PR #9647, opened 2017-01-28
Skip RAII event tests if libevent is built without event_set_mem_functions

### LINKED ISSUE #9493: event_set_mem_functions is not available on all libevents
#9387 added a new test that uses libevent's event_set_mem_functions which is sometimes not included with libevent. In particular, Gentoo only enables this when libevent is installed with the "debug" option.

Currently, this causes the build to simply fail on tests:

```
test/raii_event_tests.cpp: In member function ‘void raii_event_tests::raii_event_creation::test_method()’:
test/raii_event_tests.cpp:39:58: error: ‘event_set_mem_functions’ was not declared in this scope
     event_set_mem_functions(tag_malloc, realloc, tag_free);
                                                          ^
test/raii_event_tests.cpp: In member function ‘void raii_event_tests::raii_event_order::test_method()’:
test/raii_event_tests.cpp:68:58: error: ‘event_set_mem_functions’ was not declared in this scope
     event_set_mem_functions(tag_malloc, realloc, tag_free);
                                                          ^
Makefile:8355: recipe for target 'test/test_test_bitcoin-raii_event_tests.o' failed
```

Probably we should skip this test if libevent doesn't support overriding memory functions.

(event.h defines EVENT_SET_MEM_FUNCTIONS_IMPLEMENTED if it is available.)

luke-jr:
Fixes #9493 

I couldn't find a sane way to just mark the tests for skipping. :(

### DIFF
--- a/src/test/raii_event_tests.cpp
+++ b/src/test/raii_event_tests.cpp
@@ -3,6 +3,10 @@
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
 #include <event2/event.h>
+
+#ifdef EVENT_SET_MEM_FUNCTIONS_IMPLEMENTED
+// It would probably be ideal to define dummy test(s) that report skipped, but boost::test doesn't seem to make that practical (at least not in versions available with common distros)
+
 #include <map>
 #include <stdlib.h>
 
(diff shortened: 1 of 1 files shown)


MarcoFalke:
utACK 95f97f4, c.f. https://github.com/libevent/libevent/blob/release-2.1.8-stable/include/event2/event.h#L1547

laanwj:
utACK 95f97f4
(well, half-tested - I tested that the test is still executed when the functionality is available, I can't test the other case but I'm sure @luke-jr did)

PR #9649, opened 2017-01-28
[doc] Remove unused clang format dev script

maflcko:
The script was never used and is not maintained. Better to move it to the maintainer repo some day...

Also, update the clang format file to reflect the current coding style mentioned in the developer notes.

PR #9632, opened 2017-01-25
Add clang_static_analysis.py to help automate static analysis runs

isle2983:
In the same vein as `copyright_header.py` (#9459)  and `basic_style.py` (#9603), this script provides a consistent way to run clang static analysis checking on the codebase.

```
$ contrib/devtools/clang_static_analysis.py -h
usage: clang_static_analysis.py [-h] [-b BIN_PATH] [-r REPORT_PATH] [-j JOBS]
                                {report,check} repository

A utility for running clang static analysis on the codebase in a consistent
way.

positional arguments:
  {report,check}        Selects the output behavior. 'report' generates a
                        summary report on the issues found. 'check' compares
                        the state of the repository against a standard,
                        provides a return code for the shell and prints more
                        specific details when issues are found.
  repository            A source code repository for which the static analysis
                        is to be performed upon.

optional arguments:
  -h, --help            show this help message and exit
  -b BIN_PATH, --bin-path BIN_PATH
                        The path holding 'scan-build' and 'scan-view'
                        binaries. (Uses 'scan-build' and 'scan-view' installed
                        in PATH by default)
  -r REPORT_PATH, --report-path REPORT_PATH
                        The path for scan-build to write its report files.
                        (default=/tmp/bitcoin-scan-build/)
  -j JOBS, --jobs JOBS  The number of parallel jobs to run with 'make'.
                        (default=6)
```

The `report` subcommand is for providing a summary of the state of the checked-out branch. On current `master`, the output looks like this:
```
$ contrib/devtools/clang_static_analysis.py report .
Running command:     make clean
Running command:     /usr/share/clang/scan-build-3.5/scan-build -k -plist-html --keep-empty -o /tmp/bitcoin-scan-build/ make -j6
stderr/stdout to:    scan_build.log
This might take a few minutes...
Done.
Results in:          /tmp/bitcoin-scan-build/2017-01-25-140103-6556-1

--------------------------------------------------------------------------------
Took 143.30 seconds to analyze with scan-build
Found 3 issues:
--------------------------------------------------------------------------------
0: bench/bench.cpp:97:52 - Division by zero
1: qt/paymentrequest.pb.cc:51:24 - Called C++ object pointer is null
2: ./src/ecmult_gen_impl.h:153:5 - Value stored to 'bits' is never read
--------------------------------------------------------------------------------
Full details can be seen in a browser by running:
[...]

PR #9603, opened 2017-01-20
Add basic_style.py to automate some style checking.

isle2983:
As discussed recently in #9581 and in referenced past discussions #6839, #6156, and probably a bunch of other places too, there is concern about the long tail of effort and risk involved with these style fix-up PRs. To summarize what has been articulated by the experienced maintainers:

1) it is a lot of PR clutter for reviewers/maintainers to sift through
2) it impedes release management since style fix-ups often have difficult merge conflicts with more important content
3) they are sometimes hard to review and bugs can slip in
4) they permanently clutter up output from `git blame` and other history tools

My suggestion is that the ideal long-term strategy is to automate detection of these style issues as much as possible such that the CI system can auto-check every PR. This will ensure that the code is kept clean and consistent at source and is kept that way for all time.

Additionally, the message I am getting from what has been written is that the strategy for bringing this online needs to be done **properly** or **not at all**. It will require a conscientious effort to minimize the impact to ongoing activities and steps should be taken to make damn sure it is not the cause of any bugs. I have some thoughts, but collecting ideas from all around is good (and thanks for all that has already been written, it is quite helpful for thinking this through). I am content leaving this PR to sit until a larger collection of automation scripts are complete and we have a strategy that the maintainers are on board with.

This PR provides an example script for very basic style checks which are expressed as simple regular expressions:

```
STYLE_RULES = [
    {'title':   'No tabstops',
     'applies': ['*.c', '*.cpp', '*.h', '*.py', '*.sh'],
     'regex':   '\t',
     'fix':     '    '},
    {'title':   'No trailing whitespace on a line',
     'applies': ['*.c', '*.cpp', '*.h', '*.py', '*.sh'],
     'regex':   ' \n',
     'fix':     '\n'},
    {'title':   'No more than three consecutive newlines',
     'applies': ['*.c', '*.cpp', '*.h', '*.py', '*.sh'],
     'regex':   '\n\n\n\n',
     'fix':     '\n\n\n'},
    {'title':   'Do not end a line with a semicolon',
     'applies': ['*.py'],
     'regex':   ';\n',
     'fix':     '\n'},
    {'title':   'Do not end a line with two semicolons',
     'applies': ['*.c', '*.cpp', '*.h'],
     'regex':   ';;\n',
     'fix':     ';\n'},
]
```

[...]

PR #9459, opened 2017-01-02
Improvements to copyright_header.py and some minor copyright header tweaks.

isle2983:
This is a re-submission of PR #9453 with the TravisCI integration and language removed based on feedback.

The changes are a significant iteration to `contrib/devtools/copyright_header.py`. Like the preceding version, this script is able to parse through the source files and report on the state of the copyright headers as well as update the year.

The first major improvement is that it now has a ruleset specific for each subtree to detect the local proper header style (the subtrees being secp256k1, leveldb, univalue and ctaes).

The second major improvement is that it now makes more elegant use of Python's regular expressions to detect and manipulate the header text. This approach should be maintainable and extensible while keeping the script's execution time quick.

The third major improvement is the addition of a `check` command which gives a positive or negative acceptance for the state of the repository according to the rules of the script. When issues are found, it also gives more direct suggestions for how to resolve. The purpose is to better assist periodic maintenance of the state of the copyright headers.

Also, some minor tweaks are done to make the set of copyright headers more consistent. The addition of the MIT license header to `contrib/dev/tools/gen-manpages.sh` was previously ACKed by @nomnombtc in #9453.

PR #9658, opened 2017-01-31
Add clang_format.py to help automate code style analysis

isle2983:
The objective here is to expose data to inform the code style discussion. The tool makes it easy to change the `clang-format` binary and `.clang_format` file to help reason about the problem.

`clang_format.py` is added with the usage:

```
$ contrib/devtools/clang_format.py -h
usage: clang_format.py [-h] [-b BIN_PATH] [-s STYLE_FILE] [-j JOBS] [-f]
                       {report,check,format} [target [target ...]]

A utility for invoking clang-format to observe the state of C++ code
formatting in the repository. It produces reports of style metrics and also
can apply formatting.

positional arguments:
  {report,check,format}
                        Selects the action to be taken. 'report' produces a
                        report with analysis of the selected files taken as a
                        group. 'check' validates that the selected files match
                        the style and gives a per-file report and returns a
                        non-zero bash status if there are any format issues
                        discovered. 'format' applies the style formatting to
                        the selected files.
  target                A list of files and/or directories that select the
                        subset of files for this action. If a directory is
                        given as a target, all files contained in it and its
                        subdirectories are recursively selected. All targets
                        must be tracked in the same git repository clone.
                        (default=The current directory)

optional arguments:
  -h, --help            show this help message and exit
  -b BIN_PATH, --bin-path BIN_PATH
                        The path to the clang-format binary to be used.
                        (default=clang-format-[0-9]\.[0-9] installed in PATH
                        with the highest version number)
  -s STYLE_FILE, --style-file STYLE_FILE
                        The path to the style file to be used. (default=The
                        src/.clang_format file of the repository which holds
                        the targets)
  -j JOBS, --jobs JOBS  Parallel jobs for computing diffs. (default=6)
  -f, --force           Force proceeding with 'check' or 'format' if clang-
                        format doesn't support all parameters in the style
                        file. (default=False)
```

The `report` command looks like this when run against `master`:

```
$ contrib/devtools/clang_format.py report
--------------------------------------------------------------------------------
1466 files tracked according to 'git ls-files'
 406 files in scope according to SOURCE_FILES and ALWAYS_IGNORE settings
 406 files examined according to listed targets
--------------------------------------------------------------------------------
clang-format bin:         /usr/lib/llvm-3.8/bin/clang-format
clang-format version:     3.8.0
Using style in:           /home/isle/muhbitcoin/src/.clang-format
--------------------------------------------------------------------------------
Parallel jobs for diffs:   6
Elapsed time:              38.60s
Slowest diffs:
 33.88s for src/chainparamsseeds.h
  3.45s for src/qt/bitcoinstrings.cpp
--------------------------------------------------------------------------------
Files 100% matching:             47
Files <100% matching:           359
Formatted content MD5:      91ee2c53ca93ccca584457770dbf606c
--------------------------------------------------------------------------------
Files 90%-99% matching:         132
Files 80%-89% matching:         151
Files 70%-79% matching:          46
Files 60%-69% matching:          14
Files 50%-59% matching:          10
Files 40%-49% matching:           1
Files 30%-39% matching:           1
Files 20%-29% matching:           1
Files 10%-19% matching:           1
Files  0%- 9% matching:           2
--------------------------------------------------------------------------------

 +-------+          +------------+--------+---------+-----------+-------------+
 | score |          | pre-format |  added | removed | unchanged | post-format |
 +-------+  +-------+------------+--------+---------+-----------+-------------+
 |  83%  |  | lines |     109469 |  16807 |   17917 |     91552 |      108359 |
 +-------+  +-------+------------+--------+---------+-----------+-------------+

--------------------------------------------------------------------------------
```

It also will take a list of specific files to narrow it to a specific set instead of the entire repo.

The `check` subcommand output displays a specific report for each non-matching file in the set of targets and provides a status code to the shell:

```
$ contrib/devtools/clang_format.py check src/init.cpp
--------------------------------------------------------------------------------
1466 files tracked according to 'git ls-files'
 406 files in scope according to SOURCE_FILES and ALWAYS_IGNORE settings
   1 files examined according to listed targets
--------------------------------------------------------------------------------
A code format issue was detected in src/init.cpp
 +-------+          +------------+--------+---------+-----------+-------------+
 | score |          | pre-format |  added | removed | unchanged | post-format |
 +-------+  +-------+------------+--------+---------+-----------+-------------+
 |  90%  |  | lines |       1641 |    151 |     163 |      1478 |        1629 |
 +-------+  +-------+------------+--------+---------+-----------+-------------+
--------------------------------------------------------------------------------
These files can be auto-formatted by running:
$ contrib/devtools/clang_format.py format [target [target ...]]
--------------------------------------------------------------------------------
*** Format issues found!

```

The `format` subcommand will apply `clang-format` to set of targets much like the predecessor `clang-format.py` (which was recently removed in #9649).

The usage is overloaded for convenience:

To limit the reporting to just `src/qt/`:
```
$ contrib/devtools/clang_format.py report src/qt/
```

To check three random files:
```
$ contrib/devtools/clang_format.py check src/foo.cpp src/bar.h src/baz.cpp
```

To apply format to a single random file and also all files under a subdir:

```
$ contrib/devtools/clang_format.py format src/fizz.cpp src/buzz/
```

I have tested with all versions 3.4 through 3.9.0 on Ubuntu 14.04 and most of the same up to 3.9.1 on Debian 8. The applied formatting does have small differences between major versions (e.g. different result between 3.8.0 and 3.9.0 but the same result between 3.9.0 and 3.9.1). Some of the parameters in `src/.clang_format` aren't added until version 3.6.0. The script warns if a parameter is rejected. If you use the `--force` flag, it will proceed with 'best-effort' where the rejected parameters are dropped.

This is the fourth script in a series (The others being #9459 copyright_header.py, #9603 basic_style.py and #9632 clang_static_analysis.py) that provides a `report` command to measure the state of the codebase in the particular dimension.

A suggested trajectory from here is:

1) automating a suite of scripts to generate reports with consistent and known tool dependencies - a build matrix item in TravisCI looks like a good first thought
2) exposing metrics so they can be observed and monitored.

After that, the discussion can progress to:

1) considering some bulk actions to improve the codebase - in a safe, measured, conscientious way, of course.
2) observing the 'drift' in the wrong direction as PRs are merged
3) considering enforcement of pre-merge CI criteria where it adds clear value - the `check` subcommand in these scripts show a way that this can be done.
4) considering areas where similar automation scripts might be helpful (pep8 etc.)

Comments on all this are welcome.

### DIFF
--- a/contrib/devtools/README.md
+++ b/contrib/devtools/README.md
@@ -8,6 +8,36 @@ check-doc.py
 Check if all command line args are documented. The return value indicates the
 number of undocumented args.
 
+clang\_format.py
+===================
+
+A utility for invoking clang-format to observe the state of C++ code
+formatting in the repository. It produces reports of style metrics and also can
+apply formatting. It has three subcommands - `report`, `check`, and `format`:
+
+```
+clang_format.py [-h] [-b BIN_PATH] [-s STYLE_FILE] [-j JOBS] [-f]
+                    {report,check,format} [target [target ...]]
+```
+
+The style definition file and clang-format binary can be varied to experiment and compare results.
+
+clang\_format.py report
+-----------------------
+
+Produces a report with analysis of the selected files taken as a group.
+
+clang\_format.py check
+-----------------------
+
+Validates that the selected files match the style and gives a per-file report
+and returns a non-zero bash status if there are any format issues discovered.
+
+clang\_format.py format
+-----------------------
+
+Applies the style formatting to the selected files.
+
 clang-format-diff.py
 ===================
 
--- a/contrib/devtools/clang_format.py
+++ b/contrib/devtools/clang_format.py
@@ -1,0 +1,667 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+import sys
+import os
+import subprocess
+import time
+import argparse
+import re
+import fnmatch
+import difflib
+import hashlib
+from multiprocessing import Pool
+
+###############################################################################
+# settings for the set of files that this applies to
+###############################################################################
+
+SOURCE_FILES = ['*.cpp', '*.h']
+
+SOURCE_FILES_COMPILED = re.compile('|'.join([fnmatch.translate(match)
+                                             for match in SOURCE_FILES]))
+
+ALWAYS_IGNORE = [
+    # files in subtrees:
+    'src/secp256k1/*',
+    'src/leveldb/*',
+    'src/univalue/*',
+    'src/crypto/ctaes/*',
+]
+
+ALWAYS_IGNORE_COMPILED = re.compile('|'.join([fnmatch.translate(match)
+                                              for match in ALWAYS_IGNORE]))
+
+###############################################################################
+# obtain list of files in repo to examine
+###############################################################################
+
+GIT_LS_CMD = 'git ls-files'
+
+
+def git_ls():
+    out = subprocess.check_output(GIT_LS_CMD.split(' '))
+    return [f for f in out.decode("utf-8").split('\n') if f != '']
+
+
+def filename_is_to_be_examined(filename):
+    return (SOURCE_FILES_COMPILED.match(filename) and not
+            ALWAYS_IGNORE_COMPILED.match(filename))
+
+
+def get_filenames_in_scope(full_file_list):
+    return sorted([filename for filename in full_file_list if
+                   filename_is_to_be_examined(filename)])
+
+
+###############################################################################
+# file IO
+###############################################################################
+
+
+def read_file(filename):
+    file = open(os.path.abspath(filename), 'r')
+    contents = file.read()
+    file.close()
+    return contents
+
+
+def write_file(filename, contents):
+    file = open(os.path.abspath(filename), 'w')
+    file.write(contents)
+    file.close()
+
+
+###############################################################################
+# obtain formatted file
+###############################################################################
+
+
+def generate_style_arg(opts):
+    return '-style={%s}' % ', '.join(["%s: %s" % (k, v) for k, v in
+                                      opts.style_params.items()])
+
+
+UNKNOWN_KEY_REGEX = re.compile("unknown key '(?P<key_name>\w+)'")
+
+
+def parse_unknown_key(err):
+    if len(err) == 0:
+        return 0, None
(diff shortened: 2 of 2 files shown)


jtimon:
This seems related to #9649 ping @MarcoFalke

MarcoFalke:
The rationale for removing the predecessor was that (maybe) people
were confused which clang format script is meant for day-to-day dev
activity, so they used none. Lets hope this changes after we had the
discussion in last weeks meeting and after deleting the unused script.
So there is only one script left with a description on how to use it.
Though, I am not interested in enforcing the use of clang format, also
I won't highlight any style nits in pulls, as I think it would be
distracting.

In regard to this pull, I think this is better placed in some meta
repository, so we can use it whenever appropriate but don't confuse
regular developers by the number of scripts we put in the devtools
subdir.

I submitted a pull request to the maintainers repo to archive the old
version: https://github.com/bitcoin-core/bitcoin-maintainer-tools/pull/7
I think your script does fit really well into that repo as well.

isle2983:
I support keeping `devtools/` uncluttered for things that are clearly needed to be understood by everyone, so I am on board with it not going there at present time.

In the long run, my hope would be that every code-changing PR gets massaged developer-side prior to submission and everyone does a clean TravisCI run on the branch in their forked repo prior to submitting the PR. An efficient workflow to cut down on TravisCI round trips would also include running a pass of the same scripts on individual work machines, so having them easily available is a plus. To me, that is more of a 'developer' task rather than a 'maintainer' task - albeit an optional one at present.

The next task I am looking at is to bring the four scripts I have written into one subdirectory modeled after the rpc tests in `qa/`. I wish to consolidate the repeated code and make better OO abstractions for future extensions. Also, for the same PR, I hope to have it all working inside TravisCI in a report-information-only capacity. At that time we could decide to make it live in a different repo if it isn't too much trouble importing it for TravisCI runs on `bitcoin/bitcoin`.

laanwj:
Please add scripts like this to the maintainer tools: https://github.com/bitcoin-core/bitcoin-maintainer-tools

There are meta-tools that are not part of the Bitcoin Core release schedule and branching, so have no place in the repository. `qa/devtools` could be seen as a historical error in that regard, but we'll keep some often-used scripts there.

laanwj:
> In the long run, my hope would be that every code-changing PR gets massaged developer-side prior to submission and everyone does a clean TravisCI run on the branch in their forked repo prior to submitting the PR.

Sure, I completely agree that in the future, if it gets used by every developer it can be moved to somewhere inside the tree if that is more convenient.

But right now, this script has 0 (or 1) users. So let's add it to the maintainer tools first and see how that goes.

PR #9670, opened 2017-02-02
contrib: github-merge improvements

laanwj:
Some minor github-merge improvements I've made over time:

User interface:

- Print merge details again before signing off, to refresh your memory -  usually at least I'll have done lots of different things in the shell so this will have scrolled out a long time ago.

- Require a valid answer on the prompts. One of the requested answers must be typed, if not, the prompt will re-ask. This prevents accidentally rejecting.

Efficiency:

- Condense "accept merge" and "sign off" prompts. There's no reason to have this as two separate prompts, both are just opportunities to skip out on the merge, no action is performed in between.

Merging:

- Strip spaces from github title. This avoids redundant spaces surrounding it from getting into the commit message.

### DIFF
--- a/contrib/devtools/github-merge.py
+++ b/contrib/devtools/github-merge.py
@@ -1,5 +1,5 @@
 #!/usr/bin/env python3
-# Copyright (c) 2016 The Bitcoin Core developers
+# Copyright (c) 2016-2017 Bitcoin Core Developers
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
@@ -15,7 +15,7 @@
 # In case of a clean merge that is accepted by the user, the local branch with
 # name $BRANCH is overwritten with the merged result, and optionally pushed.
 from __future__ import division,print_function,unicode_literals
-import os
+import os,sys
 from sys import stdin,stdout,stderr
 import argparse
 import hashlib
@@ -127,6 +127,9 @@ def tree_sha512sum(commit='HEAD'):
         raise IOError('Non-zero return value executing git cat-file')
     return overall.hexdigest()
 
+def print_merge_details(pull, title, branch, base_branch, head_branch):
+    print('%s#%s%s %s %sinto %s%s' % (ATTR_RESET+ATTR_PR,pull,ATTR_RESET,title,ATTR_RESET+ATTR_PR,branch,ATTR_RESET))
+    subprocess.check_call([GIT,'log','--graph','--topo-order','--pretty=format:'+COMMIT_FORMAT,base_branch+'..'+head_branch])
 
 def parse_arguments():
     epilog = '''
@@ -171,7 +174,7 @@ def main():
     info = retrieve_pr_info(repo,pull)
     if info is None:
         exit(1)
-    title = info['title']
+    title = info['title'].strip()
     # precedence order for destination branch argument:
     #   - command line argument
     #   - githubmerge.branch setting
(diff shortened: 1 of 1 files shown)


MarcoFalke:
This is not a dev script, so it might be better located in the maintainer repo. OTOH, this is probably the most critical maintainer script, so keeping it here to have more eyes watching it, is a good thing.

MarcoFalke:
utACK 90f97df regardless.

fanquake:
NACK moving this to another repo.

utACK changes.

laanwj:
> This is not a dev script, so it might be better located in the maintainer repo. OTOH, this is probably the most critical maintainer script, so keeping it here to have more eyes watching it is a good thing.

Yes I've thought about that. We could have the main development copy in the maintainer-tools repo, and sync it to Bitcoin Core once in a while.

I use the tool with a few different repositories so I have it installed system-wide and hardly ever use the one in the Bitcoin Core tree.

contrib/devtools/github-merge.py
```diff
@@ -69,6 +69,10 @@ def ask_prompt(text):
     print("",file=stderr)
     return reply
 
+def print_merge_details(pull, title, branch, base_branch, head_branch):
+    print('%s#%s%s %s %sinto %s%s' % (ATTR_RESET+ATTR_PR,pull,ATTR_RESET,title,ATTR_RESET+ATTR_PR,branch,ATTR_RESET))
```
JeremyRubin: nit: maybe a bit nicer with some python3 isms

```python
print('{color_pr}#{pr}{nocolor} {title}{color_pr} into {branch}{nocolor}'.format(color_pr = ATTR_RESET+ATTR_PR, nocolor=ATTR_RESET, **locals()))
```

contrib/devtools/github-merge.py
```diff
@@ -242,9 +236,13 @@ def main():
         subprocess.call([GIT,'branch','-q','-D',local_merge_branch],stderr=devnull)
 
     # Push the result.
-    reply = ask_prompt("Type 'push' to push the result to %s, branch %s." % (host_repo,branch))
-    if reply.lower() == 'push':
-        subprocess.check_call([GIT,'push',host_repo,'refs/heads/'+branch])
+    while True:
+        reply = ask_prompt("Type 'push' to push the result to %s, branch %s, or 'x' to exit without pushing." % (host_repo,branch)).lower()
```
JeremyRubin: May be nice to change this to be "type the remote and branch to confirm" so that you are forced to actually read where you're pushing it :)

JeremyRubin: in a similar vein, may be good to also add an additional confirm if the branch is prefixed with a +, ie, a force push.

JeremyRubin (approved):
utack 90f97df.

A couple suggestions on further improvements, but those are addressable in another PR if desired.

contrib/devtools/github-merge.py
laanwj: Agree, but I didn't change this code in this pull only moved it.

contrib/devtools/github-merge.py
laanwj: > May be nice to change this to be "type the remote and branch to confirm" so that you are forced to actually read where you're pushing it :)

Not sure this will help much. I don't think I've ever pushed something to the wrong place with this script after I made it take the target branch from the pull metadata on github.

> in a similar vein, may be good to also add an additional confirm if the branch is prefixed with a +, ie, a force push.

We don't support force-pushing from this script, and force pushes to master and all the version branches are disabled on github.

PR #9690, opened 2017-02-05
Change 'Clear' button string to 'Reset'

da2x:
Easier to understand what the button does (it resets the graph view).

'Clear' might mean that the graph is emptied and stops updating, whereas its easier to see that you're just starting fresh with 'Reset'.

### DIFF
--- a/src/qt/forms/debugwindow.ui
+++ b/src/qt/forms/debugwindow.ui
@@ -645,7 +645,7 @@
            <item>
             <widget class="QPushButton" name="btnClearTrafficGraph">
              <property name="text">
-              <string>&amp;Clear</string>
+              <string>&amp;Reset</string>
              </property>
              <property name="autoDefault">
               <bool>false</bool>


jonasschnelli:
trivial ACK 9155241bd9b0de18f30a5589351bdcbd0c6583a6 but no strong opinion wether this solution is better/clearer.

da2x:
Users, keep in mind the many that have English UI but are not English native speakers, should be more familiar with ‘Reset’ than 'Clear'. It appears in many programs and on devices, and as it’s the default string for `<input type=reset>` which alone should mean people have seen this a lot on web forms. I’m not saying the two strings don’t mean almost the same thing, but in my opinion more people around the world will understand ‘Reset’ in this context.

laanwj:
In any case this will have to wait until 0.14 is branched off, the project is in a pre-release translation strings freeze at the moment.

MarcoFalke:
Jup, and for 0.15 we might rework this whole tab in the debug window
anyway. (Hoping that jonas's statistics collector is merged)

PR #9691, opened 2017-02-05
Init ECC context for test_bitcoin_fuzzy.

gmaxwell:
This avoids calling things like pubkey_parse with a null context argument.


laanwj:
eh yes good catch, obvious utACK ac719c9

PR #7826, opened 2016-04-06
[Qt] show conflicts of unconfirmed transactions in the UI

jonasschnelli:
Currently, the GUI shows only conflicts with transaction in a block.

This results in displaying replaced transaction (RBF) wrong (the replaced transaction is no longer in the mempool, but still in the wallet-transactionlist, conflicts are not detected).

This PR will mark unconfirmed transactions that are **not** in our mempool and have conflicts as conflicted.

PR #9697, opened 2017-02-06
[Qt] simple fee bumper with user verification

jonasschnelli:
~~Based on #9681.~~
Add "Increase transaction fee" action to the transaction table context menu.
The user can verify the new (and old) fee before sign & commit.

bildschirmfoto 2017-02-06 um 15 31 34
bildschirmfoto 2017-02-06 um 15 31 36

### DIFF
--- a/src/qt/sendcoinsdialog.h
+++ b/src/qt/sendcoinsdialog.h
@@ -100,13 +100,14 @@ Q_SIGNALS:
 };
 
 
+#define SEND_CONFIRM_DELAY   3
 
 class SendConfirmationDialog : public QMessageBox
 {
     Q_OBJECT
 
 public:
-    SendConfirmationDialog(const QString &title, const QString &text, int secDelay = 0, QWidget *parent = 0);
+    SendConfirmationDialog(const QString &title, const QString &text, int secDelay = SEND_CONFIRM_DELAY, QWidget *parent = 0);
     int exec();
 
 private Q_SLOTS:
--- a/src/qt/transactionview.cpp
+++ b/src/qt/transactionview.cpp
@@ -11,6 +11,7 @@
 #include "guiutil.h"
 #include "optionsmodel.h"
 #include "platformstyle.h"
+#include "sendcoinsdialog.h"
 #include "transactiondescdialog.h"
 #include "transactionfilterproxy.h"
 #include "transactionrecord.h"
@@ -37,7 +38,7 @@
 
 TransactionView::TransactionView(const PlatformStyle *platformStyle, QWidget *parent) :
     QWidget(parent), model(0), transactionProxyModel(0),
-    transactionView(0), abandonAction(0), columnResizingFixer(0)
+    transactionView(0), abandonAction(0), bumpFeeAction(0), columnResizingFixer(0)
 {
     // Build filter row
     setContentsMargins(0,0,0,0);
@@ -138,6 +139,7 @@ TransactionView::TransactionView(const PlatformStyle *platformStyle, QWidget *pa
 
     // Actions
     abandonAction = new QAction(tr("Abandon transaction"), this);
+    bumpFeeAction = new QAction(tr("Increase transaction fee"), this);
     QAction *copyAddressAction = new QAction(tr("Copy address"), this);
     QAction *copyLabelAction = new QAction(tr("Copy label"), this);
     QAction *copyAmountAction = new QAction(tr("Copy amount"), this);
@@ -156,6 +158,7 @@ TransactionView::TransactionView(const PlatformStyle *platformStyle, QWidget *pa
     contextMenu->addAction(copyTxPlainText);
     contextMenu->addAction(showDetailsAction);
     contextMenu->addSeparator();
+    contextMenu->addAction(bumpFeeAction);
     contextMenu->addAction(abandonAction);
     contextMenu->addAction(editLabelAction);
 
@@ -173,6 +176,7 @@ TransactionView::TransactionView(const PlatformStyle *platformStyle, QWidget *pa
     connect(view, SIGNAL(doubleClicked(QModelIndex)), this, SIGNAL(doubleClicked(QModelIndex)));
     connect(view, SIGNAL(customContextMenuRequested(QPoint)), this, SLOT(contextualMenu(QPoint)));
 
+    connect(bumpFeeAction, SIGNAL(triggered()), this, SLOT(bumpFee()));
     connect(abandonAction, SIGNAL(triggered()), this, SLOT(abandonTx()));
     connect(copyAddressAction, SIGNAL(triggered()), this, SLOT(copyAddress()));
     connect(copyLabelAction, SIGNAL(triggered()), this, SLOT(copyLabel()));
@@ -372,6 +376,7 @@ void TransactionView::contextualMenu(const QPoint &point)
     uint256 hash;
     hash.SetHex(selection.at(0).data(TransactionTableModel::TxHashRole).toString().toStdString());
     abandonAction->setEnabled(model->transactionCanBeAbandoned(hash));
+    bumpFeeAction->setEnabled(model->transactionSignalsRBF(hash));
 
     if(index.isValid())
     {
@@ -395,6 +400,21 @@ void TransactionView::abandonTx()
 
     // Update the table
     model->getTransactionTableModel()->updateTransaction(hashQStr, CT_UPDATED, false);
+}
+
+void TransactionView::bumpFee()
+{
+    if(!transactionView || !transactionView->selectionModel())
+        return;
+    QModelIndexList selection = transactionView->selectionModel()->selectedRows(0);
+
+    // get the hash from the TxHashRole (QVariant / QString)
+    uint256 hash;
+    QString hashQStr = selection.at(0).data(TransactionTableModel::TxHashRole).toString();
+    hash.SetHex(hashQStr.toStdString());
+
+    // Bump tx fee over the walletModel
+    model->bumpFee(hash);
 }
 
 void TransactionView::copyAddress()
--- a/src/qt/transactionview.cpp
+++ b/src/qt/transactionview.cpp
@@ -414,10 +414,10 @@ void TransactionView::bumpFee()
     hash.SetHex(hashQStr.toStdString());
 
     // Bump tx fee over the walletModel
-    model->bumpFee(hash);
-
-    // Update the table
-    model->getTransactionTableModel()->updateTransaction(hashQStr, CT_UPDATED, false);
+    if (model->bumpFee(hash)) {
+        // Update the table
+        model->getTransactionTableModel()->updateTransaction(hashQStr, CT_UPDATED, false);
+    }
 }
 
 void TransactionView::copyAddress()
--- a/src/qt/transactionview.cpp
+++ b/src/qt/transactionview.cpp
@@ -415,6 +415,9 @@ void TransactionView::bumpFee()
 
     // Bump tx fee over the walletModel
     model->bumpFee(hash);
+
+    // Update the table
+    model->getTransactionTableModel()->updateTransaction(hashQStr, CT_UPDATED, false);
 }
 
 void TransactionView::copyAddress()
--- a/src/qt/walletmodel.cpp
+++ b/src/qt/walletmodel.cpp
@@ -712,7 +712,7 @@ bool WalletModel::bumpFee(uint256 hash)
     std::unique_ptr<CFeeBumper> feeBump;
     {
         LOCK2(cs_main, wallet->cs_wallet);
-        feeBump.reset(new CFeeBumper(wallet, hash, 0, false, 0, true));
+        feeBump.reset(new CFeeBumper(wallet, hash, nTxConfirmTarget, false, 0, true));
     }
     if (feeBump->getResult() != BumpFeeResult::OK)
     {
--- a/src/qt/walletmodel.cpp
+++ b/src/qt/walletmodel.cpp
@@ -8,8 +8,10 @@
 #include "consensus/validation.h"
 #include "guiconstants.h"
 #include "guiutil.h"
+#include "optionsmodel.h"
 #include "paymentserver.h"
 #include "recentrequeststablemodel.h"
+#include "sendcoinsdialog.h"
 #include "transactiontablemodel.h"
 
 #include "base58.h"
@@ -17,15 +19,18 @@
 #include "keystore.h"
 #include "validation.h"
 #include "net.h" // for g_connman
+#include "policy/rbf.h"
 #include "sync.h"
 #include "ui_interface.h"
 #include "util.h" // for GetBoolArg
+#include "wallet/feebumper.h"
 #include "wallet/wallet.h"
 #include "wallet/walletdb.h" // for BackupWallet
 
 #include <stdint.h>
 
 #include <QDebug>
+#include <QMessageBox>
 #include <QSet>
 #include <QTimer>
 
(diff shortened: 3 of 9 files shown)


src/qt/walletmodel.cpp
```diff
@@ -692,6 +694,34 @@ bool WalletModel::abandonTransaction(uint256 hash) const
     return wallet->AbandonTransaction(hash);
 }
 
+bool WalletModel::transactionSignalsRBF(uint256 hash) const
+{
+    LOCK2(cs_main, wallet->cs_wallet);
+    const CWalletTx *wtx = wallet->GetWalletTx(hash);
+    if (wtx || SignalsOptInRBF(*wtx))
```
ryanofsky: wtx && SignalsOptInRBF(*wtx)

src/qt/walletmodel.cpp
```diff
@@ -692,6 +694,34 @@ bool WalletModel::abandonTransaction(uint256 hash) const
...
+bool WalletModel::bumpFee(uint256 hash, const std::function<bool(const CTransaction& tx, const CAmount& oldFee, const CAmount& newFee)> verificationCallback) const
+{
+    LOCK2(cs_main, wallet->cs_wallet);
+    CAmount nOldFee(0);
+    CAmount nNewFee(0);
+    std::shared_ptr<CWalletTx> wtxRef;
+    std::vector<std::string> vErrors;
+
+    // do a simple feebump with default confirmation target
+    CWallet::BumpFeeResult res = wallet->BumpFee(hash, 0, false, 0, true, nOldFee, nNewFee, wtxRef, vErrors, verificationCallback);
```
ryanofsky: I'm surprised it's ok to be holding onto the cs_main and cs_wallet locks while waiting for user input from a confirmation dialog, since it would seem to grind the whole bitcoin node to a halt. But maybe there are other precedents for doing this.

If you wanted to avoid it, though, I think you easily could by splitting bumpfee into two functions: one which creates the new transaction, and the other which commits this.

jonasschnelli:
Thanks for the review @ryanofsky. Yes. The current situation with holding cs_main/cs_wallet during the QMessageBox must be avoided. Using `LEAVE_CRITICAL_SECTION` and `ENTER_CRITICAL_SECTION` would be another alternative, but I'd like your idea about splitting bumpFee into a part where we create the tx and a part where we sign and commit.
The signing part could potentially be shared with `CreateTransaction` in future.

jonasschnelli:
Rebased on top of #9681.
Fixed the `cs_main` & `cs_wallet` lock problems mentioned by @ryanofsky.

src/qt/walletmodel.cpp
```diff
@@ -692,6 +697,65 @@ bool WalletModel::abandonTransaction(uint256 hash) const
...
+{
+    std::shared_ptr<CFeeBumper> feeBump;
+    {
+        LOCK2(cs_main, wallet->cs_wallet);
+        feeBump = std::make_shared<CFeeBumper>(CFeeBumper(wallet, hash, 0, false, 0, true));
+    }
+    if (feeBump->getResult() != CFeeBumper::BumpFeeResult::OK)
+    {
+        QMessageBox::critical(0, tr("Fee bump error"), tr("Increasing transaction fee failed") + "<br />(" +
+            (feeBump->getErrors().size() ? QString::fromStdString(feeBump->getErrors()[0]) : "") +")");
```
ryanofsky: Are we able to write unit tests to exercise our qt interface code? I don't know very much about qt, but googling around it seems to be possible: http://doc.qt.io/qt-5/qttestlib-tutorial3-example.html

Asking because the logic here isn't totally trivial, and this seems like something we would want automated testing for to ensure it doesn't break in the future, especially if there will be more refactorings.

src/qt/walletmodel.cpp
```diff
@@ -692,6 +697,65 @@ bool WalletModel::abandonTransaction(uint256 hash) const
...
+    if (wtx && SignalsOptInRBF(*wtx))
+        return true;
+    return false;
+}
+
+bool WalletModel::bumpFee(uint256 hash)
+{
+    std::shared_ptr<CFeeBumper> feeBump;
+    {
+        LOCK2(cs_main, wallet->cs_wallet);
```
ryanofsky: Seems like it would be nice for this code if CFeeBumper constructor, wallet SignTransaction method, and CFeeBumper commit could acquire the locks they need themselves. Not asking for a change, since I don't want to hold up #9697, but curious if there was something that got in the way of doing this.

ryanofsky (commented):
Tested ACK f7132c3b14cf714f71a2fe59241223fc30520814.

Confirmed fee bumping, error handling, and RBF detection all work correctly. The interface is very clean and convenient, and it seems like it could be logically extended to give users control of the new fee in the future.

src/qt/walletmodel.cpp
ryanofsky: I started writing unit tests for this change: https://github.com/ryanofsky/bitcoin/commit/9bfbdd88dd0f1e11f06403e81f39fd8df1d047c7

They are mostly complete, but have a few things I want to fix up. They also required changes to the testing framework which I'm going to split off and make a separate PR.

laanwj:
Adding milestone for 0.15

src/wallet/wallet.h
```diff
@@ -791,6 +791,21 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
...
+     */
+    enum BumpFeeResult
+    {
+        BumpFeeResult_OK,
+        BumpFeeResult_INVALID_ADDRESS_OR_KEY,
+        BumpFeeResult_INVALID_REQUEST,
+        BumpFeeResult_INVALID_PARAMETER,
+        BumpFeeResult_WALLET_ERROR,
+        BumpFeeResult_MISC_ERROR,
+    };
```
jtimon: The function doesn't seem to return anything and the enum is not used anywhere it seems.
Perhaps the function could start as void in this code movement and change it afterwards?

src/wallet/wallet.cpp
```diff
@@ -2313,71 +2313,85 @@ int64_t CalculateMaximumSignedTxSize(const CTransaction &tx)
...
         // implies that we can sign for every input.
-        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Transaction contains inputs that cannot be signed");
+        return -1;
     }
     return GetVirtualTransactionSize(txNew);
 }
 
 enum CWallet::BumpFeeResult CWallet::BumpFee(uint256 txid, int newConfirmTarget, bool specifiedConfirmTarget, CAmount totalFee, bool newTxReplaceable, CAmount& nOldFee, CAmount& nNewFee, std::shared_ptr<CWalletTx>& wtxNew, std::vector<std::string>& vErrors)
 {
-    if (!pwalletMain->mapWallet.count(hash)) {
```
jtimon: Did the previous commit compile without hash?
I though we wanted every commit to compile and pass tests (for git bisect but also for general history cleanness).

src/wallet/wallet.cpp
```diff
@@ -2313,71 +2313,85 @@ int64_t CalculateMaximumSignedTxSize(const CTransaction &tx)
...
     // Calculate the expected size of the new transaction.
     int64_t txSize = GetVirtualTransactionSize(*(wtx.tx));
     const int64_t maxNewTxSize = CalculateMaximumSignedTxSize(*wtx.tx);
+    if (maxNewTxSize < 0) {
+        vErrors.push_back("Transaction contains inputs that cannot be signed");
+        return CWallet::BumpFeeResult_INVALID_ADDRESS_OR_KEY;
+    }
 
     // calculate the old fee and fee-rate
-    CAmount nOldFee = wtx.GetDebit(ISMINE_SPENDABLE) - wtx.tx->GetValueOut();
```
jtimon: This could be done when making nOldFee a parameter of this function too.

src/wallet/rpcwallet.cpp
```diff
@@ -2749,37 +2749,46 @@ UniValue bumpfee(const JSONRPCRequest& request)
...
                 break;
-            case CWallet::BumpFeeResult_WALLET_ERROR:
-                throw JSONRPCError(RPC_WALLET_ERROR, vErrors[0]);
-                break;
             default:
                 throw JSONRPCError(RPC_MISC_ERROR, vErrors[0]);
                 break;
         }
     }
+
```
jtimon: Seems weird and unnecessary to move this to wallet and then move it back here. Perhaps SignTransaction/BumpFeePrepare/BumpFeeCommit should be created before BumpFee (ie BumpFee never being created to just remove it in the same PR) ?

jtimon:
utACK individual commits: 8c6316038e8ebb1877d712f29171afc4bb76a159 641c99cb6f96afa166a293917f36f3cf8f28a08a (without fully verifying code movements, but look ok) f0c98b9bfa763fbef87f1d9d37853f28e9451c87
Fast review 65e2e0085a608d9f570bc911349a5953a6ca88ee
It seems to me that the history in this PR is unnecessarily complicated and not very clean (and I believe it doesn't compile commit by commit, but I haven't checked).
Needs rebase.

ryanofsky:
@jtimon, if you are interested in reviewing this PR, the only commit worth looking at is the final one: "[Qt] simple fee bumper with user verification" (f7132c3b14cf714f71a2fe59241223fc30520814). The preceding commits which you reviewed were just pulled in from an old version of #9681, and they will disappear after #9681 is merged and this is rebased.

If you want to review the latest version of the feebumper code in #9681, that would be very helpful. #9681 is fully up to date and its commit history has been simplified.

jonasschnelli:
Rebased (since #9681 is now merged) and overhauled.
This now also works with encrypted wallets.

flack:
It's kind of hard to see in the confirmation dialog by how much the fee will actually increase. Could you work that somehow into your sentence, or maybe add a table like this:

Current fee |  0.00010595
------ | -------
increase |  0.00000960
**New fee** |  **0.00011555**

jonasschnelli:
Implemented @flack's idea.
Now it looks like:

bildschirmfoto 2017-04-26 um 09 59 11

bildschirmfoto 2017-04-26 um 10 20 40

jtimon:
Perhaps it is too much info for the confirmation, but what about showing the current feerate and the new feerate as well?

laanwj:
@jonasschnelli  that looks good!

laanwj:
I tested this out a bit on testnet; seems to work! one nit. I bumped a single transaction a few times, and it showed it as being sent multiple times:
untitled2

This could be scary to users, as it looks like the transaction is sent successfully multiple times, and will possibly be mined multiple times.

After restart it was better:
untitled3

src/qt/walletmodel.cpp
ryanofsky: Here's an updated version of the tests I wrote previously: https://github.com/ryanofsky/bitcoin/commit/71c0a80fbfe95262be3b81c45474aceaccddfad3

Feel free to cherry-pick or I can submit as a separate PR.

ryanofsky (approved):
ACK 488ca9f9c78d7117d11e7f1f0be14d51065f7110

jonasschnelli:
What would be the best solution regarding bump conflicts?

At the moment RPCs `listtransaction` does also list all previous tx in a bump (you can see the `walletconflicts` there).

Completely hiding the conflicted transactions (only own bump conflicts) would be a solution though I would provide to just "flagging" (color or opacity) the conflicts in mempool.
I once had a PR for that: https://github.com/bitcoin/bitcoin/pull/7826

Any ideas?

src/qt/sendcoinsdialog.h
```diff
@@ -100,13 +100,14 @@ private Q_SLOTS:
...
 
+#define SEND_CONFIRM_DELAY   3
 
 class SendConfirmationDialog : public QMessageBox
 {
     Q_OBJECT
 
 public:
-    SendConfirmationDialog(const QString &title, const QString &text, int secDelay = 0, QWidget *parent = 0);
+    SendConfirmationDialog(const QString &title, const QString &text, int secDelay = SEND_CONFIRM_DELAY, QWidget *parent = 0);
```
jonasschnelli: @ryanofsky: because setting `QWidget *parent = 0` is required, that would either result in swapping `QWidget *parent = 0` with `int secDelay = SEND_CONFIRM_DELAY` or passing `SEND_CONFIRM_DELAY` in `sendcoinsdialog.cpp` (as it is by now), right?
The current situation seems to be okay to me.
Or do I misunderstand something?

jonasschnelli:
Fixed @laanwj's point.
Added a commit that ensures the table model row gets updated correctly after a bump.

src/qt/sendcoinsdialog.h
ryanofsky: > Or do I misunderstand something?

No you're right. I didn't see sendcoinsdialog is also passing a this argument, so there is no way to use the default delay argument.

src/qt/transactionview.cpp
```diff
@@ -415,6 +415,9 @@ void TransactionView::bumpFee()
 
     // Bump tx fee over the walletModel
     model->bumpFee(hash);
+
+    // Update the table
+    model->getTransactionTableModel()->updateTransaction(hashQStr, CT_UPDATED, false);
```
ryanofsky: In commit "Make sure we always update the table row after a bumpfee call"

I was surprised that passing false for the `updateTransaction` `showTransaction` parameter doesn't hide the transaction, though it's good that it doesn't. Do you think this parameter should be removed (separately, not part of this change)? I'm not seeing what it does other than set a temporary state that gets immediately overwritten.

ryanofsky (commented):
ACK 1015ef40ea37a33faeded844e2da62cc6c66b0bf

src/qt/transactionview.cpp
jonasschnelli: Heh. Yes. This is really surprising and yes, I think we should remove it (in a clean follow up PR).

src/qt/walletmodel.cpp
```diff
@@ -692,6 +697,86 @@ bool WalletModel::abandonTransaction(uint256 hash) const
...
+        res = feeBump->signTransaction(wallet);
+    }
+    if (!res) {
+        QMessageBox::critical(0, tr("Fee bump error"), tr("Can't sign transaction."));
+        return false;
+    }
+    // commit the bumped transaction
+    {
+        LOCK2(cs_main, wallet->cs_wallet);
+        res = feeBump->commit(wallet);
```
TheBlueMatt: I believe this is the first use of CFeeBumper without holding cs_wallet through the object's entire lifetime, so more thourough review of that is merited. I read through it and /think/ its OK, but it would probably be good to duplicate a few of the checks in the CFeeBumper constructor in commit, specifically the HasWalletSpend/mempool descendant count, wtx.mapValue.count("replaced_by_txid"), and GetDepthInMainChain() != 0 ones.

TheBlueMatt (commented):
As for not displaying bumps, I'd advocate for just hiding them wholesale in TransactionRecord::showTransaction (unless the original, non-bumped version got confirmed).

src/qt/walletmodel.cpp
```diff
@@ -692,6 +697,86 @@ bool WalletModel::abandonTransaction(uint256 hash) const
...
+        return true;
+    return false;
+}
+
+bool WalletModel::bumpFee(uint256 hash)
+{
+    std::unique_ptr<CFeeBumper> feeBump;
+    {
+        LOCK2(cs_main, wallet->cs_wallet);
+        feeBump.reset(new CFeeBumper(wallet, hash, 0, false, 0, true));
```
jonasschnelli: Oh. Right. Here it misses `nTxConfirmTarget`. Will fix.

jonasschnelli:
Added 3 commits to address @TheBlueMatt findings.

* Make sure to pass `nConfirmTarget` during Qt fee bumps (d9afe0192c9ad83594074661d55a4242f33de462)
* Re-check the conditions during feebump "commit". Because the Qt dialog may stay open for a couple of seconds (or even minutes), we need to re-check the descendants and if the tx has been mined (717495875b995c63e3e4166245e88e8def7f1e47). This still has room to optimise. We may want to timeout a bump...
*  Don't hide rows when the bump has been cancelled (c9c48ea36fab3c0c36665289cbb4038491aa1354)

jonasschnelli:
Rebased.

src/wallet/feebumper.cpp
```diff
@@ -248,6 +255,11 @@ bool CFeeBumper::commit(CWallet *pWallet)
     }
     CWalletTx& oldWtx = pWallet->mapWallet[txid];
 
+    // make sure the transaction still has no descendants and hasen't been mined in the meantime
```
paveljanik: typo "hasen't"

paveljanik (approved):

paveljanik:
Minor nit: can we gray "Increase transaction fee" menu entry when on transaction that was already fee-bumped?

It has no reason to allow bumping the fee when it already was fee bumped and clicking on the menu entry will show

```
Increasing transaction fee failed
(Cannot bump transaction xyz which was already bumped by transaction abc)
```

paveljanik:
... and now when my tx has been mined, there is no reason to show this menu entry at all (because clicking on it will tell me it is mined already).

laanwj:
utACK a387837

I think this is ready for merging. Sure, it's not perfect yet, but it's usable and can be improved later.
It has a reasonable number of ACKs/utACKs and nits have been fixed a few times.

Anyone disagree?

ryanofsky:
utACK a387837. I agree with merging.

Changes since my last review were just rebase on updated master and 3 new commits.

PR #9699, opened 2017-02-06
Remove contrib/debian/patches subdirectory

droark:
The directory hasn’t been touched in five years and seems like something that could be added again later if needed. Remove the subdirectory.


practicalswift:
ACK ca04070

TheBlueMatt:
Have you tested if the Debian build still succeeds if these files are missing? I'd bet they don't.

laanwj:
What I sometimes wonder: is anyone using the stuff in `contrib/debian` to do debian builds at all?
Is it used for the PPAs?

TheBlueMatt:
It is (roughly) used for the PPAs, AFAIK that is the only user.

laanwj:
Okay, thanks. Yes that's enough reason to keep it around.

TheBlueMatt:
Yea, I would say trash it, but we've gotten contributors to update the PPA through it, which helps me be super lazy :).

fanquake:
Closing, seems we want to keep this.

laanwj:
Well, yes, this directory (and the accompanying README file) is a standard fixture in `contrib/debian` directories, it doesn't hurt to keep it around, though it's unlikely any patches will be added there as we could just as well patch the project itself as it's part of the same tree.

droark:
Thanks, everyone. I'll add a PR a little later that updates the README. Might as well ensure that people know to leave the directory alone. :)

PR #8456, opened 2016-08-04
[RPC] Simplified bumpfee command.

mrbandrews:
This purports to be a simplified "bumpfee" RPC command.

This pull was motivated by the discussion in #7865, where at the end, the idea was floated of submitting a simplified version of the code. I put a little thought into how a simple "bumpfee" command should work and here is what I came up with:

1) The user should specify which output to decrement in order to pay increased fee. This saves us from trying to figure out which address(es) are "change" addresses. This is preferable not only because it simplifies the code, but because as far as I can tell, the code that identifies change outputs is not particularly robust and may be modified in the future. If it's desirable to have user-friendly code that figures out which output to decrement, perhaps that logic could be placed in the GUI?

2) The command should not add new inputs. In other words, if the user can't identify an output that can be decremented in an amount sufficient to bump the fee then the command should fail. It seems likely that the vast majority of the time, identifying such an output would not be a problem. Adding new inputs complicates the code because the size of the transaction increases, perhaps significantly, so then the user would have to pay more fee for that reason as well, as opposed to just bumping the fee on the tx as it currently exists.

3) If the tx has descendant transactions, the bumped tx has to pay the fees for all of the descendants since they will get evicted from the mempool, and the rule as I understand it is that replacing a tx cannot cause the total fees in the mempool to go down.

So, the required inputs are the txid and the output index.

The optional inputs are:

```
confirmTarget - perhaps the reason the user feels the need to bump the fee is because the original confirmTarget was too slow
totalFee - for full control, the user can simply specify how many satoshis to pay as a fee on the bumped transaction (this may be useful when the parent tx has children tx'es)
```

This pull includes a python test.

PR #9701, opened 2017-02-06
Make bumpfee tests less fragile

ryanofsky:
Fix test fragility described in https://github.com/bitcoin/bitcoin/pull/8456#issuecomment-272256286

>I ran into a lot of spurious errors related to `fundrawtransaction` in the bumpfee tests while trying to add a simple test for this. fundrawtransaction would sometimes chose different sized inputs than the `small_output_fails` and `dust_to_fee tests` were expecting, so I rewrote these tests in 13960f870c1ba94f391012f9c9243c59951ea6ab to manually create RBF transactions with the right inputs.
>
>I also ran into problems with `fundrawtransaction` using up all the right-sized inputs before these tests could use them, so I changed the order of the tests so they would run before any fundrawtransaction calls.
>
>I also ran into problems with `fundrawtransaction` sometimes chosing a higher fee for the RBF transaction than the `test_rebumping` test expected, causing the bumpfee command in the test to fail with an "Invalid totalFee, must be at least ..." error. I did not figure out what was causing this, but it seemed to somehow be related to the number of `peer_node.sendtoaddress(rbf_node_address, 0.001)` calls made during the test setup.


sdaftuar:
utACK 145b702

qa/rpc-tests/bumpfee.py
```diff
@@ -259,51 +249,31 @@ def test_unconfirmed_not_spendable(rbf_node, rbf_node_address):
...
-    signedtx = node.signrawtransaction(fundtx["hex"])
-    txid = node.sendrawtransaction(signedtx["hex"])
-    return txid
-
-
-def spend_one_input(node, input_amount, outputs):
-    input = dict(sequence=BIP125_SEQUENCE_NUMBER, **next(u for u in node.listunspent() if u["amount"] == input_amount))
-    rawtx = node.createrawtransaction([input], outputs)
+def spend_one_input(node, dest_address):
+    input = dict(
```
jnewbery: Please don't use input as a variable name. input is a built-in Python function. Perhaps `txin` is better?

qa/rpc-tests/bumpfee.py
```diff
@@ -165,66 +165,56 @@ def test_bumpfee_with_descendant_fails(rbf_node, rbf_node_address, dest_address)
...
-    rbfid = create_fund_sign_send(rbf_node, {dest_address: 0.00090000})
+    rbfid = spend_one_input(rbf_node, dest_address)
     rbftx = rbf_node.gettransaction(rbfid)
-    rbf_node.settxfee(Decimal("0.00002500"))
+    feerate = Decimal("0.00025000")
+    rbf_node.settxfee(feerate)
     bumped_tx = rbf_node.bumpfee(rbfid)
-    assert bumped_tx["fee"] > 2 * abs(rbftx["fee"])
+    bumped_len = len(rbf_node.getrawtransaction(bumped_tx["txid"])) // 2
+    assert_greater_than(Decimal("0.00001000"), abs(feerate - bumped_tx["fee"] * 1000 / bumped_len))
```
jnewbery: I see that you've tried to clarify what this assert is doing, but it's still pretty obscure to me. Why should the difference between the desired feerate and the actual feerate be at least 1000 satoshis/byte? Can you add a comment explaining?

I also think that this inequality is more clearly expressed as actual_feerate >= desired_feerate + 1000 (ie the thing we're asserting is on the left side of the inequality).

qa/rpc-tests/bumpfee.py
```diff
@@ -165,66 +165,56 @@ def test_bumpfee_with_descendant_fails(rbf_node, rbf_node_address, dest_address)
...
-    rbf_node.settxfee(Decimal("0.00001000"))
-    rbfid = create_fund_sign_send(rbf_node, {dest_address: 0.00090000})
+    rbfid = spend_one_input(rbf_node, dest_address)
     rbftx = rbf_node.gettransaction(rbfid)
-    rbf_node.settxfee(Decimal("0.00002500"))
+    feerate = Decimal("0.00025000")
+    rbf_node.settxfee(feerate)
     bumped_tx = rbf_node.bumpfee(rbfid)
-    assert bumped_tx["fee"] > 2 * abs(rbftx["fee"])
+    bumped_len = len(rbf_node.getrawtransaction(bumped_tx["txid"])) // 2
```
jnewbery: I think this is more clearly expressed as:

`bumped_len = rbf_node.getrawtransaction(bumped_tx["txid"], True)["size"]` (ie not relying on size being the length of the serialized string // 2)

jnewbery (commented):
Looks great. Definitely an improvement.

I've added a few nits inline. My only other comment is that it'd be great if you could update the module docstring (after rebasing onto master) to say:

- what the individual subtests are doing
- the subtests consume discrete inputs, and there should be no shared state between subtests. They should be able to be run in any order (and if anyone adds additional subtests, they should follow that convention).

Tested ACK https://github.com/bitcoin/bitcoin/pull/9701/commits/145b702e5a2e2f6cf8766e070e25ee9a24ee931d

MarcoFalke:
utACK 145b702. Needs rebase.

qa/rpc-tests/bumpfee.py
MarcoFalke: There is a `count_bytes` helper function in util, that also checks that the bytes are hex encoded.

laanwj:
utACK, needs rebase.

qa/rpc-tests/bumpfee.py
ryanofsky: Thanks, switched to count_bytes in 779ea9c214c1cf508efea59b8ebaaec424fdac44.

qa/rpc-tests/bumpfee.py
ryanofsky: It's checking that difference between requested and actual feerate is less than 1000, not greater than 1000. Replaced 1000 with `epsilon` variable to maybe make comparison clearer in 779ea9c214c1cf508efea59b8ebaaec424fdac44.

qa/rpc-tests/bumpfee.py
ryanofsky: Renamed input->tx_input in 8ecec5f75a8b527194275835cbc0177a6ae99b3a.

ryanofsky (commented):
Rebased 145b702e5a2e2f6cf8766e070e25ee9a24ee931d -> 68a49603eea561d7c38c4d7fc7d7550e8622e1dd ([pr/bumpfee-fragile.0](https://github.com/ryanofsky/bitcoin/commits/pr/bumpfee-fragile.0) -> [pr/bumpfee-fragile.1](https://github.com/ryanofsky/bitcoin/commits/pr/bumpfee-fragile.1))
Squashed 8ecec5f75a8b527194275835cbc0177a6ae99b3a -> 45b060ef9d00f17e639c0e5d76500e5587bcc6d0 ([pr/bumpfee-fragile.2](https://github.com/ryanofsky/bitcoin/commits/pr/bumpfee-fragile.2) -> [pr/bumpfee-fragile.3](https://github.com/ryanofsky/bitcoin/commits/pr/bumpfee-fragile.3))

jnewbery:
tested ACK. I'd still prefer to see:

- a comment in the docstring about the subtests consuming discrete inputs, and that there should be no shared state between subtests.
- a comment explaining why epsilon is 1000. What is this testing?
- the call to count_bytes removed and using the RPC to get size directly. The test here is for the feerate, not that getrawtranasaction can correctly serialize into a hex string.

ryanofsky:
- Dropped count_bytes and clarified the settxfee assertion in 5876a4da9bf93c32efc42ebaf3c11f639608f8ec.
- Expanded docstring in 74d74d9f68ed3a5dd9fd268fdb0ef5985ae670a0.
- Squashed 74d74d9f68ed3a5dd9fd268fdb0ef5985ae670a0 -> a641eba17ae97ff80d7d1c272b7f86f94dfc70fd ([pr/bumpfee-fragile.4](https://github.com/ryanofsky/bitcoin/commits/pr/bumpfee-fragile.4) -> [pr/bumpfee-fragile.5](https://github.com/ryanofsky/bitcoin/commits/pr/bumpfee-fragile.5))

jnewbery:
retested ACK https://github.com/bitcoin/bitcoin/pull/9701/commits/a641eba17ae97ff80d7d1c272b7f86f94dfc70fd. Changes look good. Thanks!

ryanofsky:
Rebased a641eba17ae97ff80d7d1c272b7f86f94dfc70fd -> f85ac54e2429ad70c8bd4b2dbdf9d4107fd81f3f ([pr/bumpfee-fragile.5](https://github.com/ryanofsky/bitcoin/commits/pr/bumpfee-fragile.5) -> [pr/bumpfee-fragile.6](https://github.com/ryanofsky/bitcoin/commits/pr/bumpfee-fragile.6)) because of renames in #9956.

PR #9704, opened 2017-02-07
Store downloaded blocks prior to shutdown.

### LINKED ISSUE #9703: downloaded blocks are not stored upon shutdown
Especially noticeable during initial block download, there can be quite a number of blocks downloaded which do not get processed (i.e. not saved) when shutdown is requested, yet activation of chain (which takes much longer) is permitted to continue.

rebroad:
Fixes #9703 

Intentionally does not process blocks from disconnecting peers (although this could be changed if it is desired behavior).


fanquake:
Closing due to no discussion. Also needs a rebase if re-opening.

Issue #3921, opened 2014-03-21
build: Need options to specify BDB location


laanwj:
Many people compile BerkeleyDb 4.8 themselves and understandably don't want to install it system-wide.

I'd suggest the following names:

```
--with-bdb-libdir=
--with-bdb-includedir=
```

One can always override LDFLAGS and CFLAGS to find them manually, don't forget that you need to manualy specify optimization flags in that case ie: 

```
./configure LDFLAGS="-L/path/to/libs" CPPFLAGS="-I/path/to/includes -O2"
```

But it'd be easier if there was a flag for this as it is very common now.

Alternative would be `BDB_CFLAGS` and `BDB_LIBS`, analogous to `SSL_CFLAGS` and `SSL_LIBS`.

laanwj:
I'm trying to build Bitcoin Core on FreeBSD and would be very much helped by this. Also in our build guides we routinely override `LDFLAGS`/`CXXFLAGS` for this purpose, which shouldn't be necessary if there was an option or a specific env variable for this:

- https://github.com/bitcoin/bitcoin/blob/master/doc/build-unix.md#berkeley-db
- https://github.com/bitcoin/bitcoin/blob/master/doc/build-openbsd.md#building-bitcoin-core

Anyone up for this? I can try to do it myself but I don't know that much about m4 and will probably come up with a hacky solution.

laanwj:
Fixed by #9705

PR #9705, opened 2017-02-07
build: Add options to override BDB cflags/libs

laanwj:
Add environment settings to specify the CFLAGS and LIBS to be used for BerkeleyDB directly. These will completely by-pass autodetection in the same way as other similar flags for pkgconfig.

```
BDB_CFLAGS  C compiler flags for BerkeleyDB, bypasses autodetection
BDB_LIBS    Linker flags for BerkeleyDB, bypasses autodetection
```

Implements #3921.

Also update *BSD build guides to use these.

### DIFF
--- a/build-aux/m4/bitcoin_find_bdb48.m4
+++ b/build-aux/m4/bitcoin_find_bdb48.m4
@@ -3,68 +3,76 @@ dnl Distributed under the MIT software license, see the accompanying
 dnl file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
 AC_DEFUN([BITCOIN_FIND_BDB48],[
-  AC_MSG_CHECKING([for Berkeley DB C++ headers])
-  BDB_CPPFLAGS=
-  BDB_LIBS=
-  bdbpath=X
-  bdb48path=X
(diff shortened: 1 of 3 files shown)


jonasschnelli (approved):
utACK 5cc2ebb619b6bc7125e6a5f83f860dfcaa759d93
A nice addition to this would be listing the found BDB version (and or maybe the BDB CLAGS) in the configure summary.

laanwj:
> A nice addition to this would be listing the found BDB version (and or maybe the BDB CLAGS) in the configure summary.

I thought about that. But well we don't really detect the BDB version, there's just a 4.8-check-or-not (which gets skipped if you manually provide these environment variables). May be worth doing at some point, but not as part of this pull.
(we'd also have to decide what flags for which dependencies to list - listing them all would be a huge list)

paveljanik:
utACK https://github.com/bitcoin/bitcoin/pull/9705/commits/5cc2ebb619b6bc7125e6a5f83f860dfcaa759d93

theuni:
utACK 5cc2ebb619b6bc7125e6a5f83f860dfcaa759d93

PR #9706, opened 2017-02-07
Remove unicode characters from RPC user and password in QA tests

jnewbery:
The characters are cute, but they're difficult to type into bitcoin-cli
if you need to halt the test and manually issue RPCs to the node under
test.


laanwj:
They're there for a purpose: to test that unicode values work in the user/password. This is important for supporting non-English languages.

Why would you ever need to type them manually, instead of just pointing bitcoin-cli at the right `.conf` file??

jnewbery:
You're right - using the bitcoin-cli -conf option makes sense. Thanks!

PR #9712, opened 2017-02-07
bench: Fix initialization order in registration

### LINKED ISSUE #8910: bench_bitcoin crashes on FreeBSD and OpenBSD
`bench_bitcoin` crashes at startup, even before getting to main. It seems to have to do with the registration system:

``` c++
// BENCHMARK(foo) expands to:  benchmark::BenchRunner bench_11foo("foo", foo);
#define BENCHMARK(n) \
    benchmark::BenchRunner BOOST_PP_CAT(bench_, BOOST_PP_CAT(__LINE__, n))(BOOST_PP_STRINGIZE(n), n);
```

I don't know how the BOOST_PP_CAT stuff is working internally. This is likely an upstream issue with boost, to do with initialization order.
(FreeBSD 10.3, Boost 1.55)

laanwj:
The initialization order of global data structures in different implementation units is undefined. Making use of this is essentially gambling on what the linker does, the so-called [Static initialization order fiasco](https://isocpp.org/wiki/faq/ctors#static-init-order).

In this case it apparently worked on Linux but failed on OpenBSD and FreeBSD.

To create it on first use, make the registration structure local static to a function.

Fixes #8910.
(tested on Linux, OpenBSD, FreeBSD)

### DIFF
--- a/src/bench/bench.cpp
+++ b/src/bench/bench.cpp
@@ -9,7 +9,10 @@
 #include <iomanip>
 #include <sys/time.h>
 
-std::map<std::string, benchmark::BenchFunction> benchmark::BenchRunner::benchmarks;
+benchmark::BenchRunner::BenchmarkMap &benchmark::BenchRunner::benchmarks() {
+    static std::map<std::string, benchmark::BenchFunction> benchmarks_map;
+    return benchmarks_map;
+}
 
 static double gettimedouble(void) {
     struct timeval tv;
@@ -19,7 +22,7 @@ static double gettimedouble(void) {
 
 benchmark::BenchRunner::BenchRunner(std::string name, benchmark::BenchFunction func)
 {
-    benchmarks.insert(std::make_pair(name, func));
+    benchmarks().insert(std::make_pair(name, func));
 }
 
 void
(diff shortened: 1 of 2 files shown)


MarcoFalke:
utACK 29c53289a9e6bb34a098ea87d923968e3ac6d75d

TheBlueMatt:
utACK 29c53289a9e6bb34a098ea87d923968e3ac6d75d

paveljanik:
... and now you are toast ;-)

ACK https://github.com/bitcoin/bitcoin/pull/9712/commits/29c53289a9e6bb34a098ea87d923968e3ac6d75d

src/bench/bench.cpp
```diff
@@ -29,12 +32,9 @@ benchmark::BenchRunner::RunAll(double elapsedTimeForOne)
     std::cout << "#Benchmark" << "," << "count" << "," << "min" << "," << "max" << "," << "average" << ","
               << "min_cycles" << "," << "max_cycles" << "," << "average_cycles" << "\n";
 
-    for (std::map<std::string,benchmark::BenchFunction>::iterator it = benchmarks.begin();
-         it != benchmarks.end(); ++it) {
-
-        State state(it->first, elapsedTimeForOne);
-        benchmark::BenchFunction& func = it->second;
-        func(state);
```
jtimon: why is func(state) not called anymore?

jtimon:
utACK 29c5328

PR #9716, opened 2017-02-08
[Net] Clarity TIMEOUT_INTERVAL constant meaning.

pstratem:


src/net.h
```diff
@@ -47,7 +47,7 @@ namespace boost {
 /** Time between pings automatically sent out for latency probing and keepalive (in seconds). */
 static const int PING_INTERVAL = 2 * 60;
 /** Time after which to disconnect, after waiting for a ping response (or inactivity). */
-static const int TIMEOUT_INTERVAL = 20 * 60;
+static const int PING_TIMEOUT_INTERVAL = PING_INTERVAL * 5;
```
TheBlueMatt: Why change the number?

MarcoFalke:
Looks like time is denoted in `int64_t` elsewhere. When changing the
name, you might as well make the type `static constexpr int64_t`, so
we don't need to rely on implicit promotion whenever the constant is
used.

sipa:
I'm not convinced the constant rename helps. From the code it looks like a general timeout for received messages, not just a ping timeout?

pstratem:
@sipa it's a general timeout for having sent anything to the other peer, however the only thing guaranteeing we send peers a message at regular intervals is ping, it's critical that TIMEOUT_INTERVAL be larger than PING_INTERVAL because of this, but it's not obvious where the constants are defined from their names

laanwj:
There seems no agreement on doing this, and the issue has been dormant for months, so closing.

PR #9720, opened 2017-02-08
net: fix banning and disallow sending messages before receiving verack

theuni:
This is the last of my net issues for 0.14. As discussed with @TheBlueMatt and @gmaxwell.

Fixes for a few problems discovered while running a network stress/fuzzer:
- Remote nodes weren't always banned when they hadn't yet sent a verack. Regression from 7a8c2519015650acd51eaf42719f04e53f839bbe.
- Require a verack before sending any non-handshake messages. This is much more straightforward behavior, and allows for tests to be easily written
- Now that there's a sane model for testing, add checks for leaky messages sent out before the handshake is complete, as well as for banning in those cases.

### DIFF
--- a/qa/rpc-tests/p2p-leaktests.py
+++ b/qa/rpc-tests/p2p-leaktests.py
@@ -1,0 +1,145 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from test_framework.mininode import *
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import *
+
+'''
+Test for message sending before handshake completion
+
+A node should never send anything other than VERSION/VERACK/REJECT until it's
+received a VERACK.
+
+This test connects to a node and sends it a few messages, trying to intice it
+into sending us something it shouldn't.
+'''
+
+banscore = 10
+
+class CLazyNode(NodeConnCB):
+    def __init__(self):
+        self.connection = None
+        self.unexpected_msg = False
+        self.connected = False
+        super().__init__()
+
+    def add_connection(self, conn):
+        self.connection = conn
+
+    def send_message(self, message):
+        self.connection.send_message(message)
+
+    def bad_message(self, message):
+        self.unexpected_msg = True
+        print("should not have received message: %s" % message.command)
+
+    def on_open(self, conn):
+        self.connected = True
+
+    def on_version(self, conn, message): self.bad_message(message)
+    def on_verack(self, conn, message): self.bad_message(message)
+    def on_reject(self, conn, message): self.bad_message(message)
+    def on_inv(self, conn, message): self.bad_message(message)
+    def on_addr(self, conn, message): self.bad_message(message)
+    def on_alert(self, conn, message): self.bad_message(message)
+    def on_getdata(self, conn, message): self.bad_message(message)
+    def on_getblocks(self, conn, message): self.bad_message(message)
+    def on_tx(self, conn, message): self.bad_message(message)
+    def on_block(self, conn, message): self.bad_message(message)
+    def on_getaddr(self, conn, message): self.bad_message(message)
+    def on_headers(self, conn, message): self.bad_message(message)
+    def on_getheaders(self, conn, message): self.bad_message(message)
+    def on_ping(self, conn, message): self.bad_message(message)
+    def on_mempool(self, conn): self.bad_message(message)
+    def on_pong(self, conn, message): self.bad_message(message)
+    def on_feefilter(self, conn, message): self.bad_message(message)
+    def on_sendheaders(self, conn, message): self.bad_message(message)
+    def on_sendcmpct(self, conn, message): self.bad_message(message)
+    def on_cmpctblock(self, conn, message): self.bad_message(message)
+    def on_getblocktxn(self, conn, message): self.bad_message(message)
+    def on_blocktxn(self, conn, message): self.bad_message(message)
+
+# Node that never sends a version. We'll use this to send a bunch of messages
+# anyway, and eventually get disconnected.
+class CNodeNoVersionBan(CLazyNode):
+    def __init__(self):
+        super().__init__()
+
+    # send a bunch of veracks without sending a message. This should get us disconnected.
+    # NOTE: implementation-specific check here. Remove if bitcoind ban behavior changes
+    def on_open(self, conn):
+        super().on_open(conn)
+        for i in range(banscore):
+            self.send_message(msg_verack())
+
+    def on_reject(self, conn, message): pass
+
+# Node that never sends a version. This one just sits idle and hopes to receive
+# any message (it shouldn't!)
+class CNodeNoVersionIdle(CLazyNode):
+    def __init__(self):
+        super().__init__()
+
+# Node that sends a version but not a verack.
+class CNodeNoVerackIdle(CLazyNode):
+    def __init__(self):
+        self.version_received = False
+        super().__init__()
+
+    def on_reject(self, conn, message): pass
+    def on_verack(self, conn, message): pass
+    # When version is received, don't reply with a verack. Instead, see if the
+    # node will give us a message that it shouldn't. This is not an exhaustive
+    # list!
+    def on_version(self, conn, message):
+        self.version_received = True
+        conn.send_message(msg_ping())
+        conn.send_message(msg_getaddr())
+
+class P2PLeakTest(BitcoinTestFramework):
+    def __init__(self):
+        super().__init__()
+        self.num_nodes = 1
+    def setup_network(self):
+        extra_args = [['-debug', '-banscore='+str(banscore)]
+                      for i in range(self.num_nodes)]
+        self.nodes = start_nodes(self.num_nodes, self.options.tmpdir, extra_args)
+
+    def run_test(self):
+        no_version_bannode = CNodeNoVersionBan()
+        no_version_idlenode = CNodeNoVersionIdle()
+        no_verack_idlenode = CNodeNoVerackIdle()
+
+        connections = []
+        connections.append(NodeConn('127.0.0.1', p2p_port(0), self.nodes[0], no_version_bannode, send_version=False))
+        connections.append(NodeConn('127.0.0.1', p2p_port(0), self.nodes[0], no_version_idlenode, send_version=False))
+        connections.append(NodeConn('127.0.0.1', p2p_port(0), self.nodes[0], no_verack_idlenode))
+        no_version_bannode.add_connection(connections[0])
+        no_version_idlenode.add_connection(connections[1])
+        no_verack_idlenode.add_connection(connections[2])
+
+        NetworkThread().start()  # Start up network handling in another thread
+
+        assert(wait_until(lambda: no_version_bannode.connected and no_version_idlenode.connected and no_verack_idlenode.version_received, timeout=10))
+
+        # Mine a block and make sure that it's not sent to the connected nodes
+        self.nodes[0].generate(1)
+
+        #Give the node enough time to possibly leak out a message
+        time.sleep(5)
+
+        #This node should have been banned
+        assert(no_version_bannode.connection.state == "closed")
+
+        [conn.disconnect_node() for conn in connections]
+
+        # Make sure no unexpected messages came in
+        assert(no_version_bannode.unexpected_msg == False)
+        assert(no_version_idlenode.unexpected_msg == False)
+        assert(no_verack_idlenode.unexpected_msg == False)
+
+if __name__ == '__main__':
+    P2PLeakTest().main()
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -2596,6 +2596,36 @@ bool static ProcessMessage(CNode* pfrom, const std::string& strCommand, CDataStr
     return true;
 }
 
+static bool SendRejectsAndCheckIfBanned(CNode* pnode, CConnman& connman)
+{
+    AssertLockHeld(cs_main);
+    CNodeState &state = *State(pnode->GetId());
+
+    BOOST_FOREACH(const CBlockReject& reject, state.rejects) {
+        connman.PushMessage(pnode, CNetMsgMaker(INIT_PROTO_VERSION).Make(NetMsgType::REJECT, (std::string)NetMsgType::BLOCK, reject.chRejectCode, reject.strRejectReason, reject.hashBlock));
+    }
+    state.rejects.clear();
+
+    if (state.fShouldBan) {
+        state.fShouldBan = false;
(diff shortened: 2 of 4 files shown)


theuni:
I see that @TheBlueMatt and I managed to make almost the exact same change:
2a278cdaf6ea46dc85a61a37a12a8acd7acd5670 vs a5032b5b0c55c90a8e4df658d85d99824cf4699d. I'm happy to rebase and drop mine if his goes in first.

rebroad:
What is the risk by not merging this?

TheBlueMatt:
Concept ACK, will give this a more full review after breakfast.

instagibbs (approved):
utACK 50ce6755e024bc849ef8681f9aaa81ff3f3bca57

I'd also like the commit referred to in 0abd2ccfaa24fdd1e5d24271e2360c4d9d64b662 's commit message to include its commit hash for easier understanding.

theuni:
I got a few questions about these changes, so I've updated the commit messages to provide (I hope) better back-story/context.

src/net_processing.cpp
```diff
@@ -2594,6 +2594,35 @@ bool static ProcessMessage(CNode* pfrom, const std::string& strCommand, CDataStr
     return true;
 }
 
+static bool SendRejectsAndCheckBan(CNode* pnode, CConnman& connman)
```
theuni: Heh, i waffled back and forth on this. Sure.

src/net_processing.cpp
```diff
@@ -2594,6 +2594,35 @@ bool static ProcessMessage(CNode* pfrom, const std::string& strCommand, CDataStr
...
+{
+    AssertLockHeld(cs_main);
+    CNodeState &state = *State(pnode->GetId());
+
+    BOOST_FOREACH(const CBlockReject& reject, state.rejects)
```
theuni: Will do.

src/net_processing.cpp
TheBlueMatt: The usual fix for waffling is comment and clarify name :p

src/net_processing.cpp
kallewoof: Nit: avoid `BOOST_FOREACH` if possible (though this was moved code, so perhaps not applicable, but..).
```C++
for (const CBlockReject& reject : state.rejects) [...]
```

kallewoof (commented):
utACK, did not review tests

src/net_processing.cpp
laanwj: No, for moved code this is not applicable. To avoid confusion, let's keep code style changes, moves and bugfixes separate where possible.

src/net_processing.cpp
```diff
@@ -2706,8 +2735,11 @@ bool ProcessMessages(CNode* pfrom, CConnman& connman, const std::atomic<bool>& i
             PrintExceptionContinue(NULL, "ProcessMessages()");
         }
 
-        if (!fRet)
+        if (!fRet) {
             LogPrintf("%s(%s, %u bytes) FAILED peer=%d\n", __func__, SanitizeString(strCommand), nMessageSize, pfrom->id);
+            LOCK(cs_main);
+            SendRejectsAndCheckBan(pfrom, connman);
```
TheBlueMatt: There are a few cases it looks like we return true after setting misbehaving, so this probably needs to be checked either way (or those cases need updating - but that might result in double-logging).

theuni: I was trying to avoid the cs_main lock for each message received, but you're right that it means that we rely on a return value that's not very well-defined.

I'll just make it unconditional. Until we have parallel processing, there's not much harm in that.

TheBlueMatt: Sounds good.

TheBlueMatt:
utACK df1a32392933a4f716c53d62703a56e8d8bda9da, did not review tests.

gmaxwell (approved):
ACK (only tested lightly)

qa/rpc-tests/p2p-leaktests.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from test_framework.mininode import *
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import *
+import logging
```
jnewbery: You can remove this. logging isn't being used.

qa/rpc-tests/p2p-leaktests.py
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
```
jnewbery: Please add a docstring describing what this test case is doing.

qa/rpc-tests/p2p-leaktests.py
```diff
@@ -0,0 +1,135 @@
...
+        self.done = False
+        self.unrequested_msg = False
+
+    def add_connection(self, conn):
+        self.connection = conn
+
+    def check(self):
+        def done():
+            return self.done
+        return wait_until(done, timeout=10) and not self.unrequested_msg
```
jnewbery: I'm not entirely sure what this is achieving. I think you can remove `check()` and `self.done` entirely.

You can check directly that no unexpected messages have been received at the end of the test (see my comment below).

qa/rpc-tests/p2p-leaktests.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def send_message(self, message):
+        self.connection.send_message(message)
+
+    def bad_message(self, message):
+        self.unrequested_msg = True
```
jnewbery: Is `unexpected_message` a better name than `unrequested_message`?

qa/rpc-tests/p2p-leaktests.py
```diff
@@ -0,0 +1,135 @@
...
+        raise Exception("should not have received message: %s" % message.command)
```
jnewbery: I don't think you should be raising an exception in the network thread. Just set `unexpected_message` to True (and optionally print some debug info)

qa/rpc-tests/p2p-leaktests.py
```diff
@@ -0,0 +1,135 @@
...
+    def on_getheaders(self, conn, message): self.bad_message(message)
+    def on_ping(self, conn, message): self.bad_message(message)
+    def on_mempool(self, conn): self.bad_message(message)
+    def on_pong(self, conn, message): self.bad_message(message)
+    def on_feefilter(self, conn, message): self.bad_message(message)
+    def on_sendheaders(self, conn, message): self.bad_message(message)
+    def on_sendcmpct(self, conn, message): self.bad_message(message)
+    def on_cmpctblock(self, conn, message): self.bad_message(message)
+    def on_getblocktxn(self, conn, message): self.bad_message(message)
+    def on_blocktxn(self, conn, message): self.bad_message(message)
```
jnewbery: Sigh. This is verbose and will be incomplete every time new p2p message types are added.

I was tempted to say just override the `deliver()` message from `NodeConnCB` but that just feels ugly (`deliver()` should really be a private method and the testcases just override use the on_ callbacks).

I can't think of anything better here.

qa/rpc-tests/p2p-leaktests.py
```diff
@@ -0,0 +1,135 @@
...
+
+# Node that sends a version but not a verack.
+class CNodeNoVerackIdle(CLazyNode):
+    def __init__(self):
+        CLazyNode.__init__(self)
+
+    def on_reject(self, conn, message): pass
+    def on_verack(self, conn, message): pass
+
+    # When version is received, don't replay with a verack. Instead, see if the
```
jnewbery: s/replay/reply

qa/rpc-tests/p2p-leaktests.py
```diff
@@ -0,0 +1,135 @@
...
+    def on_version(self, conn, message):
+        self.done = True
+        conn.send_message(msg_ping())
+        conn.send_message(msg_getaddr())
+
+class P2PLeakTest(BitcoinTestFramework):
+    def add_options(self, parser):
+        parser.add_option("--testbinary", dest="testbinary",
+                          default=os.getenv("BITCOIND", "bitcoind"),
+                          help="Binary to test max block requests behavior")
```
jnewbery: I don't think this parser.add_option line is required.

qa/rpc-tests/p2p-leaktests.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def __init__(self):
+        super().__init__()
+        self.num_nodes = 1
+        self.banscore = 10
```
jnewbery: Since banscore isn't changing I'd prefer to make it a global constant.

qa/rpc-tests/p2p-leaktests.py
```diff
@@ -0,0 +1,135 @@
...
+    def setup_network(self):
+        extra_args = [['-debug', '-banscore='+str(self.banscore)]
+                      for i in range(self.num_nodes)]
+        self.nodes = start_nodes(self.num_nodes, self.options.tmpdir, extra_args)
```
jnewbery: There's only one node here. You can replace the above three lines with:

```python
self.nodes = [start_node(0, self.options.tmpdir, ['-debug', '-banscore='+str(self.banscore)])]
```

qa/rpc-tests/p2p-leaktests.py
```diff
@@ -0,0 +1,135 @@
...
+        connections = []
+        connections.append(NodeConn('127.0.0.1', p2p_port(0), self.nodes[0], no_version_bannode, send_version=False))
+        connections.append(NodeConn('127.0.0.1', p2p_port(0), self.nodes[0], no_version_idlenode, send_version=False))
+        connections.append(NodeConn('127.0.0.1', p2p_port(0), self.nodes[0], no_verack_idlenode))
+        no_version_bannode.add_connection(connections[0])
+        no_version_idlenode.add_connection(connections[1])
+        no_verack_idlenode.add_connection(connections[2])
+
+        NetworkThread().start()  # Start up network handling in another thread
+        self.nodes[0].generate(1)
```
jnewbery: not needed?

qa/rpc-tests/p2p-leaktests.py
```diff
@@ -0,0 +1,135 @@
...
+        # NOTE: implementation-specific check here. Remove if bitcoind ban behavior changes
+        for i in range(self.banscore):
+            no_version_bannode.send_message(msg_verack())
+
+        #Give the node enough time to possibly leak out a message
+        time.sleep(5)
+
+        assert(no_version_bannode.check())
+        assert(no_version_idlenode.check())
+        assert(no_verack_idlenode.check())
```
jnewbery: Might as well just check the variable directly:

```python
        assert(not no_version_bannode.unexpected_message)
        assert(not no_version_idlenode.unexpected_message)
        assert(not no_verack_idlenode.unexpected_message)
```

qa/rpc-tests/p2p-leaktests.py
```diff
@@ -0,0 +1,135 @@
...
+
```
jnewbery: I recommend you disconnect all the connections to the node at the end of the test:

```python
        # Disconnect all peers
        [conn.disconnect_node() for conn in connections]
```

jnewbery (commented):
Test looks good @theuni ! I've added a bunch of style comments.

qa/rpc-tests/p2p-leaktests.py
theuni: ok

qa/rpc-tests/p2p-leaktests.py
theuni: ok

qa/rpc-tests/p2p-leaktests.py
theuni: seems about the same to me, but sure

qa/rpc-tests/p2p-leaktests.py
theuni: ok

qa/rpc-tests/p2p-leaktests.py
theuni: yes, i hate this too.

Too much to do here but for the future: how about a base class for messages. Then there's an overridable dispatcher that forwards to the individual callbacks. So here, I would just override the dispatcher and check the message type.

Basically just another thin layer under deliver()

qa/rpc-tests/p2p-leaktests.py
theuni: ok

qa/rpc-tests/p2p-leaktests.py
theuni: ok. Definitely the c/p from the wrong test isn't needed :)

qa/rpc-tests/p2p-leaktests.py
theuni: I think i'd rather keep this local so that we can use different scores for different tests.

qa/rpc-tests/p2p-leaktests.py
theuni: ok

qa/rpc-tests/p2p-leaktests.py
theuni: Very needed, but needs a comment. We generate a block, and wait a few secs to make sure that it's not relayed to the nodes who haven't versioned/veracked yet.

qa/rpc-tests/p2p-leaktests.py
theuni: ok

qa/rpc-tests/p2p-leaktests.py
theuni: ok

theuni:
@jnewbery Thanks for the great test review. I'm not sure I'll have time to get to this before Sunday, so let's not let it hold back merge if a few more ACKs come in. I'll for sure fix up the tests post-merge if that's the case.

qa/rpc-tests/p2p-leaktests.py
MarcoFalke: Please leave it as is for now. My goal was to generalize this logic and get rid of most `setup_network` methods in test as the only thing they commonly do is fire up the requested number of nodes and nothing else. So there would be no need to overwrite this function if you want a simple network set up.

MarcoFalke: Also, it would be easier to adapt if the number of nodes changes.

theuni:
Fixed up the tests and squashed. Only the tests changed, the bitcoin code is exactly the same as before squash. I've archived the old branch here: https://github.com/theuni/bitcoin/commits/fix-ban2 in case anyone wants to compare.

I replaced my mininode changes with @TheBlueMatt's commits from #9715, so that these won't conflict with eachother.

PR #8961, opened 2016-10-18
Headers announcement for nodes that can do headers.

rebroad:
Announce blocks using headers instead of block invs to nodes that can process headers.

PR #9721, opened 2017-02-08
Headers announcement for nodes that can do headers.

rebroad:
This was raised previously (#8961) but closed due to an apparent misunderstanding - it builds upon the current announce by headers logic in that it will also detect nodes that understand headers and announce via headers to these also (I am not aware of any advantage in announcing via block invs to nodes that actively request headers and therefore can process them).

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -179,6 +179,8 @@ struct CNodeState {
     bool fPreferredDownload;
     //! Whether this peer wants invs or headers (when possible) for block announcements.
     bool fPreferHeaders;
+    //! Whether this peer can process headers for block announcements.
+    bool fCanDoHeaders;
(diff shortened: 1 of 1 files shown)


sdaftuar:
@rebroad BIP 130 has already been implemented and allows peers to opt-in to receiving block announcements via headers, specifically to maintain backwards compatibility and not require all software on the network adopt the protocol change.  This patch only makes the behavior more confusing (as it needlessly breaks BIP 130's opt-in protocol) and less flexible (as peers which make use of getheaders are now forced to receive block announcements this way as well, which could break other participants on the network).  This is why I NACKed this when you proposed it in #8961.

As an aside, I do think we could propose eliminating block inv's altogether, which has been suggested before.  But that is a topic that should first be discussed more broadly, eg on the bitcoin-dev mailing list, as it would affect all software on the peer-to-peer network.

Issue #3266, opened 2013-11-16
Should the GUI have a "you're sending again to the same place you sent before"?


gmaxwell:
when you try to send to some place you've already sent, perhaps there should be a polite "You've already sent X btc to this destination in TXN y on date Z. Do you want to continue?"  which should help people avoid some of the incidents they've had with double paying.

MarcoFalke:
It would also help to educate about address reuse.

CryptAxe:
Do either of you have any thoughts as to how to implement this? Looping through the wallet's transactions and comparing the destination address to what the user wants to send to now seems possible, but maybe not the best approach?

luke-jr:
I set a label for every payment, so whenever I go to pay an address a second time, it already loads that label up... which I take as an implicit warning (and then I go complain to whoever gave me the same address twice). Making it an explicit warning sounds like a good idea, and can probably be done with some flag on the addressbook entry easily enough.

laanwj:
Indeed, searching the address book should be enough, which is much faster than searching the transactions.

CryptAxe:
Alright heres a simple solution I am testing out. This will just check if the address you have pasted into the send coins entry address field exists in the address book already, and display a warning message if it does.

This will also display the error message if you create a receiving address from the request dialog and then try to send a payment there. I'm not sure if that is a problem because users probably aren't doing that too often.

I also noticed that the label field is not cleared when a new address without a label is pasted, the UI displays the old label so I fixed that as well.

https://github.com/CryptAxe/bitcoin/commit/0698a278274c32b2b90901561d1f8f510cd4890a

luke-jr:
1. If you type a label before the address, you don't want it cleared.
2. Consider the scenario where a user gets a loan, and saves the repayment address in his address book for later. I would consider it a bug if he got a warning when he tried to send to it the first time. So some explicit flag is needed (it can be set on the address book entry went it is used for sending).
3. "Address reuse is not recommended!" is too soft and unclear. I suggest "You have already paid this address!", and making it a hard error with a manual override (dialog with an "Ignore and send anyway" option?)

CryptAxe:
Thank you for the feedback. In this new branch I made it so that an address will be marked as used when it is sent to via UI.  If the user tries to send to that address again, a warning message will popup and require that they accept the consequences to continue. If multiple recipients are entered, each will be checked and the user will need to verify that they want to reuse each one before the send tx confirmation dialog is shown.

[...]

PR #9722, opened 2017-02-08
GUI: Display warning when attempting address reuse

CryptAxe:
Added address reuse key 'au' to destdata

Added informative QMessageBox popup warning when user attempts address reuse

### DIFF
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -263,6 +263,27 @@ void SendCoinsDialog::on_sendButton_clicked()
     QStringList formatted;
     Q_FOREACH(const SendCoinsRecipient &rcp, currentTransaction.getRecipients())
     {
+        // If user is attempting address reuse, show an informative warning
+        if (model->isUsed(rcp.address.toStdString()))
+        {
+            QString reuseWarningText = tr(
+                        "You have sent to the Bitcoin address %1 already!\n\n"
+                        "Sending to the same address twice is technically not supported and "
+                        "will reduce your own security as well as the privacy of all Bitcoin users."
+                    ).arg(rcp.address) + "\n";
+
+            QMessageBox reuseMessage("Warning: address reuse!",
+                                     reuseWarningText,
+                                     QMessageBox::Warning,
+                                     QMessageBox::Yes | QMessageBox::Escape,
+                                     QMessageBox::Cancel | QMessageBox::Default,
+                                     QMessageBox::NoButton,
+                                     this);
+
+            if (reuseMessage.exec() != QMessageBox::Yes)
+                return;
+        }
+
         // generate bold amount string
         QString amount = "<b>" + BitcoinUnits::formatHtmlWithUnit(model->getOptionsModel()->getDisplayUnit(), rcp.amount);
         amount.append("</b>");
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -268,16 +268,15 @@ void SendCoinsDialog::on_sendButton_clicked()
         {
             QString reuseWarningText = tr(
                         "You have sent to the Bitcoin address %1 already!\n\n"
-                        "Sending to the same address twice is technically not supported and "
-                        "will reduce your own security as well as the privacy of all Bitcoin users."
+                        "Address reuse is highly discouraged and will reduce "
+                        "your own security as well as the privacy of all Bitcoin users."
+                        "\n\nAre you sure you want to do this?"
                     ).arg(rcp.address) + "\n";
 
-            QMessageBox reuseMessage("Warning: address reuse!",
+            QMessageBox reuseMessage(QMessageBox::Warning,
+                                     "Warning: address reuse!",
                                      reuseWarningText,
-                                     QMessageBox::Warning,
-                                     QMessageBox::Yes | QMessageBox::Escape,
-                                     QMessageBox::Cancel | QMessageBox::Default,
-                                     QMessageBox::NoButton,
+                                     QMessageBox::Cancel | QMessageBox::Yes,
                                      this);
 
             if (reuseMessage.exec() != QMessageBox::Yes)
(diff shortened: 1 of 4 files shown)


CryptAxe:
Related to https://github.com/bitcoin/bitcoin/issues/3266

src/qt/sendcoinsdialog.cpp
```diff
@@ -263,6 +263,27 @@ void SendCoinsDialog::on_sendButton_clicked()
     QStringList formatted;
     Q_FOREACH(const SendCoinsRecipient &rcp, currentTransaction.getRecipients())
     {
+        // If user is attempting address reuse, show an informative warning
+        if (model->isUsed(rcp.address.toStdString()))
+        {
+            QString reuseWarningText = tr(
+                        "You have sent to the Bitcoin address %1 already!\n\n"
+                        "Sending to the same address twice is technically not supported and "
```
jonasschnelli: I think the part `technically not supported` is not correct.

src/qt/sendcoinsdialog.cpp
```diff
@@ -263,6 +263,27 @@ void SendCoinsDialog::on_sendButton_clicked()
...
+                        "will reduce your own security as well as the privacy of all Bitcoin users."
+                    ).arg(rcp.address) + "\n";
+
+            QMessageBox reuseMessage("Warning: address reuse!",
```
luke-jr: I can't find this constructor in the Qt docs...

src/qt/sendcoinsdialog.cpp
```diff
@@ -263,6 +263,27 @@ void SendCoinsDialog::on_sendButton_clicked()
...
+                                     reuseWarningText,
+                                     QMessageBox::Warning,
+                                     QMessageBox::Yes | QMessageBox::Escape,
```
luke-jr: Yes | Escape doesn't look like a normal combination?

src/qt/walletmodel.cpp
```diff
@@ -355,6 +355,9 @@ WalletModel::SendCoinsReturn WalletModel::sendCoins(WalletModelTransaction &tran
 
                 std::map<CTxDestination, CAddressBookData>::iterator mi = wallet->mapAddressBook.find(dest);
 
+                std::string key = "au"; // "au" prefix = "address used" in destdata
```
luke-jr: `"used"` seems more straightforward, but more importantly, this should be moved to a `static const` in a `src/wallet/` header.

src/qt/walletmodel.cpp
```diff
@@ -591,6 +594,24 @@ bool WalletModel::isSpent(const COutPoint& outpoint) const
     return wallet->IsSpent(outpoint.hash, outpoint.n);
 }
 
+bool WalletModel::isUsed(const std::string& address) const
+{
+    LOCK2(cs_main, wallet->cs_wallet);
+    for (const std::pair<CTxDestination, CAddressBookData>& item : wallet->mapAddressBook)
```
jonasschnelli: Instead of looping, would a std::map::find be possible (`std::map<CTxDestination, CAddressBookData>::iterator mi = wallet->mapAddressBook.find(dest);`?

jonasschnelli:
General concept ACK.
Am I correct that this patch does not warn for "send-agains" to addresses sent before this patch was active?

src/qt/sendcoinsdialog.cpp
TheBlueMatt: Indeed, it is supported fine, but it is highly discouraged.

luke-jr: It would not be a bug to stop scanning for transactions paying to an address once it has been used. It "works" only by accident.

MarcoFalke:
Agree that the "technically not supported" is confusing to users. We
should just mention "is (highly) discouraged".

CryptAxe:
Okay I changed the warning message because if there is confusion even here then it needs to be more clear.  I also updated the message box code so that it isn't using an obsolete constructor and removed the extra escape.

This warning will only be displayed for addresses which are used after the software has been updated. It would be possible though for someone to add an update later that would scan through the wallet transactions and mark the destinations of those as used. Also the wallet already had a function to read the destdata by using std::map find, but it wasn't being used anywhere as far as I can tell. I replaced the loop with that.

Thanks everyone for the feedback, hopefully my update https://github.com/bitcoin/bitcoin/pull/9722/commits/8f01b1f8f6e918b0a52397e12257cdc1f47dfc9d addressed all of the comments here

CryptAxe:
I am going to go through some tests now, but everything is ready to be reviewed again.

CryptAxe:
Alright I'm happy with the code. Anyone have a problem with the updated warning message?

jonasschnelli:
Code looks good to me. I dislike the `"1"`, but I guess this is just how we have to do it with the flexible per wtx map.

Though I'm unsure about the update path of this feature. If we only mark new destinations, this may result in a false awareness of address reuse because some addresses you may reuse (the old ones before the upgrade) won't show the warning and users may think it's okay to reuse them.

A solution would be, to either A) loop through the transactions that where made before the feature was active or to B) migrate at first start with the new feature (loop through all wtxs, set the `used` dest data.

I'd like to hear more opinions on that.

luke-jr:
I'd prefer B, but it would only allow one reuse before it'd get flagged, right?

CryptAxe:
Yeah the way it works right now, and with B, the code will only allow
tagging on first use. I'm fine switching to B or whatever makes it more
clear but 1 = true made sense to me.

jonasschnelli:
Any intentions to work this towards the solution "B" (migrate at first start with the new feature (loop through all wtxs, set the used dest data.)?

CryptAxe:
Yes, I was just waiting for some feedback on that before I went ahead and
added the code. Since I think it's a good idea and nobody seems to take
issue I'll finish this up next time I have a break.

jonasschnelli:
Closing due to inactivity (feel free to reopen once someone takle option B or there is consensus about doing solution A).

PR #9724, opened 2017-02-08
Qt/Intro: Add explanation of IBD process

luke-jr:
Follow-up for after 0.14 with the final fixes from #9718 along with some additional explanation text.

### DIFF
--- a/src/qt/forms/intro.ui
+++ b/src/qt/forms/intro.ui
@@ -7,7 +7,7 @@
     <x>0</x>
     <y>0</y>
     <width>674</width>
-    <height>363</height>
+    <height>415</height>
    </rect>
   </property>
   <property name="windowTitle">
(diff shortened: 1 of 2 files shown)


src/qt/forms/intro.ui
```diff
@@ -204,6 +201,36 @@
...
+     </property>
+     <property name="wordWrap">
+      <bool>true</bool>
+     </property>
+    </widget>
+   </item>
+   <item>
+    <widget class="QLabel" name="lblExplanation3">
+     <property name="text">
+      <string>If you have chosen to limit block chain storage, the historical data must still be downloaded and processed, but will be deleted afterward to keep your disk usage low.</string>
```
jonasschnelli: maybe add `(pruning)` after `If you have chosen to limit block chain storage`?

sipa:
Concept ACK post-0.14.

laanwj:
utACK f6d18f5

Issue #9430, opened 2016-12-26
'ERROR: Error reading proxy response' when using TOR on bitcoind


2SI3NX:
### Describe the issue
The problem is, the error message: 'ERROR: Error reading proxy response' in the debug.log file.
I wanted to simply run bitcoin over TOR. So in the bitcoin.conf file i had the following lines:
bind=127.0.0.1
port=8333
upnp=0
proxy=127.0.0.1:9050

Then i started bitcoind in terminal, i monitored the debug.log file and i saw that error message described above every minute or so while i was getting a new 'update tip' block.

To test if TOR was really working, i stopped the TOR service and i immediately got countless errors:
connect() to 127.0.0.1:9050 failed after select(): Connection refused (111)
connect() to 127.0.0.1:9050 failed after select(): Connection refused (111)
connect() to 127.0.0.1:9050 failed after select(): Connection refused (111)

So i realized that TOR was working, yet the original unknown error of:
'ERROR: Error reading proxy response'  occurred.

Please note: my intention was not to use a hidden tor service.
My intention was to simply torify outgoing connections as described in the first step here:
https://github.com/bitcoin/bitcoin/blob/master/doc/tor.md#1-run-bitcoin-behind-a-tor-proxy
and here:
https://bitcointalk.org/index.php?topic=623868.0

### Expected behaviour
The expected behaviour should have been that bitcoin would run over tor without a proxy error while tor is actually working.

### What version of bitcoin-core are you using?
sudo apt-add-repository ppa:bitcoin/bitcoin   (https://launchpad.net/~bitcoin/+archive/ubuntu/bitcoin)
sudo apt-get install bitcoind

### Machine specs:
- OS: Ubuntu 16.04
- Disk Type (HD/SDD): SSD

If there should be some other type of config that i am missing in bitcoin.conf please let me know.

unsystemizer:
You need to fix your proxy. Unless you can show other apps work with that proxy and Bitcoin does not, it's not a Bitcoin issue. Similar eample:
https://github.com/bitcoin-dot-org/bitcoin.org/issues/1334
Perhaps you should fix Tor first and make sure that it works as SOCKS5 proxy. At the moment you only demonstrate that the proxy is refusing connections, and not that Bitcoin is not functioning properly

Beast141:
Well the thing is, i do use the TOR daemon for other apps and it works perfectly in Ubuntu.  Thats why I am reporting this issue. Can you please try and reproduce the issue?

[...]

PR #9726, opened 2017-02-09
netbase: Do not print an error on connection timeouts through proxy

laanwj:
If a timeout happens while reading the proxy response, this effectively means we timed out while connecting to the remote node. This is very common for Tor, so do not print an error message to the log in that case.

Avoids a flood of 
```
2017-02-09 07:10:02 ERROR: Error reading proxy response
```
while running through Tor, causing users to blame the proxy.

Ref: #9430.


gmaxwell (approved):
utACK

PR #8920, opened 2016-10-14
Set minimum required Boost to 1.47.0

fanquake:
Boost is on the way out, but in the mean time, having the minimum set to 1.20.0 is misleading, and can lead to issues like #8868.

1.47.0 is the first version to include [Chrono](http://www.boost.org/users/history/version_1_47_0.html). See #8875 for some history.

PR #9727, opened 2017-02-09
Remove fallbacks for boost_filesystem < v3

laanwj:
Minimum boost version was bumped to 1.47.0 in #8920, which means the configure step won't even pass with older boost.

This version has boost filesystem v3, which means the (crappy) fallbacks for older versions can go.

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -37,9 +37,7 @@
 
 #include <boost/filesystem.hpp>
 #include <boost/filesystem/fstream.hpp>
-#if BOOST_FILESYSTEM_VERSION >= 3
 #include <boost/filesystem/detail/utf8_codecvt_facet.hpp>
-#endif
 #include <boost/scoped_array.hpp>
 
 #include <QAbstractItemView>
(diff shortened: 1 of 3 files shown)


fanquake:
Concept ACK, I'd be planning on starting this work. As you've probably noticed I have some lists of various workaround/possible changes in #8875

laanwj:
Yup, there's likely some more fallbacks that could be removed.

However I think the filesystem filebacks are especially bad. In one case (Qt) the behavior of the fallback is just plain wrong in the case of unicode, the other case looks like fragile, hardly-tested code but is luckily in the tests only.

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/9727/commits/763ef23ef4cef3456dda9c3a147a7709f308134a

laanwj:
Updated - there as another one in wallet.cpp, which I skipped over because it used `BOOST_VERSION` instead of `BOOST_FILESYSTEM_VERSION` as distinguisher.

fanquake:
utACK 056aba2

This removes all our uses of BOOST_FILESYSTEM_VERSION.
We're also down to 3 uses of BOOST_VERSION in /src.

sipa:
Posthumous utACK.

PR #9298, opened 2016-12-07
[Wallet] use CHDPubKey, don't store child priv keys in db, derive on the fly

jonasschnelli:
Adds a new database record (`"hdpubkey"`) reflected by class `CHDPubKey`.
* Results in no longer storing derived child private keys in the database
* Only the extended child public key will be stored
* If the private key gets requested, it will be derived on the fly

Dump functions are unchanged, will result in deriving the keys on the fly.
Not backward compatible. Ideally combined with HD chain split #9294.

PR #8723, opened 2016-09-14
[Wallet] Add support for flexible BIP32/HD keypath-scheme

jonasschnelli:
This adds the startup argument `-hdkeypath` which allows to set the BIP32 keypath scheme during the creation of a wallet.

This PR would allow to use keypath-scheme after BIP44, etc. to be compatible with other wallets.

This PR does **not** change the keypool mechanism. Even if the keypath would allow public-key-derivation, we still derive all keys with private-key-derivation and fill up the keypool.
Though, a PR that would enable public-key-derivation would be "in a reviewable size" once this gets merged.

PR #9728, opened 2017-02-09
Can create External HD wallet with -externalhd

NicolasDorier:
The user creates a new wallet by running `./bitcoind -externalhd=[ExtPubKey base58]`.

This make it possible to use methods like `getnewaddress`, `fundrawtransaction` and all normal wallet operations on a HD pubkey.

Software built on top of core which need to delegate signing operations to hardware wallet will have almost the same code as if signing was done by Core.

With the introduction of a standard for dealing with hardware wallet signing in the future, I expect that `signrawtransaction` will just delegate the signing to the hardware wallet.

In this way, there will be no code difference between software using third party solution for signing, and those just using core for signing.

I will use it in my own projects. My HW is giving me the ExtPubKey, and I want to use bitcoin core just for coin selection and tracking. I also did not wanted to break bunch of old code. Ping @jonasschnelli 

EDIT: @saleemrashid built HW support for Bitcoin Core on https://github.com/saleemrashid/bitcoin/tree/hardware-wallet based on this PR


src/wallet/walletdb.h
```diff
@@ -59,6 +63,10 @@ class CHDChain
         READWRITE(this->nVersion);
         READWRITE(nExternalChainCounter);
         READWRITE(masterKeyID);
+        if (this->nVersion <= SUPPORT_WATCHONLY_VERSION) {
```
jonasschnelli: `>=`?

src/wallet/wallet.cpp
```diff
@@ -93,72 +94,116 @@ CPubKey CWallet::GenerateNewKey()
...
 
-    if (!AddKeyPubKey(secret, pubkey))
-        throw std::runtime_error(std::string(__func__) + ": AddKey failed");
+    if(hasSecret) {
+        if (!AddKeyPubKey(secret, pubkey))
+            throw std::runtime_error(std::string(__func__) + ": AddKey failed");
+    }
+    else {
+        if (!AddWatchOnly(GetScriptForDestination(pubkey.GetID())))
+            throw std::runtime_error(std::string(__func__) + ": AddKey failed");
```
jonasschnelli: `s/AddKey/AddWatchOnly/`

src/wallet/wallet.cpp
```diff
@@ -93,72 +94,116 @@ CPubKey CWallet::GenerateNewKey()
...
+            metadata.hdKeypath = "m/0/0/" + std::to_string(hdChain.nExternalChainCounter) + "'";
+            metadata.hdMasterKeyID = hdChain.masterKeyID;
+            // increment childkey index
+            hdChain.nExternalChainCounter++;
+        } while (HaveWatchOnly(GetScriptForDestination(childKey.pubkey.GetID())));
+        pubKey = childKey.pubkey;
+        // update the chain model in the database
+        if (!CWalletDB(strWalletFile).WriteHDChain(hdChain))
+            throw std::runtime_error(std::string(__func__) + ": Writing HD chain model failed");
+        return false;
```
jonasschnelli: I think returning false in case of `isWatchOnly` is confusing.
IMO better set the `hasSecret` boolean with another check of `hdChain.isWatchOnly`.

jonasschnelli:
Concept ACK (will review in detail and test soon).
I think this is a great feature!
Together with #9662 this would allow better HWW/Cold-Storage interaction.

laanwj:
> The user create a new wallet by removing the old wallet.dat and running ./bitcoind -hdwatchonly=[ExtPubKey base58].

Instead of removing the wallet, it would also be possible to specify an alternative one with `-wallet`, I guess?

src/wallet/wallet.cpp
NicolasDorier: I can replace returns bool by void, and make the client responsible to test the validity of the returned Key.

But I fear that a new user would assume that DeriveNewKey already returns a valid CKey.

jonasschnelli: Yes. I once did a PR where we separate the key/pubkey records: https://github.com/bitcoin/bitcoin/pull/9298
I guess this would be the cleaner solution... but maybe later.

luke-jr:
How does this work, considering that Core is exclusively hardened derivation right now? Or can it only watch non-Core wallets? O.o

jonasschnelli:
> How does this work, considering that Core is exclusively hardened derivation right now? Or can it only watch non-Core wallets? O.o

For the hd-watch-only default keypath, we should probably use Bip44.
This PR makes much more sense if we would have the flexible key path feature https://github.com/bitcoin/bitcoin/pull/8723

NicolasDorier:
@luke-jr  for watchonly hd I am using non hardened derivation. goal is to eventually combine with flexible path.

@laanwj yes. What I mean is that if you specify -hdwatchonly on an already existing wallet, you get an error at startup.

NicolasDorier:
Travis error not related to this PR.

jonasschnelli:
@NicolasDorier: I think in order to pass travis you need to add `-hdwatchonly` to the check doc script.

NicolasDorier:
I added a commit to track P2PK as well. It would be the same behavior as normal wallet. However I am not sure it is the right approach as P2PK are obsolete.

Also, what for P2WPKH ?

NicolasDorier:
I am replicating the behavior of normal wallets, both p2pk and p2pkh are tracked.

The test is failing because I am testing wrong parameter usage, and there is no way in the test framework to assert an exception. Ping @MarcoFalke , same problem as reported by jonas on https://github.com/bitcoin/bitcoin/pull/9662

qa/rpc-tests/test_framework/util.py
```diff
@@ -352,7 +355,20 @@ def start_node(i, dirname, extra_args=None, rpchost=None, timewait=None, binary=
...
+        if assert_msg is None:
+            assert_msg = "bitcoind should have exited with an error"
+        raise AssertionError(assert_msg)
+    except AssertionError as e:
+        raise
+    except Exception as e:
+        assert('bitcoind exited' in str(e)) #node must have shutdown
+    
+
+def start_nodes(num_nodes, dirname, extra_args=None, rpchost=None, timewait=None, binary=None, ignorestderr=False):
```
MarcoFalke: Is this additional optional arg required? I don't think this will ever be used.

NicolasDorier: it is used by `assert_start_raises_init_error`

NicolasDorier:
I am at loss understanding why there is a permission denied.... I thought I maybe did not had right to /dev/null, and tried to redirect stderr to stdout instead, but same problem. Any idea ?

qa/rpc-tests/test_framework/util.py
MarcoFalke: Oh, right. I missed the plural s.

MarcoFalke:
> Any idea

Try

```sh
chmod +x qa/rpc-tests/hdwatchonly.py

**Comment by NicolasDorier** [comment]
rebased, and cleaned up the commits for better review.

**Comment by saleemrashid** [comment]
As far as I understand, Bitcoin Core uses `m/0'/0'/k'` whereas BIP 44 uses `m/0'/0/k` and `m/0'/1/k`. It makes most sense for this PR to implement `m/0/k` (and, with #9294, `m/1/k`). This allows you to use an `xpub` derived from `m/0'` (or any other account) by your hardware wallet, for example.

However, this means that you are doing one less derivation than with `xprv` in Bitcoin Core. I don't think this is too much of an issue since `m/0'/0'/k'` is in no way compatible with `m/0'/0/k` anyway.

**Comment by NicolasDorier** [comment]
@saleemrashid very good point. Will make your change once  #9294 is merged.

File: src/wallet/walletdb.h [hunk 5]
```diff
@@ -47,8 +47,12 @@ class CHDChain
 public:
     uint32_t nExternalChainCounter;
     CKeyID masterKeyID; //!< master key hash160
+    CExtPubKey masterPubKey;
+    bool isWatchOnly = false;
 
-    static const int CURRENT_VERSION = 1;
+    static const int FIRST_VERSION = 1;
+    static const int SUPPORT_WATCHONLY_VERSION = 2;
+    static const int CURRENT_VERSION = 2;
```
jonasschnelli: nit: `CURRENT_VERSION = SUPPORT_WATCHONLY_VERSION`.

src/wallet/wallet.cpp
```diff
@@ -91,71 +92,103 @@ CPubKey CWallet::GenerateNewKey()
...
+        } while (HaveKey(childKey.key.GetPubKey().GetID()));
+        secret = childKey.key;
+        pubKey = childKey.key.GetPubKey();
+        // update the chain model in the database
+        if (!CWalletDB(strWalletFile).WriteHDChain(hdChain))
+            throw std::runtime_error(std::string(__func__) + ": Writing HD chain model failed");
+    }
+    else if (IsHDWatchOnly()) {
+        CExtPubKey& masterKey = hdChain.masterPubKey;             //hd master key
+        CExtPubKey accountKey;            //key at m/0
```
jonasschnelli: To make this compatible with BIP44, can we not just use `m` as the BIP44 account key. No hardened derivation would then be involved and users could use their BIP44 xpub's to generate watchonlys.
The `externalChainChildKey` would then be `m/0` (instead of `m/0/0`)

src/wallet/wallet.cpp
```diff
@@ -91,71 +92,103 @@ CPubKey CWallet::GenerateNewKey()
...
+        CExtPubKey externalChainChildKey; //key at m/0/0
+        CExtPubKey childKey;              //key at m/0/0/<n>
+
+        // derive m/0
+        masterKey.Derive(accountKey, BIP32_NONHARDENED_KEY_LIMIT);
```
jonasschnelli: IMO this constant (`BIP32_NONHARDENED_KEY_LIMIT`) is superfluous. Just use `0`.

src/wallet/wallet.cpp
```diff
@@ -91,71 +92,103 @@ CPubKey CWallet::GenerateNewKey()
...
+
+        // derive m/0/0
+        accountKey.Derive(externalChainChildKey, BIP32_NONHARDENED_KEY_LIMIT);
+
+        // derive child key at next index, skip keys already known to the wallet
+        do {
+            externalChainChildKey.Derive(childKey, hdChain.nExternalChainCounter | BIP32_NONHARDENED_KEY_LIMIT);
```
jonasschnelli: Also here, ... just use `hdChain.nExternalChainCounter` instead of `hdChain.nExternalChainCounter | BIP32_NONHARDENED_KEY_LIMIT` (no need to bitwise OR it with 0).

src/wallet/wallet.cpp
```diff
@@ -91,71 +92,103 @@ CPubKey CWallet::GenerateNewKey()
...
     }
 
     // Compressed public keys were introduced in version 0.6.0
     if (fCompressed)
         SetMinVersion(FEATURE_COMPRPUBKEY);
 
-    CPubKey pubkey = secret.GetPubKey();
-    assert(secret.VerifyPubKey(pubkey));
+    if (!IsHDWatchOnly())
+        assert(secret.VerifyPubKey(pubkey));
```
jonasschnelli: I think there is no way to verify the correctness of the derived pubkey?

src/wallet/walletdb.h
```diff
@@ -59,6 +63,11 @@ class CHDChain
         READWRITE(this->nVersion);
         READWRITE(nExternalChainCounter);
         READWRITE(masterKeyID);
+        if (this->nVersion >= SUPPORT_WATCHONLY_VERSION) {
+            READWRITE(isWatchOnly);
+            if(isWatchOnly)
+                READWRITE(masterPubKey);
```
jonasschnelli: Somehow this PR was crashing here because of 
```
Assertion failed: (pubkey.size() == 33), function Encode, file pubkey.cpp, line 255.
```

It looks like that my tested `tpub` was non compressed? How is that even possible?
LLDB told me `pubkey = (vch = unsigned char [65] @ 0x00007fd16736976c)`.

jonasschnelli: Wait... I was using the extended private key instead. Though, I guess it should not crash in this case.

src/wallet/wallet.cpp
```diff
@@ -3618,14 +3714,41 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
         walletInstance->SetMaxVersion(nMaxVersion);
     }
 
+
+    std::string hdWatchOnly = GetArg("-hdwatchonly", "");
+    CExtPubKey extPubKey;
+    if (!hdWatchOnly.empty()) {
+        CBitcoinExtPubKey bitcoinExtPubKey;
+        auto extpubKeyPrefix = chainparams.Base58Prefix(CChainParams::Base58Type::EXT_PUBLIC_KEY);
```
jonasschnelli: Hmm.. I think this is incorrect. I could load a xpub (mainnet) on regtest.
Shouldn't `CBitcoinExtPubKey("<x|tpub>")` do the check for the correct encoding according to the chainparams?

src/wallet/wallet.cpp
```diff
@@ -3618,14 +3714,41 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
...
+        if (!bitcoinExtPubKey.SetString(hdWatchOnly.c_str(), extpubKeyPrefix.size())) {
+            InitError(_("Invalid ExtPubKey format"));
+            return NULL;
+        }
+        extPubKey = bitcoinExtPubKey.GetKey();
+    }
+
+    if (!fFirstRun && !hdWatchOnly.empty()) {
```
jonasschnelli: I got no warning when I started bitcoind again with a different `-hdwatchonly=xpub`.
Here there should be a check wether the wallet has already a watch-only xpub set, and if the user provides one, and it's different, it should warn or stop.

jonasschnelli:
This is a great PR. I really would like this to be in 0.15. It makes HWW signing much simpler and could be a pre-step for HWW support in Core.

src/wallet/wallet.cpp
saleemrashid: @jonasschnelli https://github.com/bitcoin/bitcoin/pull/9728#issuecomment-286187790

NicolasDorier:
I will rebase and pass on all your nits and better test/code the case where you pass a xpriv once #9294 get merged.

src/wallet/wallet.cpp
NicolasDorier: this is normally checked. Actually I am testing it on https://github.com/bitcoin/bitcoin/pull/9728/files#diff-08fcd10ff4d7d4f9f0249ba978b3b4d4R49 this is strange...

src/wallet/walletdb.h
NicolasDorier: How is it possible to get an extended private key here? if you provided a xpriv instead of xpub, it should have crashed at startup.

jonasschnelli: I used a `tpriv` which made this PR proceeding the hdwatchonly initialisation but then crashed at this point. Haven't looked into the details why the `tpriv` was accepted during init.

src/wallet/wallet.cpp
jonasschnelli: Ah.. Wasn't aware that @saleemrashid already pointed this out. Great.

src/wallet/walletdb.h
NicolasDorier: ah OK I know why... I do not verify the prefix

src/wallet/wallet.cpp
NicolasDorier: Interesting thing: https://github.com/bitcoin/bitcoin/blob/master/src/base58.h#L160 it does not check the actual prefix.
I think we should dump this ctor, it is using `Params()` and does not check the prefix.

src/wallet/walletdb.h
NicolasDorier: Issue fixed and tested

NicolasDorier:
1. Removed the addition of `CChainParams` into `InitLoadWallet`
2. Correctly error if not passing an extpubkey with the right network
3. Testing hdwatchonly parsing

NicolasDorier:
I am tempted to modify AvailableCoins in the case of HDWatchOnly, such that includeWatchOnly is always true. This would make it possible to use `sendtoaddress` and `fundrawtransaction` without modification for old code which was storing private keys on bitcoind before.

NicolasDorier:
Rebased:

* Support internal addresses of #9294
* Using BIP44 Friendly derivation @saleemrashid

NicolasDorier:
Rebased

saleemrashid:
ACK

instagibbs:
needs rebase, will review

NicolasDorier:
rebased

instagibbs:
Is there a way exposed to the user that their wallet is HD watch-only? Might be useful in `getwalletinfo`.

NicolasDorier:
just added info in `getwalletinfo`. I sadly can't compile because of https://github.com/bitcoin/bitcoin/issues/10622 and it seems my travis build is queued and does not want to run.

src/wallet/walletdb.cpp
```diff
@@ -57,6 +57,14 @@ bool CWalletDB::EraseTx(uint256 hash)
     return EraseIC(std::make_pair(std::string("tx"), hash));
 }
 
+bool CWalletDB::WriteKeyMeta(const CPubKey& vchPubKey, const CKeyMetadata& keyMeta)
+{
+    if (!batch.Write(std::make_pair(std::string("keymeta"), vchPubKey),
```
instagibbs: Please add braces for readability

Should you be using `WriteIC` like everywhere else?

NicolasDorier:
Ok my blinded push passed @instagibbs . Added some info in `getwalletinfo` as suggested.

src/wallet/wallet.cpp
```diff
@@ -93,77 +93,115 @@ CPubKey CWallet::GenerateNewKey(bool internal)
...
     // Compressed public keys were introduced in version 0.6.0
     if (fCompressed)
         SetMinVersion(FEATURE_COMPRPUBKEY);
 
-    CPubKey pubkey = secret.GetPubKey();
-    assert(secret.VerifyPubKey(pubkey));
+    if (!IsHDWatchOnly())
+        assert(secret.VerifyPubKey(pubkey));
 
     mapKeyMetadata[pubkey.GetID()] = metadata;
```
instagibbs: Perhaps move this into `AddKeyPubKey`

src/wallet/wallet.cpp
```diff
@@ -93,77 +93,115 @@ CPubKey CWallet::GenerateNewKey(bool internal)
...
+        pubkey = secret.GetPubKey();
     }
 
     // Compressed public keys were introduced in version 0.6.0
     if (fCompressed)
         SetMinVersion(FEATURE_COMPRPUBKEY);
 
-    CPubKey pubkey = secret.GetPubKey();
-    assert(secret.VerifyPubKey(pubkey));
+    if (!IsHDWatchOnly())
```
instagibbs: Please add braces.

test/functional/hdwatchonly.py
```diff
@@ -0,0 +1,104 @@
...
+
+        # generatetoaddress with p2pkh
+        assert_equal(unspent[0]['solvable'], True)
+
+        # generate mine to p2pk, so let's just be sure we can solve it
+        assert_equal(unspent[1]['solvable'], True)
+
+        self.stop_nodes()
+
+        # check fails gracefully if any mess up with hdwatchonly parameter
```
instagibbs: suggested rewording:

"check for graceful failure due to any invalid hdwatchonly parameters"

test/functional/hdwatchonly.py
```diff
@@ -0,0 +1,104 @@
...
+        # should not restart if hdwatchonly is different from the current one
+        assert_start_raises_init_error(0, self.options.tmpdir, ['-hdwatchonly=tpubD6NzVbkrYhZ4YTNYPw3XmSoBRZWmfn8mRerv3SEaC8UFiz5geKgCJH42cp9KUzRcfQNSuCQgdM1grUH7FgWYahWKDST3E9NYJMBwMKooTaY'], 
+        'Cannot specify hdwatchonly on an already existing wallet')
+
+        # check the hdkeypath was persisted
+        self.nodes = self.start_nodes(self.num_nodes, self.options.tmpdir, [[],[]])
+        validated_address = self.nodes[0].validateaddress('mxKeRQP6gTdCW6jHhn9FW8bGXD8W1UpR6n')
+        assert_equal(validated_address['hdkeypath'], 'm/0/1')
+
+        # check the hd key was persisted
```
instagibbs: s/was/has/

instagibbs (approved):
tACK, works as expected

NicolasDorier:
Thanks for the review @instagibbs I addressed your nit, can you re tACK once ?

instagibbs:
re-tACK https://github.com/bitcoin/bitcoin/pull/9728/commits/067b117b523208d66ade575e11ca76df964bf6ed

instagibbs:
We could actually also return the xpub in `validateaddress` as well...

instagibbs:
Included the return in `validateaddress` here: https://github.com/instagibbs/bitcoin/commit/b8e1316721a72fa7a0de4e9c8497746f022e9881

Please feel free to take.

Using this PR "in production" with no issues.

src/wallet/wallet.cpp
```diff
@@ -1868,8 +1868,18 @@ bool CWalletTx::IsTrusted() const
...
+        const auto& isMine = pwallet->IsMine(parentOut);
+        if (isMine != ISMINE_SPENDABLE)
+        {
+            // If the wallet is hdwatchonly, check if it is a key we generated
+            if (!pwallet->IsHDWatchOnly() || isMine != ISMINE_WATCH_SOLVABLE)
+                return false;
+            const auto& meta = pwallet->mapKeyMetadata;
+            auto it = meta.find(CScriptID(parentOut.scriptPubKey));
+            if (it == meta.end() ||
+                it->second.hdKeypath.empty())
```
instagibbs: I think we may need to generalize this check a bit more. If the script here is a multisig in which the watch-only wallet owns all the keys, this will return false since there is no hdKeypath.

Practically this means that spending 0-conf p2sh funds is a dicey experience. 

Perhaps something like `pwalletMain->IsHDWatchOnly(CScriptID& scriptID)` to make this check, which accounts for this situation?

instagibbs: Since we're not even required to be backwards compatible, I think adding a new piece of key metadata `fIsHDWatchOnly`, and adding that during `importaddress`(or some new call) when all keys are hdwatchonly keys(or a mixture of watchonly and stored private keys?).

NicolasDorier: Does `IsTrusted()`  returns `ISMINE_SPENDABLE` when for a multi sig output where we know all the keys?

NicolasDorier:
Rebased and included @instagibbs commit.

src/wallet/wallet.cpp
instagibbs: Yes. I think the case where any of the keys are unknown/HDWatchonly, it returns "False" if unconfirmed.

instagibbs: If you have the private keys, yes, because of check a few lines above.

NicolasDorier: So your goal is that `importaddress` scriptPubKeys should be considered trusted if we are in hdwatchonly mode ?

instagibbs: If we completely those addresses via hdwatchonly keys and/or privkeys, yes.

instagibbs: Currently in this mode you can import a privkey and spend unconfirmed outputs. Likewise you can spend unconfirmed hdwatchonly outputs. You cannot however spend a mixture of the two in a multisig, or pure hdwatchonly p2sh.

instagibbs:
Not trying to completely sidetrack this PR, but this is for future PR consideration:

@NicolasDorier https://github.com/instagibbs/bitcoin/commit/bb72d12cf6f737e4e7d43a90d7446f12a7e18a60  

This is the implementation that works for unconfirmed p2sh multisig where all keys are mixture of imported privkeys and hdwatchonly. Pretty ugly though and quite special-cased. Longer-term I'd look to extend ISMINE instead to have a value which covers this consideration and make this whole IsTrusted logic much simpler.

NicolasDorier:
I think making more general case with SPENDABLE_WATCHONLY is preferable.
I sadly don't have too much time these days, but I am interested into fixing it during 27-28.

src/wallet/wallet.cpp
```diff
@@ -93,77 +93,114 @@ CPubKey CWallet::GenerateNewKey(bool internal)
...
-    CExtKey masterKey;             //hd master key
-    CExtKey accountKey;            //key at m/0'
-    CExtKey chainChildKey;         //key at m/0'/0' (external) or m/0'/1' (internal)
-    CExtKey childKey;              //key at m/0'/0'/<n>'
-
     // try to get the master key
-    if (!GetKey(hdChain.masterKeyID, key))
+    if (GetKey(hdChain.masterKeyID, key))
+    {
+        // for now we use a fixed keypath scheme of m/0'/0'/k
```
jtimon: it seems like you are repeating yourself a little bit here. Perhaps a common function can be extracted to avoid some code duplication between this if and the next one?

NicolasDorier: I tried to think about this, but this was not as easy. I preferred keeping this way as it makes review easier: I did not touch what was already working.

NicolasDorier:
Rebased, will soon open a new PR, as discussed with @instagibbs `watchonlyhd` should be renamed into `externalhd`.  Reason is that we want the generated addresses to behave exactly as if they were done by a normal wallet which has the keys. Those generated addresses might be able to be signed correclty by a HW module contrary to watchonly.

Thus we keys generated by externalhd wallet should not be considered watchonly.

src/wallet/rpcwallet.cpp
```diff
@@ -2510,7 +2510,8 @@ UniValue getwalletinfo(const JSONRPCRequest& request)
             "  \"keypoolsize_hd_internal\": xxxx, (numeric) how many new keys are pre-generated for internal use (used for change outputs, only appears if the wallet is using this feature, otherwise external keys are used)\n"
             "  \"unlocked_until\": ttt,           (numeric) the timestamp in seconds since epoch (midnight Jan 1 1970 GMT) that the wallet is unlocked for transfers, or 0 if the wallet is locked\n"
             "  \"paytxfee\": x.xxxx,              (numeric) the transaction fee configuration, set in " + CURRENCY_UNIT + "/kB\n"
-            "  \"hdmasterkeyid\": \"<hash160>\"     (string) the Hash160 of the HD master pubkey\n"
+            "  \"hdmasterkeyid\": \"<hash160>\"   (string) the Hash160 of the HD master pubkey\n" +
```
jonasschnelli: nit: missing two whitespaces before `(string)`.

src/wallet/wallet.cpp
```diff
@@ -95,79 +95,116 @@ CPubKey CWallet::GenerateNewKey(CWalletDB &walletdb, bool internal)
...
 
-    if (!AddKeyPubKeyWithDB(walletdb, secret, pubkey)) {
-        throw std::runtime_error(std::string(__func__) + ": AddKey failed");
+    if(!IsExternalHD()) {
+        if (!AddKeyPubKeyWithDB(walletdb, secret, pubkey)) {
+            throw std::runtime_error(std::string(__func__) + ": AddKey failed");
+        }
+    }
+    else {
+        if (!AddWatchOnly(pubkey, metadata, nCreationTime))
```
jonasschnelli: nit: brackets

src/wallet/wallet.cpp
```diff
@@ -95,79 +95,116 @@ CPubKey CWallet::GenerateNewKey(CWalletDB &walletdb, bool internal)
...
+        CExtPubKey childKey;              //key at m/0/<n>
+
+        // derive m/x
+        masterKey.Derive(chainChildKey, internal ? 1 : 0);
+
+        // derive child key at next index, skip keys already known to the wallet
+        do {
+            if (internal) {
+                chainChildKey.Derive(childKey, hdChain.nInternalChainCounter);
+                metadata.hdKeypath = "m/1/" + std::to_string(hdChain.nInternalChainCounter);
```
jonasschnelli: for other reviewers: the account key level is dropped here because it's hardened in BIP44. Maybe a source code comment would be good.

jonasschnelli:
Re-Concept ACK. This should really be something we want in 0.16.
I dislike the fact that the extended pubkey needs to be provided via a startup argument, but I guess as long as there is no runtime wallet creation RPC we have to go with that.

jonasschnelli:
Needs rebase.

NicolasDorier:
@jonasschnelli  We talked about this PR at Tokyo with @instagibbs and @sipa . Basically we agreed that contrary to what this PR does, the generated keys should not be considered as WATCH_ONLY but as SPENDABLE, even if core do not have the keys, it is still signable.

This mean that we need to refactor the wallet to first decouple a new class: `WatchOnlyKeyStore` from the `CCryptoKeyStore`. Then replace `CCryptoKeyStore` by a `CExternalHDKeyStore` to handle signing.

I tried to do that in a separate PR, but this goes deep into the rabbit hole, I will need more work on it.

wtogami:
What is the status of this?

NicolasDorier:
So @saleemrashid is using it for hardware support in Bitcoin Core. I am using it in prod for quite a while, so this is stable.

However, this is not the best way to do, ideally we should do like I detailed here https://github.com/bitcoin/bitcoin/pull/9728#issuecomment-322953243 .
This however requires deeper refactoring which I started on https://github.com/bitcoin/bitcoin/pull/10980 .

This would need deeper review, and sadly I am out of time these days. So if someone wants to take it over, it would probably be better.

instagibbs:
For those interested, I'm rebasing at https://github.com/instagibbs/bitcoin/tree/externalhd2

jonasschnelli:
Rebase required...

NicolasDorier:
@instagibbs  is maintaining a rebased do you want to supersede this PR?

instagibbs:
Sure I can open a replacement PR if people are interested. I'm not actively pushing for merge at this time however.

NicolasDorier:
So the reason I put that back on the shelf, is that I think we want to do that after a major overall wallet refactoring that @sipa  proposed on https://gist.github.com/sipa/125cfa1615946d0c3f3eec2ad7f250a2#future-design

This PR is creating new addresses as watch-only, this is however not what we want. We want addresses to be `MINE`, even if we don't have private key. Managing to refactor the code to reach this goal is a can of worms which might be solved by @sipa redesign proposal.

NicolasDorier:
Closing this it seems that we need deeper wallet refactoring as specified by sipa on https://gist.github.com/sipa/125cfa1615946d0c3f3eec2ad7f250a2#future-design 

@instagibbs is maintaining https://github.com/instagibbs/bitcoin/tree/externalhd2 for those interested to continue this work.

PR #6654, opened 2015-09-09
Mempool package tracking

sdaftuar:
This is a preparatory pull to try to make reviewing #6557 easier.

In #6557, I added tracking packages of transactions to the mempool (for each tx, we track all the "descendant" transactions that depend on that tx), in order to make the mempool limiting code more effective.  This PR is a standalone implementation of mempool descendant tracking, including the policy limits on transaction chains (limiting ancestors and descendants) proposed in #6557.

I'll rebase that pull off these commits assuming we can agree on this approach.

PR #9732, opened 2017-02-09
[Trivial] Remove nonsense #undef foreach

jnewbery:
This PR removes `#undef foreach` from txmempool.h. That preprocessor
directive was added in 34628a18070064e75b35f28fd6a43d5c23832eb8
(https://github.com/bitcoin/bitcoin/pull/6654) but there doesn't appear to be a
`#define foreach` anywhere in the codebase. The `#undef directive` doesn't
actually do anything.

### DIFF
--- a/src/txmempool.h
+++ b/src/txmempool.h
@@ -20,7 +20,6 @@
 #include "sync.h"
 #include "random.h"
 
-#undef foreach
 #include "boost/multi_index_container.hpp"
 #include "boost/multi_index/ordered_index.hpp"
 #include "boost/multi_index/hashed_index.hpp"


MarcoFalke:
gives same binaries for me ACK 467df39

jtimon:
utACK 467df39

dcousens (approved):

sipa:
Obvious ACK.

This was added in #6654, without explanation.

paveljanik:
ACK https://github.com/bitcoin/bitcoin/commit/467df390f0a620859c46e49d0894b0b7f8dc55ad

laanwj:
>  but there doesn't appear to be a #define foreach anywhere in the codebase.

IIRC Qt defines this, and there was a conflict with that and some other dependency. But if it's no longer necessary that's great.

fanquake:
utACK 467df39

PR #9735, opened 2017-02-10
devtools: Handle Qt formatting characters edge-case in update-translations.py

laanwj:
If both numeric format specifiers and "others" are used, assume we're dealing with a Qt-formatted message. In the case of Qt formatting (see https://doc.qt.io/qt-5/qstring.html#arg) only numeric formats are
replaced at all. This means `(percentage: %1%)` is valid (which was introduced in #9461), without needing any kind of escaping that would be necessary for strprintf.  Without this, this function would wrongly detect `%)` as a printf format specifier (and an assertion triggers).

Also contains periodic translations update from Transifex.

### DIFF
--- a/contrib/devtools/update-translations.py
+++ b/contrib/devtools/update-translations.py
@@ -65,6 +65,14 @@ def split_format_specifiers(specifiers):
         else:
             other.append(s)
 
+    # If both numeric format specifiers and "others" are used, assume we're dealing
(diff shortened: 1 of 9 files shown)


jonasschnelli:
Looks good to me: 7179e7ca0db493dccbfd88f64738252844aab6ff

theuni:
ACK 7179e7ca0db493dccbfd88f64738252844aab6ff

PR #9736, opened 2017-02-10
Pre-0.14.0 hardcoded seeds update

laanwj:
Update hardcoded seeds list from @sipa's DNS seed.

Also update seeds tooling to Python 3.


gmaxwell (commented):
PATTERN_AGENT needs an update

laanwj:
> PATTERN_AGENT needs an update

What does it need to be updated to?

gmaxwell:
I believe this would be prudent:

PATTERN_AGENT = re.compile(r"^(\/Satoshi:0\.12\.(0|1|99)\/|\/Satoshi:0\.13\.(0|1|2|99)\/)$")

(Adds the newest versions, drops everything multiple softforks ago.)

laanwj:
Ok, updated and regenerated with that.

jonasschnelli:
A thought: we can (not sure if we should) filter the seed.txt for particular service flags (like `NODE_WITNESS`).

laanwj:
Regarding the possibility: yes, the make_seeds script filters for bit 1 at the moment. This could be extended to others:
```
    # Require service bit 1.
    ips = [ip for ip in ips if (ip['service'] & 1) == 1]
```

gmaxwell (approved):
I reproduced the nodes_main with acceptable similarity using data collected at about the same time wumpus updated the PR.   I checked that many of them were reachable, and also checked that the nodes_main didn't include any IPs I have banned (I was surprised it didn't).  Also confirmed the seeds in the source matched the text file.

laanwj:
Thanks for checking @gmaxwell

PR #9737, opened 2017-02-10
Don't disconnect feeler connections prematurely

rebroad:
This allows the usual logic (e.g. advertising address, sending alert, etc) to run as usual for feeler connections, regardless of whether the node provides the expected services.


sipa:
I don't think all that logic is needed for feeler connections anyway, and we could probably disconnect earlier in the VERSION message processing for them regardless.

fanquake:
Closing as it needs a rebase, and has a lack of comments.

PR #7562, opened 2016-02-19
Bump transaction version default to 2

btcdrak:
This should be merged after the BIP68 softfork enforces.

EDIT: BIP68 has long since activated so miners are running min 0.12.1 and therefore will mine v2 txs. Also there are sufficient nodes on the network relaying v2 txs

PR #7835, opened 2016-04-07
Version 2 transactions remain non-standard until CSV activates

sdaftuar:
Before activation, such transactions might not be reliably mined, so don't allow into the mempool.

Please tag for backport to 0.12.1

PR #7648, opened 2016-03-06
BIP9 versionbits softfork for BIP68, BIP112 and BIP113

btcdrak:
## General

This PR softforks 3 lock-time related BIPs which are all implemented in `master` as mempool-only logic at the moment.
1. [BIP68](https://github.com/bitcoin/bips/blob/master/bip-0068.mediawiki) sequence locks for relative locktime;
2. [BIP112](https://github.com/bitcoin/bips/blob/master/bip-0112.mediawiki) CHECKSEQUENCEVERIFY;
3. [BIP113](https://github.com/bitcoin/bips/blob/master/bip-0113.mediawiki) Median Time Past.
### Dependencies
- BIP68 (sequence locks) independently enforces the same semantics BIP113 (MTP)
- BIP112 (CSV) relies on BIP68
### Relay policy for BIP113

BIP113 mempool-only policy was deployed with Bitcoin Core 0.11.2 at the same time as the BIP65 CLTV softfork so the policy is in wide use (at  least 70% of nodes) as well as all the miners who upgraded to 0.11.2.
### Relay policy for v2 transactions

BIP68/112 rely on v2 transactions. Currently only v1 transactions are relayed, so it is necessary to change the relay policy to allow v2. This will be done at the same time as softfork deployment and will have the net effect that once the softfork enforces, we can be pretty sure miners will mine v2 transactions.

At a later date once enough nodes upgrade and we're sure v2 transaction will be relayed efficiently, we can bump the default transaction version in core see #7562.

PR #9739, opened 2017-02-10
Fix BIP68 activation test

jnewbery:
https://github.com/bitcoin/bitcoin/pull/7562 made version 2 transactions standard, which should have broken the test in bip68-sequence.py that version 2 transactions are non-standard before BIP 68 activation. However, there was a syntax error in the test case, so it didn't break.

In fact, it's fine for version 2 transactions to be standard before BIP 68 activation since BIP 68 was activated in July 2016, so it'd have to be a pretty big re-org to go back to pre-activated times (and even then this is a non-consensus change).

Best fix here is to remove the test that version 2 transactions are non-standard prior to BIP68 activation.

### DIFF
--- a/qa/rpc-tests/bip68-sequence.py
+++ b/qa/rpc-tests/bip68-sequence.py
@@ -51,14 +51,11 @@ class BIP68Test(BitcoinTestFramework):
         print("Running test BIP68 not consensus before versionbits activation")
         self.test_bip68_not_consensus()
 
-        print("Verifying nVersion=2 transactions aren't standard")
-        self.test_version2_relay(before_activation=True)
+        print("Verifying nVersion=2 transactions are standard")
+        self.test_version2_relay()
 
         print("Activating BIP68 (and 112/113)")
         self.activateCSV()
-
-        print("Verifying nVersion=2 transactions are now standard")
-        self.test_version2_relay(before_activation=False)
 
         print("Passed\n")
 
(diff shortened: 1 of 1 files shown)


jnewbery:
@sdaftuar

jnewbery:
@morcos has pointed out version 2 transactions were actually made standard in the CLTV PR #7648 (not #7562 as I claimed above). That means that we were relaying version 2 transactions before CLTV was activated. Oops.

Anyway, the point still stands. BIP68 is now well and truly activated, so we don't need to test pre-activation standardness.

gmaxwell:
> That means that we were relaying version 2 transactions before CSV was activated. Oops.

I think that is fine and that it was the intent-- so long as they were _valid_ CSV transactions, which they were, because the CSV rules were enforced against the mempool prior to activation.

qa/rpc-tests/bip68-sequence.py
```diff
@@ -51,15 +51,12 @@ def run_test(self):
         print("Running test BIP68 not consensus before versionbits activation")
         self.test_bip68_not_consensus()
 
-        print("Verifying nVersion=2 transactions aren't standard")
-        self.test_version2_relay(before_activation=True)
+        print("Verifying nVersion=2 transactions are standard")
+        self.test_version2_relay()
```
sdaftuar: I think we should move this to below the CSV activation.  Testing that relaying of nVersion=2 transactions post activation makes sense, whereas testing that it works pre-activation is strange -- it's okay in this case that we allow it (because on mainnet activation happened a long time ago), but that is not how the code used to behave prior to activation (because relaying those transactions prior to validation opens up some DoS vectors).

sdaftuar (changes requested):
@gmaxwell This is an aside but for the sake of anyone following this conversation, we actually didn't start relaying version 2 transactions before CSV was activated -- we merged #7835 prior to release, to prevent that from happening.  (If there is a class of transactions that we think are valid but that miners for whatever reason are likely to not mine, then that opens up some mempool-related DoS vectors.)

jnewbery:
ok, I've moved the standardness check to after BIP68 activation. I've also added a comment to say transaction version 2 is *always* standard for current bitcoin core versions to avoid any confusion for people reading the code. I've verified that the test still runs.

Once you're happy and have ACKed the changes, I'll squash for merging.

jnewbery:
@sdaftuar pointed out a missing closing parentheses. Fixed in https://github.com/bitcoin/bitcoin/pull/9739/commits/f5aba8a3c5f10bf42595cbaf0439976cd48c57b1

sdaftuar:
utACK f5aba8a

MarcoFalke:
utACK f5aba8a3c5f10bf42595cbaf0439976cd48c57b1

Issue #9564, opened 2017-01-15
dumpwallet behavior


unsystemizer:
In v0.13: `dumpwallet` "dumps all wallet keys in a human-readable format." (according to console help).

First, it's hard to tell *where* it dumps them.
```
Arguments:
1. "filename"    (string, required) The filename
```
I executed the action and couldn't find the file. Another way to look at this is "I executed the action and nothing happened".
Well, it turns out it did work, but the file ended up in the Bitcoin "root" directory (for me, `c:\Bitcoin`). 
I bet some users would think that the command simply didn't work and you can see where it goes from there.

Second, the action behavior isn't particularly intuitive. 
For example, I'm in `E:\backup` and I run 
`c:\bitcoin\bin\bitcoin-cli.exe dumpwallet wallet.txt`
I think this shouldn't dump the wallet to c:\bitcoin\wallet.txt.

`dumpwallet e:\backup\wallet.txt`works as expected (both from debug console and from the CLI).

This could be classified as a doco bug (e.g. change "The filename" to "The full path and filename"), or it could be improved in the code.

OS: English Windows 10 x64.

jonasschnelli:
More documentation is always good. Hence the fact that the Console/RPC interface if for experience users and (maybe) most of them are aware that a dump-to-filename command better gets supplied with a full absolute path.
Related: https://github.com/bitcoin/bitcoin/pull/9330

cdecker:
It appears that this was addressed in #9740, so this can be closed.

unsystemizer:
@MarcoFalke it seems this should be reopened as doesn't seem fixed (at last not on Linux; originally I reported this with 0.13.2 for Windows).

Current help:

`"1. \"filename\"    (string, required) The filename with path (either absolute or relative to bitcoind)\n"`
https://github.com/bitcoin/bitcoin/blob/v0.15.1/src/wallet/rpcdump.cpp#L600

Below we can see that relative path isn't relative to bitcoind, but to "user's home directory"

```
$ /home/user/Downloads/bitcoin-0.15.1/bin/bitcoin-cli dumpwallet github.txt
{
  "filename": "/home/user/github.txt"
}
```

Trying to use `/tmp/Downloads`:

```
$ cd /tmp
$ /home/user/Downloads/bitcoin-0.15.1/bin/bitcoin-cli dumpwallet Downloads/github.txt
{
  "filename": "/home/user/Downloads/github.txt"
}
```


In my case:

[...]

PR #8278, opened 2016-06-28
Forking daemon

ghost:
When started with cron, the sendmail process hung around waiting for its stdin, bitcoind's stdout/err, to close. Fix that using daemon(3) where possible and doing its work where not.

Also makes the forking error messages more consistent and, by using strerror, more explanatory.

PR #7759, opened 2016-03-28
[WIP] rest: Stream entire utxo set

laanwj:
This is by no means ready, but anyhow this builds on #7756 and
- Adds a streaming API to the HTTP server. This allows streaming data to the client chunk by chunk, which is useful when not the entire data is available at once or it is huge and wouldn't fit (efficiently) in memory.
- Allows downloading the entire UTXO set through `/rest/utxoset`. This is a raw dump of all outputs, the state normally hashed by `gettxoutsetinfo`. The dump is performed in the background by making use of leveldb snapshotting, so without keeping cs_main locked.
  - This can be useful for analysis purposes if you don't want to mess with bitcoin core's database
  - Filename (via content-disposition) is `utxoset-<height>-<bestblockhash>.dat`. Also a custom `X-Best-Block` and `X-Block-Height` header is added.

It matches:

``` bash
$ src/bitcoin-cli -datadir=/store/tmp/testbtc gettxoutsetinfo
{
...
  "hash_serialized": "5017f82bbb82a8199ae0fbaa9e5881a0c82575db89e6edd5b39414b35299363b",
...
}
$ wget --content-disposition http://127.0.0.1:8332/rest/utxoset 
2016-03-28 22:58:32 (44.3 MB/s) - ‘utxoset-404681-0000000000000000034854f5a3ab27cfbc220a42c75061dd13d2067cda71191d.dat’ saved [1291578967]
$ ~/bin/dsha256 utxoset-404681-0000000000000000034854f5a3ab27cfbc220a42c75061dd13d2067cda71191d.dat
5017f82bbb82a8199ae0fbaa9e5881a0c82575db89e6edd5b39414b35299363b utxoset
```
## TODO
- ~~Rebase after #7756 merged~~
- Sensibly name and split up commits
- Clean up and split up code
- Actually handle errors (you can crash a worker thread right now by disconnecting while downloading)
- The timeout actually runs _while_ downloading, causing it to break off after downloading. I don't understand why this is. You can work around it with `-rpcservertimeout=6000` or such.
- ~~UTXO set dump doesn't contain keys (?) I'm not sure this format is actually useful this way (see #7758)~~ (fixed in #7848)
- Other formats, potentially ^^

Note that the HTTP streaming API could in principle also be used for other large data (say, wallet backups), or even for websocket-like event notification.

PR #9740, opened 2017-02-11
Add friendly output to dumpwallet

aideca:
Add friendly output to dumpwallet #9564.

* Add help message.
* Show created dumpfile path.
* Add rpc test.

### DIFF
--- a/src/wallet/rpcdump.cpp
+++ b/src/wallet/rpcdump.cpp
@@ -579,7 +579,11 @@ UniValue dumpwallet(const JSONRPCRequest& request)
             "dumpwallet \"filename\"\n"
             "\nDumps all wallet keys in a human-readable format.\n"
             "\nArguments:\n"
-            "1. \"filename\"    (string, required) The filename\n"
+            "1. \"filename\"    (string, required) The filename with path (either absolute or relative to bitcoind)\n"
+            "\nResult:\n"
+            "{                           (json object)\n"
+            "  \"filename\" : {        (string) The filename with full absolute path\n"
+            "}\n"
             "\nExamples:\n"
             + HelpExampleCli("dumpwallet", "\"test\"")
             + HelpExampleRpc("dumpwallet", "\"test\"")
@@ -590,7 +594,9 @@ UniValue dumpwallet(const JSONRPCRequest& request)
     EnsureWalletIsUnlocked(pwallet);
 
     std::ofstream file;
-    file.open(request.params[0].get_str().c_str());
+    boost::filesystem::path filepath = request.params[0].get_str();
+    filepath = boost::filesystem::absolute(filepath);
+    file.open(filepath.string().c_str());
     if (!file.is_open())
         throw JSONRPCError(RPC_INVALID_PARAMETER, "Cannot open wallet dump file");
 
@@ -655,7 +661,11 @@ UniValue dumpwallet(const JSONRPCRequest& request)
     file << "\n";
     file << "# End of dump\n";
     file.close();
-    return NullUniValue;
+
+    UniValue reply(UniValue::VOBJ);
+    reply.push_back(Pair("filename", filepath.string()));
+
+    return reply;
 }
 
 
--- a/test/functional/wallet-dump.py
+++ b/test/functional/wallet-dump.py
@@ -4,6 +4,8 @@
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 """Test the dumpwallet RPC."""
 
+import os
+
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import (start_nodes, start_node, assert_equal, bitcoind_processes)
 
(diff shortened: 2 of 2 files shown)


src/wallet/rpcdump.cpp
```diff
@@ -572,7 +576,9 @@ UniValue dumpwallet(const JSONRPCRequest& request)
     EnsureWalletIsUnlocked();
 
     ofstream file;
-    file.open(request.params[0].get_str().c_str());
+    boost::filesystem::path filepath = request.params[0].get_str();
+    filepath = absolute(filepath);
```
laanwj: Interesting. Shouldn't it be only accepting absolute paths in the first place?
Otherwise it will depend on the directory that bitcoind happens to have been started in, that'd be pretty crappy API design.

MarcoFalke:
utACK 48abaeae39c0485f864dd67cc4a2fcac977bbdc3

src/wallet/rpcdump.cpp
aideca: Thanx.
IMO, it is better to inform users that path argument is from the viewpoint of bitcoind either absolute nor relative.
The problem is that users expects path argument is the viewpoint of RPC-caller, but actually the one of bitcoind. The former is not same as the latter even if absolute path if they executed on different docker.
So i added doc and output at this time.

jonasschnelli: Maybe an absolute path or if not, relative to the data-dir? If possible?

JeremyRubin: Also is there a "namespaced" name for this function? E.g., boost::filesystem::absolute? It'd be good to use the fully qualified name, otherwise it's difficult to hunt down such dependencies/lookup documentation.

I don't have a firm opinion to add to the above conversation though, except that I would add a way to get dumpwallet to serialize results across network (although dangerous if unencrypted/authenticated link) in case the user wants the wallet dump locally (RPC may even be called on a different machine let alone different directory as @aideca points out).

JeremyRubin:
utACK 48abaea.

I also want to explicitly approve of the returning of a JSON rather than a plain string, it may be useful (in future work?) to augment this call's return value with other data as well (e.g., which user owns the dump file, how many keys were dumped, etc).

src/wallet/rpcdump.cpp
laanwj: > Maybe an absolute path or if not, relative to the data-dir? If possible?

Good idea. It's our (effective) convention for command line arguments as well. I think it would be good to `chdir` the daemon to the data directory at start. This also avoids issues with the daemon holding on to resources (e.g. the starting directory could be deleted later) discussed in #8278.

>  RPC may even be called on a different machine

A way to dump the wallet over the network could be useful. The practical problem is that there is currently no way to do streaming writes to the network from RPC code, so the entire wallet would have to be dumped in memory, which would result in huge memory and parsing/generating overhead. I looked at streaming to the HTTP in a different context here: #7759.

Also yes there's a bit of a security concern though it's no different from dumpprivkey and friends...

laanwj:
This seems ready to be merged once 0.14 is branched off.

src/wallet/rpcdump.cpp
```diff
@@ -561,7 +561,11 @@ UniValue dumpwallet(const JSONRPCRequest& request)
             "dumpwallet \"filename\"\n"
             "\nDumps all wallet keys in a human-readable format.\n"
             "\nArguments:\n"
-            "1. \"filename\"    (string, required) The filename\n"
+            "1. \"filename\"    (string, required) The filename (if not full path, relative to bitcoind path)\n"
+            "\nResult:\n"
+            "{                           (json object)\n"
+            "  \"dumpfilepath\" : {        (string) The filename\n"
```
jnewbery: I'd prefer to use the name "filename" (since this isn't just the path). You could change the help text to say "The filename with full absolute path"

src/wallet/rpcdump.cpp
```diff
@@ -561,7 +561,11 @@ UniValue dumpwallet(const JSONRPCRequest& request)
             "dumpwallet \"filename\"\n"
             "\nDumps all wallet keys in a human-readable format.\n"
             "\nArguments:\n"
-            "1. \"filename\"    (string, required) The filename\n"
+            "1. \"filename\"    (string, required) The filename (if not full path, relative to bitcoind path)\n"
```
jnewbery: The text is a bit clunky here. Perhaps "The filename with path (either absolute or relative to bitcoind)" is better?

jnewbery (commented):
Definite concept ACK. A couple of nitty comments about language.

I agree that it'd be better to change the default location to be the bitcoin datadir if possible, but that might be out of scope of this small PR.

aideca:
Thank you for your comments and sorry for my late reply.

I just updated help text and qualified function name.
"chdir daemon to the data-dir" and "dump the wallet over the network" are quite interesting. But it's out of scope of this PR as jnewbery said.

achow101:
Needs rebase.

aideca:
Thanx! rebased & fixuped.

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/9740/commits/164019d611e9d67fd3e66eb7f18033c9249764d4

PR #9745, opened 2017-02-12
[RPC] Getting confirmations command

34ro:
This is aimed at saving traffic to get transaction confirmations and improve rpc throughput.

To check transaction confirmations, we needs to call "getrawtransaction" or "gettransaction".
But to check confirmations of huge number of transaction, network is crowded and latency is up.
"getrawtransaction" and "gettransaction" includes constancy infomation, though confirmations is just changeable field.
I think it is good to split off "confirmations" from "getrawtransaction".
For this, add new command "getconfirmations" to rpc command.

### DIFF
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -123,6 +123,39 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
     }
 }
 
+UniValue getconfirmations(const JSONRPCRequest& request)
+{
+    if (request.fHelp || request.params.size() != 1)
+        throw runtime_error(
+                "getconfirmations \"txid\"\n"
+                        "\nReturns the number of transaction confirmations.\n"
+                        "\nResult:\n"
+                        "n    (numeric) The number of confirmations\n"
+                        "\nExamples:\n"
+                + HelpExampleCli("getconfirmations", "\"txid\"")
+                + HelpExampleRpc("getconfirmations", "\"txid\"")
+        );
+
+    LOCK(cs_main);
+
+    uint256 hash;
+    hash.SetHex(request.params[0].get_str());
+
+    CTransactionRef tx;
+    uint256 hashBlock;
+    if (!GetTransaction(hash, tx, Params().GetConsensus(), hashBlock, true))
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string("No such mempool or blockchain transaction"));
+
+    BlockMap::iterator mi = mapBlockIndex.find(hashBlock);
+    if (mi != mapBlockIndex.end() && (*mi).second) {
+        CBlockIndex *pindex = (*mi).second;
+        if (chainActive.Contains(pindex)) {
+            return 1 + chainActive.Height() - pindex->nHeight;
+        }
+    }
+    return 0;
+}
+
 UniValue getrawtransaction(const JSONRPCRequest& request)
 {
     if (request.fHelp || request.params.size() < 1 || request.params.size() > 2)
@@ -933,6 +966,7 @@ static const CRPCCommand commands[] =
 { //  category              name                      actor (function)         okSafeMode
   //  --------------------- ------------------------  -----------------------  ----------
     { "rawtransactions",    "getrawtransaction",      &getrawtransaction,      true,  {"txid","verbose"} },
+    { "rawtransactions",    "getconfirmations",       &getconfirmations,       true,  {"txid"} },
     { "rawtransactions",    "createrawtransaction",   &createrawtransaction,   true,  {"transactions","outputs","locktime"} },
     { "rawtransactions",    "decoderawtransaction",   &decoderawtransaction,   true,  {"hexstring"} },
     { "rawtransactions",    "decodescript",           &decodescript,           true,  {"hexstring"} },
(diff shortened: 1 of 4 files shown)


jonasschnelli:
Hmm..... so this would purely be a (bandwidth) performance optimisation? Right? I haven't measured it, but I expect that most of the CPU time gets spent in `GetTransaction()` and the http-serving parts.
I'm not sure if its worth to add another RPC call to solve a rather individual issue.

34ro:
yes, bandwidth performance.
I think it is not individual issue.
Wallet users most interested in tx confirmations because it is changeable value.
And we would like to detect confirmation as soon as possible to finish payment.
For this reason, getrawtransaction is called so many time and network performance is down.

jonasschnelli:
For normal use-cases, `get(raw)transaction` seems to be fast enough... 
Constantly polling is not efficient.
Something that could be a better solution would be RPC long polling (https://github.com/bitcoin/bitcoin/pull/7949).

It would be possible, to keep the http-connection open (long poll) until the confirmation-count got changed (could be check on every Connect/Disconnect block). This would reduce the bandwidth dramatically.

kallewoof:
A wallet would not use the RPC commands directly except if the bitcoin node is local, in which case bandwidth is not an issue. For a remote / arbitrary node, it would use some form of proxy which could have a reduced-bandwidth representation of the transactions easily enough.

jonasschnelli:
> A wallet would not use the RPC commands directly except if the bitcoin node is local, in which case bandwidth is not an issue. For a remote / arbitrary node, it would use some form of proxy which could have a reduced-bandwidth representation of the transactions easily enough.

Ideally yes. But I could guess there are some users tunnelling RPC to a remote machine (bandwidth matters) or connecting to a node on a different machine but in the same/trusted network (bandwidth should not matter that much).

JeremyRubin:
Hey! Just wanted to say nice work for your first PR @34ro :) In terms of code quality and formatting I think you've done a nice job & your implementation seems correct and pragmatic. Great to see that you even added tests!

Unfortunately, I agree somewhat with @jonasschnelli with regards to the necessity of this work. I don't think that the RPC bandwidth will be so large as to cause a problem in this specific case (most transactions don't have an insane amount of vins and vouts), but I do agree in general that this is worthwhile to think about (especially if you ever wanted to connect a mobile phone or something :) ) 

However, I do have one possible way of fixing your problem (excessive bandwidth) without needing to add a whole new RPC. You could change the order of the entries in the JSON such that the non variable-length fields are first, followed by the variable length fields. This way, your low-bandwidth client can read out just the first packet, parse far enough to read the confirmations field, and then ungracefully terminate the TCP connection. That's kind of a "dirty ugly hack" so it's only half-serious.

Another possible solution to this would be somewhat like long polling -- keep the http connection open after the RPC, and also allow a parameter fWatch to be sent. On every call to connectblock, another JSON could be sent over with just the delta (e.g., {"confirmations" : 1 } then {"confirmations": 2}). This is somewhat similar to what @jonasschnelli suggested but has the benefit of not closing the tcp socket until you're actually done with it.

laanwj:
I'd say the code looks good, but yes it doesn't seem there is enough reason presented to add yet another call to the API.

> And we would like to detect confirmation as soon as possible to finish payment.

The conceptual solution to this would be some kind of notification, not faster polling. @jonasschnelli has some work in introducing long-poll notifications in RPC, see #7949. That seems a more fruitful path.

laanwj:
Just wanted to add that while a better notification method would be nice, *practically* this is already possible with `-blocknotify`. The number of confirmations of a transaction can only change when a block comes in. The thing I'm not sure about is whether race conditions - it may be possible that the wallet has not processed the block yet when your script launches. A `-walletblocknotify` that is called when the wallet processed a new block could avoid that problem.

34ro:
Thank you for many advice.
Long polling is good idea.
I agree ```-blocknotify``` is practically way to detect confirmation fast.
I'll wait until #7949 is merged.

src/test/rpc_tests.cpp
```diff
@@ -49,6 +49,9 @@ BOOST_AUTO_TEST_CASE(rpc_rawparams)
     BOOST_CHECK_THROW(CallRPC("getrawtransaction not_hex"), std::runtime_error);
     BOOST_CHECK_THROW(CallRPC("getrawtransaction a3b807410df0b60fcb9736768df5823938b2f838694939ba45f3c0a1bff150ed not_int"), std::runtime_error);
 
+    BOOST_CHECK_THROW(CallRPC("getconfirmations"), std::runtime_error);
+    BOOST_CHECK_THROW(CallRPC("getconfirmations not_hex"), std::runtime_error);
```
luke-jr: Better to put these in `qa/rpc-tests`

src/rpc/rawtransaction.cpp
```diff
@@ -123,6 +123,39 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
...
+        );
+
+    LOCK(cs_main);
+
+    uint256 hash;
+    hash.SetHex(request.params[0].get_str());
+
+    CTransactionRef tx;
+    uint256 hashBlock;
+    if (!GetTransaction(hash, tx, Params().GetConsensus(), hashBlock, true))
```
luke-jr: This requires `txindex=1`, otherwise only works on unconfirmed transactions.

Seems like this ought to be a wallet RPC?

laanwj:
Closing this for now. It has been inactive for quite some time. Let me know if you start work on this again then I'll reopen.

PR #9751, opened 2017-02-13
Mini fix bincoin against bitcoin

ghost:
:)


jonasschnelli:
You have read the change log very attentively. :)
IIRC, we once decided that we won't fix trivial typos in change logs.

DATSEC:
@jonasschnelli
No I have not read the magazine I just have the program finds all the files and text 

Very handy for mass replacing the variables or text

laanwj:
We don't fix typos in historical change logs. And this is a really old one (0.3.20.01~dfsg-1) at that.

PR #8808, opened 2016-09-25
Do not shadow variables (gcc set)

paveljanik:
Followup to #8105.

The current set of PRs made `clang` compile `Wshadow` clean. Compiling with `gcc` revealed new needed changes. `gcc` is more strict in warnings. This is non-Qt part. I do not have an environment to test `gcc`+Qt build now.

PR #9753, opened 2017-02-13
Add static_assert to prevent VARINT(<signed value>)

ryanofsky:
Using VARINT with signed types is dangerous because negative values will appear to serialize correctly, but then deserialize as positive values mod 128.

This commit changes the VARINT macro to trigger a compile error by default if called with an signed value, and it updates existing broken uses of VARINT to pass a special flag that lets them keep working with no changes in behavior.

There is some discussion about this issue here: https://github.com/bitcoin/bitcoin/pull/9693#issuecomment-278701473. I think another good change along these lines would be to make `GetSizeOfVarInt` and `WriteVarInt` throw exceptions if they are passed numbers less than 0 to serialize. But unlike this change, that would be a change in runtime behavior, and need more consideration.

### DIFF
--- a/src/chain.h
+++ b/src/chain.h
@@ -91,7 +91,7 @@ struct CDiskBlockPos
 
     template <typename Stream, typename Operation>
     inline void SerializationOp(Stream& s, Operation ser_action) {
-        READWRITE(VARINT(nFile));
+        READWRITE(VARINT(nFile, VarIntMode::NONNEGATIVE_SIGNED));
         READWRITE(VARINT(nPos));
     }
 
@@ -386,13 +386,13 @@ public:
     inline void SerializationOp(Stream& s, Operation ser_action) {
         int _nVersion = s.GetVersion();
         if (!(s.GetType() & SER_GETHASH))
-            READWRITE(VARINT(_nVersion));
+            READWRITE(VARINT(_nVersion, VarIntMode::NONNEGATIVE_SIGNED));
 
-        READWRITE(VARINT(nHeight));
+        READWRITE(VARINT(nHeight, VarIntMode::NONNEGATIVE_SIGNED));
         READWRITE(VARINT(nStatus));
         READWRITE(VARINT(nTx));
         if (nStatus & (BLOCK_HAVE_DATA | BLOCK_HAVE_UNDO))
-            READWRITE(VARINT(nFile));
+            READWRITE(VARINT(nFile, VarIntMode::NONNEGATIVE_SIGNED));
         if (nStatus & BLOCK_HAVE_DATA)
             READWRITE(VARINT(nDataPos));
         if (nStatus & BLOCK_HAVE_UNDO)
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -834,18 +834,18 @@ static void ApplyStats(CCoinsStats &stats, CHashWriter& ss, const uint256& hash,
 {
     assert(!outputs.empty());
     ss << hash;
-    ss << VARINT(outputs.begin()->second.nHeight * 2 + outputs.begin()->second.fCoinBase);
+    ss << VARINT(outputs.begin()->second.nHeight * 2 + outputs.begin()->second.fCoinBase, VarIntMode::NONNEGATIVE_SIGNED);
     stats.nTransactions++;
     for (const auto output : outputs) {
         ss << VARINT(output.first + 1);
         ss << output.second.out.scriptPubKey;
-        ss << VARINT(output.second.out.nValue);
+        ss << VARINT(output.second.out.nValue, VarIntMode::NONNEGATIVE_SIGNED);
         stats.nTransactionOutputs++;
         stats.nTotalAmount += output.second.out.nValue;
         stats.nBogoSize += 32 /* txid */ + 4 /* vout index */ + 4 /* height + coinbase */ + 8 /* amount */ +
                            2 /* scriptPubKey len */ + output.second.out.scriptPubKey.size() /* scriptPubKey */;
     }
-    ss << VARINT(0);
+    ss << VARINT(0u);
 }
 
 //! Calculate statistics about the unspent transaction output set
--- a/src/serialize.h
+++ b/src/serialize.h
@@ -296,9 +296,31 @@ uint64_t ReadCompactSize(Stream& is)
  * 2^32:           [0x8E 0xFE 0xFE 0xFF 0x00]
  */
 
-template<typename I>
+/**
+ * Mode for encoding VarInts.
+ *
+ * Currently there is no support for signed encodings. The default mode will not
+ * compile with signed values, and the legacy "nonnegative signed" mode will
+ * accept signed values, but improperly encode and decode them if they are
+ * negative. In the future, the DEFAULT mode could be extended to support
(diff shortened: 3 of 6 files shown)


fanquake:
Failed on OSX:
```
qt/qt_bitcoin_qt-bitcoin.o `test -f 'qt/bitcoin.cpp' || echo './'`qt/bitcoin.cpp
In file included from qt/bitcoin.cpp:9:
In file included from ./qt/bitcoingui.h:12:
In file included from ./amount.h:9:
./serialize.h:311:16: error: constexpr function's return type 'void' is not a literal type
constexpr void CheckVarIntMode()
               ^
1 error generated.
make[1]: *** [qt/qt_bitcoin_qt-bitcoin.o] Error 1
```

MarcoFalke:
> constexpr function's return type 'void' is not a literal type

If this is c++14 code, I wonder why the other builds succeeded, given that we are at c++11.

CryptAxe:
Are the other systems using g++ > version 5?

laanwj:
Do we explicitly request compilation for stdc++11? Would make sense to do so, as the default can (and has) changed.

theuni:
This is a weird one.

g++ accepts ```constexpr void foo()``` even when using ```-std=c++11 -pedantic -Wall -Wextra```. I assume that's a bug, but I'm not certain without looking at the language spec itself. That's the case in 4.8-5.1, anyway.

Not much we can do about that, as far as I can tell. Nothing I tried made g++ complain about it.

@laanwj I agree. It turned out to be unrelated to this issue, but I whipped up https://github.com/theuni/bitcoin/commit/9829c54de2725037ee0702306cbaa99fc9aa1826 anyway. Will PR it.

@ryanofsky You can just use a struct with a constexpr ctor instead:
```c++
template<VarIntMode Mode, typename I>
struct CheckVarIntMode
{
    constexpr CheckVarIntMode() {
        static_assert(Mode != VarIntMode::DEFAULT || std::is_unsigned<I>::value, "Unsigned type required with mode DEFAULT.");
        static_assert(Mode != VarIntMode::BROKEN_SIGNED || std::is_signed<I>::value, "Signed type required with mode BROKEN_SIGNED.");
    }
};
```

ryanofsky:
Thanks @theuni, that's a clever workaround. Added in dd094d6c0aede5d9e4e22caa441cdacbc95beede.

Squashed dd094d6c0aede5d9e4e22caa441cdacbc95beede -> d1df497083df4116ae3bf2915cea0da6a13fa274 ([varint-assert.1](https://github.com/ryanofsky/bitcoin/commits/pr/varint-assert.1) -> [varint-assert.2](https://github.com/ryanofsky/bitcoin/commits/pr/varint-assert.2))

ryanofsky:
Rebased d1df497083df4116ae3bf2915cea0da6a13fa274 -> 8950418e538d02c2eb8b0ea76b9838e867e91feb ([pr/varint-assert.2](https://github.com/ryanofsky/bitcoin/commits/pr/varint-assert.2) -> [pr/varint-assert.3](https://github.com/ryanofsky/bitcoin/commits/pr/varint-assert.3)) because of conflict with renamed nVersion variable in #8808.

theuni:
Concept ACK. Though i'd prefer to just introduce a new macro rather than a default arg:
```VARINT(foo, VarIntMode::NONNEGATIVE_SIGNED)``` -> ```VARINT_NONNEGATIVE_SIGNED(foo)```

**Comment by laanwj** [comment]
This seems to be in limbo - @ryanofsky can you please reply to @theuni's comment?

**Comment by ryanofsky** [comment]
> Though i'd prefer to just introduce a new macro rather than a default arg

I guess I prefer the opposite. I think it's good if you can look at code like:

```
ss << VARINT(i);
ss << VARINT(i, VarIntMode::DEFAULT);
ss << VARINT(i, VarIntMode::NONNEGATIVE_SIGNED);
```

and know that all 3 encodings of `i` are going to use the same underlying serialization machinery, and that you can understand the differences just by looking up the `VarIntMode` enum, skipping all the macro definition goop.

An interface like:

```
ss << VARINT(i);
ss << VARINT_DEFAULT(i);
ss << VARINT_NONNEGATIVE_SIGNED(i);
```

just seems a lot more opaque. The implementation would also be a little longer, more duplicative, and more indirect.

ryanofsky:
Rebased 8950418e538d02c2eb8b0ea76b9838e867e91feb -> 46e649b8ff6e7a66957fc269a04ef4117bd34d7a ([pr/varint-assert.3](https://github.com/ryanofsky/bitcoin/commits/pr/varint-assert.3) -> [pr/varint-assert.4](https://github.com/ryanofsky/bitcoin/commits/pr/varint-assert.4))

laanwj:
> ss << VARINT(i);
> ss << VARINT(i, VarIntMode::DEFAULT);
> ss << VARINT(i, VarIntMode::NONNEGATIVE_SIGNED);

FWIW I don't like the repetition of VarInt here, the VarIntMode:: could be inside the macro instead of outside it.

ryanofsky:
@laanwj. Implemented your suggestion in af20ce5a90aadd56d504291c754e0c60ca1f8788. Let me know if you think I should squash it.

As mentioned before though, I prefer current change because VARINT() is a simple, dumb, single macro and regular VARINT(x) cases work exactly the same as before. The only verbose case is the VarIntMode::NONNEGATIVE_SIGNED one and IMO that is verbose in a good way, because if you are reading the code you can tell that VarIntMode::NONNEGATIVE_SIGNED is just a plain c++ value, not more macro magic goop, and you can click "VarIntMode" in your IDE to go to the definition of the enum, and see documentation and details about the encoding.

ryanofsky:
Please let me know if af20ce5a90aadd56d504291c754e0c60ca1f8788 is ok and does what you want then. I think both the implementation and the resulting interface are gross, but if this is actually what you want, I'll squash it.

ryanofsky:
Still need feedback on whether to merge af20ce5a90aadd56d504291c754e0c60ca1f8788 into this PR. I don't love the macro magic it adds, and I don't see much benefit in the change since only weird legacy code should ever need to use `NONNEGATIVE_SIGNED` (normal code should just use `VARINT(value)`). But I'm fine with it if it would help move things forward.

theuni:
I would still prefer a separate macro, so that makes 3 differing opinions here :)
But as they all do the same thing, and we could argue about syntactic sugar until the end of time, utACK to either approach here.

laanwj:
> I would still prefer a separate macro, so that makes 3 differing opinions here :)

Seperate macros was also one of my preferences, FWIW:

> But I wouldn't mind explicit VARUINT_U and VARINT_I or even VARUINT and VARSINT either...
(there's no need to make things longer and longer and more verbose)

I just don't like the overly verbose construction of passing a scoped enum into everything.

ryanofsky:
Rebased 649bfb00bfb738df1b2c41c3d0c982f430c85495 -> 499d95e278f34790660a2b9baf5525e0def1485a ([pr/varint-assert.5](https://github.com/ryanofsky/bitcoin/commits/pr/varint-assert.5) -> [pr/varint-assert.6](https://github.com/ryanofsky/bitcoin/commits/pr/varint-assert.6)) due to conflict with #12683

sipa:
utACK 499d95e278f34790660a2b9baf5525e0def1485a

PR #7025, opened 2015-11-16
[Qt] refactor and optimize proxy settings behavior

jonasschnelli:
Fixes #6749 and related to #7006.

This slightly refactors the way how setting a Proxy over the UI settings panel works. It extends and used the `QValidatedLineInput` class.

Two goals of this PRs:
- Setting a invalid proxy should no longer be possible (which could lead to a startup termination)
- An Invalid proxy get the "red" invalid input-field background (similar like an invalid bitcoin address in the pay-to-address input-field)

bildschirmfoto 2015-11-16 um 11 01 17

PR #9755, opened 2017-02-13
Bugfix: Qt/Options: Restore persistent "restart required" notice

luke-jr:
This got broken when updateProxyValidationState was added.

### DIFF
--- a/src/qt/optionsdialog.cpp
+++ b/src/qt/optionsdialog.cpp
@@ -258,6 +258,9 @@ void OptionsDialog::showRestartWarning(bool fPersistent)
 void OptionsDialog::clearStatusLabel()
 {
     ui->statusLabel->clear();
+    if (model && model->isRestartRequired()) {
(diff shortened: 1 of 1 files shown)


laanwj:
What exactly is broken here?

luke-jr:
updateProxyValidationState clears it, so the persistent message is never shown in practice

laanwj:
That seems pretty serious, tagging for 0.14

jonasschnelli:
Started testing...

jonasschnelli:
Thanks @luke-jr!
Tested ACK 0b4f273e9cf7166dbd103e116b5d991e359e673d. Indeed, this is an important fix for 0.14.

jonasschnelli:
Issue was introduced in #7025.

MarcoFalke:
utACK 0b4f273e9cf7166dbd103e116b5d991e359e673d

PR #9760, opened 2017-02-14
[wallet] Remove importmulti always-true check

ryanofsky:
No change in behavior.

Remove `nLowestTimestamp <= chainActive.Tip()->GetBlockTimeMax()` check from
importmulti, which is always true because nLowestTimestamp is set to the
minimum of the most recent block time and all the imported key timestamps,
which is necessarily lower than the maximum block time.

An alternative would to make initialize nLowestTimestamp to `numeric_limits<int64_t>::max()` so the check would actually do what it appears to be intended to do, and avoid ScanForWalletTransactions calls when all the imported key timestamps are in the future.

### DIFF
--- a/src/wallet/rpcdump.cpp
+++ b/src/wallet/rpcdump.cpp
@@ -1072,7 +1072,7 @@ UniValue importmulti(const JSONRPCRequest& mainRequest)
         }
     }
 
-    if (fRescan && fRunScan && requests.size() && nLowestTimestamp <= chainActive.Tip()->GetBlockTimeMax()) {
+    if (fRescan && fRunScan && requests.size()) {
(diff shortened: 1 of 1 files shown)


morcos:
utACK ee7caa0

I'd prefer to merge this for 0.14 also so that a future code change doesn't break the rescan because there is no grace period in this check.

gmaxwell (approved):
utACK.

laanwj:
This has a merge conflict now, either due to #9778 or #9761

ryanofsky:
Rebased ee7caa09d3eb79b43e361f5c9d56836e40c64224 -> ec1267f13b7d0b9b5058c6821cf8dbf74e02d17c ([multitaut.0](https://github.com/ryanofsky/bitcoin/commits/pr/multitaut.0) -> [multitaut.1](https://github.com/ryanofsky/bitcoin/commits/pr/multitaut.1)).

jtimon:
utACK ec1267f

PR #9490, opened 2017-01-08
Replace FindLatestBefore used by importmuti with FindEarliestAtLeast.

gmaxwell:
In spite of the name FindLatestBefore used std::lower_bound to try
 to find the earliest block with a nTime greater or equal to the
 the requested value.  But lower_bound uses bisection and requires
 the input to be ordered with respect to the comparison operation.
 Block times are not well ordered.

I don't know what lower_bound is permitted to do when the data
 is not sufficiently ordered, but it's probably not good.
 (I could construct an implementation which would infinite loop...)

To resolve the issue this commit introduces a maximum-so-far to the
 block indexes and searches that.

For clarity the function is renamed to reflect what it actually does.

An issue that remains is that there is no grace period in importmulti:
 If a address is created at time T and a send is immediately broadcast
 and included by a miner with a slow clock there may not yet have been
 any block with at least time T.

The normal rescan has a grace period of 7200 seconds, but importmulti
 does not.

PR #9137, opened 2016-11-11
WIP: Allow wallet key import RPCs to track TxOut amounts on -prune nodes (on top of #9306)

ryanofsky:
Allow importprivkey, importaddress, importpubkey, and importmulti RPCs to be
called with rescan=True option on nodes with pruned blockchains (started with
-prune option). Instead of returning errors, the RPCs will now look for
transactions spending to the key in the UTXO database, and display them in the
wallet.

Because the UTXO database doesn't store full transaction information (it
discards TxIns) this change introduces a new "fIncomplete" CWalletTx member to
distinguish these from normal wallet transactions.

This change makes the -prune mode usable with wallets that need to import keys
after the initial sync.

Fixes #8497.

PR #9761, opened 2017-02-14
Use 2 hour grace period for key timestamps in importmulti rescans

ryanofsky:
@gmaxwell pointed out the lack of grace period in https://github.com/bitcoin/bitcoin/pull/9490#issue-199407998.
    
The importwallet RPC which uses key timestamps in a similar way already has a 2 hour grace period.

### DIFF
--- a/src/wallet/rpcdump.cpp
+++ b/src/wallet/rpcdump.cpp
@@ -988,7 +988,8 @@ UniValue importmulti(const JSONRPCRequest& mainRequest)
             "                                                              or the string \"now\" to substitute the current synced blockchain time. The timestamp of the oldest\n"
             "                                                              key will determine how far back blockchain rescans need to begin for missing wallet transactions.\n"
             "                                                              \"now\" can be specified to bypass scanning, for keys which are known to never have been used, and\n"
-            "                                                              0 can be specified to scan the entire blockchain.\n"
+            "                                                              0 can be specified to scan the entire blockchain. Blocks up to 2 hours before the earliest key\n"
+            "                                                              creation time of all keys being imported by the importmulti call will be scanned.\n"
             "      \"redeemscript\": \"<script>\"                            , (string, optional) Allowed only if the scriptPubKey is a P2SH address or a P2SH scriptPubKey\n"
             "      \"pubkeys\": [\"<pubKey>\", ... ]                         , (array, optional) Array of strings giving pubkeys that must occur in the output or redeemscript\n"
             "      \"keys\": [\"<key>\", ... ]                               , (array, optional) Array of strings giving private keys whose corresponding public keys must occur in the output or redeemscript\n"
@@ -1072,7 +1073,7 @@ UniValue importmulti(const JSONRPCRequest& mainRequest)
     }
 
     if (fRescan && fRunScan && requests.size() && nLowestTimestamp <= chainActive.Tip()->GetBlockTimeMax()) {
-        CBlockIndex* pindex = nLowestTimestamp > minimumTimestamp ? chainActive.FindEarliestAtLeast(nLowestTimestamp) : chainActive.Genesis();
+        CBlockIndex* pindex = nLowestTimestamp > minimumTimestamp ? chainActive.FindEarliestAtLeast(std::max<int64_t>(nLowestTimestamp - 7200, 0)) : chainActive.Genesis();
 
         if (pindex) {
             pwalletMain->ScanForWalletTransactions(pindex, true);
--- a/qa/rpc-tests/import-rescan.py
+++ b/qa/rpc-tests/import-rescan.py
@@ -2,6 +2,22 @@
 # Copyright (c) 2014-2016 The Bitcoin Core developers
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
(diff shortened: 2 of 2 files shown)


gmaxwell:
So one consequence of this is that if you use the manual pruning you will potentially prune past the time your imports will attempt to search, and then fail.

laanwj:
Concept ACK.

> So one consequence of this is that if you use the manual pruning you will potentially prune past the time your imports will attempt to search, and then fail.

Yes. Though I'm not sure that warrants any changes. After all it's the same for importwallet already.

At the least this grace period needs to be clearly documented in the RPC help so that it can be taken into account.

src/wallet/rpcdump.cpp
```diff
@@ -1072,7 +1072,7 @@ UniValue importmulti(const JSONRPCRequest& mainRequest)
     }
 
     if (fRescan && fRunScan && requests.size() && nLowestTimestamp <= chainActive.Tip()->GetBlockTimeMax()) {
-        CBlockIndex* pindex = nLowestTimestamp > minimumTimestamp ? chainActive.FindEarliestAtLeast(nLowestTimestamp) : chainActive.Genesis();
+        CBlockIndex* pindex = nLowestTimestamp > minimumTimestamp ? chainActive.FindEarliestAtLeast(nLowestTimestamp - 7200) : chainActive.Genesis();
```
morcos: I think this works fine, but does it seem slightly safer to not allow it to go negative?

ryanofsky:
>> So one consequence of this is that if you use the manual pruning you will potentially prune past the >> time your imports will attempt to search, and then fail.
> Yes. Though I'm not sure that warrants any changes. After all it's the same for importwallet already.

I would want to address this by only adding the grace period when `fPruneMode` is false. The way I see it, if you're using the rescan feature on a pruned node, you already have to be pretty careful, and while there are real improvements we can make for pruned nodes (like marking keys that could have incomplete transaction lists, or looking for spendable outputs in the utxo set in #9137), adding a grace period for importmulti rescans on pruned nodes is only a slightly useful thing to begin with, so it would be simplest for the code and for users to just keep the current behavior in this case.

> At the least this grace period needs to be clearly documented in the RPC help so that it can be taken > into account.

Will add a short note.

This change was discussed some on IRC:
https://botbot.me/freenode/bitcoin-core-dev/msg/80989053/
https://botbot.me/freenode/bitcoin-core-dev/msg/81013278/

ryanofsky:
Created #9773 to deal with the issue of pruned nodes. morcos argued against removing the grace period when fPruneMode was false because not having a grace period could potentially lead to missing transaction data, and gmaxwell argued against disabling importmulti rescans on pruned nodes because partial rescanning is an important feature for pruned nodes, so #9773 deals with the issue by returning errors from importmulti when rescans fail because blocks are missing or corrupt.

Still need to add a some documentation about the grace period, and add a min call to avoid passing negative values to FindEarliestAtLeast, but the bulk of the code in this PR (i.e. import-rescan.py) is ready for review.

laanwj:
Making the grace period depend on whether the node is pruning or not sounds like a terrible idea to me. It ties together two different options in a completely unexpected way, that's just asking for people to make mistakes.

gmaxwell:
The whole reason manual pruning takes a timestamp is so harmonize it with this, so you can not prune what you need to import. I think this could be addressed by also adding an identical grace to the manual pruning time.

ryanofsky (commented):
Added max call and updated RPC help text.

morcos:
utACK 		e662af3

Please tag 0.14
only reviewed code changes did not review test changes

src/wallet/rpcdump.cpp
```diff
@@ -1072,7 +1073,7 @@ UniValue importmulti(const JSONRPCRequest& mainRequest)
     }
 
     if (fRescan && fRunScan && requests.size() && nLowestTimestamp <= chainActive.Tip()->GetBlockTimeMax()) {
-        CBlockIndex* pindex = nLowestTimestamp > minimumTimestamp ? chainActive.FindEarliestAtLeast(nLowestTimestamp) : chainActive.Genesis();
+        CBlockIndex* pindex = nLowestTimestamp > minimumTimestamp ? chainActive.FindEarliestAtLeast(std::max<int64_t>(nLowestTimestamp - 7200, 0)) : chainActive.Genesis();
```
theuni: nit: I think this should be ```std::max<int64_t>(nLowestTimestamp - 7200, minimumTimestamp)```

I only mention because I'm unsure if 0 has a special meaning somewhere.

ryanofsky: 0 has special meaning for the CWallet::nTimeFirstKey variable, but I recently removed all outside references to that variable and made it private so that no other code that wasn't dealing with it directly would have to worry about special meanings of 0.

minimumTimestamp would work fine if it were set to 0 instead of 1, and I have some local changes to remove this variable, so I didn't want to add another use here.

theuni: Sounds good.

jtimon:
Concept ACK, although as commented in #9778, if these grace periods need to be consistent, I think we should have a constant (also used for normal rescan?).

laanwj:
> Concept ACK, although as commented in #9778, if these grace periods need to be consistent, I think we should have a constant (also used for normal rescan?).

Yes, would make sense to define a constant. No, that does not need to hold up merging this.

PR #8704, opened 2016-09-12
[RPC] Transaction details in getblock

achow101:
Adds an optional parameter extraVerbose to getblock to have transaction details displayed in a getblock call.

PR #9762, opened 2017-02-14
Add txdetails parameter to getblock.

pstratem:
Also add braces to if statements while we're here.


achow101:
Basically a duplicate of #8704

laanwj:
Huh. Yes, seemed like a deja-vu. You can make your changes after #8704 goes in, there's no reason to have two pulls open that effectively do the same.

PR #9764, opened 2017-02-15
wallet: Prevent "overrides a member function but is not marked 'override'" warnings

laanwj:
Because it is used inconsistently in a single class at least version 5.4.0 of g++ to complains about methods that don't use override. I get a ton of these warnings in everything that includes wallet.h. There is two ways to go about this: remove override from the methods having it, or add it to the methods missing it. I chose the second.
(a third would be to disable the warning completely or selectively, but I don't think that's called for)


paveljanik:
This removes approx. 2000 lines from my build logs! ;-)

ACK https://github.com/bitcoin/bitcoin/pull/9764/commits/6c5427d24ca18ce01e46242876e24f68a9c72cc3

PR #8068, opened 2016-05-18
Compact Blocks

TheBlueMatt:
This is based on #8020 and implements the BIP 152 draft from https://github.com/TheBlueMatt/bips/blob/152/bip-0152.mediawiki.

In short, it sends blocks as a set of short transaction IDs to allow nodes to not double-relay transactions in blocks. There are lots of details in both the BIP and the discussion on the bitcoin-dev ML.

There are still some TODOs in the code here, but I do not think they are blockers for this, as they are not protocol-level changes and can be implemented separately to make the protocol more effecient.

PR #7225, opened 2015-12-17
Eliminate unnecessary call to CheckBlock

sdaftuar:
`ProcessNewBlock` would return failure early if `CheckBlock` failed, before calling `AcceptBlock`.  `AcceptBlock` also calls `CheckBlock`, and upon failure would update `mapBlockIndex` to indicate that a block was failed.  By returning early in `ProcessNewBlock`, we were not marking blocks that fail a check in `CheckBlock` as permanently failed, and thus would continue to re-request and reprocess them.

This should result in one fewer call to `CheckBlock` for valid blocks, at the expense of one extra call to `AcceptBlockHeader` for blocks that fail `CheckBlock`, and it avoids reprocessing those failed blocks over and over.

PR #9765, opened 2017-02-15
Harden against mistakes handling invalid blocks

sdaftuar:
Fixes a bug in AcceptBlock() in invoking CheckBlock() with incorrect
arguments (introduced in #8068), and restores a call to CheckBlock() from ProcessNewBlock()
as belt-and-suspenders (effectively reverting #7225).

Updates the (overspecified) tests to match the slight behavior change (different reject code).

### DIFF
--- a/qa/rpc-tests/p2p-fullblocktest.py
+++ b/qa/rpc-tests/p2p-fullblocktest.py
@@ -398,7 +398,7 @@ class FullBlockTest(ComparisonTestFramework):
 
         # Extend the b26 chain to make sure bitcoind isn't accepting b26
         b27 = block(27, spend=out[7])
-        yield rejected(RejectResult(16, b'bad-prevblk'))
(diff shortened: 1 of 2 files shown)


morcos:
utACK ba803ef

morcos:
Also this is a bug fix that should be tagged 0.14 in my opinion

gmaxwell (commented):
utACK (will test)

theuni:
Grr, we really shouldn't have default arguments in consensus code.

utACK ba803efb687c2cb408176c546a544a8466c652ea.

laanwj:
utACK ba803ef

> Grr, we really shouldn't have default arguments in consensus code.

Bitfields/Enumerations instead of bool, bool, bool, ... also tends to help make code like this more readable. Multiple boolean parameters to a function make code hard to understand, always have to look at the signature to find out which one is what.

PR #9766, opened 2017-02-15
Add --exclude option to rpc-tests.py

jnewbery:
Sometimes it's useful to be able to run all test *except* some specified tests, either because those tests take a long time to execute or because you know they're broken.

This commit adds an additional `--exclude` or `-x` option to rpc-tests to exclude a comma-separated-list of tests. Usage is:

`rpc-tests.py --exclude=test1,test2,test3`
or
`rpc-tests.py --exclude test1,test2,test3`
or
`rpc-tests.py -x=test1,test2,test3`
or
`rpc-tests.py -x test1,test2,test3`

test names are given without the `.py` suffix.

This can be combined with the extended option to run the entire suite of tests, except for a single test, eg pruning.py which takes a long time:

`rpc-tests.py --extended -x pruning`

Note for reviewers: this requires and builds on top of https://github.com/bitcoin/bitcoin/pull/9657 . Please review only the last commit in this PR (921f6e1ac9b32cb80a83d97054b4cfdbbc290110)

[EDIT: edited commit hash]

### DIFF
--- a/qa/pull-tester/rpc-tests.py
+++ b/qa/pull-tester/rpc-tests.py
@@ -133,6 +133,7 @@ def main():
     Help text and arguments for individual test script:''',
                                      formatter_class=argparse.RawTextHelpFormatter)
     parser.add_argument('--coverage', action='store_true', help='generate a basic coverage report for the RPC interface')
+    parser.add_argument('--exclude', '-x', help='specify a comma-seperated-list of scripts to exclude. Do not include the .py extension in the name.')
     parser.add_argument('--extended', action='store_true', help='run the extended test suite in addition to the basic tests')
     parser.add_argument('--force', '-f', action='store_true', help='run tests even on platforms where they are disabled by default (e.g. windows).')
     parser.add_argument('--help', '-h', '-?', action='store_true', help='print help text and exit')
@@ -179,13 +180,6 @@ def main():
         # in the ALL_SCRIPTS list. Accept the name with or without .py extension.
         test_list = [t for t in ALL_SCRIPTS if
                 (t in tests or re.sub(".py$", "", t) in tests)]
-        if not test_list:
-            print("No valid test scripts specified. Check that your test is in one "
-                  "of the test lists in rpc-tests.py or run rpc-tests.py with no arguments to run all tests")
-            print("Scripts not found:")
-            print(tests)
-            sys.exit(0)
-
     else:
         # No individual tests have been specified. Run base tests, and
         # optionally ZMQ tests and extended tests.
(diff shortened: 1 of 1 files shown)


jnewbery:
@morcos

qa/pull-tester/rpc-tests.py
```diff
@@ -194,6 +195,12 @@ def main():
             # (for parallel running efficiency). This combined list will is no
             # longer sorted.
 
+    # Remove the test cases that the user has explicitly asked to exclude.
+    if args.exclude:
+        [ test_list.remove(exclude_test + ".py")
```
jtimon: why use [ ] here?

I think this would be more clear:
```
    if args.exclude:
        for exclude_test in args.exclude.split(','):
            if exclude_test + ".py" in test_list:
                test_list.remove(exclude_test + ".py")
```
And it's the same number of lines.

jtimon:
utACK f34e664 (although I left a style nit).
Yay, ```--extended --exclude=pruning```!!

MarcoFalke:
Concept ACK!

qa/pull-tester/rpc-tests.py
jnewbery: ha. If I write for loops I get review comments saying I should use list comprehensions, and if I write list comprehensions I get review comments saying I should use for loops :)

In this case you're probably right that the for loop is clearer.

jnewbery:
I've changed the list comprehension to a for loop as suggested by @jtimon . I've also moved the `if len(test_list) == 0` block lower down to catch the (unlikely) failure case where the user specifies and excludes the same test.

jnewbery:
rebased

qa/pull-tester/rpc-tests.py
```diff
@@ -198,6 +192,17 @@ def main():
             # (for parallel running efficiency). This combined list will is no
             # longer sorted.
 
+    # Remove the test cases that the user has explicitly asked to exclude.
+    if args.exclude:
+        for exclude_test in args.exclude.split(','):
+            if exclude_test + ".py" in test_list:
```
jtimon: Don't you need the following?
```
            if exclude_test in test_list:
                test_list.remove(exclude_test)
            if exclude_test + ".py" in test_list:
                test_list.remove(exclude_test + ".py")
```

Otherwise --exclude won't work if you specify the test with including .py explicitly.
Oh, sorry, the --exclude documentation explains it, never mind.

jtimon:
re utACK c578408

MarcoFalke:
utACK c5784085e9031ea1ad32a2aee897bd57a0482d24

PR #9768, opened 2017-02-15
[qa] Add logging to test_framework.py

jnewbery:
This is a WIP PR. Comments and feedback on the overall approach welcomed!

(For an overview of python's logging module and terms, see https://docs.python.org/3/library/logging.html and https://docs.python.org/3/howto/logging.html)

This PR adds logging to the test-framwork.py module. The "TestFramework" logger is added as a member of the BitcoinTestFramework class, and two handlers are attached to the logger:

- a StreamHandler() which outputs ERROR and higher level records to stderr.
- a FileHandler() which outputs all log records to a new `test_framework.log` file in the temporary test directory.

Test scripts and helper modules can log to the same Logger so there's a single log output:

- Individual test scripts can log to "TestFramework" by calling `self.log.[debug|info|warning|error|critical|exception]()`.
- Helper modules should log to child Loggers such as "TestFramework.mininode" or "TestFramework.authproxy", which will result in log records being propagated up to the "TestFramework" Logger.

I've added examples of how individual scripts and helper modules should log events in p2p-segwit.py and mininode.py.

The eventual goal is to remove all `print()` calls and replace them with logging.

A few notes on the implementation:

- a new `--loglevel` parameter can be used to log lower level records to the console, for example `--loglevel=DEBUG` will emit all log records to the console.
- the date format of `test_framework.log` is the same as bitcoind's debug.log. It should be trivial to interleave `test_framework.log` with the various `nodeX/debug.log` files for a global view of what's happening during the test.
- this should have no interaction with the existing `--tracerpc` parameter


laanwj:
If you're working on custom logging stuff anyway: what would be really nice is something that would queue up debug messages, then print them only if the test fails. This lowers the amount of spam in the Travis log in the good case, and gives full troubleshooting information in the bad case.

jnewbery:
@laanwj done: https://github.com/bitcoin/bitcoin/pull/9780

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -183,12 +179,41 @@ def main(self):
...
             sys.exit(0)
         else:
-            print("Failed")
+            self.log.warning("Failed")
             sys.exit(1)
 
+    def start_logging(self):
+        # Add logger and logging handlers
+        self.log = logging.getLogger('TestFramework')
+        self.log.setLevel(logging.DEBUG)
```
ryanofsky: I think you're better off dropping this call. The default debug level seems to be 0, and the value of DEBUG is 10 so this potentially throws away logs.

Also would drop the 3 other setLevel(logging.DEBUG) calls below.

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -183,12 +179,41 @@ def main(self):
...
+        # Create file handler to log all messages
+        fh = logging.FileHandler(self.options.tmpdir + '/test_framework.log')
+        fh.setLevel(logging.DEBUG)
+        # Create console handler to log messages to stderr. By default this logs only error messages, but can be configured with --loglevel.
+        ch = logging.StreamHandler()
+        console_log_level = getattr(logging, self.options.loglevel.upper(), None)
```
ryanofsky: Calling getattr doesn't seem to be necessary, the documentation says setLevel already accepts a string: 
https://docs.python.org/3/library/logging.html#logging.Handler.setLevel. Maybe upper() is still useful though.

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -183,12 +179,41 @@ def main(self):
...
+        self.log.addHandler(fh)
+        self.log.addHandler(ch)
+
+        if self.options.trace_rpc:
+            rpc_logger = logging.getLogger("BitcoinRPC")
+            rpc_logger.setLevel(logging.DEBUG)
+            rpc_handler = logging.StreamHandler()
+            rpc_handler.setLevel(logging.DEBUG)
+            rpc_logger.addHandler(rpc_handler)
+            rpc_logger.debug("debug")
```
ryanofsky: Junk lines?

ryanofsky:
3e7ed8c21c3f52a312a538746aebbbb06f67e3a1

jnewbery:
@ryanofsky thanks for the great feedback. I'm going to hold back on making any changes until I get some more concept ACKs/NACKs. I want to make sure people are happy going in this direction before I spend any time addressing review comments.

jnewbery:
@MarcoFalke @laanwj @jonasschnelli @sdaftuar you're the most active contributors to the qa directory. Can you have a quick look and give me a concept ACK/NACK so I know whether to continue in this direction.

No rush. Obviously you're all busy with the 0.14 release.

sdaftuar:
Concept ACK.  

I haven't reviewed this PR in detail yet, but what I'd find useful is if the default behavior when running a single test locally was to print all log messages to the console, and only suppress log messages when running the tests through the `rpc-tests.py` script.

jnewbery:
@sdaftuar the eventual aim will be to add a lot more low-level logging (eg logging when every p2p message is sent or received on the mininode connection). Those low-level logs will always be saved to test_framework.log, but by default are not printed to the console. If the user want low-level logs printed to console, they can specify --loglevel when running the script.

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -113,6 +113,8 @@ def main(self):
                           help="Directory for caching pregenerated datadirs")
         parser.add_option("--tmpdir", dest="tmpdir", default=tempfile.mkdtemp(prefix="test"),
                           help="Root directory for datadirs")
+        parser.add_option("--loglevel", dest="loglevel", default="ERROR",
```
jnewbery: Thanks. Updated.

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -113,6 +113,8 @@ def main(self):
...
+                          help="log events at this level and higher to the console")
```
jnewbery: Updated

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -140,38 +139,35 @@ def main(self):
...
             os.makedirs(self.options.tmpdir, exist_ok=False)
+            self.start_logging()
             self.setup_chain()
             self.setup_network()
             self.run_test()
             success = True
         except JSONRPCException as e:
-            print("JSONRPC error: "+e.error['message'])
-            traceback.print_tb(sys.exc_info()[2])
+            self.log.exception("JSONRPC error: "+e.error['message'])
```
jnewbery: Yep, good catch.

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -140,38 +139,35 @@ def main(self):
...
 
         if not self.options.nocleanup and not self.options.noshutdown and success:
-            print("Cleaning up")
+            self.log.info("Cleaning up")
             shutil.rmtree(self.options.tmpdir)
             if not os.listdir(self.options.root):
                 os.rmdir(self.options.root)
         else:
-            print("Not cleaning up dir %s" % self.options.tmpdir)
+            self.log.info("Not cleaning up dir %s" % self.options.tmpdir)
```
jnewbery: good point. Fixed.

qa/rpc-tests/test_framework/test_framework.py
jnewbery: I think DEBUG is good here. We shouldn't encourage people to log messages at custom levels (ie not one of debug, info, warning, error and critical), and we especially shouldn't encourage people to log at a level lower than DEBUG.

qa/rpc-tests/test_framework/test_framework.py
jnewbery: Agree. I was following the Python 'logging how to' guide here: https://docs.python.org/3/howto/logging.html#logging-to-a-file which uses getattr(). I guess that was written before 3.2, which is when setlevel() was updated to take a string argument.

I've tested the logging module and:
- we do need to use upper() to set the log levels to uppercase
- calling with an invalid string causes logging to raise a helpful exception (I was worried that it might fail silently).

So calling setlevel directly with self.options.loglevel.upper() is better.

qa/rpc-tests/test_framework/test_framework.py
jnewbery: yes. Junk lines from when I was testing this. Thanks. Removed.

jnewbery:
I've addressed @ryanofsky's review comments and repushed. A few updates since I opened this PR:

- microsecond debug logging is now set by default on all test nodes. I've removed all places in the test code where debugging is set manually, since there's no reason not to use debug logging in these test cases.
- I've added logging to the test_framework submodules
- I've replaced print functions in the individual test-cases with logging calls

I've been running with this patch set locally for some time, together with a tool to interleave test_framwork logs with individual node logs. It makes troubleshooting testcases a lot easier since I have a global view of everything that's happening in the test case (ie I can see messages being sent in by the test framework and then received by the node).

I believe this is ready for merge, but any further review comments are welcomed.

qa/rpc-tests/test_framework/test_framework.py
ryanofsky: > I think DEBUG is good here. We shouldn't encourage people to log messages at custom levels (ie not one of debug, info, warning, error and critical), and we especially shouldn't encourage people to log at a level lower than DEBUG.

Actually setting DEBUG seems to be needed here in order to be able to see debug messages at all, otherwise python seems to fall back to root logger level which is WARNING.

Still I would consider dropping the "fh.setLevel()" call below. Having that call seems less like "not encouraging" people to log messages at custom levels, and more like going out of our way to punish and confuse them.

ryanofsky (commented):
ACK 3b336aba4ef250868d6f3dc2bf741b0bbf621d39 assuming travis tests will be fixed.

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -112,6 +112,8 @@ def main(self):
                           help="Directory for caching pregenerated datadirs")
         parser.add_option("--tmpdir", dest="tmpdir", default=tempfile.mkdtemp(prefix="test"),
                           help="Root directory for datadirs")
+        parser.add_option("--loglevel", dest="loglevel", default="ERROR",
```
jnewbery: You're right. This should be INFO by default. DEBUG emits way too much to the console to be the default.

pull-tester diverts stdout from the subtests, so there's no output to console at all.

--loglevel now accepts ints as well as the named strings and can be specified with the short option -l

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -139,38 +138,35 @@ def main(self):
         success = False
         try:
             os.makedirs(self.options.tmpdir, exist_ok=False)
```
jnewbery: Good idea. I needed to put start_logging() below the os.makedirs() call, but you're right that this should be outside the large try/except block. I've now moved them both up.

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -182,12 +178,36 @@ def main(self):
                     from collections import deque
                     print("".join(deque(open(f), MAX_LINES_TO_PRINT)))
         if success:
-            print("Tests successful")
+            self.log.info("Tests successful")
             sys.exit(0)
         else:
-            print("Failed")
+            self.log.warning("Failed")
```
jnewbery: Agree. Done both.

qa/rpc-tests/test_framework/test_framework.py
jnewbery: I still don't agree. logging at non-standard levels is harder (you need to call `self.log.log(lvl, message`) rather than `self.log.debug(message)` or similar). Why would anyone want do that, and why would we want to allow people to add confusing logging levels to the test cases?

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -182,12 +178,36 @@ def main(self):
...
+    def start_logging(self):
+        # Add logger and logging handlers
+        self.log = logging.getLogger('TestFramework')
+        self.log.setLevel(logging.DEBUG)
+        # Create file handler to log all messages
+        fh = logging.FileHandler(self.options.tmpdir + '/test_framework.log')
+        fh.setLevel(logging.DEBUG)
+        # Create console handler to log messages to stderr. By default this logs only error messages, but can be configured with --loglevel.
+        ch = logging.StreamHandler()
+        ch.setLevel(self.options.loglevel.upper())
```
jnewbery: Yep. Good idea. Done (although I've split this over two lines for readability).

qa/rpc-tests/bip68-sequence.py
```diff
@@ -34,26 +34,26 @@ def run_test(self):
...
-        print("Running test BIP68 not consensus before versionbits activation")
+        self.log.info("Running test BIP68 not consensus before versionbits activation")
         self.test_bip68_not_consensus()
 
-        print("Activating BIP68 (and 112/113)")
+        self.log.info("Activating BIP68 (and 112/113)")
         self.activateCSV()
 
-        print("Verifying nVersion=2 transactions are standard.")
-        print("Note that with current versions of bitcoin software, nVersion=2 transactions are always standard (independent of BIP68 activation status).")
```
jnewbery: Because I previously thought that it was a bug that Core v0.13 (and a few earlier versions) always relayed v2 transactions - hence the 'with current versions' comment. See #9739 for the discussion. Whether or not it's a bug is moot now, and I think the 'with current verions' comment doesn't clarify.

'always' here means 'both before and after BIP 68 activation'.

qa/rpc-tests/bip68-sequence.py
```diff
@@ -34,26 +34,26 @@ def run_test(self):
...
+        self.log.info("Verifying nVersion=2 transactions are standard.")
+        self.log.info("Note that nVersion=2 transactions are always standard (independent of BIP68 activation status).")
         self.test_version2_relay()
 
-        print("Passed\n")
+        self.log.info("Passed\n")
```
jnewbery: Thanks. Done.

qa/rpc-tests/test_framework/comptool.py
```diff
@@ -21,6 +21,10 @@
 from .blockstore import BlockStore, TxStore
 from .util import p2p_port
 
+import logging
+
+logger=logging("TestFramework.comptool")
```
jnewbery: Yep. My mistake. Fixed.

jnewbery:
@ryanofsky's comments addressed and repushed: c9755da51a742c0d86572f750e16573e87fdd166 / https://github.com/jnewbery/bitcoin/tree/pr9768.3

jnewbery:
rebased: https://github.com/bitcoin/bitcoin/pull/9768/commits/2c9fc0e5f786dd4daa906f20fc322a615bc5b6e2 / https://github.com/jnewbery/bitcoin/tree/pr9768.4

jnewbery:
@laanwj @MarcoFalke I think this is a real improvement and helps debugging and fixing test cases when they fail. I've addressed https://github.com/bitcoin/bitcoin/pull/9768#issuecomment-280268516 in #9780 which builds off this PR. Any chance I can convince you to take a look at this?

qa/rpc-tests/proxy_test.py
```diff
@@ -83,13 +83,13 @@ def setup_nodes(self):
         # Note: proxies are not used to connect to local nodes
         # this is because the proxy to use is based on CService.GetNetwork(), which return NET_UNROUTABLE for localhost
         args = [
-            ['-listen', '-debug=net', '-debug=proxy', '-proxy=%s:%i' % (self.conf1.addr),'-proxyrandomize=1'],
```
ryanofsky: Line 72 `Warning: testing without local IPv6 support` is now a print again, not sure if this was intentional.

ryanofsky (commented):
utACK f546d06ac21e2953c0dc0b1d8fef459b186781ce

Thanks for supporting `-l0`.

> I think this is a real improvement and helps debugging and fixing test cases when they fail.

Definitely agree. It should be especially useful to easily see interleaved python and bitcoin debug.log output while debugging.

qa/rpc-tests/proxy_test.py
jnewbery: yes - proxy_test does a lot of work in the `__init__()` method, which is called before the temp directory is set up. Logging requires the temp directory, so we can't use `self.log` until we're in `main()` and `_start_logging()` has been called.

I left this as a `print()` rather than trying to do anything clever to coerce the test into being log friendly.

MarcoFalke:
Concept ACK. Going to give some feedback later tonight.

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -182,12 +181,38 @@ def main(self):
                     from collections import deque
                     print("".join(deque(open(f), MAX_LINES_TO_PRINT)))
         if success:
-            print("Tests successful")
+            self.log.info("Tests successful")
             sys.exit(0)
         else:
-            print("Failed")
+            self.log.error("Test failed. Test logging available at %s/test_framework.log", self.options.tmpdir)
             sys.exit(1)
```
MarcoFalke: You might need to shutdown (flush) the logs here.

qa/rpc-tests/p2p-segwit.py
```diff
@@ -908,7 +907,7 @@ def test_witness_tx_relay_before_segwit_activation(self):
         # a witness transaction ought not result in a getdata.
         try:
             self.test_node.announce_tx_and_wait_for_getdata(tx, timeout=2)
-            print("Error: duplicate tx getdata!")
+            self.log.error("Error: duplicate tx getdata!")
             assert(False)
```
MarcoFalke: You don't need the assert(False) when you pass any AssertionError below. Though maybe a fix for an unrelated follow up pr.

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -182,12 +181,38 @@ def main(self):
...
+        fh.setLevel(logging.DEBUG)
+        # Create console handler to log messages to stderr. By default this logs only error messages, but can be configured with --loglevel.
+        ch = logging.StreamHandler(sys.stdout)
+        # User can provide log level as a number or string (eg DEBUG). loglevel was caught as a string, so try to convert it to an int
+        ll = int(self.options.loglevel) if self.options.loglevel.isdigit() else self.options.loglevel.upper()
+        ch.setLevel(ll)
+        # Format logs the same as bitcoind's debug.log with microprecision (so log files can be concatenated and sorted)
+        formatter = logging.Formatter(fmt = '%(asctime)s.%(msecs)03d000 %(name)s (%(levelname)s): %(message)s', datefmt='%Y-%m-%d %H:%M:%S')
+        fh.setFormatter(formatter)
+        ch.setFormatter(formatter)
```
MarcoFalke: I think the print to stdout can use a formatter that produces smaller lines. (Maybe without the timestamp?) But no strong  opinion.

maflcko (commented):
utACK f546d06

qa/rpc-tests/invalidateblock.py
```diff
@@ -18,51 +18,49 @@ def __init__(self):
...
-        print("..and then mine a block")
+        self.log.info("..and then mine a block")
         self.nodes[2].generate(1)
-        print("Verify all nodes are at the right height")
+        self.log.info("Verify all nodes are at the right height")
         time.sleep(5)
-        for i in range(3):
-            print(i,self.nodes[i].getblockcount())
         assert(self.nodes[2].getblockcount() == 3)
         assert(self.nodes[0].getblockcount() == 4)
```
jnewbery: Done.

qa/rpc-tests/test_framework/mininode.py
```diff
@@ -1742,17 +1737,14 @@ def got_data(self):
...
-            print('got_data:', repr(e))
-            # import  traceback
-            # traceback.print_tb(sys.exc_info()[2])
+            logger.exception('got_data:', repr(e))
 
     def send_message(self, message, pushbuf=False):
         if self.state != "connected" and not pushbuf:
             raise IOError('Not connected, no pushbuf')
-        self.show_debug_msg("Send %s" % repr(message))
+            logger.debug("Send message to %s:%d: %s" % (self.dstaddr, self.dstport, repr(message)))
```
jnewbery: Oops. Fixed.

qa/rpc-tests/test_framework/test_framework.py
jnewbery: Done.

qa/rpc-tests/test_framework/test_framework.py
jnewbery: I specifically set the format of the logs to be the same as what bitcoind produces with microsecond logging.

qa/rpc-tests/p2p-segwit.py
jnewbery: That's pre-existing behaviour, which is unrelated to this PR. I agree it should be fixed in a different PR.

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -182,12 +181,38 @@ def main(self):
...
+        # add the handlers to the logger
+        self.log.addHandler(fh)
+        self.log.addHandler(ch)
+
+        if self.options.trace_rpc:
+            rpc_logger = logging.getLogger("BitcoinRPC")
+            rpc_logger.setLevel(logging.DEBUG)
+            rpc_handler = logging.StreamHandler()
```
jnewbery: No reason. Just an oversight. Now fixed.

jnewbery:
Thanks for the feedback @MarcoFalke . I believe the latest commit should address all your comments.

MarcoFalke:
Ok, will take a quick look later and plan to merge this before the meeting.

Sjors:
This works if you call `test/functional/bumpfee.py --loglevel=INFO`, but not if you call `test/functional/test_runner.py --loglevel=INFO bumpfee` (which just shows dots). Not sure if that's intended, but it did [confuse me](https://botbot.me/freenode/bitcoin-core-dev/2017-11-27/?msg=93962703&page=2).

jnewbery:
documentation is here: https://github.com/bitcoin/bitcoin/blob/master/test/README.md#test-logging :

> when run through the test_runner harness, all logs are written to test_framework.log and no logs are output to the console.

Feel free to open a PR if you think the documentation could be improved.

Sjors:
@jnewberry it's probably sufficient if `/test_runner.py --help` says that next to the `--loglevel` flag (or if it doesn't show that option).

jnewbery:
The help text for `--loglevel` comes from `test_framework.py`. `test_runner.py` grabs that help text and appends it to its own help text. I suppose that test_runner could remove the `--loglevel` help before displaying it. The relevant code is here: https://github.com/bitcoin/bitcoin/blob/a89221873a3ee2451c73b41bbe2d99d36f439d31/test/functional/test_runner.py#L258

Sjors:
Thanks, I'll take a look.

PR #9771, opened 2017-02-15
Add missing cs_wallet lock that triggers new lock held assertion

ryanofsky:
A new AssertLockHeld(cs_wallet) call was added in commit a58370e
"Dedup nTimeFirstKey update logic" (part of PR #9108).

The lock held assertion will fail when loading prexisting wallets files from
before the #9108 merge that have watch-only keys.

### DIFF
--- a/src/wallet/walletdb.cpp
+++ b/src/wallet/walletdb.cpp
@@ -559,8 +559,8 @@ DBErrors CWalletDB::LoadWallet(CWallet* pwallet)
     bool fNoncriticalErrors = false;
     DBErrors result = DB_LOAD_OK;
 
+    LOCK(pwallet->cs_wallet);
     try {
-        LOCK(pwallet->cs_wallet);
         int nMinVersion = 0;
         if (Read((string)"minversion", nMinVersion))
         {


ryanofsky:
This bug was found by @morcos. It should be tagged for 0.14 because it's a regression caused by #9108.

morcos:
ACK 0f25026

seems like it might be better to just lock cs_wallet for the whole function call, but i don't know if there is some reason it wasn't written that way.

ryanofsky:
I don't see any reason, updated the PR. 0f25026b553771817928ceb96d33d1bab942e18c -> 07afcd6379bb46ace5856f6a47a9188cf9aed2ea ([loadlock.0](https://github.com/ryanofsky/bitcoin/commits/pr/loadlock.0) -> [loadlock.1](https://github.com/ryanofsky/bitcoin/commits/pr/loadlock.1))

laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/9771/commits/07afcd6379bb46ace5856f6a47a9188cf9aed2ea

Issue #9423, opened 2016-12-25
"tinyformat: Too many conversion specifiers in format string"


rebroad:
I don't think I've seen an issue raised for this - it's more of a development issue than a runtime issue but if encounter it often, which is that if there is an incorrect number of arguments in a LogPrint() or LogPrintf() then no compile error/warning is given, it will cause the code to SIGABRT, and a gdb backtrace never provides much useful information about where the error in the code is. Only the name of the thread is given which can often leave a lot of searching or git bisecting. This common problem could be made easier to debug...

gmaxwell:
It's basically impossible to give a compile error for this unfortunately. It's only possible for printf because the compiler basically implements the whole function internally.

ryanofsky:
Could \_\_attribute__ format work?

http://clang.llvm.org/docs/AttributeReference.html#format-gnu-format
https://gcc.gnu.org/onlinedocs/gcc/Common-Function-Attributes.html

laanwj:
>  it will cause the code to SIGABRT, and a gdb backtrace never provides much useful information about where the error in the code is

That's not entirely correct. It is set up to raise an exception.

```
 #define TINYFORMAT_ERROR(reasonString) throw std::runtime_error(reasonString)
```

In the case of logging it would be OK to just log a warning message instead of raising an exception.
The various Log* could catch this exception and do that.

`strprintf` OTOH should not, because it is also used in more critical places than logging (such as formatting RPC output), and a programmer error there to specify not enough conversion specifiers is a critical problem.

laanwj:
An external script would also be able to check this, without too much trouble as the format strings must be static.

In the long run I'd prefer to abandon strprintf-ish syntax and use, at least for logging, the more C++ish way of
```
Logger(Error,"rpc") << "HTTP: starting" << rpcThreads << "worker threads";
```
This is what tinyformat does internally. It is a type-safe compatibility layer from printf-like vararg syntax to c++ formatting.

Using tinyformat was a workaround to make it possible to keep most of the logging lines the same while still avoiding libc printf's pitfalls such as variability between platforms and type-safety issues. Lack of compile-time checking of number of arguments is annoying though.

fanquake:
Fixed in #9423.

PR #9773, opened 2017-02-15
Return errors from importmulti if complete rescans are not successful

ryanofsky:
This is supposed to address the concern from @gmaxwell in #9761 that importmulti rescans on pruned nodes could unsuccessfully scan pruned blocks and fail to inform the user.

An alternative approach of just simply disabling importmulti rescans on pruned nodes was initially discussed in https://botbot.me/freenode/bitcoin-core-dev/msg/81023921/, but running being able to import keys and do partial rescans on pruned nodes was determined to be an important use-case.

~~~Posting this PR for feedback. The code has not been tested at all, and I'm actually not sure how I'm going to go about writing a unit test because there doesn't seem to be an easy way to target individual blocks for pruning inside RPC tests.~~~

### DIFF
--- a/src/wallet/rpcdump.cpp
+++ b/src/wallet/rpcdump.cpp
@@ -1074,10 +1074,31 @@ UniValue importmulti(const JSONRPCRequest& mainRequest)
 
     if (fRescan && fRunScan && requests.size() && nLowestTimestamp <= chainActive.Tip()->GetBlockTimeMax()) {
         CBlockIndex* pindex = nLowestTimestamp > minimumTimestamp ? chainActive.FindEarliestAtLeast(std::max<int64_t>(nLowestTimestamp - 7200, 0)) : chainActive.Genesis();
-
+        CBlockIndex* scannedRange = nullptr;
         if (pindex) {
-            pwalletMain->ScanForWalletTransactions(pindex, true);
+            scannedRange = pwalletMain->ScanForWalletTransactions(pindex, true);
             pwalletMain->ReacceptWalletTransactions();
+        }
+
+        if (!scannedRange || scannedRange->nHeight > pindex->nHeight) {
+            std::vector<UniValue> results = response.getValues();
+            response.clear();
+            response.setArray();
+            size_t i = 0;
+            for (const UniValue& request : requests.getValues()) {
+                // If key creation date is within the successfully scanned
+                // range, or if the import result already has an error set, let
+                // the result stand unmodified. Otherwise replace the result
+                // with an error message.
+                if (GetImportTimestamp(request, now) - 7200 >= scannedRange->GetBlockTimeMax() || results.at(i).exists("error")) {
+                    response.push_back(results.at(i));
+                } else {
+                    UniValue result = UniValue(UniValue::VOBJ);
+                    result.pushKV("success", UniValue(false));
+                    result.pushKV("error", JSONRPCError(RPC_MISC_ERROR, strprintf("Failed to rescan before time %d, transactions may be missing.", scannedRange->GetBlockTimeMax())));
+                    response.push_back(std::move(result));
+                }
+            }
         }
     }
 
--- a/src/wallet/test/wallet_tests.cpp
+++ b/src/wallet/test/wallet_tests.cpp
@@ -9,10 +9,16 @@
 #include <utility>
 #include <vector>
 
+#include "rpc/server.h"
+#include "test/test_bitcoin.h"
+#include "validation.h"
 #include "wallet/test/wallet_test_fixture.h"
 
 #include <boost/foreach.hpp>
 #include <boost/test/unit_test.hpp>
+#include <univalue.h>
+
+extern UniValue importmulti(const JSONRPCRequest& request);
 
 // how many times to run all the tests to have a chance to catch errors that only show up with particular random shuffles
 #define RUN_TESTS 100
@@ -353,6 +359,60 @@ BOOST_AUTO_TEST_CASE(ApproximateBestSubset)
     BOOST_CHECK_EQUAL(setCoinsRet.size(), 2U);
 
     empty_wallet();
+}
+
+BOOST_FIXTURE_TEST_CASE(rescan, TestChain100Setup)
+{
+    LOCK(cs_main);
+
+    // Cap last block file size, and mine new block in a new block file.
+    CBlockIndex* oldTip = chainActive.Tip();
+    GetBlockFileInfo(oldTip->GetBlockPos().nFile)->nSize = MAX_BLOCKFILE_SIZE;
+    CreateAndProcessBlock({}, GetScriptForRawPubKey(coinbaseKey.GetPubKey()));
+    CBlockIndex* newTip = chainActive.Tip();
+
+    // Verify ScanForWalletTransactions picks up transactions in both the old
+    // and new block files.
+    {
+        CWallet wallet;
+        LOCK(wallet.cs_wallet);
+        wallet.AddKeyPubKey(coinbaseKey, coinbaseKey.GetPubKey());
+        BOOST_CHECK_EQUAL(oldTip, wallet.ScanForWalletTransactions(oldTip));
+        BOOST_CHECK_EQUAL(wallet.GetImmatureBalance(), 100 * COIN);
+    }
+
+    // Prune the older block file.
+    PruneOneBlockFile(oldTip->GetBlockPos().nFile);
+    UnlinkPrunedFiles({oldTip->GetBlockPos().nFile});
+
+    // Verify ScanForWalletTransactions only picks transactions in the new block
+    // file.
+    {
+        CWallet wallet;
+        LOCK(wallet.cs_wallet);
+        wallet.AddKeyPubKey(coinbaseKey, coinbaseKey.GetPubKey());
+        BOOST_CHECK_EQUAL(newTip, wallet.ScanForWalletTransactions(oldTip));
+        BOOST_CHECK_EQUAL(wallet.GetImmatureBalance(), 50 * COIN);
+    }
+
+    {
+        CWallet wallet;
+        ::pwalletMain = &wallet;
+        UniValue key;
+        key.setObject();
+        key.pushKV("scriptPubKey", HexStr(GetScriptForRawPubKey(coinbaseKey.GetPubKey())));
+        key.pushKV("timestamp", 0);
+        key.pushKV("internal", UniValue(true));
+        UniValue keys;
+        keys.setArray();
+        keys.push_back(key);
+        JSONRPCRequest request;
+        request.params.setArray();
+        request.params.push_back(keys);
+
+        UniValue response = importmulti(request);
+        BOOST_CHECK_EQUAL(response.write(), strprintf("[{\"success\":false,\"error\":{\"code\":-1,\"message\":\"Failed to rescan before time %d, transactions may be missing.\"}}]", newTip->GetBlockTimeMax()));
+    }
 }
 
 BOOST_AUTO_TEST_SUITE_END()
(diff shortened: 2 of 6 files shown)


morcos:
Concept ACK
This looks like a reasonable approach to me.  

Don't forget that you may want to include the grace period in detecting whether it scanned enough.
Also you might want to skip the rest of the loop if ReadBlockFromDisk is false.

laanwj:
Concept ACK.

> The code has not been tested at all, and I'm actually not sure how I'm going to go about writing a unit test because there doesn't seem to be an easy way to target individual blocks for pruning inside RPC tests.

There's `pruneblockchain` but yes pruning works on a per-file, not a per-block basis, so it would need some extra effort on regtest to generate full blocks instead of empty ones otherwise at block 30000 you're still in the first file, and that RPC will do nothing.

gmaxwell:
Concept ACK. Will test/review further.

morcos:
Please rebase on #9761 so we can verify the grace period works properly

ryanofsky:
Rebased

src/wallet/wallet.cpp
```diff
@@ -1570,12 +1574,15 @@ int CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, bool fUpdate)
...
+                ret = nullptr;
+            } else if (!ret) {
+                ret = pindex;
+            }
             int posInBlock;
             for (posInBlock = 0; posInBlock < (int)block.vtx.size(); posInBlock++)
             {
-                if (AddToWalletIfInvolvingMe(*block.vtx[posInBlock], pindex, posInBlock, fUpdate))
-                    ret++;
+                AddToWalletIfInvolvingMe(*block.vtx[posInBlock], pindex, posInBlock, fUpdate);
```
morcos: I think this loop should also be skipped if ReadBlockFromDisk fails

ryanofsky: Skipped in fixmed6f6a23c7092aa21110cc77ebeb018492bacae28

ryanofsky (commented):
Add test for CWallet::ScanForWalletTransactions in e3d4f257ce0202811464d3b6860db87ba147aa24

src/wallet/rpcdump.cpp
```diff
@@ -1072,12 +1073,32 @@ UniValue importmulti(const JSONRPCRequest& mainRequest)
...
+                // If key creation date is within the successfully scanned
+                // range, or if the import result already has an error set, let
+                // the result stand unmodified. Otherwise replace the result
+                // with an error message.
+                if (GetImportTimestamp(request, now) - 7200 >= scannedRange->GetBlockTimeMax() || results.at(i).exists("error")) {
+                    response.push_back(results.at(i));
+                } else {
+                    UniValue result = UniValue(UniValue::VOBJ);
+                    result.pushKV("success", UniValue(false));
+                    result.pushKV("error", JSONRPCError(RPC_MISC_ERROR, strprintf("Failed to rescan before time %" PRId64 ", transactions may be missing.", (long long)scannedRange->GetBlockTimeMax())));
```
jonasschnelli: just curios: Is there an advantage using `PRId64` over `%ul` (rest of the code uses this)?

ryanofsky: Actually I need to change this to either remove PRId64 or the (long long) cast.

You're supposed to use PRId64 whenever you pass an int64_t type to printf. %ld or %lld aren't always correct because longs and long longs just have minimum widths in standard c++, not fixed widths that are the same on all systems.

ryanofsky:
Previously squashed e3d4f257ce0202811464d3b6860db87ba147aa24 -> 455b8f7cb2a1889053a006055e9dcad923c77029 ([multicheck.2](https://github.com/ryanofsky/bitcoin/commits/pr/multicheck.2) -> [multicheck.3](https://github.com/ryanofsky/bitcoin/commits/pr/multicheck.3))

src/wallet/rpcdump.cpp
ryanofsky: Removed cast in e49f0394f07a95e6cc5cdabe32f2c6b8f53c40fb.

ryanofsky:
Added C++ test for new importmulti RPC code in e64cdfba4a7e8a670e76c7bf770c248dc6b78c3d

Squashed e64cdfba4a7e8a670e76c7bf770c248dc6b78c3d -> b73252de549984a06aae294252d0d055aca43a4a ([multicheck.4](https://github.com/ryanofsky/bitcoin/commits/pr/multicheck.4) -> [multicheck.5](https://github.com/ryanofsky/bitcoin/commits/pr/multicheck.5))

sipa:
Note that our strprintf is a wrapper around tfm::format, and doesn't need type specs at all.

morcos:
lightly tested ACK b73252d
Only reviewed/tested code, not the tests.

jonasschnelli (approved):
utACK b73252de549984a06aae294252d0d055aca43a4a

src/wallet/rpcdump.cpp
```diff
@@ -1072,12 +1073,33 @@ UniValue importmulti(const JSONRPCRequest& mainRequest)
...
+                // If key creation date is within the successfully scanned
+                // range, or if the import result already has an error set, let
+                // the result stand unmodified. Otherwise replace the result
+                // with an error message.
+                if (GetImportTimestamp(request, now) - 7200 >= scannedRange->GetBlockTimeMax() || results.at(i).exists("error")) {
+                    response.push_back(results.at(i));
+                } else {
+                    UniValue result = UniValue(UniValue::VOBJ);
+                    result.pushKV("success", UniValue(false));
+                    result.pushKV("error", JSONRPCError(RPC_MISC_ERROR, strprintf("Failed to rescan before time %" PRId64 ", transactions may be missing.", scannedRange->GetBlockTimeMax())));
```
laanwj: Please remove the PRI64d here. It gives problems with some platforms, and as @sipa ways don't need it.

ryanofsky: Ok will change, but what are the problems with some platforms?

ryanofsky: Squashed b73252de549984a06aae294252d0d055aca43a4a -> 90671ee7e77c17c18f5ea82ea041962ef5d53f8a ([multicheck.5](https://github.com/ryanofsky/bitcoin/commits/pr/multicheck.5) -> [multicheck.6](https://github.com/ryanofsky/bitcoin/commits/pr/multicheck.6))

ryanofsky:
Rebased b73252de549984a06aae294252d0d055aca43a4a -> 90671ee7e77c17c18f5ea82ea041962ef5d53f8a ([multicheck.5](https://github.com/ryanofsky/bitcoin/commits/pr/multicheck.5) -> [multicheck.6](https://github.com/ryanofsky/bitcoin/commits/pr/multicheck.6))

src/wallet/rpcdump.cpp
ryanofsky: Removed in aaaf65a90b87c18889d177ff3f0ab8a868f849f3.

Reason I am asking about the problem you're seeing is that not using PRI64d for int64_t would prevent us from using compile time format string checking (as suggested in https://github.com/bitcoin/bitcoin/issues/9423#issuecomment-269480805) which would be much better than the runtime checking we currently use.

ryanofsky:
Squashed aaaf65a90b87c18889d177ff3f0ab8a868f849f3 -> 1425c727863b335bf98bb4b17b575e59552d38e9 ([multicheck.7](https://github.com/ryanofsky/bitcoin/commits/pr/multicheck.7) -> [multicheck.8](https://github.com/ryanofsky/bitcoin/commits/pr/multicheck.8))

MarcoFalke:
Somehow travis didn't like your commit.

ryanofsky:
> Somehow travis didn't like your commit.

Appears to be a spurious error. The travis run for b72521dd7f570b0c557439d534f7523a5be52999 which is aaaf65a90b87c18889d177ff3f0ab8a868f849f3 merged into 9828f9a9962c1bee5c343847030b9cfd87a40a5e succeeded: https://travis-ci.org/bitcoin/bitcoin/builds/202600321

Only the travis run for 64072b74ebbf7b5585b39f6c117c6ffa9dc820f0 which is 1425c727863b335bf98bb4b17b575e59552d38e9 merged into 9828f9a9962c1bee5c343847030b9cfd87a40a5e failed: https://travis-ci.org/bitcoin/bitcoin/builds/202600549

There are no differences between the two commits:

```
$ git rev-parse aaaf65a90b87c18889d177ff3f0ab8a868f849f3^{tree} 1425c727863b335bf98bb4b17b575e59552d38e9^{tree}
9e695fd32f65b1596f3ff613b96ec756bb2f1d2b
9e695fd32f65b1596f3ff613b96ec756bb2f1d2b
```

ryanofsky:
Touched CommitDate to rerun travis 1425c727863b335bf98bb4b17b575e59552d38e9 -> 68da3ee4284a07f695e992afd6cdd545604aac90 ([multicheck.8](https://github.com/ryanofsky/bitcoin/commits/pr/multicheck.8) -> [multicheck.9](https://github.com/ryanofsky/bitcoin/commits/pr/multicheck.9))

ryanofsky:
Travis run for 91b1ec2f4f27a817aee15f573afc26b7d17c8d7b which is 154025927dd5929f69f79e32818932d387c004bf merged into 9828f9a9962c1bee5c343847030b9cfd87a40a5e succeeded: https://travis-ci.org/bitcoin/bitcoin/builds/202655068

Travis run for 527205931a636f26d620af25db61dbe9c716fb38 which is 68da3ee4284a07f695e992afd6cdd545604aac90 merged into 9828f9a9962c1bee5c343847030b9cfd87a40a5e failed: https://travis-ci.org/bitcoin/bitcoin/builds/202665925

Touched CommitDate to rerun travis 68da3ee4284a07f695e992afd6cdd545604aac90 -> 99d032e5b6a5140cdb07d0d209ff7f6a74bc1244 ([multicheck.9](https://github.com/ryanofsky/bitcoin/commits/pr/multicheck.9) -> [multicheck.10](https://github.com/ryanofsky/bitcoin/commits/pr/multicheck.10))

jtimon:
Concept ACK

src/wallet/rpcdump.cpp
laanwj: > Ok will change, but what are the problems with some platforms?

The conceptual problem is that PRI64d can evaluate to anything depending on the platform, e.g. "%ponyfood". `tinyformat.h` cannot know about all these platform-dependent escape sequences, so might produce garbled output or even raise an exception. We had a lot of problems with this back in the day with windows and mingw.

Offline checking could still work, the script would just need to be smart enough to cope with a) translations b) tinyformat's specifiers and type-safety. (b) is actually a subset, not a superset of what platform printfs support (none of the insecure ones like %n) so if anything that should make it easier.

src/wallet/test/wallet_tests.cpp
```diff
@@ -355,4 +361,58 @@ BOOST_AUTO_TEST_CASE(ApproximateBestSubset)
...
+    // Cap last block file size, and mine new block in a new block file.
+    CBlockIndex* oldTip = chainActive.Tip();
+    GetBlockFileInfo(oldTip->GetBlockPos().nFile)->nSize = MAX_BLOCKFILE_SIZE;
+    CreateAndProcessBlock({}, GetScriptForRawPubKey(coinbaseKey.GetPubKey()));
+    CBlockIndex* newTip = chainActive.Tip();
+
+    // Verify ScanForWalletTransactions picks up transactions in both the old
+    // and new block files.
+    {
+        CWallet wallet;
```
paveljanik: This brings a few Wshadow warinngs:

```
Wshadow statistics: 
   1 wallet/test/wallet_tests.cpp:377:17: warning: declaration shadows a variable in namespace 'wallet_tests' [-Wshadow]
   1 wallet/test/wallet_tests.cpp:391:17: warning: declaration shadows a variable in namespace 'wallet_tests' [-Wshadow]
   1 wallet/test/wallet_tests.cpp:399:17: warning: declaration shadows a variable in namespace 'wallet_tests' [-Wshadow]

```

PR #9778, opened 2017-02-16
Add two hour buffer to manual pruning

morcos:
@gmaxwell, this is meant to address https://github.com/bitcoin/bitcoin/pull/9761#issuecomment-280316826

Should probably go in 0.14 with the rest of the grace period / importmulti changes

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -820,7 +820,8 @@ UniValue pruneblockchain(const JSONRPCRequest& request)
         throw runtime_error(
             "pruneblockchain\n"
             "\nArguments:\n"
-            "1. \"height\"       (numeric, required) The block height to prune up to. May be set to a discrete height, or to a unix timestamp to prune based on block time.\n"
+            "1. \"height\"       (numeric, required) The block height to prune up to. May be set to a discrete height, or a unix timestamp\n"
+            "                  to prune blocks whose block time is at least 2 hours older than the provided timestamp.\n"
             "\nResult:\n"
             "n    (numeric) Height of the last block pruned.\n"
             "\nExamples:\n"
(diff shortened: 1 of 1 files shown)


src/rpc/blockchain.cpp
```diff
@@ -820,7 +820,8 @@ UniValue pruneblockchain(const JSONRPCRequest& request)
         throw runtime_error(
             "pruneblockchain\n"
             "\nArguments:\n"
-            "1. \"height\"       (numeric, required) The block height to prune up to. May be set to a discrete height, or to a unix timestamp to prune based on block time.\n"
+            "1. \"height\"       (numeric, required) The block height to prune up to. May be set to a discrete height, or a unix timestamp\n"
+            "                  to prune based on block time. Timestamp based pruning automatically adds a 2 hour buffer.\n"
```
ryanofsky: I'd maybe make the comment more literal and say timestamp pruning will only prune blocks that are at least 2 hours older than the provided timestamp (instead of that timestamp pruning automatically adds a 2 hour buffer).

ryanofsky (approved):
utACK 051fc42521dd5417f34ed27a727ad045a9823764

I think it should probably not be hard to add a test for this in:
https://github.com/bitcoin/bitcoin/blob/1e92e041ddc8232bcf64f09fd70891b80ac05c66/qa/rpc-tests/pruning.py#L229

Though pruning.py is kind of a pain to run.

jonasschnelli (approved):
code review utACK 051fc42521dd5417f34ed27a727ad045a9823764

jtimon:
Maybe 7200 could be made a constant to be reused in #9761 ?
Also agree with @ryanofsky 's nit, and, yes, a test would be nice (although, yeah, prunning.py is a pita).
utACK 051fc42

morcos:
ok slightly modified the help text
otherwise unchanged from ACK'ed commit

The 7200 number was added simultaneously in 3 separate PR's, but I agree it should be cleaned up to a constant in a later PR.

morcos:
Actually my personal preference would have been to leave the details of the exact buffering mechanism used off the help text.  It seems too dependent on the exact implementation choice, and there is nothing magical about the 2 hour number.   But at this point I think we are bike shedding.

gmaxwell:
utACK  you could also leave it out of the description as far as I care, it's an implementation detail.

laanwj:
I think it's fine to mention it in the description. The two hours could be important in edge cases e.g. determining what blocks to keep with manual pruning.

laanwj:
utACK 91fb506

morcos:
This caused `pruning.py` to break.
@ryanofsky or I will fix the test.

PR #9780, opened 2017-02-16
Suppress noisy output from qa tests in Travis

jnewbery:
The QA tests are currently quite noisy even when they run successfully. They output progress information which fills up pages of screen. This might be what you want when running the tests locally, but for build/integration tools it fills the output log with spam.

This PR suppresses noisy output from the qa tests. It builds on and requires both https://github.com/bitcoin/bitcoin/pull/9657 and https://github.com/bitcoin/bitcoin/pull/9768. There are three commits specifically for this PR:

- commit 1 adds a 'quiet' option to rpc-tests.py, which suppresses the progress output to stdout.
- commit 2 updates .travis.yml to use quiet mode
- commit 3 prints out the final 1000 lines of test_framework.log if the test fails.

Example of a successful travis run: https://travis-ci.org/jnewbery/bitcoin/jobs/202373787
Example of a failed travis run: https://travis-ci.org/jnewbery/bitcoin/jobs/202374810


qa/pull-tester/rpc-tests.py
```diff
@@ -146,6 +148,13 @@ def main():
     config = configparser.ConfigParser()
     config.read_file(open(os.path.dirname(__file__) + "/tests_config.ini"))
 
+    # Set up logging handler
+    sh = logging.StreamHandler()
+    sh.setLevel((logging.INFO if args.quiet else logging.DEBUG))
```
ryanofsky: Maybe only call sh.setLevel if args.quiet is true, because logging.DEBUG can still throw away logs at levels < 10.

qa/rpc-tests/test_framework/test_framework.py
```diff
@@ -172,12 +173,15 @@ def main(self):
...
+                filenames += glob.glob(self.options.tmpdir + "/node*/regtest/debug.log")
                 MAX_LINES_TO_PRINT = 1000
                 for f in filenames:
-                    print("From" , f, ":")
-                    from collections import deque
-                    print("".join(deque(open(f), MAX_LINES_TO_PRINT)))
+                    try:
+                        print("From" , f, ":")
+                        print("".join(deque(open(f), MAX_LINES_TO_PRINT)))
+                    except:
```
ryanofsky: I think this very broad exception handler could hide information we might like to know about. I would remove it, or narrow it, or at least at stick a "traceback.print_exc()" call in here.

If this handler is supposed to handle the case where f does not exist, you could add an "if os.path.exists(f)" or similar check above.

laanwj:
Concept ACK, filtering signal/noise has been a pet peeve of me for as long as we have a travis run, but never got around to it. Thanks for doing this.

JeremyRubin:
I like the approach, but I think that it would be preferable to have some output (e.g., periods) while the tests are running. This is because of a limitation in Travis CI where if no output is seen for 10 minutes the build will fail. This has been a problem historically in the unit tests, but was only addressed (by me) by making them run more quickly such that this is not an issue.

laanwj:
Yes, printing periods or another kind of progress display would be nice. My issue is with tons of unnecessary debug output.

ryanofsky:
utACK d580707d6a2481f13e8c83c08b5635c59672ae6d

jnewbery:
@laanwj - this PR suppresses the period printing when running --quiet mode. Do you want me to put that back in?

laanwj:
> this PR suppresses the period printing when running --quiet mode. Do you want me to put that back in?

Not necessarily. Just in the mode that Travis uses, apparently it needs to print something to avoid the timeout problem that @JeremyRubin mentions.

jnewbery:
I've rebased this now that #9657 has been merged. I've also made a change so that period printing still happens in quiet mode. That should prevent any Travis timeout issues.

ryanofsky:
re-utACK 85312c075e7bf3379e0a277e3393e734f940b71c. Confirmed no changes since last rebase other than restoring the print('.').

I would definitely prefer not to be adding the catch-all except handler here, though: https://github.com/bitcoin/bitcoin/pull/9780#discussion_r101642780

qa/rpc-tests/test_framework/test_framework.py
jnewbery: Printing the tails of the log files here is simply to help if the test case fails on travis. My thinking in adding the try/except handling is that if reading a log file fails for *any* reason, we should just continue and try to print out the other log files. I don't want there to be any chance that this commit makes the logging available on travis any worse.

I've added a print out of the stack trace exception in the except branch. Does that allay your concerns here?

jnewbery:
rebased and addressed https://github.com/bitcoin/bitcoin/pull/9780#discussion_r101642780.

qa/rpc-tests/test_framework/test_framework.py
ryanofsky: The print addresses most of my concern. If something goes wrong, I would prefer to know about it.

I also think you should change the bare except to `except OSError` or `except Exception`. It isn't right to ignore exceptions like SystemExit or KeyboardInterrupt and keep looping if someone is trying to kill the program. (For reference: https://docs.python.org/3/library/exceptions.html#exception-hierarchy)

jtimon: Agree on catching only a more concrete exception, specially if you're expecting one in particular, as it seems to be the case in https://github.com/bitcoin/bitcoin/pull/9780/commits/a2a565180449b80c054c05c12c93097f1e8752c2 . Also you could print the message of the exception too (before the traceback).

jtimon:
utACK a2a565180449b80c054c05c12c93097f1e8752c2 except for @ryanofsky 's nit.

jnewbery:
Updated the exception handler to catch only OSError exceptions as suggested by @ryanofsky and rebased.

jtimon:
Needs rebase again.

jnewbery:
rebased. I think this is useful PR since it suppresses noisy output in travis when tests pass and also increases the amount of logging available when the test fails. This will be helpful in debugging intermittent functional test failures in Travis. @MarcoFalke , @laanwj - can I convince you to review this PR?

laanwj:
utACK 2c0be25

MarcoFalke:
utACK 2c0be25bd6b744cab60a3672409750545522733a. Please squash the fixup commit.

jnewbery:
squashed

PR #9783, opened 2017-02-17
release: bump gitian descriptors for a new 0.14 package cache

theuni:
We should go ahead and bump master to 0.15 as well, so that it will transition to the release branch when the time comes.

I'll PR that separately.


laanwj:
utACK 42f5f21

theuni:
Yes. I'll add that along with the bump to 0.15 for master.

achow101:
It's already in the release process doc under the "Update version in sources" section.

theuni:
@achow101 Yep, thanks for checking. We just missed the bump last time around.

PR #9785, opened 2017-02-17
Avoid variable length arrays

sipa:
This removes this warning in some G++ versions:

```c++
httprpc.cpp: In function ‘bool HTTPReq_JSONRPC(HTTPRequest*, const string&)’:
httprpc.cpp:150:13: warning: stack protector not protecting local variables: variable length buffer [-Wstack-protector]
 static bool HTTPReq_JSONRPC(HTTPRequest* req, const std::string &)
             ^~~~~~~~~~~~~~~
```

### DIFF
--- a/src/httprpc.cpp
+++ b/src/httprpc.cpp
@@ -112,7 +112,7 @@ static bool multiUserAuthorized(std::string strUserPass)
             std::string strSalt = vFields[1];
             std::string strHash = vFields[2];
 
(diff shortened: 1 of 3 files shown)


theuni:
utACK c801c82e253dc4956069b6110f65b43f538dd5a1

luke-jr:
utACK

laanwj:
utACK c801c82

gmaxwell (commented):
utACK

PR #9786, opened 2017-02-17
boost: remove iostreams includes

theuni:
They're unused and produce nasty deprecation warnings like:

NOTE: Use of this header (bool_trait_def.hpp) is deprecated


laanwj:
Trivial ACK 3301587

Issue #9601, opened 2017-01-20
Remove coin age priority and free transactions - discussion


morcos:
I've opened #9602 to remove all coin age priority functionality from the codebase.

I think this has been discussed many times and hopefully it will not be the subject of another lengthy discussion, but just in case, I've opened this issue so any discussion not related to the technical merits of the implementation can happen here and the PR itself can remain free for discussion of the actual code.

This is targeted for 0.15, and shouldn't be merged until we ship 0.14, but it came up again recently and so I figured I might as well post the PR.  Also in the event that we feel anything needs to be pre-announced with the 0.14 release this will give us a chance to think about that, although I don't think there is anything pressing.

MarcoFalke:
Looks like this is a duplicate of  #6675

DATSEC:
I am against
It works as an anti-spam addresses

luke-jr:
Concept NACK to removal from mining:

* Miner(s) still use priority. Whether or not you agree with this, policy is not up to developers to decide and force on miners.
* Coin age is a useful algorithm for altruistic miners who wish to keep the network functioning smoothly even during flood attacks.
* Coin age ensures that so long as a non-spam transaction pays a reasonable fee, it will *eventually* get mined. (RBF may eventually make this assurance unnecessary, but we're not there yet.)
* Using a single algorithm for block creation makes it significantly easier for floods to game the algorithm to DoS the entire network. Merely the existence of multiple different sorting algorithms for blocks helps avoid the situation where blocks are pure flood and legitimate transactions can't get through.
* Regardless of past complications, the current code is cleanly isolated from the rest of the code, and a mere 100 lines. If it creates a burden for future modifications, removing priority can be put back on the table and weighed against the value of those modifications (or ported by someone interested in keeping it; eg, me).
* Multiple algorithms for mining (regardless of what those algorithms may be) also avoids the situation where policy logic gets too deeply embedded into non-policy code, creating huge risks and burdens for anyone who wishes to customise it without modifying any non-policy code.

I can alternatively maintain coin-age priority in Knots only, but I do not think Core should head in this direction of pushing a single specific political agenda at the exclusion of all others. If it does go that route, it should be made clear that Core is no longer a reference implementation, and its political agenda documented.

Concept ACK to everything else (eg, removal of support for zero-fee transactions, and priority estimation)

jnewbery:
@luke-jr you have your opinions on priority, which I completely respect. However, this:

[...]

PR #9790, opened 2017-02-17
doc/release-notes: Various cleanups

luke-jr:
- Remove minor/obvious changes
- Clarify and document new options in "Minimum Fee Rate Policies"
- Attribution fixes
- Remove deprecation of miner priority sorting

### DIFF
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -107,15 +107,6 @@ Additionally the `setnetworkactive` RPC command has been added which does
 the same thing as the GUI icon. The command takes one boolean parameter,
 `true` enables networking and `false` disables it.
 
-Out-of-sync Modal Info Layer
-----------------------------
-
-When Bitcoin Core is out-of-sync on startup, a semi-transparent information
-layer will be shown over top of the normal display. This layer contains
-details about the current sync progress and estimates the amount of time
-remaining to finish syncing. This layer can also be hidden and subsequently
-unhidden by clicking on the progress bar at the bottom of the window.
-
 Support for JSON-RPC Named Arguments
 ------------------------------------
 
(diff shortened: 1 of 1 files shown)


jnewbery:
NACK. You seem to have accidentally removed the part about priority transaction sorting for mining being deprecated in Core and being removed in the next major version.

luke-jr:
Not accidentally, it's documented in the summary. There is no consensus to remove priority mining.

sipa:
Please stop this, @luke-jr. I'm aware you dislike the removal of priority mining, but it seems the vast majority of contributors to the project agree it is pointless to maintain at this point.

luke-jr:
@sipa I tried to find a compromise where nobody had to maintain it besides me, but certain people were stubborn and refused. So the status quo of "no consensus to remove" remains for now.

sipa:
@luke-jr This is not a consensus change, and it's not up to you to block this. I absolutely respect that you want priority mining to remain available, but you'll have to do it in your own fork, or develop an external tool that implements it through `prioritizetransaction` RPC calls (which seems very useful in any case as an example).

luke-jr:
@sipa I think [this comment](https://github.com/bitcoin/bitcoin/issues/9601#issuecomment-279600986) deserves proper consideration before any final decision is made. But in the case that priority is to be removed, how about simply adding to the current text, "Miners who wish to retain this feature should switch to Bitcoin Knots, where it will continue to be maintained and supported for the time being."?

sipa:
@luke-jr Ok, I've commented.

> how about simply adding to the current text, "Miners who wish to retain this feature should switch to Bitcoin Knots, where it will continue to be maintained and supported for the time being."?

I disagree with recommending the use of software based on a feature we don't consider relevant ourselves.

luke-jr:
I do consider it relevant, and even aside from that, it is up to miners to decide on their own if they consider it relevant, not to have information about its availability suppressed so they are forced out of using it.

sipa:
> I do consider it relevant

That's fine, you can recommend whatever you like. But you can't decide alone what goes into Bitcoin Core's release notes.

MarcoFalke:
Closing. This missed the deadline.

PR #9791, opened 2017-02-17
Avoid VLA in hash.h

sipa:


gmaxwell (approved):
utACK.

paveljanik:
Compiles and works together with #9789.

ACK https://github.com/bitcoin/bitcoin/pull/9791/commits/5c8fd508112016c3821b71112270afe83e6949ac

laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/9791/commits/5c8fd508112016c3821b71112270afe83e6949ac

sipa:
To clarify: my GCC does not consider this to be a VLA, but at least some Clang versions do.

laanwj:
Well I think this makes the code clearer too, so it's a win in any case.

MarcoFalke:
5c8fd508112016c3821b71112270afe83e6949ac gives same binaries with recent gcc on my arch.

utACK 5c8fd508112016c3821b71112270afe83e6949ac

PR #9794, opened 2017-02-18
Minor update to qrencode package builder

MitchellCash:
Use package name variable inside qrencode $(package)_file_name variable


fanquake:
Tested ACK 1bfe6b4

MarcoFalke:
utACK 1bfe6b4

laanwj:
What is the effect of this?

MarcoFalke:
No effect. It just seems that this style is used in the other files.

mitchellcash:
@laanwj sorry I should have been more specific in my description, but as @MarcoFalke mentions, it is more about being consistent with the other files.

PR #9798, opened 2017-02-19
Fix Issue #9775 (Check returned value of fopen)

### LINKED ISSUE #9775: Check returned value of fopen()
[util.cpp#L216](https://github.com/bitcoin/bitcoin/blob/master/src/util.cpp#L216) 
if `fopen `failed, a null file pointer could be passed to `FileWriteStr(vMsgsBeforeOpenLog->front(), fileout);`

kirit93:
Making sure file is opened before writing string to file to avoid seg faults in case open() returns a NULL pointer. 

Attempts to fix #9775

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -214,12 +214,13 @@ void OpenDebugLog()
     assert(vMsgsBeforeOpenLog);
     boost::filesystem::path pathDebug = GetDataDir() / "debug.log";
     fileout = fopen(pathDebug.string().c_str(), "a");
-    if (fileout) setbuf(fileout, NULL); // unbuffered
-
-    // dump buffered messages from before we opened the log
-    while (!vMsgsBeforeOpenLog->empty()) {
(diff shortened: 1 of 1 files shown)


src/util.cpp
```diff
@@ -164,7 +164,7 @@ class CInit
             delete ppmutexOpenSSL[i];
         OPENSSL_free(ppmutexOpenSSL);
     }
-}
+};
```
laanwj: Adding a ; here is wrong.

kirit93: I removed the semicolon in the latest commit.

MarcoFalke:
Please also remove the trailing whitespace

kirit93:
Hey I just noticed this issue has been closed. Do we still need this fix?

laanwj:
This issue hasn't been closed, we still need it. There were just some comments.

Looks good to me now. Could you squash to one commit? (e.g. follow https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#squashing-commits)

PR #9801, opened 2017-02-19
Removed redundant parameter from mempool.PrioritiseTransaction

gubatron:
The string parameter is used only for logging.

The only place where the `string` form of the hash isn't recalculated is at `src/rpc/mining.cpp` where it's obtained from the arguments, however, in this case the `uint256` hash is then parsed from the given `string`.

In every other use of the function, the `string` representation is always calculated from the `uint256 hash` object (either with `ToString()` or with `GetHash()`), so no real benefit on having this redundant parameter.

This also makes the API more consistent with other methods that only receive the `uint256` hash parameter.

### DIFF
--- a/src/txmempool.cpp
+++ b/src/txmempool.cpp
@@ -920,7 +920,7 @@ CTxMemPool::ReadFeeEstimates(CAutoFile& filein)
     return true;
 }
 
-void CTxMemPool::PrioritiseTransaction(const uint256 hash, const std::string strHash, double dPriorityDelta, const CAmount& nFeeDelta)
+void CTxMemPool::PrioritiseTransaction(const uint256& hash, double dPriorityDelta, const CAmount& nFeeDelta)
 {
     {
         LOCK(cs);
@@ -940,7 +940,7 @@ void CTxMemPool::PrioritiseTransaction(const uint256 hash, const std::string str
             }
         }
     }
-    LogPrintf("PrioritiseTransaction: %s priority += %f, fee += %d\n", strHash, dPriorityDelta, FormatMoney(nFeeDelta));
+    LogPrintf("PrioritiseTransaction: %s priority += %f, fee += %d\n", hash.ToString(), dPriorityDelta, FormatMoney(nFeeDelta));
 }
 
 void CTxMemPool::ApplyDeltas(const uint256 hash, double &dPriorityDelta, CAmount &nFeeDelta) const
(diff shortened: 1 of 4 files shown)


jonasschnelli (approved):
Makes sense. Thanks.
utACK 8e88a445dc9de71657282a461cb6b679c9079ebd

paveljanik:
ACK https://github.com/bitcoin/bitcoin/commit/8e88a445dc9de71657282a461cb6b679c9079ebd

dcousens (approved):

MarcoFalke:
utACK 8e88a44

src/txmempool.cpp
```diff
@@ -920,7 +920,7 @@ CTxMemPool::ReadFeeEstimates(CAutoFile& filein)
     return true;
 }
 
-void CTxMemPool::PrioritiseTransaction(const uint256 hash, const std::string strHash, double dPriorityDelta, const CAmount& nFeeDelta)
+void CTxMemPool::PrioritiseTransaction(const uint256 hash, double dPriorityDelta, const CAmount& nFeeDelta)
```
sdaftuar: Perhaps while you're at it you could make this `const uint256& hash`?

gubatron: done deal, thanks.

(squashed and rebased)

src/txmempool.h
```diff
@@ -555,7 +555,7 @@ class CTxMemPool
     bool HasNoInputsOf(const CTransaction& tx) const;
 
     /** Affect CreateNewBlock prioritisation of transactions */
-    void PrioritiseTransaction(const uint256 hash, const std::string strHash, double dPriorityDelta, const CAmount& nFeeDelta);
+    void PrioritiseTransaction(const uint256& hash, const std::string strHash, double dPriorityDelta, const CAmount& nFeeDelta);
```
MarcoFalke: The redundant parameter is not removed, as mentioned in the commit message.

MarcoFalke: You can run `make check` locally as a quick sanity check.

gubatron: yes, something got screwed with the rebase, re-uploaded now. Thanks.

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/9801/commits/eaea2bbb411c5d878e991d9082ca037628efce42

MarcoFalke:
utACK eaea2bb

PR #9805, opened 2017-02-19
Add seed.btc.petertodd.org to mainnet DNS seeds

petertodd:
Identical setup to my testnet DNS seed, but for mainnet.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -125,6 +125,7 @@ public:
         vSeeds.push_back(CDNSSeedData("dashjr.org", "dnsseed.bitcoin.dashjr.org")); // Luke Dashjr
         vSeeds.push_back(CDNSSeedData("bitcoinstats.com", "seed.bitcoinstats.com", true)); // Christian Decker, supports x1 - xf
         vSeeds.push_back(CDNSSeedData("bitcoin.jonasschnelli.ch", "seed.bitcoin.jonasschnelli.ch", true)); // Jonas Schnelli, only supports x1, x5, x9, and xd
+        vSeeds.push_back(CDNSSeedData("petertodd.org", "seed.btc.petertodd.org", true)); // Peter Todd, only supports x1, x5, x9, and xd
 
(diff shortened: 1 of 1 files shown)


cdecker:
ACK 3b4dd2a6768536be3a7db21353f43ba614c17377

Nice to see this added. I've been monitoring the seed for a while so let me add some additional data about the seed: it returns about 50% reachable nodes and (from my vantage point) has ~100ms latency. The stale rate is a bit high-ish, but that might be easy to fix.

petertodd:
Thanks! Didn't know anyone else was monitoring it.

I reset the database on it a week ago or so when I upgrading the seeder to @sipa's latest bitcoin-seeder code, so possibly that's why the stale rate is a bit high; maybe needs time to collect more data? Specifically it's running commit https://github.com/sipa/bitcoin-seeder/commit/d54f39a1eeb7ddb0638396755d3a76abbd4c41ef, with one small patch to also return full-RBF flagging nodes (as does my testnet seed).

cdecker:
Yeah, I added it some time ago to http://bitcoinstats.com/network/dns-servers/ :wink:

laanwj:
Concept ACK.

I suppose you sign off on the [dnsseed policy](https://github.com/bitcoin/bitcoin/blob/master/doc/dnsseed-policy.md)?

petertodd:
@laanwj Yup, ACK https://github.com/bitcoin/bitcoin/blob/99963b938ffb4ad37f4a9d275998bf1a9dc11e64/doc/dnsseed-policy.md

gmaxwell (approved):
utACK

jonasschnelli:
Tested ACK.
IPv4 and IPv6 works. Tried a couple of IPs retrived over the seeder. Crawler seems to be up to date.

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/9805/commits/3b4dd2a6768536be3a7db21353f43ba614c17377

PR #8660, opened 2016-09-03
txoutsbyaddress index (take 2)

djpnewton:
This is an attempt to revive PR #5048

I have moved the index to its own database and also added a gettxoutsbyaddress REST command

I still need to think about the layout of binary input/output for the REST command, or is bin encoding necessary?

Also still missing tests for the new REST stuff

PR #5048, opened 2014-10-05
txoutsbyaddress index

cozz:
Adds new rpc call "gettxoutsbyaddress" as requested in #4007.
Disabled by default, enabled with -txoutsbyaddressindex.

The index is built from the normal utxo on first use, reindex is not required.
For the GUI there is progress shown, on bitcoind you just need to be patient.

The qa-test includes a simple reorganization.
I tested all 4 code parts in main.cpp by commenting them out and check if the qa-test fails at the expected line.
I had to modify the rpc call for this test, as normally wrong outputs in the address index are not exposed to the user, because the rpc call relies on the normal GetCoins call.
The rest of the code has been tested in random ad-hoc testing.
I only tested linux on my machine.

Its all in 1 commit for laziness reasons, but if splitting in multiple commits would help reviewing, I could do that.

txoutsbyaddress

txoutsbyaddress3

PR #9806, opened 2017-02-20
txoutsbyaddress index (take 3)

droark:
This is an attempt to revive PR #8660 (and, by extension, PR #5048). For now, this PR simply compiles without fresh warnings or errors. Once it is confirmed that no more conflicts exist, the remaining comments/requests from #8660 will be fully addressed.


paveljanik:
Wshadow statistics: 
   1 coinsbyscript.cpp:180:2200: warning: declaration shadows a local variable [-Wshadow]

droark:
@paveljanik - Thanks! Fixed that up.

laanwj:
Thanks for reviving this. IMO this is important.

sidhujag:
Concept ACK

instagibbs:
needs rebase

droark:
@instagibbs - Rebased. Thanks for the heads up. (Side note: Is there any way to have GH tell you when a conflict occurs? That would be really handy.)

@sidhujag - Good question. I don't know offhand. I'm happy to do some benchmarking. I may need some help with that. If anybody would like to make some suggestions, I'm all ears.

All - Will get more movement on this. Life intervened for awhile and has finally slowed down enough to where I can dedicate more time to this.

sidhujag:
Thanks droark great feature btw.. this makes walletless spending very easy

droark:
Am placing a to-do list here to remind myself of what I need to do, and solicit feedback on anything people might think is missing.

- [x] Remove all Boost code from the PR.
- [ ] Address feedback from #8660 (primarily @ryanofsky). (Mostly done but I want to double check a couple of things before proceeding.)
- [ ] Probably switch names of this & that per the suggestion of @gmaxwell. ("utxoindex" is my current choice but I'm not attached to it.)
- [ ] Add a default constructor for CTxMemPool. (A recently merged PR, combined with this PR, inadvertently nukes the default constructor.)
- [ ] Check in once a day to see if a rebase is needed. (Ongoing task.)

btcdrak:
needs rebase.

droark:
@btcdrak - Thanks. Rebased.

weex:
Running this on a .bitcoin/ folder that was last used with 0.13 and was run pruned to 2000. It seems 1GB of ram is too low for this enabled as I got an out of memory error. Now on restarting I get the following after/during the "Rescanning..." step:

    ~/.bitcoin# bitcoind: coinsbyscript.cpp:53: CCoinsByScript& CCoinsViewByScript::GetCoinsByScript(const CScript&, bool): Assertion `it != cacheCoinsByScript.end()' failed.

src/init.cpp
```diff
@@ -1484,6 +1492,58 @@ bool AppInitMain(boost::thread_group& threadGroup, CScheduler& scheduler)
...
                 }
 
+                // Check -txoutindex
+                pcoinsByScriptDB->ReadFlag("txoutindex", fTxOutIndex);
+                if (IsArgSet("-txoutindex"))
+                {
+                    if (GetBoolArg("-txoutindex", false))
+                    {
+                        // build index
+                        if (!fTxOutIndex)
```
luke-jr: We also need to build the index if it's outdated. For example, if the user ran an older version to sync. (This is fixed automatically if you replace the flag with a check that the index's best block matches that of the chainstate.)

src/rest.cpp
```diff
@@ -598,6 +601,109 @@ static bool rest_getutxos(HTTPRequest* req, const std::string& strURIPart)
     return true; // continue to process further HTTP reqs on this cxn
 }
 
+static bool rest_getutxoindex(HTTPRequest* req, const std::string& strURIPart)
```
luke-jr: (did not review)

src/rpc/blockchain.cpp
```diff
@@ -876,6 +828,8 @@ UniValue gettxoutsetinfo(const JSONRPCRequest& request)
             "  \"bestblock\": \"hex\",   (string) the best block hash hex\n"
             "  \"transactions\": n,      (numeric) The number of transactions\n"
             "  \"txouts\": n,            (numeric) The number of output transactions\n"
+            "  \"addresses\": n,         (numeric) The number of addresses and scripts. Only if -txoutindex=1\n"
```
luke-jr: Call it `"uniquescripts"`?

src/rpc/blockchain.cpp
```diff
@@ -876,6 +828,8 @@ UniValue gettxoutsetinfo(const JSONRPCRequest& request)
...
+            "  \"utxoindex\": n,   (numeric) The number of output transactions. Only if -txoutindex=1\n"
```
luke-jr: Duplicates `"txouts"`. Remove and simply add an integrity check.

src/rpc/blockchain.cpp
```diff
@@ -983,6 +939,117 @@ UniValue gettxout(const JSONRPCRequest& request)
...
+            "getutxoindex ( minconf [\"address\",...] count from )\n"
+            "\nReturns a list of unspent transaction outputs by address (or script).\n"
+            "The list is ordered by confirmations in descending order.\n"
+            "Note that passing minconf=0 will include the mempool.\n"
+            "\nTo use this function, you must start bitcoin with the -txoutindex parameter.\n"
+            "\nArguments:\n"
+            "1. minconf          (numeric) Minimum confirmations\n"
+            "2. \"addresses\"    (string) A json array of bitcoin addresses (or scripts)\n"
+            "    [\n"
+            "      \"address\"   (string) bitcoin address (or script)\n"
```
luke-jr: The address abstraction ends before/when coins hit the UTXO set. UTXOs do not have and are not associated to addresses.

Rename this to `"scripts"`, and don't support specifying as addresses.

src/rpc/blockchain.cpp
```diff
@@ -983,6 +939,117 @@ UniValue gettxout(const JSONRPCRequest& request)
...
+            "  }\n"
+            "  ,...\n"
+            "]\n"
+            "\nExamples:\n"
+            + HelpExampleCli("getutxoindex", "6 \"[\\\"1PGFqEzfmQch1gKD3ra4k18PNj3tTUUSqg\\\",\\\"1LtvqCaApEdUGFkpKMM4MstjcaL4dKg8SP\\\"]\"")
+            + "\nAs a json rpc call\n"
+            + HelpExampleRpc("getutxoindex", "6, \"[\\\"1PGFqEzfmQch1gKD3ra4k18PNj3tTUUSqg\\\",\\\"1LtvqCaApEdUGFkpKMM4MstjcaL4dKg8SP\\\"]\"")
+        );
+
+    if (!fTxOutIndex)
```
luke-jr: Add braces

src/rpc/blockchain.cpp
```diff
@@ -983,6 +939,117 @@ UniValue gettxout(const JSONRPCRequest& request)
...
+        throw JSONRPCError(RPC_METHOD_NOT_FOUND, "To use this function, you must start bitcoin with the -txoutindex parameter.");
+
+    RPCTypeCheck(request.params, boost::assign::list_of(UniValue::VNUM)(UniValue::VARR)(UniValue::VNUM)(UniValue::VNUM));
+
+    UniValue vObjects(UniValue::VARR);
+    std::vector<std::pair<int, unsigned int> > vSort;
+    int nMinDepth = request.params[0].get_int();
+    UniValue inputs = request.params[1].get_array();
+
+    int nCount = 999999999;
```
luke-jr: Max [guaranteed] value of `int` is `32767`, so this could overflow.

src/rpc/blockchain.cpp
```diff
@@ -983,6 +939,117 @@ UniValue gettxout(const JSONRPCRequest& request)
...
+            script = CScript(data.begin(), data.end());
+        } else {
+            throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid Bitcoin address or script: " + input.get_str());
+        }
+
+        CCoinsByScript coinsByScript;
+        pcoinsByScript->GetCoinsByScript(script, coinsByScript);
+
+        if (nMinDepth == 0)
+            mempool.GetCoinsByScript(script, coinsByScript);
```
luke-jr: Need to document that mempool.GetCoinsByScript must only add, not replace.

src/rpc/blockchain.cpp
```diff
@@ -983,6 +939,117 @@ UniValue gettxout(const JSONRPCRequest& request)
...
+
+        CoinsByScriptToJSON(coinsByScript, nMinDepth, vObjects, vSort, true); 
+    }
+
+    UniValue results(UniValue::VARR);
+    sort(vSort.begin(), vSort.end());
```
luke-jr: Why sort? Let the caller sort if they need it sorted somehow.

src/rpc/blockchain.cpp
```diff
@@ -983,6 +939,117 @@ UniValue gettxout(const JSONRPCRequest& request)
...
+    for (unsigned int i = (unsigned int)nFrom; i < vSort.size(); i++)
```
luke-jr: `size_t`

src/rpc/rawtransaction.cpp
```diff
@@ -122,6 +122,63 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
     }
 }
 
+void CoinsByScriptToJSON(const CCoinsByScript& coinsByScript, int nMinDepth, UniValue& vObjects, std::vector<std::pair<int, unsigned int>>& vSort, bool fIncludeHex)
+{
+    BOOST_FOREACH(const COutPoint &outpoint, coinsByScript.setCoins)
```
luke-jr: Use C++11 `for`-each.

src/rpc/blockchain.cpp
```diff
@@ -983,6 +939,117 @@ UniValue gettxout(const JSONRPCRequest& request)
...
 }
 
+UniValue getutxoindex(const JSONRPCRequest& request)
+{
+    if (request.fHelp || request.params.size() < 2 || request.params.size() > 4)
+        throw std::runtime_error(
+            "getutxoindex ( minconf [\"address\",...] count from )\n"
+            "\nReturns a list of unspent transaction outputs by address (or script).\n"
+            "The list is ordered by confirmations in descending order.\n"
+            "Note that passing minconf=0 will include the mempool.\n"
```
luke-jr: Consider the case of a current UTXO that has been spent in the mempool. Under what conditions is the mempool ignored and that UTXO returned? Would you ever want to see it while also searching the mempool?

src/rpc/rawtransaction.cpp
```diff
@@ -122,6 +122,63 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
...
+        if (nMinDepth == 0)
+        {
+            LOCK(mempool.cs);
+            CCoinsViewMemPool view(pcoinsTip, mempool);
+            if (!view.GetCoins(outpoint.hash, coins))
+                continue;
+            mempool.pruneSpent(outpoint.hash, coins); // TODO: this should be done by the CCoinsViewMemPool
+        }
+        else if (!pcoinsTip->GetCoins(outpoint.hash, coins))
+            continue;
```
luke-jr: Inconsistent behaviour for UTXOs created in mined transactions vs in the mempool. We should probably consider mempool spends the same for both mempool-created UTXOs and confirmed-tx-created UTXOs.

src/rpc/rawtransaction.cpp
```diff
@@ -122,6 +122,63 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
...
+            ScriptPubKeyToJSON(coins.vout[outpoint.n].scriptPubKey, oScriptPubKey, fIncludeHex);
+
+            UniValue o(UniValue::VOBJ);
+            o.push_back(Pair("confirmations", nConfirmations));
+            o.push_back(Pair("txid", outpoint.hash.GetHex()));
+            o.push_back(Pair("vout", (int)outpoint.n));
+            o.push_back(Pair("value", ValueFromAmount(coins.vout[outpoint.n].nValue)));
+            o.push_back(Pair("scriptPubKey", oScriptPubKey));
+            o.push_back(Pair("version", coins.nVersion));
+            o.push_back(Pair("coinbase", coins.fCoinBase));
```
luke-jr: Rather call this `"generated"`

src/test/blockencodings_tests.cpp
```diff
@@ -57,7 +57,7 @@ static CBlock BuildBlockTestCase() {
 
 BOOST_AUTO_TEST_CASE(SimpleRoundTripTest)
 {
-    CTxMemPool pool;
+    CTxMemPool pool(false);
```
luke-jr: `false` is the default. No need to change all this.

src/txdb.cpp
```diff
@@ -98,7 +99,8 @@ CCoinsViewCursor *CCoinsViewDB::Cursor() const
        that restriction.  */
     i->pcursor->Seek(DB_COINS);
     // Cache key of first record
-    i->pcursor->GetKey(i->keyTmp);
+    if (!i->pcursor->Valid() || !i->pcursor->GetKey(i->keyTmp))
```
luke-jr: Braces

src/txmempool.cpp
```diff
@@ -370,6 +371,16 @@ unsigned int CTxMemPool::GetTransactionsUpdated() const
     return nTransactionsUpdated;
 }
 
+void CTxMemPool::GetCoinsByScript(const CScript& script, CCoinsByScript& coinsByScript) const
+{
```
luke-jr: Throw an exception if CTxMemPool was created without it?

src/txmempool.cpp
```diff
@@ -432,6 +443,11 @@ bool CTxMemPool::addUnchecked(const uint256& hash, const CTxMemPoolEntry &entry,
     vTxHashes.emplace_back(tx.GetWitnessHash(), newit);
     newit->vTxHashesIdx = vTxHashes.size() - 1;
 
+    if (fTxOutIndex)
```
luke-jr: Braces...

src/validation.cpp
```diff
@@ -2051,6 +2055,62 @@ void PruneAndFlush() {
     FlushStateToDisk(state, FLUSH_STATE_NONE);
 }
 
+void static UpdateAddressIndex(const CTxOut& txout, const COutPoint& outpoint, bool fInsert)
```
luke-jr: This should be moved to `coinsbyscript.cpp` and renamed.

src/validation.cpp
```diff
@@ -2051,6 +2055,62 @@ void PruneAndFlush() {
...
+    {
+        CCoinsByScript &coinsByScript = pcoinsByScript->GetCoinsByScript(txout.scriptPubKey, !fInsert);
+        if (fInsert)
+            coinsByScript.setCoins.insert(outpoint);
+        else
+            coinsByScript.setCoins.erase(outpoint);
+    }
+}
+
+void static UpdateAddressIndex(const CBlock& block, CBlockUndo& blockundo, bool fConnect)
```
luke-jr: This should be moved to `coinsbyscript.cpp` and renamed.

test/functional/rest.py
```diff
@@ -47,14 +47,14 @@ def __init__(self):
         super().__init__()
         self.setup_clean_chain = True
         self.num_nodes = 3
+        self.extra_args = [["-txoutindex"]] * 3
```
luke-jr: (didn't review tests)

test/functional/utxoindex.py
```diff
@@ -0,0 +1,95 @@
+#!/usr/bin/env python3
```
luke-jr: (didn't review tests)

luke-jr (changes requested):

droark:
@luke-jr - Thanks for the detailed feedback. I'll look over it this weekend.

dexX7:
Is there a chance this gets adopted?

droark:
@dexX7 - Thanks for checking in. Some family-related issues came up recently that I had to deal with for awhile. I'm almost done rebasing the PR and will catch up with the remaining feedback ASAP, along with some test harness changes that probably need to be folded in. The path forward looks pretty clear to me, IMO. I just need to wrap up the work.

jtimon:
I still need to read more, but...

Regarding searching by address , by scriptPubKey, or by COutPoint (tx_id, output_pos), I'm not sure whether I want them all or a subset of them.
The later doesn't require any further index. 
Searching by scriptPubKey only requires an additional scriptPubKey -> COutPoint index but not much logic.
Searching by address on the other hand...there's many types of addresses and people still propose new better ones...I wonder if perhaps the translation address -> scriptPubKey (which shouldn't require an extra index on top of the scriptPubKey one) belongs to an upper layer like the wallet and/or ui that the users of this feature implement. I'm like supporting it here, but maybe script/standard needs to be extended for the address -> scriptPubKey translator (sorry to reiterate, but I don't think we want an address index, just a one way translation function [the other way is for wallets and relay policies, or higher level protocols ala coinjoin and whatnot]). 

This all assumes you create a new scriptPubKey -> COutPoint index.

Regarding utxo vs stxo vs txo...
You can easily serve both utxo and stxo (that is, the whole txo) by just calling GetTransaction() (in validation.o) like getrawtransaction does, without needing any consideration on whether the the output is spent or not while getting the better performance if the COutPoint happens to be in the current utxo, which is very nice.
That will just complain if you search when you are pruning or not using -txindex and thus you cannot serve certain outputs unless you do a rescan which is completely out of the question.
You probably want a couple of options -utxoscriptindex and -txoscriptindex or something of the sort.
Having -utxoscriptindex but not -txoscriptindex maybe should force the error when searching for spent txo's. Having both is just reduntant. Having -txoscriptindex implies having the full historic scriptPubKey -> COutPoint index.
You don't even need to remove the spent entries from the index if you only have -utxoscriptindex if you make sure the outputs fail when -utxoscriptindex is set but -txoscriptindex is not. In that case, garbage collecting or "is spent reference counting" or "worrying about reorgs" or "adding a spent bool to the index" would probably be nice additions to have since that selection could prune them and otherwise they will accumulate over running time, even if they're just entries on an index and not the outputs themselves. That seems like it could be an important distraction in the short term.

So although I'm not against gettxoutsbyaddress and I celebrate concept ACK it, but I think it would be nicer to get gettxoutsbyoutpoint and gettxoutsbyscript reviewed and merged in that order first. But I think some other people don't like chained PRs all that much, so just take it as "if we're offering gettxoutsbyaddress, we should be able to offer gettxoutsbyoutpoint and gettxoutsbyscript 'almost for free'(tm) too".

It may seem contradictory: I'm asking you to do more with respect to utxo/stxo, but I'm suggesting to do less with respect to addresses and properly removing things for the users that want this feature but don't want to be full archives, which is something we can optimize further later after leveraging the easy important one, which is just looking in the utxo first.
The way I see it, the utxo is just the subset of the txo that can't be pruned, we happen to cache very efficiently, so it is nice to always search there first just in case and well, people usually care less about stxo because it is already spent...but is simple to also serve it or just complain when you can't or you are not expected to, sorry to reiterate.

Anyway, feel free to note the parts of my long post you like in whatever order you like best and ignore the rest. I plan to have a deeper look either way beyond this concept ack.

All the best but needs rebase.

jtimon:
Sorry, no, using GetTransaction doesn't leverage the utxo index we maintain...One would need to search in the utxo (with CCoinsViewCache::AccessCoin) first and then try with GetTransaction only if we want to also serve stxo, which should perhaps be left out of scope for this PR as a later improvement.

jtimon:
Sorry for the long post, not for being long but for being incorrect.

There's not need for a gettxobyoutpoint or similar because it already exists and is named gettxout!
That means everything I said about "optimize later, it shouldn't be hard" turns into "it's already done".

Only the slower parts were missing, which I plan to serve in #10822 which is just a draft that needs lots of testing.

To reiterate, the important parts of my previous fedback:

1) I'm more than happy to extend whatever you do for utxo for txo too ideallly almost transparently
2) Don't do an address index, do a more generic script (that means binary or lower level, hex strings on rpc) to outpoint index, and once you have the outpoint you have the whole txo (unless you don't tcindex or you prune).
3) Create a function to translate "an address" into a script so that it can be easily searched for in the outpoint -> script index. There will always be new address types your index doesn't explicitly support yet, make them support your index instead.

The PR that should help extend whatever is done for utxo to txo if you do the index: outpoint -> script; is the following:

https://github.com/bitcoin/bitcoin/pull/10822

As said it needs testing, but there's no need to wait, let's expose utxo if we can first and then txo "almost for free"(tm).

Sorry again for invading the pr in a distractive way, but this needs rebase.

braydonf:
What's the status of this PR, needs more review and testing? General concept ok?

I think it would be useful to have an index on the UTXO set for the purposes of supporting external wallets.

jnewbery:
Closing as 'up for grabs'. PR author hasn't checked in for about a year.

@droark - if you still want this, feel free to re-open and rebase!

MarcoFalke:
Picked up in " Add address-based index (attempt 4?) #14053 "

laanwj:
> Picked up in " Add address-based index (attempt 4?) #14053 "

I think those are different, right? this an indeex over the UTXO set, while that is an index over all addresses in the block chain? Or do I get this wrong.

MarcoFalke:
Indeed looks like it. I wonder why it is called take 4 then.

Though, scanning the utxo set can already be done with `scantxoutset`, no?

laanwj:
> Though, scanning the utxo set can already be done with scantxoutset, no?

Absolutely, that does the same thing. Though it can be pretty slow, which can be awkward if it's part of some user flow, so UTXO indexing is not completely out of the picture probably.

PR #9813, opened 2017-02-20
Read/write mempool.dat as a binary.

### LINKED ISSUE #9810: 0.14 not loading mempool.dat?
### Describe the issue

Restarting bitcoin, mempool starts filling from scratch.

#### If so, please list the steps to reproduce below:
1. Wait for mempool to reach for example 3000 transactions
2. Restart bitcoin
3. Watch mempool filling slowly from scratch

### Expected behaviour
Mempool would already contain ~3000 TX on startup

### Actual behaviour
It starts at 0 and fills up slowly with a few per second.

### Screenshots.
http://imgur.com/a/or82d

Windows x64 0.14rc1 binary from https://bitcoin.org/bin/bitcoin-core-0.14.0/test.rc1/

### Any extra information that might be useful in the debugging process.
I may not be aware of how the mempool.dat thing works and there could just be a misunderstanding.

paveljanik:
`mempool.dat` is a binary file and thus it should be read/written as such.

Fixes #9810.

I can't test on Windows though.

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -4180,7 +4180,7 @@ static const uint64_t MEMPOOL_DUMP_VERSION = 1;
 bool LoadMempool(void)
 {
     int64_t nExpiryTimeout = GetArg("-mempoolexpiry", DEFAULT_MEMPOOL_EXPIRY) * 60 * 60;
-    FILE* filestr = fopen((GetDataDir() / "mempool.dat").string().c_str(), "r");
+    FILE* filestr = fopen((GetDataDir() / "mempool.dat").string().c_str(), "rb");
     CAutoFile file(filestr, SER_DISK, CLIENT_VERSION);
     if (file.IsNull()) {
         LogPrintf("Failed to open mempool file from disk. Continuing anyway.\n");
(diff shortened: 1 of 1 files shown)


laanwj:
Obviously correct ACK https://github.com/bitcoin/bitcoin/pull/9813/commits/171fc91f061d4a980eedfa522e302b8598408cc5

```
<paveljanik> binary on write?
<paveljanik> Windows...
<wumpus> that could be the explanation, good catch! how is the file opened?
<wumpus> if its opened in text (instead of binary) mode, windows will convert \n to \r\n
<paveljanik> it is written with "w" and opened with "r"
<wumpus> that should be "wb" and "rb"
<wumpus> so you've found the issue, congrats
```

I wonder if we have any more instances of this.

theuni:
Good catch. utACK 171fc91f061d4a980eedfa522e302b8598408cc5

achow101:
ACK 171fc91f061d4a980eedfa522e302b8598408cc5

Tested on windows 10 and ubuntu 16.04.

You *may* need to delete the original mempool.dat for this to work though. I'm not entirely sure and may have accidentally screwed up that part when I was testing it. I know for sure that with mempool.dat files created with this fix it works fine.

gmaxwell:
utACK.

gmaxwell (approved):
utACK

fanquake:
utACK 171fc91

jonasschnelli:
utACK 171fc91f061d4a980eedfa522e302b8598408cc5

PR #9815, opened 2017-02-21
Trivial: use EXIT_ codes instead of magic numbers

benma:
To be consistent with other exit() calls.

### DIFF
--- a/src/bitcoind.cpp
+++ b/src/bitcoind.cpp
@@ -136,17 +136,17 @@ bool AppInit(int argc, char* argv[])
         if (!AppInitBasicSetup())
         {
             // InitError will have been called with detailed error, which ends up on console
-            exit(1);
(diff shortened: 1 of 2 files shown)


rabidus (approved):

laanwj:
Yeah why not, wonder why these were not caught in similar previous pulls. Anyhow, utACK https://github.com/bitcoin/bitcoin/pull/9815/commits/a87d02adad2d566927136694dc0d7b202d72691d

benma:
They were added after the previous PR taking care of those.

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/9815/commits/a87d02adad2d566927136694dc0d7b202d72691d

laanwj:
> They were added after the previous PR taking care of those.

Good point.

PR #9819, opened 2017-02-21
Remove harmless read of unusued priority estimates

morcos:
It was being discarded immediately anyway


sdaftuar:
utACK bc8fd12

laanwj:
utACK bc8fd12. There's no need to parse this data at all.

PR #9820, opened 2017-02-21
Fix pruning test broken by 2 hour manual prune window

ryanofsky:
Regression happened in 91fb506 Add two hour buffer to manual pruning (#9778)


morcos:
Needs backport to 0.14

ACK 874c736

Thanks for fixing!

PR #7209, opened 2015-12-14
test: don't override BITCOIND and BITCOINCLI if they're set

laanwj:
In rpc-tests.py, don't override BITCOIND and BITCOINCLI if they're already set. Makes it possible to run the tests with either another tree or the GUI.

PR #9823, opened 2017-02-22
qa: Set correct path for binaries in rpc tests

maflcko:
Probably copy paste error introduced in  #9657.

### DIFF
--- a/qa/pull-tester/rpc-tests.py
+++ b/qa/pull-tester/rpc-tests.py
@@ -219,7 +219,7 @@ def run_tests(test_list, src_dir, build_dir, exeext, jobs=1, enable_coverage=Fal
 
     tests_dir = src_dir + '/qa/rpc-tests/'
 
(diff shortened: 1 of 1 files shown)


jnewbery:
tested ACK. Out of interest, was this causing issues?

Minor suggestion: you've used the str.format() method here, and the line immediately below uses a C-style string specifier for the same variable. Perhaps use the same syntax for both? (I don't care which)

MarcoFalke:
> was this causing issues

Yes, when running the tests with a binary of different name, such as
`bitcoin-qt`. See #7209 or try `BITCOIND=bitcoin-qt
./qa/pull-tester/rpc-tests.py`.

MarcoFalke:
re style nit: Personally I think the new style format method looks nicer, but I think everyone is free to use whatever they prefer as long as changes only happen when the code is touched anyway. (No "qa: Unify format methods" pulls please)

PR #9817, opened 2017-02-21
Fix segfault crash when shutdown the GUI in disablewallet mode

jonasschnelli:
Should fix #9816.
Was introduced in https://github.com/bitcoin/bitcoin/pull/9190.

PR #9824, opened 2017-02-22
qa: Check return code when stopping nodes

maflcko:
We had a segfault slip through recently (#9817) because the test framework just ignores the return code.

### DIFF
--- a/qa/rpc-tests/fundrawtransaction.py
+++ b/qa/rpc-tests/fundrawtransaction.py
@@ -469,7 +469,9 @@ class RawTransactionsTest(BitcoinTestFramework):
         # locked wallet test
         self.nodes[1].encryptwallet("test")
         self.nodes.pop(1)
-        stop_nodes(self.nodes)
+        stop_node(self.nodes[0], 0)
+        stop_node(self.nodes[1], 2)
+        stop_node(self.nodes[2], 3)
 
         self.nodes = start_nodes(self.num_nodes, self.options.tmpdir)
         # This test is not meant to test fee estimation and we'd like
--- a/qa/rpc-tests/rpcbind_test.py
+++ b/qa/rpc-tests/rpcbind_test.py
@@ -35,11 +35,9 @@ class RPCBindTest(BitcoinTestFramework):
             base_args += ['-rpcallowip=' + x for x in allow_ips]
         binds = ['-rpcbind='+addr for addr in addresses]
         self.nodes = start_nodes(self.num_nodes, self.options.tmpdir, [base_args + binds], connect_to)
-        try:
-            pid = bitcoind_processes[0].pid
-            assert_equal(set(get_bind_addrs(pid)), set(expected))
-        finally:
-            stop_nodes(self.nodes)
+        pid = bitcoind_processes[0].pid
+        assert_equal(set(get_bind_addrs(pid)), set(expected))
+        stop_nodes(self.nodes)
 
     def run_allowip_test(self, allow_ips, rpchost, rpcport):
         '''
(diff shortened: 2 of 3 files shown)


laanwj:
Yes, makes sense to check that. utACK https://github.com/bitcoin/bitcoin/pull/9824/commits/fac3af7f1c6a1d2ec25ed93201c9dec7bf9bb7bf (also for 0.14)

shyii (approved):
Hey

laanwj:
@shyii I'm not sure what you're trying to do, but please don't abuse the code review system for sending idle messages.

jnewbery:
Concept ACK. We should definitely assert the return code on shutdown.

This causes at least the following two tests to fail:

- fundrawtransaction.py
- rpcbind_test.py

fundrawtransaction is failing because it's calling stop_nodes() on a subset of the running nodes:

```py
        self.nodes.pop(1)
        stop_nodes(self.nodes)
```
The `encryptwallet()` RPC actually stops bitcoind. That means that node 1 is still in the util module's global `bitcoind_processes` dictionary, so a call to `stop_nodes()` hits your assert.

It can be fixed by just calling `stop_node()` for each of the nodes individually.

The failure in rpcbind_test.py is due to this:

```py
        try:
            # connect to node through non-loopback interface
            node = get_rpc_proxy(rpc_url(0, "%s:%d" % (rpchost, rpcport)), 0)
            node.getnetworkinfo()
        finally:
            node = None # make sure connection will be garbage collected and closed
            stop_nodes(self.nodes)
```

I don't think we need to wrap those calls in try/finallys since the `BitcoinTestFramework.main()` method takes care of cleaning up the nodes for us.

I've pushed fixes for both of those failures to https://github.com/jnewbery/bitcoin/commit/cbebbcf010e2f7986e349b1a21121b6cdaf182c2. Feel free to squash into your commit if you're happy with the changes.

Generally, tracking `bitcoin_processes` as a global dictionary in util.py smells (any global variables in util.py smells). The nodes should be members of TestFramework, and it should be the test framework's responsibility to clean them up (and assert return codes) at the end of the test run.

And one question on code style: I know that assert_equal() is used all over the qa test directory. Why? Why is
`assert_equal(ret, 0)`
preferred over
`assert ret == 0, "unexpected return code %s" % ret`?

Why is
`assert_equal(len(bitcoind_processes.values()), 0)`
preferred over
`assert not bitcoind_processes, "bitcoind processes still running: %s" % str(bitcoind_processes)`?

In generally, `assert` seems to be treated as a function throughout the qa directory instead of a statement. It would be nice if we could change that.

laanwj:
> In generally, assert seems to be treated as a function throughout the qa directory instead of a statement. > It would be nice if we could change that.

I think the idea on the longer run is to move away from using bare assert everywhere and to something that resembles a real testing framework (maybe `unittest`). Testing frameworks provide [helper functions](https://docs.python.org/3/library/unittest.html#assert-methods) such as `assert_equal` because those can automatically print the expected and passed-in value, without having to provide a custom message everywhere.

MarcoFalke:
Thanks for the fixups, I will include them in this pull. I will also
adjust the code style of the second assertion.

In regard to your question:

assert_equal(a, b) is the lazy way to assert equality but still have a
somewhat useful assert message which includes both values that are
unequal.

assert as a statement should only be used with a single variable of
type boolean as expression or with an additional string that provides
further debug infos (such as the ones you mentioned above). I assume
no one uses the hand crafted strings because assertions are expected
to fire off basically never and the additional effort required to put
into crafting the string is not considered worthy. (In the rare case a
assertion fires, a developer is usually required to jump into the code
at that location anyway...)

jnewbery:
@laanwj @MarcoFalke thanks. Makes sense.

> assert as a statement should only be used with a single variable of
type boolean as expression or with an additional string that provides
further debug infos

Sounds sensible. We're quite a long way off that though!
```
→ cat qa/rpc-tests/*py | grep -c "assert("
364
```

PR #9828, opened 2017-02-22
Avoid -Wshadow warnings in wallet_tests

ryanofsky:
Warnings introduced by #9773 and reported by @paveljanik in #9773 and #9827.


paveljanik:
Isn't the other way shorter (ie. changing only inside blocks)?

But anyway, this fixes the issue, yes. ACK https://github.com/bitcoin/bitcoin/pull/9828/commits/09fe346a4fb9c2b6c4f1bc1318494885708226cd

ryanofsky:
The other way is shorter but would conflict with other pending prs. Also I think it is better for global variables to use longer names than locals.

paveljanik:
@ryanofsky Yes, I agree.

dcousens (approved):

MarcoFalke:
utACK 09fe346, I think we should give #8808 a try.

PR #9831, opened 2017-02-23
build: force a c++ standard to be specified

theuni:
@laanwj suggested this here https://github.com/bitcoin/bitcoin/pull/9753#issuecomment-281651607.

Newer compilers may switch to newer standards by default. For example, gcc6 uses std=gnu++14 by default.

Note that this makes testing c++14 impractical. Until we're ready to investigate that transition, I don't think that's a problem.

I'll attempt to upstream the macro changes post-merge.

### DIFF
--- a/build-aux/m4/ax_cxx_compile_stdcxx.m4
+++ b/build-aux/m4/ax_cxx_compile_stdcxx.m4
@@ -57,8 +57,14 @@ AC_DEFUN([AX_CXX_COMPILE_STDCXX], [dnl
         [$3], [mandatory], [ax_cxx_compile_cxx$1_required=true],
         [$3], [optional], [ax_cxx_compile_cxx$1_required=false],
         [m4_fatal([invalid third argument `$3' to AX_CXX_COMPILE_STDCXX])])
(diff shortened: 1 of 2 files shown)


laanwj:
Yes - the rationale for this is that we want to avoid c++14 use to creep into the source code before we've decided we want c++14. I think it's too soon for that. It'd be nice to not bump our minimum clang/g++ requirements for a while.

sipa:
Concept ACK

laanwj:
Tested ACK 9829c54: `-std=c++11` is indeed being passed to compiles after this patch.

paveljanik:
And it is still passed on systems where it is needed. The only difference here is a configure output missing the following line:
```
checking whether g++ supports C++11 features by default... no
```

ACK https://github.com/bitcoin/bitcoin/pull/9831/commits/9829c54de2725037ee0702306cbaa99fc9aa1826

laanwj:
> `checking whether g++ supports C++11 features by default... no`

I think that's to be expected, because we no longer care about the default, we always specify it.

MarcoFalke:
Concept ACK

theuni:
@paveljanik Yes, that's the intended change.

PR #9832, opened 2017-02-23
[qa] assert_start_raises_init_error

NicolasDorier:
assert_start_raises_init_error make it possible to test expected error condition when bitcoind is starting.

ping @MarcoFalke and @jonasschnelli mentioned the need on https://github.com/bitcoin/bitcoin/pull/9662 and https://github.com/bitcoin/bitcoin/pull/9728

### DIFF
--- a/qa/rpc-tests/test_framework/util.py
+++ b/qa/rpc-tests/test_framework/util.py
@@ -19,6 +19,7 @@ import http.client
 import random
 import shutil
 import subprocess
+import tempfile
 import time
 import re
 import errno
@@ -329,7 +330,7 @@ def _rpchost_to_args(rpchost):
         rv += ['-rpcport=' + rpcport]
     return rv
 
-def start_node(i, dirname, extra_args=None, rpchost=None, timewait=None, binary=None):
+def start_node(i, dirname, extra_args=None, rpchost=None, timewait=None, binary=None, stderr=None):
     """
     Start a bitcoind and return RPC connection to it
     """
@@ -338,7 +339,7 @@ def start_node(i, dirname, extra_args=None, rpchost=None, timewait=None, binary=
         binary = os.getenv("BITCOIND", "bitcoind")
     args = [ binary, "-datadir="+datadir, "-server", "-keypool=1", "-discover=0", "-rest", "-mocktime="+str(get_mocktime()) ]
     if extra_args is not None: args.extend(extra_args)
-    bitcoind_processes[i] = subprocess.Popen(args)
+    bitcoind_processes[i] = subprocess.Popen(args, stderr=stderr)
     if os.getenv("PYTHON_DEBUG", ""):
         print("start_node: bitcoind started, waiting for RPC to come up")
     url = rpc_url(i, rpchost)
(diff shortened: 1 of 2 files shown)


laanwj:
Makes sense.
utACK https://github.com/bitcoin/bitcoin/pull/9832/commits/5737be099f83dfb367a342c36f7f0ae5ea31bc29

jonasschnelli:
Thanks.
utACK 5737be099f83dfb367a342c36f7f0ae5ea31bc29.
This is also required for #9662.

qa/rpc-tests/test_framework/util.py
```diff
@@ -338,7 +339,10 @@ def start_node(i, dirname, extra_args=None, rpchost=None, timewait=None, binary=
         binary = os.getenv("BITCOIND", "bitcoind")
     args = [ binary, "-datadir="+datadir, "-server", "-keypool=1", "-discover=0", "-rest", "-mocktime="+str(get_mocktime()) ]
     if extra_args is not None: args.extend(extra_args)
-    bitcoind_processes[i] = subprocess.Popen(args)
+    if stderr is not None:
```
jnewbery: I don't think this if/else construction is needed. subprocess.Popen's stderr argument defaults to None, so you can just call `bitcoind_processes[i] = subprocess.Popen(args, stderr=stderr)`

qa/rpc-tests/test_framework/util.py
```diff
@@ -352,6 +356,28 @@ def start_node(i, dirname, extra_args=None, rpchost=None, timewait=None, binary=
...
+    log_stderr = tempfile.SpooledTemporaryFile(max_size=2**16)
+    try:
+        node = start_node(i, dirname, extra_args, stderr=log_stderr)
+        stop_node(node, i)
+        if expected_msg is None:
+            assert_msg = "bitcoind should have exited with an error"
+        else:
+            assert_msg = "bitcoind should have exited with expected error " + expected_msg
+        raise AssertionError(assert_msg)
+    except AssertionError as e:
```
jnewbery: yuck! raising an exception in expectation of catching it in the very next line.

I think try/except/else is a much cleaner construction, as in:

```python
try:
    thing_we_expect_to_fail()
except ExceptionType as e:
    do_some_asserts_on_the_exception_raised()
else:
    # oops, we were expecting an exception
    assert False, "we were expecting an exception!"
```

qa/rpc-tests/test_framework/util.py
```diff
@@ -352,6 +356,28 @@ def start_node(i, dirname, extra_args=None, rpchost=None, timewait=None, binary=
 
     return proxy
 
+def assert_start_raises_init_error(i, dirname, extra_args=None, expected_msg=None):
+    log_stderr = tempfile.SpooledTemporaryFile(max_size=2**16)
```
jnewbery: Consider opening the file using the `with log_stderr as...` form. That means the file will be closed automatically (even if an exception is raised on the way), and you won't have to close it yourself with the `finally` branch at the bottom.

qa/rpc-tests/test_framework/util.py
```diff
@@ -352,6 +356,28 @@ def start_node(i, dirname, extra_args=None, rpchost=None, timewait=None, binary=
...
+        stop_node(node, i)
+        if expected_msg is None:
+            assert_msg = "bitcoind should have exited with an error"
+        else:
+            assert_msg = "bitcoind should have exited with expected error " + expected_msg
+        raise AssertionError(assert_msg)
+    except AssertionError as e:
+        raise
+    except Exception as e:
+        assert('bitcoind exited' in str(e)) #node must have shutdown
```
jnewbery: minor nit: assert is a statement, not a function. This should be:

assert 'bitcoind exited' in str(e)

NicolasDorier:
nits of @jnewbery addressed

MarcoFalke:
Please squash.

NicolasDorier:
squashed @MarcoFalke

qa/rpc-tests/test_framework/util.py
```diff
@@ -352,6 +353,25 @@ def start_node(i, dirname, extra_args=None, rpchost=None, timewait=None, binary=
 
     return proxy
 
+def assert_start_raises_init_error(i, dirname, extra_args=None, expected_msg=None):
+    with tempfile.SpooledTemporaryFile(max_size=2**16) as log_stderr:
+        try:
+            node = start_node(i, dirname, extra_args, stderr=log_stderr)
+            stop_node(node, i)
```
MarcoFalke: Nit: No need to stop the node here. This is done by the test framework.

MarcoFalke:
Post merge utACK 025dec0e5bf001ba297f7430affe4098627ea5ce

qa/rpc-tests/test_framework/util.py
NicolasDorier: It does need to be stopped, or this function would not be side effect free.

MarcoFalke: Right this line is only executed in exceptional circumstances, in which case the whole frameworks shuts down.

PR #9010, opened 2016-10-25
Split up AppInit2 into multiple phases, daemonize after datadir lock errors

laanwj:
Split up AppInit2 into multiple steps: This allows doing some of the steps before daemonization and some later.

To avoid the issue described in https://github.com/bitcoin/bitcoin/issues/9009#issuecomment-255988169:
- Before daemonization, just probe the data directory lock and print an early error message if possible.
- After daemonization get the data directory lock again and hold on to it until exit. This creates a slight window for a race condition to happen, however this condition is harmless: it will at most make the process exit without printing a message to console as it will be detected after daemonization.

Solves #9009:

```
$ src/bitcoind -testnet -daemon
Bitcoin server starting
$ src/bitcoind -testnet -daemon
Error: Cannot obtain a lock on data directory /home/orion/.bitcoin/testnet3. Bitcoin Core is probably already running.
```

As for the AppInit2 split: This is something I've been wanting to do for a while. Note that, to reduce risk of regressions, this is just mechanic code movement and doesn't change ordering, besides the `daemon()` invocation.

Some smarter things could be done such as:
- Move basic context initialization from `AppInitBasicSetup()` to `SetupEnvironment()` where possible, especially things such as the windows DEP activation
- Unify `InitParameterInteraction()` and `AppInitParameterInteraction()` if/where possible. I'm not sure why this is done in two steps with just `AppInitBasicSetup()` in between right now.
- Do something smarter than `// Variables internal to initialization process only` (e.g. an initialization context structure). Some variables there may be completely avoided by moving the code to extract the arguments.
- Splitting up further steps, this will allow factoring out `ENABLE_WALLET` sections from `libbitcoin_server.a` (which contains init.cpp) and remove the circular dependencies (#7965)
- It is now possible to do unit testing on our initialization steps in isolation.

However I do not intend to do these things in this pull.

PR #9833, opened 2017-02-23
Trivial: fix comments referencing AppInit2

benma:
It was refactored into multiple functions in
0cc8b6bc44bea29e24fa4e13d8a9bbe4f1483680.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -118,10 +118,6 @@ static const char* FEE_ESTIMATES_FILENAME="fee_estimates.dat";
 // threads that should only be stopped after the main network-processing
 // threads have exited.
 //
-// Note that if running -daemon the parent process returns from AppInit2
-// before adding any threads to the threadGroup, so .join_all() returns
-// immediately and the parent exits from main().
-//
 // Shutdown for Qt is very similar, only it uses a QTimer to detect
 // fRequestShutdown getting set, and then does the normal Qt
 // shutdown thing.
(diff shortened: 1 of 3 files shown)


src/init.cpp
```diff
@@ -188,7 +188,7 @@ void Shutdown()
     if (!lockShutdown)
         return;
 
-    /// Note: Shutdown() must be able to handle cases in which AppInit2() failed part of the way,
+    /// Note: Shutdown() must be able to handle cases in which AppInitMain() failed part of the way,
```
laanwj: let's just remove mention of a function here and call it "initialization". It may be any of the initialization-related functions that failed.

benma: Done (also in bitcoin.cpp).

For my information: how should I address feedback: with new commits for easier review, or with amending? I amended this time.

laanwj: For documentation/string commits like this, amending is best IMO.
For more involved commits, it can make sense to keep a history, at least until merge.

src/init.cpp
```diff
@@ -118,7 +118,7 @@ static const char* FEE_ESTIMATES_FILENAME="fee_estimates.dat";
 // threads that should only be stopped after the main network-processing
 // threads have exited.
 //
-// Note that if running -daemon the parent process returns from AppInit2
+// Note that if running -daemon the parent process returns from AppInitMain()
```
laanwj: This is no longer true, by the way. Since #9010, the -daemon handling is in bitcoind.cpp, not in init.cpp.

benma: Thanks. The comment seems to not be true anyway, so I remove it altogether.

sipa:
utACK ef9f495e931fad1e356757d6a40092596aa0205c

PR #9834, opened 2017-02-23
qt: clean up initialize/shutdown signals

benma:
- Change initializeResult(int) to initializeResult(bool) to avoid
   implicit type conversion.
 - Use EXIT_FAILURE and EXIT_SUCCESS instead of magic numbers.
 - Remove the argument from shutdownResult(int); it was called with a
   constant argument.

### DIFF
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -178,8 +178,8 @@ public Q_SLOTS:
     void shutdown();
 
 Q_SIGNALS:
-    void initializeResult(int retval);
-    void shutdownResult(int retval);
+    void initializeResult(bool success);
+    void shutdownResult();
     void runawayException(const QString &message);
 
 private:
@@ -223,8 +223,8 @@ public:
     WId getMainWinId() const;
 
 public Q_SLOTS:
-    void initializeResult(int retval);
-    void shutdownResult(int retval);
+    void initializeResult(bool success);
+    void shutdownResult();
     /// Handle runaway exceptions. Shows a message box with the problem and quits the program.
     void handleRunawayException(const QString &message);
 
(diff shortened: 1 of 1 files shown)


laanwj:
> it was called with a constant argument.

That seems like a bug. It should be called with 0 in case of an error. Though I guess shutdown can't really fail (without crashing).

benma:
It calls a different signal in that case: `runawayException`, which calls exit(EXIT_FAILURE) instead of QT's quit().

laanwj:
Yes, indeed.

benma:
@jonasschnelli could you please review this?

src/qt/bitcoin.cpp
```diff
@@ -302,7 +302,7 @@ void BitcoinCore::shutdown()
         threadGroup.join_all();
         Shutdown();
         qDebug() << __func__ << ": Shutdown finished";
-        Q_EMIT shutdownResult(1);
```
jonasschnelli: Agree with @laanwj: This was a bug.

jonasschnelli:
Tested ACK 5b528d746c841aa868a305f9c1a586f539b531cb

src/qt/bitcoin.cpp
benma: Was a bug, or is a bug? Because it is still not called with (0), and I am not sure under what circumstances it should be, seeing that the runaway exception signal is called when shutdown fails.

jonasschnelli: The int return value was only use over qDebug which we can argue what it should have been there.
But this PRs solution (remove the shutdown result value) seems to make most sense.

benma:
Thank you for the review and merge! :dancing_men:

PR #9845, opened 2017-02-23
RPC: cleanups in rpc/server

jtimon:
There's no reason to declare the following functions in rpc/server.h when they're defined in wallet/rpcwallet.cpp. They can be declared in wallet/rpcwallet.h instead:

HelpRequiringPassphrase
EnsureWalletIsAvailable
EnsureWalletIsUnlocked

Also, there's no reason to use extern for functions declared in rpc/server.h. It is really never needed (see http://stackoverflow.com/questions/11712707/extern-functions-in-c-vs-c/11712820#11712820 ), but it seems more strange when those functions are also defined in rpc/server.cpp (in the same module).

### DIFF
--- a/src/rpc/server.h
+++ b/src/rpc/server.h
@@ -185,16 +185,16 @@ extern CRPCTable tableRPC;
  * Utilities: convert hex-encoded Values
  * (throws error if not hex).
  */
-extern uint256 ParseHashV(const UniValue& v, std::string strName);
-extern uint256 ParseHashO(const UniValue& o, std::string strKey);
-extern std::vector<unsigned char> ParseHexV(const UniValue& v, std::string strName);
-extern std::vector<unsigned char> ParseHexO(const UniValue& o, std::string strKey);
+uint256 ParseHashV(const UniValue& v, std::string strName);
+uint256 ParseHashO(const UniValue& o, std::string strKey);
+std::vector<unsigned char> ParseHexV(const UniValue& v, std::string strName);
+std::vector<unsigned char> ParseHexO(const UniValue& o, std::string strKey);
 
-extern CAmount AmountFromValue(const UniValue& value);
-extern UniValue ValueFromAmount(const CAmount& amount);
-extern double GetDifficulty(const CBlockIndex* blockindex = NULL);
-extern std::string HelpExampleCli(const std::string& methodname, const std::string& args);
-extern std::string HelpExampleRpc(const std::string& methodname, const std::string& args);
+CAmount AmountFromValue(const UniValue& value);
+UniValue ValueFromAmount(const CAmount& amount);
+double GetDifficulty(const CBlockIndex* blockindex = NULL);
+std::string HelpExampleCli(const std::string& methodname, const std::string& args);
+std::string HelpExampleRpc(const std::string& methodname, const std::string& args);
 
 bool StartRPC();
 void InterruptRPC();


sipa:
utACK

dcousens (approved):

kallewoof:
utACK 073613b

jonasschnelli (approved):
utACK 073613b36a8780dd720e2a17f1314980173e200d

paveljanik:
utACK 073613b

laanwj:
Good idea, rpcserver is supposed to be bitcoin-agnostic, those methods don't belong there.

jtimon:
There's also GetDifficulty (declared in rpc/server.h, defined in rpc/blockchain.cpp, used in several rpc places) which isn't taken care of here. I wasn't sure what to do with it.

NicolasDorier:
utACK 073613b36a8780dd720e2a17f1314980173e200d

MarcoFalke:
utACK 073613b36a8780dd720e2a17f1314980173e200d

src/rpc/server.h
```diff
@@ -185,20 +185,17 @@ extern CRPCTable tableRPC;
...
-extern uint256 ParseHashV(const UniValue& v, std::string strName);
-extern uint256 ParseHashO(const UniValue& o, std::string strKey);
-extern std::vector<unsigned char> ParseHexV(const UniValue& v, std::string strName);
-extern std::vector<unsigned char> ParseHexO(const UniValue& o, std::string strKey);
+uint256 ParseHashV(const UniValue& v, std::string strName);
+uint256 ParseHashO(const UniValue& o, std::string strKey);
+std::vector<unsigned char> ParseHexV(const UniValue& v, std::string strName);
+std::vector<unsigned char> ParseHexO(const UniValue& o, std::string strKey);
 
 extern int64_t nWalletUnlockTime;
```
jnewbery: Can you also move `nWalletUnlockTime` to rpcwallet.h as an extern declaration? You'll need to `#include wallet/rpcwallet.h` in rpc/misc.cpp, protected by `#ifdef ENABLE_WALLET`

jnewbery (approved):
tested ACK. Looks good.

> There's also GetDifficulty (declared in rpc/server.h, defined in rpc/blockchain.cpp, used in several rpc places) which isn't taken care of here

Could that become a member function of CBlockIndex in chain.cpp?

src/rpc/server.h
laanwj: Good catch. Agreed.

laanwj:
Needs rebase (and last nit-fix)

jtimon:
Rebased.
Rebasing makes the first commit unnecessary after #8775 "a4356328 Move wallet RPC declarations to rpcwallet.h" . It seems nWalletUnlockTime was moved in #8775 too (see "2e518e31 Move nWalletUnlockTime to CWallet::nRelockTime, and name timed task unique per CWallet" ).
This leaves the PR to a single commit removing the unnecessary 'extern' in functions, trivial to review.

MarcoFalke:
> removing the unnecessary 'extern' in functions

I think there are other functions with 'extern'. You might as well do it
for the whole code base.

jtimon:
@MarcoFalke I guess at that point I should create another PR since the title "RPC: cleanups in rpc/server" wouldn't fit anymore. I'm not planning to do that in the short term, but you are right. I would happily ACK such a PR to replace this one if anybody does it. I was just working on rpcwallet and rpc/server and saw easy things to fix without knowing about #8775 . I may end up doing it myself if nobody else jumps at it, but for now I'll just leave this as is.

laanwj:
> I think there are other functions with 'extern'. You might as well do it for the whole code base.

For whoever does that: do add instruction in `developer-notes.md` on when to use `extern` and when not to. It seems safe to add extern to these cases, just redundant, so I'd see this as quite low priority.

src/rpc/server.h
```diff
@@ -185,16 +185,16 @@ extern CRPCTable tableRPC;
...
-extern std::string HelpExampleCli(const std::string& methodname, const std::string& args);
-extern std::string HelpExampleRpc(const std::string& methodname, const std::string& args);
+uint256 ParseHashV(const UniValue& v, std::string strName);
+uint256 ParseHashO(const UniValue& o, std::string strKey);
+std::vector<unsigned char> ParseHexV(const UniValue& v, std::string strName);
+std::vector<unsigned char> ParseHexO(const UniValue& o, std::string strKey);
+
+CAmount AmountFromValue(const UniValue& value);
+UniValue ValueFromAmount(const CAmount& amount);
+double GetDifficulty(const CBlockIndex* blockindex = NULL);
```
laanwj: Why is GetDifficulty still here? :-)

jtimon:
extern should never be used for functions, but I'm not sure that is worth to add "please don't add extern to functions because it doesn't do anything" to the developer notes, at least not more than "please don't add variables that aren't used".
Agreed removing extern is low priority, as said, the most interesting parts of this PR were already done by #8775

Should we merge this anyway (since it's trivial, has been reviewed and it's not disruptive) or should I close it?

jtimon:
Closing, it seems at this point would be preferable to remove extern from all functions from the project at once (I bet there's not that many, but I didn't looked yet) as @MarcoFalke suggests.
In some way, all utACKs for this were just partial acks for merged #8775

laanwj:
I think removing GetDifficulty from server.h would have made a lot of sense here, along with the other cleanups as those lines were touched anyway, but indeed it's very low priority.

jtimon:
Well, as said I wasn't sure what to do with GetDifficulty, so this PR never moved it.

laanwj:
No need to be clever here, if it's defined in `src/rpc/blockchain.cpp` the interface should be declared in `src/rpc/blockchain.h`. Will file a PR, see #10095.

Issue #9825, opened 2017-02-22
Intermittent FAIL: test/test_bitcoin in Travis


laanwj:
This seems to happen at least on 0.14:
```
FAIL: test/test_bitcoin
=======================
test_bitcoin: /home/travis/build/bitcoin/bitcoin/depends/x86_64-unknown-linux-gnu/share/../include/boost/thread/pthread/recursive_mutex.hpp:113: void boost::recursive_mutex::lock(): Assertion `!pthread_mutex_lock(&m)' failed.
Running 228 test cases...
unknown location(0): fatal error: in "rpc_tests/rpc_rawparams": signal: SIGABRT (application abort requested)
test/rpc_tests.cpp(29): last checkpoint
test_bitcoin: key.cpp:300: void ECC_Start(): Assertion `secp256k1_context_sign == __null' failed.
unknown location(0): fatal error: in "rpc_tests/rpc_format_monetary_values": signal: SIGABRT (application abort requested)
...
*** 37 failures are detected in the test module "Bitcoin Test Suite"
test_bitcoin: /home/travis/build/bitcoin/bitcoin/depends/x86_64-unknown-linux-gnu/share/../include/boost/thread/pthread/condition_variable_fwd.hpp:102: boost::condition_variable::~condition_variable(): Assertion `!ret' failed.
```

Somehow it fails to lock a mutex even before the tests start, and all tests after that fail. I cannot reproduce it locally.

laanwj:
Another one: https://travis-ci.org/bitcoin/bitcoin/jobs/204158785
```
test_bitcoin: tpp.c:62: __pthread_tpp_change_priority: Assertion `new_prio == -1 || (new_prio >= __sched_fifo_min_prio && new_prio <= __sched_fifo_max_prio)' failed.
Running 228 test cases...
unknown location(0): fatal error: in "rpc_tests/rpc_rawparams": signal: SIGABRT (application abort requested)
test/rpc_tests.cpp(29): last checkpoint
...
```

laanwj:
Another one on 0.14: https://travis-ci.org/bitcoin/bitcoin/jobs/204167533
```
FAIL: test/test_bitcoin
=======================
Running 228 test cases...
unknown location(0): fatal error: in "rpc_tests/rpc_rawsign": signal: illegal operand; address of failing instruction: 0x2b73cbf4cf3b
test/rpc_tests.cpp(29): last checkpoint
```

laanwj:
[...]

PR #9851, opened 2017-02-24
[do not merge] travis gdb parachute for #9825

laanwj:
Hopefully this can give us a useful traceback.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,20 +21,12 @@ env:
     - PYTHON_DEBUG=1
     - WINEDEBUG=fixme-all
   matrix:
-# ARM
-    - HOST=arm-linux-gnueabihf PACKAGES="g++-arm-linux-gnueabihf" DEP_OPTS="NO_QT=1" CHECK_DOC=1 GOAL="install" BITCOIN_CONFIG="--enable-glibc-back-compat --enable-reduce-exports"
-# Win32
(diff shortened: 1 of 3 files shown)


paveljanik:
Looks like gdb is missing

laanwj:
yep, added that now

laanwj:
Hm, to make this iterate faster, could change the builds to only build test_bitcoin. Let's just wait for this to finish first to see if anything else needs changing.

laanwj:
Ugh the no-wallet build was rebuilding all dependencies (including Qt). Cancelled that. Hopefully it won't do that every time.

laanwj:
Looks like I caught it in the act now:
```
test_bitcoin: /home/travis/build/bitcoin/bitcoin/depends/i686-pc-linux-gnu/share/../include/boost/thread/pthread/recursive_mutex.hpp:113: void boost::recursive_mutex::lock(): Assertion `!pthread_mutex_lock(&m)' failed.
Program received signal SIGABRT, Aborted.
Thread 41 (Thread 0xec3a7b40 (LWP 21516)):
Python Exception <class 'gdb.MemoryError'> Cannot access memory at address 0x24: 
#0  0xf7fdacd9 in ?? ()
Error while running hook_stop:
Cannot access memory at address 0x24
0xf7fdacd9 in ?? ()
```
Unfortunately, the traceback is everything but useful.

laanwj:
I'm slightly optimistic that this has been solved by #9875. If the bug rears its head again I'll reopen.

ryanofsky:
Maybe another thing to add would be --loglevel=all to test_bitcoin
http://www.boost.org/doc/libs/1_63_0/libs/test/doc/html/boost_test/utf_reference/rt_param_reference/log_level.html

PR #9842, opened 2017-02-23
Fix RPC failure testing (continuation of #9707)

jnewbery:
Commit 9db8eecac1c713c760c0217b6acb7455c657fa8b improved the
assert_raises_jsonrpc() function for better testing of RPC failure
modes. This commit continues the job by removing remaining broken
try-except RPC testing from individual test cases and replacing it
with calls to assert_raises_jsonrpc().

Note that this does not fully complete the job. This commit replaces the try/except code with assert_raises_jsonrpc() only for those test cases where no changes to bitcoind were required (ie the test cases weren't necessarily fully testing the behaviour, but the behaviour was correct anyway). For any test cases where fixing the test code has revealed bugs or inconsistencies, I'll raise a separate issue/PR (eg #9738, #9714, #9841, #9710). There are a few more of those to come.

PR #9853, opened 2017-02-24
Fix error codes from various RPCs

jnewbery:
(Completes #9707 and #9842)

I've changed the qa test code to properly assert error codes and messages for failed RPC calls. That's revealed a few RPCs which were returning incorrect or misleading error codes. This PR fixes all of those cases.

It also adds commenting around the definition of the JSON RPC error constants to warn against using RPC_INVALID_REQUEST and RPC_METHOD_NOT_FOUND for application-layer errors.

### DIFF
--- a/qa/rpc-tests/bumpfee.py
+++ b/qa/rpc-tests/bumpfee.py
@@ -128,7 +128,7 @@ def test_segwit_bumpfee_succeeds(rbf_node, dest_address):
 def test_nonrbf_bumpfee_fails(peer_node, dest_address):
     # cannot replace a non RBF transaction (from node which did not enable RBF)
     not_rbfid = create_fund_sign_send(peer_node, {dest_address: 0.00090000})
-    assert_raises_message(JSONRPCException, "not BIP 125 replaceable", peer_node.bumpfee, not_rbfid)
+    assert_raises_jsonrpc(-4, "not BIP 125 replaceable", peer_node.bumpfee, not_rbfid)
 
 
 def test_notmine_bumpfee_fails(rbf_node, peer_node, dest_address):
@@ -148,7 +148,7 @@ def test_notmine_bumpfee_fails(rbf_node, peer_node, dest_address):
     signedtx = rbf_node.signrawtransaction(rawtx)
     signedtx = peer_node.signrawtransaction(signedtx["hex"])
     rbfid = rbf_node.sendrawtransaction(signedtx["hex"])
-    assert_raises_message(JSONRPCException, "Transaction contains inputs that don't belong to this wallet",
+    assert_raises_jsonrpc(-4, "Transaction contains inputs that don't belong to this wallet",
                           rbf_node.bumpfee, rbfid)
 
 
@@ -159,7 +159,7 @@ def test_bumpfee_with_descendant_fails(rbf_node, rbf_node_address, dest_address)
     tx = rbf_node.createrawtransaction([{"txid": parent_id, "vout": 0}], {dest_address: 0.00020000})
     tx = rbf_node.signrawtransaction(tx)
     txid = rbf_node.sendrawtransaction(tx["hex"])
-    assert_raises_message(JSONRPCException, "Transaction has descendants in the wallet", rbf_node.bumpfee, parent_id)
+    assert_raises_jsonrpc(-8, "Transaction has descendants in the wallet", rbf_node.bumpfee, parent_id)
 
 
 def test_small_output_fails(rbf_node, dest_address):
@@ -174,7 +174,7 @@ def test_small_output_fails(rbf_node, dest_address):
                             Decimal("0.00100000"),
                             {dest_address: 0.00080000,
                              get_change_address(rbf_node): Decimal("0.00010000")})
-    assert_raises_message(JSONRPCException, "Change output is too small", rbf_node.bumpfee, rbfid, {"totalFee": 20001})
+    assert_raises_jsonrpc(-4, "Change output is too small", rbf_node.bumpfee, rbfid, {"totalFee": 20001})
 
 
 def test_dust_to_fee(rbf_node, dest_address):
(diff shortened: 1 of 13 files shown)


qa/rpc-tests/bumpfee.py
```diff
@@ -128,7 +128,7 @@ def test_segwit_bumpfee_succeeds(rbf_node, dest_address):
 def test_nonrbf_bumpfee_fails(peer_node, dest_address):
     # cannot replace a non RBF transaction (from node which did not enable RBF)
     not_rbfid = create_fund_sign_send(peer_node, {dest_address: 0.00090000})
-    assert_raises_message(JSONRPCException, "not BIP 125 replaceable", peer_node.bumpfee, not_rbfid)
+    assert_raises_jsonrpc(-4, "not BIP 125 replaceable", peer_node.bumpfee, not_rbfid)
```
ryanofsky: Is there a plan to switch from integer literals to constants in a later PR? Seems like it would be nice to say `RPC_WALLET_ERROR` instead of `-4`.

laanwj: Yes that would be nice, but is orthogonal to this change, so if you want to do that it'd indeed be a different PR.

src/wallet/rpcwallet.cpp
```diff
@@ -2753,33 +2753,33 @@ UniValue bumpfee(const JSONRPCRequest& request)
     CWalletTx& wtx = pwalletMain->mapWallet[hash];
 
     if (pwalletMain->HasWalletSpend(hash)) {
-        throw JSONRPCError(RPC_MISC_ERROR, "Transaction has descendants in the wallet");
+        throw JSONRPCError(RPC_WALLET_ERROR, "Transaction has descendants in the wallet");
```
ryanofsky: In cases where bumpfee is called erroneously with a transaction that shouldn't be bumped, invalid parameter seems more apt than wallet error.

Probably the most important thing you want to know when you get an error message from an API is whether the error is caused by something that you control and need to fix, or whether it's caused by something out of your control (like an implementation bug, or corrupt data, or a temporary failure). RPC_INVALID_PARAMETER clearly tells you that you need to fix something, while RPC_WALLET_ERROR is vague but suggests that something's wrong with the wallet.

src/wallet/rpcdump.cpp
```diff
@@ -338,11 +338,11 @@ UniValue removeprunedfunds(const JSONRPCRequest& request)
     vector<uint256> vHashOut;
 
     if(pwalletMain->ZapSelectTx(vHash, vHashOut) != DB_LOAD_OK) {
-        throw JSONRPCError(RPC_INTERNAL_ERROR, "Could not properly delete the transaction.");
+        throw JSONRPCError(RPC_WALLET_ERROR, "Could not properly delete the transaction.");
     }
 
     if(vHashOut.empty()) {
-        throw JSONRPCError(RPC_INTERNAL_ERROR, "Transaction does not exist in wallet.");
+        throw JSONRPCError(RPC_WALLET_ERROR, "Transaction does not exist in wallet.");
```
ryanofsky: Maybe invalid parameter here as well.

src/wallet/rpcdump.cpp
```diff
@@ -338,11 +338,11 @@ UniValue removeprunedfunds(const JSONRPCRequest& request)
     vector<uint256> vHashOut;
 
     if(pwalletMain->ZapSelectTx(vHash, vHashOut) != DB_LOAD_OK) {
-        throw JSONRPCError(RPC_INTERNAL_ERROR, "Could not properly delete the transaction.");
+        throw JSONRPCError(RPC_WALLET_ERROR, "Could not properly delete the transaction.");
```
ryanofsky: ZapSelect returns a DBErrors enum value which seems like it could be mapped to an RPC error (would also be nice to include in the error string).

src/rpc/blockchain.cpp
```diff
@@ -829,7 +834,7 @@ UniValue pruneblockchain(const JSONRPCRequest& request)
             + HelpExampleRpc("pruneblockchain", "1000"));
 
     if (!fPruneMode)
-        throw JSONRPCError(RPC_METHOD_NOT_FOUND, "Cannot prune blocks because node is not in prune mode.");
+        throw JSONRPCError(RPC_MISC_ERROR, "Cannot prune blocks because node is not in prune mode.");
```
ryanofsky: Some of these MISC_ERROR cases make it seem like we could benefit from an new RPC_INVALID_STATE value. (Blockchain too short is another case below.)

ryanofsky (commented):
utACK b34faf156d49705faa335641cfbfc0efd60bb781, definitely an improvement.

qa/rpc-tests/bumpfee.py
jnewbery: yep, I'll put that in the todo list. I'm working on a primitives.py for bitcoin messages, datastructures and constants. Seems like a good place for the RPC error constants to live.

src/wallet/rpcwallet.cpp
jnewbery: > Probably the most important thing you want to know when you get an error message from an API is whether the error is caused by something that you control and need to fix, or whether it's caused by something out of your control

Nicely put. I'll update this to RPC_INVALID_PARAMETER

src/wallet/rpcdump.cpp
jnewbery: I agree we could do better, but let's leave that for a future PR.

I think this is strictly an improvement because currently the API is returning RPC_INTERNAL_ERROR even in cases where there's no internal error.

src/wallet/rpcdump.cpp
jnewbery: ACK

src/rpc/blockchain.cpp
jnewbery: Sounds sensible, but again let's leave that for a future PR.

laanwj: Yes, there is no reason why we couldn't add new status values. There's no need to make everything fit in the arbitrary bag of codes that exists now, which has been there unchanged for a long while.

jnewbery:
Squashed @ryanofsky's nits and rebased.
Old tip: https://github.com/bitcoin/bitcoin/pull/9853/commits/5158a1e2a2587f9eba04d0f2a81744a9ecbe766d / https://github.com/jnewbery/bitcoin/tree/pr9853.1
New tip: https://github.com/bitcoin/bitcoin/pull/9853/commits/3ff485e32f7a1d068e799d323d3531ede22873f1 / https://github.com/jnewbery/bitcoin/tree/pr9853.2

laanwj:
utACK 3ff485e

MarcoFalke:
Concept ACK.

Style Nit: In blockchain.cpp you are missing braces around the
multiline if block

Also, I was wondering if this requires a mention in the release notes...

laanwj:
> Also, I was wondering if this requires a mention in the release notes...

It does.

ryanofsky:
utACK 3ff485e32f7a1d068e799d323d3531ede22873f1

Confirmed only minor changes since previous ACK.

jnewbery:
Release notes updated. Otherwise no change.

PR #6380, opened 2015-07-06
rpc: Accept strings in AmountFromValue

laanwj:
Accept strings containing decimal values, in addition to bare values.

note: untested

PR #3759, opened 2014-02-27
Add option `-rpcamount` for monetary amounts as strings/satoshis in RPC

laanwj:
See this report: https://bitcoin.stackexchange.com/questions/22716/bitcoind-sendfrom-round-amount-error

I've been unable to reproduce it in either master or 0.8.6. However, in theory this could happen on architectures that have imprecise floating point types. Even though bitcoind itself doesn't rely on floating point types anywhere, Spirit JSON does use doubles internally for parsing values like 123.456 (and so do we in ValueFromAmount / AmountFromValue).

My first attempt was to try to fix this and make the parser use a fixed-point or string representation for JSON numbers (ie https://github.com/bitcoin/bitcoin/blob/master/src/json/json_spirit_reader_template.h#L499 ), but turned out to be a deep rabbit hole and doesn't solve the issue at the client side.

So this is my solution: add an option `-rpcstringamounts` to represent monetary amounts as strings in RPC. With this option any monetary amount will be emitted as a string with fixed 8 decimals, very easy to parse. Also this changes the parsing to accept the same format (by using the existing ParseMoney function).

Example getinfo:

```
{
    ...
    "balance" : "0.29099995",
    ...
}
```

Commits should be self-documenting.

PR #9855, opened 2017-02-25
RPC: Use integer satoshis instead BTC with decimals

jtimon:
There's been talks about using satoshis in RPC instead of BTC, which requires decimals and don't work well with json.

On the other hand, such a change would break the RPC API as well as many tests.
This is a draft on how we could move there (doing it only for 2 of the 8 RPC calls that would need to be changed). An useDecimalAmounts optional param is added. If true, the user can introduce BTC amounts as before. If false, the user needs to introduce the amounts in satoshis.

Interestingly enough, for signrawtransaction, useDecimalAmounts can default to false without breaking any tests (which means the amounts in useDecimalAmounts are not being currently tested).
In contrast, createrawtransaction would break many tests.

We could slowly adapt the tests to use satoshis and once that's done, at some point in the future we could break the API, remove all useDecimalAmounts params and not allow to BTC amounts anymore.

I didn't want go further before getting some feedback.


da2ce7:
Concept ACK.  I think that it is important to have an API Break at some point. Moving away from Decimal Amounts is very desirable.

jgarzik:
General agreement, with a notable caveat:

Older but widely deployed versions of the jansson library will truncate JSON numbers that are integers larger than unsigned 32-bit.

luke-jr:
Dislike making another numbered argument.

If we care about jansson bugs, I suspect throwing a ".0" on the end would workaround them.

jgarzik:
On the floating point side, you're limited by the 'double' C data type.

Some people just throw up their hands and store numbers inside of strings :)

JeremyRubin:
One idea:

pass a "multiplier" argument to the RPCs such that users can select arbitrary number systems. Default it to Bitcoin or Satoshis. This future proofs the API design in the event of any popular system (or even prepping for fractional satoshis...)


This could also be useful for people wishing to pay at an exchange rate :)

TheBlueMatt:
This is the kind of thing I think should be an option in the named
arguments list, but probably not added to every single RPC command's
positional parameters.

luke-jr:
`double` is large enough for any satoshi amounts on most systems (and any exceptions are likely to not support Core at all right now).

laanwj:
NACK doing it in this way.

- I'd prefer "return amounts as strings". It would be more consistent.  "Integers" don't exist in JSON. A JSON parser could still convert large numbers to float silently. Returning the values as strings means the program can use whatever internal format they want - it moves parsing away from the JSON parser, which is usually hard to override.

We already added handling of these for input in `AmountFromValue` in #6380, with the idea of making that an option later.

- If you really want to go ahead with this please see my implementation in #3759. It was much less invasive, effectively only two functions were touched `AmountFromValue`  and `ValueFromAmount`, and it allowed the user to select between four variants. 

There is no need to change every single call site of those functions for this!

jtimon:
Replaced by #9882

PR #9856, opened 2017-02-25
Terminate immediately when allocation fails

theuni:
Addresses #9854.

This installs a handler for failed allocations. Usually, a std::bad_alloc would be thrown, but we can instead opt to terminate immediately. See http://en.cppreference.com/w/cpp/memory/new/set_new_handler .

Our only manual malloc/free comes from prevector. Ideally we'd just use new[]/delete[] there, but the benchmark shows a noticeable difference, and I'm very hesitant to make that kind of change post-rc2.

As an alternative, this could even just be an assert() for 0.14.

For reference, here's the alternative using new/delete: https://github.com/theuni/bitcoin/commit/0c1b4107b918d457376dcd154f7e87d9d647bf91

Edit: g++/libstdc++ didn't support get_new_handler() yet, so assert it is.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -797,6 +797,19 @@ ServiceFlags nLocalServices = NODE_NETWORK;
 
 }
 
+[[noreturn]] static void new_handler_terminate()
+{
+    // Rather than throwing std::bad-alloc if allocation fails, terminate
+    // immediately to (try to) avoid chain corruption.
+    // Since LogPrintf may itself allocate memory, set the handler directly
+    // to terminate first.
+    std::set_new_handler(std::terminate);
+    LogPrintf("Error: Out of memory. Terminating.\n");
+
+    // The log was successful, terminate now.
+    std::terminate();
+};
+
 bool AppInitBasicSetup()
 {
     // ********************************************************* Step 1: setup
@@ -849,6 +862,9 @@ bool AppInitBasicSetup()
     // Ignore SIGPIPE, otherwise it will bring the daemon down if the client closes unexpectedly
     signal(SIGPIPE, SIG_IGN);
 #endif
+
+    std::set_new_handler(new_handler_terminate);
+
     return true;
 }
 
--- a/src/prevector.h
+++ b/src/prevector.h
@@ -5,6 +5,7 @@
 #ifndef _BITCOIN_PREVECTOR_H_
 #define _BITCOIN_PREVECTOR_H_
 
+#include <assert.h>
 #include <stdlib.h>
 #include <stdint.h>
 #include <string.h>
@@ -170,10 +171,15 @@ private:
             }
         } else {
             if (!is_direct()) {
+                /* FIXME: Because malloc/realloc here won't call new_handler if allocation fails, assert
+                    success. These should instead use an allocator or new/delete so that handlers
+                    are called as necessary, but performance would be slightly degraded by doing so. */
                 _union.indirect = static_cast<char*>(realloc(_union.indirect, ((size_t)sizeof(T)) * new_capacity));
+                assert(_union.indirect);
                 _union.capacity = new_capacity;
             } else {
                 char* new_indirect = static_cast<char*>(malloc(((size_t)sizeof(T)) * new_capacity));
+                assert(new_indirect);
                 T* src = direct_ptr(0);
                 T* dst = reinterpret_cast<T*>(new_indirect);
                 memcpy(dst, src, size() * sizeof(T));


sipa:
```
./prevector.h:178:21: error: ‘get_new_handler’ is not a member of ‘std’

                     std::get_new_handler()();
```

theuni:
@sipa grr, works for me locally. I'll try adding the include, but I'm guessing that libstdc++ didn't support it for 4.8 :(

JeremyRubin:
Why not make prevector throw std::bad_alloc or std::length_error? That way it's at least recoverable in future code, and makes it a bit more "API Consistent" with std::vector...

theuni:
@JeremyRubin The issue in #9854 is that exactly that bad_alloc is thrown but not handled correctly. We don't actually attempt to catch bad_alloc specifically anywhere. And logically it doesn't make sense to... if we've got an allocation problem, we need to shut down.

luke-jr:
The default action of exceptions is to abort the program, no? So we must be catching it at least indirectly... Why not just add an explicit catch for bad_alloc there? (or ideally, be specific about what exceptions we want to catch..)

laanwj:
ACK on the assert() in prevector.

I really don't like setting this global handler that crashes the program. I hope this is a temporary measure for 0.14?

laanwj:
Thinking of it, no, using assert for error handling is a bad idea. We should support compiling without assertions at some point.
If you really want to terminate the program immediatelly please log a message and call abort or so () :/

TheBlueMatt:
@laanwj I hope we do keep this longer-term, actually...std::bad_alloc, in the design of our entire codebase, seems massively dangerous to me (and I've wished we could do something like this for a while - just didn't know it existed). I'm happy if we remove it later, but the amount of auditing we'd have to do to check everywhere we call new, even through stl, seems like a rather significant effort.

laanwj:
@TheBlueMatt okay, yes, after discussion IRC I agree seems the least bad solution for now :/

JeremyRubin:
Ah, I see. I was going to suggest get_new_handler, then I actually read the above discussion ;)

I'm mostly just uncomfortable with using assert for error handling.

Can we not implement something similar to get_new_handler ourselves for unsupported platforms?

TheBlueMatt:
@laanwj yea, I totally agree there are places we don't want this (mostly deserialization comes to mind), but I believe we mostly fixed that by now?

src/init.cpp
```diff
@@ -797,6 +797,19 @@ ServiceFlags nLocalServices = NODE_NETWORK;
...
 }
 
+[[noreturn]] static void new_handler_terminate()
+{
+    // Rather than throwing std::bad-alloc if allocation fails, terminate
+    // immediately to (try to) avoid chain corruption.
+    // Since LogPrintf may itself allocate memory, set the handler directly
+    // to terminate first.
+    std::set_new_handler(std::terminate);
+    LogPrintf("Error: Out of memory. Terminating.\n");
```
kallewoof: Before this attempt, you could do `fputs(stderr, "Error: Out of memory. Terminating.\n");` which, I believe, does not call `malloc`. At least it didn't when I wrote my own `malloc` a few years ago, but it may be different per platform. A nice to have for when a node goes down with no log entries about why.

dcousens (approved):

src/init.cpp
JeremyRubin: I don't think this is threadsafe (ie, a concurrent OOM could cause it to terminate in the middle of a LogPrintf or print two LogPrintfs). I would make the handler have some `switch(atomic counter++)`. Then it is both re-entrant and threadsafe. (note that set_new_handler is threadsafe itself, just is threadsafe in this usage)

I think this sort of implementation should be ok:

```c++
static std::thread::id first_entrant{std::this_thread::get_id()};
static bool first_pass {true};
if (first_entrant == std::this_thread::get_id()) {
    if (first_pass) {
        first_pass = false;
        // Single faulted
        std::terminate();
    } else {
        // Double faulted
    }
} else {
    sleep();
}
```


Either of the above should be safe for concurrent OOM as well as double OOM faults.

sdaftuar:
nit: Perhaps we should add a comment to the `assert` lines introduced in prevector, mentioning their purpose/rationale?  (I'm pretty sure if I were reading the code in the future and hadn't seen this discussion, I'd definitely overlook the significance and implications.)

src/init.cpp
theuni: Those are both reasonable suggestions, but imo they're overthinking this a bit. If we OOM, the important thing is to shutdown, logging should be best-effort. I'd rather not complicate this too much.

sipa:
utACK

PR #9859, opened 2017-02-26
Make TestBlockValidity optional in CreateNewBlock

jameshilliard:
This is an alternative to #9858

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -381,6 +381,8 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-proxyrandomize", strprintf(_("Randomize credentials for every proxy connection. This enables Tor stream isolation (default: %u)"), DEFAULT_PROXYRANDOMIZE));
     strUsage += HelpMessageOpt("-rpcserialversion", strprintf(_("Sets the serialization of raw transaction or block hex returned in non-verbose mode, non-segwit(0) or segwit(1) (default: %d)"), DEFAULT_RPC_SERIALIZE_VERSION));
     strUsage += HelpMessageOpt("-seednode=<ip>", _("Connect to a node to retrieve peer addresses, and disconnect"));
+    if (showDebug)
+        strUsage += HelpMessageOpt("-testcnbvalid=<n>", strprintf(_("Run validity test when generating block template. (default: %d)"), DEFAULT_TEST_CNBVALID));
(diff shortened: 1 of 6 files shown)


src/validation.h
```diff
@@ -131,6 +131,7 @@ static const int64_t MAX_FEE_ESTIMATION_TIP_AGE = 3 * 60 * 60;
 /** Default for -permitbaremultisig */
 static const bool DEFAULT_PERMIT_BAREMULTISIG = true;
 static const bool DEFAULT_CHECKPOINTS_ENABLED = true;
+static const bool DEFAULT_TEST_CNBVALID = false;
```
luke-jr: At the very least, this should default to `true`.

src/init.cpp
```diff
@@ -381,6 +381,8 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-proxyrandomize", strprintf(_("Randomize credentials for every proxy connection. This enables Tor stream isolation (default: %u)"), DEFAULT_PROXYRANDOMIZE));
     strUsage += HelpMessageOpt("-rpcserialversion", strprintf(_("Sets the serialization of raw transaction or block hex returned in non-verbose mode, non-segwit(0) or segwit(1) (default: %d)"), DEFAULT_RPC_SERIALIZE_VERSION));
     strUsage += HelpMessageOpt("-seednode=<ip>", _("Connect to a node to retrieve peer addresses, and disconnect"));
+    if (showDebug)
```
luke-jr: Needs braces

luke-jr (changes requested):

src/init.cpp
jameshilliard: This is the style used above for single debug config options

src/validation.h
jameshilliard: This seems like it should just be a debug option that's off by default since it's something that should never be needed unless there's an internal bug.

src/test/miner_tests.cpp
```diff
@@ -199,6 +199,7 @@ BOOST_AUTO_TEST_CASE(CreateNewBlock_validity)
 
     LOCK(cs_main);
     fCheckpointsEnabled = false;
+    fRunTestBlockValidity = true;
```
jtimon: Instead of using the global for the tests, true can be passed in the tests directly to CreateNewBlock() as in #9858 . All other calls of CreateNewBlock() can use the global fRunTestBlockValidity

fanquake:
Going to close this for now. Needs a rebase, and hasn't had much discussion. Can be re-opened if/when necessary.

PR #9861, opened 2017-02-26
Trivial: Debug log ambiguity fix for peer addrs

keystrike:
This line can be misinterpreted as loading wallet addresses which is confusing, especially when the wallet is disabled.

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -2210,8 +2210,9 @@ bool CConnman::Start(CScheduler& scheduler, std::string& strNodeError, Options c
     SetBestHeight(connOptions.nBestHeight);
 
     clientInterface = connOptions.uiInterface;
-    if (clientInterface)
-        clientInterface->InitMessage(_("Loading addresses..."));
+    if (clientInterface) {
+        clientInterface->InitMessage(_("Loading P2P addresses..."));
+    }
     // Load addresses from peers.dat
     int64_t nStart = GetTimeMillis();
     {


kallewoof:
I'm unsure why the filename is needed, esp since it is printed in both outcomes later. If it is needed, you would need filename for banlist.dat load as well (line 2228), no?

keystrike:
The filename isn't as important as stating that p2p addresses are being loaded rather than wallet addresses. I added the filename just to be sure it was clear. This line happens shortly after "mapAddressBook.size() = 1" in debug.log, so it isn't obvious what these addresses are -- especially confusing when -disablewallet is set.

It may help if I explain why I noticed this. Previous versions of bitcoin would log setKeyPool.size, mapWallet.size, etc., even when the wallet was disabled, but that has now been fixed. So this line now stands out.

src/net.cpp
```diff
@@ -2211,7 +2211,7 @@ bool CConnman::Start(CScheduler& scheduler, std::string& strNodeError, Options c
 
     clientInterface = connOptions.uiInterface;
     if (clientInterface)
-        clientInterface->InitMessage(_("Loading addresses..."));
+        clientInterface->InitMessage(_("Loading p2p addresses from peers.dat..."));
```
MarcoFalke: No need to mention the file name in a translated message. Also missing curly braces {} according to style guide.

src/net.cpp
```diff
@@ -2210,8 +2210,9 @@ bool CConnman::Start(CScheduler& scheduler, std::string& strNodeError, Options c
     SetBestHeight(connOptions.nBestHeight);
 
     clientInterface = connOptions.uiInterface;
-    if (clientInterface)
-        clientInterface->InitMessage(_("Loading addresses..."));
+    if (clientInterface) {
+        clientInterface->InitMessage(_("Loading p2p addresses..."));
```
laanwj: s/p2p/P2P

MarcoFalke:
After fixing the nit, please squash your commits according to https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#squashing-commits

keystrike:
I'm sorry, I made a mistake when trying to squash my commits. Would someone please help?

laanwj:
keystrike: the git 'reflog' has a backup of all your past HEADs, so nothing will be lost, you can restore your old state if you can find it (with `reset --hard <commitid>`) and re-do from there.

Having said that, for small changes the easiest solution is sometimes to start over and just re-do your change and copy the commit message.

PR #9865, opened 2017-02-26
Change bitcoin address in RPC help message

marijnfs:
Change bitcoin address in RPC help message to an invalid address, so people don't accidentally send coins there (like I did).

### DIFF
--- a/src/rpc/misc.cpp
+++ b/src/rpc/misc.cpp
@@ -347,11 +347,11 @@ UniValue verifymessage(const JSONRPCRequest& request)
             "\nUnlock the wallet for 30 seconds\n"
             + HelpExampleCli("walletpassphrase", "\"mypassphrase\" 30") +
             "\nCreate the signature\n"
-            + HelpExampleCli("signmessage", "\"1D1ZrZNe3JUo7ZycKEYQQiQAWd9y54F4XZ\" \"my message\"") +
+            + HelpExampleCli("signmessage", "\"1D1ZrZNe3JUo7ZycKEYQQiQAWd9y54F4XX\" \"my message\"") +
             "\nVerify the signature\n"
-            + HelpExampleCli("verifymessage", "\"1D1ZrZNe3JUo7ZycKEYQQiQAWd9y54F4XZ\" \"signature\" \"my message\"") +
+            + HelpExampleCli("verifymessage", "\"1D1ZrZNe3JUo7ZycKEYQQiQAWd9y54F4XX\" \"signature\" \"my message\"") +
             "\nAs json rpc\n"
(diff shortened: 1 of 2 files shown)


laanwj:
Concept ACK. Hope you didn't lose a lot on this, it's not one of our addresses :(

Implementation-nit: please define a constant for this.

MarcoFalke:
I think the address was first mentioned in #3184.

src/rpc/misc.cpp
```diff
@@ -347,11 +347,11 @@ UniValue verifymessage(const JSONRPCRequest& request)
             "\nUnlock the wallet for 30 seconds\n"
             + HelpExampleCli("walletpassphrase", "\"mypassphrase\" 30") +
             "\nCreate the signature\n"
-            + HelpExampleCli("signmessage", "\"1D1ZrZNe3JUo7ZycKEYQQiQAWd9y54F4XZ\" \"my message\"") +
+            + HelpExampleCli("signmessage", "\"1D1ZrZNe3JUo7ZycKEYQQiQAWd9y54F4XX\" \"my message\"") +
```
MarcoFalke: I think you can use `DummyAddress(Params())` instead to get an invalid address.

marijnfs: Interesting. Though, I wanted to make the minimal change possible (I changed one letter) to keep things simpler. Do you think calling another function is necessary for this? It would be confusing if the address changes on each run for example.

JeremyRubin:
What if someone is trying to learn something and uses one of the documentation addresses  and gets "invalid address" and thinks something is wrong with their node/service? I think I prefer having a valid address in documentation for this reason.

My proposal to fix it would be that it uses an address that is owned by the wallet in question (specifically marked for help output), if the node has a wallet. That way any command that a user tries running from the help output without understanding could actually do something.

laanwj:
I strongly prefer using an invalid address here. You are not supposed to copy/paste the address, it can happen and it's better if it fails. Prevents a lot of panic "where did I just send to".

Using an address from the current wallet is a fair suggestion, but many of these calls are not coupled to the wallet at all, and shouldn't be.

JeremyRubin:
yeah on further thought failing is clearly better, even if it confuses someone once.

src/rpc/misc.cpp
laanwj: I think this is a good fix for 0.14 actually.
We can consider a cleaner solution later.

laanwj:
ACK 83ac719

rebroad:
@marijnfs sje3000@gmail.com is the email address of the user who probably has your BTC. Rather sneaky change to put their address (which is being used!) into the code, IMHO.

PR #9869, opened 2017-02-27
Move comment to right spot

RHavar:
...and kill some extraneous whitespace

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -465,21 +465,21 @@ bool CWallet::Verify()
         } catch (const boost::filesystem::filesystem_error&) {
             // failure is ok (well, not really, but it's not worse than what we started with)
         }
-        
+
         // try again
         if (!bitdb.Open(GetDataDir())) {
             // if it still fails, it probably means we can't even create the database env
             return InitError(strprintf(_("Error initializing wallet database environment %s!"), GetDataDir()));
         }
     }
-    
+
     if (GetBoolArg("-salvagewallet", false))
     {
         // Recover readable keypairs:
         if (!CWalletDB::Recover(bitdb, walletFile, true))
             return false;
     }
-    
+
     if (boost::filesystem::exists(GetDataDir() / walletFile))
     {
         CDBEnv::VerifyResult r = bitdb.Verify(walletFile, CWalletDB::Recover);
@@ -494,7 +494,7 @@ bool CWallet::Verify()
         if (r == CDBEnv::RECOVER_FAIL)
             return InitError(strprintf(_("%s corrupt, salvage failed"), walletFile));
     }
-    
+
     return true;
 }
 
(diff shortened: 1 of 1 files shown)


src/wallet/wallet.cpp
```diff
@@ -2145,11 +2145,11 @@ bool CWallet::SelectCoinsMinConf(const CAmount& nTargetValue, const int nConfMin
     setCoinsRet.clear();
     nValueRet = 0;
 
-    // List of values less than target
+
     pair<CAmount, pair<const CWalletTx*,unsigned int> > coinLowestLarger;
     coinLowestLarger.first = std::numeric_limits<CAmount>::max();
     coinLowestLarger.second.first = NULL;
-    vector<pair<CAmount, pair<const CWalletTx*,unsigned int> > > vValue;
+    vector<pair<CAmount, pair<const CWalletTx*,unsigned int> > > vValue; // List of values less than target
```
dcousens: `std::`

src/wallet/wallet.cpp
```diff
@@ -2145,11 +2145,11 @@ bool CWallet::SelectCoinsMinConf(const CAmount& nTargetValue, const int nConfMin
     setCoinsRet.clear();
     nValueRet = 0;
 
-    // List of values less than target
+
     pair<CAmount, pair<const CWalletTx*,unsigned int> > coinLowestLarger;
     coinLowestLarger.first = std::numeric_limits<CAmount>::max();
     coinLowestLarger.second.first = NULL;
-    vector<pair<CAmount, pair<const CWalletTx*,unsigned int> > > vValue;
+    std::vector<pair<CAmount, pair<const CWalletTx*,unsigned int> > > vValue; // List of values less than target
```
benma: I think the previous comment location was correct, as it is not only about this line, but all the lines that follow (the BOOST_FOREACH block below, which populates the vector). The few lines above is also needed for that.

Adding `std::` only here is a bit pointless, the whole file uses vector, set, pair etc. unqualified. Maybe a separate PR could explicitly qualify all of them, if preferred. Personally, I like it.

src/wallet/wallet.cpp
benma: I think the previous comment location was correct, as it is not only about this line, but all the lines that follow (the BOOST_FOREACH block below, which populates the vector). The few lines above is also needed for that.

Adding `std::` only here is a bit pointless, the whole file uses vector, set, pair etc. unqualified. Even the `pair` on the same line is not qualified.
Maybe a separate PR could explicitly qualify all of them, if preferred.  Personally, I would like it.

src/wallet/wallet.cpp
```diff
@@ -2149,7 +2149,7 @@ bool CWallet::SelectCoinsMinConf(const CAmount& nTargetValue, const int nConfMin
     pair<CAmount, pair<const CWalletTx*,unsigned int> > coinLowestLarger;
     coinLowestLarger.first = std::numeric_limits<CAmount>::max();
     coinLowestLarger.second.first = NULL;
-    vector<pair<CAmount, pair<const CWalletTx*,unsigned int> > > vValue; // List of values less than target
+    std::vector<pair<CAmount, pair<const CWalletTx*,unsigned int> > > vValue; // List of values less than target
```
laanwj: This is already done (for the entire file) by @kallewoof in https://github.com/bitcoin/bitcoin/pull/9643

PR #9871, opened 2017-02-27
Add a tree sha512 hash to merge commits

sipa:
This adds a SHA512 hash of the resulting merged tree to merge commit messages. Assuming gpg uses a sufficiently strong hash internally, this results in our gpg signed commits avoiding any potential SHA1 issues w.r.t. the commit's resulting contents (but NOT its history).

The hash can be verified by running `git ls-tree --full-tree -r --name-only HEAD | LANG=C sort | xargs -n 1 sha512sum | sha512sum` in the repository root.

### DIFF
--- a/contrib/devtools/github-merge.py
+++ b/contrib/devtools/github-merge.py
@@ -18,6 +18,7 @@ from __future__ import division,print_function,unicode_literals
 import os
 from sys import stdin,stdout,stderr
 import argparse
+import hashlib
 import subprocess
 import json,codecs
 try:
@@ -68,6 +69,27 @@ def ask_prompt(text):
     reply = stdin.readline().rstrip()
     print("",file=stderr)
     return reply
+
+def tree_sha512sum():
+    files = sorted(subprocess.check_output([GIT, 'ls-tree', '--full-tree', '-r', '--name-only', 'HEAD']).splitlines())
+    overall = hashlib.sha512()
+    for f in files:
+        intern = hashlib.sha512()
+        fi = open(f, 'rb')
+        while True:
+            piece = fi.read(65536)
+            if piece:
+                intern.update(piece)
+            else:
+                break
+        fi.close()
+        dig = intern.hexdigest()
+        overall.update(dig.encode("utf-8"))
+        overall.update("  ".encode("utf-8"))
+        overall.update(f)
+        overall.update("\n".encode("utf-8"))
+    return overall.hexdigest()
+
 
 def parse_arguments():
     epilog = '''
@@ -157,6 +179,9 @@ def main():
     subprocess.check_call([GIT,'checkout','-q','-b',local_merge_branch])
 
     try:
+        # Go up to the repository's root.
+        toplevel = subprocess.check_output([GIT,'rev-parse','--show-toplevel']).strip()
+        os.chdir(toplevel)
         # Create unsigned merge commit.
         if title:
             firstline = 'Merge #%s: %s' % (pull,title)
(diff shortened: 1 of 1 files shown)


petertodd:
Given how few lines of code this requires and how easy it is to independently calculate, concept ACK as a quick fix.

da2ce7:
Creating a 'git-sha512-treehashes.txt' file in the root folder of the repo.  That is a list of every sha512 tree hash for every merge commit (in this reop history).

Each new merge commit should append it's shar512-tree-hash to this file.  This would make falsifying sha1 root hash far-harder as modifying any file will update the root-hash from two different locations.

TheBlueMatt:
Hmm, what we really need to do is have a second PGP signature over just this hash in the commit message that can be verified completely independently from git.

contrib/devtools/github-merge.py
```diff
@@ -69,6 +70,27 @@ def ask_prompt(text):
...
+    for f in files:
+        intern = hashlib.sha512()
+        fi = open(f, 'rb')
+        while True:
+            piece = fi.read(65536)
+            if piece:
+                intern.update(piece)
+            else:
+                break
+        fi.close()
```
kallewoof: Nit: use `with` instead of `open`/`close`:
```Python
with open(f, 'rb') as fi:
    while True:
        piece = fi.read(65536)
        if piece:
            intern.update(piece)
        else:
            break
```

kallewoof (commented):
utACK fa89670

petertodd:
@TheBlueMatt The PGP signature on a git commit signs the entire commit object, including the message and thus SHA512 hash. So the SHA512 tree hash can be verified directly; you don't need another PGP signature.

TheBlueMatt:
@petertodd oh? IIRC it just signs a few SHA1 hashes, not the message directly.

petertodd:
@TheBlueMatt Nope, look at my OpenTimestamps code to timestamp git commits: https://github.com/opentimestamps/opentimestamps-client/blob/master/ots-git-gpg-wrapper#L87

sipa:
It signs the SHA1 hash of the entire commit object (excluding the
signature, of course).

petertodd:
@sipa Nope, that's simply not how git works. The commit object itself is what's piped to GnuPG, not just the hash of that object.

sipa:
@petertodd I'm aware. But GnuPG then takes a SHA1 hash of that.

petertodd:
@sipa GnuPG uses whatever hash algorithm you've told it to use; defaults used to be SHA1, but that's both easy to change, and IIRC the default is now SHA256.

sipa:
@petertodd I don't think so (but I'd be glad to be wrong). Look at the actual data with `git cat-file -p <commit>`. It does not include the otherwise mandatory hash type line, so I assume gpg prepends that before verification. That can't work if it's anything but the default. git does not use detached signatures.

petertodd:
@sipa It's missing because it's not a clearsigned signature; it's an ascii encoded detached signature.

sipa:
Hmm, I read conflicting information. I hope I'm wrong.

petertodd:
@sipa Don't worry, you are wrong. :)

Like I said above, look at the git integration code I linked above from the OpenTimestamps client. Notably, the command line flags -bsau are passed to gpg by git when signing, which creates an ascii-encoded detached signature.

dcousens:
Relevant https://stackoverflow.com/questions/23584990/what-data-is-being-signed-when-you-git-commit-gpg-sign-key-id ?

petertodd:
@dcousens They're talking about the equivalent of segwit vs how Bitcoin works now; it's not relevant to what we're talking about here.

dcousens:
@petertodd no worries,  did `git` change or is the comment more related to the fact that the GPG signing process used here is different?

FWIW,  I also saw https://grimoire.ca/git/detached-sigs

petertodd:
@dcousens The "detached sigs" your links are referring to isn't the same thing as the "detached sigs" we're talking about above.

sipa:
Cool, so this is much more useful than I thought :)

achow101:
So as long as the merge commit signers use SHA512 (or something like that, maybe SHA256) for GPG this solution is essentially perfect (besides git upgrading to a new hash function).

laanwj:
I'm not sure how this works: `git ls-tree --full-tree -r --name-only HEAD` returns a list of filenames. Is it enough assurance to hash that, you don't need the file contents?

sipa:
I read in the actual files, and hash those, and then hash the result as like the result of sha512sum (look at the command line).

achow101:
@laanwj it does that command to grab the list of files. Then it opens each file in the list and hashes it.

laanwj:
> I read in the actual files, and hash those, and then hash the result as like the result of sha512sum (look at the command line).

Oh, then things are alright again, thanks :)

laanwj:
Tested in: https://github.com/bitcoin/bitcoin/commit/c7e57ce98154bccc0941f078296ba7b2eaf23e90

contrib/devtools/github-merge.py
```diff
@@ -175,14 +200,29 @@ def main():
...
+            message += '\n\nTree-SHA512: ' + first_sha512
+        except subprocess.CalledProcessError as e:
+            printf("ERROR: Unable to compute tree hash")
+            exit(4)
+        try:
+            subprocess.check_call([GIT,'commit','--amend','-m',message.encode('utf-8')])
+        except subprocess.CalledProcessError as e:
+            printf("ERROR: Cannot update message.",file=stderr)
+            exit(4)
+        second_sha512 = tree_sha512sum()
```
MarcoFalke: Why is the above call to tree_sha512sum() guarded by a try except clause and this one is not?

Anyway, I think the right place to put the second check would be `verify-commits.sh`.

MarcoFalke:
Concept ACK

jonasschnelli (approved):
utACK fa89670d34ac7839e7e2fe6f59fdfd6cc76c4483

achow101:
utACK fa89670

TheBlueMatt:
We should probably have some better handling for symlinks, no?

sipa:
@TheBlueMatt I think that may be overkill, but ok... how would you construct something that's easily shell-verifiable?

laanwj:
> We should probably have some better handling for symlinks, no?

What about just rejecting symlinks? We don't use them in this repository and probably shouldn't as not all OSes have that concept.

TheBlueMatt:
@sipa not sure about shell-verifiable, but git-cat has some symlinks options. I'm more than happy to reject them wholesale, however, given the dubious value and obvious annoyance/risk.

petertodd:
ACK to rejecting symlinks.

We don't need to go overkill here and make this a scheme every project could adopt.

laanwj:
I'm just going to merge this - this is a nice and simple change that alleviates the direct worries about SHA1 compromise. More advanced things or functionality to reject symlinks can be added to the script later.

PR #9873, opened 2017-02-27
[do not merge] upload test_bitcoin executable for #9825

laanwj:
More troubleshooting for #9825, to find out if there is something wrong with the failing executable.


laanwj:
I'm slightly optimistic that this has been solved by #9875. If the bug rears its head again I'll reopen.

PR #9888, opened 2017-02-28
travis: Verify commits only for one target

maflcko:
This abuses the CHECK_DOC flag to serve as a general indicator to
execute platform independent sanity checks.


MarcoFalke:
Motivation is that we don't have a red travis for all platforms when there are issues with a platform independent check.

laanwj:
Yes, it makes sense to do checks like this only on one target.

laanwj:
Ping @theuni

theuni:
Makes sense. utACK.

Issue #9655, opened 2017-01-31
Add editor button bitcoin.conf in wallet bitcoin


DATSEC:
I noticed that in linux there is no additional menus and windows do not understand where it is located. 
I think it would be not bad to have a button and set the config manually. 
In the future, this button can be used to edit the wallet.

jonasschnelli:
I can't 100% follow the text above. Do you mean a button (like the debug.log button we have) which directly opens the `bitcoin.conf` file in a text-editor?
This would indeed be a neat function.

fanquake:
This this was fixed by #9890.

PR #9890, opened 2017-02-28
Add a button to open the config file in a text editor

ghost:
This button was described in issue  #9655 

I added it to the options dialog

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -417,6 +417,22 @@ void openDebugLogfile()
         QDesktopServices::openUrl(QUrl::fromLocalFile(boostPathToQString(pathDebug)));
 }
 
+bool openBitcoinConf()
+{
+    boost::filesystem::path pathConfig = GetConfigFile(BITCOIN_CONF_FILENAME);
+
+    /* Create the file */
+    boost::filesystem::ofstream configFile(pathConfig, std::ios_base::app);
+    
+    if (!configFile.good())
+        return false;
+    
+    configFile.close();
+    
+    /* Open bitcoin.conf with the associated application */
+    return QDesktopServices::openUrl(QUrl::fromLocalFile(boostPathToQString(pathConfig)));
+}
+
 void SubstituteFonts(const QString& language)
 {
 #if defined(Q_OS_MAC)
--- a/src/qt/forms/optionsdialog.ui
+++ b/src/qt/forms/optionsdialog.ui
@@ -692,17 +692,34 @@
    <item>
     <layout class="QHBoxLayout" name="horizontalLayout_Buttons">
      <item>
-      <widget class="QPushButton" name="resetButton">
-       <property name="toolTip">
-        <string>Reset all client options to default.</string>
-       </property>
-       <property name="text">
-        <string>&amp;Reset Options</string>
-       </property>
-       <property name="autoDefault">
-        <bool>false</bool>
-       </property>
-      </widget>
+      <layout class="QVBoxLayout" name="verticalLayout_Buttons">
+       <item>
+        <widget class="QPushButton" name="openBitcoinConfButton">
+         <property name="toolTip">
+          <string>Open the %1 configuration file from the working directory.</string>
+         </property>
+         <property name="text">
+          <string>Open Configuration File</string>
+         </property>
+         <property name="autoDefault">
+          <bool>false</bool>
+         </property>
+        </widget>
+       </item>
+       <item>
+        <widget class="QPushButton" name="resetButton">
+         <property name="toolTip">
+          <string>Reset all client options to default.</string>
+         </property>
+         <property name="text">
+          <string>&amp;Reset Options</string>
+         </property>
+         <property name="autoDefault">
+          <bool>false</bool>
+         </property>
+        </widget>
+       </item>
+      </layout>
      </item>
      <item>
       <spacer name="horizontalSpacer_1">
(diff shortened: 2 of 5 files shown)


laanwj:
Concept ACK

jonasschnelli:
Nice.
Concept ACK. Will test soon.

jonasschnelli:
* I think it should create the file if `bitcoin.conf` does not exists (right now, there is no reaction if the file does not exists)
* The buttons should probably aligned/distributed vertically (horizontally it will eat up a lot of space in non en_EN lang)

bildschirmfoto 2017-03-01 um 22 26 46
bildschirmfoto 2017-03-01 um 22 26 32

ericshawlinux:
@jonasschnelli Ok

ericshawlinux:
@jonasschnelli Is this what you had in mind for the layout? I wanted to check before pushing

button_demo

jonasschnelli:
@ericshawlinux: I think this looks better and is more flexible.

ericshawlinux:
I think so too. I amended my commit from the other day

ericshawlinux:
(I just fixed a mistake I made, I accidentally misplaced a spacer while using Qt-designer. It is back where it belongs now)

luke-jr:
I wonder if users might get configured that none of their options show up in the config file...? Not really sure how one might improve on this, though.

jonasschnelli:
> I wonder if users might get configured that none of their options show up in the config file...? Not really sure how one might improve on this, though.

Yes. We need to watch out here. The configuration levels are complex.
1. startup arguments (`-arg=val`)
2. bitcoin.conf
3. Qt configuration (GUI)

If a user opens "the" configuration file, he might expect to see all existing configuration values (including the GUI and startup one).

We should probably label the button "open experts configuration file" or similar.

ericshawlinux:
Well it says "Active command-line options override the above options" so maybe if we put the button to open the file above that, the user will know they are different.

jonasschnelli:
An alternative would be to first show an QMessageBox with some infos about how the bitcoin.conf works and after that alert box, open the bitcoin.conf file.

ericshawlinux:
Okay. How does this sound?
"The configuration is used to specify advanced user options less any command-line or Qt options. Any command-line options will override this configuration file."

ericshawlinux:
I added the information dialog

src/qt/guiutil.cpp
```diff
@@ -417,6 +417,18 @@ void openDebugLogfile()
         QDesktopServices::openUrl(QUrl::fromLocalFile(boostPathToQString(pathDebug)));
 }
 
+void openBitcoinConf()
+{
+    boost::filesystem::path pathConfig = GetConfigFile(BITCOIN_CONF_FILENAME);
+
+    /* Create the file */
+    std::fstream fs(pathConfig.c_str(), std::fstream::app);
```
jonasschnelli: Gitian Windows Build is failing here:
```
qt/guiutil.cpp: In function ‘void GUIUtil::openBitcoinConf()’:
qt/guiutil.cpp:425:58: error: no matching function for call to ‘std::basic_fstream<char>::basic_fstream(const value_type*, const openmode&)’
     std::fstream fs(pathConfig.c_str(), std::fstream::app);
                                                          ^
qt/guiutil.cpp:425:58: note: candidates are:
In file included from /home/ubuntu/build/bitcoin/depends/i686-w64-mingw32/share/../include/boost/filesystem/operations.hpp:45:0,
                 from /home/ubuntu/build/bitcoin/depends/i686-w64-mingw32/share/../include/boost/filesystem.hpp:17,
                 from qt/guiutil.h:19,
                 from qt/guiutil.cpp:5:
/usr/include/c++/4.8/fstream:831:7: note: std::basic_fstream<_CharT, _Traits>::basic_fstream(const string&, std::ios_base::openmode) [with _CharT = char; _Traits = std::char_traits<char>; std::string = std::basic_string<char>; std::ios_base::openmode = std::_Ios_Openmode]
       basic_fstream(const std::string& __s,
```

jonasschnelli:
Currently the build is failing on Windows (over Gitian):
https://bitcoin.jonasschnelli.ch/build/53

Otherwise this works fine. Tested on OSX and Ubuntu.

bildschirmfoto 2017-03-17 um 16 09 09
bildschirmfoto 2017-03-17 um 16 09 13

bildschirmfoto 2017-03-17 um 16 17 40
bildschirmfoto 2017-03-17 um 16 17 32

src/qt/guiutil.cpp
ericshawlinux: ok, I can look at it on monday

ericshawlinux:
whoops

ericshawlinux:
okay, I had an issue with git for a few minutes, but I think I fixed it now.

as for the code, I changed the mode from std::fstream::app to std::ios_base::app that builds okay with linux, and I think with windows, since I read the error message you sent me.

jonasschnelli:
Thanks... started again a gitian build: https://bitcoin.jonasschnelli.ch/build/60

ericshawlinux:
Still didn't work. I guess I didn't look closely enough at the error before. Now, I think I need to call path.native() instead of path.c_str() based on the constructor type.

I will push that soon

laanwj:
Please use `qstringToBoostPath` and `boostPathToQString` from GUIUtil to convert between QString and `boost::filesystem`.  They avoid UTF-8 issues that will arise when using other methods.

ericshawlinux:
ok

ericshawlinux:
@laanwj I think for using fstream, I needed to use boost::filesystem::path.native() because it doesn't take a Qstring. But I changed the other line below that to use that function you mentioned.
The purpose of using the fstream was so I can ensure the file exists before opening it, and creating a new one if it doesn't. Would I need to find another solution, in order to avoid using an ordinary string& ?

ericshawlinux:
thanks, I figured it out using boost and didn't need to convert the path at all. It should build on windows now

ericshawlinux:
@jonasschnelli could you do the gitian build on your site again? I am pretty confident it will work this time after Laan advised me, I changed it to use boost path and not a c string.

laanwj:
Code looks good to me now, although I wonder about error handling: what happens if the config file is not writable?

ericshawlinux:
When I change the permissions, It seems to ask me how I want to open the file, which doesn't make sense. I can add some logic and show a QMessageBox::critical() dialog if it fails to open the file.

ericshawlinux:
Now an error appears when the file cannot be opened.

error2

fanquake:
Testing 541f8d8 on OS X.
Appearance in the GUI:
screen shot 1
Pre opening message:
screen shot 2
Unable to open file error message:
screen shot 4

I'm wondering if we can word the pre-opening message any better. Could we use Core rather than qt, or is that more confusing?

MarcoFalke:
Maybe replace 'qt options' with 'gui settings'?

laanwj:
Some comments on messages:
- Agree with @MarcoFalke , "Qt options" should be "GUI settings". Except when directly referring to the widget toolkit we avoid referring to Qt, as that's an implementation detail.
- "config file" in the error message should be "configuration file" (as the button) or just the filename, "bitcoin.conf". There's no reason to shorten here, and it could make it more difficult for translators.
- Instead of "less any..." I'd use "except any...". Seems more common English (at least to me as non-native speaker.)

ericshawlinux:
I agree, saying "less any..." is not common, but to make it more descriptive and clear I reworded the whole message the best that I could. How is this?
"The configuration file is used to specify advanced user options which override GUI settings. Additionally, any command-line options will override this configuration file."

jonasschnelli:
Tested ACK 9ab9e7d1 (https://bitcoin.jonasschnelli.ch/build/69)

PR #9891, opened 2017-02-28
depends: make osx output deterministic

theuni:
ld64 is threaded, and uses a worker for each CPU to parse input files. But
there's a bug in the parser causing dependencies to be calculated differently
based on which files have already been parsed.

As a result, builders with more CPUs are more likely to see non-determinism.

This looks to have been fixed in a newer version of ld64, so just disable
threading for now. There's no noticible slowdown.

### DIFF
--- a/depends/packages/native_cctools.mk
+++ b/depends/packages/native_cctools.mk
@@ -38,7 +38,8 @@ $(package)_cxx=$($(package)_extract_dir)/toolchain/bin/clang++
 endef
 
 define $(package)_preprocess_cmds
-  cd $($(package)_build_subdir); ./autogen.sh
+  cd $($(package)_build_subdir); ./autogen.sh && \
+  sed -i.old "/define HAVE_PTHREADS/d" ld64/src/ld/InputFiles.h
 endef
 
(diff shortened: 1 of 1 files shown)


paveljanik:
@theuni Brilliant investigation! Thank you!

For more info, the unpacked code is available at https://github.com/froydnj/ld64/blob/master/src/ld/InputFiles.cpp#L716

utACK https://github.com/bitcoin/bitcoin/pull/9891/commits/9e4d842afc67c63448cbbe30e5cea906e9c87f7e

jonasschnelli:
Great, Thanks @theuni.
utACK 9e4d842afc67c63448cbbe30e5cea906e9c87f7e

laanwj:
Wow, this is a really sneaky bug. Thanks for catching it. utACK https://github.com/bitcoin/bitcoin/commit/9e4d842afc67c63448cbbe30e5cea906e9c87f7e

achow101:
Post merge ACK. Tested by running 50 gitian builds with this patch applied and the hashes of each build match.

PR #9894, opened 2017-03-01
remove 'label' filter for rpc command help

instagibbs:
No idea why it's there(which gives me some pause) since it's a sourceforge-era commit, but it seems useless and filters out any command with "label" in the name.

### DIFF
--- a/src/rpc/server.cpp
+++ b/src/rpc/server.cpp
@@ -193,9 +193,6 @@ std::string CRPCTable::help(const std::string& strCommand) const
     {
         const CRPCCommand *pcmd = command.second;
         string strMethod = pcmd->name;
-        // We already filter duplicates, but these deprecated screw up the sort order
-        if (strMethod.find("label") != string::npos)
-            continue;
         if ((strCommand != "" || pcmd->category == "hidden") && strMethod != strCommand)
             continue;
         try


src/rpc/server.cpp
```diff
@@ -194,8 +194,6 @@ std::string CRPCTable::help(const std::string& strCommand) const
         const CRPCCommand *pcmd = command.second;
         string strMethod = pcmd->name;
         // We already filter duplicates, but these deprecated screw up the sort order
```
dcousens: is this comment relevant?

instagibbs: It used to have other deprecated commands, long-ago deleted. I'm not actually sure what the comment is trying to say.

laanwj: Yes please remove the comment too

laanwj:
Eh yes this is old cruft, we certainly don't want to remove *label* commands from the help. 
You've potentially saved me a hour of frustrated debugging for #7729.
utACK https://github.com/bitcoin/bitcoin/pull/9894/commits/91321e63e0dbdabae682b7455329fded8aa576af.

src/rpc/server.cpp
laanwj: Yes please remove the comment too

instagibbs:
Removed the comment as well.

I was a bit surprised that this had never come up for #7729, heh

MarcoFalke:
utACK 6665977

jnewbery:
Tested ACK https://github.com/bitcoin/bitcoin/pull/9894/commits/666597798c07d9df06681be63d2b2f51f0d5ca8a

Travis has been timing out all day. The failure isn't due to this PR.

jonasschnelli:
utACK 666597798c07d9df06681be63d2b2f51f0d5ca8a

paveljanik:
ACK https://github.com/bitcoin/bitcoin/commit/666597798c07d9df06681be63d2b2f51f0d5ca8a

luke-jr:
For historical reference: IIRC, the RPCs were originally *label*, and had been replaced by the account variants.

laanwj:
> For historical reference: IIRC, the RPCs were originally label, and had been replaced by the account variants.

Interesting. So we're going full circle.

PR #9895, opened 2017-03-01
Turn TryCreateDirectory() into TryCreateDirectories()

benma:
Use case: TryCreateDirectory(GetDataDir() / "blocks" / "index") would
fail if the blocks directory was not explicitly created before.

The line that did so was in a weird location and could be removed as a
result.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -653,21 +653,21 @@ bool RenameOver(fs::path src, fs::path dest)
 }
 
 /**
- * Ignores exceptions thrown by Boost's create_directory if the requested directory exists.
+ * Ignores exceptions thrown by Boost's create_directories if the requested directory exists.
  * Specifically handles case where path p exists, but it wasn't possible for the user to
  * write to the parent directory.
  */
-bool TryCreateDirectory(const fs::path& p)
+bool TryCreateDirectories(const fs::path& p)
 {
     try
     {
-        return fs::create_directory(p);
+        return fs::create_directories(p);
     } catch (const fs::filesystem_error&) {
         if (!fs::exists(p) || !fs::is_directory(p))
             throw;
     }
 
-    // create_directory didn't create the directory, it had to have existed already
+    // create_directories didn't create the directory, it had to have existed already
     return false;
 }
 
--- a/src/dbwrapper.cpp
+++ b/src/dbwrapper.cpp
@@ -108,7 +108,7 @@ CDBWrapper::CDBWrapper(const fs::path& path, size_t nCacheSize, bool fMemory, bo
             leveldb::Status result = leveldb::DestroyDB(path.string(), options);
             dbwrapper_private::HandleError(result);
         }
-        TryCreateDirectory(path);
+        TryCreateDirectories(path);
         LogPrintf("Opening LevelDB in %s\n", path.string());
     }
     leveldb::Status status = leveldb::DB::Open(options, path.string(), &pdb);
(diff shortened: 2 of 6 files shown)


benma:
The blocks directory is now created here:

https://github.com/benma/bitcoin/blob/de85040fc552178815a71cacb00d16faff994c9a/src/dbwrapper.cpp#L52
via
https://github.com/benma/bitcoin/blob/11049f4fe62606d1b0380a9ef800ac130f0fbadf/src/txdb.cpp#L70

and is created the first time it is needed with the initialization of CBlockTreeDB in init.cpp.

dcousens (approved):
utACK

src/util.cpp
```diff
@@ -610,15 +610,15 @@ bool RenameOver(boost::filesystem::path src, boost::filesystem::path dest)
...
  * Specifically handles case where path p exists, but it wasn't possible for the user to
  * write to the parent directory.
  */
-bool TryCreateDirectory(const boost::filesystem::path& p)
+bool TryCreateDirectories(const boost::filesystem::path& p)
 {
     try
     {
-        return boost::filesystem::create_directory(p);
+        return boost::filesystem::create_directories(p);
```
laanwj: Is this utility function needed at all? I think the definition of `fs::create_directories` is already exactly what we want here:  http://www.boost.org/doc/libs/1_61_0/libs/filesystem/doc/reference.html#create_directories

```
Returns: true if a new directory was created, otherwise false.

Throws: As specified in Error reporting.
```

benma: The function is apparently needed, the comment explains why:
```c++
/**
 * Ignores exceptions thrown by Boost's create_directory if the requested directory exists.
 * Specifically handles case where path p exists, but it wasn't possible for the user to
 * write to the parent directory.
 */
```
 
While I can't recreate this condition, digging up the issue that introduced this talks about some edge cases where it does happen (Truecrypt, ...?): https://github.com/bitcoin/bitcoin/issues/432

So we should probably leave it.

laanwj: Ugh. I think not being able to rely on the postcondition is incredibly ugly, and should simply be patched in boost upstream. But okay...

fanquake:
This needs a rebase.

benma:
Rebased.

benma:
@dcousens @laanwj could you please re-review (no changes, just a rebase) and merge?

ryanofsky (commented):
utACK e94c40c770a5de6c95a8c53cfb1717a35c008d83, nice change.

benma:
@laanwj bump :)

sipa:
utACK e94c40c770a5de6c95a8c53cfb1717a35c008d83

MarcoFalke:
utACK e94c40c

src/util.cpp
```diff
@@ -653,15 +653,15 @@ bool RenameOver(fs::path src, fs::path dest)
 }
 
 /**
- * Ignores exceptions thrown by Boost's create_directory if the requested directory exists.
+ * Ignores exceptions thrown by Boost's create_directories if the requested directory exists.
```
flack: Since you're touching this line anyways, you could also remove the reference to Boost, which isn't used here anymore (also, the comment at the end of the function could be updated)

benma: But it is using Boost.

`namespace fs = boost::filesystem;`.

I fixed the comment at the end of the function.

flack: hm, ok. I was under the impression that this was changed in https://github.com/benma/bitcoin/commit/bac5c9cf643e9333479ac667426d0b70f8f3aa7f, but I guess I misread

PR #9897, opened 2017-03-01
AppInitMain: split initialization of Connman into a new function

benma:
An effort to reduce the size of AppInitMain().

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1147,6 +1147,61 @@ bool AppInitSanityChecks()
     return LockDataDirectory(true);
 }
 
+static bool InitConnman(bool listen)
+{
+    // ********************************************************* Step 6a: connection manager initialization
+    assert(!g_connman);
+    g_connman = std::unique_ptr<CConnman>(new CConnman(GetRand(std::numeric_limits<uint64_t>::max()), GetRand(std::numeric_limits<uint64_t>::max())));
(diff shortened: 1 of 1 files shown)


benma:
Rebased.

dcousens:
concept ACK

laanwj:
Ping @cfields (as he wrote the Connman code).

theuni (changes requested):
Mm, I like the consolidation, but this kinda confuses the interface and gets in the way of the clean instantiation. The things currently done before starting connman are actually just config options that haven't been moved in yet.

What I'd much prefer (just hadn't gotten around yet) is to add:
CConnman::Options::vWhitelist
CConnman::Options::vBinds
CConnman::Options::vSeedNodes

I think we can combine the approaches by keeping the creation as-is, and having InitConnman() parse the config options and return a CConnman::Options.

benma:
@theuni 

Thanks, that makes a whole lot of sense. I started the process with `vSeedNodes` in #10467, and will revisit this PR once the rest of the initialization is cleaned up.

ryanofsky:
#10496 seems to be the next PR in this progression.

laanwj:
Needs rebase after #10467 (I guess?)

benma:
Might revisit with a new PR some day.

PR #9899, opened 2017-03-01
Add jonasschnellis new NIST-P 256 signing key

jonasschnelli:
Hash: SHA512

Removed old expired keys.
Added a new NIST-P 256 subkey that will be used to sign git merges and gitian builds.
-----BEGIN PGP SIGNATURE-----

iHUEARMKAB0WIQRDU2HzXdDCNVJGfeBxOm5iFuoefwUCWLbGCwAKCRBxOm5iFuoe
fzm+AQCW5rLFUQ29eUG0MvEyqI+DYpJIDSzELZWmaxCERB5ewgD6Au9iaRozesPb
KVQ3CYim9FyAJFFg3gALKPNHkv51OW8=
=Yp9o
-----END PGP SIGNATURE-----


MarcoFalke:
I don't think it is possible to delete keys when they have been made
public once. Also, it should be sufficient to upload the new subkeys
to a keyserver instead of updating the key here. (We can't possibly
keep up with updating the keys here anyway, considering that every
couple of days there are new signatures, revocations, subkeys, etc...)

MarcoFalke:
Hmm, I can't verify your signature on a vanilla xenial box with the gpg apt.

```sh
gpg --verify 9899 
gpg: Signature made Wed Mar  1 13:00:59 2017 UTC using ? key ID 16EA1E7F
gpg: Can't check signature: unknown pubkey algorithm
```


So I think this will not only break travis, but also make it harder for ppl to verify.

jonasschnelli:
It requires gpg >= 2.1.17 with ECDSA support to verify those signatures.
I think a project like Bitcoin Core should no longer only-rely on RSA for code sign and commit signatures.
Github and many other projects have started to support ECDSA.

gmaxwell:
I do not know of anyone with a founded belief P256 offers more security that 4096 bit RSA. But thats fine, not all keys here need to be equally secure.

But it is an issue that there is no support for this in the stable version of GPG and so many systems cannot support this key and probably won't until after there is a stable GPG with ECC support (I believe this will 2.2) which has been out for a while-- which will probably be next year.

MarcoFalke:
> requires gpg >= 2.1.17 

2.1.0 should be enough, which actually ships with xenial and later. Though, I'd still prefer to wait a few months/years for the reasons @gmaxwell mentioned.

jonasschnelli:
Okay. Fair enough,... then I (grudgingly) generate again a RSA4096 key.

jonasschnelli:
Closing...

PR #9903, opened 2017-03-02
Docs: add details to -rpcclienttimeout doc

ian-kelling:


ian-kelling:
How I I verified the 0 value disables timeout: Bitcoin passes the value directly to libevent, reading libevent source code shows that 0 disables it. I did a manual test to verify that it would at least go longer than the 900 default, which it did.

jnewbery:
utACK. Seems reasonable.

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/9903/commits/90a1d9a55d81029deb482888a9ddaa364f4a2c73

BTW: I just tested all 0 (unlimited), 900 (the default value) and even 2000. It all worked as expected.

PR #9904, opened 2017-03-02
test: Fail if InitBlockIndex fails

laanwj:
If InitBlockIndex fails, then it will segfault later. Same for the later ActivateBestChain. BOOST_REQUIRE the result, so that an error will be reported and the test case aborted.

### DIFF
--- a/src/test/test_bitcoin.cpp
+++ b/src/test/test_bitcoin.cpp
@@ -69,11 +69,11 @@ TestingSetup::TestingSetup(const std::string& chainName) : BasicTestingSetup(cha
         pblocktree = new CBlockTreeDB(1 << 20, true);
         pcoinsdbview = new CCoinsViewDB(1 << 23, true);
         pcoinsTip = new CCoinsViewCache(pcoinsdbview);
-        InitBlockIndex(chainparams);
+        BOOST_REQUIRE(InitBlockIndex(chainparams));
         {
             CValidationState state;
             bool ok = ActivateBestChain(state, chainparams);
-            BOOST_CHECK(ok);
+            BOOST_REQUIRE(ok);
         }
         nScriptCheckThreads = 3;
         for (int i=0; i < nScriptCheckThreads-1; i++)


MarcoFalke:
utACK 64854666f57d9f94269000f10897858e8c60d99f. Thanks for introducing
the first BOOST_REQUIRE, this was overdue.

src/test/test_bitcoin.cpp
```diff
@@ -69,11 +69,11 @@ TestingSetup::TestingSetup(const std::string& chainName) : BasicTestingSetup(cha
         pblocktree = new CBlockTreeDB(1 << 20, true);
         pcoinsdbview = new CCoinsViewDB(1 << 23, true);
         pcoinsTip = new CCoinsViewCache(pcoinsdbview);
-        InitBlockIndex(chainparams);
+        BOOST_REQUIRE(InitBlockIndex(chainparams));
         {
             CValidationState state;
             bool ok = ActivateBestChain(state, chainparams);
-            BOOST_CHECK(ok);
+            BOOST_REQUIRE(ok);
```
jnewbery: Why not `BOOST_REQUIRE(ActivateBestChain(state, chainparams));` on one line?

jnewbery (approved):
Looks good. Tested ACK 64854666f57d9f94269000f10897858e8c60d99f. One nit.

src/test/test_bitcoin.cpp
MarcoFalke: smaller diff and we don't want to imply assertions with side effects are our default coding style :P

benma: Then you should also do the same with the above BOOST_REQUIRE :)

MarcoFalke: Yeah, it is not consistent, but I don't think it matters too much.

jnewbery: > we don't want to imply assertions with side effects are our default coding style

:thumbsup:

laanwj: Meh, in the tests it's pretty normal to check something with side effects. This is only an "assertion" in the abstract sense, not the C/C++ sense that it will potentially be disabled (new: tests in release mode, with disabled tests!).

Anyhow let's not nitpick this two-line change to death.

PR #9905, opened 2017-03-02
[contrib] gh-merge: Move second sha512 check to the end

maflcko:
This moves the check down, so it is run after the signature was created. Otherwise it seems to me that this would not do anything.

### DIFF
--- a/contrib/devtools/github-merge.py
+++ b/contrib/devtools/github-merge.py
@@ -212,10 +212,6 @@ def main():
         except subprocess.CalledProcessError as e:
             printf("ERROR: Cannot update message.",file=stderr)
             exit(4)
-        second_sha512 = tree_sha512sum()
-        if first_sha512 != second_sha512:
-            print("ERROR: Tree hash changed unexpectedly",file=stderr)
-            exit(4)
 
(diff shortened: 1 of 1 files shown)


sipa:
utACK

laanwj:
> Otherwise it seems to me that this would not do anything.

It does work, as you're not supposed to change the tree during your validation session.

Moving it to the end before the GPG signing seems more consistent to me too, though. utACK https://github.com/bitcoin/bitcoin/pull/9905/commits/fa1b11c2671f4013c34e7f272685a6242419b9b6

contrib/devtools/github-merge.py
```diff
@@ -270,6 +266,11 @@ def main():
             print("Not signing off on merge, exiting.",file=stderr)
             exit(1)
 
+        second_sha512 = tree_sha512sum()
```
laanwj: I'd somewhat prefer to do this check

- after signing off on the merge
- but before putting the actual GPG signature

This avoids having a signed commit around (even though it's throwing it away) from something that isn't supposed to be.

MarcoFalke: done

PR #9906, opened 2017-03-03
Disallow copy constructor CReserveKeys

instagibbs:
In the case where you use a copy constructor(no instance in codebase now), Bad Things can happen, including crash.

### DIFF
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -1009,6 +1009,10 @@ public:
         pwallet = pwalletIn;
     }
 
+    CReserveKey() = default;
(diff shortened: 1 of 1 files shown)


laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/9906/commits/562965b3a3aca93a07ad65a3c6d21ecbc48f448d

gmaxwell (approved):
utACK

sipa:
You probably want to delete assignment operators too, or at least invoke ReturnKey before assigning.

instagibbs:
Went ahead and made it non-copyable altogether.

PR #9910, opened 2017-03-03
Docs: correct and elaborate -rpcbind doc

ian-kelling:
The default was incorrect unless -rpcallowip was also specified.

The code implementing the behavior described in the doc is fairly readable (src/httpserver.cpp, method HTTPBindAddresses). I also tested manually to make sure.


laanwj:
Makes sense, thanks for documenting this.

jonasschnelli:
utACK e5b449c3e8f3b27796629e4adfcb59ed0b212f80

PR #9497, opened 2017-01-09
CCheckQueue Unit Tests

JeremyRubin:
This PR builds on https://github.com/bitcoin/bitcoin/pull/9495 to unit test the CCheckQueue for correctness.

The cases covered in these tests are:

1. Standard usage
2. Failing checks are caught
3. Prior blocks failing don't interfere with future blocks
4. No Memory leakage (all check destructors are called before new blocks allowed, memory is freed).
5. Thread Safety

PR #8574, opened 2016-08-24
[Wallet] refactor CWallet/CWalletDB/CDB

jonasschnelli:
Try to hide `CDB`/`bitdb` behind `CWalletDB`.
This is a first chunk of changes (does not cover everything to avoid an oversized PR).

**Prepare for full wallet database abstraction which is required for smooth migration to a different wallet database format** (must be supported in parallel).

We probably shouldn't stay forever with Berkley-DB4.8.

PR #9911, opened 2017-03-03
Wshadow: various gcc fixes

paveljanik:
Additional `-Wshadow` fixes for gcc 4.8.5.

### DIFF
--- a/src/torcontrol.cpp
+++ b/src/torcontrol.cpp
@@ -662,8 +662,8 @@ void TorController::reconnect_cb(evutil_socket_t fd, short what, void *arg)
 }
 
 /****** Thread ********/
-struct event_base *gBase;
-boost::thread torControlThread;
+static struct event_base *gBase;
+static boost::thread torControlThread;
 
 static void TorControlThread()
 {
--- a/src/torcontrol.cpp
+++ b/src/torcontrol.cpp
@@ -662,26 +662,26 @@ void TorController::reconnect_cb(evutil_socket_t fd, short what, void *arg)
 }
 
 /****** Thread ********/
-struct event_base *base;
+struct event_base *gBase;
 boost::thread torControlThread;
 
 static void TorControlThread()
 {
-    TorController ctrl(base, GetArg("-torcontrol", DEFAULT_TOR_CONTROL));
+    TorController ctrl(gBase, GetArg("-torcontrol", DEFAULT_TOR_CONTROL));
 
-    event_base_dispatch(base);
+    event_base_dispatch(gBase);
 }
 
 void StartTorControl(boost::thread_group& threadGroup, CScheduler& scheduler)
 {
-    assert(!base);
+    assert(!gBase);
 #ifdef WIN32
     evthread_use_windows_threads();
 #else
     evthread_use_pthreads();
 #endif
-    base = event_base_new();
-    if (!base) {
+    gBase = event_base_new();
+    if (!gBase) {
         LogPrintf("tor: Unable to create event_base\n");
         return;
     }
@@ -691,18 +691,18 @@ void StartTorControl(boost::thread_group& threadGroup, CScheduler& scheduler)
 
 void InterruptTorControl()
 {
-    if (base) {
+    if (gBase) {
         LogPrintf("tor: Thread interrupt\n");
-        event_base_loopbreak(base);
+        event_base_loopbreak(gBase);
     }
 }
 
 void StopTorControl()
 {
-    if (base) {
+    if (gBase) {
         torControlThread.join();
-        event_base_free(base);
-        base = 0;
+        event_base_free(gBase);
+        gBase = 0;
     }
 }
 
--- a/src/streams.h
+++ b/src/streams.h
@@ -404,8 +404,8 @@ public:
         return (*this);
     }
 
-    void GetAndClear(CSerializeData &data) {
-        data.insert(data.end(), begin(), end());
+    void GetAndClear(CSerializeData &d) {
+        d.insert(d.end(), begin(), end());
         clear();
     }
 
--- a/src/test/checkqueue_tests.cpp
+++ b/src/test/checkqueue_tests.cpp
@@ -46,7 +46,7 @@ struct FakeCheckCheckCompletion {
 
 struct FailingCheck {
     bool fails;
-    FailingCheck(bool fails) : fails(fails){};
+    FailingCheck(bool _fails) : fails(_fails){};
     FailingCheck() : fails(true){};
     bool operator()()
     {
@@ -411,15 +411,15 @@ BOOST_AUTO_TEST_CASE(test_CheckQueueControl_Locks)
             std::unique_lock<std::mutex> l(m);
             tg.create_thread([&]{
                     CCheckQueueControl<FakeCheck> control(queue.get());
-                    std::unique_lock<std::mutex> l(m);
+                    std::unique_lock<std::mutex> ll(m);
                     has_lock = true;
                     cv.notify_one();
-                    cv.wait(l, [&]{return has_tried;});
+                    cv.wait(ll, [&]{return has_tried;});
                     done = true;
                     cv.notify_one();
                     // Wait until the done is acknowledged
                     //
-                    cv.wait(l, [&]{return done_ack;});
+                    cv.wait(ll, [&]{return done_ack;});
                     });
             // Wait for thread to get the lock
             cv.wait(l, [&](){return has_lock;});
--- a/src/test/script_tests.cpp
+++ b/src/test/script_tests.cpp
@@ -32,7 +32,7 @@
 // Uncomment if you want to output updated JSON tests.
 // #define UPDATE_JSON_TESTS
 
-static const unsigned int flags = SCRIPT_VERIFY_P2SH | SCRIPT_VERIFY_STRICTENC;
+static const unsigned int gFlags = SCRIPT_VERIFY_P2SH | SCRIPT_VERIFY_STRICTENC;
 
 unsigned int ParseScriptFlags(std::string strFlags);
 std::string FormatScriptFlags(unsigned int flags);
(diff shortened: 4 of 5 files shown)


paveljanik:
gcc version 5.3.1 20160301 [gcc-5-branch revision 233849] is completely `-Wshadow` clean.

gmaxwell (commented):
utACK (lightly reviewed)

MarcoFalke:
Any reason this is split into three commits with mostly the same
commit messages?

paveljanik:
Yes. It is WIP. I'm adding commit by commit for various gcc/clang version I test so people do not see warnings. I still have to test one more clang version I have installed on one system.

Wil squash afterwards of course.

paveljanik:
BTW, this PR is open also for others who see some `-Wshadow` warnings on their preferred compiler. Having all fixes in one PR should lower the impact of turning the warnings on by default.

laanwj:
Sigh, more Wshadow changes. Is this ever done?

paveljanik:
... and squashed as both clang version tested are `-Wshadow` clean.

`clang` seems to be much better or people use it more when developing for Bitcoin Core ;-)

laanwj:
> clang seems to be much better or people use it more when developing for Bitcoin Core ;-)

Aside: In my case, yes. FreeBSD defaults to clang nowadays, and on Linux I use clang specifically for Bitcoin Core because it compiles faster and uses less memory (=more parallelism with same amount of RAM).

GCC is supposed to produce slightly better code for x86 though, or at least that used to be the case years ago. So for production builds it makes sense to use that (which we do in gitian). I think @theuni was discussing switching all the builds to clang at some point though because it makes cross-compiling easier (also clang support for Win32 is supposed to be fairly good, whereas GCC is [giving us a lot of trouble](https://github.com/bitcoin/bitcoin/projects/1) recently.) But it's not possible to say generally that one of them is better.

dcousens (approved):
utACK

paveljanik:
Added fix for #8574's local variable redeclaration (aka shadowing).

gmaxwell:
paveljanik have you compiled with a make clean lately? I'm flooded with  streams.h:407:44: warning: declaration of ‘data’ shadows a member of 'this' [-Wshadow]

paveljanik:
@gmaxwell This should be already fixed in this PR - https://github.com/bitcoin/bitcoin/pull/9911/files#diff-eb3b977a68473a9d20093cbe90c659e6L407

BTW - I always clean compile...

jtimon:
utACK bfc6be9

src/torcontrol.cpp
```diff
@@ -662,26 +662,26 @@ void TorController::reconnect_cb(evutil_socket_t fd, short what, void *arg)
 }
 
 /****** Thread ********/
-struct event_base *base;
```
TheBlueMatt: What is this shadowing? (and can we make it static?)

paveljanik: With gcc4.8.5:
```
bitcoin@linux-5bz0:~/bitcoin> make CC=gcc
Making all in src
make[1]: Entering directory '/home/bitcoin/bitcoin/src'
make[2]: Entering directory '/home/bitcoin/bitcoin/src'
  CXX      libbitcoin_server_a-torcontrol.o
/usr/include/boost/date_time/time_system_counted.hpp: In instantiation of 'static boost::date_time::counted_time_system<time_rep>::time_rep_type boost::date_time::counted_time_system<time_rep>::add_days(const time_rep_type&, const date_duration_type&) [with time_rep = boost::date_time::counted_time_rep<boost::posix_time::millisec_posix_time_system_config>; boost::date_time::counted_time_system<time_rep>::time_rep_type = boost::date_time::counted_time_rep<boost::posix_time::millisec_posix_time_system_config>; boost::date_time::counted_time_system<time_rep>::date_duration_type = boost::gregorian::date_duration]':
/usr/include/boost/date_time/time.hpp:141:45:   required from 'boost::date_time::base_time<T, time_system>::time_type boost::date_time::base_time<T, time_system>::operator+(const date_duration_type&) const [with T = boost::posix_time::ptime; time_system = boost::date_time::counted_time_system<boost::date_time::counted_time_rep<boost::posix_time::millisec_posix_time_system_config> >; boost::date_time::base_time<T, time_system>::time_type = boost::posix_time::ptime; boost::date_time::base_time<T, time_system>::date_duration_type = boost::gregorian::date_duration]'
/usr/include/boost/date_time/posix_time/date_duration_operators.hpp:33:37:   required from here
torcontrol.cpp:665:20: warning: shadowed declaration is here [-Wshadow]
 struct event_base *base;
                    ^
/usr/include/boost/date_time/time_system_counted.hpp: In instantiation of 'static boost::date_time::counted_time_system<time_rep>::time_rep_type boost::date_time::counted_time_system<time_rep>::add_time_duration(const time_rep_type&, boost::date_time::counted_time_system<time_rep>::time_duration_type) [with time_rep = boost::date_time::counted_time_rep<boost::posix_time::millisec_posix_time_system_config>; boost::date_time::counted_time_system<time_rep>::time_rep_type = boost::date_time::counted_time_rep<boost::posix_time::millisec_posix_time_system_config>; boost::date_time::counted_time_system<time_rep>::time_duration_type = boost::posix_time::time_duration]':
/usr/include/boost/date_time/time.hpp:161:64:   required from 'boost::date_time::base_time<T, time_system>::time_type boost::date_time::base_time<T, time_system>::operator+(const time_duration_type&) const [with T = boost::posix_time::ptime; time_system = boost::date_time::counted_time_system<boost::date_time::counted_time_rep<boost::posix_time::millisec_posix_time_system_config> >; boost::date_time::base_time<T, time_system>::time_type = boost::posix_time::ptime; boost::date_time::base_time<T, time_system>::time_duration_type = boost::posix_time::time_duration]'
/usr/include/boost/date_time/posix_time/conversion.hpp:30:48:   required from here
torcontrol.cpp:665:20: warning: shadowed declaration is here [-Wshadow]
```

TheBlueMatt: Hmm, that doesnt really answer my question, thanks gcc...Is that trying to say that boost has some "base" object that we're shadowing? Either way lets make the new gBase static.

paveljanik: Hmm. I do not see any `base` there 8)

laanwj: Oh no, so we're at renaming random variables now without even understanding why just to shut up warnings.

laanwj:
To be honest I'd prefer to disable -Wshadow again. Even a few days after enabling it, it seems to me it requires too many disjointed, seemingly random changes, and continuous maintenance after almost every thing that gets merged after new incomprehensible warnings pop up with *some* gcc or clang version.

src/torcontrol.cpp
paveljanik: You are overreacting. The problem is elsewhere, of course. The original author is using `base` all over the source code and arguments of methods conflicts with the global (now static). I think that renaming the global and making it static is the best solution here. Yes, discovered because of turning on `-Wshadow`.

MarcoFalke:
We might keep it enabled, after all it could help to detect valid
bugs. Though, I agree that those "fix various gcc warnings" with no
apparent rationale other than "The warning disappears when compiling
with this specific version of boost and gcc-x.x.x" are not
particularly helpful. (I am not against all changes in this pull, but
we should only commit the fixes that make sense and keep the commits
itself at a frequency that does not hinder the other development
work.)

laanwj:
> We might keep it enabled, after all it could help to detect valid bugs.

It could, sure. In an ideal situation where compilers had a single, sane (or even close) definition of shadowing I'd be all for it. 
However I just searched and there are >25 recent pulls already dealing with Wshadow changes:  https://github.com/bitcoin/bitcoin/pulls?utf8=%E2%9C%93&q=is%3Apr%20is%3Aclosed%20wshadow
This is becoming a bit too much.
And these kind of changes could also result in bugs when applied carelessly, e.g. forgetting to check binaries for equality on some platform.

Also changes in boost or other dependencies could suddenly trigger these in our code. E.g. on the version of boost on FreeBSD I see various warnings inside boost too.

laanwj:
@gmaxwell noted that there are also sub-options for Wshadow. Maybe we can reduce the false positives/weird cases this way.
```
  -Wshadow-local which warns if a local variable shadows another local
  variable or parameter,

  -Wshadow-compatible-local which warns if a local variable shadows
  another local variable or parameter whose type is compatible with that
  of the shadowing variable.
```

paveljanik:
We had some false positive/weird case?

The `base` case is a gcc bug. It correctly identified shadowing but emitted wrong log lines...

paveljanik:
Rebased.

paveljanik:
@laanwj Can you please upload your FreeBSD build log for me with V=1 so I can investigate your warnings?

paveljanik:
Added fix for #9497's shadowing issues:

```
test/checkqueue_tests.cpp:49:23: warning: declaration shadows a field of 'checkqueue_tests::FailingCheck' [-Wshadow]
    FailingCheck(bool fails) : fails(fails){};
                      ^
test/checkqueue_tests.cpp:48:10: note: previous declaration is here
    bool fails;
         ^
test/checkqueue_tests.cpp:414:50: warning: declaration shadows a local variable [-Wshadow]
                    std::unique_lock<std::mutex> l(m);
                                                 ^
test/checkqueue_tests.cpp:411:42: note: previous declaration is here
            std::unique_lock<std::mutex> l(m);
                                         ^
2 warnings generated.
```

laanwj:
```
1201f4dac236eb92dade95ae8dcb832b374c38463e1180177433b7c500dd56a7  /tmp/compare/bitcoind.0cebfcffb480e63527c225bf9c46db06a3816d98.stripped
1201f4dac236eb92dade95ae8dcb832b374c38463e1180177433b7c500dd56a7  /tmp/compare/bitcoind.58e94bf6bb3981bca466ff83994ca2a06630a828.stripped
1201f4dac236eb92dade95ae8dcb832b374c38463e1180177433b7c500dd56a7  /tmp/compare/bitcoind.d9f7224cde0aa9415efd945bc1c9124628a59002.stripped
e813793f57842ece8ec04d101fb202a30414c3a6d6801b3038884e5efeb8d437  /tmp/compare/bitcoind.946ae52d0361058a1f5e70d48d6bec191217cf4d.stripped
e813793f57842ece8ec04d101fb202a30414c3a6d6801b3038884e5efeb8d437  /tmp/compare/bitcoind.ce01e6226ce52b88c644b0cb89852278edbdf33b.stripped
```
Looks like d9f7224cde0aa9415efd945bc1c9124628a59002 changes the bitcoind executable. 
ACK on the others, if you make a pull with just those I'll merge it.

paveljanik:
@laanwj Please look at d9f7224cde0aa9415efd945bc1c9124628a59002 ;-) Of course I can rewrite it by renaming the local variable instead of reusing the upper local which is not used after at all.

Is it OK for you? I have chosen this solution because it is crystal clean and less lines etc.

MarcoFalke:
Let's not reuse iterators. This will cause issues with nested for
loops, where the inner loops uses the iterator of the outer loop. I'd
rather prefer to keep it the same instead of changing one issue
against another.

paveljanik:
@MarcoFalke Can you please rephrase what do you prefer? Do you agree with renaming the first or the second `mi` to e.g. `mIter`?

MarcoFalke:
Change one of them to `mit`?

laanwj:
I'm not going to merge any shadowing fixes if they change the binary.

In this case I'd prefer if you did the renaming in a way that just renames
one of the iterators instead of reusing it which changes the code (thanks
MarcoFalke for checking)

On Mar 17, 2017 7:21 PM, "MarcoFalke" <notifications@github.com> wrote:

Change one of them to mit?

—
<https://github.com/bitcoin/bitcoin/pull/9911#issuecomment-287433600>, or mute
the thread

paveljanik:
@MarcoFalke Yes, but your example is not the code which is used right now (and this what the reason I did not understand what you wrote). There are two sequential uses, there is no loop in the second part and no nested loop at all. But yes, to be future-safe and same-binary-now, I have changed the first iterator name to `mit` in an additional commit for easier checking.

laanwj:
@paveljanik thanks, going to re-check.

PR #9916, opened 2017-03-04
Fix msvc compiler error C4146 (minus operator applied to unsigned type)

kobake:
On msvc14, the error C4146 (unary minus operator applied to unsigned type, result still unsigned) had been occured.
Minus operator should be applied to signed type.

### DIFF
--- a/src/bloom.cpp
+++ b/src/bloom.cpp
@@ -254,8 +254,8 @@ void CRollingBloomFilter::insert(const std::vector<unsigned char>& vKey)
         if (nGeneration == 4) {
             nGeneration = 1;
         }
-        uint64_t nGenerationMask1 = -(uint64_t)(nGeneration & 1);
-        uint64_t nGenerationMask2 = -(uint64_t)(nGeneration >> 1);
+        uint64_t nGenerationMask1 = 0 - (uint64_t)(nGeneration & 1);
+        uint64_t nGenerationMask2 = 0 - (uint64_t)(nGeneration >> 1);
         /* Wipe old entries that used this generation number. */
         for (uint32_t p = 0; p < data.size(); p += 2) {
             uint64_t p1 = data[p], p2 = data[p + 1];
--- a/src/test/util_tests.cpp
+++ b/src/test/util_tests.cpp
@@ -321,7 +321,7 @@ BOOST_AUTO_TEST_CASE(test_ParseInt32)
     BOOST_CHECK(ParseInt32("1234", &n) && n == 1234);
     BOOST_CHECK(ParseInt32("01234", &n) && n == 1234); // no octal
     BOOST_CHECK(ParseInt32("2147483647", &n) && n == 2147483647);
-    BOOST_CHECK(ParseInt32("-2147483648", &n) && n == -2147483648);
+    BOOST_CHECK(ParseInt32("-2147483648", &n) && n == (-2147483647 - 1)); // (-2147483647 - 1) equals INT_MIN
     BOOST_CHECK(ParseInt32("-1234", &n) && n == -1234);
     // Invalid values
     BOOST_CHECK(!ParseInt32("", &n));


src/bloom.cpp
```diff
@@ -254,8 +254,8 @@ void CRollingBloomFilter::insert(const std::vector<unsigned char>& vKey)
         if (nGeneration == 4) {
             nGeneration = 1;
         }
-        uint64_t nGenerationMask1 = -(uint64_t)(nGeneration & 1);
-        uint64_t nGenerationMask2 = -(uint64_t)(nGeneration >> 1);
+        uint64_t nGenerationMask1 = (uint64_t)(-(int64_t)(nGeneration & 1));
+        uint64_t nGenerationMask2 = (uint64_t)(-(int64_t)(nGeneration >> 1));
```
laanwj: This works around the error, but I don't think it's very elegant. What about just formulating it explicitly:
```
        uint64_t nGenerationMask1 = ~(uint64_t)(nGeneration & 1) + 1;
        uint64_t nGenerationMask2 = ~(uint64_t)(nGeneration >> 1) + 1;
```

laanwj: Or even just:
```
        uint64_t nGenerationMask1 = 0 - (uint64_t)(nGeneration & 1);
        uint64_t nGenerationMask2 = 0 - (uint64_t)(nGeneration >> 1);
```
Not sure this is any more readable than the above though :)

kobake: Thanks. I confirmed three ways work well as same in my local msvc14.
The latter you show is simple but not feel as explicit bit calculation.
I like the former that is readable clearly as two's complement formula, then I'll adopt it.

kobake: Applied. (commit --amend && push -f)

gmaxwell:
ugh. MSVC produces the most absurd errors for perfectly valid, strictly conforming code. I don't have a problem with silencing it in this way that the code currently is (though IMO it makes it less clear).

The prior fix of casting to an integer can actually result in undefined behavior (at least generally if not actually in our case due to the sizes we use). a=INT_MIN; a = -a; is undefined. (highlighting the foolishness of MSFT's war on bit manipulation. :P -- as someone trying to address this warning in an 'obvious way' would potentially introduce undefined behavior as a result!)

laanwj:
Agreed, that this is necessary at all is silly. There's no ambiguity here. Unary minus -x equivalent to 0-x which is well-defined also for unsigned values which are defined to wrap around.

kobake:
Yes, we should take care not to meet undefined behavior.
Now in this case, even if nGeneration has INT_MIN value, it's no problem because the target values to be applied two's complement formula are "(nGeneration & 1)" or "(nGeneration >> 1)"; they are 32bit values and their most significant bit is always zero, of course it's true even after the cast.

gmaxwell:
https://docs.microsoft.com/en-us/cpp/error-messages/compiler-warnings/compiler-warning-level-2-c4146

Seems to indicate that 0-(uint64_t)(nGeneration & 1); will silence it-- if so I think that would be preferable.

kobake:
OK, I just understood. I'll adopt ```0 - x``` style.
FYI, msvc14 (vs2015) treats C4146 not as warning but as error by default because of ```/sdl``` option.

laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/9916/commits/40aca33add8f108caa517a952f6c8555a63f6f17

laanwj:
Amended the commit message to put a newline between the title and body, otherwise `git log` in short formats doesn't work as expected.

kobake:
Oh, I see. I've inserted new line as you say. Thanks for helpful tips.

kobake:
I also fixed another C4146 error in the commit below that is about unit test.
https://github.com/kobake/bitcoin-msvc/commit/3c73d820e2355b36ea6bd6c4b78e4afa82637b8f

Should I contain the commit in this PR #9916, or create an another PR for the commit?

laanwj:
It's another minor thing, I'd say add it here.

kobake:
Thanks, I added.

laanwj:
re-UTack https://github.com/bitcoin/bitcoin/pull/9916/commits/8e0720bdb9141b00b4c00893b8139904c67ddacf

paveljanik:
utACK https://github.com/bitcoin/bitcoin/pull/9916/commits/8e0720bdb9141b00b4c00893b8139904c67ddacf

PR #9919, opened 2017-03-04
UNIX sockets support for RPC

laanwj:
Add functionality for RPC over UNIX sockets, on platforms that support UNIX sockets. This is specified with `:unix:/path/to/socket` to `-rpcbind` and `-rpcconnect`. By default the socket path is relative to the data directory.

For the server side this works as-is.

For the client side (bitcoin-cli) this requires a small patch to libevent to be able to pass in a file descriptor of an existing connection (https://github.com/libevent/libevent/pull/479).

Even without the client change this could be useful though. For example the Python RPC tests could connect through UNIX socket to avoid port collisions ~~(TODO: make a Python example)~~. This has been implemented.

Also the overhead of using UNIX sockets should be lower than using local TCP (see https://stackoverflow.com/questions/14973942/performance-tcp-loopback-connection-vs-unix-domain-socket).

### Example

Server:

```bash
src/bitcoind -printtoconsole -datadir=/store/tmp/testbtc -connect=0 -debug=rpc -debug=http -rpcbind=:unix:socket
...
2017-03-05 10:29:31 Binding RPC on UNIX socket /store/tmp/testbtc/socket
...
```

Client:

```bash
src/bitcoin-cli -rpcconnect=":unix:socket" -datadir=/store/tmp/testbtc getinfo
{
  "version": 149900,
  ...
}
```

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -136,6 +136,7 @@ BITCOIN_CORE_H = \
   support/allocators/zeroafterfree.h \
   support/cleanse.h \
   support/events.h \
+  support/evunix.h \
   support/lockedpool.h \
   sync.h \
   threadsafety.h \
@@ -211,6 +212,9 @@ libbitcoin_server_a_SOURCES = \
   validationinterface.cpp \
   versionbits.cpp \
   $(BITCOIN_CORE_H)
+if BUILD_EVUNIX
+libbitcoin_server_a_SOURCES += support/evunix.cpp
+endif
 
 if ENABLE_ZMQ
 libbitcoin_zmq_a_CPPFLAGS = $(BITCOIN_INCLUDES) $(ZMQ_CFLAGS)
@@ -377,6 +381,9 @@ bitcoind_LDADD += $(BOOST_LIBS) $(BDB_LIBS) $(SSL_LIBS) $(CRYPTO_LIBS) $(MINIUPN
 
 # bitcoin-cli binary #
 bitcoin_cli_SOURCES = bitcoin-cli.cpp
+if BUILD_EVUNIX
+bitcoin_cli_SOURCES += support/evunix.cpp
+endif
 bitcoin_cli_CPPFLAGS = $(AM_CPPFLAGS) $(BITCOIN_INCLUDES) $(EVENT_CFLAGS)
 bitcoin_cli_CXXFLAGS = $(AM_CXXFLAGS) $(PIE_FLAGS)
 bitcoin_cli_LDFLAGS = $(RELDFLAGS) $(AM_LDFLAGS) $(LIBTOOL_APP_LDFLAGS)
--- a/src/httpserver.cpp
+++ b/src/httpserver.cpp
@@ -663,16 +663,21 @@ CService HTTPRequest::GetPeer()
     evhttp_connection* con = evhttp_request_get_connection(req);
     CService peer;
     if (con) {
+#ifdef HAVE_SOCKADDR_UN
+        // evhttp has no way to query what bindsocket a connection
+        // came in on, so we need this low-level code to be able to
+        // correctly mark UNIX socket connections as coming from localhost.
+        struct bufferevent *bev = evhttp_connection_get_bufferevent(con);
+        if (bev && evunix_is_conn_from_unix(bev)) {
+            // As we have no way to signify "UNIX localhost" here, just return
+            // IPv6 localhost.
+            return LookupNumeric("::1", 0);
+        }
+#endif
         // evhttp retains ownership over returned address string
         const char* address = "";
         uint16_t port = 0;
         evhttp_connection_get_peer(con, (char**)&address, &port);
-        if (!strcmp(address, "localhost")) {
-            /* Special: will get here for UNIX sockets. As we have no way to indicate that,
-             * just pass localhost IPv6.
-             */
-            address = "::1";
-        }
         peer = LookupNumeric(address, port);
     }
     return peer;
--- a/src/httpserver.cpp
+++ b/src/httpserver.cpp
@@ -21,12 +21,14 @@
 #include <signal.h>
 #include <future>
 
+#include <boost/algorithm/string/predicate.hpp>
 #include <event2/event.h>
 #include <event2/http.h>
 #include <event2/thread.h>
 #include <event2/buffer.h>
 #include <event2/util.h>
 #include <event2/keyvalq_struct.h>
+#include "support/evunix.h"
 
 #ifdef EVENT__HAVE_NETINET_IN_H
 #include <netinet/in.h>
@@ -181,6 +183,10 @@ static WorkQueue<HTTPClosure>* workQueue = 0;
 std::vector<HTTPPathHandler> pathHandlers;
 //! Bound listening sockets
 std::vector<evhttp_bound_socket *> boundSockets;
+#ifdef HAVE_SOCKADDR_UN
+//! UNIX sockets to clean up on shutdown
+std::vector<boost::filesystem::path> cleanupUNIXSockets;
+#endif
 
 /** Check if a network address is allowed to access the HTTP server */
 static bool ClientAllowed(const CNetAddr& netaddr)
@@ -331,6 +337,8 @@ static bool HTTPBindAddresses(struct evhttp* http)
     } else if (mapMultiArgs.count("-rpcbind")) { // Specific bind address
         const std::vector<std::string>& vbind = mapMultiArgs.at("-rpcbind");
         for (std::vector<std::string>::const_iterator i = vbind.begin(); i != vbind.end(); ++i) {
+            if (boost::starts_with(*i, RPC_ADDR_PREFIX_UNIX)) // Skip UNIX sockets here
+                continue;
             int port = defaultPort;
             std::string host;
             SplitHostPort(*i, port, host);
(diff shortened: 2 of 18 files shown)


src/Makefile.am
```diff
@@ -377,6 +381,9 @@ bitcoind_LDADD += $(BOOST_LIBS) $(BDB_LIBS) $(SSL_LIBS) $(CRYPTO_LIBS) $(MINIUPN
 
 # bitcoin-cli binary #
 bitcoin_cli_SOURCES = bitcoin-cli.cpp
+if BUILD_EVUNIX
```
laanwj: @theuni This is pretty ugly, but I don't quite know how to do this otherwise. My first intuition was to add the file to `libbitcoin_util_a_SOURCES`, however these can't depend on libevent as that is shared between all executables. There is no library that is shared between just `bitcoind` and `bitcoin-cli` and adding one just for this seems overkill.

gmaxwell:
Neat.

paveljanik:
This does not compile with `libevent2-2.0.22` which is OK with the current master.

```
bitcoin-cli.cpp:214:13: error: use of undeclared identifier `evhttp_connection_base_bufferevent_new'
            evhttp_connection_base_bufferevent_new(base.get(), __null, bev, "", 0)
            ^
1 error generated.
```

laanwj:
@paveljanik Yes, that needs a version `#ifdef`, that call doesn't exist in older versions of libevent, and on newer ones (until my patch lands) it ignores the current connection and tries to open a new one so that trick doesn't work. I'd recommend testing with my patched libevent. Alternatively comment out the bitcoin-cli change and test connecting with something else:
- `nc -U <socket>` works, for example
- curl also has a ` --unix-socket` parameter, but haven't tried it yet

TheBlueMatt:
For merge we should probably just not have the client version, then?

Still, this is cool - lots of better auth controls can be applied to
sockets than IPs :)

laanwj:
> For merge we should probably just not have the client version, then?

Would prefer putting it behind `#ifdef EXPERIMENTAL_LIBEVENT` or such, with a comment that it can be replaced with the appropriate version comparison when it lands in a version. That way, people that compile their own stuff can use it already.

laanwj:
> Still, this is cool - lots of better auth controls can be applied to
sockets than IPs :)

Indeed. Default permissions of the socket will be `700` due to our umask setting, which seems okay:
```
srwx------  1 user user          0 Mar  5 11:40 socket
```

jonasschnelli:
This is great!
Concept ACK.

laanwj:
Added a commit that makes the tests' RPC run over UNIX sockets on platforms that support it, except for a few tests that make explicit assumptions about running over TCP/IP.

jnewbery:
I've been trying to test this locally by running the qa tests. Something that I do quite often is to run the tests with --noshutdown, wait for the test to fail, and use bitcoin-cli to interactively run RPCs on the test nodes:

`bitcoin.conf -conf=<test temp directory>/node<x>/bitcoin.conf`

With this PR, I'm not able to do that anymore because the conf file doesn't contain the username/password, and even when I add them back in, I get the error:

```
/tmp/user/1000/test_6o70u2b/28404/node0
→ ~/bitcoin/src/bitcoin-cli -rpcconnect=":unix:rpc_socket" -datadir=/tmp/user/1000/test_6o70u2b/28404/node0 help
[warn] evhttp_connection_connect_: connection to "" failed: No such file or directory
error: couldn't connect to server: unknown (code -1)
(make sure server is running and you are connecting to the correct RPC port)
/tmp/user/1000/test_6o70u2b/28404/node0
→ ll
total 16
drwxrwxr-x 3 ubuntu ubuntu 4096 Mar  6 23:31 ./
drwxrwxr-x 4 ubuntu ubuntu 4096 Mar  6 23:11 ../
-rw-rw-r-- 1 ubuntu ubuntu   79 Mar  6 23:20 bitcoin.conf
drwx------ 5 ubuntu ubuntu 4096 Mar  6 23:26 regtest/
srwx------ 1 ubuntu ubuntu    0 Mar  6 23:11 rpc_socket=
```

Do you have a better way of connecting to a test node during qa testing?

laanwj:
> With this PR, I'm not able to do that anymore because the conf file doesn't contain the username/password, and even when I add them back in, I get the error:

Right, I could change it to add the options to the bitcoin.conf. 

> And even when I add them back in, I get the error:

The above should work, though. That's exactly how I've been testing it. The connection failure is strange. Did you patch libevent so that `bitcoin-cli` is able to connect to UNIX sockets?

laanwj:
@jnewbery Okay, added two commits:
- add `BITCOIN_TEST_RPC_TCP` environment variable to force tests to use TCP instead of UNIX sockets for RPC
- write connection info to `bitcoin.conf` instead of providing it on the command line

laanwj:
Needed rebase, so rebased and squashed the squasme: commits.

jonasschnelli:
F.I.Y:
I wanted to test this via binaries built in gitian.
I could start the daemon (wanted to test in conjunction with Q): ` ./bitcoin-0.14.99/bin/bitcoin-qt -debug=rpc -debug=http -rpcbind=nix:socket --regtest`

However the CLI responded with:
`error: RPC was asked to connect to a UNIX socket. However, the version of libevent that this utility was compiled with has no support for that.`

Maybe we should bump libevent in the dependencies directory?

laanwj:
> However the CLI responded with:

Yes, that's expected. `bitcoin-cli` support needs this patch: https://github.com/laanwj/libevent/commit/403b793771341460b262edb1abc31fae267174bd.patch  as well as defining `-DLIBEVENT_EXPERIMENTAL`

You should still be able to use the socket with other clients, for example from Python or with curl.

laanwj:
Rebased (RPC tests logging changes #9768)

JeremyRubin:
Oops, I accidentally did my review of this code on the other PR. See here https://github.com/bitcoin/bitcoin/pull/9979#pullrequestreview-31071186

ryanofsky (commented):
utACK 695e854a0444bb5c81b1422494451f17db0b5f1e, though code is pretty out of date

jnewbery:
I've had a quick read through the changes to the functional tests, but I'll defer doing a full review until this is rebased. A few general comments:

- I'd prefer a command line argument over and environment variable for `BITCOIN_TEST_RPC_TCP`, partly because adding a command line argument means the help text will be updated and the option will be better documented.
- there's a response to your libevent PR here: https://github.com/libevent/libevent/pull/479. It'd be good to the functionality merged into libevent so we can avoid the `LIBEVENT_EXPERIMENTAL` stuff
- I also think it would be better to make the changes to the functional test framework after TestNode changes in #10082 are in. I really dislike having state for nodes floating around in util.py (eg the new `conninfo_for()` function)

src/Makefile.am
theuni: I think this is OK. I think we'll end up with a few wrappers for libevent, so moving those to their own internal helper lib will  make sense.

src/httpserver.cpp
```diff
@@ -619,6 +667,12 @@ CService HTTPRequest::GetPeer()
         const char* address = "";
         uint16_t port = 0;
         evhttp_connection_get_peer(con, (char**)&address, &port);
+        if (!strcmp(address, "localhost")) {
+            /* Special: will get here for UNIX sockets. As we have no way to indicate that,
+             * just pass localhost IPv6.
+             */
+            address = "::1";
```
theuni: +1. This seems like another good use-case for NET_INTERNAL from #10446 (or something similar). As a bonus, that would allow us to use associative containers for unix socket addresses, since each path would have its own hash.

src/support/evunix.cpp
```diff
@@ -0,0 +1,107 @@
...
+    }
+    int fd = socket(AF_UNIX, SOCK_STREAM, 0);
+    if (fd < 0) {
+        return -1;
+    }
+    if (connect(fd, (struct sockaddr*)&addr, sizeof(addr)) < 0) {
+        close(fd);
+        return -1;
+    }
+    evutil_make_socket_nonblocking(fd);
```
theuni: need to disable SIGPIPE here too?

src/httpserver.cpp
```diff
@@ -505,6 +547,12 @@ void StopHTTPServer()
         event_base_free(eventBase);
         eventBase = 0;
     }
+#ifdef HAVE_SOCKADDR_UN
+    //! Clean up UNIX sockets on shutdown
+    for (const auto& path: cleanupUNIXSockets) {
+         evunix_remove_socket(path);
```
theuni: should probably .clear() in case this gets called twice.

theuni (commented):
Concept ACK. I added a few comments, and @ryanofsky beat me to several others.

laanwj:
Closing this for now. I'll probably pick it up again for 0.16.

Sjors:
Concept ACK. c-lightning does this too FWIW.

This was [briefly discussed](http://www.erisian.com.au/bitcoin-core-dev/log-2018-10-25.html#l-901) during the meeting today. Tag as Up for Grabs?

MarcoFalke:
Marked as "up for grabs"

laanwj:
The server side should be easy to pick up.
As for the client side I have opened a libevent issue to track this here: https://github.com/libevent/libevent/issues/907
Unfortunately there is very little interest from their side for this.

vinipsmaker:
Does it have to be HTTP? Would it be easier to implement if it was pure (line delimited) JSON-RPC over UNIX domain sockets?

githubnew-pro:
> قابلیت های RPC را از طریق سوکت های UNIX ، در سیستم عامل های پشتیبانی از سوکت های UNIX ، اضافه کنید. این با `:unix:/path/to/socket`to `-rpcbind`و مشخص شده است `-rpcconnect`. به طور پیش فرض مسیر سوکت نسبت به فهرست داده است.
> 
> برای سمت سرور این همانطور که هست کار می کند.
> 
> برای طرف مشتری (bitcoin-cli) این نیاز به یک وصله کوچک دارد تا بتواند در توصیف پرونده اتصال موجود ( [libevent / libevent # 479](https://github.com/libevent/libevent/pull/479) ) عبور کند.
> 
> حتی بدون تغییر مشتری ، این می تواند مفید باشد. به عنوان مثال آزمایش های RPC پایتون می توانند از طریق سوکت UNIX متصل شوند تا از برخورد بندر جلوگیری شود~(TODO: یک مثال پایتون بسازید)~. این اجرا شده است.
> 
> همچنین سربار استفاده از سوکتهای UNIX باید کمتر از استفاده از TCP محلی باشد (به https://stackoverflow.com/questions/14973942/performance-tcp-loopback-connection-vs-unix-domain-socket مراجعه [کنید](https://stackoverflow.com/questions/14973942/performance-tcp-loopback-connection-vs-unix-domain-socket) ).
> 
> ### مثال
> سرور:
> 
> ```shell
> src / bitcoind -printtoconsole -datadir = / store / tmp / testbtc -connect = 0 -debug = rpc -debug = http -rpcbind =: unix: سوکت
> ...
> 05/07/2017 10:29:31 اتصال RPC در سوکت / فروشگاه / tmp / testbtc / سوکت UNIX
> ...
> ```
> 
> مشتری:
> 
> ```shell
> src / bitcoin-cli -rpcconnect = " : unix: socket " -datadir = / store / tmp / testbtc getinfo
> {
>   " نسخه " : 149900 ،
>   ...
> }
> ```

michaelfolkson:
> Does it have to be HTTP? Would it be easier to implement if it was pure (line delimited) JSON-RPC over UNIX domain sockets?

This was discussed today on [IRC](https://gnusha.org/bitcoin-core-dev/2021-09-22.log). In summary, this is what c-lightning does for its UNIX socket support and what Core could potentially do in future for UNIX socket support as an alternative approach to getting a PR merged in libevent. Any non HTTP based UNIX socket support would need to be in addition to the existing HTTP based JSON-RPC rather than as a replacement.

ryanofsky:
> i don't see any problem personally with adding support for unix sockets rpc server / client that uses plain json-rpc lines, if someone wants to work on that

This seems ok, but without URLs, you would have to decide how to handle wallet endpoints. Maybe create different sockets for different wallets like `<datadir>/wallets/<walletname>/rpc.sock`, or use one socket and accept an extra wallet RPC param.

JeremyRubin:
> Any non HTTP based UNIX socket support would need to be in addition to the existing HTTP based JSON-RPC rather than as a replacement.

contestable, i think it would be great to deprecate then remove http support & leave it up to external proxy services.

PR #9920, opened 2017-03-05
[0.14] Allow any subkey in verify-commits (BlueMatt)

maflcko:
This replaces the fingerprints of the trusted keys which are subkeys by the fingerprint of their root key and then relaxes verify-commits to accept the root key or any of its subkeys.

Stolen from #9880 

@laanwj This should fix the travis issue. You can merge the same branch to master via https://github.com/bitcoin/bitcoin/compare/master...MarcoFalke:Mf1703-verifySubkeys


laanwj:
This would prohibit designating certain subkeys for the specific purpose of signing commits? On the other hand the added overhead (and travis failures) of keeping track of that on the project level may not be worth it. So I'm okay with this.

MarcoFalke:
Jup, the script no longer encourages you to use only the specifically hardcoded subkey for signing commits. You are now free to create several subkeys and use any of them (or even the root key, which I would not recommend).

laanwj:
ACK, verified that the added key fingerprints 133EAC179436F14A5CF1B794860FEB804E669320 B8B3F1C0E58C15DB6A81D30C3648A882F4316B9B are the correct ones.

jonasschnelli:
Post merge tested ACK 589cd63b303158b565d49d2ce26c9e46b0bba8bd.
Verfied sipa's (133EAC179436F14A5CF1B794860FEB804E669320) and MarcoFalkes (B8B3F1C0E58C15DB6A81D30C3648A882F4316B9B ) new key.

PR #9921, opened 2017-03-05
build: Probe MSG_DONTWAIT in the same way as MSG_NOSIGNAL

laanwj:
#### build: Probe MSG_DONTWAIT in the same way as MSG_NOSIGNAL

Instead of the WIN32-specific workaround, detect lack of `MSG_DONTWAIT` in the build system. This allows other platforms without `MSG_DONTWAIT` to work too.

#### build: cleanup: define MSG_DONTWAIT/MSG_NO_SIGNAL locall

Define MSG_DONTWAIT and MSG_NO_SIGNAL in the implementation files that use them (`net.cpp` and `netbase.cpp`), instead of compat.h which is included all over the place. 

This avoids putting them in the global namespace, as defining them as 0 is a hack that works for our specific usage, but it is not a general solution. 

Also makes sure they are defined only once so the `!defined(MSG_x)` guard can go.

#### compat: use `unsigned int` instead of `u_int`

`u_int` (without size specifier) is not available on some platforms (not sure what standard it's supposed to be part of), we don't use it anywhere else, and it doesn't hurt to simply write `unsigned int` out here.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -558,6 +558,15 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#include <sys/socket.h>]],
  [ AC_MSG_RESULT(no)]
 )
 
+dnl Check for MSG_DONTWAIT
+AC_MSG_CHECKING(for MSG_DONTWAIT)
+AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#include <sys/socket.h>]],
+ [[ int f = MSG_DONTWAIT; ]])],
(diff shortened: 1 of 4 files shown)


theuni:
Whoops, missed the assignment here (saw your IRC ping though, thanks!). utACK a4d1c9f0417e25b0aac937f07edf2bb642b84f09. Yes, Much of this stuff goes away with libevent. I like the move out of compat as well. For anything that libevent doesn't end up absorbing, it'd be nice to make everything else local the same way.

PR #9924, opened 2017-03-05
[UI Styling] Left-Align Tab Bar - optionsdialog.ui

SCDeveloper:
Aesthetics.

### DIFF
--- a/src/qt/forms/optionsdialog.ui
+++ b/src/qt/forms/optionsdialog.ui
@@ -13,6 +13,10 @@
   <property name="windowTitle">
    <string>Options</string>
   </property>
+  <property name="styleSheet">
+   <string notr="true">QTabWidget:tab-bar { alignment: left }
+QTabBar { alignment: left; }</string>
+  </property>
   <property name="modal">
    <bool>true</bool>
   </property>


SCDeveloper:
screen shot 2017-03-05 at 20 40 07

SCDeveloper:
Merge alongside #9925

luke-jr:
It's already left-aligned on platforms that use left-aligned tab bars... Your screenshot looks like macOS. Does macOS not match what our tab bar looks like?

SCDeveloper:
No this is a bug on OSX, and I have fixed it :D

jonasschnelli:
I don't think we should do this.
OSX uses a center-aligned tab-bar.

Example form the settings:
bildschirmfoto 2017-03-06 um 09 27 44

jonasschnelli:
I don't think we should do this.
OSX uses a center-aligned tab-bar.

Example form the settings:
bildschirmfoto 2017-03-06 um 09 27 44

SCDeveloper:
Well I think you should considering that it appears left-aligned on all other operating systems.

It also looks a damn site better.

laanwj:
I agree with @jonasschnelli, we should simply use the OS-provided layout. On MacOSX it's what users expect.

SCDeveloper:
I am an OSX user and it isn't what I expected.

luke-jr:
> OK, then let the OS simply control how the application is styled and relinquish control over it. 

That's exactly what we (and any reasonable software) do...

SCDeveloper:
Even if it looks bad?

laanwj:
> Even if it looks bad and differs from how it looks on all the other operating systems?

Yes. Why would someone using some platform care how it looks in other platforms at all?

laanwj:
Closing these, we disagree on the direction the GUI should take so it's not going to go anywhere.

SCDeveloper:
Because people don't just use one operating system. I have Windows, Linux and OSX.

luke-jr:
If you dislike how it looks, reconfigure your OS or use one that lets you.

PR #9926, opened 2017-03-06
Rename argument name 'data' to prevent shadowing.

kobake:
Fix gcc warning such as ```streams.h:407:44: warning: declaration of ‘data’ shadows a member of 'this'```.

The fixing file "streams.h" is included in so many sources and I saw the number of the warnings were 89.

### DIFF
--- a/src/streams.h
+++ b/src/streams.h
@@ -404,8 +404,8 @@ public:
         return (*this);
     }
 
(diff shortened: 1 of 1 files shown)


MarcoFalke:
I think this is done in #9911

kobake:
Just as you say. Thanks.

PR #9928, opened 2017-03-06
Allow verify-commit.sh to just verify the HEAD commit (Use non-recursive verification by default)

jonasschnelli:
Currently, `verify-commits.sh` leads to a segfault 11 on my machine because of the limits of recursions (proof is [here](https://bitcoin.jonasschnelli.ch/st.txt)).

Until we have ported this to python (or fixed it in another way), we should offer a only-head-commit verification.

### DIFF
--- a/contrib/verify-commits/verify-commits.sh
+++ b/contrib/verify-commits/verify-commits.sh
@@ -25,10 +25,14 @@ IS_SIGNED () {
 	if ! git -c "gpg.program=${DIR}/gpg.sh" verify-commit $1 > /dev/null 2>&1; then
 		return 1;
 	fi
+  if [ $2 = false ]; then
+    echo "Only verify HEAD commit signature"
+    return 0;
+  fi
 	local PARENTS
 	PARENTS=$(git show -s --format=format:%P $1)
-	for PARENT in $PARENTS; do
-		if IS_SIGNED $PARENT; then
+	for PARENT in $PARENT 1; do
+		if IS_SIGNED $PARENT $2; then
 			return 0;
 		fi
 		break
@@ -44,13 +48,19 @@ IS_SIGNED () {
 	return 1;
 }
 
-if [ x"$1" = "x" ]; then
-	TEST_COMMIT="HEAD"
+if [ "$1" = "yes" ] || [ "$1" = "1" ]; then
+	RECURSIVE=true
 else
-	TEST_COMMIT="$1"
+	RECURSIVE=false
 fi
 
-IS_SIGNED "$TEST_COMMIT"
+if [ x"$2" = "x" ]; then
+	TEST_COMMIT="HEAD"
+else
+	TEST_COMMIT="$2"
+fi
+
+IS_SIGNED "$TEST_COMMIT" $RECURSIVE
 RES=$?
 if [ "$RES" = 1 ]; then
 	if ! "$HAVE_FAILED"; then


laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/9928/commits/08473ea60e667a80f7bc06696b6039ccd7b248d0

TheBlueMatt:
Instead of defaulting to only enforce the top commit be signed, can we just remove the recursion? I changed it a while back to only follow the first merge branch which enforces structure on our git tree as well as checks signatures all the way down, so this should be trivial.

contrib/verify-commits/verify-commits.sh
```diff
@@ -25,10 +25,14 @@ IS_SIGNED () {
 	if ! git -c "gpg.program=${DIR}/gpg.sh" verify-commit $1 > /dev/null 2>&1; then
 		return 1;
 	fi
+  if [ $2 = false ]; then
```
TheBlueMatt: The rest of the file uses tabs, you've introduced spaces.

laanwj:
> Instead of defaulting to only enforce the top commit be signed, can we just remove the recursion?

Sure, go ahead. The only point here is to stop travis from failing ASAP. Checking just the top commit is a workaround until the script can be fixed properly. 

If you want to properly fix the script in one go, that's great, but please do it soon because the failling Travis is an annoyance and people will just ignore this check if it reports false positive after false positive.

MarcoFalke:
@laanwj, travis specifically fails because bbd7579 and 589cd63 are not
yet merged into master, not due to excessive recursion. Those are part
of #9880, so imo either #9880 should be merged or the two commits
cherry-picked into a separate pull.

laanwj:
@MarcoFalke Thanks. It was not clear to me that #9880 was a fix. I held off merging it because I first wanted it to pass again before making the test more comphrehensive.

MarcoFalke:
Note that the checks added in #9880 are off by default.

MarcoFalke:
Concept NACK. It is important that all commits are properly signed, changing it such that only HEAD needs to be signed is a loss in security by default.

laanwj:
> Concept NACK. It is important that all commits are properly signed, changing it such that only HEAD needs to be signed is a loss in security by default.

It was still a better security check than failing-all-the-time.

Anyhow, closing in favor of #9932

MarcoFalke:
We should not use travis for security checks at all. I think the check was
added to give maintainers a timely warning in case of an erroneous push.

MarcoFalke:
I think @jonasschnelli's issue is not yet solved. Can you check with master
again, Jonas?

jonasschnelli:
Two problems I have on my local machine:
* `./contrib/verify-commits/verify-commits.sh: line 68: sha512sum: command not found` (fixed it by changing `sha512sum` into `shasum -a 512`.
* `./contrib/verify-commits/verify-commits.sh 
Using verify-commits data from /Users/jonasschnelli/Documents/bitcoin/bitcoin/./contrib/verify-commits
Segmentation fault: 11`
  (still have the recursive limit crash)

Either we remove the recursion (how easy?) or we migrate it to python. The recursion problem seems to be also happen on non-OSX operating systems.

TheBlueMatt:
Yea, I think we should remove the recursion (should be easy, I'll look later today) and have a fallback for missing sha512sum to try the perl shasum. I'm not convinced we should only check the top commit.

PR #9930, opened 2017-03-06
Trivial: Correct indentation and bracing

matthias-g:
This corrects the indentation after a while loop was removed in c5a8b1b946b1ab0bb82bd4270b2a40f5731abcff

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -708,7 +708,7 @@ bool CNode::ReceiveMsgBytes(const char *pch, unsigned int nBytes, bool& complete
             handled = msg.readData(pch, nBytes);
 
         if (handled < 0)
-                return false;
+            return false;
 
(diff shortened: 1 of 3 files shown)


laanwj:
There are some more sillly uses of indentation/bracing left over in network code, for example: 
https://github.com/bitcoin/bitcoin/blob/master/src/net.cpp#L1295

Agree this needs to be cleaned up at some point, let's try to do it in one pass.

fanquake:
Note that this isn't a move-only change, two sets of brackets are being removed. Add ?w=1 to the URL to see.

MarcoFalke:

fanquake:
Needs a rebase.

laanwj:
Thanks. utACK 31a14d4

PR #9937, opened 2017-03-07
rpc: Prevent `dumpwallet` from overwriting files

### LINKED ISSUE #9934: dumpwallet walletfile overwrite footgun
19:02 < $user> gmaxwell: I have the new version. I opened the console. I ran dumpwallet, and it's not giving anything back to me. I passed wallet.dat and /Users/pawn/Library/Application\ Support/Bitcoin/wallet.dat as an argument.

laanwj:
Prevent arbitrary files from being overwritten by `dumpwallet`. There have been reports that users have overwritten wallet files this way. It may also avoid other security issues.

Fixes #9934. Adds mention to release notes and adds a test.

### DIFF
--- a/src/wallet/rpcdump.cpp
+++ b/src/wallet/rpcdump.cpp
@@ -600,7 +600,7 @@ UniValue dumpwallet(const JSONRPCRequest& request)
     if (request.fHelp || request.params.size() != 1)
         throw std::runtime_error(
             "dumpwallet \"filename\"\n"
-            "\nDumps all wallet keys in a human-readable format.\n"
+            "\nDumps all wallet keys in a human-readable format to a server-side file. This does not allow overwriting existing files.\n"
             "\nArguments:\n"
             "1. \"filename\"    (string, required) The filename with path (either absolute or relative to bitcoind)\n"
             "\nResult:\n"
@@ -616,9 +616,19 @@ UniValue dumpwallet(const JSONRPCRequest& request)
 
     EnsureWalletIsUnlocked(pwallet);
 
-    std::ofstream file;
     boost::filesystem::path filepath = request.params[0].get_str();
     filepath = boost::filesystem::absolute(filepath);
+
+    /* Prevent arbitrary files from being overwritten. There have been reports
+     * that users have overwritten wallet files this way:
+     * https://github.com/bitcoin/bitcoin/issues/9934
+     * It may also avoid other security issues.
+     */
+    if (boost::filesystem::exists(filepath)) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, filepath.string() + " already exists. If you are sure this is what you want, move it out of the way first");
+    }
+
+    std::ofstream file;
     file.open(filepath.string().c_str());
     if (!file.is_open())
         throw JSONRPCError(RPC_INVALID_PARAMETER, "Cannot open wallet dump file");
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -83,6 +83,9 @@ Low-level RPC changes
   * `getwalletinfo`
   * `getmininginfo`
 
+- `dumpwallet` no longer allows overwriting files. This is a security measure
+  as well as prevents dangerous user mistakes.
+
 Credits
 =======
 
--- a/test/functional/wallet-dump.py
+++ b/test/functional/wallet-dump.py
@@ -7,7 +7,7 @@
 import os
 
 from test_framework.test_framework import BitcoinTestFramework
-from test_framework.util import assert_equal
+from test_framework.util import (assert_equal, assert_raises_jsonrpc)
 
 
 def read_dump(file_name, addrs, hd_master_addr_old):
@@ -104,6 +104,9 @@ class WalletDumpTest(BitcoinTestFramework):
         assert_equal(found_addr, test_addr_count)
         assert_equal(found_addr_chg, 90*2 + 50)  # old reserve keys are marked as change now
         assert_equal(found_addr_rsv, 90*2) 
+
+        # Overwriting should fail
+        assert_raises_jsonrpc(-8, "already exists", self.nodes[0].dumpwallet, tmpdir + "/node0/wallet.unencrypted.dump")
 
 if __name__ == '__main__':
     WalletDumpTest().main ()


sipa:
concept ACK

jonasschnelli:
Important fix.
utACK 1307cd79b859f01a35d68dea9a674e56451f75f5.
IMO this is a backport candidate.

MarcoFalke:
utACK 1307cd79b859f01a35d68dea9a674e56451f75f5

paveljanik:
The change as written now allows to test the existence of any file.

gmaxwell:
only downside I see to this is an in automated backup that leaves files behind, your backup will stop and you might not notice. (I think that is a preferable failure mode to what we have now)

laanwj:
> The change as written now allows to test the existence of any file.

That seems preferable to overwriting any file, right? I tend to agree that being able to write arbitrary non-existing files is still a security risk, but it is tons better than what we have now.

Do you have a suggestion that would avoid this usage but still offers the same functionality?

> only downside I see to this is an in automated backup that leaves files behind, your backup will stop and you might not notice. (I think that is a preferable failure mode to what we have now)

Suggestion: simply append a counter. This is safer, too.

Note that it throws an exception when the attempt fails, so if you don't notice you're not handling errors at all, which is dangerous for a ton of other reasons too.

laanwj:
Updated the RPC help, too.

laanwj:
> Do you have a suggestion that would avoid this usage but still offers the same functionality?

We could make the check even stricter, only allowing writing of files within the data directory, or even better a `dumpwallet` directory within the data directory. But that has a much larger impact to the API. Could do that for 0.15 but certainly not for backporting.

paveljanik:
What about this: if there is a file already present, add a timestamp to the file name and save.

laanwj:
Yeah that could work...

laanwj:
Closing this, feel free to implement a different solution, have no time for this at the moment.

laanwj:
Reopening this, I think this is basically a good fix, and was closed for unimportant concerns.

src/wallet/rpcdump.cpp
```diff
@@ -616,9 +616,19 @@ UniValue dumpwallet(const JSONRPCRequest& request)
...
-    std::ofstream file;
     boost::filesystem::path filepath = request.params[0].get_str();
     filepath = boost::filesystem::absolute(filepath);
+
+    /* Prevent arbitrary files from being overwritten. There have been reports
+     * that users have overwritten wallet files this way:
+     * https://github.com/bitcoin/bitcoin/issues/9934
+     * It may also avoid other security issues.
+     */
+    if (boost::filesystem::exists(filepath)) {
```
promag: What if a `dumpwallet` is executed after this check with the same filename, or the filename is created by other means? IMO we should use `fopen` with `wx` mode where `x` stands for:

> File access mode flag "x" can optionally be appended to "w" or "w+" specifiers. This flag forces the function to fail if the file exists, instead of overwriting it. (C11)

and drop `exists()` test.

promag:
And there is also the option to make the filename optional, where the default value is `/some/base/path/ + wallet-name + some-timestamp + .extension` and return the filename.

src/wallet/rpcdump.cpp
```diff
@@ -600,7 +600,7 @@ UniValue dumpwallet(const JSONRPCRequest& request)
     if (request.fHelp || request.params.size() != 1)
         throw std::runtime_error(
             "dumpwallet \"filename\"\n"
-            "\nDumps all wallet keys in a human-readable format.\n"
+            "\nDumps all wallet keys in a human-readable format to a server-side file. This does not allow overwriting existing files.\n"
             "\nArguments:\n"
             "1. \"filename\"    (string, required) The filename with path (either absolute or relative to bitcoind)\n"
```
jnewbery: This help text is misleading. The relative path is relative to the directory from where bitcoind was started.

(really I think we should disallow relative paths entirely unless they're relative to some fixed point like the datadir)

src/wallet/rpcdump.cpp
```diff
@@ -616,9 +616,19 @@ UniValue dumpwallet(const JSONRPCRequest& request)
 
     EnsureWalletIsUnlocked(pwallet);
 
-    std::ofstream file;
     boost::filesystem::path filepath = request.params[0].get_str();
     filepath = boost::filesystem::absolute(filepath);
+
+    /* Prevent arbitrary files from being overwritten. There have been reports
```
jnewbery: nit: I don't think this needs such a long comment. Just say we don't allow files to be overwritten for safety.

src/wallet/rpcdump.cpp
```diff
@@ -616,9 +616,19 @@ UniValue dumpwallet(const JSONRPCRequest& request)
...
+     * that users have overwritten wallet files this way:
+     * https://github.com/bitcoin/bitcoin/issues/9934
+     * It may also avoid other security issues.
+     */
+    if (boost::filesystem::exists(filepath)) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, filename.string() + " already exists. If you are sure this is what you want, move it out of the way first");
```
jnewbery: `filename` should be `filepath`?

jnewbery (commented):
Concept ACK, but currently fails to build :(

A few comments inline.

src/wallet/rpcdump.cpp
laanwj: I agree relative paths should be relative to the datadir. However I am not changing that functionality here, and am also not touching this comment here.

laanwj:
> And there is also the option to make the filename optional, where the default value is /some/base/path/ + wallet-name + some-timestamp + .extension and return the filename.

Yes, many other things are possible. I'm not going to do that here. That's the reason I closed this before. **However** I only want to solve the overwrite problem here, which is not solved by making up filenames, only by refusing to overwrite files.

>

src/wallet/rpcdump.cpp
laanwj: Meh, better too long than too short.

src/wallet/rpcdump.cpp
laanwj: I'm not terribly worried about race conditions here, just about overwriting existing files such as the wallet.

src/wallet/rpcdump.cpp
laanwj: Yes, that'd make sense.

src/wallet/rpcdump.cpp
promag: Yes, the comment should have paranoia label..

jnewbery:
Tested ACK 0cd9273fd959c6742574259d026039f7da0309a2

jnewbery:
Requires release notes (although this is fixing up undefined behaviour, there may be users who currently rely on that behaviour)

laanwj:
> Requires release notes (although this is fixing up undefined behaviour, there may be users who currently rely on that behaviour)

A release notes change is already part of the PR.

MarcoFalke:
@laanwj I also pushed this to the backports pull for 0.15.1

PR #7823, opened 2016-04-06
Add index to wallet UTXO

joaopaulofonseca:
This PR implements an index of unspent transaction outputs (UTXOs).

This approach optimizes `CWallet::AvailableCoins`, by doing some of its work each time a wallet transaction changes. For instance, using `CWallet::IsMine` earlier for each transaction output, it manages to keep track only of UTXOs related to wallet addresses.

That brings large improvements on RPCs such as `listunspent` and `fundrawtransaction`. Depending on the size of your wallet and number of UTXOs, performance gains may be up to 98%.

This way, instead of going through all the wallet transactions, the search for UTXOs is made only on this updated list of outputs.

Depends on #7822 to ensure functionality behaviour remains the same.

PR #8952, opened 2016-10-18
Add query options to listunspent RPC call

pedrobranco:
This PR adds a new optional JSON options for `listunspent` RPC call: 
- Return unspents greater or equal than a specific amount in BTC: `minimumAmount` (default = 0).
- Return unspents lower or equal than a specific amount in BTC: `maximumAmount` (default = `MAX_MONEY` = 21000000.0 BTC).
- Return unspents with a total number lower or equal than a specific number: `maximumCount` (default=0=unlimited).
- Return unspents which total is greater or equal than a specific amount in BTC: `minimumSumAmount` (default = `MAX_MONEY` = 21000000.0 BTC).

``` bash
> bitcoin-cli help listunspent
...
5. query options    (json, optional) JSON with query options
    {
      "minimumAmount"    (numeric or string, default=0) Minimum value of each UTXO in BTC
      "maximumAmount"    (numeric or string, default=21000000=unlimited) Maximum value of each UTXO in BTC
      "maximumCount"     (numeric or string, default=0=unlimited) Maximum number of UTXOs
      "minimumSumAmount" (numeric or string, default=21000000=unlimited) Minimum sum value all UTXOs in BTC
    }
...
```

When selecting unspents on client side in a wallet with a large set of small unspents makes the call `listunspent` slow. Using with `minimumAmount` option could improve the performance of the RPC call, performance which depends on the size of wallet, the distribution  of the unspents and the `minimumAmount` selected.

Example:

``` bash
> bitcoin-cli listunspent 0 9999999 '[]' true '{ "minimumAmount": 10, "maximumAmount": 80, "minimumSumAmount": 50, "maximumCount": 3 }'
```

**Note:** this PR also changes/simplifies the code in `AvailableCoins` to **make the longest calls later in execution time**, which also could improve performance in some `listunspent` results.

PR #9939, opened 2017-03-07
Cache vout IsMine() value on `CWallet::AvailableCoins()`

joaopaulofonseca:
This PR tries to improve the performance of the `listunspent`, `fundrawtransaction`, `sendtoaddress` and other RPC calls by improving `CWallet::AvailableCoins()` iteration through the wallet transactions.

E.g. If a given address is an outpoint of N transactions, I guess there's no point to check N times if the address `IsMine()`.

This PR supersedes #7823 although it's a different approach. In that PR, caching accurately those values was a little tricky because of all the scenarios that could make the cache invalidated (conflicts, double-spends, reorganizations, abandoned transactions, etc, ...).

Still, with this approach I got performance gains up to 90% on `CWallet::AvailableCoins()` for large wallets.

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -1952,6 +1952,8 @@ void CWallet::AvailableCoins(std::vector<COutput>& vCoins, bool fOnlySafe, const
 
     {
         LOCK2(cs_main, cs_wallet);
+
+        std::map<CScript, isminetype> mapOutputIsMine;
         for (std::map<uint256, CWalletTx>::const_iterator it = mapWallet.begin(); it != mapWallet.end(); ++it)
         {
             const uint256& wtxid = it->first;
@@ -2010,8 +2012,13 @@ void CWallet::AvailableCoins(std::vector<COutput>& vCoins, bool fOnlySafe, const
             }
 
             for (unsigned int i = 0; i < pcoin->tx->vout.size(); i++) {
-                isminetype mine = IsMine(pcoin->tx->vout[i]);
-                if (!(IsSpent(wtxid, i)) && mine != ISMINE_NO &&
+                auto inserted = mapOutputIsMine.emplace(pcoin->tx->vout[i].scriptPubKey, ISMINE_NO);
+                if (inserted.second) {
+                   inserted.first->second = IsMine(pcoin->tx->vout[i]);
+                }
+                isminetype mine = inserted.first->second;
+
+                if (mine != ISMINE_NO && !(IsSpent(wtxid, i)) &&
                     !IsLockedCoin((*it).first, i) && (pcoin->tx->vout[i].nValue > 0 || fIncludeZeroValue) &&
                     (!coinControl || !coinControl->HasSelected() || coinControl->fAllowOtherInputs || coinControl->IsSelected(COutPoint((*it).first, i))))
                         vCoins.push_back(COutput(pcoin, i, nDepth,


src/wallet/wallet.cpp
```diff
@@ -2080,8 +2082,15 @@ void CWallet::AvailableCoins(vector<COutput>& vCoins, bool fOnlyConfirmed, const
...
-                if (!(IsSpent(wtxid, i)) && mine != ISMINE_NO &&
+                isminetype mine = ISMINE_NO;
+                std::map<CScript, isminetype>::const_iterator im = mapOutputIsMine.find(pcoin->tx->vout[i].scriptPubKey);
+                if (im != mapOutputIsMine.end()) {
+                    mine = im->second;
+                } else {
+                    mine = mapOutputIsMine[pcoin->tx->vout[i].scriptPubKey] = IsMine(pcoin->tx->vout[i]);
+                }
+
+                if (mine != ISMINE_NO && !(IsSpent(wtxid, i)) &&
```
pedrobranco: I there's an ACK from the community, I would love to see this giant `if` break down with several early continues (to facilitate the reading), as applied [here](https://github.com/bitcoin/bitcoin/pull/8952/files#diff-b2bb174788c7409b671c46ccc86034bdR2066).

NicolasDorier:
Please check https://github.com/bitcoin/bitcoin/pull/9167, morcos is working on it

joaopaulofonseca:
@pedrobranco yes, that `if` is a little mess. I'll also take a look at it.

@NicolasDorier If I understood the @morcos PR correctly, it's related to the `isminetype` of the transaction inputs and not the outputs, which we need here to find our `utxos`.

NicolasDorier:
my bad I read too fast.

NicolasDorier:
Why not having the cache scoped to the WalletTransaction type ?
Instead of 
```
IsMine(pcoin->tx->vout[i])
````
we would have
```
pcoin->IsMine(i)
````
where WalletTransaction::IsMine would cache the result at the WalletTransaction level.

1. this would make it more readable, 
2. this cache could be populated everywhere in the code when IsMine is called,
3. the cache would span accross different calls to AvailableCoins, making AvailableCoins even faster.

File: src/wallet/wallet.cpp [hunk 2]
```diff
@@ -2080,8 +2082,15 @@ void CWallet::AvailableCoins(vector<COutput>& vCoins, bool fOnlyConfirmed, const
...
             for (unsigned int i = 0; i < pcoin->tx->vout.size(); i++) {
-                isminetype mine = IsMine(pcoin->tx->vout[i]);
-                if (!(IsSpent(wtxid, i)) && mine != ISMINE_NO &&
+                isminetype mine = ISMINE_NO;
+                std::map<CScript, isminetype>::const_iterator im = mapOutputIsMine.find(pcoin->tx->vout[i].scriptPubKey);
+                if (im != mapOutputIsMine.end()) {
+                    mine = im->second;
+                } else {
+                    mine = mapOutputIsMine[pcoin->tx->vout[i].scriptPubKey] = IsMine(pcoin->tx->vout[i]);
+                }
```
ryanofsky: Can simplify this as:

```
+                auto inserted = mapOutputIsMine.emplace(pcoin->tx->vout[i].scriptPubKey, ISMINE_NO);
+                if (inserted.second) {
+                   inserted.first->second = IsMine(pcoin->tx->vout[i]);
+                }
+                isminetype mine = inserted.first->second;
```

src/wallet/wallet.cpp
ryanofsky: This would seem like a good idea, especially if the continues were documented with new comments. Not really on-topic for this PR, however.

ryanofsky (commented):
utACK dca8d8926e655ba6639e748bcd32c560a5539b88

Code seems fine, but I don't know very much about the performance implications of this change, and would leave it for others to discuss. NicolasDorier's suggestion seems like a reasonable alternative, too.

joaopaulofonseca:
@NicolasDorier that's a good suggestion. Actually a while ago I was trying to create a cache to store these values, but at the wallet level. But couldn't manage to successfully achieve a solution with both reliability and speed. 

But I will look into your approach.

@ryanofsky Ok, I like it. I'll simplify that piece of code.

RHavar:
I've tested 46cbc0afa0060f0065dd48f8e7e20fe413f5b10c

Before median time of listunspent after a few runs on my wallet was 14.4 seconds, now it's 12.8s.

So it seems faster, but not earth shatteringly so. For reference my wallet has about ~520k transactions in it, where of them ~168k are mine (outgoing transactions).

laanwj:
@RHavar thanks for testing!

joaopaulofonseca:
@RHavar thanks a lot for your feedback! The current implementation of `listunspent` is a little tricky to optimize without a larger refactor. 

I've tried to optimize each of the two heavier calls within `listunspent`: `IsMine()` and `IsSpent()`. But without optimizing both, they will only fix different use cases separately, which will behave very differently for **large wallet** vs **large amount of unspents**.

I haven't had much time to work on this lately, but is a problem I'd like to come up with a clean solution in the near future, since it's an issue affecting almost everyone with large wallets.

ryanofsky (commented):
utACK 46cbc0afa0060f0065dd48f8e7e20fe413f5b10c. Code is slightly simplified since last review, and change looks good assuming performance improvements warrant it.

NicolasDorier:
utACK  46cbc0a, it is simple enough. Can keep wallet level caching for later PR.

ryanofsky:
Change is simple and has multiple acks, but needs to be rebased.

MarcoFalke:
Not rebased in 30 days.

This is now up for grabs.

joaopaulofonseca:
Sorry guys, haven't had much time to work on this lately.. Although this improves the speed, the last tests I did showed there are still some edge cases where the unspent count differs which means that the cache is not accurate.. I'll try to pick this up in the future, and reopen if needed.

ryanofsky:
Could add [up for grabs](https://github.com/bitcoin/bitcoin/issues?q=label%3A%22Up+for+grabs%22) label

PR #9943, opened 2017-03-07
Make script.py wildcard importable.

jnewbery:
This PR is part of a larger effort to make the test_framework API more intuitive. See https://github.com/bitcoin/bitcoin/pull/9876 for background.

Several testcases use a wildcard import from test_framework.script. That's not ideal as it pollutes the namespace with all of the names defined in test_framework.script, including anything that script.py has imported itself.

This PR adds an `__all__` module variable to script.py which indexes all of the names that should commonly be required by individual test cases. That makes the wildcard import a well-defined API and test cases can wildcard import without worrying about importing unexpected names. If a test case needs to import something from script.py which isn't included in `__all__`, it can explicitly import it.

I've not included `hash160()` or `CScriptOp()` in `__all__` because `hash160()` is a helper function which I think should be defined elsewhere (ideally in the same module as `sha256()`, `ripemd160()` and `hash256()`), and `CScriptOp` should not commonly be required by testcases (the only place it's currently used is in `GetP2PKHScript()` in p2p-segwit.py, which arguably should be pulled up into the script.py module).

### DIFF
--- a/test/functional/p2p-fullblocktest.py
+++ b/test/functional/p2p-fullblocktest.py
@@ -18,6 +18,7 @@ from test_framework.blocktools import *
 import time
 from test_framework.key import CECKey
 from test_framework.script import *
+from test_framework.script import hash160
 import struct
 
 class PreviousSpendableOutput(object):
--- a/test/functional/p2p-segwit.py
+++ b/test/functional/p2p-segwit.py
@@ -8,6 +8,7 @@ from test_framework.mininode import *
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import *
 from test_framework.script import *
+from test_framework.script import hash160, CScriptOp
 from test_framework.blocktools import create_block, create_coinbase, add_witness_commitment, get_witness_script, WITNESS_COMMITMENT_HEADER
 from test_framework.key import CECKey, CPubKey
 import time
--- a/test/functional/segwit.py
+++ b/test/functional/segwit.py
@@ -8,7 +8,8 @@ from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import *
 from test_framework.mininode import sha256, CTransaction, CTxIn, COutPoint, CTxOut, COIN, ToHex, FromHex
 from test_framework.address import script_to_p2sh, key_to_p2pkh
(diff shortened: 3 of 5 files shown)


fanquake:
Needs a rebase.

test/functional/p2p-fullblocktest.py
```diff
@@ -18,6 +18,7 @@
 import time
 from test_framework.key import CECKey
 from test_framework.script import *
+from test_framework.script import hash160
```
jimmysong: It seems a little strange that hash160 is in script.py and hash256/sha256 are in mininode.py can we put all three in one place (probably mininode.py)? You're changing most of the places where this is being imported (smartfees.py and test_framework/address.py seem to be the only other ones). It might be worth just changing this, too.

jimmysong (commented):
It looks like smartfees.py, bip9-softforks.py and bip65-cltv-p2p.py could also use the same import in `test/functional`. Also, address.py and blocktools.py could use the same import cleanup in `test/functional/test_framework`

test/functional/p2p-fullblocktest.py
jnewbery: Yes, I agree that they should all live in the same place. My preference would be for utils.py to actually be pure utility functions and for the hash functions to live in there.

I haven't moved it in this commit because that seemed the minimal change. We should move hash160 to live alongside the other functions in a future PR that cleans up utils.py and mininode.py.

jnewbery:
I didn't change the imports in smartfees.py, bip9-softforks.py and bip65-cltv-p2p.py because there's no harm in leaving them as explicit. I only moved segwit.py from exlicit to wildcard because it was importing an absurdly long list of names.

for imports within the test_framework module itself, I think we should aim for them all to be explicit. I have updated address.py. It was previously importing hash256 and sha256 indirectly via script.py. I've now changed it so that it imports them directly from mininode.py.

jimmysong:
ACK d5cc6ff31755877eb899bc45497a240e2432615f

jimmysong (approved):

jnewbery:
I'd love to tidy up imports in the test framework, but there doesn't seem to be general enthusiasm for this. Closing for now.

PR #9032, opened 2016-10-27
test: Add format-dependent comparison to bctest

laanwj:
This splits the output comparison for `bitcoin-tx` into two steps:
- First, check for data mismatch, parsing the data as json or hex depending on the extension of the output file
- Then, check if the literal string matches

For either of these cases give a different error.

This prevents wild goose chases when e.g. a trailing space doesn't match exactly (by helping narrow down the problem), and makes sure that both test output and examples are valid data of the purported format.

PR #9945, opened 2017-03-07
Improve logging in bctest.py if there is a formatting mismatch

jnewbery:
#9032 added format comparison checking to bctest.py.

If the formats don't match, the test doesn't print the difference between the expected output and the actual output, which makes troubleshooting and fixing the test difficult.

This small changes the logic so that if the formatting comparison fails, we go on to compare the data, and then fail the test if either of those comparisons fail.

### DIFF
--- a/src/test/bctest.py
+++ b/src/test/bctest.py
@@ -10,6 +10,7 @@ import sys
 import binascii
 import difflib
 import logging
+import pprint
 
 def parse_output(a, fmt):
     """Parse the output according to specified format.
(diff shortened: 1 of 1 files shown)


laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/9945/commits/6c1fb73dd1744b7281e5d506c14307daf1b72e1f

MarcoFalke:
utACK 6c1fb73dd1744b7281e5d506c14307daf1b72e1f

Issue #9933, opened 2017-03-07
Compile error: "clientversion.cpp:42:10: fatal error: 'build.h' file not found"


pinheadmz:
OSX 10.10.5
git clone latest master (19be26afe3d04783a92d032b55bf3fb1e2ae63cc) 

I'm not sure what happened because I've been able to compile bitcoin from source on this machine recently until today. 

[config.log](http://pastebin.com/DgY6PbKF)

I've seen on other issues that openssl might be the issue:

```
$ brew info openssl
openssl: stable 1.0.2k (bottled) [keg-only]
SSL/TLS cryptography library
https://openssl.org/
/usr/local/Cellar/openssl/1.0.2j (1,695 files, 12M)
  Poured from bottle on 2016-10-04 at 10:12:41
/usr/local/Cellar/openssl/1.0.2k (1,696 files, 12M)
  Poured from bottle on 2017-03-06 at 15:09:14
From: https://github.com/Homebrew/homebrew-core/blob/master/Formula/openssl.rb
==> Dependencies
Build: makedepend ✘
==> Options
--universal
	Build a universal binary
--without-test
	Skip build-time tests (not recommended)
==> Caveats
A CA file has been bootstrapped using certificates from the SystemRoots
keychain. To add additional certificates (e.g. the certificates added in
the System keychain), place .pem files in
  /usr/local/etc/openssl/certs

and run
  /usr/local/opt/openssl/bin/c_rehash

This formula is keg-only, which means it was not symlinked into /usr/local.

Apple has deprecated use of OpenSSL in favor of its own TLS and crypto libraries

If you need to have this software first in your PATH run:
  echo 'export PATH="/usr/local/opt/openssl/bin:$PATH"' >> ~/.bash_profile

For compilers to find this software you may need to set:
    LDFLAGS:  -L/usr/local/opt/openssl/lib
    CPPFLAGS: -I/usr/local/opt/openssl/include
For pkg-config to find this software you may need to set:
    PKG_CONFIG_PATH: /usr/local/opt/openssl/lib/pkgconfig

```

... but I did set those flags (output of configure):

```
Options used to compile and link:
  with wallet   = yes
  with gui / qt = no
  with zmq      = no
  with test     = yes
  with bench    = yes
  with upnp     = yes
  debug enabled = no
  werror        = no

  target os     = darwin
  build os      = darwin

  CC            = gcc
  CFLAGS        = -g -O2
[...]

PR #5872, opened 2015-03-09
configure: BITCOIN_SUBDIR_TO_INCLUDE: Improve compatibility with paths including space and multiline cpp output

luke-jr:
This fixes a number of bugs in the BITCOIN_SUBDIR_TO_INCLUDE macro:
- tr was deleting all 'n' and 'r' characters in addition to the backslash, due to too much escaping.
- Escaped spaces in paths were ignored, parsing the space as a delimiter
- Multiline output from cpp (used when there are a number of includes being listed) weren't handled correctly.

The new sed takes two steps:
1. Translate escaped newlines into a space, to work with single lines.
2. Search for the file we want and return the complete path for it.

PR #9946, opened 2017-03-07
Fix build errors if spaces in path or parent directory

pinheadmz:
See Issue https://github.com/bitcoin/bitcoin/issues/9933

Some work started in PR https://github.com/bitcoin/bitcoin/pull/5872, but I can't understand what's going on in there...

It fixed the problem I was having (compiling in OSX 10.10.5) if there were spaces anywhere in the absolute path of the source code. Successfully compiled and passed all rpc-tests in paths with and without spaces. 

[GNU Make with spaces in filenames seems to be a general issue.](https://stackoverflow.com/questions/9838384/can-gnu-make-handle-filenames-with-spaces)

I will continue to bang on this and see if I break anything else.

### DIFF
--- a/test/functional/test_runner.py
+++ b/test/functional/test_runner.py
@@ -87,7 +87,7 @@ BASE_SCRIPTS= [
 ZMQ_SCRIPTS = [
     # ZMQ test can only be run if bitcoin was built with zmq-enabled.
     # call test_runner.py with -nozmq to explicitly exclude these tests.
-    "zmq_test.py"]
+    'zmq_test.py']
 
 EXTENDED_SCRIPTS = [
     # These tests are not run by the travis build process.
@@ -203,9 +203,9 @@ def main():
         sys.exit(0)
 
     if args.help:
-        # Print help for test_runner.py, then print help of the first script and exit.
+        # Print help for test_runner.py, then print help of the first script (with args removed) and exit.
         parser.print_help()
-        subprocess.check_call((config["environment"]["SRCDIR"] + '/test/functional/' + test_list[0]).split() + ['-h'])
+        subprocess.check_call([(config["environment"]["SRCDIR"] + '/test/functional/' + test_list[0].split()[0])] + ['-h'])
         sys.exit(0)
 
     run_tests(test_list, config["environment"]["SRCDIR"], config["environment"]["BUILDDIR"], config["environment"]["EXEEXT"], args.jobs, args.coverage, passon_args)
@@ -296,9 +296,10 @@ class TestHandler:
             port_seed = ["--portseed={}".format(len(self.test_list) + self.portseed_offset)]
             log_stdout = tempfile.SpooledTemporaryFile(max_size=2**16)
             log_stderr = tempfile.SpooledTemporaryFile(max_size=2**16)
+            test_argv = t.split()
             self.jobs.append((t,
                               time.time(),
-                              subprocess.Popen((self.tests_dir + t).split() + self.flags + port_seed,
+                              subprocess.Popen([self.tests_dir + test_argv[0]] + test_argv[1:] + self.flags + port_seed,
                                                universal_newlines=True,
                                                stdout=log_stdout,
                                                stderr=log_stderr),
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -167,8 +167,8 @@ BITCOIN_CORE_H = \
 
 obj/build.h: FORCE
 	@$(MKDIR_P) $(builddir)/obj
-	@$(top_srcdir)/share/genbuild.sh $(abs_top_builddir)/src/obj/build.h \
-	  $(abs_top_srcdir)
+	@$(top_srcdir)/share/genbuild.sh "$(abs_top_builddir)/src/obj/build.h" \
+	  "$(abs_top_srcdir)"
 libbitcoin_util_a-clientversion.$(OBJEXT): obj/build.h
 
 # server: shared between bitcoind and bitcoin-qt


laanwj:
```
FileNotFoundError: [Errno 2] No such file or directory: '/home/travis/build/bitcoin/bitcoin/build/bitcoin-i686-pc-linux-gnu/qa/rpc-tests/txn_doublespend.py --mineblock'
```
It's trying to execute the script + parameter as one command.

pinheadmz:
Ah yeah some of the tests in the list include --arguments in a string with the command. I wonder why it only failed on Travis builds with python3-zmq? I added two lines that separate the --arguments into list elements so hopefully `subprocess.Popen()` likes that better on those builds.

theuni:
Rather than adding string parsing into the mix here, how about making the tests a pair of ['filename', 'args'] instead?

pinheadmz:
I had to change a few other lines to work with the new format  ['filename', 'args'] but it's working for me

MarcoFalke:

pinheadmz:
@MarcoFalke I will, I just want to make sure I was taking @theuni direction correctly on this last modification

fanquake:
ping @pinheadmz . This needs a rebase & commits squashed.

pinheadmz:
@fanquake happy to do that now but I was hoping to get feedback from @theuni on latest commit: reformat all the test names to lists? Or parse as strings? (Previous commit)

MarcoFalke:
I think it is fine to have each functional test in the format ['file_name.py', 'arg'] ...

pinheadmz:
wow I really fucked up this rebase, sorry, trying again

pinheadmz:
ok there we go, going to test building in my own directory with spaces to double-check

theuni:
@pinheadmz yes, that's what I had in mind, thanks.

test/functional/test_runner.py
```diff
@@ -205,7 +206,7 @@ def main():
     if args.help:
         # Print help for test_runner.py, then print help of the first script and exit.
         parser.print_help()
-        subprocess.check_call((config["environment"]["SRCDIR"] + '/test/functional/' + test_list[0]).split() + ['-h'])
+        subprocess.check_call([config["environment"]["SRCDIR"] + '/qa/rpc-tests/' + test_list[0][0]] + ['-h'])
```
theuni: rebase issue

test/functional/test_runner.py
```diff
@@ -248,6 +249,7 @@ def run_tests(test_list, src_dir, build_dir, exeext, jobs=1, enable_coverage=Fal
     results = BOLD[1] + "%s | %s | %s\n\n" % ("TEST".ljust(max_len_name), "PASSED", "DURATION") + BOLD[0]
     for _ in range(len(test_list)):
         (name, stdout, stderr, passed, duration) = job_queue.get_next()
+        name = " ".join(name)
```
theuni: What changed here?

pinheadmz: At this point, `name` is a list not a string. So this turns ['testname', '--arg'] back into 'testname --arg' before it is displayed

test/functional/test_runner.py
pinheadmz: fixed. sorry/thanks. I guess there's no test to fail for requesting help!

test/functional/test_runner.py
theuni: Ah, of course.

Mind keeping them separate so that's more obvious? e.g. rename to test and print test[0] + test[1:] ?

jnewbery:
Sorry I'm late to this party. @theuni asked me to take a look at this.

I really don't like the approach of changing the list of tests into a list of lists. It's torturing the test_runner code for the sake of a couple of test cases which take arguments. It's also borked the formatting of the printout report (look at how `max_len_name` gets set to see how).

I think it's much better (and less intrusive) to keep the test list as a list of strings, and then just change the Popen() call to be something like:

```python
test_argv = t.split()
self.jobs.append((t,
	      time.time(),
	      subprocess.Popen([self.tests_dir + test_argv[0]] + test_argv[1:] + self.flags + port_seed,
			       universal_newlines=True,
			       stdout=log_stdout,
			       stderr=log_stderr),
	      log_stdout,
	      log_stderr))
```

This is very similar to what Popen() was calling before, it's just that split() is only being called on the test name string, rather than the complete path. I think that fixes your original problem of test cases not working when there's a space in the path.

Can you test that and see if it works for you?

pinheadmz:
Let me check my backup drive... ;-) I got all these messages to squash! Should've trusted my gut. Wouldn't be hard to re-write anyway and @jnewbery's suggested code is cleaner anyway.

pinheadmz:
Found a backup! https://github.com/bitcoin/bitcoin/pull/9946/commits/33894380c0b0a2ed2f46997c685e75125badd205 is how it was before, with test names as strings and using string parsing instead of lists.

jnewbery:
Looks like https://github.com/bitcoin/bitcoin/pull/9946/commits/33894380c0b0a2ed2f46997c685e75125badd205 is the current tip. Do you still have the old commit? Can you push to a new branch on github?

pinheadmz:
sorry, https://github.com/bitcoin/bitcoin/pull/9946/commits/16008c4d8dfd1539cefd98feb45d83e6389efd27

jnewbery:
I like https://github.com/bitcoin/bitcoin/pull/9946/commits/16008c4d8dfd1539cefd98feb45d83e6389efd27 much better. My comment would be that rather than what you're doing:
-  concatenate the base path with the test string
- then split on " --"

instead you should:

- split the test string on whitespace (this is the default for string.split())
- concatenate the base path with the first entry in the list
- add that concatenated value with the rest of the list

That seems reasonable. You're splitting the test string into arguments based on whitespace (which is what a shell does anyway!)

pinheadmz:
This seems to work fine on my computer, installing and testing into directory "bitcoin dev". Although I do get a lot of timeout errors from some of the tests... any advice on that?

theuni:
@pinheadmz Thanks for sticking with this, and sorry for the back and forth. @jnewbery has been doing tons of work on these tests lately, so I'll defer to his preference.

jnewbery:
Glad it works. I'll happily review this once you've pushed the new commit.

Regarding the timeouts: the first thing I'd check is whether there are any bitcoin processes running. Sometimes the framework fails to clean up old processes if the test fails, which can lead to issues next time you run a test. If that's not the issue, I've beefed up the test logs recently. You should be able to run `combine_logs.py <test directory> -c | less -r` to get a combined view of what happened in the test.

And if that doesn't help, please open a new issue with the logs and @ me and I'll take a look.

pinheadmz:
squashed! 1de311d64d477fa2a220aa77840f86f4deabebb6

jnewbery:
utACK https://github.com/bitcoin/bitcoin/commit/1de311d64d477fa2a220aa77840f86f4deabebb6 modulo my single nit.

theuni:
utACK b1f584dbc185fc9961d4d8d16680ca4041f3d1f3

PR #9947, opened 2017-03-08
Add true/false return value to prioritisetransaction

jonasschnelli:
`prioritisetransaction` currently always return true, even if the transactions is not in the mempool.
This PR fixes that by returning false in case of a failed prioritization.

### DIFF
--- a/src/rpc/mining.cpp
+++ b/src/rpc/mining.cpp
@@ -268,7 +268,7 @@ UniValue prioritisetransaction(const JSONRPCRequest& request)
             "                  The fee is not actually paid, only the algorithm for selecting transactions into a block\n"
             "                  considers the transaction as it would have paid a higher (or lower) fee.\n"
             "\nResult:\n"
-            "true              (boolean) Returns true\n"
+            "true              (boolean) Returns true if prioritization was possible, otherwise false\n"
             "\nExamples:\n"
             + HelpExampleCli("prioritisetransaction", "\"txid\" 10000")
             + HelpExampleRpc("prioritisetransaction", "\"txid\", 10000")
@@ -279,7 +279,7 @@ UniValue prioritisetransaction(const JSONRPCRequest& request)
     uint256 hash = ParseHashStr(request.params[0].get_str(), "txid");
     CAmount nAmount = request.params[1].get_int64();
 
-    mempool.PrioritiseTransaction(hash, nAmount);
+    return mempool.PrioritiseTransaction(hash, nAmount);
     return true;
 }
 
--- a/src/txmempool.cpp
+++ b/src/txmempool.cpp
@@ -895,7 +895,7 @@ CTxMemPool::ReadFeeEstimates(CAutoFile& filein)
     return true;
 }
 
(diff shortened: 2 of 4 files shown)


src/txmempool.cpp
```diff
@@ -895,7 +895,7 @@ CTxMemPool::ReadFeeEstimates(CAutoFile& filein)
     return true;
 }
 
-void CTxMemPool::PrioritiseTransaction(const uint256& hash, const CAmount& nFeeDelta)
+bool CTxMemPool::PrioritiseTransaction(const uint256& hash, const CAmount& nFeeDelta)
 {
     {
         LOCK(cs);
```
bitkevin: we should check the tx hash to find out if it's exist in `mapTx` before add `nFeeDelta` to `mapDeltas`.

jonasschnelli: Good point.. I'll fix that.

jonasschnelli: Oh wait. This seems to be intentional. This allows to increase fees (add a fee delta) before the transaction hits the mempool.

dcousens:
As discussed,  the return value would probably better approximate 
whether the transaction was in the memory pool",  as prioritization should always occur.

src/rpc/mining.cpp
```diff
@@ -279,7 +279,7 @@ UniValue prioritisetransaction(const JSONRPCRequest& request)
     uint256 hash = ParseHashStr(request.params[0].get_str(), "txid");
     CAmount nAmount = request.params[1].get_int64();
 
-    mempool.PrioritiseTransaction(hash, nAmount);
+    return mempool.PrioritiseTransaction(hash, nAmount);
     return true;
```
laanwj: the `return true` can go here

jnewbery:
I don't understand why this is required. As mentioned above, prioritisetransaction can be used for transactions which are not yet in the mempool, and then the fee delta is:

- used to determine whether the transaction should be allowed into the mempool (the `ApplyDeltas()` call in `AcceptToMemoryPoolWorker()`)
- applied to the transaction as it enters the mempool (`mapTx.modify(newit, update_fee_delta(deltas.second));` in `CTxMemPool::addUnchecked()`)

If you really need to check whether the transaction is already in the mempool you can use getrawmempool.

MarcoFalke:
Agree with jnewbery. Imo it is confusing for prioritisetransaction to
return a bool. (One could think that false means the call failed, which is
not true.)

There are other methods to check if a transaction is in the mempool.

src/txmempool.cpp
gmaxwell: And an important feature: without the ability to add deltas before a transaction exists, you may not be able to prioritize a transaction at all (because of rejection).

ericshawlinux:
You could return a struct that contains a bool, that way the said bool value is given a name, making the intention clear.

jonasschnelli:
I think @jnewbery and @MarcoFalke are right. I misunderstood it's behaviour. It can prioritise future transaction and I wasn't aware of that fact.

An option would be to return true/false wether the transaction was in the mempool or not...

jonasschnelli:
> There are other methods to check if a transaction is in the mempool.

@MarcoFalke is right here.
Closing...

PR #9963, opened 2017-03-09
util: Properly handle errors during log message formatting

laanwj:
Instead of having an exception propagate into the program (which at worst causes a crash) when an error happens while formatting a log message, just print a message to the log. This message clearly indicates what log message was formatted wrongly, and what error happened during formatting it.

Addresses #9423.

#### Before
```
2017-03-09 10:58:20 

************************
EXCEPTION: St13runtime_error       
tinyformat: Too many conversion specifiers in format string       
bitcoin in AppInit()       



************************
EXCEPTION: St13runtime_error       
tinyformat: Too many conversion specifiers in format string       
bitcoin in AppInit()       

2017-03-09 10:58:20 Shutdown: In progress...
2017-03-09 10:58:20 scheduler thread interrupt
2017-03-09 10:58:20 Shutdown: done
```
(and process exits)

#### After
```
2017-03-09 10:51:50 Error "tinyformat: Too many conversion specifiers in format string" while formatting log message: Erasing %s %s
```
(and process continues)

### DIFF
--- a/src/tinyformat.h
+++ b/src/tinyformat.h
@@ -123,7 +123,7 @@ namespace tinyformat {}
 namespace tfm = tinyformat;
 
 // Error handling; calls assert() by default.
-#define TINYFORMAT_ERROR(reasonString) throw std::runtime_error(reasonString)
+#define TINYFORMAT_ERROR(reasonString) throw tinyformat::format_error(reasonString)
 
 // Define for C++11 variadic templates which make the code shorter & more
 // general.  If you don't define this, C++11 support is autodetected below.
@@ -163,6 +163,13 @@ namespace tfm = tinyformat;
 #endif
 
 namespace tinyformat {
+
+class format_error: public std::runtime_error
+{
+public:
+    format_error(const std::string &what): std::runtime_error(what) {
+    }
+};
 
 //------------------------------------------------------------------------------
 namespace detail {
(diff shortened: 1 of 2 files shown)


TheBlueMatt:
Looks good, though I'd prefer if we narrow the catch and change TINYFORMAT_ERROR to throw a more narrow exception type. Not for any specific concern, but...man I hate C++ exceptions.

laanwj:
So are there any errors during log message formatting that raise `runtime_error` that you'd want to escalate? We don't have any sensible/specific exception handling so personally I'd prefer to keep it like this.

laanwj:
Ehh oops I guess it makes sense to have the catch() only around `tfm::format`, though, not LogPrintStr (which does I/O).

laanwj:
Okay, updated so that only the formatting is guarded and there is a single LogPrintStr call which can still fail for legit reasons such as disk full...

TheBlueMatt:
Heh, I knew it needed a tighter catch somehow :P. Anyway, if you dont want to have something more specific that's fine.

utACK 3b092bd9b6b3953d5c3052d57e4827dbd85941fd

laanwj:
LIke this? I don't think this is not a very interesting place to get started defining more specific exceptions, but we have to start somewhere...

TheBlueMatt:
Yea, agreed. I do like that better, though I suppose we really need to come up with some kind of coherent exception-use design at some point.

PR #9974, opened 2017-03-10
Add basic Qt wallet test

ryanofsky:
This adds a basic unit test for the Qt send coins dialog. I have more changes (for a future PR) that extend this to cover new Qt bumpfee code in #9697.

Previously, I brought up the topic of qt unit tests in https://github.com/bitcoin/bitcoin/pull/9817#issuecomment-281351497 and https://github.com/bitcoin/bitcoin/pull/9697#discussion_r102971669.

### DIFF
--- a/src/qt/test/test_main.cpp
+++ b/src/qt/test/test_main.cpp
@@ -7,7 +7,6 @@
 #endif
 
 #include "chainparams.h"
-#include "key.h"
 #include "rpcnestedtests.h"
 #include "util.h"
 #include "uritests.h"
@@ -36,7 +35,6 @@ extern void noui_connect();
 // This is all you need to run all the tests
 int main(int argc, char *argv[])
 {
-    ECC_Start();
     SetupEnvironment();
     SetupNetworking();
     SelectParams(CBaseChainParams::MAIN);
@@ -66,6 +64,5 @@ int main(int argc, char *argv[])
     if (QTest::qExec(&test4) != 0)
         fInvalid = true;
 
-    ECC_Stop();
     return fInvalid;
 }
--- a/src/qt/test/test_main.cpp
+++ b/src/qt/test/test_main.cpp
@@ -23,12 +23,22 @@
 
 #include <openssl/ssl.h>
 
-#if defined(QT_STATICPLUGIN) && QT_VERSION < 0x050000
+#if defined(QT_STATICPLUGIN)
 #include <QtPlugin>
+#if QT_VERSION < 0x050000
 Q_IMPORT_PLUGIN(qcncodecs)
 Q_IMPORT_PLUGIN(qjpcodecs)
 Q_IMPORT_PLUGIN(qtwcodecs)
 Q_IMPORT_PLUGIN(qkrcodecs)
+#else
+#if defined(QT_QPA_PLATFORM_XCB)
+Q_IMPORT_PLUGIN(QXcbIntegrationPlugin);
+#elif defined(QT_QPA_PLATFORM_WINDOWS)
+Q_IMPORT_PLUGIN(QWindowsIntegrationPlugin);
+#elif defined(QT_QPA_PLATFORM_COCOA)
+Q_IMPORT_PLUGIN(QCocoaIntegrationPlugin);
+#endif
+#endif
 #endif
 
 extern void noui_connect();
--- a/src/qt/test/test_main.cpp
+++ b/src/qt/test/test_main.cpp
@@ -14,9 +14,10 @@
 
 #ifdef ENABLE_WALLET
 #include "paymentservertests.h"
+#include "wallettests.h"
 #endif
 
-#include <QCoreApplication>
+#include <QApplication>
 #include <QObject>
 #include <QTest>
 
@@ -43,8 +44,8 @@ int main(int argc, char *argv[])
     bool fInvalid = false;
 
     // Don't remove this, it's needed to access
-    // QCoreApplication:: in the tests
-    QCoreApplication app(argc, argv);
+    // QApplication:: and QCoreApplication:: in the tests
+    QApplication app(argc, argv);
     app.setApplicationName("Bitcoin-Qt-test");
 
     SSL_library_init();
@@ -67,6 +68,12 @@ int main(int argc, char *argv[])
     if (QTest::qExec(&test4) != 0) {
         fInvalid = true;
     }
+#ifdef ENABLE_WALLET
+    WalletTests test5;
+    if (QTest::qExec(&test5) != 0) {
+        fInvalid = true;
+    }
+#endif
 
     return fInvalid;
 }
--- a/src/qt/test/test_main.cpp
+++ b/src/qt/test/test_main.cpp
@@ -50,19 +50,23 @@ int main(int argc, char *argv[])
     SSL_library_init();
 
     URITests test1;
-    if (QTest::qExec(&test1) != 0)
+    if (QTest::qExec(&test1) != 0) {
         fInvalid = true;
+    }
 #ifdef ENABLE_WALLET
     PaymentServerTests test2;
-    if (QTest::qExec(&test2) != 0)
+    if (QTest::qExec(&test2) != 0) {
         fInvalid = true;
+    }
 #endif
     RPCNestedTests test3;
-    if (QTest::qExec(&test3) != 0)
+    if (QTest::qExec(&test3) != 0) {
         fInvalid = true;
+    }
     CompatTests test4;
-    if (QTest::qExec(&test4) != 0)
+    if (QTest::qExec(&test4) != 0) {
         fInvalid = true;
+    }
 
     return fInvalid;
 }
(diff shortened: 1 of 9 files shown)


luke-jr:
I almost wonder if it might be useful to have a RPC wrapper for the GUI (possibly in addition to this) so we can just use the RPC tests with it as well.

fanquake:
Failed with:
```
==============================================
   Bitcoin Core 0.14.99: src/test-suite.log
==============================================
# TOTAL: 2
# PASS:  1
# SKIP:  0
# XFAIL: 0
# FAIL:  1
# XPASS: 0
# ERROR: 0
.. contents:: :depth: 2
FAIL: qt/test/test_bitcoin-qt
=============================
This application failed to start because it could not find or load the Qt platform plugin "xcb"
in "".
Reinstalling the application may fix this problem.
```

laanwj:
Concept ACK

MarcoFalke:
Concept ACK for sure, as we basically don't have any gui tests right now. In the long term I'd prefer a solution that runs on a higher abstraction level. Your current approach is pretty much mimicking functional tests on a low (unit test) level.

ryanofsky:
> I'd prefer a solution that runs on a higher abstraction level. Your current approach is pretty much mimicking functional tests on a low (unit test) level.

I don't think I understand. I'm not doing anything original here, just following the approach of http://doc.qt.io/qt-5/qttestlib-tutorial3-example.html.

Is it that instead of the writing a test that sends qt events, you would prefer to see a test that starts an x server and sends keyboard and mouse events?

jonasschnelli:
I really like this.
Concept ACK.

MarcoFalke:
> Is it that instead of the writing a test that sends qt events, you would
prefer to see a test that starts an x server and sends keyboard and mouse
events?

Indeed. I was dreaming of a gui test framework where you just press a
"record this" button. Then records discrete actions (e.g. "Press: Receive
tab", "Press: Copy address", other gui: "Press: Send tab", "Insert:
$address", ...) instead of the exact mouse and keyboard movements, such
that it works on different screen resolutions. Though, we also need a way
to wait for the gui to catch up on longer tasks such as importmulti with
rescan, so I don't think this is a weekend project.

src/qt/test/test_main.cpp
```diff
@@ -46,7 +45,7 @@ int main(int argc, char *argv[])
 
     // Don't remove this, it's needed to access
     // QCoreApplication:: in the tests
```
fanquake: Comment needs removing/updating?

fanquake:
Still failing Travis with the same issue, @theuni should probably take a look here.

src/qt/test/test_main.cpp
```diff
@@ -65,7 +64,11 @@ int main(int argc, char *argv[])
     CompatTests test4;
     if (QTest::qExec(&test4) != 0)
         fInvalid = true;
+#ifdef ENABLE_WALLET
+    WalletTests test5;
+    if (QTest::qExec(&test5) != 0)
```
MarcoFalke: style nit: We use braces now; Always.

src/qt/test/test_main.cpp
ryanofsky: It's still accurate, but I expanded the comment in f4d28fbe1b8041ce358d80ecb0e1fa074eee359c.

src/qt/test/test_main.cpp
ryanofsky: Added in b61b34c89dbe3946e21147dcfe620a4a6ff536ed

ryanofsky (commented):
Finally got test to pass on travis.

jonasschnelli:
A lock is missing.
```
diff --git a/src/qt/test/wallettests.cpp b/src/qt/test/wallettests.cpp
index 81b358e..d63f4ec 100644
--- a/src/qt/test/wallettests.cpp
+++ b/src/qt/test/wallettests.cpp
@@ -73,6 +73,7 @@ void WalletTests::walletTests()
     test.CreateAndProcessBlock({}, GetScriptForRawPubKey(test.coinbaseKey.GetPubKey()));
     bitdb.MakeMock();
     CWallet wallet("wallet_test.dat");
+    LOCK(wallet.cs_wallet);
     bool firstRun;
     wallet.LoadWallet(firstRun);
     wallet.SetAddressBook(test.coinbaseKey.GetPubKey().GetID(), "", "receive");
```
How could this pass travis with the missing lock?

Will properly ACK after apply the diff above.

Question: is it possible to create a screenshot (or a bitmap/JPG/PNG) of the UI elements you open?
A such form of an UI export would be great. We could add this then to a CI and have plenty of screenshots from various operating systems in order to verify the UI correctness.

ryanofsky:
> A lock is missing.

Will add. I'm not sure the way the lock is acquired in the diff is exactly right, because cs_main is supported to be acquired before cs_wallet, and some wallet methods acquire both.

> How could this pass travis with the missing lock?

Locally, the test works fine for me with the missing lock. On travis, every configuration we have either specifies NO_QT or NO_WALLET, or omits RUN_TESTS, so neither PaymentServerTests code nor this new WalletTests code currently executes.

> Question: is it possible to create a screenshot (or a bitmap/JPG/PNG) of the UI elements you open?
A such form of an UI export would be great. We could add this then to a CI and have plenty of screenshots from various operating systems in order to verify the UI correctness.

Something like this should be possible. To help debug this change, I frequently added `widget.show(); QEventLoop().exec();` calls to be able to see and interact with individual test widgets, so it should be possible to capture images also (maybe with [QWidget::grab](http://doc.qt.io/qt-5/qwidget.html#grab)?)

ryanofsky:
Fixed locking bug in 739f19ca969c2ae351e18d2bbb424237830f54f1

Squashed 739f19ca969c2ae351e18d2bbb424237830f54f1 -> 9576b015a107b98fc950c574ed01d993b388d7c9 ([pr/qt-test.7](https://github.com/ryanofsky/bitcoin/commits/pr/qt-test.7) -> [pr/qt-test.8](https://github.com/ryanofsky/bitcoin/commits/pr/qt-test.8))

MarcoFalke:
utACK 9576b015a107b98fc950c574ed01d993b388d7c9

jonasschnelli:
Tested ACK 9576b015a107b98fc950c574ed01d993b388d7c9

jnewbery:
This is great. Hopefully I'll have coverage running soon so we can see how much extra code you've managed to exercise.

Post-merge tested ACK https://github.com/bitcoin/bitcoin/commit/9576b015a107b98fc950c574ed01d993b388d7c9.

paveljanik:
BTW, compiling this is not possible with gcc-4.8.5, the default gcc on openSUSE 42.2:

```
/usr/include/boost/signals2/connection.hpp: In function 'uint256 {anonymous}::SendCoins(CWallet&, SendCoinsDialog&, const CBitcoinAddress&, CAmount)':
/usr/include/boost/signals2/connection.hpp:234:7: error: 'boost::signals2::scoped_connection::scoped_connection(const boost::signals2::scoped_connection&)' is private
       scoped_connection(const scoped_connection &other);
       ^
qt/test/wallettests.cpp:47:6: error: within this context
     });
      ^
Makefile:7991: recipe for target 'qt/test/qt_test_test_bitcoin_qt-wallettests.o' failed
make[2]: *** [qt/test/qt_test_test_bitcoin_qt-wallettests.o] Error 1
```

ryanofsky:
Possible fix might be to change:

```
boost::signals2::scoped_connection c = wallet.NotifyTransactionChanged.connect(...);
```

to:

```
boost::signals2::scoped_connection c(wallet.NotifyTransactionChanged.connect(...));
```

paveljanik:
Hmm, all gcc's on openSUSE 42.2, ie. 4.8.5, 5.3.1 and 6.2.1, fail on the current master...

paveljanik:
With https://github.com/bitcoin/bitcoin/pull/9974#issuecomment-287550867 change, the build is OK. Will you please PR it?

paveljanik:
And another issue here. On plain Debian Jessie, with the gcc version 4.9.2 (Debian 4.9.2-10), this doesn't compile with:

```
qt/test/wallettests.cpp:34:6: error: no matching function for call to 'QTimer::singleShot(int, Qt::TimerType, {anonymous}::ConfirmSend()::<lambda()>)'
     });
      ^
qt/test/wallettests.cpp:34:6: note: candidates are:
In file included from /usr/include/x86_64-linux-gnu/qt5/QtCore/QTimer:1:0,
                 from ./qt/sendcoinsdialog.h:13,
                 from qt/test/wallettests.cpp:7:
/usr/include/x86_64-linux-gnu/qt5/QtCore/qtimer.h:81:17: note: static void QTimer::singleShot(int, const QObject*, const char*)
     static void singleShot(int msec, const QObject *receiver, const char *member);
                 ^
/usr/include/x86_64-linux-gnu/qt5/QtCore/qtimer.h:81:17: note:   no known conversion for argument 2 from 'Qt::TimerType' to 'const QObject*'
/usr/include/x86_64-linux-gnu/qt5/QtCore/qtimer.h:82:17: note: static void QTimer::singleShot(int, Qt::TimerType, const QObject*, const char*)
     static void singleShot(int msec, Qt::TimerType timerType, const QObject *receiver, const char *member);
                 ^
/usr/include/x86_64-linux-gnu/qt5/QtCore/qtimer.h:82:17: note:   candidate expects 4 arguments, 3 provided
Makefile:7991: recipe for target 'qt/test/qt_test_test_bitcoin_qt-wallettests.o' failed
make[2]: *** [qt/test/qt_test_test_bitcoin_qt-wallettests.o] Error 1
```

PR #9977, opened 2017-03-11
QA: getblocktemplate_longpoll.py should always use >0 fee tx

sdaftuar:
I saw a test failure that looked like it was due to a 0-fee transaction being generated.


MarcoFalke:
utACK 655df06fb67b558325d0301b7c543ea04bf2899b

PR #9981, opened 2017-03-13
Bloom: Consider witness script pushes in bloom filter check.

maaku:
Allows SPV nodes to construct bloom filters that match pushdata in the scriptWitness.

### DIFF
--- a/src/bloom.cpp
+++ b/src/bloom.cpp
@@ -196,6 +196,13 @@ bool CBloomFilter::IsRelevantAndUpdate(const CTransaction& tx)
             if (data.size() != 0 && contains(data))
                 return true;
         }
+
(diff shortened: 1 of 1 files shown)


sipa:
Is there any use case for this?

maaku:
For something like channel closure watching, you can give a filter that only matches certain pathways of the spend script (by matching against a pubkey that's only used on that path, for example).

sipa:
Ok, but this is not sufficient. A full solution would need:
* A way for lightweight client to request witnesses.
* A way for full nodes to prove that the provided witnesses are part of the block (by giving the branches of the witness tree and the coinbase).
* A way to match against witnesses like your PR does, but optionally.

Without having all 3 pieces, you're just going to degrade the Bloom filter (even) faster than before, without any way the client can use the result.

PR #9983, opened 2017-03-13
tests: Convert selected tests to using named RPC arguments

laanwj:
Convert selected tests to use named arguments in RPC calls. This covers a few of the important ones:

- `assumevalid.py`
- `blockchain.py`
- `merkle_blocks.py`
- `segwit.py`

As well as the test framework itself, `util.py`.

This makes invocations easier to read especially if booleans or lots of arguments are involved.

#### Reviewing

To review I'd suggest using the command:

    git diff --word-diff --word-diff-regex='[^[:space:],\(\)=]+'

This will regard the added argument names as one word, making it trivial to see what is added.

#### Process

As doing this is a bit of a finnicky process, I'm doing this a few tests at a time. I may add more to this PR later. In case people want to help, I created [this list](https://gist.github.com/laanwj/f55638598c697452199014b4faa6d679) of RPC calls with argument names to avoid having to refer to the individual help all the time.

To make sure a test has no use of positional arguments left you can add these lines to `__call__(self, *args, **argsn)` in `authproxy.py`:
```python
        if args:
            raise ValueError('TEST: supporting named arguments')
```
Every use of positional arguments will then raise that exception.

### DIFF
--- a/qa/rpc-tests/assumevalid.py
+++ b/qa/rpc-tests/assumevalid.py
@@ -88,7 +88,7 @@ class SendHeadersTest(BitcoinTestFramework):
 
         # Build the blockchain
         self.tip = int(self.nodes[0].getbestblockhash(), 16)
-        self.block_time = self.nodes[0].getblock(self.nodes[0].getbestblockhash())['time'] + 1
+        self.block_time = self.nodes[0].getblock(blockhash=self.nodes[0].getbestblockhash())['time'] + 1
 
         self.blocks = []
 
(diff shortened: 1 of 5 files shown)


MarcoFalke:
> This makes invocations easier to read especially if booleans or lots of arguments are involved.

Concept ACK. Though I think we don't need named args for calls that
only have a single and obvious param, such as generate( $num ) or any
single param call that starts with set*.

To aid reviewers even more, maybe in future pulls, a project wide
search and replace of eligible calls makes sense. Potentially one s&r
per commit, but no strong opinion on this.

laanwj:
> Concept ACK. Though I think we don't need named args for calls that
only have a single and obvious param, such as generate( $num ) or any
single param call that starts with set*.

I've thought about that. But some of the functions called with one argument are actually multi-arg functions of which only the first is used. And it may happen that more arguments get added to a function in the future. So I thought it better to be consistent and do all of them to move to a fully name-based API.

> To aid reviewers even more, maybe in future pulls, a project wide
> search and replace of eligible calls makes sense. Potentially one s&r
> per commit, but no strong opinion on this.

Yeah, maybe. The advantage of doing it per test is that it is possible to verify that all calls have been converted without having to convert the entire project, which would conflict with every possible PR that changes tests.

MarcoFalke:
> And it may happen that more arguments get added to a function in the future. So I thought it better to be consistent and do all of them to move to a fully name-based API.

Hmm. As long as downstream projects use the interface with positional
args, we can't do that. Though, maybe we should require named args for
new calls and whenever we extend an existing API call to encourage the
name-based API and then switch completely at some point in the future?

laanwj:
> Hmm. As long as downstream projects use the interface with positional args, we can't do that.

I don't say we should deprecate the positional arguments any time soon, just try to use it as our own best practice to use named arguments. Mixing them is kind of messy, IMO.

laanwj:
Closing, too much work to rebase all the time.

PR #9984, opened 2017-03-13
devtools: Make github-merge compute SHA512 from git, instead of worktree

laanwj:
This changes tree_sha512sum() to requests the objects for hashing from git instead of from the working tree.

The change should make the process more deterministic (it hashes what will be pushed) and hopefully avoids the frequent miscomputed SHA512's that happen now.

As an added bonus this can now compute treehashes for non-HEAD commits, although this functionality is not used.

### DIFF
--- a/contrib/devtools/github-merge.py
+++ b/contrib/devtools/github-merge.py
@@ -78,24 +78,53 @@ def get_symlink_files():
             ret.append(f.decode('utf-8').split("\t")[1])
     return ret
 
-def tree_sha512sum():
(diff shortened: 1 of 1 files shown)


MarcoFalke:
utACK a327e8ea30e10abdce910a3e37c78e00d0918548

laanwj:
@MarcoFalke thanks
I checked this against many recent merge commits and am now confident enough that this computes the same output as the old algorithm, to merge this.

PR #9985, opened 2017-03-13
[QT] Show more descriptive label for pay to yourself entries

NicolasDorier:
This show a more descriptive label for Pay To Yourself transaction entries in bitcoin QT. Showing the label of the first Output.

Before:
before

After:
after

Pay to yourself is common when using Bitcoin Core for tracking coins going from and to Watch-Only addresses. For example, in the case of TumblerBit, it will make it  easier to see when a coins is transitioning throught different states: Wallet =>Alice Escrow => Alice Offer => Alice Offer Redeem.

### DIFF
--- a/src/qt/transactionrecord.cpp
+++ b/src/qt/transactionrecord.cpp
@@ -103,8 +103,13 @@ QList<TransactionRecord> TransactionRecord::decomposeTransaction(const CWallet *
         {
             // Payment to self
             CAmount nChange = wtx.GetChange();
-
-            parts.append(TransactionRecord(hash, nTime, TransactionRecord::SendToSelf, "",
+            std::string address;
+            CTxDestination destAddress;
+            if (wtx.tx->vout.size() >= 1 && ExtractDestination(wtx.tx->vout[0].scriptPubKey, destAddress))
+            { 
+                address = CBitcoinAddress(destAddress).ToString();
+            }
+            parts.append(TransactionRecord(hash, nTime, TransactionRecord::SendToSelf, address,
                             -(nDebit - nChange), nCredit - nChange));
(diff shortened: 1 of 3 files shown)


laanwj:
Concept ACK

src/qt/transactionrecord.cpp
```diff
@@ -103,8 +103,13 @@ QList<TransactionRecord> TransactionRecord::decomposeTransaction(const CWallet *
         {
             // Payment to self
             CAmount nChange = wtx.GetChange();
-
-            parts.append(TransactionRecord(hash, nTime, TransactionRecord::SendToSelf, "",
+            std::string address;
+            CTxDestination destAddress;
+            if (wtx.tx->vout.size() >= 1 && ExtractDestination(wtx.tx->vout[0].scriptPubKey, destAddress))
```
jonasschnelli: This list only the first to-yourself-output. But I think this is okay.

da2ce7:
For 'send to yourself' transactions, maybe the row-background shading should be tinted a different colour. So it is more visually apparent what transactions leave the wallet?

luke-jr:
I think it may be a good idea to just remove send-to-self as a thing. In some cases (eg, CoinJoins), it is indistinguishable from send-to-someone-else + receive-from-someone-else in the same transaction...

NicolasDorier:
@da2ce7 there is the left column that already have a logo to distinguish.
@luke-jr I think this is a good idea, will try that

NicolasDorier:
@luke-jr an alternative would be to not consider watch-only as "self".

luke-jr:
CoinJoins aren't necessarily watch-only.

NicolasDorier:
I think https://github.com/bitcoin/bitcoin/pull/10007 is superior alternative.
Keeping open this one until it is clear we will not use it.

Issue #9192, opened 2016-11-20
Context menu options that are not available should be disabled (in transactions tab)


nopara73:
### Describe the issue
In Bitcoin QT, under the Transactions tab when right clicking on the transaction the context menu shows more option then it can always handle.

### Can you reliably reproduce the issue?
Yes.

#### If so, please list the steps to reproduce below:  
0. Open QT.  
1. Make an incoming transaction without label.  
2. Go to the Transactions tab.  
3. Rigth click on a transaction and select "Copy label".  

### Actual behaviour
The clipboard will be empty.  

### Expected behaviour
Copies the label. (What does not exist, of course.)

### What version of bitcoin-core are you using?
v0.13.1, Ubuntu PPA.  

### Any extra information that might be useful in the debugging process.
This behaviour is more problematic in edge cases.  
Like when you send funds with JoinMarket. In this case for example the "Copy address" option does not work.  

### Suggestion
I think when something cannot be copied that should not be present in the context menu in the first place.  
  
![](http://i.imgur.com/hLfAgeF.png)

laanwj:
> I think when something cannot be copied that should not be present in the context menu in the first place.

Yes, disabling the option would be the preferable thing to do in that case.

PR #9991, opened 2017-03-14
listreceivedbyaddress Filter Address

NicolasDorier:
Supersede https://github.com/bitcoin/bitcoin/pull/9503 created by @JeremyRubin , I will maintain it.

### DIFF
--- a/src/rpc/client.cpp
+++ b/src/rpc/client.cpp
@@ -43,6 +43,7 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "listreceivedbyaddress", 0, "minconf" },
     { "listreceivedbyaddress", 1, "include_empty" },
     { "listreceivedbyaddress", 2, "include_watchonly" },
+    { "listreceivedbyaddress", 3, "address_filter" },
     { "listreceivedbyaccount", 0, "minconf" },
     { "listreceivedbyaccount", 1, "include_empty" },
     { "listreceivedbyaccount", 2, "include_watchonly" },
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -1403,6 +1403,16 @@ UniValue ListReceived(CWallet * const pwallet, const UniValue& params, bool fByA
         if(params[2].get_bool())
             filter = filter | ISMINE_WATCH_ONLY;
 
+    bool has_filtered_address = false;
+    CTxDestination filtered_address = CNoDestination();
+    if (!fByAccounts && params.size() > 3) {
+        if (!IsValidDestinationString(params[3].get_str())) {
+            throw JSONRPCError(RPC_WALLET_ERROR, "address_filter parameter was invalid");
+        }
+        filtered_address = DecodeDestination(params[3].get_str());
+        has_filtered_address = true;
+    }
+
     // Tally
     std::map<CTxDestination, tallyitem> mapTally;
     for (const std::pair<uint256, CWalletTx>& pairWtx : pwallet->mapWallet) {
@@ -1421,6 +1431,10 @@ UniValue ListReceived(CWallet * const pwallet, const UniValue& params, bool fByA
             if (!ExtractDestination(txout.scriptPubKey, address))
                 continue;
 
+            if (has_filtered_address && !(filtered_address == address)) {
+                continue;
+            }
+
             isminefilter mine = IsMine(*pwallet, address);
             if(!(mine & filter))
                 continue;
@@ -1437,10 +1451,24 @@ UniValue ListReceived(CWallet * const pwallet, const UniValue& params, bool fByA
     // Reply
     UniValue ret(UniValue::VARR);
     std::map<std::string, tallyitem> mapAccountTally;
-    for (const std::pair<CTxDestination, CAddressBookData>& item : pwallet->mapAddressBook) {
-        const CTxDestination& dest = item.first;
-        const std::string& strAccount = item.second.name;
-        std::map<CTxDestination, tallyitem>::iterator it = mapTally.find(dest);
+
+    // Create mapAddressBook iterator
+    // If we aren't filtering, go from begin() to end()
+    auto start = pwallet->mapAddressBook.begin();
+    auto end = pwallet->mapAddressBook.end();
+    // If we are filtering, find() the applicable entry
+    if (has_filtered_address) {
+        start = pwallet->mapAddressBook.find(filtered_address);
+        if (start != end) {
+            end = std::next(start);
+        }
+    }
+
+    for (auto item_it = start; item_it != end; ++item_it)
+    {
+        const CTxDestination& address = item_it->first;
+        const std::string& strAccount = item_it->second.name;
+        auto it = mapTally.find(address);
         if (it == mapTally.end() && !fIncludeEmpty)
             continue;
 
@@ -1466,7 +1494,7 @@ UniValue ListReceived(CWallet * const pwallet, const UniValue& params, bool fByA
             UniValue obj(UniValue::VOBJ);
             if(fIsWatchonly)
                 obj.pushKV("involvesWatchonly", true);
-            obj.pushKV("address",       EncodeDestination(dest));
+            obj.pushKV("address",       EncodeDestination(address));
             obj.pushKV("account",       strAccount);
             obj.pushKV("amount",        ValueFromAmount(nAmount));
             obj.pushKV("confirmations", (nConf == std::numeric_limits<int>::max() ? 0 : nConf));
@@ -1511,15 +1539,15 @@ UniValue listreceivedbyaddress(const JSONRPCRequest& request)
         return NullUniValue;
     }
 
-    if (request.fHelp || request.params.size() > 3)
+    if (request.fHelp || request.params.size() > 4)
         throw std::runtime_error(
-            "listreceivedbyaddress ( minconf include_empty include_watchonly)\n"
+            "listreceivedbyaddress ( minconf include_empty include_watchonly address_filter )\n"
             "\nList balances by receiving address.\n"
             "\nArguments:\n"
             "1. minconf           (numeric, optional, default=1) The minimum number of confirmations before payments are included.\n"
             "2. include_empty     (bool, optional, default=false) Whether to include addresses that haven't received any payments.\n"
             "3. include_watchonly (bool, optional, default=false) Whether to include watch-only addresses (see 'importaddress').\n"
-
+            "4. address_filter    (string, optional) If present, only return information on this address.\n"
             "\nResult:\n"
             "[\n"
             "  {\n"
@@ -1541,6 +1569,7 @@ UniValue listreceivedbyaddress(const JSONRPCRequest& request)
             + HelpExampleCli("listreceivedbyaddress", "")
             + HelpExampleCli("listreceivedbyaddress", "6 true")
             + HelpExampleRpc("listreceivedbyaddress", "6, true, true")
+            + HelpExampleRpc("listreceivedbyaddress", "6, true, true, \"1M72Sfpbz1BPpXFHz9m3CdqATR44Jvaydd\"")
         );
 
     ObserveSafeMode();
@@ -3837,7 +3866,7 @@ static const CRPCCommand commands[] =
     { "wallet",             "listaddressgroupings",             &listaddressgroupings,          {} },
     { "wallet",             "listlockunspent",                  &listlockunspent,               {} },
     { "wallet",             "listreceivedbyaccount",            &listreceivedbyaccount,         {"minconf","include_empty","include_watchonly"} },
-    { "wallet",             "listreceivedbyaddress",            &listreceivedbyaddress,         {"minconf","include_empty","include_watchonly"} },
+    { "wallet",             "listreceivedbyaddress",            &listreceivedbyaddress,         {"minconf","include_empty","include_watchonly","address_filter"} },
     { "wallet",             "listsinceblock",                   &listsinceblock,                {"blockhash","target_confirmations","include_watchonly","include_removed"} },
     { "wallet",             "listtransactions",                 &listtransactions,              {"account","count","skip","include_watchonly"} },
     { "wallet",             "listunspent",                      &listunspent,                   {"minconf","maxconf","addresses","include_unsafe","query_options"} },
--- a/test/functional/wallet_listreceivedby.py
+++ b/test/functional/wallet_listreceivedby.py
@@ -45,10 +45,44 @@ class ReceivedByTest(BitcoinTestFramework):
         assert_array_result(self.nodes[1].listreceivedbyaddress(11), {"address": addr}, {}, True)
 
         # Empty Tx
-        addr = self.nodes[1].getnewaddress()
+        empty_addr = self.nodes[1].getnewaddress()
         assert_array_result(self.nodes[1].listreceivedbyaddress(0, True),
-                            {"address": addr},
-                            {"address": addr, "account": "", "amount": 0, "confirmations": 0, "txids": []})
+                            {"address": empty_addr},
+                            {"address": empty_addr, "account": "", "amount": 0, "confirmations": 0, "txids": []})
+
+        #Test Address filtering
+        #Only on addr
+        expected = {"address":addr, "account":"", "amount":Decimal("0.1"), "confirmations":10, "txids":[txid,]}
+        res = self.nodes[1].listreceivedbyaddress(minconf=0, include_empty=True, include_watchonly=True, address_filter=addr)
+        assert_array_result(res, {"address":addr}, expected)
+        assert_equal(len(res), 1)
+        #Error on invalid address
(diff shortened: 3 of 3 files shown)


NicolasDorier:
1. Rebased and fixed conflicts
2. Addressed @kallewoof nits
3. Add the new parameter to `client.cpp`

kallewoof (approved):
utACK 8c373d5feedb428d234278d484e5d2a932010fae

NicolasDorier:
@luke-jr @jnewbery nit addressed, can you re-ACK?

qa/rpc-tests/receivedby.py
```diff
@@ -62,10 +62,22 @@ def run_test(self):
...
         assert_array_result(self.nodes[1].listreceivedbyaddress(0,True),
-                           {"address":addr},
-                           {"address":addr, "account":"", "amount":0, "confirmations":0, "txids":[]})
+                           {"address":empty_addr},
+                           {"address":empty_addr, "account":"", "amount":0, "confirmations":0, "txids":[]})
+
+        #Test Address filtering
+        #Only on addr
+        expected = {"address":addr, "account":"", "amount":Decimal("0.1"), "confirmations":10, "txids":[txid,]}
+        res = self.nodes[1].listreceivedbyaddress(0, True, True, addr)
```
jnewbery: Nit: Consider using named arguments here, instead of positional arguments (since the meaning of the arguments is not clear without having to go back to the definition of listreceivedbyaddress).

qa/rpc-tests/receivedby.py
```diff
@@ -62,10 +62,22 @@ def run_test(self):
...
         #Empty Tx
-        addr = self.nodes[1].getnewaddress()
+        empty_addr = self.nodes[1].getnewaddress()
         assert_array_result(self.nodes[1].listreceivedbyaddress(0,True),
-                           {"address":addr},
-                           {"address":addr, "account":"", "amount":0, "confirmations":0, "txids":[]})
+                           {"address":empty_addr},
+                           {"address":empty_addr, "account":"", "amount":0, "confirmations":0, "txids":[]})
+
+        #Test Address filtering
```
jnewbery: I'd prefer this test if `other_addr` (currently declared on line 78) received some funds before you did your tests on `listreceivedbyaddress()` with a filter address. This isn't currently testing the case where the wallet contains multiple addresses with funds and you want to see the transactions received by one of those addresses. Ideally the test case would be:

- define addr and send funds to it (already done for you in lines 42-43)
- define other_addr and send funds to it
- call `listreceivedbyaddress()` filtering on `addr`. Assert only one transaction is returned.
- call `listreceivedbyaddress()` filtering on `other_addr`. Assert only one transaction is returned.
- call `listreceivedbyaddress()` with no filtering. Assert both transactions are returned.

jnewbery:
Looks good. A couple of nits and a suggestion on improving the test case.

JeremyRubin:
utAck

qa/rpc-tests/receivedby.py
NicolasDorier: Listreceivedbyaddress is not known at compile time, and thus has no definition in python. This is interpreted at runtime. So I can't really use named argument. Would inline comment be good ?

NicolasDorier:
@jnewbery I improved the tests, and fixed the alignement. I can't use named parameter in the test though.

qa/rpc-tests/receivedby.py
jnewbery: You can use named arguments in the RPC calls. See https://github.com/bitcoin/bitcoin/pull/9983 for example.

NicolasDorier:
added named parameters to one call to listreceivedbyaddress  in the tests for better readability.

jnewbery:
line 73 `minConf` needs to be `minconf`.

with that change, tested ACK.

NicolasDorier:
@jnewbery thanks done ;)

EthanHeilman:
utACK - the code looks good and this is very useful functionality.

NicolasDorier:
@TheBlueMatt I fixed nits and added tests, can you re-ACK ?

NicolasDorier:
https://github.com/bitcoin/bitcoin/pull/7729 should also work for TB purpose.

nopara73:
ACK - works fine.  
This could be (not sure if should be) combined with fixing this issue for correct GUI experience: https://github.com/bitcoin/bitcoin/issues/9192

NicolasDorier:
Closing. I am using a workaround now, I caching the whole `listtransaction` at every block. This call would have been super useful though.

jnewbery:
:disappointed: I'm happy to reACK if you decide to reopen this!

NicolasDorier:
@jnewbery, if you have interest, I am reopening, I guess it does not hurt to keep it opens, it is not high maintenance PR. I was thinking to allow multiple addresses to be passed in the filter, instead of only one.

For NTumbleBit, I changed the design, I have a method called `GetTransactions(address)` which should send back the received transactions as well as the outputs one of this address. `listreceivedbyaddress` would only least what the address received.

The way I ended up implemeting my need is by caching `listtransaction` then for each transaction in it, caching `gettransaction` in a internal database. I now have good performance, and I am not sure this RPC Method would improve my case.

sipa:
@NicolasDorier Do you actually need the wallet functionality (balance tracking, UTXO management, coin selection, ...), or just a means to be notified of certain transactions?

nopara73:
@sipa Need the wallet functionality for NTumbleBit.

sipa:
Can you clarify what functionality it relies on?

nopara73:
A lot, but I am not sure which are relevant to this discussion with the workaround. @NicolasDorier will have a better insight. https://github.com/NTumbleBit/NTumbleBit/tree/master/NTumbleBit.TumblerServer/Services/RPCServices

src/wallet/rpcwallet.cpp
```diff
@@ -1206,6 +1206,17 @@ UniValue ListReceived(CWallet * const pwallet, const UniValue& params, bool fByA
         if(params[2].get_bool())
             filter = filter | ISMINE_WATCH_ONLY;
 
+    bool fFilterAddress = false;
+    CTxDestination filterAddress =  CNoDestination();
+    if (!fByAccounts && params.size() > 3) {
```
jonasschnelli: Use a simpler check:
```cpp
CBitcoinAddress address(request.params[3].get_str());
    if (!address.IsValid())
```

src/wallet/rpcwallet.cpp
```diff
@@ -1240,10 +1255,24 @@ UniValue ListReceived(CWallet * const pwallet, const UniValue& params, bool fByA
...
     UniValue ret(UniValue::VARR);
     std::map<std::string, tallyitem> mapAccountTally;
-    for (const std::pair<CBitcoinAddress, CAddressBookData>& item : pwallet->mapAddressBook) {
-        const CBitcoinAddress& address = item.first;
-        const std::string& strAccount = item.second.name;
-        std::map<CBitcoinAddress, tallyitem>::iterator it = mapTally.find(address);
+
+    // Create mapAddressBook iterator
+    // If we aren't filtering, go from begin() to end()
+    auto start = pwallet->mapAddressBook.begin();
```
jonasschnelli: This loop-or-find logic seems to be a bit wired.
I probably would have factored out the JSON object packing and either called the new function from within the range based loop or after the `find()`.

jonasschnelli:
Needs rebase.

Concept ACK. I think the by address filter is in general useful.

NicolasDorier:
@sipa just saw now your previous question.
As far as TB is concerned, I am using balance tracking, UTXO management, and coin selection management of Bitcoin Core for funding the escrow transactions.

I am also using Core as a block explorer: There is some well known address on which I would like to be notified is money is sent *to* it, or sent *from*. The method allows me to know when something is sent to the address.
However, I still need a new rpc method for knowing if something is sent *from* an address.

Until I have the two features, my only way is to poll periodically `listtransaction` then fetch all transaction with `gettransaction`, and see if the relevant addresses are inside. This work well but is very cumbersome.

test/functional/receivedby.py
```diff
@@ -62,10 +62,42 @@ def run_test(self):
...
+
+        #Test Address filtering
+        #Only on addr
+        expected = {"address":addr, "account":"", "amount":Decimal("0.1"), "confirmations":10, "txids":[txid,]}
+        res = self.nodes[1].listreceivedbyaddress(minconf=0, include_empty=True, include_watchonly=True, only_address=addr)
+        assert_array_result(res, {"address":addr}, expected)
+        assert_equal(len(res), 1)
+        #Another address receive money
+        res = self.nodes[1].listreceivedbyaddress(0, True, True)
+        assert_equal(len(res), 3) #Right now 3 entries
```
luke-jr: Note, this needs to be 2->3 (instead of 3->4) due to removal of the "default address".

luke-jr:
Rebased here: https://github.com/luke-jr/bitcoin/tree/listreceivedbyaddress-filtered

src/wallet/rpcwallet.cpp
```diff
@@ -1206,6 +1206,17 @@ UniValue ListReceived(CWallet * const pwallet, const UniValue& params, bool fByA
         if(params[2].get_bool())
             filter = filter | ISMINE_WATCH_ONLY;
 
+    bool fFilterAddress = false;
+    CTxDestination filterAddress =  CNoDestination();
+    if (!fByAccounts && params.size() > 3) {
+        filterAddress =  CBitcoinAddress(params[3].get_str()).Get();
+        CTxDestination nulladdress = CNoDestination();
+        if (filterAddress == nulladdress) {
```
promag: Just `if (filterAddress == CNoDestination())`?

src/wallet/rpcwallet.cpp
```diff
@@ -1206,6 +1206,17 @@ UniValue ListReceived(CWallet * const pwallet, const UniValue& params, bool fByA
         if(params[2].get_bool())
             filter = filter | ISMINE_WATCH_ONLY;
 
+    bool fFilterAddress = false;
+    CTxDestination filterAddress =  CNoDestination();
+    if (!fByAccounts && params.size() > 3) {
+        filterAddress =  CBitcoinAddress(params[3].get_str()).Get();
```
promag: Remove extra space.

src/wallet/rpcwallet.cpp
```diff
@@ -1206,6 +1206,17 @@ UniValue ListReceived(CWallet * const pwallet, const UniValue& params, bool fByA
         if(params[2].get_bool())
             filter = filter | ISMINE_WATCH_ONLY;
 
+    bool fFilterAddress = false;
+    CTxDestination filterAddress =  CNoDestination();
```
promag: Remove extra space.

src/rpc/client.cpp
```diff
@@ -44,6 +44,7 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "listreceivedbyaddress", 0, "minconf" },
     { "listreceivedbyaddress", 1, "include_empty" },
     { "listreceivedbyaddress", 2, "include_watchonly" },
+    { "listreceivedbyaddress", 3, "only_address" },
```
promag: Not sure what others think, but I kind of prefer `filter_address`.

src/wallet/rpcwallet.cpp
```diff
@@ -1206,6 +1206,17 @@ UniValue ListReceived(CWallet * const pwallet, const UniValue& params, bool fByA
...
+    if (!fByAccounts && params.size() > 3) {
+        filterAddress =  CBitcoinAddress(params[3].get_str()).Get();
+        CTxDestination nulladdress = CNoDestination();
+        if (filterAddress == nulladdress) {
+            throw JSONRPCError(RPC_WALLET_ERROR, "only_address parameter was invalid");
```
promag: Missing test `assert_raises_rpc_error`.

src/wallet/rpcwallet.cpp
```diff
@@ -1314,15 +1343,15 @@ UniValue listreceivedbyaddress(const JSONRPCRequest& request)
...
         throw std::runtime_error(
-            "listreceivedbyaddress ( minconf include_empty include_watchonly)\n"
+            "listreceivedbyaddress ( minconf include_empty include_watchonly only_address )\n"
             "\nList balances by receiving address.\n"
             "\nArguments:\n"
             "1. minconf           (numeric, optional, default=1) The minimum number of confirmations before payments are included.\n"
             "2. include_empty     (bool, optional, default=false) Whether to include addresses that haven't received any payments.\n"
             "3. include_watchonly (bool, optional, default=false) Whether to include watch-only addresses (see 'importaddress').\n"
-
+            "4. only_address      (string, optional) If present, only return information on this address. Otherwise, return all information.\n"
```
promag: Drop `Otherwise, return all information`?

src/wallet/rpcwallet.cpp
```diff
@@ -1240,10 +1255,24 @@ UniValue ListReceived(CWallet * const pwallet, const UniValue& params, bool fByA
...
+    auto end = pwallet->mapAddressBook.end();
+    // If we are filtering, find() the applicable entry
+    if (fFilterAddress) {
+        start = pwallet->mapAddressBook.find(filterAddress);
+        if (start != end) {
+            end = std::next(start);
+        }
+    }
+
+    for(auto item_it = start; item_it != end; ++item_it)
```
promag: Nit, space after `for`, `{` on this line.

promag (commented):
Some minor comments, new code could use current convention, for instance, s/filterAddress/filter_address.

NicolasDorier:
Closing this I workedaround the need of this

JeremyRubin:
@NicolasDorier I think this should reopen, AFAIK LN people would like this too... @cfromknecht @Roasbeef

NicolasDorier:
ok rebasing

NicolasDorier:
Rebased, fixed the nits of @luke-jr and @promag .

I renamed `only_address` to `address_filter` so that if later we decide it is also possible to pass an array of address instead of a single address, we don't need to rename anything. It is more generic.

src/wallet/rpcwallet.cpp
NicolasDorier: This does not apply anymore.

src/wallet/rpcwallet.cpp
```diff
@@ -1403,6 +1403,16 @@ UniValue ListReceived(CWallet * const pwallet, const UniValue& params, bool fByA
         if(params[2].get_bool())
             filter = filter | ISMINE_WATCH_ONLY;
 
+    bool fFilterAddress = false;
```
kallewoof: The code convention has changed a bit recently. Now, we don't put the type as prefix, and we try to use snake case. I would rename this `has_filter_address` and the below to `filter_address`. You can leave existing vars as is, if you don't want the diff to grow for no reason though (like `fByAccounts` below).

src/wallet/rpcwallet.cpp
```diff
@@ -3837,7 +3866,7 @@ static const CRPCCommand commands[] =
     { "wallet",             "listaddressgroupings",             &listaddressgroupings,          {} },
     { "wallet",             "listlockunspent",                  &listlockunspent,               {} },
     { "wallet",             "listreceivedbyaccount",            &listreceivedbyaccount,         {"minconf","include_empty","include_watchonly"} },
-    { "wallet",             "listreceivedbyaddress",            &listreceivedbyaddress,         {"minconf","include_empty","include_watchonly"} },
+    { "wallet",             "listreceivedbyaddress",            &listreceivedbyaddress,         {"minconf","include_empty","include_watchonly","address_filter" } },
```
kallewoof: No space after `"` (end of line) -- compare to other lines.

kallewoof (approved):
utACK

NicolasDorier:
Addressed @kallewoof nits

kallewoof (approved):
utACK f087613719026bcd5cba95ec64c19361fcc71ecf

laanwj:
utACK f08761371

test/functional/wallet_listreceivedby.py
```diff
@@ -45,10 +45,44 @@ def run_test(self):
...
         # Empty Tx
-        addr = self.nodes[1].getnewaddress()
+        empty_addr = self.nodes[1].getnewaddress()
         assert_array_result(self.nodes[1].listreceivedbyaddress(0, True),
-                            {"address": addr},
-                            {"address": addr, "account": "", "amount": 0, "confirmations": 0, "txids": []})
+                            {"address": empty_addr},
+                            {"address": empty_addr, "account": "", "amount": 0, "confirmations": 0, "txids": []})
+
+        #Test Address filtering
```
luke-jr: Space after `#` is expected now it seems.

test/functional/wallet_listreceivedby.py
```diff
@@ -45,10 +45,44 @@ def run_test(self):
...
+        #Only on addr
+        expected = {"address":addr, "account":"", "amount":Decimal("0.1"), "confirmations":10, "txids":[txid,]}
```
luke-jr: Space after `:` is expected now it seems.

PR #9994, opened 2017-03-15
[trivial/doc] clarify current blockchain size

askmike:
EDIT: typo in commit message, not an english native!

### DIFF
--- a/doc/README.md
+++ b/doc/README.md
@@ -3,7 +3,7 @@ Bitcoin Core 0.14.99
 
 Setup
 ---------------------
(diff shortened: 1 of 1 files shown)


askmike:
Will open a new PR without typo!

sipa:
You can modify existing pull request, so you don't need to create a new one. Simply push additional commits to the same branch (though you may be asked to squash and force push).

askmike:
Ah, will do that next time! Thanks for the tip :)

PR #9999, opened 2017-03-15
[LevelDB] Plug leveldb logs to bitcoin logs

NicolasDorier:
Implementation ~copy paste~ inspired by https://github.com/bitcoin/bitcoin/blob/57b34599b2deb179ff1bd97ffeab91ec9f904d85/src/leveldb/util/posix_logger.h#L28

ping @laanwj

### DIFF
--- a/src/dbwrapper.cpp
+++ b/src/dbwrapper.cpp
@@ -14,6 +14,64 @@
 #include <leveldb/filter_policy.h>
 #include <memenv.h>
 #include <stdint.h>
+#include <algorithm>
+
+class CBitcoinLevelDBLogger : public leveldb::Logger {
+public:
+    // This code is adapted from posix_logger.h, which is why it is using vsprintf.
+    // Please do not do this in normal code
+    virtual void Logv(const char * format, va_list ap) override {
+            if (!LogAcceptCategory("leveldb"))
+                return;
+            char buffer[500];
+            for (int iter = 0; iter < 2; iter++) {
+                char* base;
+                int bufsize;
+                if (iter == 0) {
+                    bufsize = sizeof(buffer);
+                    base = buffer;
+                }
+                else {
+                    bufsize = 30000;
+                    base = new char[bufsize];
+                }
+                char* p = base;
+                char* limit = base + bufsize;
+
+                // Print the message
+                if (p < limit) {
+                    va_list backup_ap;
+                    va_copy(backup_ap, ap);
+                    // Do not use vsnprintf elsewhere in bitcoin source code, see above.
+                    p += vsnprintf(p, limit - p, format, backup_ap);
+                    va_end(backup_ap);
+                }
+
+                // Truncate to available space if necessary
+                if (p >= limit) {
+                    if (iter == 0) {
+                        continue;       // Try again with larger buffer
+                    }
+                    else {
+                        p = limit - 1;
+                    }
+                }
+
+                // Add newline if necessary
+                if (p == base || p[-1] != '\n') {
+                    *p++ = '\n';
+                }
+
+                assert(p <= limit);
+                base[std::min(bufsize - 1, (int)(p - base))] = '\0';
+                LogPrintStr(base);
+                if (base != buffer) {
+                    delete[] base;
+                }
+                break;
+            }
+    }
+};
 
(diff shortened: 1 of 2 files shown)


laanwj:
Nice! utACK https://github.com/bitcoin/bitcoin/pull/9999/commits/9d21779d9e18546c027d6fe7064289d0e7491efa

src/dbwrapper.cpp
```diff
@@ -14,6 +14,62 @@
...
+            base = new char[bufsize];
+        }
+        char* p = base;
+        char* limit = base + bufsize;
+
+        // Print the message
+        if (p < limit) {
+            va_list backup_ap;
+            va_copy(backup_ap, ap);
+            p += vsnprintf(p, limit - p, format, backup_ap);
```
laanwj: nit: I'd prefer not to use `vsnprintf` here. We've purged this from the code when introducing tinyformat. Wonder if we can do something based on out `strprintf`. Though bridging between C varargs and C++ is likely going to be difficult, and this solution as-taken-from LevelDB should be fine. It's just fairly ugly code and should not be taken as a hint to contributors that using system printf is okay in other places.

laanwj: Maybe just add a comment in that regard, mentioning where the code comes from and why and explaining the above.

NicolasDorier:
Added some comments about the ugly vsprintf.  Bridging varargs to use `strprintf` is difficult and error proned. I do not think it is worth it.

src/dbwrapper.h
```diff
@@ -44,6 +45,11 @@ const std::vector<unsigned char>& GetObfuscateKey(const CDBWrapper &w);
 
 };
 
+class CBitcoinLogger : public leveldb::Logger {
```
laanwj: CBitcoinLevelDBLogger maybe? The name is too general otherwise, we should reserve this class name for ourselves.

NicolasDorier:
Renamed. Can Travis be restarted ? I do not think the error is related to this PR.

src/dbwrapper.h
```diff
@@ -44,6 +45,11 @@ const std::vector<unsigned char>& GetObfuscateKey(const CDBWrapper &w);
 
 };
 
+class CBitcoinLevelDBLogger : public leveldb::Logger {
```
sipa: Why is this class in the .h file? It seems to be only used internally in dbwrapper.cpp.

NicolasDorier: well, h for declaration and cpp for implementation, at least that is what I always thought. You mean just putting the class directly in the cpp ?

sipa: .h is for the externally accessible parts of a cpp file. This class doesn't need to be exposed to anything, so it can just go inside the cpp file.

src/init.cpp
```diff
@@ -430,7 +430,7 @@ std::string HelpMessage(HelpMessageMode mode)
         strUsage += HelpMessageOpt("-limitdescendantsize=<n>", strprintf("Do not accept transactions if any ancestor would have more than <n> kilobytes of in-mempool descendants (default: %u).", DEFAULT_DESCENDANT_SIZE_LIMIT));
         strUsage += HelpMessageOpt("-bip9params=deployment:start:end", "Use given start/end times for specified BIP9 deployment (regtest-only)");
     }
-    std::string debugCategories = "addrman, alert, bench, cmpctblock, coindb, db, http, libevent, lock, mempool, mempoolrej, net, proxy, prune, rand, reindex, rpc, selectcoins, tor, zmq"; // Don't translate these and qt below
+    std::string debugCategories = "addrman, alert, bench, cmpctblock, coindb, db, http, libevent, lock, mempool, mempoolrej, net, proxy, prune, rand, reindex, rpc, selectcoins, tor, zmq, leveldb"; // Don't translate these and qt below
```
jnewbery: Please retain alphabetical ordering (leveldb should be between http and libevent)

src/dbwrapper.h
laanwj: Right, .h is for *interface* declaration. Although with C++ it's not always clear what should be part of the interface and what not (especially when tests are involved), this one could just as well be private to the .cpp.

NicolasDorier:
fixed nits

Issue #9996, opened 2017-03-15
Running python tests fails


NicolasDorier:
After doing a `git clean -dfx` and rebuilding, the python tests started failing:

Test output:
```
root@0f51c0a75128:/home/bitcoin/bitcoin# python3 qa/pull-tester/rpc-tests.py receivedby
.
receivedby.py:
2017-03-15 02:31:58.943000 TestFramework (INFO): Initializing test directory /tmp/testmvbdk1ho/592
2017-03-15 02:31:59.240000 TestFramework (ERROR): Unexpected exception caught during testing
Traceback (most recent call last):
  File "/home/bitcoin/bitcoin/qa/rpc-tests/test_framework/test_framework.py", line 145, in main
    self.setup_chain()
  File "/home/bitcoin/bitcoin/qa/rpc-tests/test_framework/test_framework.py", line 49, in setup_chain
    initialize_chain(self.options.tmpdir, self.num_nodes, self.options.cachedir)
  File "/home/bitcoin/bitcoin/qa/rpc-tests/test_framework/util.py", line 258, in initialize_chain
    wait_for_bitcoind_start(bitcoind_processes[i], rpc_url(i), i)
  File "/home/bitcoin/bitcoin/qa/rpc-tests/test_framework/util.py", line 216, in wait_for_bitcoind_start
    raise Exception('bitcoind exited with status %i during initialization' % process.returncode)
Exception: bitcoind exited with status 1 during initialization
2017-03-15 02:31:59.264000 TestFramework (INFO): Stopping nodes

stderr:
: Error opening block database.
Please restart with -reindex or -reindex-chainstate to recover.
Traceback (most recent call last):
  File "/home/bitcoin/bitcoin/qa/rpc-tests/receivedby.py", line 147, in <module>
    ReceivedByTest().main()
  File "/home/bitcoin/bitcoin/qa/rpc-tests/test_framework/test_framework.py", line 162, in main
    stop_nodes(self.nodes)
  File "/home/bitcoin/bitcoin/qa/rpc-tests/test_framework/util.py", line 380, in stop_nodes
    for i, node in enumerate(nodes):
TypeError: 'NoneType' object is not iterable

Pass: False, Duration: 1 s

TEST          | PASSED | DURATION

receivedby.py | False  | 1 s

ALL           | False  | 1 s (accumulated)

Runtime: 1 s
root@0f51c0a75128:/home/bitcoin/bitcoin#
```

Debug Logs:
```
[...]

PR #10000, opened 2017-03-15
[DO NOT MERGE][LevelDB] Do no crash if filesystem can't fsync

NicolasDorier:
Solve https://github.com/bitcoin/bitcoin/issues/9996.
There should be some kind of warning if fsync is not supported. It is unclear how to implement that easily though since we don't have access to any logger here.

I encountered the problem as I could not run bitcoin on a network share on cifs. (I could previously, which is strange)

It seems that https://www.kernel.org/doc/Documentation/filesystems/cifs/CHANGES Version 1.57 seems to indicate that cifs should not fail, but just not guarantee that the server properly synched on the drive. This is not what I am experiencing.

Ping @laanwj

### DIFF
--- a/src/leveldb/util/env_posix.cc
+++ b/src/leveldb/util/env_posix.cc
@@ -231,7 +231,7 @@ class PosixWritableFile : public WritableFile {
       if (fd < 0) {
         s = IOError(dir, errno);
       } else {
-        if (fsync(fd) < 0) {
+        if (fsync(fd) < 0 && errno != EINVAL) {
           s = IOError(dir, errno);
         }
         close(fd);


laanwj:
To be clear: this happens only happens while syncing a **directory** handle in `SyncDirIfManifest()`, syncing normal fds must still succeed.

> It is unclear how to implement that easily though since we don't have access to any logger here.

I'd say we should warn *once* when this happens, that data syncing can not be guaranteed over network filesystems.

Even if it succeeds all that fsync can (generally) guarantee with network filesystems is that the data has been written to the server, not that the server has written it to disk

However if there is no access to a logger from there it becomes difficult...

luke-jr:
Shouldn't we at least log a warning in this case? Maybe even fail if pruning is enabled? (Presumably it means a crash is likely to corrupt the index?)

NicolasDorier:
It should not crash. There is reasons to use shared folder, one is in my posted issue. 
If there is a corruption, then the user need to reindex. This is strictly better than just preventing to run at all.

I agree for a warning. We can attempt fsync into bitcoin code on a dummy folder to see if it is enabled, and if not, provide a warning.

luke-jr:
Pruning nodes don't have data to reindex from. I know if I downloaded 100 GB, and then had to re-do the download due to this, I'd be pretty annoyed... Better to find out sooner so I can use a different filesystem and avoid the re-download.

laanwj:
No, it should never crash/fail on this. If people run a datadirectory on a network FS they take some extra risk. There's nothing that we can do in that case.

laanwj:
> I agree for a warning. We can attempt fsync into bitcoin code on a dummy folder to see if it is enabled, and if not, provide a warning.

Yeah - maybe just try fsync() on the data directory at startup? No need even to make a dummy dir.

NicolasDorier:
for using `fsync` on the data directory I need to use `LEVELDB_PLATFORM_WINDOWS` and `port.h` of level db in `dbwrapper.h` it feels dirty but I do not see another way. Any suggestion ?

laanwj:
> for using fsync on the data directory I need to use LEVELDB_PLATFORM_WINDOWS and port.h of level db in dbwrapper.h it feels dirty but I do not see another way. Any suggestion ?

To do that, best is probably to check separately `fsync()` is available in `configure.ac`. LevelDB compilation flags are not available while compiling our files.

>Another way is to test in DB::Open, log the warning, and then make special case in the CBitcoinLevelDBLogger that if the log message start is "Warning, fsync directory disabled...", it should log as error. But it also feels dirty.

Yes. Unfortunately they don't deal in log levels. Making an exception for a single message is indeed ugly. You could do it linux-kernel `printk` style and add a prefix (for example `=e=<message>`) to promote something to error.

NicolasDorier:
The easiest would be prefixed log.
I am wondering if logging it really matter actually. As it is failure of fsync happens only if it is a folder. fdatasync works as expected.

The fsync on a folder is only done if Sync is called on a MANIFEST file. Which seems to be rather rare.

laanwj:
> The fsync on a folder is only done if Sync is called on a MANIFEST file. Which seems to be rather rare. Will check my assertion soon.

The manpage says

>       Calling fsync() does not necessarily ensure that the entry in the directory containing the file has also reached disk.  For that an explicit fsync() on a file descriptor for the directory  is also needed.

So this seems only important after adding new files (or deleting them), which is probably why it's only done on MANIFEST changes.

NicolasDorier:
So I think we do not really to show a log as this is rare occurence. (I want to still check that with dbcache to 0 though)

I also want to note also that on windows, running bitcoin core from a shared SMB folder works fine.

NicolasDorier:
I think in any case we should merge it. Windows version works fine on SMB file system, only the linux version crash.

laanwj:
@NicolasDorier I agree

marukka:
@NicolasDorier What exactly are you doing to get the Windows version to work on a SMB share? In bug #10018 I had the exact opposite experience using a SMB share from a Server 2016 file server failover cluster with a Server 2016 Remote Desktop Services server publishing Bitcoin Core as a RemoteApp. Did you map the share as a drive letter as opposed to using a UNC path?

NicolasDorier:
> Did you map the share as a drive letter as opposed to using a UNC path?

Yes I did. Indeed // path did not work, I think this is why I did it.

NicolasDorier:
Where to ACK that, here or in https://github.com/bitcoin-core/leveldb/pull/16 ?

NicolasDorier:
merged on https://github.com/bitcoin-core/leveldb/pull/16

PR #10007, opened 2017-03-16
[QT] Remove SendToSelf, and break down its payouts

NicolasDorier:
Alternative to https://github.com/bitcoin/bitcoin/pull/9985 following @luke-jr suggestion. I think it looks great.

The rational is that pay to yourself is mostly done by third party tool, like tumbler bit and joinmarket using watch-only addresses to track payments.

I aggregate all the inputs into one entry with debit.
If their is only one input, the description of the line is taken from its address.
And one entry per output with credit.

As you can see in this example, you can follow where the coins flow from the escrow, to the redeem (timeout) or offer. 

Before:
before

After:
capture

### DIFF
--- a/src/qt/addresstablemodel.cpp
+++ b/src/qt/addresstablemodel.cpp
@@ -422,6 +422,8 @@ bool AddressTableModel::removeRows(int row, int count, const QModelIndex &parent
  */
 QString AddressTableModel::labelForAddress(const QString &address) const
 {
+    if (address.isEmpty())
+        return QString();
     {
         LOCK(wallet->cs_wallet);
         CBitcoinAddress address_parsed(address.toStdString());
--- a/src/qt/transactionrecord.cpp
+++ b/src/qt/transactionrecord.cpp
@@ -101,12 +101,56 @@ QList<TransactionRecord> TransactionRecord::decomposeTransaction(const CWallet *
 
         if (fAllFromMe && fAllToMe)
         {
-            // Payment to self
-            CAmount nChange = wtx.GetChange();
+            TransactionRecord sub;
 
-            parts.append(TransactionRecord(hash, nTime, TransactionRecord::SendToSelf, "",
-                            -(nDebit - nChange), nCredit - nChange));
-            parts.last().involvesWatchAddress = involvesWatchAddress;   // maybe pass to TransactionRecord as constructor argument
+            // Add one entry for all the outputs
+            for (unsigned int nOut = 0; nOut < wtx.tx->vout.size(); nOut++)
+            {
+                const CTxOut& txout = wtx.tx->vout[nOut];
+                sub = TransactionRecord(hash, nTime);
+                sub.idx = nOut;
+                sub.involvesWatchAddress = involvesWatchAddress;
+                CTxDestination address;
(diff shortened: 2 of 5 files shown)


jonasschnelli:
Nice.
Concept ACK.

luke-jr:
> If their is only one input, the description of the line is taken from its address.

Uh, concept NACK this part. Inputs don't have addresses... Not sure what to do for a description when we only have some of the inputs, but this isn't it.

NicolasDorier:
@luke-jr maybe you have a better way to reach my objective, which I think is reasonable use case.

Take a look at this scenario:

TB

If Alice manage to pay 1.0 Tumbler by the Client Fullfill transaction here is what I would like to see in the wallet:

```
Tumbler  [1.0]
Offer       [-1.0]
Offer       [1.0]
Escrow    [-1.0]
Escrow    [1.0]
(n/a)        [-1.0] 
```

As you can see, it would be possible to follow the coins.

luke-jr:
We don't even support labels for inputs (nor would it make sense to, since they are considered fungible). Labels shown are always for the outputs only.

(I have no idea how to interpret your image, or what you're trying to do there.)

NicolasDorier:
Ah. These are transaction with top of the line being the conditions the input solved. And bottom the output.

NicolasDorier:
In the screenshot of the PR you see several tumbling session where alice put the coin into escrow, but timeout and get the coin back through redeem.

luke-jr:
I don't see that there's any possible confusion over a case with one input and one output: if they're both ours, we display the transaction as a send (with label from the output address), and also a receive (again, with label from the output address). Fairly straight-forward there. If there are merely more outputs, each one is a "send".

The confusion comes in when you only control a subset of the inputs. This isn't particular to send-to-self, though: even if you don't get any of the outputs, it's still confusing. How do we handle that right now? Seems logical to keep the same behaviour in this PR - if that behaviour isn't sane, another PR can change it...

NicolasDorier:
If you only control a subset of the inputs, then this case is not reached because `fAllFromMe` is `false`.

luke-jr:
Well, then at least this case is straightforward: Show each output as a "send" with the output's label unless it's considered change. :)

NicolasDorier:
@luke-jr it is what I am doing already.

NicolasDorier:
Closing, I ended up making NTumbleBit its own CLI commands for querying the state. Bitcoin QT is useless for showing any layer 2 information without this commit.

laanwj:
Sorry to hear you couldn't get agreement on this.

NicolasDorier:
@laanwj  I think this should be discussed at a more fundamental level: Does Bitcoin Core should consider at all the need of layer 2 users? 

The conflict point was that the notion of "input address" should be non existent in layer 1. But it makes perfect sense for layer 2. Which beg the question of: Does bitcoin core QT solely focus on layer 1.

But the payment-to-self, which is very used for layer 2, are actually not useful at all for layer 1.

sipa:
@NicolasDorier In the above when referring to Bitcoin Core, you mean its built-in wallet?

NicolasDorier:
yes, I mean specifically Bitcoin-QT user interface in this case.

Issue #10004, opened 2017-03-15
After max chain of unconfirmed change transactions, last tx is missing from memory until rescan


pinheadmz:
## Summary:
https://github.com/bitcoin/bitcoin/blob/57b34599b2deb179ff1bd97ffeab91ec9f904d85/doc/release-notes/release-notes-0.12.0.md

_"Bitcoin Core 0.12 also introduces new default policy limits on the length and size of unconfirmed transaction chains that are allowed in the mempool (generally limiting the length of unconfirmed chains to 25 transactions, with a total size of 101 KB)."_

1) Generate a chain of transactions with unconfirmed change outputs until this maximum is hit.
2) The last tx in this chain does is not accessible from `gettransaction`, does not appear in `listunspent`  and will not be mined into a block until the node is restarted with `-rescan` (luckily 0.14.0 saves mempool to disk so the tx is "found" on rescan)

## Demonstration:

Start node, generate 101 blocks so only a single 50 BTC output is mature enough to spend:
```
$ ./bitcoind -daemon -regtest
Bitcoin server starting
$ ./bitcoin-cli -regtest getinfo
{
  "version": 140000,
  "protocolversion": 70015,
  "walletversion": 130000,
  "balance": 0.00000000,
  ...
}
$ ./bitcoin-cli -regtest generate 101
[
  "271d2c725a2279976bee7650bf20f0a16ccbbc626e07150fcb5e8e8429909b09", 
  ... 
  "13e07e2376aa2ce1d161497e05156632262571a17bb6e1cbd333392e23b973cc"
]
$ ./bitcoin-cli -regtest getinfo
{
  "version": 140000,
  "protocolversion": 70015,
  "walletversion": 130000,
  "balance": 50.00000000,
  ...
}
```
Now attempt to create 30 x 0.5 BTC transactions, forcing bitcoind to spend unconfirmed change outputs in a chain until the limit is hit:
```
$ for n in {1..30}; do ./bitcoin-cli -regtest sendtoaddress 2NCY1Q1Mu6V5uha7fDzGNYYPUs5F1zSdXCL 0.5; done
fb15572accafe0b6d2b7b2806f31bda16280c7d97a91d99ca929d4236e5099c1
c7cea6a964ab01f20bfc54e647fc2ea9b86f1a73f64ae547dcb5fb77a9071180
1cca3556fa953a0826d34d5096353a6809df1479996caff9e47954b1f6f81dce
[...]

Issue #9752, opened 2017-02-13
Max unconfirmed chainlength 


Joukehofman:
### Describe the issue
Because transaction malleability isn't that much of an issue lately, we've decided to let the wallet spend zero conf change. It has happened that we create a transaction just moments before a significant increase in (higher paying) transactions. Because of #9645, there is a chance that you end up with an unconfirmed chain of 24 child transactions. If you let the wallet produce a 25th child, it will return an error because the mempool won't accept it.  This is also problematic in the case we would like to cpfp the whole chain, because other nodes won't relay the transaction to the miners.

#### If so, please list the steps to reproduce below:
1. Predict an increase in transactions. (You need a crystal ball)
2. Create transaction
3. Create more transactions
4. ?
5. Profit

### Expected behaviour
Ideally it would stop if the chain is 24 transactions long, so one could try a cpfp transaction to compensate for the fees.

### Actual behaviour
It will create a transaction chain of 25 transaction, making it hard to have the whole chain confirmed in a certain time.

### What version of bitcoin-core are you using?
Official binary: 0.13.1

Joukehofman:
A quick way would be to have a setting for allowed length of unconfirmed transactions in the mempool, and push the cpfp-transaction on an other node.

jonasschnelli:
I think https://github.com/bitcoin/bitcoin/pull/9262 addresses that issue (partially?).
This fix is available in 0.13.2 (as well as in 0.14).
But #9262 is still disabled by default (`DEFAULT_WALLET_REJECT_LONG_CHAINS = false;`)

@Joukehofman:
You could try to upgrade to 0.13.2 and enable `-walletrejectlongchains=true`.

instagibbs:
Yes `-walletrejectlongchains` sounds like what you want. It will simply reject the proposed transaction, allowing you to try something else.

Joukehofman:
It is indeed what I want! Thanks for mentioning @jonasschnelli and fixing it @instagibbs !

instagibbs:
Your use-case might be an argument for stopping 1 shorter even, to allow for a single CPFP in the mempool(with that one hitting the chain limit). As-is it might not quite work.

Joukehofman:
Ah, I thought that was the intention, but you only catch it if mempool doesn't accept the transaction. But, this option, combined with a setting for allowed mempool length would solve my issue in a nice way.

instagibbs:
[...]

PR #9290, opened 2016-12-06
Make RelayWalletTransaction attempt to AcceptToMemoryPool.

gmaxwell:
This resolves an issue where a wallet transaction which failed to
 relay previously because it couldn't make it into the mempool
 will not try again until restart, even though mempool conditions
 may have changed. Rebroadcast of other transactions is unchanged and works as it always did.

Abandoned and known-conflicted transactions are skipped.

Some concern was expressed that there may be users with many
 unknown conflicts would waste a lot of CPU time trying to
 add them to their memory pools over and over again.  But I am
 doubtful these users exist in any number, if they do exist
 they have worse problems, and they can mitigate any performance
 issue this might have by abandoning the transactions in question.

PR #10015, opened 2017-03-16
Wallet should reject long chains by default

jnewbery:
#9262 introduced `-walletrejectlongchains`, which prevents the wallet from adding transactions which would form a chain which is too long to be accepted by the mempool. However, it's set to false by default.

There are legitimate reasons for wanting the wallet to accept longer chains than the mempool, and #9290 does ensure that wallet transactions will be resubmitted to the mempool eventually. However, in the mainline case the behaviour is confusing for the user and the wallet should reject the long chains. See for instance #10004 and #9752.

This PR sets `-walletrejectlongchains` to true by default, and makes the wallet reject transactions that form a chain up to (maxlength - 1). This means that the user has one final chance to add a transaction to the chain by restarting with `-walletrejectlongchains` set to false. This could be helpful if a user wants to use CPFP to unstick a transaction package with too low fees.

Tests also updated to verify new behaviour.

This PR also changes the error message returned if Coin Selection fails. Currently, if Coin Selection fails, then the error returned to the user is "Insufficient Funds". This is misleading as Coin Selection can fail for other reasons, such as if the transaction would form a chain that is too long to enter the mempool. This commit changes the error message returned from "Insufficient Funds" to "Coin Selection Failed" and updates the test scripts to expect the correct error message.

### DIFF
--- a/qa/rpc-tests/fundrawtransaction.py
+++ b/qa/rpc-tests/fundrawtransaction.py
@@ -316,7 +316,7 @@ class RawTransactionsTest(BitcoinTestFramework):
         rawtx   = self.nodes[2].createrawtransaction(inputs, outputs)
         dec_tx  = self.nodes[2].decoderawtransaction(rawtx)
 
-        assert_raises_jsonrpc(-4, "Insufficient funds", self.nodes[2].fundrawtransaction, rawtx)
+        assert_raises_jsonrpc(-4, "Coin Selection Failed", self.nodes[2].fundrawtransaction, rawtx)
 
         ############################################################
         #compare fee of a standard pubkeyhash transaction
@@ -472,7 +472,7 @@ class RawTransactionsTest(BitcoinTestFramework):
         rawTx = self.nodes[1].createrawtransaction(inputs, outputs)
         # fund a transaction that requires a new key for the change output
         # creating the key must be impossible because the wallet is locked
-        assert_raises_jsonrpc(-4, "Insufficient funds", self.nodes[1].fundrawtransaction, rawtx)
+        assert_raises_jsonrpc(-4, "Coin Selection Failed", self.nodes[1].fundrawtransaction, rawtx)
 
         #refill the keypool
         self.nodes[1].walletpassphrase("test", 100)
--- a/qa/rpc-tests/wallet.py
+++ b/qa/rpc-tests/wallet.py
@@ -359,23 +359,34 @@ class WalletTest (BitcoinTestFramework):
         # So we should be able to generate exactly chainlimit txs for each original output
         sending_addr = self.nodes[1].getnewaddress()
         txid_list = []
-        for i in range(chainlimit*2):
+        for i in range((chainlimit - 1 ) * 2):
             txid_list.append(self.nodes[0].sendtoaddress(sending_addr, Decimal('0.0001')))
-        assert_equal(self.nodes[0].getmempoolinfo()['size'], chainlimit*2)
-        assert_equal(len(txid_list), chainlimit*2)
+        assert_equal(self.nodes[0].getmempoolinfo()['size'], (chainlimit - 1) * 2)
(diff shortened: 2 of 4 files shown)


qa/rpc-tests/fundrawtransaction.py
```diff
@@ -316,7 +316,7 @@ def run_test(self):
         rawtx   = self.nodes[2].createrawtransaction(inputs, outputs)
         dec_tx  = self.nodes[2].decoderawtransaction(rawtx)
 
-        assert_raises_jsonrpc(-4, "Insufficient funds", self.nodes[2].fundrawtransaction, rawtx)
+        assert_raises_jsonrpc(-4, "Coin Selection Failed", self.nodes[2].fundrawtransaction, rawtx)
```
lontivero: The new error message is a bit confussing to me because it doesn't provides a reason or error code. IMO if something can fail for two different reasons, it should let the caller know the reason in order to let him take the appropiate actions, show the best error message, etc.

jnewbery: I agree it'd be better to have more precise error codes/messages returned. However, I think this is an improvement since we're returning "Insufficient funds" when the problem may be something different.

laanwj: Agree that if there are really insufficient funds it'd be good to reply that. A specific error is better than a non-specific one *if the message is correct*. Better an aspecific error than an incorrect one though. An incorrect "insufficient funds" error could set the caller on a wild goose chase or get them to panic. In that case it's better to return something general which prompts the caller to do their own research.

NicolasDorier:
NACK on "Coin Selection Failed". It would break existing code (I rely in some project on the Insufficient Funds message), and I think "Insufficient funds" is correct. If you can't create a transaction because the chain is too long, then you really do not have enough liquidity.

qa/rpc-tests/fundrawtransaction.py
NicolasDorier: I think "Insufficient funds" is correct. If the user does not have enough liquidity to not do a long chain, then he does not have enough funds.

jnewbery:
@NicolasDorier I strongly disagree with your definition of 'liquidity' here. In the too-long-chains case, the coin selection is being rejected by *local* policy. If the user was able to get the transaction to a miner who was prepared to mine a long chain, then it would be a perfectly valid transaction. I also think the most obvious reading of "insufficient funds" is that there are not enough funds in the wallet, not that the funds are temporarily unusable for technical reasons.

That said, I do have sympathy for client applications which expect the interface to bitcoind to remain stable. It's just that in this case the interface is wrong and should be corrected. If your client application is propagating the error message to the end user that there are insufficient funds, then it's likely to cause alarm and anxiety for users who may think that their funds have been lost.

Ideally, we'd have error messages that were specific and correct, and that's where we should aim to get to. The choice here is between log messages that are (specific and potentially incorrect) and log messages that are (general and correct). Log messages that are general and correct are preferable.

NicolasDorier:
@jnewbery well in my case, the Insufficient Fee is interpreted into a proper Enum value in my client code.
Making this change mean that my code will throw a UnknownException, interpreted like a bug, instead of a Insufficient Fee, which delay the same action for later. (and if long chains, this is how it should be handled as well)

That said this PR should solve my issue https://github.com/bitcoin/bitcoin/issues/9874 .
At least "Insufficient Fee" so you do not break older code, and maybe add "Too long chain" as a new error message in the case that this PR attempt to fix.

NicolasDorier:
I advice "too-long-mempool-chain" as it is the same error message sent by `sendrawtransaction`.

jnewbery:
Doesn't seem to be much appetite for this. There was a bit of disagreement on IRC here: https://botbot.me/freenode/bitcoin-core-dev/2017-03-16/?msg=82546050&page=3

Closing for now.

PR #10017, opened 2017-03-16
combine_logs.py - aggregates log files from multiple bitcoinds during functional tests.

jnewbery:
This commit adds a proof-of-concept tool for combining log files from multiple instances of bitcoinds as well as the test_framework.log file when running a functional test. This gives a combined view of what the test framework and all bitcoin instances were doing during the test.

There are 3 options for output: plain text, colorized text for posix terminals, and html:

Plaintext:

screenshot from 2017-03-16 17-29-16

Colorized text:

screenshot from 2017-03-16 17-29-44

HTML:

screenshot from 2017-03-16 17-31-40

Possible future enhancements:
- add some very simple javascript to the HTML output to allow filtering of log events by source, severity, timestamp, regex, etc.
- have the test_runner generate combined logs automatically if a test fails.
- have automated build tools like travis to generate and push an html log of a testcase if it fails

### DIFF
--- a/test/functional/combine_logs.py
+++ b/test/functional/combine_logs.py
@@ -1,0 +1,111 @@
+#!/usr/bin/env python3
+"""Combine logs from multiple bitcoin nodes as well as the test_framework log.
+
+This streams the combined log output to stdout. Use combine_logs.py > outputfile
+to write to an outputfile."""
+
+import argparse
+from collections import defaultdict, namedtuple
+import glob
+import heapq
+import os
+import re
+import sys
+
+# Matches on the date format at the start of the log event
+TIMESTAMP_PATTERN = re.compile(r"^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d{6}")
+
+LogEvent = namedtuple('LogEvent', ['timestamp', 'source', 'event'])
+
+def main():
+    """Main function. Parses args, reads the log files and renders them as text or html."""
+
+    parser = argparse.ArgumentParser(usage='%(prog)s [options] <test temporary directory>', description=__doc__)
+    parser.add_argument('-c', '--color', dest='color', action='store_true', help='outputs the combined log with events colored by source (requires posix terminal colors. Use less -r for viewing)')
+    parser.add_argument('--html', dest='html', action='store_true', help='outputs the combined log as html. Requires jinja2. pip install jinja2')
+    args, unknown_args = parser.parse_known_args()
+
+    if args.color and os.name != 'posix':
+        print("Color output requires posix terminal colors.")
+        sys.exit(1)
+
+    if args.html and args.color:
+        print("Only one out of --color or --html should be specified")
+        sys.exit(1)
+
+    # There should only be one unknown argument - the path of the temporary test directory
+    if len(unknown_args) != 1:
+        print("Unexpected arguments" + str(unknown_args))
+        sys.exit(1)
+
+    log_events = read_logs(unknown_args[0])
+
+    print_logs(log_events, color=args.color, html=args.html)
+
+def read_logs(tmp_dir):
+    """Reads log files.
+
+    Delegates to generator function get_log_events() to provide individual log events
+    for each of the input log files."""
+
+    files = [("test", "%s/test_framework.log" % tmp_dir)]
+    for i, logfile in enumerate(glob.glob("%s/node*/regtest/debug.log" % tmp_dir)):
+        files.append(("node%d" % i, logfile))
+
+    return heapq.merge(*[get_log_events(source, f) for source, f in files])
+
+def get_log_events(source, logfile):
+    """Generator function that returns individual log events.
+
+    Log events may be split over multiple lines. We use the timestamp
+    regex match as the marker for a new log event."""
+    try:
+        with open(logfile, 'r') as infile:
+            event = ''
+            timestamp = ''
+            for line in infile:
+                # skip blank lines
+                if line == '\n':
+                    continue
+                # if this line has a timestamp, it's the start of a new log event.
+                time_match = TIMESTAMP_PATTERN.match(line)
+                if time_match:
+                    if event:
+                        yield LogEvent(timestamp=timestamp, source=source, event=event.rstrip())
+                    event = line
+                    timestamp = time_match.group()
+                # if it doesn't have a timestamp, it's a continuation line of the previous log.
+                else:
+                    event += "\n" + line
+            # Flush the final event
+            yield LogEvent(timestamp=timestamp, source=source, event=event.rstrip())
+    except FileNotFoundError:
+        print("File %s could not be opened. Continuing without it." % logfile, file=sys.stderr)
+
+def print_logs(log_events, color=False, html=False):
+    """Renders the iterator of log events into text or html."""
+    if not html:
+        colors = defaultdict(lambda: '')
+        if color:
+            colors["test"] = "\033[0;36m"   # CYAN
+            colors["node0"] = "\033[0;34m"  # BLUE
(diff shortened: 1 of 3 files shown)


laanwj:
Nice! Concept ACK

practicalswift:
@jnewbery Good stuff! Just a minor nit - consider running `flake8` on `combine_logs.py` and fix the warnings to bring it in line with PEP-8 :-)

These are the warnings:
```
      1  B901 blind except: statement
      1  E221 multiple spaces before operator
      1  E231 missing whitespace after ','
     10  E251 unexpected spaces around keyword / parameter equals
      5  E261 at least two spaces before inline comment
      6  E262 inline comment should start with '# '
      3  E302 expected 2 blank lines, found 1
     12  E501 line too long (… > 79 characters)
      1  E701 multiple statements on one line (colon)
      1  E999 SyntaxError: invalid syntax
```

jnewbery:
Thanks @practicalswift . I've addressed the flake8 warnings except:

```
4     E302 expected 2 blank lines, found 1
1     E305 expected 2 blank lines after class or function definition, found 1
15    E501 line too long (80 > 79 characters)
```

which I don't care about.

MarcoFalke:
Indeed. Really nice!

TheBlueMatt:
Awesome! Can we get travis to use this when it dumps failed debug logs (I believe its via test_framework.py).

jnewbery:
@TheBlueMatt that's definitely possible. I'd envision travis pushing the html logs to S3 or similar and printing a link in its build log.

But first things first: this needs a bit of review. Once it's merged we can look at integrating it into our build process.

MarcoFalke:
Tested double-rainbow ACK

screenshot from 2017-03-19 10-44-05

qa/pull-tester/combine_logs.py
```diff
@@ -0,0 +1,111 @@
...
+import re
+import sys
+
+# Matches on the date format at the start of the log event
+TIMESTAMP_PATTERN = re.compile(r"^\d{4}-\d{2}-\d{2} \d{2}:\d{2}:\d{2}\.\d{6}")
+
+LogEvent = namedtuple('LogEvent', ['timestamp', 'source', 'event'])
+
+def main():
+    """Main function. Parses args, reads the log files and reners them as text or html."""
```
paveljanik: reners -> renders

qa/pull-tester/combine_logs.py
```diff
@@ -0,0 +1,111 @@
...
+
+    parser = argparse.ArgumentParser(usage='%(prog)s [options] <test temporary directory>', description=__doc__)
+    parser.add_argument('-c', '--color', dest='color', action='store_true', help='outputs the combined log with events colored by source (requires posix terminal colors. Use less -r for viewing)')
+    parser.add_argument('--html', dest='html', action='store_true', help='outputs the combined log as html. Requires jinja2. pip install jinja2')
+    args, unknown_args = parser.parse_known_args()
+
+    if args.color and os.name != 'posix':
+        print("color output requires posix terminal colors.")
```
paveljanik: color -> Color

qa/pull-tester/combine_logs.py
```diff
@@ -0,0 +1,111 @@
...
+            colors["node2"] = "\033[0;31m"  # RED
+            colors["node3"] = "\033[0;33m"  # YELLOW
+            colors["reset"] = "\033[0;0m"   # WHITE
+
+        for event in log_events:
+            print("{0} {1: <5} {2} {3}".format(colors[event.source.rstrip()], event.source, event.event, colors["reset"]))
+
+    else:
+        try:
+            import jinja2
```
paveljanik: This new dependency should be mentioned somewhere.

paveljanik (commented):
Concept ACK!

This is nice!

qa/pull-tester/combine_logs.py
laanwj: It's mentioned in the --help message:

> parser.add_argument('--html', dest='html', action='store_true', help='outputs the combined log as html. Requires jinja2. pip install jinja2')

As it's only required when `--html` is provided and not by default I'd say that is enough.

jnewbery: I agree. HTML rendering is not expected to be the mainline use case. If you do want to render as HTML, the documentation is exactly where it should be, at the point of use (and with instructions on how to resolve the problem):

```python
        try:
            import jinja2
        except ImportError:
            print("jinja2 not found. Try `pip install jinja2`")
            sys.exit(1)
```

jnewbery:
Thanks for the review @paveljanik . Nits addressed in https://github.com/bitcoin/bitcoin/pull/10017/commits/af802c8b45f6fa913dcc0442a48343cb75a521e5.

qa/pull-tester/combine_logs.py
paveljanik: Yes.

qa/pull-tester/combine_logs.py
```diff
@@ -0,0 +1,111 @@
...
+    """Renders the iterator of log events into text or html."""
+    if not html:
+        colors = defaultdict(lambda: '')
+        if color:
+            colors["test"] = "\033[0;36m"   # CYAN
+            colors["node0"] = "\033[0;34m"  # BLUE
+            colors["node1"] = "\033[0;32m"  # GREEN
+            colors["node2"] = "\033[0;31m"  # RED
+            colors["node3"] = "\033[0;33m"  # YELLOW
+            colors["reset"] = "\033[0;0m"   # WHITE
```
laanwj: Nit: `\033[0m` suffices, and the WHITE comment should probably go as the reset-color depends on the terminal settings.

qa/pull-tester/combine_logs.py
```diff
@@ -0,0 +1,111 @@
...
+                        yield LogEvent(timestamp=timestamp, source=source, event=event.rstrip())
+                    event = line
+                    timestamp = time_match.group()
+                # if it doesn't have a timestamp, it's a continuation line of the previous log.
+                else:
+                    event += "\n" + line
+            # Flush the final event
+            yield LogEvent(timestamp=timestamp, source=source, event=event.rstrip())
+    except FileNotFoundError:
+        print("File %s could not be opened. Continuing without it." % logfile, file=sys.stderr)
```
laanwj: Not a bug in this tool, but I noticed a divergence with the sorting of messages: logging from the test framework is in local time, logging from the nodes is in UTC. This causes, in Europe, all the test messages to sink to the end. I think all logging should be UTC.

laanwj (approved):

laanwj:
The script needs to be moved for #9956

qa/pull-tester/combine_logs.py
jnewbery: Yes, you're right - all logging should be in UTC. I've pushed a commit which makes the functional test framework log in UTC. Can you retest with the latest tip?

qa/pull-tester/combine_logs.py
jnewbery: Thanks - fixed

jnewbery:
I've rebased and moved the script to /test/functional, addressed @laanwj's nit https://github.com/bitcoin/bitcoin/pull/10017#discussion_r107131372 and updated logging in the functional test framework to always log in UTC.

jnewbery:
fixup commits squashed https://github.com/jnewbery/bitcoin/tree/pr10017.1 -> https://github.com/jnewbery/bitcoin/tree/pr10017.2

laanwj:
Re-tested, looks good to me now, everything is in the right order. I think this can turn out to be really useful in debugging synchronization issues. Removing the [POC] tag.

PR #10023, opened 2017-03-17
[tests] remove maxblocksinflight.py (functionality covered by other test)

jnewbery:
**EDIT: we can remove the maxblocksinflight test entirely. See https://github.com/bitcoin/bitcoin/pull/10023#issuecomment-293891932**

~maxblocksinflight.py had a very strange structure where the~
~fucntionality of the test was a method of the NodeConnCB object. This~
~commit tidies up the structure of the test and brings it in line with~
~the other functional tests.~


fanquake:
Needs a rebase.

jnewbery:
rebased

sdaftuar:
Restructuring this sounds good, but now that we don't send getdata's in response to inv's anymore, perhaps this test should be rewritten to use headers for announcements instead?

jnewbery:
You're correct. On closer inspection, this test case is actually testing nothing. The only thing it asserts on is that the node hasn't sent us too many getdata requests, but since nodes no longer sends getdata messages in response to block invs, it's just testing that 0 < 256:

```
./maxblocksinflight.py
2017-04-12 22:07:31.285000 TestFramework (INFO): Initializing test directory /tmp/user/1000/testk493pd32/12847
2017-04-12 22:07:31.563000 TestFramework.mininode (INFO): Connecting to Bitcoin Node: 127.0.0.1:13956
2017-04-12 22:07:33.574000 TestFramework (INFO): Round 0: success (total requests: 0)
2017-04-12 22:07:35.577000 TestFramework (INFO): Round 1: success (total requests: 0)
2017-04-12 22:07:37.582000 TestFramework (INFO): Round 2: success (total requests: 0)
2017-04-12 22:07:39.611000 TestFramework (INFO): Round 3: success (total requests: 0)
2017-04-12 22:07:39.611000 TestFramework (INFO): Stopping nodes
```

the headers-first sync'ing is covered pretty well by sendheaders.py. Should we just retire this test case?

sdaftuar:
Well, I'd say that if it's not too much work, we should just fix this test case to test the maxblocksinflight behavior correctly.  Ie, announce 17 or more block headers, and verify we only get 16 getdata requests.

jnewbery:
This line: https://github.com/bitcoin/bitcoin/blob/70f6f56e9dde0d0eb1a51e70c051b8fa3ba9535f/test/functional/sendheaders.py#L517 and the block below test that when there are more than 16 headers messages sent, the peer will only send get_data for 16 of them.

I think that covers the maxblocksinflight test, but let me know if you think it needs expanding.

jnewbery:
@sdaftuar - as discussed, we can remove maxblocksinflight entirely.

test/functional/maxblocksinflight.py
```diff
@@ -1,93 +0,0 @@
...
-MAX_REQUESTS = 128
-
-class TestManager(NodeConnCB):
-    def on_getdata(self, conn, message):
-        self.log.debug("got getdata %s" % repr(message))
-        # Log the requests
-        for inv in message.inv:
-            if inv.hash not in self.blockReqCounts:
-                self.blockReqCounts[inv.hash] = 0
-            self.blockReqCounts[inv.hash] += 1
```
jimmysong: If I'm understanding correctly, this method doesn't get called anymore since `getdata` isn't the response to `inv`?

test/functional/maxblocksinflight.py
```diff
@@ -1,93 +0,0 @@
...
-                    current_invs = []
-            if len(current_invs) > 0:
-                self.connection.send_message(msg_inv(current_invs))
-
-            # Wait and see how many blocks were requested
-            time.sleep(2)
-
-            total_requests = 0
-            with mininode_lock:
-                for key in self.blockReqCounts:
```
jimmysong: And this should be an empty dict, correct?

test/functional/maxblocksinflight.py
```diff
@@ -1,93 +0,0 @@
...
-                    total_requests += self.blockReqCounts[key]
-                    if self.blockReqCounts[key] > 1:
-                        raise AssertionError("Error, test failed: block %064x requested more than once" % key)
-            if total_requests > MAX_REQUESTS:
```
jimmysong: And finally, `total_requests` is 0 every time?

jimmysong (approved):
ACK 5f4bcf28ef8a1b775b12c9ff06367fd6656e91c3

This removes a test that doesn't do anything.

test/functional/maxblocksinflight.py
jnewbery: correct

test/functional/maxblocksinflight.py
jnewbery: yes

test/functional/maxblocksinflight.py
jnewbery: exactly, and 0 <= 128 last time I checked, so this test always passes.

PR #10027, opened 2017-03-18
Set to nullptr after delete

practicalswift:


earonesty (commented):
LGTM: seems like a case where things like unique_ptr, etc. should be used instead of bare pointers.

jtimon:
ACK d93b97fbcfdd5cd255a456ee4d8d6d8020103862

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/10027/commits/d93b97fbcfdd5cd255a456ee4d8d6d8020103862

PR #10029, opened 2017-03-18
Fix parameter naming inconsistencies between .h and .cpp files

practicalswift:
Inconsistencies prior to this commit:

```
* serializeFlags vs serialFlags
src/core_io.h:std::string EncodeHexTx(const CTransaction& tx, const int serializeFlags = 0);
src/core_write.cpp:std::string EncodeHexTx(const CTransaction& tx, const int serialFlags)

* statusOut vs outStatus
src/rpc/server.h:bool RPCIsInWarmup(std::string *statusOut);
src/rpc/server.cpp:bool RPCIsInWarmup(std::string *outStatus)

* hashesToUpdate vs vHashesToUpdate
src/txmempool.h:    void UpdateTransactionsFromBlock(const std::vector<uint256> &hashesToUpdate);
src/txmempool.cpp:void CTxMemPool::UpdateTransactionsFromBlock(const std::vector<uint256> &vHashesToUpdate)

* nPruneUpToHeight vs nManualPruneHeight
src/validation.h:void PruneBlockFilesManual(int nPruneUpToHeight);
src/validation.cpp:void PruneBlockFilesManual(int nManualPruneHeight)
```

### DIFF
--- a/src/core_write.cpp
+++ b/src/core_write.cpp
@@ -114,9 +114,9 @@ std::string ScriptToAsmStr(const CScript& script, const bool fAttemptSighashDeco
     return str;
 }
 
-std::string EncodeHexTx(const CTransaction& tx, const int serialFlags)
+std::string EncodeHexTx(const CTransaction& tx, const int serializeFlags)
 {
-    CDataStream ssTx(SER_NETWORK, PROTOCOL_VERSION | serialFlags);
+    CDataStream ssTx(SER_NETWORK, PROTOCOL_VERSION | serializeFlags);
     ssTx << tx;
     return HexStr(ssTx.begin(), ssTx.end());
 }
--- a/src/rpc/server.h
+++ b/src/rpc/server.h
@@ -68,7 +68,7 @@ void SetRPCWarmupStatus(const std::string& newStatus);
 void SetRPCWarmupFinished();
 
 /* returns the current warmup state.  */
-bool RPCIsInWarmup(std::string *statusOut);
+bool RPCIsInWarmup(std::string *outStatus);
 
 /**
  * Type-check arguments; throws JSONRPCError if wrong type given. Does not check that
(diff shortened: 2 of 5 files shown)


paveljanik:
Report univalue upstream.

practicalswift:
@paveljanik Thanks for noticing! Removed univalue inconsistency from this PR.

paveljanik:
```
bitcoin-10029 pavel$ git grep hashesToUpdate
src/txmempool.cpp:// for each entry, look for descendants that are outside hashesToUpdate, and
bitcoin-10029 pavel$ 
```

practicalswift:
@paveljanik Thanks! Fixed, squashed and pushed! :-)

src/core_io.h
```diff
@@ -26,7 +26,7 @@ std::vector<unsigned char> ParseHexUV(const UniValue& v, const std::string& strN
 
 // core_write.cpp
 std::string FormatScript(const CScript& script);
-std::string EncodeHexTx(const CTransaction& tx, const int serializeFlags = 0);
+std::string EncodeHexTx(const CTransaction& tx, const int serialFlags = 0);
```
paveljanik: Here, I prefer the longer name. Flags are about the serialization not about the "serial" number or so. But kets wait for other opinions before changing it right away ;-)

laanwj: Yes definitely use serializeFlags.

laanwj:
Why would we care about divergences of internal and external parameter name? I think this unnecessarily opens the window for naming bikeshedding discussions.

MarcoFalke:
I think the downside is that it looks kinda odd in the doxygen
auto-suggest search bar:

PruneBlockFilesManual
> PruneBlockFilesManual(int nManualPruneHeight): validation.cpp
> PruneBlockFilesManual(int nPruneUpToHeight): validation.cpp

src/core_io.h
practicalswift: @paveljanik @laanwj Good point! Fixed, squashed and pushed! :-)

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/10029/commits/97b82136740fc625acd2251fe722fd4f0872c244

MarcoFalke:
utACK 97b8213

fanquake:
utACK 97b8213

jonasschnelli:
utACK #97b82136740fc625acd2251fe722fd4f0872c244

PR #10035, opened 2017-03-19
Fix init README format to render correctly on github

jlopp:
Needs preformatted tags


laanwj:
why close?

jlopp:
Sorry, I realized I was requesting to merge into 0.14 rather than master, which I assume was incorrect.

PR #10036, opened 2017-03-19
Fix init README format to render correctly on github

jlopp:
Just need some preformatted tags to fix the github rendering.

### DIFF
--- a/contrib/init/README.md
+++ b/contrib/init/README.md
@@ -1,12 +1,12 @@
 Sample configuration files for:
-
+```
 SystemD: bitcoind.service
 Upstart: bitcoind.conf
 OpenRC:  bitcoind.openrc
          bitcoind.openrcconf
 CentOS:  bitcoind.init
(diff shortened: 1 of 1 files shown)


fanquake:
In future, please just force push changes to an existing PR rather than open a new one.

jlopp:
Changes are the same as the other PR; I didn't know that you could change the base target via github UI.

paveljanik:
Hmm, I do not understand why we have both `doc/init.md` and `contrib/init/README.md`. And why do they share the list of files in `contrib/init`? And why OS X's plist is not in the list in `doc/init.md`.

laanwj:
Documentation in `doc/init.md` is for users.

The README.md in the directory is meant as a description of that directory, and why it exists. It could (but does not currently) point to the documentation.

paveljanik:
But the change itself is OK ;-)

fanquake:
trivial ACK b99fbad

PR #10037, opened 2017-03-20
Trivial: Fix typo in help getrawtransaction RPC

keystrike:
This fixes two typos in rawtransaction.cpp. When getrawtransaction is called and verbose is false, serialized hex-encoded data is returned. Only when it is true is the JSON object returned. There is also a grammatical error.


paveljanik:
Good catch!

ACK https://github.com/bitcoin/bitcoin/pull/10037/commits/05a9f22358be27c08f7e164d05eeeaf0386073cb

fanquake:
utACK 05a9f22

MarcoFalke:
Tagging for backport.

PR #10038, opened 2017-03-20
Add mallocinfo mode to `getmemoryinfo` RPC

laanwj:
This adds a mode argument to `getmemoryinfo`. By default the output will remain the same. However if a mode argument of `mallocinfo` is provided the result of glibc `malloc_info` (if available) will be returned as a string, as-is.

This output contains detailed malloc heap statistics, and is useful for tracking heap usage over time or troubleshooting memory fragmentation issues.

### DIFF
--- a/src/rpc/misc.cpp
+++ b/src/rpc/misc.cpp
@@ -20,6 +20,9 @@
 #endif
 
 #include <stdint.h>
+#ifdef HAVE_MALLOC_INFO
+#include <malloc.h>
+#endif
 
 #include <boost/assign/list_of.hpp>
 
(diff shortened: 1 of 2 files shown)


src/rpc/misc.cpp
```diff
@@ -505,13 +531,27 @@ UniValue getmemoryinfo(const JSONRPCRequest& request)
...
             + HelpExampleRpc("getmemoryinfo", "")
         );
-    UniValue obj(UniValue::VOBJ);
-    obj.push_back(Pair("locked", RPCLockedMemoryInfo()));
-    return obj;
+
+    std::string mode = (request.params.size() < 1 || request.params[0].isNull()) ? "stats" : request.params[0].get_str();
+    if (mode == "stats") {
+        UniValue obj(UniValue::VOBJ);
+        obj.push_back(Pair("locked", RPCLockedMemoryInfo()));
```
jonasschnelli: Don't we need to escape this in some ways? Or is that handled by UniValue?

laanwj: That's handled by UniValue. The whole object is returned as a JSON string.

laanwj:
As for the GLIBC 2.10 symbol requirement, the minimum version of glibc in our symbol checker is 2.11, so this is acceptable for binary releases: https://github.com/bitcoin/bitcoin/blob/master/contrib/devtools/symbol-check.py#L43

jonasschnelli:
Nice. To bad there is no clang equivalent of `malloc_info()`. This will be linux only.
Tested ACK e141aa4ba604ff22c68454112501c166d3e892c9
https://bitcoin.jonasschnelli.ch/build/PR/10038


bildschirmfoto 2017-03-20 um 12 12 06

bildschirmfoto 2017-03-20 um 12 16 54

laanwj:
> Nice. To bad there is no clang equivalent of malloc_info(). This will be linux only.

Clang on Linux simply uses glibc. But yeah, other libc's will either have equivalent calls or maybe not, it's specific to the malloc implementation.

fanquake:
Tested on OSX ACK e141aa4
screen shot

paveljanik:
postmerge ACK (was merged in the middle of my testing)

laanwj:
> postmerge ACK (was merged in the middle of my testing)

Yeah sorry for that, getting this in soon is kind of important due to all the reported OOM issues with 0.14.

TheBlueMatt:
postumous utACK e141aa4ba604ff22c68454112501c166d3e892c9

luke-jr:
Is there a reason not to just add it as an additional key in the result?

laanwj:
> Is there a reason not to just add it as an additional key in the result?

I did that first, but having a huge encoded XML string field in the default JSON object output was ugly. Also this information is potentially slow to compile, so it's better to have it on an separate entry point.

PR #10039, opened 2017-03-20
Fix compile errors with Qt 5.3.2 and Boost 1.55.0

ryanofsky:
Both errors were introduced in #9974 and reported by @paveljanik

### DIFF
--- a/src/qt/test/wallettests.cpp
+++ b/src/qt/test/wallettests.cpp
@@ -22,7 +22,9 @@ namespace
 //! Press "Yes" button in modal send confirmation dialog.
 void ConfirmSend()
 {
-    QTimer::singleShot(0, Qt::PreciseTimer, []() {
+    QTimer* timer = new QTimer;
+    timer->setSingleShot(true);
+    QObject::connect(timer, &QTimer::timeout, []() {
         for (QWidget* widget : QApplication::topLevelWidgets()) {
             if (widget->inherits("SendConfirmationDialog")) {
                 SendConfirmationDialog* dialog = qobject_cast<SendConfirmationDialog*>(widget);
@@ -32,6 +34,7 @@ void ConfirmSend()
             }
         }
     });
+    timer->start(0);
 }
 
 //! Send coins to address and return txid.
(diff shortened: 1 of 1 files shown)


paveljanik:
Going to test this. Thanks!

paveljanik:
Compiles OK on Debian Jessie and openSUSE (gcc 4.8.5, 5.3.1 and 6.2.1).

ACK
https://github.com/bitcoin/bitcoin/pull/10039/commits/b5bec4e330fc7201d989663b4dbc6a1e620dd0f9

pstratem:
This chance breaks compiling on debian 8.7 with qt4 ( 4:4.8.6+git64-g5dc8b2b+dfsg-3+deb8u1 ).

I suspect one version works in qt4 and the other in qt5.

pstratem:
@ryanofsky ping

jonasschnelli:
Maybe it's time to think about dropping support to compile against Qt4? We already dropped support during runtime...

laanwj:
> Maybe it's time to think about dropping support to compile against Qt4? We already dropped support during runtime...

The issue for discussion that is #8263

src/qt/test/wallettests.cpp
```diff
@@ -22,7 +22,9 @@ namespace
 //! Press "Yes" button in modal send confirmation dialog.
 void ConfirmSend()
 {
-    QTimer::singleShot(0, Qt::PreciseTimer, []() {
+    QTimer* timer = new QTimer;
+    timer->setSingleShot(true);
+    QObject::connect(timer, &QTimer::timeout, []() {
```
gmaxwell: This fails to compile for me on debian testing with qt/test/wallettests.cpp:36:6: error: no matching function for call to ‘QObject::connect(QTimer*&, void (QTimer::*)(), 
                  {anonymous}::ConfirmSend()::<lambda()>)’ });

The compile is successful if I revert this commit.

ryanofsky: #10098 fixes this

PR #9483, opened 2017-01-06
Complete hybrid full block SPV mode

jonasschnelli:
This is the complete patch-set for the **hybrid full block SPV mode**.

If one enables the SPV mode with `-spv=1` it does...
* ...first sync all headers (no block downloads during that phase)
* ...requests and persist all blocks that are relevant for the wallet (down to the dept of the older wallet key)
* ...scan the block for relevant transactions and flag them with `validated = false` (visible in `listtransactions` etc).
* ... continue with IBD (initial block download) after all wallet relevant blocks have been processed

Pure full block SPV mode is possible by setting `-autorequestblocks=0`, in that mode, no blocks for validating the chain will be downloaded, resulting in a SPV only mode.

For better testing, this PR also includes a bump to 0.0005 for the default fallback fee.

Including all required GUI changes and RPC tests:

Screenshots:

bildschirmfoto 2017-01-06 um 17 21 24

untitled-1

bildschirmfoto 2017-01-06 um 17 34 09

untitled-2

PR #10040, opened 2017-03-20
wallet: use headers chain for anti fee sniping

### LINKED ISSUE #10020: Setting nLockTime on all transactions allows offline clients to be fingerprinted
An unsynchronized client sets nLockTime to its current height (or possibly slightly further back). There is a privacy implication which was not discussed in pull #2340. Since offline clients have a different chainActive.Height(), it is possible to link different wallets to individual clients. This additional metadata can tie everything back to the client's blockchain state even when different wallets are swapped in.

One possible fix would be to not set nLockTime on a transaction if the client has not recently connected to the network.

ghost:


gmaxwell:
Does leak that you were in IBD, but I think there isn't much that could be done about that.  One possibility would be to use the best headers height which you'll have very early in sync... but it would require reasoning about some corner cases. @petertodd

jonasschnelli:
I think taking the headers height as @gmaxwell describes would make more sense. #9483 has some simple pre-work that would be reusable (https://github.com/bitcoin/bitcoin/pull/9483/commits/1687a0e5bc958bd204644f2d5ad3647ab461d65d).

rat4:
@jonasschnelli indeed

MarcoFalke:
W
​asn't the original issue talking about a client that is offline in a sense
of air gap (but was connected once to the network) and less about a client
that is currently syncing?

In which case the current solution does not help either, because the
headers' height can be assumed to be identifier just as unique as the chain
height. I think the best we can do for offline clients is to not set it at
all.​

sipa:
@MarcoFalke I guess we can do both. When offline (= tip header too old), don't use anti fee sniping; otherwise, use tip header information.

keystrike:
Yes, in the initial issue I described this because of the ability to fingerprint clients which are not synchronized for some time and will not synchronize in the future. The goal is to get rid of this unique metadata. I agree with @sipa's suggestion.

laanwj:
Needs rebase; and comments addressed.

keystrike:
I've made a comment on this issue [here](https://github.com/bitcoin/bitcoin/issues/10020#issuecomment-311339203).

JohnDolittle:
Rebased and awaiting review.

petertodd:
In addition to the IsInitialChainDownload() check, it may be good to change the randomization thing to pick a random block in a much wider range, rather than just a few blocks back. Then make the IsInitialChainDownload() check also activate that randomization for all txs.

JohnDolittle:
I'd prefer to remove the randomization thing or leave it as is.

jonasschnelli:
The extra headers CChain object is not really required (at least not for this use case).
I would go back to your original simple implementation of anti-fee sniping and just use `pindexBestHeader` as the header you get your height from.

JohnDolittle:
Removed that headers CChain.

petertodd:
Concept NACK

W/ the various SHA256 alts that have been launched, and will be launched, it's quite possible that you have a headers chain with more blocks in it than the actual Bitcoin chain. This is even possible if that chain isn't the actual most-work chain, as you might not know about that chain.

I think re: privacy, changing the anti-fee sniping mechanism to occasionally pick nLockTime's from a much larger range is probably the better way to help the privacy problem.

ryanofsky:
Perhaps this issue could be tagged https://github.com/bitcoin/bitcoin/labels/Up%20for%20grabs if it's not still active. If I understand the suggestion, it's basically just to replace 100 with a much larger value here:

https://github.com/bitcoin/bitcoin/blob/22e301a3d56dc9e6878380ee92c7d19ca43119d2/src/wallet/wallet.cpp#L2668

laanwj:
Closing this, seems inactive.

MarcoFalke:
Picked up in " wallet: Avoid leaking nLockTime fingerprint when anti-fee-sniping #15039 "

Issue #10041, opened 2017-03-21
ec_privkey_export_der() seems to be called as compressed mode ALWAYS. Is it on purpose?


kobake:
## What version of bitcoin-core are you using?
GitHub latest master. (3192e5278abca7c1f3b4a2a7f77a0ce941c73985)

### "compressed" argument of ec_privkey_export_der() seems to be bool-typed.
ec_privkey_export_der function has a "compressed" argument and it accepts bool-typed value (0 or not 0) as far as I see the implementation of the function.

https://github.com/bitcoin/bitcoin/blob/master/src/key.cpp#L70
https://github.com/bitcoin/bitcoin/blob/master/src/secp256k1/contrib/lax_der_privatekey_parsing.c#L56
```cpp
if (compressed) {
    ....
} else {
    ....
}
```

### Passed values for the "compressed" are NOT ZERO ALWAYS.
Actually, however, the passed values for the argument "compressed" are SECP256K1_EC_COMPRESSED or SECP256K1_EC_UNCOMPRESSED only, and Both of them are NOT ZERO.

- SECP256K1_EC_COMPRESSED equals 0x102. (NOT ZERO)
- SECP256K1_EC_UNCOMPRESSED equals 0x2. (NOT ZERO)

https://github.com/bitcoin/bitcoin/blob/master/src/key.cpp#L149
```cpp
ret = ec_privkey_export_der(secp256k1_context_sign,
  (unsigned char*)&privkey[0], &privkeylen, begin(),
  fCompressed ? SECP256K1_EC_COMPRESSED : SECP256K1_EC_UNCOMPRESSED);
```

https://github.com/bitcoin/bitcoin/blob/master/src/secp256k1/include/secp256k1.h#L147
```cpp
#define SECP256K1_FLAGS_TYPE_COMPRESSION (1 << 1)
#define SECP256K1_FLAGS_BIT_COMPRESSION (1 << 8)
#define SECP256K1_EC_COMPRESSED (SECP256K1_FLAGS_TYPE_COMPRESSION | SECP256K1_FLAGS_BIT_COMPRESSION)
#define SECP256K1_EC_UNCOMPRESSED (SECP256K1_FLAGS_TYPE_COMPRESSION)
```

## My idea: Ignoring fCompressed variable.
I think ```fCompressed``` makes no sense when calling ec_privkey_export_der(). Though, to keep compatibility, ec_privkey_export_der() should be called as compressed mode only as in the past, I think.

[...]

PR #10043, opened 2017-03-21
Fix bug in CKey DER encoding

jonasschnelli:
At the moment, due to a mistake in our code, we DER export uncompressed keys as compressed.
Addresses #10041.

### DIFF
--- a/src/key.cpp
+++ b/src/key.cpp
@@ -147,7 +147,7 @@ CPrivKey CKey::GetPrivKey() const {
     size_t privkeylen;
     privkey.resize(279);
     privkeylen = 279;
-    ret = ec_privkey_export_der(secp256k1_context_sign, (unsigned char*)&privkey[0], &privkeylen, begin(), fCompressed ? SECP256K1_EC_COMPRESSED : SECP256K1_EC_UNCOMPRESSED);
+    ret = ec_privkey_export_der(secp256k1_context_sign, (unsigned char*)&privkey[0], &privkeylen, begin(), fCompressed);
     assert(ret);
     privkey.resize(privkeylen);
     return privkey;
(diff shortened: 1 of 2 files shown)


src/key.cpp
```diff
@@ -135,6 +135,7 @@ bool CKey::SetPrivKey(const CPrivKey &privkey, bool fCompressedIn) {
     if (!ec_privkey_import_der(secp256k1_context_sign, (unsigned char*)begin(), &privkey[0], privkey.size()))
         return false;
     fCompressed = fCompressedIn;
+    assert(privkey[1] == (fCompressed ? 0x81 : 0x82));
```
jonasschnelli: Not sure if this is correct and if we should abort if this the assumption is not true.

NicolasDorier:
I am quite surprised we even export private key in DER, I never had a need for it. Why is it used ?

jonasschnelli:
AFAIK the only place where we use DER private keys is when storing/retrieving from the wallet.dat BDB database. We should probably change that, but reading DER must be supported at least.

laanwj:
Thanks for the fix!
How does this affect users of the wallet or RPC? Does this need backports?

jonasschnelli:
> How does this affect users of the wallet or RPC? Does this need backports?

I'm not entirely sure. I haven't analysed the full ramification.
It looks like, that we only use the DER encoding during serialisation to and deserialisation from the database. Internally it should be treated correctly. Walletdump, etc. should be correct.

At the moment, I'm not sure what happens if you import an uncompressed WIF priv key. Or what happened when we introduced the bug with already existing uncompressed DER encoded keys in the wallet.

src/key.cpp
paveljanik: Looks like we do not call this function at all anyway...

gmaxwell:
(as was discussed in the last meeting)

This has no effect, we never read it. It's just idiocy in the format that a public key is saved inside the private key at all. I think it should just always set compressed here.  We use the public key outside of the private key for everything interesting.

laanwj:
OK so this is not really a bug, just a peculiarity of our data format.
I think then that we should close this and keep things as-is (and later on, switch to a better storage format)

jonasschnelli:
> OK so this is not really a bug, just a peculiarity of our data format.
I think then that we should close this and keep things as-is (and later on, switch to a better storage format)

Yes. Agree. We could optimise this out maybe together with the new, non backward compatible HD internal/external database format.

jonasschnelli:
Closing

PR #9508, opened 2017-01-10
Remove unused Python imports

practicalswift:
The commit in this PR removes unused Python imports.

PR #10047, opened 2017-03-22
[tests] Remove unused variables and imports

practicalswift:


fanquake:
I'm surprised there are some many to remove again, given this was only done two months ago (#9508).

jonasschnelli:
utACK d3073c232a7a0ad182cf4d9eb1495c79795669e9

jonasschnelli:
Maybe add commit from #10049 here? Trivial PRs tend to distract and therefore combining should be okay and desirable.

jnewbery:
Tested ACK d3073c232a7a0ad182cf4d9eb1495c79795669e9

@fanquake - there's been a lot of churn in the test directory recently, which is how some of these have crept in.

Longer term, I expect we'll continue to see unused imports until we have a sensible import scheme for the functional tests such as https://github.com/bitcoin/bitcoin/pull/9876#issuecomment-283198485. Static analysis tools can't work out wildcard imports and linters like flake8 throw up many errors when wildcards are used, so I expect no-one is using linters before committing.

practicalswift:
@jonasschnelli Good point! Added the commit from #10049 into this PR :-)

fanquake:
@jnewbery fair enough.

utACK 3897459

paveljanik:
utACK 3897459

MarcoFalke:
utACK 3897459

Issue #9923, opened 2017-03-05
Whitelisting outgoing connections


laanwj:
I think it should be possible to optionally whitelist outgoing connections - specifically those that are explicitly created through connect, addnode.
Context: I regularly need to bypass this code in testing setups to be able to sync non-internet connected nodes: https://github.com/bitcoin/bitcoin/blob/master/src/net_processing.cpp#L1720

Also in the interest of symmetry I don't see why whitelisting would only apply to incoming connections.

jnewbery:
As it happens, I have a branch that allows nodes to be whitelisted through an RPC: https://github.com/jnewbery/bitcoin/tree/updatenode .

Context: the RPC is called updatenode so that it can be extended in future, eg to update the banscore for a node, or disconnect a node based on node id (at the moment, nodes can only be disconnected based on ip address).

I haven't written any testcases for this yet, but feel free to take it if it's helpful for your testing.

laanwj:
That seems useful in general.

I'm a little afraid that won't work for my use-case because of the race between toggling 'whitelist' to 1 and making the connection. The other side will immediately send a `getheaders`, which could get ignored if it's not in time (and IIRC it won't send any, ever again, after that).

laanwj:
Another option here would be to make the `-whitelist` netmask apply to outgoing connections as well. It seems straightforward but that may be an unexpected change of behavior.

Sjors:
Duplicate of #4565

PR #10051, opened 2017-03-22
adhere to `-whitelist` for outbound connection

rubensayshi:
unsure why this wasn't already being done, seems sensible to me that you also want to be able to whitelist nodes that you make an outgoing connection to?

trying to figure if there should be any tests added but I can't seem to find tests that really cover ConnectNode specifically at all?

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -396,6 +396,7 @@ CNode* CConnman::ConnectNode(CAddress addrConnect, const char *pszDest, bool fCo
         CNode* pnode = new CNode(id, nLocalServices, GetBestHeight(), hSocket, addrConnect, CalculateKeyedNetGroup(addrConnect), nonce, pszDest ? pszDest : "", false);
         pnode->nServicesExpected = ServiceFlags(addrConnect.nServices & nRelevantServices);
         pnode->AddRef();
+        pnode->fWhitelisted = CConnman::IsWhitelistedRange((CNetAddr)pnode->addr);
 
         return pnode;
     } else if (!proxyConnectionFailed) {


gmaxwell:
Outbound connections are already privileged in a number of ways, and the white-listing by default has strange effects on transaction propagation that you probably don't want for outbound connections.

rubensayshi:
the reason for wanting to whitelist the outbound connection is cuz I'm using `connect` to connect only to 1 specific node.  
without whitelisting the trickling mechanism can cause transactions to take several seconds before being relayed.

laanwj:
I agree this would be useful. 
There should be *some* mechanism for whitelisting outgoing connections.
Closing because this is a duplicate of #9923.

laanwj:
Oops, missed that this was a PR, not an issue.

theuni:
@laanwj Agreed on all points. I've hacked this in for myself a few times for addnode=/connect= nodes. But as you said, tacking it on to whitelist is moving in the wrong direction.

I'd much prefer to get rid of whitelist and replace it with explicit privileges, generated at connection time based on inbound/outbound, config options, etc. That way there's less guessing in the later code, and there's a single place for the logic.

I can whip that up if you agree.

gmaxwell:
> without whitelisting the trickling mechanism can cause transactions to take several seconds before being relayed.

Whitelisting isn't really the correct fix for single outbound connect=1 resulting in slower relay.

laanwj:
> I'd much prefer to get rid of whitelist and replace it with explicit privileges, generated at connection time based on inbound/outbound, config options, etc. That way there's less guessing in the later code, and there's a single place for the logic.

I'd prefer that too. Something like, say, a bit field. The current whitelist is both too fine-grained and too course-grained. It's just vague what it does and different people want different things from it, and we can't change it without breaking other people's use cases.

rubensayshi:
hmm, for me the whitelist makes sense cuz I can easily ensure it covers my local servers without too much ufss

laanwj:
There seems to be no agreement to do this, and quite a lot of time elapsed without new discussion, so I'm closing. Looking forward to alternative solutions as were discussed in this PR.

PR #10053, opened 2017-03-22
[test] Allow functional test cases to be skipped

jnewbery:
Currently, functional test cases can either pass or fail. There are
occasions when it is helpful to skip tests, for example if the
system they are running on does not meet the requirements for the test.
The rest of the test suite can run without being marked as a failure.

This PR adds framework for tests to skip if their requirements
aren't met.

This PR also updates rpcbind_test.py to skip if it detects that the system
does not meet requirements (linux, ipv4 non-loopback address, ipv6 support).

This is required to have travis nightly cron jobs run without failure #10052.

### DIFF
--- a/test/functional/test_runner.py
+++ b/test/functional/test_runner.py
@@ -248,7 +248,7 @@ def run_tests(test_list, src_dir, build_dir, exeext, jobs=1, enable_coverage=Fal
     job_queue = TestHandler(jobs, tests_dir, test_list, flags)
 
     max_len_name = len(max(test_list, key=len))
-    results = BOLD[1] + "%s | %s | %s\n\n" % ("TEST".ljust(max_len_name), "PASSED ", "DURATION") + BOLD[0]
+    results = BOLD[1] + "%s | %s | %s\n\n" % ("TEST".ljust(max_len_name), "STATUS ", "DURATION") + BOLD[0]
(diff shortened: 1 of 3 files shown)


MarcoFalke:
utACK 0a6781dbc728abdd294fc199a6522caf592cbe56

fanquake:
Tested 0a6781d on OS X.
Current behaviour:
screen shot 2
This PR:
screen shot 1

test/functional/test_runner.py
```diff
@@ -245,20 +248,20 @@ def run_tests(test_list, src_dir, build_dir, exeext, jobs=1, enable_coverage=Fal
     job_queue = TestHandler(jobs, tests_dir, test_list, flags)
 
     max_len_name = len(max(test_list, key=len))
-    results = BOLD[1] + "%s | %s | %s\n\n" % ("TEST".ljust(max_len_name), "PASSED", "DURATION") + BOLD[0]
+    results = BOLD[1] + "%s | %s | %s\n\n" % ("TEST".ljust(max_len_name), "PASSED ", "DURATION") + BOLD[0]
```
```vim
s/PASSED /STATUS /

jnewbery: Agree. Done.

MarcoFalke:
Please squash the fixup.

PR #10056, opened 2017-03-22
[zmq] Call va_end() on va_start()ed args.

kallewoof:
> A function that invokes `va_start`, shall also invoke `va_end` before it returns.
([http://www.cplusplus.com/reference/cstdarg/va_start/](http://www.cplusplus.com/reference/cstdarg/va_start/))

> If there is no corresponding call to `va_start` or `va_copy`, or if` va_end` is not called before a function that calls `va_start` or `va_copy` returns, the behavior is undefined.
[http://en.cppreference.com/w/cpp/utility/variadic/va_end](http://en.cppreference.com/w/cpp/utility/variadic/va_end)

### DIFF
--- a/src/zmq/zmqpublishnotifier.cpp
+++ b/src/zmq/zmqpublishnotifier.cpp
@@ -30,6 +30,7 @@ static int zmq_send_multipart(void *sock, const void* data, size_t size, ...)
         if (rc != 0)
         {
             zmqError("Unable to initialize ZMQ msg");
+            va_end(args);
             return -1;
         }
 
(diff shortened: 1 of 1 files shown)


gmaxwell:
ugh. see this is why we need to test of sparc. :P

laanwj:
utACK, haven't checked though whether this covers all cases.

paveljanik:
utACK https://github.com/bitcoin/bitcoin/pull/10056/commits/5ba61f0034597ffeedc8c44af5a6ec670b095d83

@laanwj It covers all returns there.

jonasschnelli:
utACK 5ba61f0034597ffeedc8c44af5a6ec670b095d83.
I've check all the usages of `va_list` (`va_start`, `va_copy`) and expect of the ones covered in this PR we are good.

laanwj:
> ugh. see this is why we need to test of sparc. :P

That reminds me, ideally we'd use typesafe c++11 variadic templates instead of C varags legacy mess.

PR #10057, opened 2017-03-22
[init] Deduplicated sigaction() boilerplate

tjps:


paveljanik:
utACK https://github.com/bitcoin/bitcoin/pull/10057/commits/f21bddeaad8b88a30ed02e99ac76fde816c5954f

Thanks for extending comment about the usage of atomic variables.

laanwj:
Makes sense. utACK https://github.com/bitcoin/bitcoin/pull/10057/commits/f21bddeaad8b88a30ed02e99ac76fde816c5954f

Please remove "trivial" from the commit message and PR title though. Trivial means that there are effectively no code changes, only comment or variable name changes. This is explained in [CONTRIBUTING.md](https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#contributor-workflow).

tjps:
Updated the commit message.  Just to be clear, after having read CONTRIBUTING.md, should I include the area [prefix] in just the PR title, or the commit message as well?

sipa:
utACK

laanwj:
> Updated the commit message. Just to be clear, after having read CONTRIBUTING.md, should I include the area [prefix] in just the PR title, or the commit message as well?

Yes, the doc is unclear about that but in principle formatting for the PR title is the same as the commit message. The only difference is that the PR title summarizes a group of commits instead of one commit. If there is only one commit then they can (and should usually) be the same.

PR #10058, opened 2017-03-22
No need to use OpenSSL malloc/free

tjps:
The set of objects used in the OpenSSL locking callback do not need to be allocated using the OpenSSL allocator.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -117,26 +117,24 @@ std::atomic<bool> fReopenDebugLog(false);
 CTranslationInterface translationInterface;
 
 /** Init OpenSSL library multithreading support */
-static CCriticalSection** ppmutexOpenSSL;
+static std::unique_ptr<CCriticalSection[]> ppmutexOpenSSL;
 void locking_callback(int mode, int i, const char* file, int line) NO_THREAD_SAFETY_ANALYSIS
 {
     if (mode & CRYPTO_LOCK) {
-        ENTER_CRITICAL_SECTION(*ppmutexOpenSSL[i]);
+        ENTER_CRITICAL_SECTION(ppmutexOpenSSL[i]);
     } else {
-        LEAVE_CRITICAL_SECTION(*ppmutexOpenSSL[i]);
+        LEAVE_CRITICAL_SECTION(ppmutexOpenSSL[i]);
     }
 }
 
-// Init
+// Singleton for wrapping OpenSSL setup/teardown.
 class CInit
 {
 public:
     CInit()
     {
         // Init OpenSSL library multithreading support
-        ppmutexOpenSSL = (CCriticalSection**)OPENSSL_malloc(CRYPTO_num_locks() * sizeof(CCriticalSection*));
-        for (int i = 0; i < CRYPTO_num_locks(); i++)
-            ppmutexOpenSSL[i] = new CCriticalSection();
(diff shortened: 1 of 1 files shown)


fanquake:
Please be aware the we use the [trivial] label for changes that are essentially non-code changes. i.e fixing typos. This PR would not be considered trivial.

tjps:
Ack. will remove the trivial marker when I update the PR for the apparently broken test.

src/util.cpp
```diff
@@ -160,9 +159,8 @@ class CInit
         RAND_cleanup();
         // Shutdown OpenSSL library multithreading support
         CRYPTO_set_locking_callback(NULL);
-        for (int i = 0; i < CRYPTO_num_locks(); i++)
-            delete ppmutexOpenSSL[i];
-        OPENSSL_free(ppmutexOpenSSL);
```
jonasschnelli: AFAIK it was slightly wrong to free the ppmutexOpenSSL via `OPENSSL_free`. IMO only objects alloced by `OPENSSL_malloc()` should use `OPENSSL_free()`. This change makes sense.

jonasschnelli:
utACK 0765b6c9f08c5bac848ba8ac7c03471513fa4654

src/util.cpp
laanwj: It is allocated via OPENSSL_malloc
```
ppmutexOpenSSL = (CCriticalSection**)OPENSSL_malloc(CRYPTO_num_locks() * sizeof(CCriticalSection*));
```

jonasschnelli: > It is allocated via OPENSSL_malloc

Oh. Right.. it was. Nervermind then.

tjps:
Bump, for one less OpenSSL eyesore

src/util.cpp
```diff
@@ -117,7 +117,7 @@ std::atomic<bool> fReopenDebugLog(false);
 CTranslationInterface translationInterface;
 
 /** Init OpenSSL library multithreading support */
-static CCriticalSection** ppmutexOpenSSL;
+static std::vector<std::unique_ptr<CCriticalSection>> ppmutexOpenSSL;
```
JeremyRubin: Probably cleaner to just use a `std::unique_ptr<CCriticalSection[]>`

src/util.cpp
```diff
@@ -127,16 +127,15 @@ void locking_callback(int mode, int i, const char* file, int line) NO_THREAD_SAF
...
 class CInit
 {
 public:
     CInit()
     {
         // Init OpenSSL library multithreading support
-        ppmutexOpenSSL = (CCriticalSection**)OPENSSL_malloc(CRYPTO_num_locks() * sizeof(CCriticalSection*));
         for (int i = 0; i < CRYPTO_num_locks(); i++)
-            ppmutexOpenSSL[i] = new CCriticalSection();
+            ppmutexOpenSSL.emplace_back(new CCriticalSection());
```
JeremyRubin: and then here `ppmutexOpenSSL.reset(new CCriticalSection[CRYPTO_num_locks()]);` I think is correct.

tjps: You are correct.  I always like shaving lines off.

sipa:
utACK 6d5dd60c881b790f1b5fc5ba470605b6bd7e2fe7

PR #8801, opened 2016-09-23
[trivial] Switching from Boost for-each macros to C++11 for-each

tjps:
Switch all usages of the BOOST_FOREACH/BOOST_REVERSE_FOREACH/Q_FOREACH macros to the native C++11 for-each.  In the few usages of the BOOST_REVERSE_FOREACH macro I still use a Boost reversing adaptor to achieve the same effect.

PR #10059, opened 2017-03-23
[trivial] Removed one Boost usage and headers in util.cpp

tjps:


fanquake:
If your going to start removing specific Boost functions, I'd suggest looking over the codebase and seeing if it makes sense to remove all usages in one go, to save lots of similar pull requests. It also make review easier.

In this case, we only have two usages of ```ends_with```, in core_read.cpp and netbase.cpp.

There are 5 usages of ```starts_with``` (the 16 in leveldb/ don't count) in bitcoind.cpp, core_read.cpp, netbase.cpp, wallet/rpcdump.cpp and util.ccp as you've removed here.

I'd suggest updating this PR to remove all usages of both.

tjps:
That's a good idea, I was originally just offloading unrelated changes in smaller commits for a larger commit I'm working on.  Will update this with those other usages.

In the past I've been told that sweeping refactors (such as my for-each PR: https://github.com/bitcoin/bitcoin/pull/8801) are generally discouraged and to update things as they are touched.  Obviously this isn't as large of a set of changes, so I'm still working on finding the right balance.

Thanks for the pointer.

fanquake:
@tjps Yes large sweeping changes that touch many files, and disrupt many other pull requests at the same time are discouraged. As you say, it's finding the right balance between hundreds of trivial refactors, and trying to change 100 files at a time.

src/util.cpp
```diff
@@ -364,7 +362,7 @@ void ParseParameters(int argc, const char* const argv[])
         }
 #ifdef WIN32
         boost::to_lower(str);
-        if (boost::algorithm::starts_with(str, "/"))
+        if (str.size() > 0 && str[0] == '/')
```
laanwj: The algorithm predicate way of writing things is much more modern C++.
If we'd do this, we should just provide our own implementation of `boost::algorithm::starts_with`.
But I don't think replacing that is urgent.

tjps: I agree that not replacing it is urgent, this originally stemmed from a one-off from another patch I was working on.  Due to @fanquake's suggestion I looked at replacing all usages, and found cases where there was more than a single-character prefix/postfix match happening.

I'm dropping this PR altogether, not worth the change.

PR #10060, opened 2017-03-23
[Qt] Ensure an item exists on the rpcconsole stack before adding

achow101:
Ensures that there is an item on the rpcconsole stack before adding something to the current stack so that a segmentation fault does not occur.

Currently Bitcoin Core will just crash if there happens to be a newline somewhere in the command followed by some non-whitespace characters which are followed by a space. This causes a segfault by attempting to get the current stack from the empty stack vector. This fix ensures that that current stack always exists so that the segfault does not happen.

The crashing behavior can be tested by using this command in the rpcconsole:
```
decoderawtransaction 01000000 03e7a6f3 a93af3ae 49518cbf 8600b799 b04a8b8a ae5145f0 b25df06e 
  175d472a a
```
(copy and paste it as one line)

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -175,6 +175,10 @@ bool RPCConsole::RPCParseCommandLine(std::string &strResult, const std::string &
             nDepthInsideSensitive = 1;
             filter_begin_pos = chpos;
         }
+        // Make sure stack is not empty before adding something
+        if (stack.empty()) {
+            stack.push_back(std::vector<std::string>());
+        }
         stack.back().push_back(strArg);
     };
 


fanquake:
Confirmed that Core does segfault when using the above command in the rpcconsole, and that it no longer happens with the PR'd fix.

src/qt/rpcconsole.cpp
```diff
@@ -175,6 +175,11 @@ bool RPCConsole::RPCParseCommandLine(std::string &strResult, const std::string &
             nDepthInsideSensitive = 1;
             filter_begin_pos = chpos;
         }
+        // Make sure stack is not empty before adding something
+        if(stack.empty())
```
paveljanik: space after `if` please and left brace on the same line (or maybe one line `if` without braces - see https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md).

paveljanik:
Concept ACK

jonasschnelli:
Thanks for pointing this out.
Hmm... is there a reason to allow newlines in the command-line?
I don't think so.
Maybe this fix is more robust:
```
diff --git a/src/qt/rpcconsole.cpp b/src/qt/rpcconsole.cpp
index 7a0d0b3..c2f6434 100644
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -190,6 +190,7 @@ bool RPCConsole::RPCParseCommandLine(std::string &strResult, const std::string &
     };
 
     std::string strCommandTerminated = strCommand;
+    std::replace(strCommandTerminated.begin(),strCommandTerminated.end(),'\n',' ');
     if (strCommandTerminated.back() != '\n')
         strCommandTerminated += "\n";
     for (chpos = 0; chpos < strCommandTerminated.size(); ++chpos)
```

achow101:
@jonasschnelli The segfault bug happens with more than just a newline in the command. It also happens if an end parenthesis `)` is also in the command without a start parenthesis `(` before it.

src/qt/rpcconsole.cpp
```diff
@@ -175,6 +175,10 @@ bool RPCConsole::RPCParseCommandLine(std::string &strResult, const std::string &
             nDepthInsideSensitive = 1;
             filter_begin_pos = chpos;
         }
+        // Make sure stack is not empty before adding something
+        if(stack.empty()) {
```
sipa: Nit: space after `if` please.

sipa:
utACK, with nit.

jonasschnelli:
Thanks @achow101 for clarification.
Tested ACK 4df76e270caa9d828179cae1c7a8918d6f91ec21

jonasschnelli:
Needs backport.

PR #10068, opened 2017-03-24
[Wallet] FundRawTransaction accepts preset non-wallet inputs

NicolasDorier:
This PR makes it possible to call `FundRawTransaction` with pre filled inputs not belonging to the wallet.
This is very useful for AnyOneCanPay scenario, where one of a third party only cover part of a transaction.

The necessary information to complete the transaction is taken from of the mempool and coinview.

A follow up PR might allow the client of this API to pass previous `TxOuts` corresponding to the inputs.  This would be useful in scenario involving 2nd layer protocols, where the the inputs's of the transaction have not broadcasted yet.

A typical example would involves Alice and Bob wanting to fund 0.5 each to the payment channel. Alice would give her input, and Bob would be able to complete the missing amount via `FundRawTransaction`.

I personally need this PR for an atomic swap of colored coins. The user wants to exchange Bitcoin against some of his colored coins, to do so, he put his colored coin input into the transaction + the bitcoin outputs, and pass it to our service. Our service would then check that the rate is correct, and cover the bitcoin outputs with his own inputs by using a call to `FundRawTransaction`.

### DIFF
--- a/src/bench/coin_selection.cpp
+++ b/src/bench/coin_selection.cpp
@@ -48,7 +48,7 @@ static void CoinSelection(benchmark::State& state)
             addCoin(1000 * COIN, wallet, vCoins);
         addCoin(3 * COIN, wallet, vCoins);
 
-        std::set<std::pair<const CWalletTx*, unsigned int> > setCoinsRet;
+        std::set<CInputCoin> setCoinsRet;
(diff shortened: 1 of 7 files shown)


NicolasDorier:
rebased

PR #10069, opened 2017-03-24
[QA] Fix typo in fundrawtransaction test

NicolasDorier:
Ping @jnewbery introduced on https://github.com/bitcoin/bitcoin/commit/dab804c18a427901684ebe936b2069a97e04a268

### DIFF
--- a/test/functional/fundrawtransaction.py
+++ b/test/functional/fundrawtransaction.py
@@ -472,7 +472,7 @@ class RawTransactionsTest(BitcoinTestFramework):
         rawTx = self.nodes[1].createrawtransaction(inputs, outputs)
         # fund a transaction that requires a new key for the change output
         # creating the key must be impossible because the wallet is locked
-        assert_raises_jsonrpc(-4, "Insufficient funds", self.nodes[1].fundrawtransaction, rawtx)
+        assert_raises_jsonrpc(-4, "Keypool ran out, please call keypoolrefill first", self.nodes[1].fundrawtransaction, rawTx)
 
         #refill the keypool
         self.nodes[1].walletpassphrase("test", 100)


laanwj:
I don't get it, how did this typo pass the tests?

NicolasDorier:
@laanwj becuse `rawtx` is an actual variable used above, which is planned to fail with this reason.

test/functional/fundrawtransaction.py
```diff
@@ -472,7 +472,7 @@ def run_test(self):
         rawTx = self.nodes[1].createrawtransaction(inputs, outputs)
         # fund a transaction that requires a new key for the change output
         # creating the key must be impossible because the wallet is locked
-        assert_raises_jsonrpc(-4, "Insufficient funds", self.nodes[1].fundrawtransaction, rawtx)
```
dcousens: `rawtx` is also defined, see above?

laanwj:
Oh wow. Two variables with the same name and different casing, this is confusing.

jnewbery:
ouch. Mea culpa. Good spot @NicolasDorier.

Tested ACK 803e6a35037e7371f767585fec36ba17bdd36f4a

laanwj:
I think the underlying problem should be solved too. We shouldn't use `rawTx` and `rawtx` in a single function, this is the code equivalent of a minefield.

MarcoFalke:
utACK 803e6a3. Agree that the underlying problem should be solved as well. As a first step you can get rid of the camelCase and use a single name for the variable. In a second step, the scopes could be limited to the minimum necessary. (Maybe refactor the blocks into separate functions?)

jonasschnelli:
utACK 803e6a35037e7371f767585fec36ba17bdd36f4a

sipa:
utACK

dcousens (approved):

MarcoFalke:
This has too much review to keep open any longer. Lets do the follow up cleanup in separate pulls.

NicolasDorier:
Follow up https://github.com/bitcoin/bitcoin/pull/10083

PR #10072, opened 2017-03-24
Remove sources of unreliablility in extended functional tests

jnewbery:
This PR removes two sources of unreliability that were causing extended test cases to fail intermittently on travis:

- bip9-softforks stop-starts bitcoind twice, but does not wait for the p2p connection to mininode to re-open. This would lead to race conditions where the following call to getblocktemplate() would sometimes fail because there were no p2p connections open to the node. This commit also removes the shutil.rmtree() call that was blatting the test_framework.log file.
- forknotify asserts that an alert has been written to a file and fails immediately if the file is empty. This would lead to race conditions where the assert would sometimes be hit before bitcoind had written to the file.

### DIFF
--- a/test/functional/bip9-softforks.py
+++ b/test/functional/bip9-softforks.py
@@ -200,16 +200,14 @@ class BIP9SoftForksTest(ComparisonTestFramework):
         yield TestInstance([[block, False]])
 
         # Restart all
-        self.test.block_store.close()
+        self.test.clear_all_connections()
         stop_nodes(self.nodes)
-        shutil.rmtree(self.options.tmpdir)
+        shutil.rmtree(self.options.tmpdir + "/node0")
         self.setup_chain()
         self.setup_network()
-        self.test.block_store = BlockStore(self.options.tmpdir)
-        self.test.clear_all_connections()
         self.test.add_all_connections(self.nodes)
-        NetworkThread().start() # Start up network handling in another thread
-
+        NetworkThread().start()
+        self.test.test_nodes[0].wait_for_verack()
 
     def get_tests(self):
         for test in itertools.chain(
--- a/test/functional/forknotify.py
+++ b/test/functional/forknotify.py
@@ -3,6 +3,8 @@
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 """Test the -alertnotify option."""
+import os
+import time
 
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import *
@@ -41,11 +43,18 @@ class ForkNotifyTest(BitcoinTestFramework):
         self.nodes[1].generate(1)
         self.sync_all()
 
+        # Give bitcoind 10 seconds to write the alert notification
+        timeout = 10.0
+        while timeout > 0:
+            if os.path.exists(self.alert_filename) and os.path.getsize(self.alert_filename):
+                break
+            time.sleep(0.1)
+            timeout -= 0.1
+        else:
+            assert False, "-alertnotify did not warn of up-version blocks"
+
         with open(self.alert_filename, 'r', encoding='utf8') as f:
             alert_text = f.read()
-
-        if len(alert_text) == 0:
-            raise AssertionError("-alertnotify did not warn of up-version blocks")
 
         # Mine more up-version blocks, should not get more alerts:
         self.nodes[1].generate(1)


test/functional/forknotify.py
```diff
@@ -41,11 +42,17 @@ def run_test(self):
         self.nodes[1].generate(1)
         self.sync_all()
 
-        with open(self.alert_filename, 'r', encoding='utf8') as f:
-            alert_text = f.read()
+        # Give bitcoind 10 seconds to write the alert notification
+        timeout = 10
+        while True:
```
JeremyRubin: Maybe a bit cleaner to do something like (a couple of changes, feel free to pick and choose -- I was torn on the os.path.exists+getsize v.s. continual reading with open).

```python
        for t in xrange(100):
            if os.path.exists(self.alert_filename) and os.path.getsize(self.alert_filename):
                break
            if t != 99: time.sleep(0.1) 
        else:
            raise AssertionError("-alertnotify did not warn of up-version blocks")
        with open(self.alert_filename, 'r', encoding='utf8') as f:
            alert_text = f.read()
```

test/functional/forknotify.py
```diff
@@ -41,11 +42,17 @@ def run_test(self):
         self.nodes[1].generate(1)
         self.sync_all()
 
-        with open(self.alert_filename, 'r', encoding='utf8') as f:
-            alert_text = f.read()
+        # Give bitcoind 10 seconds to write the alert notification
+        timeout = 10
```
JeremyRubin: This should rarely actually take 10 seconds, if the code is correct, right? Maybe 1 second will be a bit better in the case where it is actually broken & you're trying to fix it.

JeremyRubin (approved):
utack! A couple suggestions but nothing major.

test/functional/bip9-softforks.py
```diff
@@ -200,16 +200,14 @@ def test_BIP(self, bipName, activated_version, invalidate, invalidatePostSignatu
         yield TestInstance([[block, False]])
 
         # Restart all
-        self.test.block_store.close()
+        self.test.clear_all_connections()
         stop_nodes(self.nodes)
-        shutil.rmtree(self.options.tmpdir)
+        shutil.rmtree(self.options.tmpdir + "/node0")
```
jnewbery: Not that I know of.

I don't like the fact that this test is using shutil and removing files itself. This change just makes it slightly less damaging (ie it's only removing the datadir rather than the entire tmpdir)

test/functional/bip9-softforks.py
```diff
@@ -200,16 +200,14 @@ def test_BIP(self, bipName, activated_version, invalidate, invalidatePostSignatu
...
         self.setup_chain()
         self.setup_network()
-        self.test.block_store = BlockStore(self.options.tmpdir)
-        self.test.clear_all_connections()
         self.test.add_all_connections(self.nodes)
-        NetworkThread().start() # Start up network handling in another thread
-
+        NetworkThread().start()
```
jnewbery: No, we need to wait for the network thread to start before receiving the verack (the network thread is what catches the verack and calls the callback).

test/functional/bip9-softforks.py
```diff
@@ -200,16 +200,14 @@ def test_BIP(self, bipName, activated_version, invalidate, invalidatePostSignatu
...
+        self.test.test_nodes[0].wait_for_verack()
```
jnewbery: wait_for_verack() is called by the comparison test framework at the start of the test run (comptest.py line 300).

test/functional/forknotify.py
jnewbery: Depends. Travis catches a lot of timing edge cases. I don't think there's any problem making the timeout something high like 10 seconds since the while loop will break as soon as the file is written.

test/functional/forknotify.py
jnewbery: I think this is equivalent.

test/functional/forknotify.py
```diff
@@ -41,11 +42,17 @@ def run_test(self):
...
+        while True:
+            with open(self.alert_filename, 'r', encoding='utf8') as f:
+                alert_text = f.read()
+            if len(alert_text) > 0:
+                break
 
-        if len(alert_text) == 0:
-            raise AssertionError("-alertnotify did not warn of up-version blocks")
+            assert timeout > 0, "-alertnotify did not warn of up-version blocks"
```
jnewbery: I don't think there's any difference. No-one runs the test cases with `-O`

test/functional/forknotify.py
JeremyRubin: yeah; mostly this was motivated by being able to get rid of the stateful `timeout = 10; assert timeout > 0; timeout -= 0.1` combo.
This is also fine.
```python
        for t in xrange(100):
            with open(self.alert_filename, 'r', encoding='utf8') as f:
                alert_text = f.read()
            if alert_text:
                break
            time.sleep(0.1) 
        else:
            raise AssertionError("-alertnotify did not warn of up-version blocks")
```

test/functional/forknotify.py
JeremyRubin: using asserts as control flow is generally an antipattern (I think that the for/else loop is a bit clearer in any case).

jnewbery: ah. I see what you're saying now. Yes, I'll remove the assert from the control flow.

jnewbery:
Pushed a new version using while/else and not using assert for control flow.

MarcoFalke:
utACK a4fd89f. Going to merge this, so that the failures are a thing of the past.

test/functional/forknotify.py
```diff
@@ -41,12 +43,19 @@ def run_test(self):
...
 
+        # Give bitcoind 10 seconds to write the alert notification
+        timeout = 10.0
+        while timeout > 0:
+            if os.path.exists(self.alert_filename) and os.path.getsize(self.alert_filename):
+                break
+            time.sleep(0.1)
+            timeout -= 0.1
+        else:
+            assert False, "-alertnotify did not warn of up-version blocks"
```
MarcoFalke: @jnewbery nit: Any reason you changed the raise AssertionError to assert False?

jnewbery: No good reason. Equivalent behaviour but you're probably right that `raise AssertionError` is better style here.

jnewbery:
🎉 first successful daily run of extended tests on Travis: https://travis-ci.org/bitcoin/bitcoin/builds/217786785

Thanks for merging @MarcoFalke

PR #10074, opened 2017-03-25
Block size/weight fraud proofs

luke-jr:
**Status: Untested, could use further optimisation**

Implements [BIP 180](https://github.com/bitcoin/bips/blob/master/bip-0180.mediawiki) to a usable state. Only provides proofs, but also implements and uses verification internally.

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -21,6 +21,7 @@
 #include "policy/policy.h"
 #include "primitives/block.h"
 #include "primitives/transaction.h"
+#include "proofs/size.h"
 #include "random.h"
 #include "tinyformat.h"
 #include "txmempool.h"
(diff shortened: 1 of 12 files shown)


buuser:
NACK

This is an unnecessary addition to the codebase and consensus protocol.

jonasschnelli:
@buuser: please add some technical arguments or observations. Simply saying no is not appropriate here.

src/net_processing.cpp
```diff
@@ -2591,6 +2592,50 @@ bool static ProcessMessage(CNode* pfrom, const std::string& strCommand, CDataStr
...
+                if (pindex->IsValid(BLOCK_VALID_SCRIPTS)) {
+                    connman.PushMessage(pfrom, msgMaker.Make(NetMsgType::FRAUD, pindex->GetBlockHash(), VARINT(0)));
+                    break;
+                }
+                if (pindex->nStatus & BLOCK_FAILED_VALID) {
+                    if ((pindex->nStatus & BLOCK_VALID_MASK) >= BLOCK_VALID_TRANSACTIONS) {
+                        // These are only reached when block size/weight are known to be good
+                    } else {
+                        CBlock block;
+                        const bool ret = ReadBlockFromDisk(block, pindex, chainparams.GetConsensus());
```
NicolasDorier: probably need to test `(pIndex->nStatus & BLOCK_HAVE_DATA)` in case of pruned node.

sdaftuar:
I think BIP 180 is an interesting observation, but I don't think that it makes sense to merge an implementation of it into Bitcoin Core.

For BIP 180 support to be useful, Bitcoin Core would need to go out of its way to download and store invalid blocks to be able to generate proofs for light clients.  Currently, we make no effort to do this (in fact, from a very quick glance at the code in this PR, I don't see how this code can possibly work, because blocks whose invalidity is detected in `CheckBlock` or `ContextualCheckBlock` aren't even written to disk, and that's where the size checks are done).

While we could attempt to "fix" that narrow issue, I also don't think it's worth pursuing as a goal: BIP 180 only makes sense in the context of a non-cooperative forking chain (as a cooperative one would e.g. provide signaling so that SPV clients could choose the headers chain they wish to follow), and if we assume non-cooperation, then BIP 180 wouldn't make sense if it's easy to subvert -- which it is.  For instance, suppose the first block in the forking chain had too many sigops but passed the blocksize/weight checks.  Bitcoin Core would not download any blocks that build off such an invalid block hash, because we'd know the chain to be invalid from the headers, and then wouldn't be able to provide any BIP 180 proofs to light clients.

We could try to rearchitect Bitcoin Core's consensus logic to try to make such fraud proofs available, by downloading later invalid blocks on a known-to-be-invalid chain and storing them, but this is a big change and IMO not worth doing.  I think we have a number of obstacles to overcome in order to make this work (see eg #9530, I think the improvements discussed there will already be review-intensive) and adding everything we need to ensure that BIP 180 would always work in Bitcoin Core seems like too much of a burden.

Anyone who wanted to provide BIP 180-proofs to the network could do so as a separate service, and I think this is actually substantially easier to implement outside Bitcoin Core (since if you're providing these proofs as a separate service, you can solve the DoS issues more easily than we can do in Bitcoin Core, where the assumptions are different).

Concept NACK.

NicolasDorier:
I think that a better, less intrusive way to do it, is to make it works in a "gossip" manner.
Instead of having the client asking fraud proof for all the blocks, it would be up to the node to broadcast fraud messages unsollicited to other peers.

A way of doing that would be: If node detect a fraud which pass PoW, then store in RAM up to X proofs, then broadcast them unsollicited. (like addr messages)

This would make it more trivial and less intrusive to implement. And even pruned node would be able to do it.

jnewbery:
It'd be great to see BIP180 as a separate application or a fork of bitcoind, but concept NACK merging this to core for the reasons @sdaftuar sets out above. Adding this kind of functionality and the architecture changes required to make it work place too much of a maintenance burden on bitcoin core.

jnewbery:
Alternatively, you could offer this as a PR to https://github.com/BitcoinUnlimited/BitcoinUnlimited . They already have all the necessary functionality to propogate and store invalid blocks, but I hear they've been having stability issues of late.

gmaxwell:
Is there any willingness to support such a thing in any existing lite-clients?

I think suhas' concern could be addressed by constructing a proof at the time we reject the block and saving it--- but the matter is moot unless someone will use it.

As an aside, thanks for going and implementing. It's good to see a version of this that isn't just talk.

NicolasDorier:
pinging @voisine

luke-jr:
Closing due to lack of interest.

jtimon:
I am very interested in this, but I thought you said there was a flaw in the design on the mailing list.

PR #9987, opened 2017-03-13
Remove unused code

practicalswift:

PR #6149, opened 2015-05-16
Buffer log messages and explicitly open logs

ajweiss:
Prevents stomping on debug logs in datadirs that are locked by other
instances and lost parameter interaction messages that can get wiped by
ShrinkDebugFile().

The log is now opened explicitly and all emitted messages are buffered
until this open occurs.  The version message and log cut have also been
moved to the earliest possible sensible location to prevent messages from
appearing "before the cut."

Thanks @jgarzik for setting me on the right track.

PR #10075, opened 2017-03-25
Remove unused C++ code not covered by unit tests

practicalswift:


src/rpc/protocol.cpp
```diff
@@ -114,13 +114,3 @@ bool GetAuthCookie(std::string *cookie_out)
         *cookie_out = cookie;
     return true;
 }
-
-void DeleteAuthCookie()
```
laanwj: Huh, it's a mistake that this is unused; the auth cookie is supposed to be removed on shutdown.

src/wallet/wallet.cpp
```diff
@@ -3495,22 +3495,6 @@ bool CWallet::LoadDestData(const CTxDestination &dest, const std::string &key, c
     return true;
 }
 
-bool CWallet::GetDestData(const CTxDestination &dest, const std::string &key, std::string *value) const
```
laanwj: This is part of the `*DestData` API, please don't remove it just because it's not used at the moment. Same reasoning as here: https://github.com/bitcoin/bitcoin/pull/9987#discussion_r106381766

src/net.cpp
```diff
@@ -2616,7 +2598,6 @@ int CConnman::GetBestHeight() const
 }
 
 unsigned int CConnman::GetReceiveFloodSize() const { return nReceiveFloodSize; }
-unsigned int CConnman::GetSendBufferSize() const{ return nSendBufferMaxSize; }
```
laanwj: CConnman is recent code and part of an ongoing effort by @theuni, please discuss with him before removing things.

practicalswift:
Trivia: This PR combined with #9987 removes 0.134 % (173 of 128 675 lines of code) of the Bitcoin C++ code base :-)

```
$ git grep "" | egrep '\.(cpp|h):' | wc -l
128675
```

src/chainparamsbase.cpp
```diff
@@ -103,8 +103,3 @@ std::string ChainNameFromCommandLine()
         return CBaseChainParams::TESTNET;
     return CBaseChainParams::MAIN;
 }
-
-bool AreBaseParamsConfigured()
```
laanwj: ACK on this one, apparently the last use of this went away in #6149/27d760580456d206c5a02ef29797f296f510099c.

src/key.cpp
```diff
@@ -131,14 +131,6 @@ void CKey::MakeNewKey(bool fCompressedIn) {
     fCompressed = fCompressedIn;
 }
 
-bool CKey::SetPrivKey(const CPrivKey &privkey, bool fCompressedIn) {
```
laanwj: ACK on this one. This seems to have been unused since #5227/36fa4a78acac0ae6bb0e95c6ef78630120a28bdd.

src/netaddress.cpp
```diff
@@ -179,12 +179,6 @@ bool CNetAddr::IsLocal() const
    return false;
 }
 
-bool CNetAddr::IsMulticast() const
```
laanwj: ACK on this one. Cannot find when this was *ever* used, and I don't think we need this now or on the forseeable future.

src/wallet/wallet.cpp
practicalswift: @laanwj Thanks for reviewing. I see - so basically a test is needed here :-)

src/netaddress.cpp
```diff
@@ -570,11 +564,6 @@ std::string CService::ToString() const
     return ToStringIPPort();
 }
 
-void CService::SetPort(unsigned short portIn)
```
laanwj: Setting ports seems to be reasonably useful part of the API, unless all of CService is made immutable I'd suggest keeping this.

src/net.cpp
practicalswift: @laanwj I see. @theuni, would you mind adding a test for this one? :-)

src/wallet/db.cpp
```diff
@@ -463,15 +463,6 @@ void CDBEnv::CloseDb(const std::string& strFile)
     }
 }
 
-bool CDBEnv::RemoveDb(const std::string& strFile)
```
laanwj: Not 100% sure about this one. We don't use it, and haven't used it for a long time. From what I see it was introduced in eed1785f701be93ac2464e854c2a7de1f748ef84 (pull# unknown) and never used. Would deleting wallets be something we'd ever support? (e.g. as part of multiwallet). It is an extremely dangerous thing to do, on the other hand for watch-only wallets it'd be useful.

practicalswift: @laanwj I'd vote for removing it until someone finds a need for it. Motivation: such a dangerous operation should be tested :-)

src/rpc/protocol.cpp
practicalswift: @laanwj How should I handle `DeleteAuthCookie()` in this PR? I'd suggest keeping this PR as-is and handle the re-introduction of `StopRPC()`/`DeleteAuthCookie()` in a separate PR. Makes sense?

practicalswift:
@laanwj I've now addressed your review items. Please let me know if any further adjustments are needed :-)

src/net.cpp
theuni: @practicalswift Yes, thanks, that's a good idea.

src/rpc/server.cpp
```diff
@@ -322,6 +322,7 @@ void StopRPC()
 {
     LogPrint("rpc", "Stopping RPC\n");
     deadlineTimers.clear();
+    DeleteAuthCookie();
```
fanquake: I think this could/should be broken out into it's own PR. It's wasn't obvious right away why this was being added.

src/netaddress.cpp
fanquake: Looks like it was introduced in #735 (5 years ago) and yes possibly never used.

src/rpc/server.cpp
laanwj: At least it's a separate commit in this PR. But splitting it to a separate PR would make sense, yes.

practicalswift:
@fanquake @laanwj Good points! Re-introduction of auth cookie removal on shutdown has been moved to #10139 :-)

practicalswift:
Let me know if there are any suggestions on further tweaks needed for this PR :-)

fanquake:
This needs a rebase.

practicalswift:
@fanquake Thanks for the notification. Now rebased!

Let me know if anything needs to be changed in this PR.

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/10075/commits/b51aaf1c42d9936ddbe8ec48a0d9b675221ab7ba

PR #10076, opened 2017-03-25
[qa] combine_logs: Use ordered list for logfiles

maflcko:
According to the documentation of glob, "results are returned in arbitrary order". Thus, create each path separately by string formatting.

### DIFF
--- a/test/functional/combine_logs.py
+++ b/test/functional/combine_logs.py
@@ -6,8 +6,8 @@ to write to an outputfile."""
 
 import argparse
 from collections import defaultdict, namedtuple
-import glob
 import heapq
+import itertools
 import os
 import re
 import sys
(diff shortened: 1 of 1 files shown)


test/functional/combine_logs.py
```diff
@@ -6,11 +6,11 @@
 
 import argparse
 from collections import defaultdict, namedtuple
-import glob
 import heapq
 import os
 import re
 import sys
+from test_framework.util import MAX_NODES
```
jnewbery: micronit: blank line between the standard library imports and local project imports please.

jnewbery:
Good catch! Tested ACK https://github.com/bitcoin/bitcoin/pull/10076/commits/fa32e5c16b7f281c500ef6064e6d0e0329465c74 with a micronit.

jnewbery (approved):

jnewbery:
Actually, thinking about this some more, I don't think I like the dependency on test_framework.utils here. At the moment combine_logs can be run from anywhere, even on a system that doesn't have bitcoin or the test_framework installed. You can tar the tmp directory, upload it somewhere and then do the post-processing on another system. I'd like to preserve that if possible.

Can you change this to use glob and then inspect the name of the directory to get the node index?

jnewbery (changes requested):
Please remove dependency on test_framework

test/functional/combine_logs.py
MarcoFalke: @jnewbery Removed this import

jnewbery (approved):
Looks great. Thanks Marco.

Tested ACK fa4535df574067700ed7d2227ae03605075b8730

PR #10077, opened 2017-03-25
[qa] Add setnetworkactive smoke test

maflcko:
This adds coverage to the currently uncovered rpc setnetworkactive.

Steps to verify the new coverage:

```sh
./test/functional/test_runner.py --coverage net | grep setnetworkactive && echo Failed! || echo Success!
```


jnewbery:
Looks great. Tested ACK https://github.com/bitcoin/bitcoin/pull/10077/commits/fa697b719286c599103d04c7055b7c666e71f09d

Only comment would be that nodehandling.py could be merged into this so we have a single test script for testing the net.cpp RPCs. But also fine to have them as separate tests.

PR #10081, opened 2017-03-26
Add 'dumpbootstrap' RPC for creating bootstrap.dat files.

dooglus:
I don't know if there's any interest in merging this or not, but I find it useful. It will dump a `bootstrap.dat` file from a synced node that can be used to sync another node.

You can dump any part of the blockchain using the optional parameters. By default it will dump the whole blockchain.

    $ bitcoin-cli help dumpbootstrap
    dumpbootstrap <destination> [endblock] [startblock=0]

    Creates a bootstrap format block dump of the blockchain in destination, which can be a directory or a path with filename, up to the given endblock number.

    Arguments:
    1. destination  (string, required) Pathname of file to write to. If a directory is use, 'bootstrap.dat' is created in that directory.
    2. endblock     (numeric, optional, defaults to the last block in the active chain) Height of last block to dump.
    3. startblock   (numeric, optional, default=0) Height of first block to dump.

    Result:

    Examples:
    > bitcoin-cli dumpbootstrap "/tmp"
    > curl --user myusername --data-binary '{"jsonrpc": "1.0", "id":"curltest", "method": "dumpbootstrap", "params": ["/tmp"] }' -H 'content-type: text/plain;' http://127.0.0.1:8332/

    $ bitcoin-cli dumpbootstrap /tmp/ 10
    dumped 11 blocks from 0 to 10 into /tmp/bootstrap.dat

    $ bitcoin-cli dumpbootstrap /tmp 10 11
    error code: -1
    error message:
    Start block number out of range.

    $ bitcoin-cli dumpbootstrap /tmp/ 30 20
    dumped 11 blocks from 20 to 30 into /tmp/bootstrap.dat

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -7,6 +7,7 @@
 #include "chain.h"
 #include "chainparams.h"
 #include "checkpoints.h"
+#include "clientversion.h"
 #include "coins.h"
 #include "consensus/validation.h"
 #include "validation.h"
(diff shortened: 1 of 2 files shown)


sipa:
This does the same as linearize.py?

cdecker:
I like the simplicity of the implementation, but I wonder how many people will find it useful.

dooglus:
@sipa If linearize.py creates bootstrap.dat format files from a synced blockchain then I guess so. I hadn't heard of linearize.py.

Are you talking about `contrib/linearize/linearize-data.py`? That's the closest I can find in this repository.

I find it convenient to use `dumpbootstrap` to dump the portion of the blockchain I want into a linear file but understand if you don't want the functionality duplicated.

laanwj:
I think this could be useful in some cases.

However remembering the recent discussion on `dumpwallet` (#9937, #9934), I'd prefer not to have more RPCs that create or write server-side files. It introduces security risks, can load to dangerous situations (e.g. overwriting the wallet) and also interferes with future process sandboxing.

The ideal solution to this would be to stream the data over HTTP instead, which I attempted to introduce in #7759. Unfortunately I had no time to continue that.

And as said, this is already what the `linarize` tooling does. There is not an urgent need to have this functionality in bitcoind.

laanwj:
Let's close this for now. Tooling to generate `bootstrap.dat` is available, after all, and it doesn't seem like something that has a lot of users, so I'm not convinced it makes sense to merge this.

PR #10078, opened 2017-03-25
[qa] Fundraw: Use named args to limit scope of names

maflcko:
Thanks to @laanwj, it is now possible to use named args to limit the scope of names in the fundrawtransaction script. In a first step this is done for the variables `inputs`, `outputs`, which have previously been overwritten for each test.

Hint to get a more pleasing diff:

```git
git diff HEAD~ --word-diff-regex=[^[:space:]]

PR #10083, opened 2017-03-26
[QA] Renaming rawTx into rawtx

NicolasDorier:
Follow up from https://github.com/bitcoin/bitcoin/pull/10069

fundrawtransaction.py is using two variables: `rawtx` and `rawTx`, which can be easily confused  (see https://github.com/bitcoin/bitcoin/pull/10069). This PR rename everything in `rawtx`.

It was discussed to have separate function for each test case of `fundrawtransaction.py`, however, it seems a bit overkill to me, I think this PR is good enough. (Other followup https://github.com/bitcoin/bitcoin/pull/10078)

### DIFF
--- a/test/functional/fundrawtransaction.py
+++ b/test/functional/fundrawtransaction.py
@@ -322,8 +322,8 @@ class RawTransactionsTest(BitcoinTestFramework):
         #compare fee of a standard pubkeyhash transaction
         inputs = []
         outputs = {self.nodes[1].getnewaddress():1.1}
-        rawTx = self.nodes[0].createrawtransaction(inputs, outputs)
-        fundedTx = self.nodes[0].fundrawtransaction(rawTx)
+        rawtx = self.nodes[0].createrawtransaction(inputs, outputs)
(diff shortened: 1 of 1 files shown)


dcousens (approved):

jonasschnelli:
utACK dd5be2c7b35f0ecd1650a7f8cf08e692f6f41ccb

laanwj:
Yes, utACK, thanks, this should have been the solution in the first place.

droark:
Just curious, any reason why rawtx was chosen over rawTx? Personally, I prefer the latter. Making everything lowercase obfuscates the fact that multiple words/abbreviations are being used. There's also some inconsistency in the code's variable names at that point. Not a big deal, just something I'm curious about.

MarcoFalke:
@droark Feel free to use `raw_tx` if you think that it is clearer than `rawtx`. Though, the python code should stay free of camelCase in variable names of local scope.

droark:
Fair enough. Thanks for clarifying.

PR #10085, opened 2017-03-26
Docs: remove 'noconnect' option

jlopp:
As far as I can tell, "noconnect" is never actually parsed as an argument in the codebase.


luke-jr:
-no* is the same as -*=0

jlopp:
@luke-jr can you point out where the logic is that translates a "no*" parameter into a "*=0" parameter?

MarcoFalke:
https://github.com/bitcoin/bitcoin/blob/111849345bb5140f86b48e730ceab4bff45fa2e9/src/util.cpp#L339

jlopp:
Aha; thanks for the clarification.

sipa:
@jlopp If it wasn't clear that -noX is an alias for -X=0, perhaps that should be documented more clearly in the `-help` output?

MarcoFalke:
utACK d5690f1ab87678278d622ea82f3fa569314a2144. I think this makes it
easier to understand.

jlopp:
@sipa Yes, possibly - I'll think about it. I'm currently working on a project involving Bitcoin Core's configuration options, which is why I'm currently exploring this area of the codebase.

sipa:
@jlopp, regardless, I agree with @MarcoFalke that this PR probably improves things, despite not fixing anything technicall incorrect. I suggest reopening.

jlopp:
Sure, done.

laanwj:
`-noconnect` works fine and I use it a lot for internal tests. It's a little bit consistent to not mention it explicitly in the help I guess but please don't say that things don't work if they do!

PR #10090, opened 2017-03-26
Update bitcoin.conf with example for pruning

coinables:
This option is becoming more popular recently, and I propose an example to be shown in the bitcoin.conf.

### DIFF
--- a/contrib/debian/examples/bitcoin.conf
+++ b/contrib/debian/examples/bitcoin.conf
@@ -131,6 +131,13 @@
 # be validated sooner.
 #paytxfee=0.00
 
+# Enable pruning to reduce storage requirements by deleting old blocks. 
+# This mode is incompatible with -txindex and -rescan.
+# 0 = default (no pruning).
+# 1 = allows manual pruning via RPC.
+# >=550 = target to stay under in MiB. 
+#prune=550
+
 # User interface options
 
 # Start Bitcoin minimized


laanwj:
Concept ACK on adding the example.

Explicitly mentioning 3GB is not a good idea. These numbers peppered around the documentation tend to get outdated sooner than later. The [size of the UTXO set](http://statoshi.info/dashboard/db/unspent-transaction-output-set) is growing every day, and that can't be pruned.

contrib/debian/examples/bitcoin.conf
```diff
@@ -131,6 +131,10 @@
 # be validated sooner.
 #paytxfee=0.00
 
+# Prune your node to the last n blocks. 550 is the minimum. 
+# Pruning can drastically reduce the amount of disk space required to run bitcoin down to ~3GB.
```
laanwj: You have the unit wrong, it's not in blocks but in MiB. Maybe you can borrow some description text from the help message:
```
  -prune=<n>
       Reduce storage requirements by enabling pruning (deleting) of old
       blocks. This allows the pruneblockchain RPC to be called to
       delete specific blocks, and enables automatic pruning of old
       blocks if a target size in MiB is provided. This mode is
       incompatible with -txindex and -rescan. Warning: Reverting this
       setting requires re-downloading the entire blockchain. (default:
       0 = disable pruning blocks, 1 = allow manual pruning via RPC,
       >550 = automatically prune block files to stay under the
       specified target size in MiB)
```

contrib/debian/examples/bitcoin.conf
```diff
@@ -131,8 +131,10 @@
 # be validated sooner.
 #paytxfee=0.00
 
-# Prune your node to the last n blocks. 550 is the minimum. 
-# Pruning can drastically reduce the amount of disk space required to run bitcoin down to ~3GB.
+# Enable pruning to reduce storage requirments by deleting old blocks.
```
paveljanik: requirements

contrib/debian/examples/bitcoin.conf
```diff
@@ -131,8 +131,10 @@
 # be validated sooner.
 #paytxfee=0.00
 
-# Prune your node to the last n blocks. 550 is the minimum. 
-# Pruning can drastically reduce the amount of disk space required to run bitcoin down to ~3GB.
+# Enable pruning to reduce storage requirments by deleting old blocks. 
+# Value is in MiB. 550 is the minimum.
```
paveljanik: 550 is the minimum, but you describe entering 1...

contrib/debian/examples/bitcoin.conf
```diff
@@ -131,8 +131,10 @@
 # be validated sooner.
 #paytxfee=0.00
 
-# Prune your node to the last n blocks. 550 is the minimum. 
-# Pruning can drastically reduce the amount of disk space required to run bitcoin down to ~3GB.
+# Enable pruning to reduce storage requirments by deleting old blocks. 
+# Value is in MiB. 550 is the minimum. 
+# This mode is incompatible with -txindex and -rescan.
+# default = 0, 1 = allows manual pruning via RPC, >550 = target to stay under in MiB.
```
paveljanik: >=550

paveljanik (changes requested):

sipa:
ACK after squashing.

contrib/debian/examples/bitcoin.conf
```diff
@@ -131,6 +131,13 @@
 # be validated sooner.
 #paytxfee=0.00
 
+# Enable pruning to reduce storage requirments by deleting old blocks. 
+# This mode is incompatible with -txindex and -rescan.
+# 0 = default(no pruning).
```
paveljanik: space before "(". And please squash.

contrib/debian/examples/bitcoin.conf
```diff
@@ -131,6 +131,13 @@
 # be validated sooner.
 #paytxfee=0.00
 
+# Enable pruning to reduce storage requirments by deleting old blocks.
```
paveljanik: requirments -> requirements

paveljanik (changes requested):

fanquake:
utACK c0651cc

laanwj:
I think we should move this file somewhere else - `doc/examples` maybe? After all it is a general example of a `bitcoin.conf` and there is nothing debian-specific to it.

PR #10092, opened 2017-03-27
Update consensus.h

1Hyena:
Fix to the block size issue.


gmaxwell:
please, trolling like this is exhausting. Submitting broken code that doesn't even pass its own self-tests and would totally break the network, over and over again... is just a waste of everyone's time. It's not funny, people are working here.

JavierGonzalez:
Gavin please stop limiting the transactions per second to make SEGWIT necessary to open the door to stealing the commissions from the miners.

Miners need this incentive to protect the network from government attacks.

Stop fucking Bitcoin.

Thank you.

PR #10093, opened 2017-03-27
[Qt] Don't add arguments of sensitive command to console window

jonasschnelli:
At the moment, we hide sensitive command arguments from the console history but not from the console window.
This tiny change will also hide the arguments from the console window.

Especially for "importmulti", this may be a little bit annoying because if one executes a command with invalid arguments, you need to start type in everything again.

Ideally we would only filter the command if sensitive arguments have been used (complicated to implement).

List of sensitive commands:
```cpp
const QStringList historyFilter = QStringList()
    << "importprivkey"
    << "importmulti"
    << "signmessagewithprivkey"
    << "signrawtransaction"
    << "walletpassphrase"
    << "walletpassphrasechange"
    << "encryptwallet";

}
```

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -825,7 +825,7 @@ void RPCConsole::on_lineEdit_returnPressed()
 
         cmdBeforeBrowsing = QString();
 
-        message(CMD_REQUEST, cmd);
+        message(CMD_REQUEST, QString::fromStdString(strFilteredCmd));
         Q_EMIT cmdRequest(cmd);
 
         cmd = QString::fromStdString(strFilteredCmd);


luke-jr:
Why?

jonasschnelli:
> Why?

Even if we don't persist the console windows content, one may not want to expose private keys (and the wallet passphrase) on the screen. IMO "shell"-like interpreters should never reveal passwords "on screen" after they have been typed-in.

fanquake:
Quickly tested. Used the same command with both `importprivkey("some_key")`.
Using master (12af74b289f8cdc6caf850dc6c802f9936b1e8b3) command params are displayed on screen.
With this PR (7278537), they are not. 
Master:
master

This PR:

laanwj:
> Even if we don't persist the console windows content, one may not want to expose private keys (and the wallet passphrase) on the screen. IMO "shell"-like interpreters should never reveal passwords "on screen" after they have been typed-in.

Agree, it is somewhat useful to prevent shoulder-surfing.

ryanofsky:
ACK 72785377306764ab999e4bab13b821e4e5ea11eb.

Code change is trivial. Tested and confirmed it removes arguments onscreen in addition to ones in history (which were already removed before).

sipa:
Lightly-tested ACK.

PR #8824, opened 2016-09-27
Refactor TxToJSON() and ScriptPubKeyToJSON()

jnewbery:
This PR removes almost all of the `TxToJSON()` and `ScriptPubKeyToJSON()` code from bitcoin-rpc and places it in bitcoin-common, removing about 80 lines of duplicate code. The only part that can't be moved into bitcoin-common is parsing block contextual information (confirmations and blocktime), which are not available to bitcoin-common code.

This PR should be merged along with https://github.com/bitcoin/bitcoin/pull/8817 , which ensures that witness data gets returned from calls to TxToUniv()

PR #10095, opened 2017-03-27
refactor: Move GetDifficulty out of `rpc/server.h`

laanwj:
`GetDifficulty` has no business in `rpcserver.h`. Define it in the interface header of the implementation unit `rpcblockchain` where it is defined.

Also modernize the signature to:

    double GetDifficulty(const CBlockIndex* blockindex = nullptr);

(remove `extern`, replace `NULL` with `nullptr`)

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -122,6 +122,7 @@ BITCOIN_CORE_H = \
   protocol.h \
   random.h \
   reverselock.h \
+  rpc/blockchain.h \
   rpc/client.h \
   rpc/protocol.h \
   rpc/server.h \
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -27,6 +27,7 @@
 #include "policy/policy.h"
 #include "rpc/server.h"
 #include "rpc/register.h"
+#include "rpc/blockchain.h"
 #include "script/standard.h"
 #include "script/sigcache.h"
 #include "scheduler.h"
--- a/src/rest.cpp
+++ b/src/rest.cpp
@@ -9,6 +9,7 @@
 #include "primitives/transaction.h"
 #include "validation.h"
 #include "httpserver.h"
+#include "rpc/blockchain.h"
 #include "rpc/server.h"
 #include "streams.h"
 #include "sync.h"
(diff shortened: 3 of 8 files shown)


jnewbery:
Tested ACK https://github.com/bitcoin/bitcoin/pull/10095/commits/e6dcfeec05d6527ab148550836937ff435e6449c

Can `RPCNotifyBlockChange()` (https://github.com/bitcoin/bitcoin/pull/10095/files#diff-a0fb7350bf105d107147e0d10f727f47L202) also be declared in rpc/blockchain.h ?

jonasschnelli:
utACK e6dcfeec05d6527ab148550836937ff435e6449c

laanwj:
@jnewbery Done

laanwj:
I see rest.cpp also uses a few functions from this file:
```
extern void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry);
extern UniValue blockToJSON(const CBlock& block, const CBlockIndex* blockindex, bool txDetails = false);
extern UniValue mempoolInfoToJSON();
extern UniValue mempoolToJSON(bool fVerbose = false);
extern void ScriptPubKeyToJSON(const CScript& scriptPubKey, UniValue& out, bool fIncludeHex);
extern UniValue blockheaderToJSON(const CBlockIndex* blockindex);
```
Going to add them to the header as well.

jnewbery:
Looks good to me. Tested ACK c7e9aa7c1a92adba141958e5f86d267f34cc9508

dcousens (approved):

jtimon:
ScriptPubKeyToJSON and TxToJSON are defined in https://github.com/bitcoin/bitcoin/blob/master/src/rpc/rawtransaction.cpp#L37 not rpc/blockchain.cpp. Either move the declaration or the definition for both to be in the same .h and .cpp

utACK c7e9aa7c1a92adba141958e5f86d267f34cc9508 modulo nit.

laanwj:
@jtimon Good catch. I'll move them to their own header.

jnewbery:
Or just refactor them out entirely since they're duplicate code: https://github.com/bitcoin/bitcoin/pull/8824

laanwj:
Yes makes sense I'll just remove them from the header and ignore them for
this PR.

laanwj:
Done, now those two lines remain in `rest.cpp`, will leave it to another pull to deal with them (I did remove the redundant `extern` though).

PR #10096, opened 2017-03-27
Check that all test scripts in test/functional are being run

jnewbery:
This commit checks that all of the python files in the test/functional
directory are listed in test_runner.py.

Suggested by @MarcoFalke here: https://github.com/bitcoin/bitcoin/pull/10073#issuecomment-289214313

This doesn't change the directory structure. We can do that in a separate PR if we decide that's a good idea.


MarcoFalke:
utACK 29d6634a691044cb5d87173c029c8da309c0102e

PR #10098, opened 2017-03-27
Make qt wallet test compatible with qt4

ryanofsky:
Unlike Qt5, the Qt4 signals implementation doesn't allow a signal to be
directly connected to a c++ lambda expression. Work around this by defining a
Callback QObject with a virtual method that can forward calls to a closure.

The Qt4 error was reported by @pstratem in https://github.com/bitcoin/bitcoin/pull/10039#issuecomment-289248763

### DIFF
--- a/src/qt/callback.h
+++ b/src/qt/callback.h
@@ -1,0 +1,30 @@
+#ifndef BITCOIN_QT_CALLBACK_H
+#define BITCOIN_QT_CALLBACK_H
+
+#include <QObject>
+
+class Callback : public QObject
+{
+    Q_OBJECT
+public Q_SLOTS:
+    virtual void call() = 0;
+};
+
+template <typename F>
+class FunctionCallback : public Callback
+{
+    F f;
+
+public:
+    FunctionCallback(F f_) : f(std::move(f_)) {}
+    ~FunctionCallback() override {}
+    void call() override { f(this); }
+};
+
+template <typename F>
+FunctionCallback<F>* makeCallback(F f)
+{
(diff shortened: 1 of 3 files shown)


pstratem:
tested but i have no idea about qt so uh tested-almost-ack ?

jonasschnelli:
This is a really nice approach how to keep lambda-ish syntax for Qt4 signals.
utACK e9a64615c8e18692a775765787f404266767260b.
Will test soon.

laanwj:
> tested but i have no idea about qt so uh tested-almost-ack ?

You don't need to have an idea about qt to confirm whether the test compiles and passes on qt4 after this (which I guess you did?).

jonasschnelli:
Tested ACK e9a64615c8e18692a775765787f404266767260b.
Could successfully compile this on my Ubuntu 14.04.5 QT4 VM.

src/qt/callback.h
```diff
@@ -0,0 +1,30 @@
+#ifndef BITCOIN_QT_CALLBACK_H
```
paveljanik: Copyright header missing.

paveljanik (commented):
utACK https://github.com/bitcoin/bitcoin/commit/e9a64615c8e18692a775765787f404266767260b

Can't test with Qt4 though.

gmaxwell (approved):
tested ACK. (without this, master does not compile on my debian host)

droark:
Just curious, does this need to be backported to 0.14? I noticed that a mempool test was updated and figured this might be worthwhile too.

PR #10104, opened 2017-03-28
linearize script: Option to use RPC cookie

### LINKED ISSUE #10103: Linearize script does not use rpc cookie
It should

achow101:
This PR adds the a `datadir` option to the linearize configuration file to allow linearize-hashes to look for and use the RPC cookie file in the datadir. It will still allow rpcuser and rpcpassword to be specified and those options will take precendence if both rpcuser/pass and datadir config options are set.

closes #10103

### DIFF
--- a/contrib/linearize/linearize-hashes.py
+++ b/contrib/linearize/linearize-hashes.py
@@ -16,6 +16,8 @@ import json
 import re
 import base64
 import sys
+import os
+import os.path
 
 settings = {}
 
@@ -93,6 +95,14 @@ def get_block_hashes(settings, max_blocks_per_call=10000):
 
 		height += num_blocks
 
+def get_rpc_cookie():
+	# Open the cookie file
+	with open(os.path.join(os.path.expanduser(settings['datadir']), '.cookie'), 'r') as f:
+		combined = f.readline()
+		combined_split = combined.split(":")
+		settings['rpcuser'] = combined_split[0]
+		settings['rpcpassword'] = combined_split[1]
+
 if __name__ == '__main__':
 	if len(sys.argv) != 2:
 		print("Usage: linearize-hashes.py CONFIG-FILE")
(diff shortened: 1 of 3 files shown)


jonasschnelli:
Nice.
utACK 476c6d4c8dcf41f9e3accf5be1b84fdd4718046b

laanwj:
Makes sense, utACK

laanwj:
@pstratem can you confirm that this implements what you requested in #10103?

MarcoFalke:
utACK 476c6d4

contrib/linearize/linearize-hashes.py
```diff
@@ -93,6 +93,14 @@ def get_block_hashes(settings, max_blocks_per_call=10000):
 
 		height += num_blocks
 
+def get_rpc_cookie(settings):
+	# Open the cookie file
+	with open(settings['datadir'] + '/.cookie', 'r') as f:
```
pstratem: open(os.path.join(os.path.expanduser(settings['datadir']), '/.cookie'), 'r')

pstratem (changes requested):

contrib/linearize/linearize-hashes.py
achow101: done

pstratem:
@achow101 need to import os.path

pstratem:
my bad
`diff --git a/contrib/linearize/linearize-hashes.py b/contrib/linearize/linearize-hashes.py
index 721d204..db8eb70 100755
--- a/contrib/linearize/linearize-hashes.py
+++ b/contrib/linearize/linearize-hashes.py
@@ -16,6 +16,8 @@ import json
 import re
 import base64
 import sys
+import os
+import os.path
 
 settings = {}
 
@@ -93,9 +95,9 @@ def get_block_hashes(settings, max_blocks_per_call=10000):
 
                height += num_blocks
 
-def get_rpc_cookie(settings):
+def get_rpc_cookie():
        # Open the cookie file
-       with open(os.path.join(os.path.expanduser(settings['datadir']), '/.cookie'), 'r') as f:
+       with open(os.path.join(os.path.expanduser(settings['datadir']), '.cookie'), 'r') as f:
                combined = f.readline()
                combined_split = combined.split(":")
                settings['rpcuser'] = combined_split[0]
@@ -150,6 +152,6 @@ if __name__ == '__main__':
 
        # Get the rpc user and pass from the cookie if the datadir is set
        if use_datadir:
-               get_rpc_cookie(settings)
+               get_rpc_cookie()
 
        get_block_hashes(settings)
`

achow101:
@pstratem Is that right?

PR #10105, opened 2017-03-28
[tests] fixup - make all Travis test runs quiet, non just cron job runs

jnewbery:
#9780 incorrectly only makes Travis nightly cron jobs run with `--quiet` option. All travis jobs should use `--quiet` option.


MarcoFalke:
utACK 91f1f19674e9cc8a0a34df73484ffcefd6227798

PR #10109, opened 2017-03-28
Remove SingleNodeConnCB

jnewbery:
This PR merges the NodeConnCB and SingleNodeConnCB into a single
class (called NodeConnCB). The original intent for the NodeConnCB was to
be able to have a python 'mininode' connect to multiple running
bitcoinds. This has never been used and can be achieved more easily by
having multiple NodeConns backed by a common datastore if it is ever
needed.

The changes in mininode.py are just code moves (and merging the two
classes into a single class). The code changes in the individual test
cases are changing the subclasses to subclass from NodeConnCB instead of
SingleNodeConnCB. There is a lot of duplicate code in the subclasses
that can be removed in future commits.

@sdaftuar as discussed - should be an easy review.

### DIFF
--- a/test/functional/test_framework/mininode.py
+++ b/test/functional/test_framework/mininode.py
@@ -1476,25 +1476,9 @@ class NodeConnCB(object):
         self.deliver_sleep_time = None
         # Remember the services our peer has advertised
         self.peer_services = None
-
-    def set_deliver_sleep_time(self, value):
-        with mininode_lock:
-            self.deliver_sleep_time = value
-
-    def get_deliver_sleep_time(self):
-        with mininode_lock:
-            return self.deliver_sleep_time
-
-    # Spin until verack message is received from the node.
-    # Tests may want to use this as a signal that the test can begin.
-    # This can be called from the testing thread, so it needs to acquire the
-    # global lock.
-    def wait_for_verack(self):
-        while True:
-            with mininode_lock:
-                if self.verack_received:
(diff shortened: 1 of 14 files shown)


sdaftuar:
utACK 159fe88.  Thanks for tackling!

test/functional/test_framework/mininode.py
```diff
@@ -1506,17 +1490,36 @@ def deliver(self, conn, message):
...
+    def on_cmpctblock(self, conn, message): pass
+    def on_feefilter(self, conn, message): pass
+    def on_getaddr(self, conn, message): pass
+    def on_getblocks(self, conn, message): pass
+    def on_getblocktxn(self, conn, message): pass
+    def on_getdata(self, conn, message): pass
+    def on_getheaders(self, conn, message): pass
+    def on_headers(self, conn, message): pass
+    def on_mempool(self, conn): pass
+    def on_open(self, conn): pass
```
MarcoFalke: style nit: Any reason you don't mention on_pong here?

maflcko (commented):
utACK 159fe88abfadf67410578c145e631819cf50b660

test/functional/test_framework/mininode.py
jnewbery: I put all the no-op callbacks in this block and the callbacks that do something below. Happy to move them around if you think that's unclear.

PR #10111, opened 2017-03-28
[WIP] Remove NetworkThread from individual test cases

jnewbery:
This commit starts the NetworkThread automatically in the test
framework, which means that the individual test cases don't need to
start the thread after they've added the p2p connections.

It also updates the NodeConnCB.add_connection() method to wait for
verack by default. This means the test writer only needs to call
add_connection() in order to have a p2p connection ready for use.

This builds on #10109 . Only the 'Remove NetworkThread from individual test cases' commit needs to be reviewed

As discussed @sdaftuar - should be another easy review.

### DIFF
--- a/test/functional/README.md
+++ b/test/functional/README.md
@@ -52,10 +52,6 @@ events of interest arrive.
 * You can pass the same handler to multiple ```NodeConn```'s if you like, or pass
 different ones to each -- whatever makes the most sense for your test.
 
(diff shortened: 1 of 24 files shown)


jnewbery:
Closing this for now. I still think this is the right way to go, but I need to sequence the commits differently. NodeConn and NodeConnCB need to be initialized differently to avoid the timing issues that are happening on Travis.

PR #10113, opened 2017-03-28
Fix test_bitcoin-qt when DISPLAY is unset

ryanofsky:
Issue reported by @TheBlueMatt in #10110

### DIFF
--- a/src/qt/test/test_main.cpp
+++ b/src/qt/test/test_main.cpp
@@ -19,7 +19,9 @@
 
 #include <QApplication>
 #include <QObject>
+#include <QPluginLoader>
 #include <QTest>
+#include <QtDebug>
 
 #include <openssl/ssl.h>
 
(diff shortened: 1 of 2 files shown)


src/qt/test/test_main.cpp
```diff
@@ -41,6 +43,15 @@ Q_IMPORT_PLUGIN(QCocoaIntegrationPlugin);
 #endif
 #endif
 
+static bool UsesXcb() {
+    for (QObject* plugin: QPluginLoader::staticInstances()) {
```
theuni: How does this react if xcb is available but not default? For ex, if egl or wayland would otherwise be used?

laanwj:
I think the `make check` tests should never try to use the DISPLAY (or the computer's GUI otherwise, e.g. egl/wayland).
There's some place for tests that require rendering (usually combined with a "monkey" to mimic user interaction) but these should IMO be separate.

src/qt/test/test_main.cpp
ryanofsky: > How does this react if xcb is available but not default? For ex, if egl or wayland would otherwise be used?

This is a good point, I suspect the check is broken in this case.

ryanofsky (commented):
> I think the make check tests should never try to use the DISPLAY (or the computer's GUI otherwise, e.g. through egl/wayland/win32/quartz/...).
There's a place for tests that require rendering (usually combined with a "monkey" to mimic user interaction) but these should IMO be separate when we have them.

The test doesn't actually require any rendering, it's just failing to initialize QApplication (which must be calling some libxcb functions) and crashing on startup.

I'm closing this in favor of an alternative fix in #10117 though, which modifies "make check" to run the test under [xvfb](https://en.wikipedia.org/wiki/Xvfb) if it's available.

PR #10114, opened 2017-03-29
[tests] sync_with_ping should assert that ping hasn't timed out

jnewbery:
sync_with_ping() currently returns false if the timeout expires, and it is
the caller's responsibility to fail the test. However, none of the tests
currently assert on sync_with_ping()'s return code. This commit adds an
assert to sync_with_ping so the test will fail if the timeout expires.

This commit also removes all the duplicate implementations of
sync_with_ping() from the individual tests.

~Only the second commit here needs to be reviewed. The first commit https://github.com/bitcoin/bitcoin/commit/159fe88abfadf67410578c145e631819cf50b660 is #10109.~ __EDIT: #10109 is now merged.__

__EDIT: The first commit in this PR ("Add send_await_disconnect() method to p2p-compactblocks.py") is a test case bug fix that needs to be merged before merging the sync_with_ping() change__

Should fix the intermittent failure reported here: https://github.com/bitcoin/bitcoin/pull/10073#issuecomment-289466653

sync'ing and pinging @sdaftuar

### DIFF
--- a/test/functional/assumevalid.py
+++ b/test/functional/assumevalid.py
@@ -190,7 +190,8 @@ class AssumeValidTest(BitcoinTestFramework):
         # Send all blocks to node1. All blocks will be accepted.
         for i in range(2202):
             node1.send_message(msg_block(self.blocks[i]))
-        node1.sync_with_ping()  # make sure the most recent block is synced
+        # Syncing 2200 blocks can take a while on slow systems. Give it plenty of time to sync.
+        node1.sync_with_ping(120)
         assert_equal(self.nodes[1].getblock(self.nodes[1].getbestblockhash())['height'], 2202)
 
         # Send blocks to node2. Block 102 will be rejected.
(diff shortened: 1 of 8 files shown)


sdaftuar:
I think travis may have caught a bug in the p2p-compactblocks.py test, which is now throwing an error that the sync with ping failed?

jnewbery:
Oh good. We've caught our first fish.

I've pushed a commit that changes p2p-compactblock to use a new method send_await_disconnect(), which awaits a disconnect rather than a pong. 

The commit adds the method to p2p-compactblocks.py, but a future commit could move it to mininode since it could be useful more generally. The commit reduces the p2p-compactblock runtime by 30 seconds.

I've added the new commit _before_ the commit that changes sync_with_ping() so git_bisect doesn't break.

laanwj:
utACK 6a18bb9

sdaftuar:
utACK 6a18bb9

PR #8348, opened 2016-07-17
Trivial: Segwit: Don't call IsWitnessEnabled from ContextualCheckBlock

jtimon:
IsWitnessEnabled() should not be called from libconsensus code, only from outside what's already encapsulated.
The 3 callers already `AssertLockHeld(cs_main);`, this is my grep:

```
./src/main.h:448:bool ContextualCheckBlock(const CBlock& block, CValidationState& state, CBlockIndex *pindexPrev);
./src/main.cpp:3525:bool ContextualCheckBlock(const CBlock& block, CValidationState& state, CBlockIndex * const pindexPrev)
./src/main.cpp:3602:        return state.DoS(100, error("ContextualCheckBlock(): cost limit failed"), REJECT_INVALID, "bad-blk-cost");
./src/main.cpp:3689:    if ((!CheckBlock(block, state, chainparams.GetConsensus(), GetAdjustedTime())) || !ContextualCheckBlock(block, state, pindex->pprev)) {
./src/main.cpp:3780:    if (!ContextualCheckBlock(block, state, pindexPrev))
```

I could have said this during segwit review instead of now, sorry @sipa.

PR #10116, opened 2017-03-29
Consensus: use IsWitnessEnabled in ContextualCheckBlock

roconnor-blockstream:
Replaced the SegWit deployment check in `ContextualCheckBlock` with a call to `IsWitnessEnabled` instead.


JeremyRubin:
NACK. 

Was explicitly reverted here.
https://github.com/bitcoin/bitcoin/pull/8348

PR #10117, opened 2017-03-29
Fix "make check" on systems with qt and libxcb but no active DISPLAY

ryanofsky:
Issue reported by @TheBlueMatt in #10110

This is an alternative to the fix in #10113. Instead of modifying test_bitcoin-qt to try to detect whether it's possible to run GUI tests with a check that is probably not reliable, it just adds a --skip-gui-tests option that can be controlled externally. Then, it uses autoconf checks to see if passing the option is necessary.


theuni:
@ryanofsky This seems like overkill. How about https://github.com/theuni/bitcoin/commit/b493819ac800c6ba2c34b15ab9a11c8bd2d0c1f2 ?

Afaik, the minimal platform is available everywhere, and seems to be enough to make this work.

(this needs a bit of extra configure logic though, so that we don't disable all of qt in the rare event that qminimal is missing).

ryanofsky:
Interesting, I didn't know about the minimal platform. If you like, I can use your change as a starting point and try to add the needed configuration logic. I'd also like to keep optional support for the regular platforms, because it can be useful to show or interact with widgets while debugging the test.

theuni:
Sure. You can just make it minimal by default, and overridable to something else.

Fwiw, you can use either "-platform minimal" or use the QT_QPA_PLATFORM env. The env was easier for a poc, but maybe you'll want to set the platform arg if there's not one set already.

ryanofsky:
Closing this in favor of #10142.

> this needs a bit of extra configure logic though, so that we don't disable all of qt in the rare event that qminimal is missing

It seems easier to just require the minimal platform, unless there are cases where you think this isn't a reasonable requirement. Adding extra configure logic to make the check for `-lqminimal` nonfatal in a static Qt build isn't enough to know whether the `setenv(QT_QPA_PLATFORM)` call will be safe, so the test could still fail without more runtime code, which seems not worth writing.

PR #9494, opened 2017-01-09
Introduce an ArgsManager class encapsulating cs_args, mapArgs  and mapMultiArgs

jtimon:
Continues #9243 by @TheBlueMatt 
Introduce an ArgsManager class encapsulating cs_args, mapArgs  and mapMultiArgs.

To reduce disruption, the used functions are kept as wrapers. The wrapers can be removed later, see https://github.com/jtimon/bitcoin/compare/0.13-args-class...jtimon:0.13-args-remove-wrapers for the less disruptive ones to remove.

PR #10119, opened 2017-03-30
Util: Remove ArgsManager wrappers:

jtimon:
- ParseParameters
- ReadConfigFile
- ForceSetArg
- SoftSetArg

Includes:

- [ ] Introduce an ArgsManager class encapsulating cs_args, mapArgs and mapMultiArgs #9494 

From the wrappers newly introduced with ArgsManager, these seem currently the easiest to remove.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -105,8 +105,6 @@ const char * const BITCOIN_CONF_FILENAME = "bitcoin.conf";
 const char * const BITCOIN_PID_FILENAME = "bitcoind.pid";
 
 ArgsManager argsGlobal;
-static std::map<std::string, std::vector<std::string> > _mapMultiArgs;
-const std::map<std::string, std::vector<std::string> >& mapMultiArgs = _mapMultiArgs;
 bool fPrintToConsole = false;
 bool fPrintToDebugLog = true;
 
@@ -384,7 +382,7 @@ void ArgsManager::ParseParameters(int argc, const char* const argv[])
 {
     LOCK(cs_args);
     mapArgs.clear();
-    _mapMultiArgs.clear();
+    mapMultiArgs.clear();
 
     for (int i = 1; i < argc; i++)
     {
@@ -412,7 +410,7 @@ void ArgsManager::ParseParameters(int argc, const char* const argv[])
         InterpretNegativeSetting(str, strValue);
 
         mapArgs[str] = strValue;
-        _mapMultiArgs[str].push_back(strValue);
+        mapMultiArgs[str].push_back(strValue);
     }
 }
 
@@ -473,7 +471,7 @@ void ArgsManager::ForceSetArg(const std::string& strArg, const std::string& strV
 {
     LOCK(cs_args);
     mapArgs[strArg] = strValue;
-    _mapMultiArgs[strArg].push_back(strValue);
+    mapMultiArgs[strArg].push_back(strValue);
 }
 
 void ParseParameters(int argc, const char* const argv[])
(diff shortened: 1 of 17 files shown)


JeremyRubin (commented):
utack

src/util.cpp
```diff
@@ -378,63 +374,94 @@ void ParseParameters(int argc, const char* const argv[])
...
     else
         return SoftSetArg(strArg, std::string("0"));
 }
 
-void ForceSetArg(const std::string& strArg, const std::string& strValue)
+void ArgsManager::ForceSetArg(const std::string& strArg, const std::string& strValue)
 {
     LOCK(cs_args);
     mapArgs[strArg] = strValue;
+    mapMultiArgs[strArg].push_back(strValue);
```
jtimon: I believe push_back, but it's a good point, I am not sure now.

src/util.h
```diff
@@ -133,6 +131,16 @@ inline bool IsSwitchChar(char c)
 #endif
 }
 
+class ArgsManager
+{
+protected:
+    CCriticalSection cs_args;
+    std::map<std::string, std::string> mapArgs;
+    std::map<std::string, std::vector<std::string> > mapMultiArgs;
```
jtimon: I was just using the same that was used before. We can move to multimap, what would be the gain?

JeremyRubin: No real gain, just slightly simpler (because multi_map is intended to be used I think exactly in this type of case).

src/util.cpp
```diff
@@ -378,63 +374,94 @@ void ParseParameters(int argc, const char* const argv[])
...
 }
 
+bool IsArgSet(const std::string& strArg)
```
jtimon: The less disruptive to remove are removed in https://github.com/bitcoin/bitcoin/pull/10119 I'm happy to squash that in https://github.com/bitcoin/bitcoin/pull/9494
Others are more disruptive to remove, but I'm working on doing it for SoftSetBoolArg and IsArgSet in https://github.com/jtimon/bitcoin/compare/0.14-args-wrappers...jtimon:0.14-args-isargset-wrapper but I'm not sure everybody will like that way of doing it.

jtimon:
Rebased.

src/util.cpp
```diff
@@ -378,63 +374,94 @@ void ParseParameters(int argc, const char* const argv[])
         InterpretNegativeSetting(str, strValue);
 
         mapArgs[str] = strValue;
-        _mapMultiArgs[str].push_back(strValue);
+        mapMultiArgs[str].push_back(strValue);
     }
 }
 
-bool IsArgSet(const std::string& strArg)
+std::vector<std::string> ArgsManager::ArgsAt(const std::string& strArg)
```
jtimon: See https://github.com/bitcoin/bitcoin/pull/9494#discussion_r95578281 for some reason all your reviews are appearing in #10119 instead of #9494 . Can you make the comments there for other reviewers to get them? Remember after #9494 this PR (#10119) is a single commit (that could be squashed to #9494 or not)

src/util.h
jtimon: I thought it was more for when you wanted the elements on the vector to be ordered in certain way or something. Leaving the same structure seems easier to review, but once encapsulated it is easy to make further improvements to ArgsManager's internal implementation.

src/util.cpp
JeremyRubin: oh -- I didn't understand 'includes' to mean builds on the PR. gotcha

jtimon:
Incorportated in #9494

PR #10120, opened 2017-03-30
util: Work around (virtual) memory exhaustion on 32-bit w/ glibc

laanwj:
glibc-specific: On 32-bit systems set the number of arenas to 1. By default, since glibc 2.10, the C library will create up to two heap arenas per core. This is known to cause excessive virtual address space usage in our usage. Work around it by setting the maximum number of arenas to 1.

I documented this before in [Reducing bitcoind memory usage](https://gist.github.com/laanwj/efe29c7661ce9b6620a7) but with the reported OOM issues in 0.14 it seems more pressing to do it by default when there is low virtual memory space (e.g. on 32 bit).

To test:
```
bitcoin-cli getmemoryinfo mallocinfo | grep "<heap"
```
This will show an entry per heap.

Tested on 32-bit ARM, 4-core:
```xml
<heap nr="0">
```

Tested on 64-bit x86, 6-core:
```xml
<heap nr="0">
<heap nr="1">
<heap nr="2">
<heap nr="3">
<heap nr="4">
<heap nr="5">
<heap nr="6">
<heap nr="7">
<heap nr="8">
<heap nr="9">
<heap nr="10">
<heap nr="11">
<heap nr="12">
```

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -574,6 +574,14 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#include <malloc.h>]],
  [ AC_MSG_RESULT(no)]
 )
 
+dnl Check for mallopt(M_ARENA_MAX) (to set glibc arenas)
+AC_MSG_CHECKING(for mallopt M_ARENA_MAX)
+AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[#include <malloc.h>]],
+ [[ mallopt(M_ARENA_MAX, 1); ]])],
+ [ AC_MSG_RESULT(yes); AC_DEFINE(HAVE_MALLOPT_ARENA_MAX, 1,[Define this symbol if you have mallopt with M_ARENA_MAX]) ],
+ [ AC_MSG_RESULT(no)]
+)
+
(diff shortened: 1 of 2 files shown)


gmaxwell (approved):
utACK.

JeremyRubin:
utACK.

laanwj:
> I'm not familiar with the intricacies here, so I'll defer judgement on the root issue.

The root issue is that freed object stay around for a while for reuse, so having 8 heaps means that they can potentially stick around 8 times. This is extremely bad considering e.g. how large the consecutive objects are that leveldb allocates.

> The implementation suggests that it's meant to apply to glibc only, but there's no obvious check for that. Is M_ARENA_MAX itself glibc specific, or do we need some macro to check for it?

As always, feel free to suggest a better way to do the autoconf stuff. Yes, M_ARENA_MAX is specific to glibc (or anything deriving from it).

theuni:
I didn't mean to imply that anything was incorrect, just wanted to clarify the assumptions. Thanks for explaining. Code looks good. utACK 625488ace5c41ccba1b68323a72588000782f820

PR #10121, opened 2017-03-30
Add missing header file

shitakee:
We need to include <sys/types.h> because we use ssize_t on the file.

### DIFF
--- a/src/leveldb/db/db_iter.cc
+++ b/src/leveldb/db/db_iter.cc
@@ -2,6 +2,7 @@
 // Use of this source code is governed by a BSD-style license that can be
 // found in the LICENSE file. See the AUTHORS file for names of contributors.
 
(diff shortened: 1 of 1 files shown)


JeremyRubin:
Hi @shitakee! Thanks for making a contribution! Please feel free to reach out to me if you want to chat about how to contribute to bitcoin core.

Unfortunately, in this case, this looks like an upstream problem for leveldb, so we typically won't merge it here unless it's causing a specific problem. Are you failing to build somewhere?

fanquake:
Indeed, @shitakee if you could re-submit this PR to our upstream leveldb repo (https://github.com/bitcoin-core/leveldb), including some more information about the specific build error/issue you are seeing, that would be appreciated.

fanquake:
Closing, can be resubmitted upstream.

PR #10124, opened 2017-03-30
[test] Suppress test logging spam

jnewbery:
__EDIT2: reimplemented on top of #10123 now that #9424 is merged. This PR is now ready for review. Only the second commit is to be reviewed. The first commit is #10123__

~__EDIT: reduced scope of this PR. I'm closing #10123 until #9424 is merged, but that shouldn't block reducing the test_framework log spam__~

This commit reduces spammy logging by the test framework and the
bitcoind instances in a test run:

- truncate logging send/receive message in mininode to 200 characters.
mininode was previously logging the entire message sent received, which
can be up to 1MB for a full block.
- exclude logging for leveldb and libevent in the bitcoind instances

requires #10123

### DIFF
--- a/test/functional/test_framework/mininode.py
+++ b/test/functional/test_framework/mininode.py
@@ -1749,7 +1749,7 @@ class NodeConn(asyncore.dispatcher):
     def send_message(self, message, pushbuf=False):
         if self.state != "connected" and not pushbuf:
             raise IOError('Not connected, no pushbuf')
-        logger.debug("Send message to %s:%d: %s" % (self.dstaddr, self.dstport, repr(message)))
+        self._log_message("send", message)
         command = message.command
         data = message.serialize()
         tmsg = self.MAGIC_BYTES[self.network]
@@ -1771,8 +1771,18 @@ class NodeConn(asyncore.dispatcher):
                 self.messagemap[b'ping'] = msg_ping_prebip31
         if self.last_sent + 30 * 60 < time.time():
             self.send_message(self.messagemap[b'ping']())
-        logger.debug("Received message from %s:%d: %s" % (self.dstaddr, self.dstport, repr(message)))
+        self._log_message("receive", message)
         self.cb.deliver(self, message)
+
+    def _log_message(self, direction, msg):
+        if direction == "send":
+            log_message = "Send message to "
+        elif direction == "receive":
+            log_message = "Received message from "
+        log_message += "%s:%d: %s" % (self.dstaddr, self.dstport, repr(msg)[:500])
+        if len(log_message) > 500:
+            log_message += "... (msg truncated)"
+        logger.debug(log_message)
 
     def disconnect_node(self):
         self.disconnect = True
--- a/test/functional/test_framework/util.py
+++ b/test/functional/test_framework/util.py
@@ -315,7 +315,7 @@ def start_node(i, dirname, extra_args=None, rpchost=None, timewait=None, binary=
     datadir = os.path.join(dirname, "node"+str(i))
     if binary is None:
         binary = os.getenv("BITCOIND", "bitcoind")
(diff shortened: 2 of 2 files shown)


test/functional/test_framework/mininode.py
```diff
@@ -1768,7 +1768,7 @@ def got_message(self, message):
                 self.messagemap[b'ping'] = msg_ping_prebip31
         if self.last_sent + 30 * 60 < time.time():
             self.send_message(self.messagemap[b'ping']())
-        logger.debug("Received message from %s:%d: %s" % (self.dstaddr, self.dstport, repr(message)))
+        logger.debug("Received message from %s:%d: %.200s" % (self.dstaddr, self.dstport, repr(message)))
```
JeremyRubin: 200 seems a bit small, and I'd imagine if we get a big error we don't expect we'd want to see it all? Maybe a slightly better way to do it is to filter on message type -- can you detect things that are big that we don't really want, e.g. messgae.startswith('bigmsgstartswith')?

also I prefer doing repr(message)[:200] than format string hacking as I always have to re-read format string rules to remember what they do.

JeremyRubin:
utack

test/functional/test_framework/mininode.py
jnewbery: Yes, we could do something more intelligent here, but I think this is at least an improvement on current behaviour. At the moment, a test like pruning.py , which sends hundreds of full blocks, will result in a log file that is GBs in size.

I can't think of a way to detect things that we don't want. A tx message could be up to 1MB, so should we filter them all out? How about block messages? I think the best we can do is truncate everything. Why 200 bytes? Because it seemed to me that 200 bytes would cover the most important information in the message, but I'm happy to bump it up a bit if you think that'd be useful.

I've changed the string formatting to a string slice to truncate the log as suggested.

laanwj:
Needs rebase after #10123

jnewbery:
Rebased

sipa:
Concept ACK

test/functional/test_framework/mininode.py
```diff
@@ -1771,7 +1771,7 @@ def got_message(self, message):
                 self.messagemap[b'ping'] = msg_ping_prebip31
         if self.last_sent + 30 * 60 < time.time():
             self.send_message(self.messagemap[b'ping']())
-        logger.debug("Received message from %s:%d: %s" % (self.dstaddr, self.dstport, repr(message)))
+        logger.debug("Received message from %s:%d: %s" % (self.dstaddr, self.dstport, repr(message)[:500]))
```
MarcoFalke: Maybe indicate with three dots, that the message is truncated?

maflcko (commented):
utACK b6c9a9a08bd58f53cbf2adbd58573348c63be79e

test/functional/test_framework/mininode.py
jnewbery: I've added this in the latest fixup commit. I also added a _log_message() method to avoid code duplication.

test/functional/test_framework/mininode.py
```diff
@@ -1749,7 +1749,7 @@ def got_data(self):
     def send_message(self, message, pushbuf=False):
         if self.state != "connected" and not pushbuf:
             raise IOError('Not connected, no pushbuf')
-        logger.debug("Send message to %s:%d: %s" % (self.dstaddr, self.dstport, repr(message)[:500]))
+        self._log_message(True, message)
```
MarcoFalke: nit: Instead of a unnamed boolean, you could pass in the string "Sent" or "Received" directly?

jnewbery: sure

jnewbery:
@MarcoFalke nit addressed.

Let me know if you're happy for merge and I'll squash the fixup commits.

jnewbery:
commits squashed.

PR #10129, opened 2017-03-30
scheduler: fix sub-second precision with boost < 1.50

theuni:
For systems with older boost, cpu usage has been ramped up since 735d9b5362aeca34c0e62006986fe9d82c24ca08, due to a time miscalculation. This was never encountered previously because this is the first use of a sub-second repeat interval for a scheduled event.

### DIFF
--- a/src/scheduler.cpp
+++ b/src/scheduler.cpp
@@ -23,7 +23,9 @@ CScheduler::~CScheduler()
 #if BOOST_VERSION < 105000
 static boost::system_time toPosixTime(const boost::chrono::system_clock::time_point& t)
 {
-    return boost::posix_time::from_time_t(boost::chrono::system_clock::to_time_t(t));
+    // Creating the posix_time using from_time_t loses sub-second precision. So rather than exporting the time_point to time_t,
+    // start with a posix_time at the epoch (0) and add the milliseconds that have passed since then.
+    return boost::posix_time::from_time_t(0) + boost::posix_time::milliseconds(boost::chrono::duration_cast<boost::chrono::milliseconds>(t.time_since_epoch()).count());
 }
 #endif
 


src/scheduler.cpp
```diff
@@ -23,7 +23,8 @@ CScheduler::~CScheduler()
 #if BOOST_VERSION < 105000
 static boost::system_time toPosixTime(const boost::chrono::system_clock::time_point& t)
 {
-    return boost::posix_time::from_time_t(boost::chrono::system_clock::to_time_t(t));
+    auto millis = boost::chrono::duration_cast<boost::chrono::milliseconds>(t.time_since_epoch()).count() % 1000;
+    return boost::posix_time::from_time_t(boost::chrono::system_clock::to_time_t(t)) + boost::posix_time::milliseconds(millis);
```
laanwj: This adds the current time, modulo milliseconds, to the converted time point passed in?
Please add an explanatory comment, it's far from obvious to me why this works.

laanwj: Ah not the current time, but also `t`. Makes somewhat more sense now.
What is the reason that this won't work directly?
```
return boost::posix_time::milliseconds(boost::chrono::duration_cast<boost::chrono::milliseconds>(t.time_since_epoch()).count());
```

JeremyRubin:
Is every 500 too frequent in 735d9b5 (given that it is now, as you point out, the most frequently scheduled task)? Should we also make that a second (independent of this fix)?

src/scheduler.cpp
theuni: ```from_time_t()``` (which seems to be the only reasonable way we have to create the posix_time without involving clocks, strings, or locales) is seconds since the epoch, so we lose the milliseconds there. The change here grabs them off of the input time, and adds them back after creating the posix_time.

> What is the reason that this won't work directly?
> ```
> return boost::posix_time::milliseconds(boost::chrono::duration_cast<boost::chrono::milliseconds>(t.time_since_epoch()).count());
> ```

This would return a duration rather than a time point. It would work if it were added to a time point at the epoch, though:
```
return boost::posix_time::from_time_t(0) + boost::posix_time::milliseconds(boost::chrono::duration_cast<boost::chrono::milliseconds>(t.time_since_epoch()).count());
```
Tested both, either is fine by me. Do you have a preference?

I'll add a comment either way.

theuni:
Updated with @laanwj's suggestion, and added a comment.

@JeremyRubin the 500msec matches the previous behavior, I'm unsure where that value came from. Any particular reason to suggest bumping it to a second, or just to reduce the overhead?

laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/10129/commits/e025246fe27f2d4bcdb6f4e9b5af9084dcc2e0a8

PR #10106, opened 2017-03-28
bitcoin-tx: Fix missing range check

awemany:
The number of arguments is not checked in MutateTxAddOutAddr(..), meaning
that

> ./bitcoin-tx -create outaddr=

accessed the vStrInputParts vector beyond its bounds.

PR #10130, opened 2017-03-30
bitcoin-tx input verification (awemany, jnewbery)

jnewbery:
Check number of arguments passed to bitcoin-tx in outaddr and outpubkey.

This PR also updates the bitcoin-tx tests to check bitcoin-tx stderr and verify that expected errors are raised.

Adds test cases for calling bitcoin-tx with wrong number of separators in outaddr and outpubkey.

### DIFF
--- a/src/bitcoin-tx.cpp
+++ b/src/bitcoin-tx.cpp
@@ -242,6 +242,9 @@ static void MutateTxAddOutAddr(CMutableTransaction& tx, const std::string& strIn
     std::vector<std::string> vStrInputParts;
     boost::split(vStrInputParts, strInput, boost::is_any_of(":"));
 
+    if (vStrInputParts.size() != 2)
+        throw std::runtime_error("TX output missing or too many separators");
+
     // Extract and validate VALUE
     CAmount value = ExtractAndValidateValue(vStrInputParts[0]);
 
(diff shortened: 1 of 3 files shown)


jnewbery:
Replaces #10106. @gmaxwell @instagibbs @laanwj @jtimon

jnewbery:
@theuni has suggested that checking the stderr for error text may cause problems due to unicode, eols, etc, and that we'd be better off defining our own exit codes in bitcoin-tx and testing on those. I think it's ok to test for text in stderr for now, but we a future PR should define exit codes and test them in bctest.py.

laanwj:
I don't disagree with @theuni on that in general, however checking for text in error messages is perfectly fine for tests that are supposed to test these errors. Especially for a user-facing tool. We do a similar thing for checking e.g. RPC errors where we check both the code and the message.

laanwj:
utACK 19ecd1e

gmaxwell:
@awemany Fix moved over here, your review/testing would be very welcome!

gmaxwell (approved):
ACK (did not test test, e.g. by introducing a bug).

theuni:
As mentioned in the comments: for reference, I suggested using exit codes rather than comparing text because (here specifically), Wine piles nonsense about window creation into stderr when running without DISPLAY set. For that reason, we can't test that "stderr == expected", rather we have to test "expected in stderr". That's not great, but it's not a huge deal either.

utACK 19ecd1e2e13514c8a92a2a4337c2568c3d5d13e5

PR #10133, opened 2017-03-31
Clean up calculations of pcoinsTip memory usage

morcos:
In #10126 we changed the calculation of when to flush so that the requested dbcache memory limit would be respected even by the peak usage occurring at flush.

Without doubling the default dbcache limit, this would lead to an effective default of half the previous which I believe is too low.

First commit doubles the default and applies the DB_PEAK_USAGE_FACTOR in 2 other places.
The debug logging of the usage still reports the current usage.

The second commit lowers the defaults slightly to ensure 0.14.1 has a lower memory footprint than 0.13.2.

The third commit makes the delayed flushing on large dbcaches a bit less aggressive, so the limit is less likely to be violated in practice.

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -2005,10 +2005,10 @@ bool static FlushStateToDisk(CValidationState &state, FlushStateMode mode, int n
         nLastSetChain = nNow;
     }
     int64_t nMempoolSizeMax = GetArg("-maxmempool", DEFAULT_MAX_MEMPOOL_SIZE) * 1000000;
-    int64_t cacheSize = pcoinsTip->DynamicMemoryUsage() * 2; // Compensate for extra memory peak (x1.5-x1.9) at flush time.
+    int64_t cacheSize = pcoinsTip->DynamicMemoryUsage() * DB_PEAK_USAGE_FACTOR;
     int64_t nTotalSpace = nCoinCacheUsage + std::max<int64_t>(nMempoolSizeMax - nMempoolUsage, 0);
     // The cache is large and we're within 10% and 100 MiB of the limit, but we have time now (not in the middle of a block processing).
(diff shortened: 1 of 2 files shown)


JeremyRubin:
utack 99b814f.

src/validation.cpp
```diff
@@ -3639,7 +3639,7 @@ bool CVerifyDB::VerifyDB(const CChainParams& chainparams, CCoinsView *coinsview,
             }
         }
         // check level 3: check for inconsistencies during memory-only disconnect of tip blocks
-        if (nCheckLevel >= 3 && pindex == pindexState && (coins.DynamicMemoryUsage() + pcoinsTip->DynamicMemoryUsage()) <= nCoinCacheUsage) {
+        if (nCheckLevel >= 3 && pindex == pindexState && (coins.DynamicMemoryUsage() + DB_PEAK_USAGE_FACTOR * pcoinsTip->DynamicMemoryUsage()) <= nCoinCacheUsage) {
```
sipa: The changes in the rollback test are never flushed, so there is no need for compensating for a peak?

morcos: agreed.  assumed coins was built on top of pcoinstip, missed that.  will change.

morcos:
Fixed sipa's issue.

After discussion I decided to remove the lowering of maxmempool until we are more confident that maxes sense.

I also made the flushing buffer a bit smarter.

sdaftuar:
utACK 1b55e07b7a61a9e6c299cf4c40fde80fa715d440

We should tag this for 0.14.1

sipa:
Benchmarking.

laanwj:
utACK 1b55e07

BTW @fanquake if you tag things for backport please also add the "needs backport" tag, this allows for filtering what has not been backported yet.

PR #10134, opened 2017-03-31
[qa] Fixes segwit block relay test after inv-direct-fetch was disabled

sdaftuar:
This test was passing because we never fetch blocks if we only receive
an inv and not the header (after 037159cebf1eae4445050cec029986514ed4e9e2),
and this test wasn't delivering the header.

I noticed the test falsely passing when I intentionally broke this behavior in one of my branches.

### DIFF
--- a/test/functional/p2p-segwit.py
+++ b/test/functional/p2p-segwit.py
@@ -61,16 +61,6 @@ class TestNode(NodeConnCB):
             self.send_message(msg)
         self.wait_for_getdata()
 
-    def announce_block(self, block, use_header):
-        with mininode_lock:
-            self.last_message.pop("getdata", None)
-        if use_header:
-            msg = msg_headers()
-            msg.headers = [ CBlockHeader(block) ]
-            self.send_message(msg)
-        else:
-            self.send_message(msg_inv(inv=[CInv(2, block.sha256)]))
-
     def request_block(self, blockhash, inv_type, timeout=60):
         with mininode_lock:
             self.last_message.pop("block", None)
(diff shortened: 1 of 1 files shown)


jnewbery:
`announce_block()` is only called from one place. I'd be tempted to remove the method entirely and explicitly call `send_message(<msg_header>)` from the test case so it's obvious to the reader what's happening.

sdaftuar:
Updated to remove `announce_block()`.

test/functional/p2p-segwit.py
```diff
@@ -1092,13 +1082,18 @@ def test_block_relay(self, segwit_activated):
...
             # or via parallel download (with an indeterminate delay from processing the announcement)
             # so to test that a block is NOT requested, we could guess a time period to sleep for,
             # and then check. We can avoid the sleep() by taking advantage of transaction getdata's
             # being processed after block getdata's, and announce a transaction as well,
             # and then check to see if that particular getdata has been received.
-            self.old_node.announce_block(block4, use_header=False)
+            # Since 0.14, inv's will only be responded to with a getheaders, so send a header
+            # to announce this block.
+            msg = msg_headers()
+            msg.headers = [ CBlockHeader(block4) ]
```
jnewbery: nit: spaces aren't required around list elements within brackets: https://www.python.org/dev/peps/pep-0008/#pet-peeves

jnewbery (approved):
Tested ACK c8e316bc3dd152a2d36a947195fe4b96ca4d34db

One small style nit. Up to you whether you take it or leave it.

test/functional/p2p-segwit.py
sdaftuar: I prefer the space there, otherwisethetextlookstoosquishedtome, so I'll leave this (until we merge a style guide for the project that calls for removing them).

sdaftuar:
Rebased

MarcoFalke:
utACK f19abd90537fe71e757fdbfc52cd2288e967bb6e

PR #7179, opened 2015-12-07
net: Fix sent reject messages for blocks and transactions

laanwj:
Ever since we #5913 have been sending invalid reject messages for transactions and blocks.

PR #10135, opened 2017-03-31
[p2p] Send the correct error code in reject messages

jnewbery:
This PR corrects a few places were incorrect error codes are send in reject messages, or where reject messages are sent when they shouldn't be:

- if a block failed to validate due to CheckQueue returning false, we should send a reject message with REJECT_INVALID
- if we reject a block because it's a fork from the chain before a checkpoint block that we've already processed, we should send a reject message with REJECT_CHECKPOINT
- if we reject a block because we don't have it's parent block, we shouldn't send a reject message (because the block is not necessarily invalid).

@sdaftuar we discussed this. Let me know if you have any questions.

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -855,8 +855,8 @@ void PeerLogicValidation::BlockChecked(const CBlock& block, const CValidationSta
 
     int nDoS = 0;
     if (state.IsInvalid(nDoS)) {
-        if (it != mapBlockSource.end() && State(it->second.first)) {
-            assert (state.GetRejectCode() < REJECT_INTERNAL); // Blocks are never rejected with internal reject codes
+        // Don't send reject message with code 0 or an internal reject code.
+        if (it != mapBlockSource.end() && State(it->second.first) && state.GetRejectCode() > 0 && state.GetRejectCode() < REJECT_INTERNAL) {
             CBlockReject reject = {(unsigned char)state.GetRejectCode(), state.GetRejectReason().substr(0, MAX_REJECT_MESSAGE_LENGTH), hash};
             State(it->second.first)->rejects.push_back(reject);
             if (nDoS > 0 && it->second.second)
@@ -1942,7 +1942,7 @@ bool static ProcessMessage(CNode* pfrom, const std::string& strCommand, CDataStr
             LogPrint("mempoolrej", "%s from peer=%d was not accepted: %s\n", tx.GetHash().ToString(),
                 pfrom->id,
                 FormatStateMessage(state));
-            if (state.GetRejectCode() < REJECT_INTERNAL) // Never send AcceptToMemoryPool's internal codes over P2P
+            if (state.GetRejectCode() > 0 && state.GetRejectCode() < REJECT_INTERNAL) // Never send AcceptToMemoryPool's internal codes over P2P
                 connman.PushMessage(pfrom, msgMaker.Make(NetMsgType::REJECT, strCommand, (unsigned char)state.GetRejectCode(),
                                    state.GetRejectReason().substr(0, MAX_REJECT_MESSAGE_LENGTH), inv.hash));
             if (nDoS > 0) {
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -1914,7 +1914,7 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
                                REJECT_INVALID, "bad-cb-amount");
 
     if (!control.Wait())
-        return state.DoS(100, false);
+        return state.DoS(100, error("%s: CheckQueue failed", __func__), REJECT_INVALID, "block-validation-failed");
     int64_t nTime4 = GetTimeMicros(); nTimeVerify += nTime4 - nTime2;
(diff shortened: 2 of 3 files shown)


jnewbery:
Extended test run succeeds for me locally.

src/net_processing.cpp
```diff
@@ -855,8 +855,8 @@ void PeerLogicValidation::BlockChecked(const CBlock& block, const CValidationSta
 
     int nDoS = 0;
     if (state.IsInvalid(nDoS)) {
-        if (it != mapBlockSource.end() && State(it->second.first)) {
-            assert (state.GetRejectCode() < REJECT_INTERNAL); // Blocks are never rejected with internal reject codes
```
sdaftuar: nit: I don't feel strongly but we could leave this assert in, no?

src/validation.cpp
```diff
@@ -3083,7 +3085,7 @@ static bool AcceptBlockHeader(const CBlockHeader& block, CValidationState& state
         CBlockIndex* pindexPrev = NULL;
         BlockMap::iterator mi = mapBlockIndex.find(block.hashPrevBlock);
         if (mi == mapBlockIndex.end())
-            return state.DoS(10, error("%s: prev block not found", __func__), 0, "bad-prevblk");
+            return state.DoS(10, error("%s: prev block not found", __func__), 0, "prev-blk-not-found");
```
sdaftuar: Sorry I didn't think of this before, but perhaps we were sending a reject message here specifically because we're assigning DoS points.  Suppressing the reject message will just cause the peer to have less information about why they're being disconnected.

I'm not sure I think BIP 61 is a great idea (in particular, the lack of versioning makes things difficult) but the motivation of it was in part to let peers know why they're being banned.  So maybe it makes more sense to use a REJECT_INVALID message here, rather than omit the reject message altogether, even though the block might not actually be invalid if its missing parent was provided? 

Really, though, I don't care what we do here, just raising the point in case others have an opinion.  Your change here is fine with me.

sdaftuar (approved):
One nit you can take or leave, and one point of information.  utACK 5d08c9c579ba8cc7b684105c6a08263992b08d52

src/net_processing.cpp
jnewbery: The if statement has been modified to check that `state.GetRejectCode() < REJECT_INTERNAL` (so this assert can now never fail)

src/validation.cpp
jnewbery: I'd rather extend BIP61 than deliberately send a reject message that's incorrect (the block here isn't INVALID, it just doesn't connect with our chain).

If you prefer that we send a reject message, I think we should at least change the error code to a 0x4* value and change the message to 'prev-blk-not-found'.

gmaxwell:
I don't really have any objection to making the behavior more correct, but I would rather more or less stop sending reject messages.

There is only one or two places where I'm aware of _any_ implementation doing anything remotely useful with them (some bitcoinj users use transaction rejection to figure out that their txn hasn't relayed at all, which isn't reliable but is arguably better than nothing).   They've been a source of serious bugs several times, add to code complexity, and like other error handling are fairly hard to get precise behavior from without more or less setting in stone algorithms that don't need to be set in stone.   Plus they waste bandwidth...
 
I've personally never found them useful in troubleshooting an issue.

I can't help but think that _any_ productive use for them could be better accomplished by some other means.

jnewbery:
@gmaxwell - I've heard that argument from several people, and I'm mostly on board with it. Thanks for documenting it here.

reject messages are somewhat useful in black box testing. We can send in invalid messages and assert that bitcoin rejected them _for the right reason_. The old style 'comparison' test cases use this a lot. That's not a justification for spending a heap of time on improving them, but I think there's no harm in small fixes like this. And if no-one's using them in live deployments, then it should be an uncontentious change, right? :)

Definitely agree with improving error-handling in general. I don't know exactly what that would look like, but it's beyond the scope of this PR.

For my own education, do you have references to the serious bugs that reject messages have caused?

gmaxwell:
> For my own education, do you have references to the serious bugs that reject messages have caused?

(PR #4903)  was mostly what I was thinking of there, also the original implementation of reject messages themselves printed unsanitized bytes to the logs and IIRC also didn't limit the size of strings, though I believe both of these were fixed before merge.  We have also had many PRs for where we broke sending reject messages (or sent the 'wrong' one)-- (PR #7179 comes to mind.)

For tests, we should eventually consider something more targeted. (I have thoughts there but it's too far a tangent for this discussion.-- My comment here was in vague support of your changes but also advising that I think we should be thinking about alternatives going forward.)

laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/10135/commits/5d08c9c579ba8cc7b684105c6a08263992b08d52

> reject messages are somewhat useful in black box testing. We can send in invalid messages and assert that bitcoin rejected them for the right reason.

Also for things that do custom transaction broadcasting (for example [bitcoin-submittx](https://github.com/laanwj/bitcoin-submittx) ) it's marginally useful to see if transactions have been accepted or not, and why.  I wouldn't care deeply if they were removed, though I don't see any urgency in doing so.

There were indeed some bugs in initial deployment of this feature (including, as I remember, one that could cause a reject message to be sent for a reject message), but I'd say those have been ironed out by now. Attack surface is mostly introduced by parsing/accepting messages not by emitting them, and we don't do anything with them besides optional logging (which could be in a separate logging category from NET).

PR #10089, opened 2017-03-26
Fix shadowing of 'what' as described in #10080.

dooglus:
@paveljanik is this what you meant? I didn't really understand your request.

PR #10136, opened 2017-04-01
build: Disable Wshadow warning

laanwj:
This warning was enabled by default in #8808 but it's a [continuing](https://github.com/bitcoin/bitcoin/pull/9911#issuecomment-285171447) [source](https://github.com/bitcoin/bitcoin/pull/10089#issuecomment-289369688) of [annoyance](https://github.com/bitcoin/bitcoin/pull/9911#issuecomment-285179129) for me and other developers. I'm sick of sounding like a broken record, so disable it again.


jnewbery:
concept ACK (with apologies to @paveljanik!)

If we can find a more intelligent way of managing these warnings so they're not as annoying, I'm all for it. At the moment it feels like the Wshadow tail is wagging the dog.

MarcoFalke:
utACK 2c83911

This was clearly the wrong way to prevent variable name shadows.

jonasschnelli:
utACK 2c839114010396b517dc5146065b133f578e7094

laanwj:
> This was clearly the wrong way to prevent variable name shadows.

Agreed. And there is still value in compiling with Wshadow periodically (with a carefully chosen compiler) and seeing if it finds any bug.

paveljanik:
Postmerge ACK. The real problem is that people use ancient compilers and I have not tested them all (only few selected by some distros as their default compilers).

FWIW: I'll compile all my builds with `-Wshadow` to catch all new occurrences. And of course, thanks for patience :-)

PR #10107, opened 2017-03-28
Remove unused variable. Remove accidental trailing semicolons in Python code

practicalswift:

PR #9607, opened 2017-01-20
Remove redundant semicolons in Python code

practicalswift:

PR #10137, opened 2017-04-01
Remove unused import. Remove accidental trailing semicolons.

practicalswift:


jnewbery:
Again? Wasn't this already done here: #10107 

and here: #9607

Can we limit this kind of PR to one at the beginning of each release cycle?

practicalswift:
@jnewbery Seems like the same issues are being re-introduced again and again:-)

jnewbery:
Seems like it! Can I suggest the following:

- keep one PR open with a title like "WIP: post 0.14.1 python code style nits"
- update it with all the style nits you find during the release cycle
- rebase it as soon as 0.14.1 is branched from master.
- only then ask for review/merge.

Thank you for your efforts to clean up the code. It's appreciated, but let's do it in a way that doesn't create undue work for reviewers and maintainers.

MarcoFalke:
Agree with @jnewbery. This is basically the same as with Wshadow: As long as we can not enforce a specific style, such pull requests tend to harm development on the master branch. I'd suggest to do a "Fix Wshadow" and "Remove trailing semicolons" pull request after each branch off of master. It is totally fine to have style nits sit around in master and clear them up prior to a release.

jonasschnelli:
Sorry, my fault.
Though I don't think its worth to do an extra cleanup PR. Maybe someone can fix it when he touched that file for a better reason.

practicalswift:
A few additional unused imports were introduced during the last few days - removed those, rebased and pushed :-)

PR #10142, opened 2017-04-03
Run bitcoin_test-qt under minimal QPA platform

ryanofsky:
Fixes broken "make check" reported by @TheBlueMatt  in https://github.com/bitcoin/bitcoin/issues/10110

Fix was suggested and initially implemented by @theuni in https://github.com/bitcoin/bitcoin/pull/10117#issuecomment-290275236

### DIFF
--- a/build-aux/m4/bitcoin_qt.m4
+++ b/build-aux/m4/bitcoin_qt.m4
@@ -130,6 +130,8 @@ AC_DEFUN([BITCOIN_QT_CONFIGURE],[
       if test "x$bitcoin_cv_need_acc_widget" = "xyes"; then
         _BITCOIN_QT_CHECK_STATIC_PLUGINS([Q_IMPORT_PLUGIN(AccessibleFactory)], [-lqtaccessiblewidgets])
       fi
+      _BITCOIN_QT_CHECK_STATIC_PLUGINS([Q_IMPORT_PLUGIN(QMinimalIntegrationPlugin)],[-lqminimal])
+      AC_DEFINE(QT_QPA_PLATFORM_MINIMAL, 1, [Define this symbol if the minimal qt platform exists])
       if test x$TARGET_OS = xwindows; then
         _BITCOIN_QT_CHECK_STATIC_PLUGINS([Q_IMPORT_PLUGIN(QWindowsIntegrationPlugin)],[-lqwindows])
         AC_DEFINE(QT_QPA_PLATFORM_WINDOWS, 1, [Define this symbol if the qt platform is windows])
--- a/src/qt/test/wallettests.cpp
+++ b/src/qt/test/wallettests.cpp
@@ -68,6 +68,18 @@ QModelIndex FindTx(const QAbstractItemModel& model, const uint256& txid)
 }
 
 //! Simple qt wallet tests.
+//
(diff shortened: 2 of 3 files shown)


laanwj:
Concept ACK

sipa:
Concept ACK

laanwj:
@TheBlueMatt can you check if this solves your issue #10110?

src/qt/test/wallettests.cpp
```diff
@@ -68,6 +68,18 @@ QModelIndex FindTx(const QAbstractItemModel& model, const uint256& txid)
 }
 
 //! Simple qt wallet tests.
+//
+// Test widgets can be debugged interactively calling show() on them and
+// manually running the event loop, e.g.:
```
theuni: Can this be done automatically based on the platform selected? Or does it require manual placement of the .show()s? If not, that would at least make it so that you can avoid a recompile for debug cycles.

build-aux/m4/bitcoin_qt.m4
```diff
@@ -130,6 +130,8 @@ AC_DEFUN([BITCOIN_QT_CONFIGURE],[
       if test "x$bitcoin_cv_need_acc_widget" = "xyes"; then
         _BITCOIN_QT_CHECK_STATIC_PLUGINS([Q_IMPORT_PLUGIN(AccessibleFactory)], [-lqtaccessiblewidgets])
       fi
+      _BITCOIN_QT_CHECK_STATIC_PLUGINS([Q_IMPORT_PLUGIN(QMinimalIntegrationPlugin)],[-lqminimal])
```
theuni: We need to make it so that this doesn't signal failed qt detection if libqminimal isn't found. It'd be fine if it were just test_bitcoin-qt disabled, but this would also disable the bitcoin-qt build which has no dependency on the lib.

theuni (changes requested):
Concept ACK and ACK after fixing up the configure check.

theuni:
@laanwj I can confirm that it does, I'm seeing the issue locally as well.

build-aux/m4/bitcoin_qt.m4
ryanofsky: > We need to make it so that this doesn't signal failed qt detection if libqminimal isn't found.

What should happen if minimal isn't found, and why we can't just make it a requirement? I looked into trying making it make it optional, but it seemed like a mess, see my comment https://github.com/bitcoin/bitcoin/pull/10117#issuecomment-291175634

src/qt/test/wallettests.cpp
ryanofsky: >Can this be done automatically

It would probably be possible to test code that would show all widgets. So far I've found it useful just to show selected widgets when debugging.

laanwj: As I understand it the issues troubleshooted here are rare enough that it's fine to have the developer insert his own .show()s at appropriate places. It's just useful to have this comment in that case.

jonasschnelli:
Tested ACK on debian 8 with Qt5.3 headless:
bf1026412da73b13ada3b4543e88938057362eb7

laanwj:
Thanks for testing. Then I'm going to merge this to prevent test failures for affected devs; the build system integration can be improved later.

PR #10144, opened 2017-04-03
Prioritisetransaction wasn't always updating ancestor fee

sdaftuar:
The prioritisetransaction rpc is supposed to update both the descendant and ancestor modified fees, but in the case of transactions with in-mempool descendants, the ancestor fees of those descendants was incorrect.

First commit updates a test to include a test case, and the second addresses the bug.


gmaxwell (approved):
utACK.

Should probably get a 0.14 tag.

jnewbery:
utACK 9bef02e36525d0eed4e2e31678b3ff04bbb0b8cb.

I haven't fully reviewed all the mempool ancestor/descendant fee accounting, but this looks correct.

dcousens (approved):

laanwj:
utACK 9bef02e

PR #10146, opened 2017-04-04
Better error handling for submitblock

gmaxwell:
Check transaction count early in submitblock.

There is no point in even hashing a submitted block which doesn't have a coinbase transaction.

This also results in more useful error reporting on corrupted input.

Thanks to rawodb for the bug report.


laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/10146/commits/30f30c0f99d21d6c8476c759104de5bb3218a188

jonasschnelli:
utACK 30f30c0f99d21d6c8476c759104de5bb3218a188

paveljanik:
utACK 30f30c0

laanwj:
Reminder: this needs a functional test to prevent regression, but not necessary in this PR.

jnewbery:
Regression test here: #10190

PR #10150, opened 2017-04-04
[rpc] Add logging rpc

jnewbery:
Adds an RPC to get/set active logging categories.

First commit allows all categories except libevent to be reconfigured during runtime.
Second commit modifies `InitHTTPServer()` to allow leveldb logging to be reconfigured during runtime.


src/httpserver.cpp
```diff
@@ -385,13 +385,7 @@ bool InitHTTPServer()
...
     event_set_log_callback(&libevent_log_cb);
 #if LIBEVENT_VERSION_NUMBER >= 0x02010100
-    // If -debug=libevent, set full libevent debugging.
-    // Otherwise, disable all libevent debugging.
-    if (LogAcceptCategory(BCLog::LIBEVENT)) {
-        event_enable_debug_logging(EVENT_DBG_ALL);
-    } else {
-        event_enable_debug_logging(EVENT_DBG_NONE);
-    }
+    event_enable_debug_logging(EVENT_DBG_ALL);
```
laanwj: Don't do this! This existed for a reason. What you did will reduce libevent performance extremely by always enabling their internal debugging.

laanwj:
Concept ACK ofcourse

jonasschnelli:
Oh. That's great.
Concept ACK. Will test soon.

src/httpserver.cpp
jnewbery: Are you sure? I've tested this and logging is only enabled when libevent is built in debug mode:

https://github.com/libevent/libevent/blob/2e52bace9f9998826bd3819af328efc8d18decf9/log-internal.h#L46

and

https://github.com/libevent/libevent/blob/2e52bace9f9998826bd3819af328efc8d18decf9/log-internal.h#L86

I would have thought that if we're running libevent in debug, then performance is already hosed.

If I'm wrong, I left this as a separate commit so we can leave it out if needed. The first commit forbids trying to change libevent logging during runtime and stands on its own.

leveldb: #9999 always hands a logging class to leveldb at startup, even if logging is disabled. I don't know if that affects performance. This PR doesn't change that behaviour.

laanwj: Yes, it is possible to compile libevent wholesale without debug messages. However, most distributions enable them.
Even if it is compiled with debug messages, normally they just sink at the libevent side before even being formatted: https://github.com/libevent/libevent/blob/master/log-internal.h#L80
Remember that libevent debugging is extremely noisy. It always is - the mask is currently unused - there are no specific flags.
There will be significant overhead to dropping the messages at our side, after formatting. That's worth it for the questionable advantage of being able to enable it during runtime.

jnewbery: Turns out libevent logging can be changed at runtime: https://github.com/libevent/libevent/blob/2e52bace9f9998826bd3819af328efc8d18decf9/whatsnew-2.1.txt#L269

I've pushed a new commit that leaves `InitHTTPServer()` unchanged and uses `event_enable_debug_logging()` to update logging.

jnewbery:
Pushed a new commit that I hope addresses @laanwj's point here: https://github.com/bitcoin/bitcoin/pull/10150#discussion_r109834797

src/rpc/misc.cpp
```diff
@@ -555,6 +556,65 @@ UniValue getmemoryinfo(const JSONRPCRequest& request)
...
+            "1. \"include\" (array of strings) add debug logging for these categories.\n"
+            "2. \"exclude\" (array of strings) remove debug logging for these categories.\n"
+            "\nResult: <categories>  (string): a list of the logging categories that are active.\n"
+            "\nExamples:\n"
+            + HelpExampleCli("logging", "\"[\\\"all\\\"]\" \"[\\\"http\\\"]\"")
+            + HelpExampleRpc("logging", "[\"all\"], \"[libevent]\"")
+        );
+    }
+
+    if (request.params.size() > 0 && request.params[0].isArray()) {
```
laanwj: It looks like this code is almost duplicated with the exclude case (except the |/& and true/false). Might make sense to roll this into a function that converts a list of flag strings into a bitmask and apply that wholesale.

jnewbery: sounds sensible. Done

src/rpc/misc.cpp
```diff
@@ -555,6 +556,61 @@ UniValue getmemoryinfo(const JSONRPCRequest& request)
...
+
+    // Update libevent logging. If the library version doesn't allow it, UpdateHTTPServerLogging() is a no-op.
+    UpdateHTTPServerLogging(logCategories & BCLog::LIBEVENT);
+
+    UniValue result(UniValue::VOBJ);
+    std::vector<CLogCategoryActive> vLogCatActive = ListActiveLogCategories();
+    for (const auto& logCatActive : vLogCatActive) {
+        result.pushKV(logCatActive.category, logCatActive.active);
+    }
+    result.pushKV("WARNING", "libevent logging cannot be updated when using libevent before v2.1.1.");
```
laanwj: If you add this warning I'd suggest it be context-sensitive:
- libevent logging is being updated 
- the libevent version is older than 2.1.1
Though you could also make it an error in that case!

src/rpc/misc.cpp
laanwj: Yup, thanks, much better

src/rpc/misc.cpp
jnewbery: I initially tried doing something like that, but it's a bit tricky:

- we only want to raise an error if *only* libevent logging is changed (ie if the user wants to include or exclude "all", then we shouldn't throw an error simply because "all" includes libevent).
- I've already applied the include/excude bitmasks by this point, and we need the before/after state to know whether to throw.

I'll see if I can come up with something sensible.

laanwj: I'm perfectly fine with removing this warning from this pull and doing that later. I just don't think it should be added unconditionally.

> I've already applied the include/excude bitmasks by this point, and we need the before/after state to know whether to throw.

Sure. `if((before ^ after) & BCLog::LIBEVENT)` would work :-) Though it's probably best to do the check inside UpdateHTTPServerLogging  because there you have the previous state, the desired new state, and the libevent information available.

> we only want to raise an error if only libevent logging is changed 

That's a good point, and indeed speaks for making it a warning instead.

jnewbery:
Pushed an updated version that hopefully addresses @laanwj's comments. Changes:

- `UpdateHTTPServerLogging()` is now called by `InitHTTPServer()`. This contains the `#if LIBEVENT_VERSION_NUMBER >= 0x02010100` preprocessor logic to one place.
- `UpdateHTTPServerLogging()` now removes the `BCLog::LIBEVENT` bit flag if the libevent version does not support debug logging. That means that a call to the logging RPC will return the correct value for libevent (0 if libevent logging is not supported).
- the logging RPC will throw an error if:
    - the libevent version does not support debug logging; and
    - the user tries to update *only* the `BCLog::LIBEVENT` flag

src/httpserver.cpp
```diff
@@ -435,6 +430,21 @@ bool InitHTTPServer()
...
+#if LIBEVENT_VERSION_NUMBER >= 0x02010100
+    if (enable) {
+        event_enable_debug_logging(EVENT_DBG_ALL);
+    } else {
+        event_enable_debug_logging(EVENT_DBG_NONE);
+    }
+    return true;
+#else
+    // Can't update libevent logging if version < 02010100
+    logCategories &= ~BCLog::LIBEVENT;
```
laanwj: Looks good to me. One nit: I don't think this global flag should be changed here as a side-effect. Maybe move the "disable the flag in logCategories if this function returns false" logic outside the function.

jnewbery: ok, changed so that the caller updates the logCategories bitfield (happens in two places - ` InitHTTPServer()` and the `logging()` rpc).

laanwj:
Tested ACK 7fd50c3

rebroad:
The help message could IMHO make it clearer how to enable and disable logging. I've tried "logging exclude estimatefee", "logging estimatefee false", "logging estimatefee exclude", and they all give a "Error parsing JSON" error.

Is it possible to toggle logging from the console?

PR #10153, opened 2017-04-04
logging: Fix off-by-one for shrinkdebugfile default

maflcko:
Reference: 

```
src/qt/bitcoin-qt -?|grep -C 1 shrinkdebug

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1185,7 +1185,7 @@ bool AppInitMain(boost::thread_group& threadGroup, CScheduler& scheduler)
 #ifndef WIN32
     CreatePidFile(GetPidFile(), getpid());
 #endif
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/init.cpp [hunk 1]
```diff
@@ -1185,7 +1185,7 @@ bool AppInitMain(boost::thread_group& threadGroup, CScheduler& scheduler)
 #ifndef WIN32
     CreatePidFile(GetPidFile(), getpid());
 #endif
-    if (GetBoolArg("-shrinkdebugfile", logCategories != BCLog::NONE)) {
+    if (GetBoolArg("-shrinkdebugfile", logCategories == BCLog::NONE)) {
```
laanwj: I never realized that shrinkdebugfile depends on the debug logging.

laanwj:
utACK faab624

PR #10154, opened 2017-04-04
init: Remove redundant logging code

maflcko:
The help message for 'nodebug' is misleading at best, as the default for debug is `0`. Otherwise we could add help messages for 'nodisablewallet', ...

Furthermore, there can never be an option in the map that starts with 'no', so no need to bloat the code with a `GetBoolArg()` that is always false.


laanwj:
Good catch

jonasschnelli:
utACK faafa801e885ec20c8e4307f1dc1617ee0c04998

laanwj:
utACK faafa80

PR #10155, opened 2017-04-05
build: Deduplicate version numbers

laanwj:
Release process streamlining: This removes all the duplicate version numbers from the build and makes bumping the version number a matter of changing one place instead of five (e.g. https://github.com/bitcoin/bitcoin/commit/47d24c85b71b854939522238f8e44c891d8674af).

If it is preferred to keep the version information in the README's then those could be generated from .in as well. My reason for not having done this is because people might want to read those before running the build system.


MarcoFalke:
utACK 8b9911fe0181d07683ca37bf011d0feec9137ab3

src/clientversion.h
```diff
@@ -9,24 +9,9 @@
...
-
-//! Set to true for release, false for prerelease or test build
-#define CLIENT_VERSION_IS_RELEASE false
-
-/**
- * Copyright year (2009-this)
- * Todo: update this when changing our copyright comments in the source
- */
-#define COPYRIGHT_YEAR 2017
+#if !defined(CLIENT_VERSION_MAJOR) || !defined(CLIENT_VERSION_MINOR) || !defined(CLIENT_VERSION_REVISION) || !defined(CLIENT_VERSION_BUILD) || !defined(CLIENT_VERSION_IS_RELEASE) || !defined(COPYRIGHT_YEAR)
```
laanwj: It might make sense to factor this out of the `HAVE_CONFIG_H` else, as to catch the case where config.h (for some reason) doesn't define the full set of client version information.

JeremyRubin: Yes I think that would be useful in case a builder modifies their config inappropriately.

doc/release-process.md
```diff
@@ -24,6 +24,7 @@ Before every major release:
 * Update hardcoded [seeds](/contrib/seeds/README.md), see [this pull request](https://github.com/bitcoin/bitcoin/pull/7415) for an example.
 * Update [`BLOCK_CHAIN_SIZE`](/src/qt/intro.cpp) to the current size plus some overhead.
 * Update `src/chainparams.cpp` chainTxData with statistics about the transaction count and rate.
+* Update version of `contrib/gitian-descriptors/*.yml`: usually one'd want to do this on master after branching off the release - but be sure to at least do it before a new major release
```
JeremyRubin: Possible to update this text while you're editing it for less ambiguity? (maybe this should be a separate PR)
suggestion:
> Update version of `contrib/gitian-descriptors/*.yml` on master after branching off the release before any new major release. Updating for minor releases is optional

JeremyRubin (approved):
utAck 8b9911f

doc/release-process.md
laanwj: Thanks for trying to improve the wording. That's not correct, though. It should only be done for major releases, it's just that there is flexibility in the exact point in the cycle at which it is done.

src/clientversion.h
theuni: Agreed.

theuni (approved):
utACK 8b9911fe0181d07683ca37bf011d0feec9137ab3, or with the change suggested in e8eead11a48e6367fc1aad4fcd4c73ecccc2c49f. Having the doxyfile autogen'd seems helpful by itself.

fanquake:
utACK 8b9911f

laanwj:
Factored out the version information check in clientversion.h, and squashed.

PR #10156, opened 2017-04-05
Fix for issues with startup and multiple monitors on windows.

AllanDoensen:
I noticed a small bug in bitcoin-qt with multiple physical displays. This bug affects me because I use a Windows 10 laptop that sometimes uses a second display. 

I can only reproduce this issue on windows 10. I could not reproduce this issue on Ubuntu 16.04 LTS with dual displays running in a VirtualBox VM.

This is a pull request for a fix for this issue. 

To demonstrate the bug:
1. Install Bitcoin-qt on a system with multiple monitors (i.e. a laptop attached to a second screen). I have only reproduced this issue on windows 10 with ‘extend these displays’ configured.   
2. Start bitcoin-qt and place it on the second screen. 
3. Close bitcoin-qt (it will cache it’s position on the second screen).
4. Remove (physically unplug) the second screen.
5. Start bitcoin-qt.
6. You will find that bitcoin-qt starts in the cached second screen position and cannot be found on the only active screen. It just disappears. You need to realise that it is off screen and then do some fancy control keys to get it to come back to the primary display.   

This is not a critical bug, but when it happens the user experience is not good, hence this fix. 

As for the code changes….The position of bitcoin-qt is saved into QSettings on exit. When bitcoin-qt restarts it grabs this position and places itself at that location. However it never tests if this position actually still exists. So I have added a test that if the location does not exist on any of the current screens then it re-positions it to the default location. 

I also use abs() when calculating the center position. This is because there is the potential for the size of the application to be larger then the size of the current screen, abs() should handle that corner case. 

I have tested this change on Windows 10 & Ubuntu 16.04 LTS.


jonasschnelli:
Makes sense.
Code looks good (re-sets the Widget to main screen center if the larger portion of Widget is not on a screen).
utACK 18968187da375b1a71af102c13b0a9e5d7ea65ef

laanwj:
This doesn't just affect windows but any setup with variable number monitors. Thanks for the fix.
(I think there's an reported issue for this open somewhere but can't find it)

sipa:
Does this apply to 0.14?

AllanDoensen:
@sipa This applies to 0.14. This bug has existed for many years and is in all bitcoin & alt-coins that I am aware of.

laanwj:
> This is because there is the potential for the size of the application to be larger then the size of the current screen, abs() should handle that corner case.

Thinking of it, I don't think abs fixes this problem. 
E.g. say the screen is 10x10, and the window 20x20. With the code without abs, the position of the window will be ((10-20)/2, (10-20)/2) or (-5,-5): screen will be centered in the center of the window instead of the other way around.
```
+-----------------|
| window          |
| +------------+  |
| |            |  |
| |  screen    |  |
| |            |  |
| +------------+  |
|                 |
+-----------------|
```
This is not useful, as agreed. However abs would cause the window to be at (5,5), which reflects the coordinates over both axes and causes something like
```
+------------+
| screen     |
|   +-----------------+
|   |  window|        |
+---|--------+        | 
    |                 |
    |                 |
    +-----------------+
```
I'm not sure that makes sense. What would be the reasoning to put it there? A part of the window will still be outside the screen.

If the size of the application is larger than the size of the screen, wouldn't it be better to resize the application to fullscreen? Or alternatively, just ignore the hint, and leave the application at the default size.

AllanDoensen:
@laanwj The screen is placed in a location that the user can easily grab the top & edges. This allows the user to choose a new size.  If you do not do the abs() then the user does not have access to a single edge. The user cannot cannot easily grab the application and re-size/move it...hence the move and the abs() . 

While we could also resize the application to fit, I think if you do that there are many other considerations and the code starts to get messy. I did contemplate it, but this is really an extreme corner case we are talking about here. 

The important point is that the user has a means of easily interacting with the application when it starts. By placing the application top right comer on the screen guarantees that the user can move & resize that application. Without the abs() that becomes much harder because the user has no edge to select and resize.

laanwj:
> The screen is placed in a location that the user can easily grab the top & edges.

That's not guaranteed to be the case. The window could be twice the size of the screen, for example.

> While we could also resize the application to fit, I think if you do that there are many other considerations and the code starts to get messy. I did contemplate it, but this is really an extreme corner case we are talking about here.

What about my proposal to just go to default settings in that case? I think that is the best fallback if the settings in the QSettings don't make sense (sometimes they get corrupted too), just ignore them.

AllanDoensen:
Yes, that is acceptable.

AllanDoensen:
Should I update it or do you other want the say first?

laanwj:
That's up to you. To be clear I'm fine with the fix as it is now, as at least it improves things compared to how it is now, but I just think it could be even more robust.

laanwj:
Looks good to me now, utACK after squash.

luke-jr:
Seems the right behaviour here is to let the WM give it the default position. (Arguably we shouldn't be saving it at all, but that's another issue.)

laanwj:
Yeah yeah, let's just agree that this is an improvement over not being able to find the window.
Whether to store the window position at all is a completely different discussion that doesn't belong here.

jonasschnelli:
@AllanDoensen: commit history is messed up. Maybe do a `git fetch origin; git reset --hard origin/master` and cherry-pick the relevant commit(s).

laanwj:
Oops. I think I have the previous version of this saved somewhere.

laanwj:
Merged via 51833a1 (retrieved the previous good version and squashed)

Issue #9836, opened 2017-02-23
Tor Single Onion Service support


keystrike:
When enabled, this will allow the bitcoin client to set up its hidden service as a [Single Onion Service](https://blog.torproject.org/blog/whats-new-tor-0298).  This will improve the speed of the service at the expense of its anonymity by removing the three hops between the service and its Rendezvous Point. Client anonymity will not be affected by this change.

gmaxwell:
We should be careful with this because many users actually intend to get anonymity.  Perhaps it would be correct to use this when we are also taking v4 connections from the open internet, since in the case the anonymity is compromised by the other connections.

keystrike:
Yes, optional and default to off to retain the status quo is the good choice. There is the consideration that v4+onion can degrade anonymity due correlation, although there is currently no warning if serving on both networks. In that case it wouldn't matter if this was on, but since users are most familiar with the standard onion hidden service then the expected behavior should remain as is. The option can exist for those who want to run open nodes on v4+onion and make the best use of their bandwidth in supporting the p2p network.

laanwj:
Any idea how to request this from the Tor control socket?

keystrike:
[ADD_ONION](https://github.com/torproject/torspec/blob/master/control-spec.txt) takes flags=NonAnonymous.

laanwj:
@keystrike Thanks, that seems simple to add.

kewde:
While this is not a Bitcoin Core related issue, it's definitely worth noting this in the documentation of this feature.

> Tor instances can either be in anonymous hidden service mode, or
  non-anonymous single onion service mode. All hidden services on the same
  tor instance have the same anonymity. To guard against unexpected loss
  of anonymity, Tor checks that the ADD_ONION "NonAnonymous" flag matches
  the current hidden service anonymity mode. The hidden service anonymity
  mode is configured using the Tor options HiddenServiceSingleHopMode and
  HiddenServiceNonAnonymousMode. If both these options are 1, the
  "NonAnonymous" flag must be provided to ADD_ONION. If both these options
  are 0 (the Tor default), the flag must NOT be provided.

https://github.com/torproject/torspec/blob/master/control-spec.txt#L1516

[...]

PR #10161, opened 2017-04-06
Support for Tor's Single Onion Service

rodentrabies:
Implements #9836 

- [x] Add flag to *ADD_ONION* command
- [x] Provide a way for external configuration

### DIFF
--- a/doc/tor.md
+++ b/doc/tor.md
@@ -102,7 +102,9 @@ affect the number of available .onion nodes.
 This new feature is enabled by default if Bitcoin Core is listening (`-listen`), and
 requires a Tor connection to work. It can be explicitly disabled with `-listenonion=0`
 and, if not disabled, configured using the `-torcontrol` and `-torpassword` settings.
-To show verbose debugging information, pass `-debug=tor`.
+To show verbose debugging information, pass `-debug=tor`. Hidden service can also be
+run in Single Onion Service mode which was introduced in Tor version 0.2.9.8 and exchanges
+service anonymity for connection speed up. This can be enabled with `-listenonion=2`.
 
 Connecting to Tor's control socket API requires one of two authentication methods to be 
 configured. For cookie authentication the user running bitcoind must have write access 
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -381,7 +381,7 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-externalip=<ip>", _("Specify your own public address"));
     strUsage += HelpMessageOpt("-forcednsseed", strprintf(_("Always query for peer addresses via DNS lookup (default: %u)"), DEFAULT_FORCEDNSSEED));
     strUsage += HelpMessageOpt("-listen", _("Accept connections from outside (default: 1 if no -proxy or -connect)"));
-    strUsage += HelpMessageOpt("-listenonion", strprintf(_("Automatically create Tor hidden service (default: %d)"), DEFAULT_LISTEN_ONION));
+    strUsage += HelpMessageOpt("-listenonion", strprintf(_("Run Tor hidden service (0 - disabled, 1 - normal mode, 2 - single onion service, default: %s)"), DEFAULT_LISTEN_ONION));
     strUsage += HelpMessageOpt("-maxconnections=<n>", strprintf(_("Maintain at most <n> connections to peers (default: %u)"), DEFAULT_MAX_PEER_CONNECTIONS));
     strUsage += HelpMessageOpt("-maxreceivebuffer=<n>", strprintf(_("Maximum per-connection receive buffer, <n>*1000 bytes (default: %u)"), DEFAULT_MAXRECEIVEBUFFER));
     strUsage += HelpMessageOpt("-maxsendbuffer=<n>", strprintf(_("Maximum per-connection send buffer, <n>*1000 bytes (default: %u)"), DEFAULT_MAXSENDBUFFER));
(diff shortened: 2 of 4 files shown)


src/torcontrol.cpp
```diff
@@ -372,6 +372,8 @@ class TorController
     bool reconnect;
     struct event *reconnect_ev;
     float reconnect_timeout;
+    /** Flag for Single Onion Service */
+    bool non_anonymous;
```
laanwj: +1 on calling this flag non_anonymous

src/torcontrol.cpp
```diff
@@ -414,6 +416,9 @@ TorController::TorController(struct event_base* _base, const std::string& _targe
         LogPrint(BCLog::TOR, "tor: Reading cached private key from %s\n", GetPrivateKeyFile());
         private_key = pkf.second;
     }
+    // Set Single Onion Service flag
+    // TODO: provide a way to configure it externally
+    non_anonymous = false;
```
laanwj: As for option: maybe `-listenonion=2`?

mrwhythat: Yes, thanks for the advice. That should be OK. Will finish it this evening.

src/torcontrol.cpp
```diff
@@ -414,6 +416,12 @@ TorController::TorController(struct event_base* _base, const std::string& _targe
         LogPrint(BCLog::TOR, "tor: Reading cached private key from %s\n", GetPrivateKeyFile());
         private_key = pkf.second;
     }
+    // Set Single Onion Service flag
+    if (mode == 1)
```
mrwhythat: Having `-listenonion=<n>` flag requires allowed values checks and isn't obvious enough as for anonymous/non-anonymous service option. Maybe it would be better to just add another boolean flag instead, with some obvious name?

laanwj: Yeah could do that. Though I think the number of options of bitcoind is kind of getting out of hand, and this seems quite obscure so I don't mind hiding it a bit.

mrwhythat:
Added allowed values check and mentioned it in documentation.

mrwhythat:
I am unable to reproduce this failure on my machine. Any advise on what might cause it?

doc/tor.md
```diff
@@ -102,7 +102,9 @@ affect the number of available .onion nodes.
 This new feature is enabled by default if Bitcoin Core is listening (`-listen`), and
 requires a Tor connection to work. It can be explicitly disabled with `-listenonion=0`
 and, if not disabled, configured using the `-torcontrol` and `-torpassword` settings.
-To show verbose debugging information, pass `-debug=tor`.
+To show verbose debugging information, pass `-debug=tor`. Hidded service can also be
```
kewde: language: Hidded -> Hidden

src/init.cpp
```diff
@@ -381,7 +381,7 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-externalip=<ip>", _("Specify your own public address"));
     strUsage += HelpMessageOpt("-forcednsseed", strprintf(_("Always query for peer addresses via DNS lookup (default: %u)"), DEFAULT_FORCEDNSSEED));
     strUsage += HelpMessageOpt("-listen", _("Accept connections from outside (default: 1 if no -proxy or -connect)"));
-    strUsage += HelpMessageOpt("-listenonion", strprintf(_("Automatically create Tor hidden service (default: %d)"), DEFAULT_LISTEN_ONION));
+    strUsage += HelpMessageOpt("-listenonion", strprintf(_("Run Tor hidden service (0 - disabled, 1 - normal mode, 2 - single onion service, default: %d)"), DEFAULT_LISTEN_ONION));
```
kewde: I'm assuming the problem will be with %d and int64_t. I see you've changed DEFAULT_LISTEN_ONION from a bool to int64_t but did not change it here.

The format for int64_t (to (str)printf etc) is typically
- "I64d" (Win32/Win64)
-  "lld" (32-bit platforms) 
- "ld" (64-bit platforms) 

I suspect int64_t might not be the right type to use for DEFAULT_LISTEN_ONION.

mrwhythat: Oh, right, that was lame. Thanks for pointing it out.

doc/tor.md
mrwhythat: Thanks for a thorough review!

src/init.cpp
laanwj: We don't use the OS' printf function but tinyformat, so you must use %d for any kind of integer. It doesn't understand most of the size specifiers and ignores them as it's typesafe and regards the actual argument types.

src/torcontrol.h
```diff
@@ -11,7 +11,7 @@
 #include "scheduler.h"
 
 extern const std::string DEFAULT_TOR_CONTROL;
-static const bool DEFAULT_LISTEN_ONION = true;
+static const std::string DEFAULT_LISTEN_ONION = "0";
```
laanwj: String constant cannot be defined reliably in headers. It must be `extern` and defined in a c++ file. Otherwise it will get included into every single compilation unit that includes this header. Sometimes that runs into linker errors or causes executable bloat at the least.

src/torcontrol.cpp
```diff
@@ -395,7 +397,7 @@ class TorController
     static void reconnect_cb(evutil_socket_t fd, short what, void *arg);
 };
 
-TorController::TorController(struct event_base* _base, const std::string& _target):
+TorController::TorController(struct event_base* _base, const std::string& _target, const std::string& mode):
```
laanwj: Please make this interface take an `enum` mode, to make the available choices clear in the API.
Conversion between the argument and the enum should be done outside this class at the call site
(for example `TorControlThread()`)

laanwj:
Interesting discussion in #9836 as to whether this is a good idea.

src/torcontrol.cpp
mrwhythat: Thought of `enum` in the first place, but then Ocham's razor principle, don't multiply entities etc:) Anyway, done.

src/torcontrol.h
mrwhythat: Yes thanks, my bad. It's been a long time without c++ preprocessor.

mrwhythat:
@laanwj , I applied changes according to both of your notes, but considering discussion in #9836 , I should probably close this PR. I'll leave the branch anyway, in case we need a more robust and general solution in future.

laanwj:
@mrwhythat Agreed. And sure, makes sense to keep the branch, maybe someone wants to add this patch manually, doesn't hurt to have the code available.

PR #10165, opened 2017-04-07
[Wallet] Refactoring by using CInputCoin instead of std::pair

NicolasDorier:
The code was using `std::pair<const CWalletTx*, unsigned int>` until now. Not only it is ugly to read, and easy to mess up, but it needlessly pass `CWalletTx` all around the code when it is not needed.

This code will simplify another PR extending `FundRawTransaction`. (https://github.com/bitcoin/bitcoin/pull/10068)

This is divided in 3 commits easy to review.

1. Just rename `std::pair<const CWalletTx*, unsigned int>` to `CInputCoin` by using a type alias
2. Transform `CInputCoin` into a class.
3. Rewrite some code which can be simplified thanks to `CInputCoin`

### DIFF
--- a/src/wallet/test/wallet_tests.cpp
+++ b/src/wallet/test/wallet_tests.cpp
@@ -29,7 +29,7 @@ extern UniValue importmulti(const JSONRPCRequest& request);
 
 std::vector<std::unique_ptr<CWalletTx>> wtxn;
 
-typedef std::set<std::pair<const CWalletTx*,unsigned int> > CoinSet;
+typedef std::set<CInputCoin> CoinSet;
 
 BOOST_FIXTURE_TEST_SUITE(wallet_tests, WalletTestingSetup)
 
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -65,8 +65,8 @@ const uint256 CMerkleTx::ABANDON_HASH(uint256S("00000000000000000000000000000000
 
 struct CompareValueOnly
 {
-    bool operator()(const std::pair<CAmount, std::pair<const CWalletTx*, unsigned int> >& t1,
-                    const std::pair<CAmount, std::pair<const CWalletTx*, unsigned int> >& t2) const
+    bool operator()(const std::pair<CAmount, CInputCoin>& t1,
+                    const std::pair<CAmount, CInputCoin>& t2) const
     {
         return t1.first < t2.first;
     }
@@ -2032,7 +2032,7 @@ void CWallet::AvailableCoins(std::vector<COutput>& vCoins, bool fOnlySafe, const
     }
 }
 
-static void ApproximateBestSubset(const std::vector<std::pair<CAmount, std::pair<const CWalletTx*,unsigned int> > >& vValue, const CAmount& nTotalLower, const CAmount& nTargetValue,
+static void ApproximateBestSubset(const std::vector<std::pair<CAmount, CInputCoin> >& vValue, const CAmount& nTotalLower, const CAmount& nTargetValue,
                                   std::vector<char>& vfBest, CAmount& nBest, int iterations = 1000)
 {
     std::vector<char> vfIncluded;
@@ -2079,16 +2079,16 @@ static void ApproximateBestSubset(const std::vector<std::pair<CAmount, std::pair
 }
 
 bool CWallet::SelectCoinsMinConf(const CAmount& nTargetValue, const int nConfMine, const int nConfTheirs, const uint64_t nMaxAncestors, std::vector<COutput> vCoins,
-                                 std::set<std::pair<const CWalletTx*,unsigned int> >& setCoinsRet, CAmount& nValueRet) const
+                                 std::set<CInputCoin>& setCoinsRet, CAmount& nValueRet) const
 {
     setCoinsRet.clear();
     nValueRet = 0;
 
     // List of values less than target
-    std::pair<CAmount, std::pair<const CWalletTx*,unsigned int> > coinLowestLarger;
+    std::pair<CAmount, CInputCoin> coinLowestLarger;
     coinLowestLarger.first = std::numeric_limits<CAmount>::max();
     coinLowestLarger.second.first = NULL;
-    std::vector<std::pair<CAmount, std::pair<const CWalletTx*,unsigned int> > > vValue;
+    std::vector<std::pair<CAmount, CInputCoin> > vValue;
     CAmount nTotalLower = 0;
 
     random_shuffle(vCoins.begin(), vCoins.end(), GetRandInt);
@@ -2109,7 +2109,7 @@ bool CWallet::SelectCoinsMinConf(const CAmount& nTargetValue, const int nConfMin
         int i = output.i;
         CAmount n = pcoin->tx->vout[i].nValue;
 
-        std::pair<CAmount,std::pair<const CWalletTx*,unsigned int> > coin = std::make_pair(n,std::make_pair(pcoin, i));
+        std::pair<CAmount,CInputCoin> coin = std::make_pair(n,std::make_pair(pcoin, i));
 
         if (n == nTargetValue)
         {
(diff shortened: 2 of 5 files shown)


laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/10165/commits/2f5eb873dda9d1561c233a2313322a53afea5161

src/wallet/wallet.h
```diff
@@ -475,7 +475,38 @@ class CWalletTx : public CMerkleTx
 };
 
 
-using CInputCoin = std::pair<const CWalletTx*, unsigned int>;
+class CInputCoin {
+public:
+    CInputCoin()
```
ryanofsky: Would suggest deleting default constructor and IsNull method. Their presence seem to make this class more complex and precarious than is necessary, and could lead to bugs and unintended behavior.

The only place where a nullable `CInputCoin` seems to be used is in one `coinLowestLarger` local variable, and you could more easily declare this variable as `boost::optional<CInputCoin> coinLowestLarger;` (or [std::optional](http://en.cppreference.com/w/cpp/utility/optional) in the future) instead of building in a layer of optionality into the type.

ryanofsky (commented):
utACK 2f5eb873dda9d1561c233a2313322a53afea5161, nice cleanup

src/wallet/wallet.h
NicolasDorier: I don't think using `optional` would make the code more clear. I never saw it used anywhere in bitcoin.
What about using `CInputCoin*` for coinLowestLarger instead?

ryanofsky: > I don't think using optional would make the code more clear.

I think optional's useful. It saves work of having to build optionality into types you define, and more importantly it allows the same type to be optional in some contexts and required in others, which I think is exactly what you want here. std::optional is part of c++17 so eventually it will start becoming more commonly used.

Anyway, this was just a suggestion, so feel free to ignore it if you have a different preference.

> What about using CInputCoin* for coinLowestLarger instead?

Using a pointer was actually going to be my first suggestion, but I didn't see a way to make that work because coinLowestLarger is assigned from a local variable. But I think this would be an improvement if it could work.

NicolasDorier:
I needed to rebase, I ended up using `boost::optional` and getting ride of `IsNull()`.

btcdrak:
utACK 7e0b365

laanwj:
> I needed to rebase, I ended up using boost::optional 

Was about to ask "why not std::optional" but apparently that's C++17. Well, easy to replace when we adapt that - the concept is great.

src/wallet/wallet.h
```diff
@@ -475,7 +475,7 @@ class CWalletTx : public CMerkleTx
 };
 
 
-
+using CInputCoin = std::pair<const CWalletTx*, unsigned int>;
```
JeremyRubin: Prefer to use a typedef (I actually prefer using personally, but some devs are unfamiliar with it and you don't need it as CInputCoin is not templated)

Also, prefer a non-Hungarian name if you're introducing a new class.
See https://github.com/bitcoin/bitcoin/pull/9472#discussion_r95775684 -- Hungarians for new code discouraged.

src/wallet/wallet.cpp
```diff
@@ -2106,22 +2105,21 @@ bool CWallet::SelectCoinsMinConf(const CAmount& nTargetValue, const int nConfMin
...
         }
-        else if (n < nTargetValue + MIN_CHANGE)
+        else if (coin.txout.nValue < nTargetValue + MIN_CHANGE)
         {
             vValue.push_back(coin);
-            nTotalLower += n;
+            nTotalLower += coin.txout.nValue;
         }
-        else if (n < coinLowestLarger.first)
+        else if (coinLowestLarger.IsNull() || coin.txout.nValue < coinLowestLarger.txout.nValue)
```
JeremyRubin: I'm a bit confused here -- isn't coinLowestLarger always null at this point?

src/wallet/wallet.h
```diff
@@ -475,7 +475,38 @@ class CWalletTx : public CMerkleTx
...
+        {
+            outpoint = COutPoint(walletTx->GetHash(), i);
+            txout = walletTx->tx->vout[i];
+        }
+    }
+    bool IsNull() const
+    {
+        return outpoint.IsNull() && txout.IsNull();
+    }
+    COutPoint outpoint;
```
JeremyRubin: prefer not to have fields intermixed with functions.

src/wallet/wallet.h
```diff
@@ -477,21 +477,21 @@ class CWalletTx : public CMerkleTx
...
 public:
-    CInputCoin()
-    {
-    }
     CInputCoin(const CWalletTx* walletTx, unsigned int i)
     {
-        if (walletTx != nullptr && i < walletTx->tx->vout.size())
+        if (walletTx == nullptr)
+            throw std::invalid_argument("walletTx should not be null");
+        if (i < walletTx->tx->vout.size())
```
JeremyRubin: Can you invert the logic here `if (!P) throw` is a bit easier to read.

Might also be good to make it a `const CWalletTx&` so as to make sure that de-referencability is checked everywhere that this is called.

JeremyRubin (commented):
Couple nits/questions (be sure to show outdated) but overall Concept Ack.

src/wallet/wallet.h
```diff
@@ -475,7 +475,38 @@ class CWalletTx : public CMerkleTx
...
+            txout = walletTx->tx->vout[i];
+        }
+        else
+        {
+            throw std::out_of_range("The output index is out of range");
+        }
+    }
+
+    COutPoint outpoint;
+    CTxOut txout;
```
bulldozer00: Why are these data members public and not private?

src/wallet/test/wallet_tests.cpp
```diff
@@ -29,7 +29,7 @@ extern UniValue importmulti(const JSONRPCRequest& request);
 
 std::vector<std::unique_ptr<CWalletTx>> wtxn;
 
-typedef std::set<std::pair<const CWalletTx*,unsigned int> > CoinSet;
+typedef std::set<CInputCoin> CoinSet;
```
bulldozer00: Think about replacing old style "typedef  X Y " with the more intuitive C++11 "using Y = X" style

NicolasDorier: better to do in a separate PR. I want to keep it focused on `CInputCoin`.

src/wallet/wallet.h
NicolasDorier: because I use them outside the class.

src/wallet/wallet.h
NicolasDorier: Ah indeed, C# habits... Just replaced by `if(!walletTx)`.

I would like to do `const CWalletTx&` in a separate PR, as it need more code change, and I need this PR for another one depending on it.

JeremyRubin: Ah, I was unclear -- I actually don't mind a `== nullptr` (although current state is cleaner), I was referring to the `throw out_of_range`. It's easier if a constructor is: `if (! preconditionA) throw A; if (!preconditionB) throw B; ....;` rather than `if (! preconditionA) throw A; if (preconditionB)...; else throw B;`

NicolasDorier: @JeremyRubin indeed, much clearer. I updated the PR.

NicolasDorier:
Travis fail unrelated to this PR, @laanwj can you kick the travis box one time to fix the issue ?

src/wallet/test/wallet_tests.cpp
laanwj: We don't use that style anywhere, introducing it here is not a good idea. I'd prefer to save that for new code.

src/wallet/wallet.h
```diff
@@ -475,7 +475,34 @@ class CWalletTx : public CMerkleTx
 };
 
 
-
+class CInputCoin {
```
laanwj: It is not visible on github but local diff shows me that line endings are weird here:
```diff
-using CInputCoin = std::pair<const CWalletTx*, unsigned int>;
+class CInputCoin {^M
+public:^M
```
^M is 13, so I suppose these are DOS line endings. Looks like it is only in `src/wallet/wallet.h`, and only the newly  introduced code.
suggest running:
```bash
$ dos2unix src/wallet/wallet.h
dos2unix: converting file src/wallet/wallet.h to Unix format ...
```

NicolasDorier: oops. Yes, developping on visual studio... Fixed.

PR #10166, opened 2017-04-07
Ignore Doxyfile generated from Doxyfile.in template.

paveljanik:
After #10155, `doc/Doxyfile` is generated by `configure`. git-ignore it.

### DIFF
--- a/doc/.gitignore
+++ b/doc/.gitignore
@@ -1,0 +1,1 @@
+Doxyfile


laanwj:
Yes makes sense, utACK https://github.com/bitcoin/bitcoin/pull/10166/commits/9eaf7189d80552130aa2c9981706180d29638bb9

PR #10169, opened 2017-04-07
[tests] Remove func test code duplication

jnewbery:
Test cases that use bitcoind's p2p interface use `NodeConnCB` or a subclass to access the p2p interface. A lot of the test cases use a subclass  of `NodeConnCB` and implement their own callbacks or helper methods. In many cases those local implementations are duplicates.

This PR adds sensible callbacks and helper functions to the base NodeConnCB class, and removes the subclass override methods. Some tests still require custom behaviour and so need to define or override some of the class methods.

The first commit adds the sensible default callbacks and helper methods to NodeConnCB. The second commit removes any duplicate implementations in the individual testcsases.

Net code change is ~ minus 300 lines

### DIFF
--- a/test/functional/maxuploadtarget.py
+++ b/test/functional/maxuploadtarget.py
@@ -10,54 +10,24 @@ if uploadtarget has been reached.
 if uploadtarget has been reached.
 * Verify that the upload counters are reset after 24 hours.
 """
+from collections import defaultdict
+import time
 
 from test_framework.mininode import *
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import *
-import time
 
-# TestNode: bare-bones "peer".  Used mostly as a conduit for a test to sending
-# p2p messages to a node, generating the messages in the main testing logic.
 class TestNode(NodeConnCB):
     def __init__(self):
         super().__init__()
-        self.connection = None
-        self.ping_counter = 1
-        self.last_pong = msg_pong()
-        self.block_receive_map = {}
-
-    def add_connection(self, conn):
-        self.connection = conn
-        self.peer_disconnected = False
+        self.block_receive_map = defaultdict(int)
 
     def on_inv(self, conn, message):
         pass
 
-    # Track the last getdata message we receive (used in the test)
-    def on_getdata(self, conn, message):
-        self.last_getdata = message
-
     def on_block(self, conn, message):
         message.block.calc_sha256()
-        try:
-            self.block_receive_map[message.block.sha256] += 1
-        except KeyError as e:
-            self.block_receive_map[message.block.sha256] = 1
-
-    def wait_for_disconnect(self):
-        def disconnected():
-            return self.peer_disconnected
-        return wait_until(disconnected, timeout=10)
-
-    # Wrapper for the NodeConn's send_message function
-    def send_message(self, message):
-        self.connection.send_message(message)
-
-    def on_pong(self, conn, message):
-        self.last_pong = message
-
-    def on_close(self, conn):
-        self.peer_disconnected = True
+        self.block_receive_map[message.block.sha256] += 1
 
 class MaxUploadTest(BitcoinTestFramework):
  
@@ -183,33 +153,26 @@ class MaxUploadTest(BitcoinTestFramework):
         stop_node(self.nodes[0], 0)
         self.nodes[0] = start_node(0, self.options.tmpdir, ["-whitelist=127.0.0.1", "-maxuploadtarget=1", "-blockmaxsize=999000"])
 
-        #recreate/reconnect 3 test nodes
-        test_nodes = []
-        connections = []
-
-        for i in range(3):
-            test_nodes.append(TestNode())
-            connections.append(NodeConn('127.0.0.1', p2p_port(0), self.nodes[0], test_nodes[i]))
-            test_nodes[i].add_connection(connections[i])
+        #recreate/reconnect a test node
+        test_nodes = [TestNode()]
+        connections = [NodeConn('127.0.0.1', p2p_port(0), self.nodes[0], test_nodes[0])]
+        test_nodes[0].add_connection(connections[0])
 
         NetworkThread().start() # Start up network handling in another thread
-        [x.wait_for_verack() for x in test_nodes]
+        test_nodes[0].wait_for_verack()
 
         #retrieve 20 blocks which should be enough to break the 1MB limit
         getdata_request.inv = [CInv(2, big_new_block)]
         for i in range(20):
-            test_nodes[1].send_message(getdata_request)
-            test_nodes[1].sync_with_ping()
-            assert_equal(test_nodes[1].block_receive_map[big_new_block], i+1)
+            test_nodes[0].send_message(getdata_request)
+            test_nodes[0].sync_with_ping()
+            assert_equal(test_nodes[0].block_receive_map[big_new_block], i+1)
 
         getdata_request.inv = [CInv(2, big_old_block)]
-        test_nodes[1].send_message(getdata_request)
-        test_nodes[1].wait_for_disconnect()
-        assert_equal(len(self.nodes[0].getpeerinfo()), 3) #node is still connected because of the whitelist
+        test_nodes[0].send_and_ping(getdata_request)
+        assert_equal(len(self.nodes[0].getpeerinfo()), 1) #node is still connected because of the whitelist
 
-        self.log.info("Peer 1 still connected after trying to download old block (whitelisted)")
-
-        [c.disconnect_node() for c in connections]
+        self.log.info("Peer still connected after trying to download old block (whitelisted)")
 
 if __name__ == '__main__':
     MaxUploadTest().main()
--- a/test/functional/maxuploadtarget.py
+++ b/test/functional/maxuploadtarget.py
@@ -44,15 +44,6 @@ class TestNode(NodeConnCB):
         except KeyError as e:
             self.block_receive_map[message.block.sha256] = 1
 
-    # Spin until verack message is received from the node.
-    # We use this to signal that our test can begin. This
-    # is called from the testing thread, so it needs to acquire
-    # the global lock.
-    def wait_for_verack(self):
-        def veracked():
-            return self.verack_received
-        return wait_until(veracked, timeout=10)
-
     def wait_for_disconnect(self):
         def disconnected():
             return self.peer_disconnected
--- a/test/functional/p2p-compactblocks.py
+++ b/test/functional/p2p-compactblocks.py
@@ -19,64 +19,31 @@ class TestNode(NodeConnCB):
     def __init__(self):
         super().__init__()
         self.last_sendcmpct = []
-        self.last_headers = None
-        self.last_inv = None
-        self.last_cmpctblock = None
         self.block_announced = False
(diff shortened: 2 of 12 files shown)


test/functional/test_framework/mininode.py
```diff
@@ -1548,44 +1579,69 @@ def on_version(self, conn, message):
...
         self.connection = conn
 
-    # Wrapper for the NodeConn's send_message function
+    def wait_for_disconnect(self, timeout=60):
+        test_function = lambda: not self.connected
+        assert wait_until(test_function, timeout=timeout)
+
+    # Message receiving helper methods
+
+    def sync(self, test_function, timeout=60):
```
jimmysong: This function seems to duplicate a lot of what `wait_until` above does. Maybe it makes sense to do something like:

    def sync(self, test_function, timeout=60):
        if not wait_until(test_function, timeout=timeout):
            AssertionError("Sync failed to complete")

test/functional/test_framework/mininode.py
```diff
@@ -1548,44 +1579,69 @@ def on_version(self, conn, message):
...
-    ##################
+    # Connection helper methods
 
     def add_connection(self, conn):
         self.connection = conn
 
-    # Wrapper for the NodeConn's send_message function
+    def wait_for_disconnect(self, timeout=60):
+        test_function = lambda: not self.connected
+        assert wait_until(test_function, timeout=timeout)
```
jimmysong: This seems like it should also use `sync`:

    self.sync(test_function, timeout)

test/functional/test_framework/mininode.py
```diff
@@ -1548,44 +1579,69 @@ def on_version(self, conn, message):
...
         self.sync_with_ping()
 
     # Sync up with the node
     def sync_with_ping(self, timeout=60):
-        def received_pong():
-            return (self.last_pong.nonce == self.ping_counter)
         self.send_message(msg_ping(nonce=self.ping_counter))
-        success = wait_until(received_pong, timeout=timeout)
+        test_function = lambda: self.last_message.get("pong") and self.last_message["pong"].nonce == self.ping_counter
+        success = wait_until(test_function, timeout = timeout)
```
jimmysong: Also looks like it should use `self.sync` instead and maybe get rid of the `if not success` below.

test/functional/test_framework/comptool.py
```diff
@@ -192,9 +192,7 @@ def disconnected():
         return wait_until(disconnected, timeout=10)
 
     def wait_for_verack(self):
-        def veracked():
-            return all(node.verack_received for node in self.test_nodes)
-        return wait_until(veracked, timeout=10)
+        [node.wait_for_verack() for node in self.test_nodes]
```
jimmysong: Any reason not to return the value like this?

    return all(node.wait_for_verack() for node in self.test_nodes)

test/functional/sendheaders.py
```diff
@@ -155,43 +137,27 @@ def check_last_announcement(self, headers=None, inv=None):
...
             return
 
-        test_function = lambda: self.last_getdata != None and [x.hash for x in self.last_getdata.inv] == hash_list
-        assert(wait_until(test_function, timeout=timeout))
-        return
-
-    def wait_for_disconnect(self, timeout=60):
-        test_function = lambda: self.disconnected
+        test_function = lambda: "getdata" in self.last_message and [x.hash for x in self.last_message["getdata"].inv] == hash_list
         assert(wait_until(test_function, timeout=timeout))
```
jimmysong: `self.sync` looks more appropriate here.

test/functional/sendheaders.py
```diff
@@ -581,25 +536,21 @@ def run_test(self):
...
             # Send a header that doesn't connect, check that we get a getheaders.
             with mininode_lock:
-                test_node.last_getheaders = None
+                test_node.last_message.pop("getheaders", None)
             test_node.send_header_for_blocks([blocks[i%len(blocks)]])
             test_node.wait_for_getheaders(timeout=1)
 
         # Eventually this stops working.
-        with mininode_lock:
-            self.last_getheaders = None
```
jimmysong: Any reason to get rid of this and the one 5 lines below? I expected `self.last_message.pop`.

test/functional/p2p-segwit.py
```diff
@@ -35,79 +35,21 @@ def get_virtual_size(witness_block):
...
+            self.last_message.pop("getdata", None)
         self.send_message(msg_inv(inv=[CInv(1, tx.sha256)]))
         self.wait_for_getdata(timeout)
-        return
 
     def announce_block_and_wait_for_getdata(self, block, use_header, timeout=60):
         with mininode_lock:
-            self.last_getdata = None
-            self.last_getheaders = None
+            self.last_message.pop("getdata", None)
```
jimmysong: how about `getheaders`?

test/functional/p2p-leaktests.py
```diff
@@ -121,7 +112,9 @@ def run_test(self):
 
         NetworkThread().start()  # Start up network handling in another thread
 
-        assert(wait_until(lambda: no_version_bannode.connected and no_version_idlenode.connected and no_verack_idlenode.version_received, timeout=10))
+        assert wait_until(lambda: no_version_bannode.ever_connected, timeout=10)
+        assert wait_until(lambda: no_version_idlenode.ever_connected, timeout=10)
+        assert wait_until(lambda: no_verack_idlenode.version_received, timeout=10)
```
jimmysong: `self.sync`?

test/functional/p2p-feefilter.py
```diff
@@ -78,7 +72,7 @@ def run_test(self):
         test_node.clear_invs()
 
         # Set a filter of 15 sat/byte
-        test_node.send_filter(15000)
+        test_node.send_and_ping(msg_feefilter(15001))
```
jimmysong: Why did the constant change from 15000 to 15001?

test/functional/p2p-compactblocks.py
```diff
@@ -43,51 +35,33 @@ def on_close(self, conn):
...
-    def on_block(self, conn, message):
-        self.last_block = message
-
     def on_cmpctblock(self, conn, message):
-        self.last_cmpctblock = message
         self.block_announced = True
-        self.last_cmpctblock.header_and_shortids.header.calc_sha256()
-        self.set_announced_blockhashes.add(self.last_cmpctblock.header_and_shortids.header.sha256)
+        self.last_message["cmpctblock"].header_and_shortids.header.calc_sha256()
+        self.set_announced_blockhashes.add(self.last_message["cmpctblock"].header_and_shortids.header.sha256)
```
jimmysong: Please let me know if this is inappropriate, but the object attribute `set_announced_blockhashes` is a bit confusing. Would changing the name to something like `announced_blockhashes_set` be appropriate here?

test/functional/p2p-compactblocks.py
```diff
@@ -103,8 +77,7 @@ def send_header_for_blocks(self, new_blocks):
     def request_headers_and_sync(self, locator, hashstop=0):
         self.clear_block_announcement()
         self.get_headers(locator, hashstop)
-        assert(wait_until(self.received_block_announcement, timeout=30))
-        assert(self.received_block_announcement())
+        assert wait_until(self.received_block_announcement, timeout=30)
```
jimmysong: `self.sync`?

test/functional/p2p-compactblocks.py
```diff
@@ -214,14 +187,14 @@ def check_announcement_of_new_block(node, peer, predicate):
             with mininode_lock:
                 assert predicate(peer), (
                     "block_hash={!r}, cmpctblock={!r}, inv={!r}".format(
-                        block_hash, peer.last_cmpctblock, peer.last_inv))
+                        block_hash, peer.last_message.get("cmpctblock", None), peer.last_message.pop("inv", None)))
```
jimmysong: Why `pop` instead of `get` for the inv message?

test/functional/maxuploadtarget.py
```diff
@@ -192,33 +153,26 @@ def run_test(self):
...
         self.nodes[0] = start_node(0, self.options.tmpdir, ["-whitelist=127.0.0.1", "-maxuploadtarget=1", "-blockmaxsize=999000"])
 
-        #recreate/reconnect 3 test nodes
-        test_nodes = []
-        connections = []
-
-        for i in range(3):
-            test_nodes.append(TestNode())
-            connections.append(NodeConn('127.0.0.1', p2p_port(0), self.nodes[0], test_nodes[i]))
-            test_nodes[i].add_connection(connections[i])
```
jimmysong: Am I correct in assuming you're creating 1 node instead of 3 because the other two aren't being used?

jimmysong (commented):
ACK e74cf5028bbe6bc7fe520647860590f855eb8b3e

This is my first review, so please feel free to let me know if I'm doing something wrong. Overall, the storing of the last network messages in `NodeConnCB.last_message` makes everything a lot simpler. I have tested the code by running the tests locally, and everything seems to work. There are a few trivial issues that I've commented inline.

jnewbery:
Thanks for the excellent review @jimmysong . I've responded to most of your points. I'm just running the extended test suite locally before pushing up to github.

I've decided to remove the `sync()` method from NodeConnCB entirely for now, since it's just duplicating the functionality of `wait_until()` and doesn't actually need to be a method. I'm still not entirely happy with `wait_until()`, but I think it's better to not duplicate code for no good reason in this PR.

jnewbery:
rebased, review comments addressed and pushed

test/functional/sendheaders.py
jimmysong: makes sense.

test/functional/p2p-feefilter.py
jimmysong: I don't see this yet? Maybe missing a changeset or I'm missing something?

jimmysong (commented):
utACK

I'll test after the 15001 is addressed.

test/functional/p2p-feefilter.py
jnewbery: argh! Fixed now

jimmysong:
ACK 2a52ae63bfdde948250df1c876dcdd5af99f03b5

jimmysong (approved):

test/functional/test_framework/mininode.py
```diff
@@ -1548,43 +1579,56 @@ def on_version(self, conn, message):
...
+    def wait_for_getdata(self, timeout=60):
+        test_function = lambda: self.last_message.get("getdata")
+        assert wait_until(test_function, timeout=timeout)
+
+    def wait_for_getheaders(self, timeout=60):
+        test_function = lambda: self.last_message.get("getheaders")
+        assert wait_until(test_function, timeout=timeout)
+
+    def wait_for_inv(self, expected_inv, timeout=60):
+        test_function = lambda: self.last_message.get("inv") and self.last_message["inv"] != expected_inv
```
MarcoFalke: @jnewbery Can you explain why the second operand is `!=` and not `==`?

jnewbery: This was a code move from p2p-segwit.py to here, but you're right that this is a bug.

Fix here: #10318

Issue #10174, opened 2017-04-10
Cursor bein' all, like: I ain't gotta be doin' shizzle that be makin' sense.


KibbledJiveElkZoo:
I am using Windows 8 and Bitcoin Core version 0.14.0.  In the operating system's settings for the mouse I have the "Snap To" option ("Automatically move pointer to the default button in a dialog box.") enabled.

When I click on the "Send" tab at the top of Bitcoin Core the cursor/pointer snaps to the send button (which I kind of think is a bit ill advised, possibly).  Then, when another tab is selected (either "Overview", "Receive" or "Transactions"), the cursor is moved to the place on the screen, the place in the program, where the Send tab's Send button was.  So you move from the Send tab to another tab, and just like when you selected the Send tab and the cursor was moved to where the Send button is, your cursor gets moved to that same spot, even though you just selected a different tab to become active.  O.o

I could be entirely wrong, but . . . me thinks this is what the experto-Bitcoiner-toshi types call: "a bug".

I am rather excited that I might be the first one to bitch about this because I want to be cited in the release notes as a contributor or advisor or "peon shit nothing" or whatever.  Please, some big huge important Bitcoin Core developer type person, agree with me that this is a bug and do all the work and fix it and cite me as being the most important person since Isaac Newton in some official press release or version release notes.  Thank you and good night.

laanwj:
Looks like the same issue as #4840. Bitcoin Core has no code for changing the cursor position as that is indeed terribly annoying behavior.

Apparently this is a windows feature:
```
<TheSeven> windows itself has such a "feature"
<TheSeven> you can enable it somewhere in the mouse settings
<TheSeven> if something opens a dialog, the cursor will automatically jump to the default button of that dialog
<TheSeven> IIUC that user has thar weird feature enabled, and is just complaining that it's somehow picking the wrong buttons to jump to
```

KibbledJiveElkZoo:
I am closing this "issue" because I was told to by my Oracle.

PR #10177, opened 2017-04-10
Changed "Send" button default status from true to false

KibbledJiveElkZoo:
Changed the "Send" button's default status from true to false to prevent
quirky Windows autofocus behavior.

### DIFF
--- a/src/qt/forms/sendcoinsdialog.ui
+++ b/src/qt/forms/sendcoinsdialog.ui
@@ -1235,7 +1235,7 @@
         <bool>false</bool>
        </property>
        <property name="default">
-        <bool>true</bool>
+        <bool>false</bool>
        </property>
       </widget>
      </item>


laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/10177/commits/8c3e6c6987f3f1d123201d401ace2de9970aa29d

KibbledJiveElkZoo:
This change prevents the cursor from being repositioned in undesired places as mentioned in #10174.

KibbledJiveElkZoo:
@laanwj Does my massive contribution warrant my inclusion in the 0.14.2 release "Credits"?

achow101:
@KibbledJiveElkZoo "massive" (not really). Your change will be included and credited in Bitcoin Core 0.15.0. It is not a big enough bug fix to be included in 0.14.2.

laanwj:
Everyone that contributes to a release warrants being included in the credits in the "contributed to this release directly" list, no exceptions.

But as @achow101 says this wasn't part of 0.14.2, so it won't be in those. Usually for the name we automatically use the name you used for the commit author, so "KibbledJiveElkZoo".

I guess we could consider this for 0.14.3. In any case it will be in 0.15.0.

KibbledJiveElkZoo:
I thank you both for the trouble of expending time babysitting my nearly equally massive vain insecurity.

PR #10182, opened 2017-04-11
[scheduler] Switched CScheduler to C++11 threading primitives

tjps:
Switched CScheduler off of Boost threading primitives to std::thread/condition_variable/chrono.

### DIFF
--- a/src/scheduler.cpp
+++ b/src/scheduler.cpp
@@ -7,8 +7,6 @@
 #include "reverselock.h"
 
 #include <assert.h>
-#include <boost/bind.hpp>
-#include <utility>
 
 CScheduler::CScheduler() : nThreadsServicingQueue(0), stopRequested(false), stopWhenEmpty(false)
 {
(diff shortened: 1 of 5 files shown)


src/scheduler.cpp
```diff
@@ -71,10 +49,11 @@ void CScheduler::serviceQueue()
             taskQueue.erase(taskQueue.begin());
 
             {
-                // Unlock before calling f, so it can reschedule itself or another task
+                // Unlock while calling f, so it can reschedule itself or another task
                 // without deadlocking:
-                reverse_lock<boost::unique_lock<boost::mutex> > rlock(lock);
+                lock.unlock();
                 f();
+                lock.lock();
```
dcousens: is `f` guaranteed not to throw?

laanwj:
>  I've been tempted down the reverse_lock RAII path many times before, but in my experience the potential usage points for such a class in a codebase are usually too few to justify its existence.

`reverse_lock` was introduced to fix an issue with the lock not getting released in case of exceptions, and you may have reintroduced that issue by effectively reverting the code to previous state. I'd prefer to just keep it. It works, there are tests.

tjps:
I realize that now based on @dcousens comment.  I'll add it back in.

src/scheduler.cpp
tjps: You make a good point, I totally missed the exception safety as a feature of reverse_lock. Added it back.

theuni:
@tjps Nice work.

Unfortunately though, this is a bit more complicated than just switching out the primitives.

boost::thread is interruptible, meaning that the scheduler thread as well as the workers will terminate soonafter a call to threadGroup.interrupt_all(). When switching to std::thread and std::condition_variable, we have to handle that manually instead. That's a good thing though, less magic going on behind the scenes!

So init.cpp will need to learn how to stop the workers and the handler gracefully before making the switch. The order is important as well: worker threads may have callbacks into objects that have been destructed at exit, so they should be interrupted early in the shutdown process.

tjps:
Sorry to let this slide - all valid points raised.  Scuttling this PR to cleanup the backlog.

PR #10183, opened 2017-04-11
[RPC] Don't default rescan on private/public key imports.

KibbledJiveElkZoo:
Don't default rescan on private/public key imports.

Turn the default rescan for importprivkey, importaddress, importpubkey,
importmulti to false to prevent accidental rescanning that is not
cancellable.

Manual rescan should be triggered either through -rescan when restarting
th Bitcoin Core program or manually by setting one of the parameters to
true at the end of batch import.

Accidental rescans should be avoided.

### DIFF
--- a/src/wallet/rpcdump.cpp
+++ b/src/wallet/rpcdump.cpp
@@ -1041,7 +1041,7 @@ UniValue importmulti(const JSONRPCRequest& mainRequest)
     const UniValue& requests = mainRequest.params[0];
 
     //Default options
-    bool fRescan = false;
+    bool fRescan = true;
 
     if (mainRequest.params.size() > 1) {
         const UniValue& options = mainRequest.params[1];
--- a/src/wallet/rpcdump.cpp
+++ b/src/wallet/rpcdump.cpp
@@ -110,7 +110,7 @@ UniValue importprivkey(const JSONRPCRequest& request)
         strLabel = request.params[1].get_str();
 
     // Whether to perform rescan after import
-    bool fRescan = true;
+    bool fRescan = false;
     if (request.params.size() > 2)
         fRescan = request.params[2].get_bool();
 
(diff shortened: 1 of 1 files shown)


kallewoof:
Existing software depending on the default value may stop working correctly with this change. Not sure I agree that an accidental rescan is worse than a missed rescan. Maybe the rescan process should be cancelable somehow?

MarcoFalke:
An alternative would be to get rid of the default and require rescan to be
set explicitly.

KibbledJiveElkZoo:
An accidental scan will lock the wallet thread and the RPC thread until the process is completed, what currently can be a multi-hour long process.  A non-cancelable process.  The only way it can be canceled is to forcefully interrupt the program which may cause data failure.

sipa:
I'd rather drop the ability to skip reindex entirely, and require a timestamp for all imports. That way there is no possibility for confusion, and in the common case (it's a recently created key you're importing) it doesn't take particularly long anyway.

I do agree we need to make reindexing abortable.

KibbledJiveElkZoo:
It is my perspective that an option to skip indexing for never-used newly imported keys should be available.

CWallet.ScanForWalletTransactions should support scanning within specified block height ranges.  It should also have a circuit breaker for aborting after a step is finished.

Another functionality I think would be highly useful is having an on-demand RPC rescanwallet command that supports start/stop and block height ranges.

My first focus was to bring up the issue of defaults and making sure they are sane and they don’t force a user to kill the process.

(The build fails because the test requires either default true for importmulti (accidental rescan) or explicitly setting rescan to true in the test.)

laanwj:
This is a long-existing API so there will be software relying on this. The new call `importmulti` was designed with the better idea that everything should be explicit, but retrofitting this to the old API is a pain.

So I would prefer a way to abort reindexing. That would directly address the problem. Even if the default is changed or removed it's easy enough to make a mistake.

sipa:
> It is my perspective that an option to skip indexing for never-used newly imported keys should be available.

Yes, importmulti with timestamp "now" does this.

KibbledJiveElkZoo:
I undid the change of the default for importmulti so that the pull will pass the test.

PR #10185, opened 2017-04-11
[0.14] Mention dbcache memory changes in release notes

sdaftuar:
This is my attempt to document the dbcache memory changes... ping @sipa @morcos

### DIFF
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -50,6 +50,24 @@ activate at a time when no segwit transactions could be included.
 Since many miners are now including the segwit commitment this concern
 no longer applies.
 
+UTXO memory accounting
+----------------------
+
+Memory usage for the UTXO cache is being calculated more accurately, so that
+the configured limit (`-dbcache`) will be respected when memory usage peaks
+during cache flushes.  The memory accounting in prior releases is estimated to
+only account for half the actual peak utilization.
+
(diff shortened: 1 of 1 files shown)


gmaxwell:
The defaults were resulting in >1200 MB peak w/ mempool sharing. Might want to specifically name "such as systems with 1GB or less memory".

laanwj:
>  and suggest the saving memory doc for more information

Yes linking to [reducing-bitcoind-memory-usage.md](https://gist.github.com/laanwj/efe29c7661ce9b6620a7) may make sense, although it's out of date again with regard to dbcache defaults.

sdaftuar:
@laanwj Perhaps we should just incorporate that gist into the `doc/` directory, and then others can share in the burden of keeping it up to date?

sdaftuar:
Updated

laanwj:
@sdaftuar maybe. The doc directory isn't the easiest place to do
collaborative editing either, though, and multiple versions of the document
will exist then (on branches/tags) while I've tried to cover a larger range
of versions in one. Maybe the bitcoin-core wiki? (The same we've used for
release note editing for 0.14.0)

gmaxwell:
ACK (and feel free to changing around the citation, no opinion there)

sipa:
ACK

PR #10187, opened 2017-04-11
tests: Fix test_runner return value in case of skipped test

laanwj:
Currently test_runner reports an error if a test case is skipped. This is not how it should be, only failed tests should cause it to fail.
(hopefully this fixes travis)

### DIFF
--- a/test/functional/test_runner.py
+++ b/test/functional/test_runner.py
@@ -293,7 +293,7 @@ def run_tests(test_list, src_dir, build_dir, exeext, jobs=1, enable_coverage=Fal
         logging.debug("Cleaning up coverage data")
         coverage.cleanup()
 
-    all_passed = all(map(lambda test_result: test_result.status == "Passed", test_results))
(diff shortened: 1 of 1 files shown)


MarcoFalke:
Thanks! utACK e96462fbece87448619ba23f4917de2bfc257ee6.

jnewbery:
tested ACK e96462fbece87448619ba23f4917de2bfc257ee6

nit: `was_successful` is a misnomer. Perhaps `did_not_fail` is better?

MarcoFalke:
Going to merge now to fix travis. We can figure out the naming issue the next time this file is touched.

laanwj:
> nit: was_successful is a misnomer. Perhaps did_not_fail is better?

I considered that but found it too literal. There may be more statuses in the future, some considered a failure some successful. Note that I didn't use "has_passed" either.

jtimon:
post-merge utACK e96462fbece87448619ba23f4917de2bfc257ee6

PR #10193, opened 2017-04-12
scripted-diff: Remove #include <boost/foreach.hpp>

jtimon:
EDIT: Originally this was only going to remove 'BOOST_FOREACH' and '#include <boost/foreach.hpp>' from src/qr, then src/wallet too, but by popular demand, the scope is increased to fully remove #include <boost/foreach.hpp> the whole project.

Apart from a few small self documented commits in preparation for the scripted commits, the content of the scripted scripts themselves is:

```
sed -i 's/BOOST_FOREACH *(\(.*\),/for (\1 :/' ./src/*.h ./src/*.cpp ./src/*/*.h ./src/*/*.cpp ./src/*/*/*.h ./src/*/*/*.cpp ;
sed -i 's/BOOST_REVERSE_FOREACH(\(.*\), \(.*\))/for (\1 : reverse_iterate(\2))/' ./src/*.h ./src/*.cpp ./src/*/*.h ./src/*/*.cpp ;
sed -i ':a;N;$!ba;s/#include <boost\/foreach.hpp>\n//' ./src/*.h ./src/*.cpp ./src/*/*.h ./src/*/*.cpp ./src/*/*/*.h ./src/*/*/*.cpp
```

Dependencies:

- [x] devtools/net: add a verifier for scriptable changes. Use it to make CNode::id private. https://github.com/bitcoin/bitcoin/pull/10189
- [x] scripted-diff: Remove BOOST_FOREACH, Q_FOREACH and PAIRTYPE #10502


dcousens:
concept ACK,  but maybe follow [after verification/approval] with a removal of the includes too?

practicalswift:
concept ACK

Very good!

jtimon:
Updated OP and code increasing the scope to fully ```remove #include <boost/foreach.hpp>``` as discussed here and on IRC.

Although I expect travis to pass, I encountered a problem and "fixed it" by commenting part of prevector_tests.cpp in https://github.com/bitcoin/bitcoin/pull/10193/commits/cc40182b533dbe6282e7c6cb8d2dae61f9a7ead4 I tried https://github.com/bitcoin/bitcoin/pull/10193/commits/eb83e88d55cd3a4b230d9c99e6d8a97135167c74 and some trivial edits to the commented lines and I don't know what else to try. If anybody can help with this it would be very welcomed.

If we cannot solve this soon or reverse_iterator.hpp is not acceptable even temporarily for some reason I suggest to go back to remove BOOST_FOREACH everywhere and #include <boost\/foreach.hpp> everywhere except for the few places that use BOOST_REVERSE_FOREACH as a first step. But I hope we don't need to do that.

src/reverse_iterator.hpp
```diff
@@ -0,0 +1,39 @@
...
+template <typename T>
+class reverse_range
+{
+    const T &x;
+    
+public:
+    reverse_range(const T &x) : x(x) {}
+    
+    auto begin() const -> decltype(this->x.rbegin())
+    {
```
dcousens: is the `decltype` even needed?

src/init.cpp
```diff
@@ -601,7 +601,7 @@ void CleanupBlockRevFiles()
     // keeping a separate counter.  Once we hit a gap (or if 0 doesn't exist)
     // start removing block files.
     int nContigCounter = 0;
-    BOOST_FOREACH(const PAIRTYPE(std::string, fs::path)& item, mapBlockFiles) {
+    for (const PAIRTYPE(std::string, fs::path)& item : mapBlockFiles) {
```
tjps: PAIRTYPE only exists in service of BOOST_FOREACH, would it make sense to remove it in this PR as well?

gmaxwell:
Luke points out that if only travis is running this it is a vulnerability because we'll be falsely confident in the changes.   Other people should run it, but then we're exposed because I assume many people have a workflow that will git pull / git diff and run scripts without reading all the commit messages. spudowiar  suggested on IRC that the review script ask you to confirm the script it's going to run.  And I think when it's run outside of travis thats what it should do, and it answers the above concerns... reviewers can check this without creating a gratuitous invisible script vector.

MarcoFalke:
The issue that you can not trust travis for the result of the run (i.e. red vs green) is true regardless of the scripted-diff script. Though, I agree with @gmaxwell that travis serves as a good tool to notify pr authors of issues with the commit. Also agree that the interactive approach should prevent most evil scripts from being run.

src/reverse_iterator.hpp
jtimon: Yes, if I remove it I get ```‘begin’ function uses ‘auto’ type specifier without trailing return type```.
If I replace the auto with T::reverse_iterator I get ```need ‘typename’ before ‘T::reverse_iterator’ because ‘T’ is a dependent scope```.

dcousens: @jtimon right, that is a C++14 feature it appears.

jtimon:
It seems some more general comments really belong in #10189 .
I agree with the concerns. I think anyone giving an utACK to an "scripted-diff" (or whatever we chose as the prefix) commit should read the script and either review the changes (that's what I did for #10189) or run the script locally. Even if the script in #10189 facilitates review and will help travis warn you when your commit is not complete anymore after rebase (for example, because of new uses of something you are replacing), it is not a replacement for review. I think #10189 's script helps more writers than reviewers (I often want to do "this and nothing else" but some times more things end up in the same commit by accident), but it also establishes a clear format for commit messages for commits using scripts and that makes review easier (because reviewers can run the script themselves or at least review thinking the commit shouldn't do anything else than what the script says [although that king of review doesn't guarantee the change is complete]).

@tjps I tried to remove PAIRTYPE but it seems Q_FOREACH needs it too.
I'm removing Q_FOREACH too to see what people think it seems we may lose some performance with that, see https://www.dvratil.cz/2015/06/qt-containers-and-c11-range-based-loops/
Also, needed rebase.

jtimon:
Rebased. I still don't know what's wrong with the reverse iterator, so I separated the first 3 commits in #10502

jtimon:
Needed rebase, see https://github.com/bitcoin/bitcoin/pull/10502#issuecomment-305664766

jtimon:
Needed rebase.
Also, it seems I solved the reverse iterator problem in prevector_tests.cpp thanks to @theuni 's suggestion.

sipa:
Needs rebase.

jtimon:
Rebased, we still don't know if the fix @theuni suggested is needed only needed on prevector_tests.cpp as I though or everywhere as he thinks.

If he is right, ideally all 5 lines changed in https://github.com/bitcoin/bitcoin/pull/10193/commits/ad087bd496e46b4ae93b96089c0d609f183d3e4d (aka 'scripted-diff: Remove BOOST_REVERSE_FOREACH' for when the rebase) deserve their own test that fails with that very commit.

Otherwise I think this is ready to merge squashes aside. Hopefully this is now much easier to review once the simpler yet most disruptive parts have been merged with #10502.

theuni:
Concept ACK, but NACK until reverse_range/reverse_iterator have their own tests as @jtimon mentioned above. I believe that the current usage is broken due to scope issues with reverse_iterate in a range-based-for. A language lawyer may prove otherwise, but it's unnecessarily risky without tests to back it up.

jtimon:
I don't mean writting their own test for reverse_range/reverse_iterator, at that point I think I'd rather just delete it.

What I mean is that if the 5 changes in https://github.com/bitcoin/bitcoin/commit/ad087bd496e46b4ae93b96089c0d609f183d3e4d are incorrect (as you think they are and unlike those in https://github.com/bitcoin/bitcoin/pull/10193/commits/e01c26c1f81a6ba5949e8292e42979e26be971aa ) then it is unfortunate that no test catches the at least 5 errors that this PR is supposedly introducing.

src/.clang-format
```diff
@@ -23,7 +23,7 @@ ContinuationIndentWidth: 4
 Cpp11BracedListStyle: true
 DerivePointerAlignment: false
 DisableFormat:   false
-ForEachMacros:   [ foreach, Q_FOREACH, BOOST_FOREACH, BOOST_REVERSE_FOREACH ]
+ForEachMacros:   [ foreach ]
```
MarcoFalke: Is this macro still used?

jtimon: I think it was never used but it was there from the beginning ( https://github.com/jtimon/bitcoin/commit/2887bffcfdc138f53c60091ab2906790971c9af5 ). @sipa do you remember why we add it? Perhaps we can Remove the full ForEachMacro line ?

jtimon:
As pointed out on IRC by @theuni it seems we neither need to do the same as in prevector_tests everywhere, it is not even needed in prevector_tests.cpp itself with a small fix in prevector. Thanks again.

jtimon:
Updated with @theuni 's latest proposed fix and fully removing ForEachMacros from clang-format following @MarcoFalke 's hint.

theuni:
NACK retracted after some local experimentation. This is effectively tested in the prevector tests, which now work as expected.

utACK other than the header name nit.

src/reverse_iterator.hpp
```diff
@@ -0,0 +1,39 @@
+// Taken from https://gist.github.com/arvidsson/7231973
```
MarcoFalke: Can we just use this here? I couldn't find that this is MIT licensed.

jtimon: I assumed it is public domain, but perhaps we should ask the author?

jtimon:
Fixed @theuni 's nit.

sipa:
utACK b1268a19d0b80401339ede2188abbd389f8d7fb0

practicalswift:
utACK b1268a1 modulo licensing approval from @arvidsson :-)

src/reverse_iterator.hpp
arvidsson: Yeah, it's public domain.

practicalswift: Thanks @arvidsson!

fanquake:
ACK b1268a1

~140 Boost includes left, with the majority being in /test code.

laanwj:
utACK b1268a1

paveljanik:
This PR brought few shadowing warnings:

```
./reverse_iterator.h:20:22: warning: declaration shadows a field of 'reverse_range<T>' [-Wshadow]
```

jtimon:
Sorry about that. It seems I don't get those warnings locally?

paveljanik:
@jtimon You have to build with `-Wshadow`.

jtimon:
@paveljanik any reason why that's not the default if we want to comply with it?

paveljanik:
Long story short: compilers are evolving and support this option differently. Sometimes old versions warn about correct code etc.

PR #10194, opened 2017-04-12
Remove unecessary friend keyword from the class definition

bulldozer00:
The fewer the friends in a C++ code base, the better. Increased readability, less invasive coupling, better style. Also, if used, friend funcs should only declared within the scope of a class, and then defined outside of the class.

### DIFF
--- a/src/coins.h
+++ b/src/coins.h
@@ -134,7 +134,7 @@ public:
     }
 
     //! equality test
-    friend bool operator==(const CCoins &a, const CCoins &b) {
+    bool operator==(const CCoins &a, const CCoins &b) {
          // Empty CCoins objects are always equal.
          if (a.IsPruned() && b.IsPruned())
              return true;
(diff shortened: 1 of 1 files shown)


laanwj:
This breaks the build:
```
In file included from txmempool.h:17:0,
                 from blockencodings.cpp:12:
coins.h:137:53: error: ‘bool CCoins::operator==(const CCoins&, const CCoins&)’ must take exactly one argument
     bool operator==(const CCoins &a, const CCoins &b) {
                                                     ^
coins.h:147:53: error: ‘bool CCoins::operator!=(const CCoins&, const CCoins&)’ must take exactly one argument
     bool operator!=(const CCoins &a, const CCoins &b) {
                                                     ^
```

bulldozer00:
My bad. Put the change in too quick. Should be good to go now?

MarcoFalke:
Please try a `make check`  on your machine.

bulldozer0:
Will do. Won't be able to try/fix locally until /later on today.

bulldozer00:
Ok guys, the unit test that I broke was a good one. By attempting to compare a const object with a non-const one for euality, it showed me that the operator==() and operator!=() functions must be declared as "friends" of the class. Rather than be pedantic and changing the code trivially so that the friend declaration remains within the class but the definition be pulled outside of it, I'm going to close out this pull request.

Sorry for the inconvenience. It's my first try at learning Git, the GitHub workflow, and the Bitcoin build system. I promise to be more careful in the future.

PR #10197, opened 2017-04-12
[tests] Functional test warnings

jnewbery:
Two changes:

1. Issue warnings when running test_runner:

- if there are python files in the directory that are not listed in test_runner (this is downgraded from an error)
- if there is already a bitcoind process running on the system (uses pidof - only works for linux)
- if there is a cache directory

2. delete cache directory every time test_runner is started. Adds a parameter --keepcache to override the default behaviour and keep the cache directory from the previous run.

Rebuilding the cache takes about 21 seconds on my pc. This will be reduced to just under 2 seconds once #10082 has been fully merged.

### DIFF
--- a/test/functional/test_runner.py
+++ b/test/functional/test_runner.py
@@ -244,6 +244,17 @@ def main():
     run_tests(test_list, config["environment"]["SRCDIR"], config["environment"]["BUILDDIR"], config["environment"]["EXEEXT"], args.jobs, args.coverage, passon_args)
 
 def run_tests(test_list, src_dir, build_dir, exeext, jobs=1, enable_coverage=False, args=[]):
+    # Warn if bitcoind is already running (unix only)
+    try:
+        if subprocess.check_output(["pidof", "bitcoind"]) is not None:
+            print("%sWARNING!%s There is already a bitcoind process running on this system. Tests may fail unexpectedly due to resource contention!" % (BOLD[1], BOLD[0]))
+    except (OSError, subprocess.SubprocessError):
+        pass
+
+    # Warn if there is a cache directory
+    cache_dir = "%s/test/cache" % build_dir
+    if os.path.isdir(cache_dir):
+        print("%sWARNING!%s There is a cache directory here: %s. If tests fail unexpectedly, try deleting the cache directory." % (BOLD[1], BOLD[0], cache_dir))
 
     #Set env vars
     if "BITCOIND" not in os.environ:
@@ -252,7 +263,7 @@ def run_tests(test_list, src_dir, build_dir, exeext, jobs=1, enable_coverage=Fal
     tests_dir = src_dir + '/test/functional/'
 
     flags = ["--srcdir={}/src".format(build_dir)] + args
-    flags.append("--cachedir=%s/test/cache" % build_dir)
+    flags.append("--cachedir=%s" % cache_dir)
 
     if enable_coverage:
         coverage = RPCCoverage()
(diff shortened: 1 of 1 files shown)


test/functional/test_runner.py
```diff
@@ -407,9 +422,7 @@ def check_script_list(src_dir):
     python_files = set([t for t in os.listdir(script_dir) if t[-3:] == ".py"])
     missed_tests = list(python_files - set(map(lambda x: x.split()[0], ALL_SCRIPTS + NON_SCRIPTS)))
     if len(missed_tests) != 0:
-        print("The following scripts are not being run:" + str(missed_tests))
-        print("Check the test lists in test_runner.py")
-        sys.exit(1)
+        print("%sWARNING!%s The following scripts are not being run: %s. Check the test lists in test_runner.py." % (BOLD[1], BOLD[0], str(missed_tests)))
```
MarcoFalke: I'd prefer to keep this an error, so that a failing travis will quickly tell the author of a new script that it is not being run.

MarcoFalke:
utACK 3fd09cb beside the comment

jnewbery:
@MarcoFalke I've found the error-and-exit behaviour annoying and intrusive since we introduced it. I'm often working on test cases outside a git repository and either want to add a new test case or modify an existing test case and keep a backup in the same directory (my exact setup is that I'm using rsync to copy my git repo into a directory on a VM, and old files don't automatically get deleted when I checkout a new branch and rsync it to the VM). The fact that test_runner fails to run when it detects a new file in the directory is annoying.

Remember that the point of this check is so new test cases don't get added to the repo without being added to test_runner. Only one person needs to notice that the warning is being raised either locally or on Travis for it to have done its job. For that, a warning should be enough.

MarcoFalke:
@jnewbery Sounds reasonable.

test/functional/test_runner.py
```diff
@@ -241,9 +242,23 @@ def main():
...
+        shutil.rmtree("%s/test/cache" % config["environment"]["BUILDDIR"], ignore_errors=True)
+
     run_tests(test_list, config["environment"]["SRCDIR"], config["environment"]["BUILDDIR"], config["environment"]["EXEEXT"], args.jobs, args.coverage, passon_args)
 
 def run_tests(test_list, src_dir, build_dir, exeext, jobs=1, enable_coverage=False, args=[]):
+    # Warn if bitcoind is already running (unix only)
+    try:
+        if subprocess.check_output(["pidof", "bitcoind"]) is not None:
+            print("%sWARNING!%s There is already a bitcoind process running on this system. Tests may fail unexpectedly due to resource contention!" % (BOLD[1], BOLD[0]))
+    except:
```
```py
except (OSError, subprocess.SubprocessError):
```

should be enough?

test/functional/test_runner.py
MarcoFalke: Maybe make this warning an error on travis?

```py
if os.getenv('TRAVIS')=='true':
  # On travis this warning is an error to prevent merging incomplete commits into master
  sys.exit(1)
```

jnewbery:
Addressed @MarcoFalke's nits

Sjors:
If a MacOS user installs `pidof` via Homebrew, `subprocess.check_output(["pidof", "bitcoind"])` will return `b''` which isn't `None`. This then causes the "There is already a bitcoind process running on this system" to always appear. Not sure if that's worth fixing.

jnewbery:
> Not sure if that's worth fixing.

Certainly! PRs welcome 🙂

PR #10198, opened 2017-04-12
[tests] Remove is_network_split from functional test framework

jnewbery:
The `split_network()` and `join_network()` functions in bitcoin test framework are problematic:

1. They make assumptions about the network topology (4 nodes, either connected linearly A-B-C-D or split A-B // C-D), and force these assumptions on *all* test cases. The `setup_network()` and `sync_all()` functions use `is_network_split`, which means that many test cases have to set this variable for no other reason than to make those functions work. Removing the requirement to set `is_network_split` means many test cases don't need to override the `setup_network()` method, which removes some 200 loc. `split_network()` and `join_network()` are only used by 2 test cases - that's a lot of redundant code for the benefit of two test cases.
2. They're horribly inefficient. They both stop all 4 nodes, then restart. That takes ~10 seconds on my pc.
3. They may have unexpected side-effects. stop-starting removes ephemeral state on the node, and that behaviour may change between releases (eg mempool persistence).

This PR:

- removes the `is_network_split` variable
- makes `setup_network()` and `sync_all()` more generic by making fewer assumptions about the test network topology
- uses a new `disconnect_nodes()` helper function in `split_network()` and `join_network()`

This requires #10143, which isn't yet merged

benefits:

1. removes ~200 lines of unnecessary test setup code
2. getchaintips and listsinceblock now run faster:

before:
```
 ./listsinceblock.py && ./getchaintips.py 
2017-04-12 15:43:53.949000 TestFramework (INFO): Initializing test directory /tmp/user/1000/test58b95puo/15463
2017-04-12 15:44:05.686000 TestFramework (INFO): lastblockhash=798a876d18cb546c59e76d7d0f62611b588dd6046ba989e5800cc8a981e025db
2017-04-12 15:44:15.959000 TestFramework (INFO): Stopping nodes
2017-04-12 15:44:24.526000 TestFramework (INFO): Cleaning up
2017-04-12 15:44:24.529000 TestFramework (INFO): Tests successful
2017-04-12 15:44:24.590000 TestFramework (INFO): Initializing test directory /tmp/user/1000/testk_xibf92/15632
2017-04-12 15:44:46.263000 TestFramework (INFO): Stopping nodes
2017-04-12 15:44:54.658000 TestFramework (INFO): Cleaning up
2017-04-12 15:44:54.667000 TestFramework (INFO): Tests successful

```
after:
```
2017-04-12 15:43:12.662000 TestFramework (INFO): Initializing test directory /tmp/user/1000/testqi2c2skb/15342
2017-04-12 15:43:14.616000 TestFramework (INFO): lastblockhash=7c70eee0979069504d6059c3cd9658fc8c57d8ae30a3872716c1f9a244a53ae1
2017-04-12 15:43:14.834000 TestFramework (INFO): Stopping nodes
2017-04-12 15:43:23.205000 TestFramework (INFO): Cleaning up
2017-04-12 15:43:23.207000 TestFramework (INFO): Tests successful
2017-04-12 15:43:23.270000 TestFramework (INFO): Initializing test directory /tmp/user/1000/testmxve6ryf/15399
2017-04-12 15:43:25.363000 TestFramework (INFO): Stopping nodes
2017-04-12 15:43:33.648000 TestFramework (INFO): Cleaning up
2017-04-12 15:43:33.658000 TestFramework (INFO): Tests successful
```
excluding the test shutdown time (which will be improved by #10082), the change is from 44s to 4.2s (10x speedup).

3. bitcoin test framework is now more generic, and can be extended later to automatically spin up different test network topologies.

### DIFF
--- a/test/functional/bip65-cltv.py
+++ b/test/functional/bip65-cltv.py
@@ -12,15 +12,12 @@ class BIP65Test(BitcoinTestFramework):
         super().__init__()
         self.num_nodes = 3
         self.setup_clean_chain = False
+        self.extra_args = [[], ["-blockversion=3"], ["-blockversion=4"]]
 
     def setup_network(self):
-        self.nodes = []
-        self.nodes.append(start_node(0, self.options.tmpdir, []))
-        self.nodes.append(start_node(1, self.options.tmpdir, ["-blockversion=3"]))
-        self.nodes.append(start_node(2, self.options.tmpdir, ["-blockversion=4"]))
+        self.setup_nodes()
         connect_nodes(self.nodes[1], 0)
         connect_nodes(self.nodes[2], 0)
-        self.is_network_split = False
         self.sync_all()
 
     def run_test(self):
--- a/test/functional/bip68-sequence.py
+++ b/test/functional/bip68-sequence.py
@@ -21,16 +21,11 @@ class BIP68Test(BitcoinTestFramework):
         super().__init__()
         self.num_nodes = 2
         self.setup_clean_chain = False
-
-    def setup_network(self):
-        self.nodes = []
-        self.nodes.append(start_node(0, self.options.tmpdir))
-        self.nodes.append(start_node(1, self.options.tmpdir, ["-acceptnonstdtxn=0"]))
-        self.is_network_split = False
-        self.relayfee = self.nodes[0].getnetworkinfo()["relayfee"]
-        connect_nodes(self.nodes[0], 1)
+        self.extra_args = [[], ["-acceptnonstdtxn=0"]]
 
     def run_test(self):
+        self.relayfee = self.nodes[0].getnetworkinfo()["relayfee"]
+
         # Generate some coins
         self.nodes[0].generate(110)
 
--- a/test/functional/bipdersig-p2p.py
+++ b/test/functional/bipdersig-p2p.py
@@ -47,12 +47,6 @@ class BIP66Test(ComparisonTestFramework):
         super().__init__()
         self.num_nodes = 1
 
-    def setup_network(self):
-        # Must set the blockversion for this test
-        self.nodes = start_nodes(self.num_nodes, self.options.tmpdir,
-                                 extra_args=[['-whitelist=127.0.0.1', '-blockversion=2']],
-                                 binary=[self.options.testbinary])
-
     def run_test(self):
         test = TestManager(self, self.options.tmpdir)
         test.add_all_connections(self.nodes)
--- a/test/functional/bipdersig.py
+++ b/test/functional/bipdersig.py
@@ -12,15 +12,12 @@ class BIP66Test(BitcoinTestFramework):
         super().__init__()
         self.num_nodes = 3
         self.setup_clean_chain = False
+        self.extra_args = [[], ["-blockversion=2"], ["-blockversion=3"]]
 
     def setup_network(self):
-        self.nodes = []
-        self.nodes.append(start_node(0, self.options.tmpdir, []))
-        self.nodes.append(start_node(1, self.options.tmpdir, ["-blockversion=2"]))
-        self.nodes.append(start_node(2, self.options.tmpdir, ["-blockversion=3"]))
+        self.setup_nodes()
         connect_nodes(self.nodes[1], 0)
         connect_nodes(self.nodes[2], 0)
-        self.is_network_split = False
         self.sync_all()
 
     def run_test(self):
--- a/test/functional/blockchain.py
+++ b/test/functional/blockchain.py
@@ -35,12 +35,6 @@ class BlockchainTest(BitcoinTestFramework):
         self.setup_clean_chain = False
         self.num_nodes = 2
 
-    def setup_network(self, split=False):
-        self.nodes = start_nodes(self.num_nodes, self.options.tmpdir)
-        connect_nodes_bi(self.nodes, 0, 1)
-        self.is_network_split = False
-        self.sync_all()
-
     def run_test(self):
         self._test_gettxoutsetinfo()
         self._test_getblockheader()
--- a/test/functional/disablewallet.py
+++ b/test/functional/disablewallet.py
@@ -18,11 +18,7 @@ class DisableWalletTest (BitcoinTestFramework):
         super().__init__()
         self.setup_clean_chain = True
         self.num_nodes = 1
-
-    def setup_network(self, split=False):
-        self.nodes = start_nodes(self.num_nodes, self.options.tmpdir, [['-disablewallet']])
-        self.is_network_split = False
-        self.sync_all()
+        self.extra_args = [["-disablewallet"]]
 
     def run_test (self):
         x = self.nodes[0].validateaddress('3J98t1WpEZ73CNmQviecrnyiWrnqRhWNLy')
--- a/test/functional/getblocktemplate_proposals.py
+++ b/test/functional/getblocktemplate_proposals.py
@@ -73,10 +73,6 @@ class GetBlockTemplateProposalTest(BitcoinTestFramework):
         self.num_nodes = 2
         self.setup_clean_chain = False
 
-    def setup_network(self):
-        self.nodes = self.setup_nodes()
-        connect_nodes_bi(self.nodes, 0, 1)
-
     def run_test(self):
         node = self.nodes[0]
         node.generate(1) # Mine a block to leave initial block download
--- a/test/functional/importmulti.py
+++ b/test/functional/importmulti.py
@@ -12,9 +12,8 @@ class ImportMultiTest (BitcoinTestFramework):
         self.num_nodes = 2
         self.setup_clean_chain = True
 
-    def setup_network(self, split=False):
-        self.nodes = start_nodes(2, self.options.tmpdir)
-        self.is_network_split=False
+    def setup_network(self):
+        self.setup_nodes()
 
     def run_test (self):
         self.log.info("Mining blocks...")
--- a/test/functional/importprunedfunds.py
+++ b/test/functional/importprunedfunds.py
@@ -14,12 +14,6 @@ class ImportPrunedFundsTest(BitcoinTestFramework):
         self.setup_clean_chain = True
         self.num_nodes = 2
 
-    def setup_network(self, split=False):
-        self.nodes = start_nodes(self.num_nodes, self.options.tmpdir)
-        connect_nodes_bi(self.nodes,0,1)
-        self.is_network_split=False
-        self.sync_all()
-
     def run_test(self):
         self.log.info("Mining blocks...")
         self.nodes[0].generate(101)
--- a/test/functional/invalidateblock.py
+++ b/test/functional/invalidateblock.py
@@ -8,20 +8,15 @@ from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import *
 
 class InvalidateTest(BitcoinTestFramework):
-    
-        
+
     def __init__(self):
         super().__init__()
         self.setup_clean_chain = True
         self.num_nodes = 3
 
     def setup_network(self):
-        self.nodes = []
-        self.is_network_split = False 
-        self.nodes.append(start_node(0, self.options.tmpdir))
-        self.nodes.append(start_node(1, self.options.tmpdir))
-        self.nodes.append(start_node(2, self.options.tmpdir))
-        
+        self.setup_nodes()
+
     def run_test(self):
         self.log.info("Make sure we repopulate setBlockIndexCandidates after InvalidateBlock:")
         self.log.info("Mine 4 blocks on Node 0")
--- a/test/functional/mempool_limit.py
+++ b/test/functional/mempool_limit.py
@@ -9,30 +9,25 @@ from test_framework.util import *
 
 class MempoolLimitTest(BitcoinTestFramework):
 
-    def setup_network(self):
-        self.nodes = []
-        self.nodes.append(start_node(0, self.options.tmpdir, ["-maxmempool=5", "-spendzeroconfchange=0"]))
-        self.is_network_split = False
-        self.sync_all()
-        self.relayfee = self.nodes[0].getnetworkinfo()['relayfee']
-
     def __init__(self):
         super().__init__()
         self.setup_clean_chain = True
         self.num_nodes = 1
-
-        self.txouts = gen_return_txouts()
+        self.extra_args = [["-maxmempool=5", "-spendzeroconfchange=0"]]
 
     def run_test(self):
+        txouts = gen_return_txouts()
+        relayfee = self.nodes[0].getnetworkinfo()['relayfee']
+
         txids = []
-        utxos = create_confirmed_utxos(self.relayfee, self.nodes[0], 91)
+        utxos = create_confirmed_utxos(relayfee, self.nodes[0], 91)
 
         #create a mempool tx that will be evicted
         us0 = utxos.pop()
         inputs = [{ "txid" : us0["txid"], "vout" : us0["vout"]}]
         outputs = {self.nodes[0].getnewaddress() : 0.0001}
         tx = self.nodes[0].createrawtransaction(inputs, outputs)
-        self.nodes[0].settxfee(self.relayfee) # specifically fund this tx with low fee
+        self.nodes[0].settxfee(relayfee) # specifically fund this tx with low fee
         txF = self.nodes[0].fundrawtransaction(tx)
         self.nodes[0].settxfee(0) # return to automatic fee selection
         txFS = self.nodes[0].signrawtransaction(txF['hex'])
@@ -42,7 +37,7 @@ class MempoolLimitTest(BitcoinTestFramework):
         base_fee = relayfee*100
         for i in range (3):
             txids.append([])
-            txids[i] = create_lots_of_big_transactions(self.nodes[0], self.txouts, utxos[30*i:30*i+30], 30, (i+1)*base_fee)
+            txids[i] = create_lots_of_big_transactions(self.nodes[0], txouts, utxos[30*i:30*i+30], 30, (i+1)*base_fee)
 
         # by now, the tx should be evicted, check confirmation state
         assert(txid not in self.nodes[0].getrawmempool())
--- a/test/functional/mempool_spendcoinbase.py
+++ b/test/functional/mempool_spendcoinbase.py
@@ -22,13 +22,7 @@ class MempoolSpendCoinbaseTest(BitcoinTestFramework):
         super().__init__()
         self.num_nodes = 1
         self.setup_clean_chain = False
-
-    def setup_network(self):
-        # Just need one node for this test
-        args = ["-checkmempool"]
-        self.nodes = []
-        self.nodes.append(start_node(0, self.options.tmpdir, args))
-        self.is_network_split = False
+        self.extra_args = [["-checkmempool"]]
 
     def run_test(self):
         chain_height = self.nodes[0].getblockcount()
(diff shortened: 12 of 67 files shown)


jnewbery:
closing until #10143 is merged

jnewbery:
#10143 is now merged, so rebasing and re-opening this.

@kallewoof - you've been working on listsinceblock recently. Do you mind reviewing this and letting me know what you think?

test/functional/test_framework/util.py
```diff
@@ -385,6 +385,17 @@ def set_node_times(nodes, t):
...
         node.setmocktime(t)
 
+def disconnect_nodes(from_connection, node_num):
+    timeout = 5
+    while True:
+        peer_ids = [peer['id'] for peer in from_connection.getpeerinfo() if "testnode%d" % node_num in peer['subver']]
+        if not peer_ids:
+            break
+        [from_connection.disconnectnode(nodeid=peer_id) for peer_id in peer_ids]
+        assert timeout > 0.1, "timed out waiting for disconnect"
```
jimmysong: This seems like it would be made better as the for loop condition:

    for i in range(50): 
        ...
        time.sleep(0.1)
    else:
        assert 0, "timed out waiting for disconnect"

At least for me, that's a bit easier to read.

test/functional/test_framework/util.py
```diff
@@ -385,6 +385,17 @@ def set_node_times(nodes, t):
     for node in nodes:
         node.setmocktime(t)
 
+def disconnect_nodes(from_connection, node_num):
+    timeout = 5
+    while True:
+        peer_ids = [peer['id'] for peer in from_connection.getpeerinfo() if "testnode%d" % node_num in peer['subver']]
+        if not peer_ids:
+            break
+        [from_connection.disconnectnode(nodeid=peer_id) for peer_id in peer_ids]
```
jimmysong: This happens every time through the loop. Is that your intention? Doesn't this only have to be done once?

test/functional/test_framework/test_framework.py
```diff
@@ -85,15 +85,15 @@ def split_network(self):
...
-            sync_mempools(self.nodes)
+    def sync_all(self, node_groups=None):
+        if not node_groups:
+            if hasattr(self, "is_network_split") and self.is_network_split:
+                node_groups = [self.nodes[:2], self.nodes[2:]]
+            else:
+                node_groups = [self.nodes]
+
+        [sync_blocks(group) for group in node_groups]
+        [sync_mempools(group) for group in node_groups]
```
jimmysong: If you're going to do this, why not just a single for loop?

    for group in node_groups:
        sync_blocks(group)
        sync_mempools(group)

For me, that's a little easier to read.

test/functional/bip68-sequence.py
```diff
@@ -21,15 +21,11 @@ def __init__(self):
...
         self.num_nodes = 2
         self.setup_clean_chain = False
+        self.extra_args = [[], ["-acceptnonstdtxn=0"]]
 
-    def setup_network(self):
-        self.nodes = []
-        self.nodes.append(start_node(0, self.options.tmpdir))
-        self.nodes.append(start_node(1, self.options.tmpdir, ["-acceptnonstdtxn=0"]))
+    def run_test(self):
         self.relayfee = self.nodes[0].getnetworkinfo()["relayfee"]
```
jimmysong: Shouldn't this go in `__init__`?

test/functional/bip68-sequence.py
```diff
@@ -21,15 +21,11 @@ def __init__(self):
...
-        connect_nodes(self.nodes[0], 1)
```
jimmysong: Also, since setup_nodes isn't called here, doesn't this need to be called at some point?

test/functional/bipdersig.py
```diff
@@ -12,12 +12,10 @@ def __init__(self):
         super().__init__()
         self.num_nodes = 3
         self.setup_clean_chain = False
+        self.extra_args = [["-blockversion=2"], ["-blockversion=3"]]
```
jimmysong: Shouldn't this have an empty list first? Node 0 has no opts, Node 1 has blockversion=2, etc?

test/functional/mempool_limit.py
```diff
@@ -9,20 +9,17 @@
...
-        self.relayfee = self.nodes[0].getnetworkinfo()['relayfee']
-
     def __init__(self):
         super().__init__()
         self.setup_clean_chain = True
         self.num_nodes = 1
+        self.extra_args = [["-maxmempool=5", "-spendzeroconfchange=0"]]
 
+    def run_test(self):
         self.txouts = gen_return_txouts()
```
jimmysong: shouldn't this be in `__init__`?

test/functional/mempool_spendcoinbase.py
```diff
@@ -22,12 +22,7 @@ def __init__(self):
         super().__init__()
         self.num_nodes = 1
         self.setup_clean_chain = False
-
-    def setup_network(self):
-        # Just need one node for this test
```
jimmysong: keep comment?

test/functional/blockchain.py
```diff
@@ -31,11 +31,6 @@ def __init__(self):
         self.setup_clean_chain = False
         self.num_nodes = 2
 
-    def setup_network(self, split=False):
-        self.nodes = start_nodes(self.num_nodes, self.options.tmpdir)
-        connect_nodes_bi(self.nodes, 0, 1)
```
jimmysong: doesn't this need to be called at some point?

test/functional/getblocktemplate_proposals.py
```diff
@@ -73,10 +73,6 @@ def __init__(self):
         self.num_nodes = 2
         self.setup_clean_chain = False
 
-    def setup_network(self):
-        self.nodes = self.setup_nodes()
-        connect_nodes_bi(self.nodes, 0, 1)
```
jimmysong: isn't this necessary?

test/functional/importprunedfunds.py
```diff
@@ -14,11 +14,6 @@ def __init__(self):
         self.setup_clean_chain = True
         self.num_nodes = 2
 
-    def setup_network(self, split=False):
-        self.nodes = start_nodes(self.num_nodes, self.options.tmpdir)
-        connect_nodes_bi(self.nodes,0,1)
```
jimmysong: doesn't this need a connection?

test/functional/bip65-cltv.py
```diff
@@ -12,12 +12,10 @@ def __init__(self):
...
 
     def setup_network(self):
-        self.nodes = []
-        self.nodes.append(start_node(0, self.options.tmpdir, []))
-        self.nodes.append(start_node(1, self.options.tmpdir, ["-blockversion=3"]))
-        self.nodes.append(start_node(2, self.options.tmpdir, ["-blockversion=4"]))
+        self.setup_nodes()
         connect_nodes(self.nodes[1], 0)
         connect_nodes(self.nodes[2], 0)
         self.sync_all()
```
jimmysong: do you need the `sync_all` ?

test/functional/bipdersig.py
```diff
@@ -12,12 +12,10 @@ def __init__(self):
...
 
     def setup_network(self):
-        self.nodes = []
-        self.nodes.append(start_node(0, self.options.tmpdir, []))
-        self.nodes.append(start_node(1, self.options.tmpdir, ["-blockversion=2"]))
-        self.nodes.append(start_node(2, self.options.tmpdir, ["-blockversion=3"]))
+        self.setup_nodes()
         connect_nodes(self.nodes[1], 0)
         connect_nodes(self.nodes[2], 0)
         self.sync_all()
```
jimmysong: Necessary?

test/functional/importmulti.py
```diff
@@ -12,8 +12,8 @@ def __init__(self):
         self.num_nodes = 2
         self.setup_clean_chain = True
 
-    def setup_network(self, split=False):
-        self.nodes = start_nodes(2, self.options.tmpdir)
+    def setup_network(self):
+        self.setup_nodes()
```
jimmysong: This is the same as the parent, no? Can be deleted?

test/functional/invalidateblock.py
```diff
@@ -8,19 +8,15 @@
...
         self.setup_clean_chain = True
         self.num_nodes = 3
 
     def setup_network(self):
-        self.nodes = []
-        self.nodes.append(start_node(0, self.options.tmpdir))
-        self.nodes.append(start_node(1, self.options.tmpdir))
-        self.nodes.append(start_node(2, self.options.tmpdir))
-        
+        self.setup_nodes()
```
jimmysong: Deleted?

test/functional/mempool_limit.py
```diff
@@ -9,20 +9,17 @@
...
 
-    def run_test(self):
+        self.relayfee = self.nodes[0].getnetworkinfo()['relayfee']
```
jimmysong: Also in `__init__`?

test/functional/p2p-acceptblock.py
```diff
@@ -98,17 +98,13 @@ def __init__(self):
...
         # Node0 will be used to test behavior of processing unrequested blocks
         # from peers which are not whitelisted, while Node1 will be used for
         # the whitelisted case.
-        self.nodes = []
-        self.nodes.append(start_node(0, self.options.tmpdir,
-                                     binary=self.options.testbinary))
-        self.nodes.append(start_node(1, self.options.tmpdir,
-                                     ["-whitelist=127.0.0.1"],
-                                     binary=self.options.testbinary))
+        self.nodes = self.setup_nodes()
```
jimmysong: delete?

test/functional/p2p-segwit.py
```diff
@@ -183,17 +183,13 @@ def __init__(self):
...
-        self.nodes.append(start_node(0, self.options.tmpdir, ["-whitelist=127.0.0.1"]))
-        # Start a node for testing IsStandard rules.
-        self.nodes.append(start_node(1, self.options.tmpdir, ["-whitelist=127.0.0.1", "-acceptnonstdtxn=0"]))
+        self.nodes = self.setup_nodes()
         connect_nodes(self.nodes[0], 1)
-
-        # Disable segwit's bip9 parameter to simulate upgrading after activation.
-        self.nodes.append(start_node(2, self.options.tmpdir, ["-whitelist=127.0.0.1", "-bip9params=segwit:0:0"]))
         connect_nodes(self.nodes[0], 2)
+        self.sync_all()
```
jimmysong: not needed?

test/functional/p2p-mempool.py
```diff
@@ -72,11 +72,10 @@ def __init__(self):
         super().__init__()
         self.setup_clean_chain = True
         self.num_nodes = 2
+        self.extra_args = [["-peerbloomfilters=0"]]
 
     def setup_network(self):
-        # Start a node with maxuploadtarget of 200 MB (/24h)
-        self.nodes = []
-        self.nodes.append(start_node(0, self.options.tmpdir, ["-peerbloomfilters=0"]))
+        self.setup_nodes()
```
jimmysong: delete?

test/functional/pruning.py
```diff
@@ -331,6 +322,14 @@ def wallet_test(self):
     def run_test(self):
         self.log.info("Warning! This test requires 4GB of disk space and takes over 30 mins (up to 2 hours)")
         self.log.info("Mining a big blockchain of 995 blocks")
+
+        # Determine default relay fee
+        self.relayfee = self.nodes[0].getnetworkinfo()["relayfee"]
+
+        # Cache for utxos, as the listunspent may take a long time later in the test
+        self.utxo_cache_0 = []
+        self.utxo_cache_1 = []
```
jimmysong: why did this move from `__init__`?

test/functional/rawtransactions.py
```diff
@@ -24,20 +24,9 @@ def __init__(self):
...
-
-        #connect to a local machine for debugging
-        #url = "http://bitcoinrpc:DP6DvqZtqXarpeNWyN3LZTFchCCyCUuHwNF7E8pX99x1@%s:%d" % ('127.0.0.1', 18332)
-        #proxy = AuthServiceProxy(url)
-        #proxy.url = url # store URL on proxy for info
-        #self.nodes.append(proxy)
-
-        connect_nodes_bi(self.nodes,0,1)
-        connect_nodes_bi(self.nodes,1,2)
+        super().setup_nodes()
```
jimmysong: `self.setup_nodes()` or `super().setup_network()`?

jimmysong (commented):
Concept ACK

Added a bunch of nits.

kallewoof:
ACK b9e9ef2415e3df508b0c0663502e60196bc37f0f

Nice, -231 lines of code. :) The listsinceblock stuff looks fine to me. I only ran tests but only briefly looked at code changes for other tests.

jimmysong:
@fanquake Can you label this Tests?

test/functional/test_framework/util.py
jnewbery: It *should* only need to be called once, but I called this every loop in case either:

- `disconnectnode` fails for one of the nodes and we need to rerun it.
- a new connection to the node is added while we're in the loop.

I don't think either of those should happen, so we can probably just remove this.

test/functional/test_framework/util.py
jnewbery: yes, makes sense. Changed

test/functional/rawtransactions.py
jnewbery: should be `super().setup_network()`. Fixed

test/functional/pruning.py
jnewbery: Because I'd really prefer `__init__()` to only set up the network topology. (eventually I think we should try to move away from individual test cases overriding the `__init__()` method entirely).

test/functional/p2p-segwit.py
jnewbery: probably not, but I don't see any harm in calling `sync_all()` once all connections have been added.

test/functional/p2p-mempool.py
jnewbery: Yes, and I can also reduce the number of nodes in this test to 1.

test/functional/p2p-acceptblock.py
jnewbery: No. The base class `setup_network()` would connect the two nodes, which we don't want in this test, so it is correct to override here. However, there is a bug here - I shouldn't assign `self.nodes` to the return value, which is None.

test/functional/mempool_spendcoinbase.py
jnewbery: meh. Should be clear from the `self.num_nodes` assignment in `__init__()`. I don't think we need this comment (and if we do need it, then there should be a similar comment in every other test case.

test/functional/mempool_limit.py
jnewbery: As below. I'd prefer `__init__()` to just be for setting up the test network topology.

However, these could be changed to be local variables instead of object variables.

test/functional/mempool_limit.py
jnewbery: ditto

test/functional/invalidateblock.py
jnewbery: `setup_nodes()` starts the nodes, but doesn't add connections between them. That's what we need here.

test/functional/getblocktemplate_proposals.py
jnewbery: This is now done by the base class `setup_network()`

test/functional/bipdersig.py
jnewbery: correct. Fixed

jnewbery (commented):
@jimmysong thanks for the review. I've addressed you nits. I'll rebase the commits locally and then push the new version.

test/functional/bip68-sequence.py
jnewbery: The base class calls `setup_network()`, which calls `setup_nodes()`

test/functional/bipdersig.py
jnewbery: probably not, but harmless

test/functional/importmulti.py
jnewbery: `setup_network()` in the base class also connects the nodes, which we don't want to do for this test.

jnewbery:
Rebased and added a fixup commit to address the nits.

I'm planning to squash these all down to a single commit before merge.

jimmysong:
utACK d64f739b3bc78e67212a252a690556bca8098745

I'll test now and test again once you squash.

jimmysong (approved):

kallewoof:
utACK d64f739b3bc78e67212a252a690556bca8098745

laanwj:
utACK d64f739, please squash

MarcoFalke:
Strong concept ACK. Going to review after squash.

jnewbery:
Commits squashed. There should be no difference between https://github.com/bitcoin/bitcoin/commit/d64f739b3bc78e67212a252a690556bca8098745 and 20f67c90c001123ab5bca7e70382c5fe26362aa3. Compare https://github.com/jnewbery/bitcoin/tree/pr10198.1 and https://github.com/jnewbery/bitcoin/tree/pr10198.2

test/functional/bipdersig-p2p.py
```diff
@@ -46,12 +46,8 @@ class BIP66Test(ComparisonTestFramework):
     def __init__(self):
         super().__init__()
         self.num_nodes = 1
-
-    def setup_network(self):
         # Must set the blockversion for this test
-        self.nodes = start_nodes(self.num_nodes, self.options.tmpdir,
-                                 extra_args=[['-whitelist=127.0.0.1', '-blockversion=2']],
-                                 binary=[self.options.testbinary])
+        self.extra_args = [[], ['-whitelist=127.0.0.1', '-blockversion=2']]
```
MarcoFalke: This does not work for the `ComparisionFramework`, as it is later overwritten. I don't understand why this commit does not fail, but I'd still like the bug fixed in this pull.


Additionally, I'd like to see a sanity check in start_nodes:

```diff
diff --git a/test/functional/test_framework/util.py b/test/functional/test_framework/util.py
index 7730e14..9186c3c 100644
--- a/test/functional/test_framework/util.py
+++ b/test/functional/test_framework/util.py
@@ -354,6 +354,8 @@ def start_nodes(num_nodes, dirname, extra_args=None, rpchost=None, timewait=None
     """
     if extra_args is None: extra_args = [ None for _ in range(num_nodes) ]
     if binary is None: binary = [ None for _ in range(num_nodes) ]
+    assert_equal(len(extra_args), num_nodes)
+    assert_equal(len(binary), num_nodes)
     rpcs = []
     try:
         for i in range(num_nodes):

File: test/functional/disablewallet.py [hunk 23]
```diff
@@ -19,11 +19,6 @@ def __init__(self):
         self.setup_clean_chain = True
         self.num_nodes = 1
```
MarcoFalke: Please add the missing self.extra_args here, and also add an additional test in this file:
```py
assert_raises_jsonrpc(-12, "Keypool ran out", lambda: self.nodes[0].generate(1))
```

maflcko (commented):
utACK 20f67c90c001123ab5bca7e70382c5fe26362aa3, but needs feedback addressed.

test/functional/bipdersig-p2p.py
jnewbery: It turns out the comment was wrong (and had presumably been copy-pasted from another test). blockversion is not required in this test, since the bitcoin node is not generating any blocks. I've removed the extra_args entirely.

I've also added your suggested asserts.

test/functional/disablewallet.py
jnewbery: yes, I've put the extra_args back in. Good spot.

I don't understand why adding the "Keypool ran out" assert should be part of this PR?

MarcoFalke:
Ok, feel free to squash. Planning to merge this.

jnewbery:
squashed and rebased. I'm currently running the extended test locally.

PR #10202, opened 2017-04-13
[Wallet] FundRawTransaction can fund a transaction with preset inputs found in the CoinView

NicolasDorier:
This PR makes it possible to call FundRawTransaction with pre filled inputs not belonging to the wallet.
This is very useful for AnyOneCanPay scenario, where one of a third party only cover part of a transaction.

The necessary information to complete the transaction is taken from of the mempool and coinview.

A typical example would involves Alice and Bob wanting to fund 0.5 each to the payment channel. Alice would give her input, and Bob would be able to complete the missing amount via FundRawTransaction.

A follow up PR will allow the client `fundrawtransaction` to pass previous TxOuts corresponding to the inputs. This is necessary for filling transaction in some 2nd layer protocols, where the the inputs's of the transaction to fund have not yet been broadcasted.

This supersede my previous attempt at https://github.com/bitcoin/bitcoin/pull/10068 which I closed, as it was harder to review.

The first two commits are strict refactoring.
The third is the one checking information in the coinview.
Rest is about tests.

### DIFF
--- a/src/wallet/coincontrol.h
+++ b/src/wallet/coincontrol.h
@@ -7,6 +7,7 @@
 
 #include "primitives/transaction.h"
 #include "wallet/wallet.h"
+#include <map>
 
 /** Coin Control Features. */
 class CCoinControl
@@ -76,8 +77,24 @@ public:
         vOutpoints.assign(setSelected.begin(), setSelected.end());
     }
 
+    void AddKnownCoins(const CInputCoin& coin)
+    {
+        knownCoins.insert(std::make_pair(coin.outpoint, coin));
+    }
+
+    boost::optional<CInputCoin> FindKnownCoin(const COutPoint& outpoint) const
+    {
+        boost::optional<CInputCoin> foundCoin;
+        auto it = knownCoins.find(outpoint);
+        if (it != knownCoins.end())
+            foundCoin = it->second;
+        return foundCoin;
+    }
+
 private:
     std::set<COutPoint> setSelected;
+    //! A map of known UTXO
+    std::map<COutPoint, CInputCoin> knownCoins;
 };
 
 #endif // BITCOIN_WALLET_COINCONTROL_H
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -2622,6 +2622,21 @@ UniValue listunspent(const JSONRPCRequest& request)
     return results;
 }
 
+boost::optional<CInputCoin> FindInCoinView(const COutPoint& outpoint)
+{
+    LOCK2(cs_main, mempool.cs);
+    boost::optional<CInputCoin> foundCoin;
+    CCoinsViewMemPool coinsTipMempool(pcoinsTip, mempool);
+    CCoinsViewCache view(&coinsTipMempool);
+    CCoins coins;
+    if (!view.GetCoins(outpoint.hash, coins))
+        return foundCoin;
+    if (!coins.IsAvailable(outpoint.n))
+        return foundCoin;
+    foundCoin = CInputCoin(coins.vout[outpoint.n], outpoint);
+    return foundCoin;
+}
+
 UniValue fundrawtransaction(const JSONRPCRequest& request)
 {
     CWallet * const pwallet = GetWalletForJSONRPCRequest(request);
@@ -2780,6 +2795,8 @@ UniValue fundrawtransaction(const JSONRPCRequest& request)
         coinControl.Select(txin.prevout);
         boost::optional<CInputCoin> foundCoin;
         foundCoin = pwallet->FindCoin(txin.prevout);
+        if(!foundCoin)
+            foundCoin = FindInCoinView(txin.prevout);
         if(!foundCoin)
             throw JSONRPCError(RPC_WALLET_ERROR, _("Insufficient funds"));
         coinControl.AddKnownCoins(*foundCoin);
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -21,6 +21,7 @@
 #include "wallet/feebumper.h"
 #include "wallet/wallet.h"
 #include "wallet/walletdb.h"
+#include "wallet/coincontrol.h"
 
 #include <stdint.h>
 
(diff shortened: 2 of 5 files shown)


TheBlueMatt (commented):
Concept ACK.

Its really quite unfortunate we cant properly calculate fee when using most outputs here. Maybe its possible to assume, at a minimum, a compressed pubkey and standard signature size for "normal" things like P2PH? Alternatively, could force the user to provide some additional data, like a set of pubkeys needed.

Also, Windows line endings in a few commits :p

NicolasDorier:
Fixed the end-of-line and fixed `wallet.cpp` to not break QT coin control as suggested by @TheBlueMatt .
In a follow up PR I will clean QT coin control dialog to use `AddKnownCoin`. I briefly tried to do that, it simplified quite a lot. I prefer keeping it in another PR though.

src/wallet/coincontrol.h
```diff
@@ -76,8 +77,24 @@ class CCoinControl
         vOutpoints.assign(setSelected.begin(), setSelected.end());
     }
 
+    void AddKnownCoins(const CInputCoin& coin)
+    {
+        knownCoins.insert(std::make_pair(coin.outpoint, coin));
+    }
+
+    boost::optional<CInputCoin> FindKnownCoin(const COutPoint& outpoint) const
```
jonasschnelli: Can we avoid `boost::optional`?
Maybe `std::shared_ptr<CInputCoin>` should do it?
Maybe switch the map keys to std::pair<uint256, int>?

src/wallet/coincontrol.h
```diff
@@ -76,8 +77,24 @@ class CCoinControl
...
+        auto it = knownCoins.find(outpoint);
+        if (it != knownCoins.end())
+            foundCoin = it->second;
+        return foundCoin;
+    }
+
 private:
     std::set<COutPoint> setSelected;
+    //! A map of known UTXO
+    std::map<COutPoint, CInputCoin> knownCoins;
```
jonasschnelli: Why not use `setSelected` (and maybe extend/migrate it to fit your use-case), AFAIK they serve the same purpose (selecting an desired input over the GUI).

I think this is not something we should fix in a follow-up PR because it kinda messes up the design.

jonasschnelli:
Concept ACK.
I think we should not have two code parts for handling pre-selected-inputs and I think we should avoid using `boost::optional` (if reasonable possible).

src/wallet/coincontrol.h
NicolasDorier: why? We are already using `boost::optional` from https://github.com/bitcoin/bitcoin/pull/10165/files#diff-b2bb174788c7409b671c46ccc86034bdR2088 . This will be implemented in `std::` in C++0x17 as well. I was kind of against it on https://github.com/bitcoin/bitcoin/pull/10165, but it turns out to be very handy.

src/wallet/coincontrol.h
NicolasDorier: I briefly tried, but it changes lot's of stuff in the QT Code. I prefer doing it in a separate PR, as I fear the changes will be too hard to review.

The design of this PR will not be greatly impacted once I switch setSelected to use CInputCoin everywhere. Basically only this block https://github.com/bitcoin/bitcoin/pull/10202/files#diff-b2bb174788c7409b671c46ccc86034bdR2247  will need to be removed.

src/wallet/coincontrol.h
NicolasDorier: Another solution is

`bool FindKnownCoin(coinst COutPoint&, CInputCoin& output)`

kanzure:
> Its really quite unfortunate we cant properly calculate fee when using most outputs here. Maybe its possible to assume, at a minimum, a compressed pubkey and standard signature size for "normal" things like P2PH? Alternatively, could force the user to provide some additional data, like a set of pubkeys needed.

How about user gives an estimated final transaction size (minus whatever fundrawtransaction adds) when calling fundrawtransaction? Then ignore all "unsolvable" inputs or whatever.

Alternatively, what about allowing fundrawtransaction feeRate to be 0 even for "unsolvable" inputs etc in the transaction, and user sets another parameter like ignoreFee--- because perhaps the user has arranged for fee to come from somewhere else, or maybe the user plans to delete one of the outputs to convert the output to fee.

NicolasDorier:
@kanzure I think nothing has to be done to calculate fee correctly. Currently, the user can just fill dummy input of the right size into the scriptSig, this would correctly evaluate the fees.

NicolasDorier:
I rebased, and rewrote this part https://github.com/bitcoin/bitcoin/pull/10202/files#diff-b2bb174788c7409b671c46ccc86034bdR2250 to be more diff friendly so you can see it is trivially right.

NicolasDorier:
@jonasschnelli I did not managed to find good alternative to `boost::optional`.  I tried `bool FindKnownCoin(coinst COutPoint&, CInputCoin& output)` but since CInputCoin can't be in uninitialized state, it does not work.

I don't think it is such a big deal, we use `boost::optional` elsewhere and `std::optional` will eventually exist.

jonasschnelli:
Needs rebase.

NicolasDorier:
I workedaround this need for this.

domob1812:
This would indeed be very useful.  Was there a particular reason for closing this, or was it simply closed because it was no longer needed for you?  Would it make sense for me to take up the patch, rebase it and resubmit?  Or are there any technical objections to the code / approach itself?

NicolasDorier:
@domob1812 I closed it because for my projects I decided to stop using RPC's wallet API completely, and did my own wallet. It was the fastest way to reach my goal, so I did not wanted to spend time rebasing this over and over. Feel free to clone this PR and take care of it.

domob1812:
@NicolasDorier:  Thanks for the answer, that makes sense.  I also have other, more important things to work on for now, but then I'll consider taking over this PR when I run into the need to do this on my own projects.

whitebreadapple:
This was not lisas reply hacked

JeremyRand:
Am I correct in inferring that if https://github.com/bitcoin/bitcoin/pull/17211 gets merged, that would eliminate the need for this PR?  Or does this PR do something that that PR doesn't do?

PR #10204, opened 2017-04-13
[rpc] rename disconnectnode argument

jnewbery:
split off from #10143 for backport

The fact that this argument is currently called 'node' and the help text says `(string, required) The node (see getpeerinfo for nodes)` is misleading at best. It currently only accepts a string which must match on the CNode::addrName, which is the IP address/port.


gmaxwell (approved):
utACK

laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/10204/commits/883154cbcb306dcc6205fe349c7056ced1f6c5fc

PR #10212, opened 2017-04-14
Make sure parameter names in .cpp and .h files are in sync

practicalswift:
Doxygen gets confused by parameter naming mismatches.

Changed from `txnConflicted` to `vtxConflicted` in:

```c++
src/validationinterface.h:    virtual void BlockConnected(const std::shared_ptr<const CBlock> &block, const CBlockIndex *pindex, const std::vector<CTransactionRef> &vtxConflicted) {}
src/wallet/wallet.cpp:void CWallet::BlockConnected(const std::shared_ptr<const CBlock>& pblock, const CBlockIndex *pindex, const std::vector<CTransactionRef>& vtxConflicted) {
src/wallet/wallet.h:    void BlockConnected(const std::shared_ptr<const CBlock>& pblock, const CBlockIndex *pindex, const std::vector<CTransactionRef>& vtxConflicted) override;
```

Changed from `pWalletIn` to `pWallet` in:

```c++
src/wallet/feebumper.cpp:CFeeBumper::CFeeBumper(const CWallet *pWallet, const uint256 txidIn, int newConfirmTarget, bool specifiedConfirmTarget, CAmount totalFee, bool newTxReplaceable)
src/wallet/feebumper.h:    CFeeBumper(const CWallet *pWallet, const uint256 txidIn, int newConfirmTarget, bool specifiedConfirmTarget, CAmount totalFee, bool newTxReplaceable);
```

Changed from `pWalletNonConst` to `pWallet` in:

```c++
src/wallet/feebumper.cpp:bool CFeeBumper::commit(CWallet *pWallet)
src/wallet/feebumper.h:    bool commit(CWallet *pWallet);
```

### DIFF
--- a/src/validationinterface.h
+++ b/src/validationinterface.h
@@ -34,7 +34,7 @@ class CValidationInterface {
 protected:
     virtual void UpdatedBlockTip(const CBlockIndex *pindexNew, const CBlockIndex *pindexFork, bool fInitialDownload) {}
     virtual void TransactionAddedToMempool(const CTransactionRef &ptxn) {}
-    virtual void BlockConnected(const std::shared_ptr<const CBlock> &block, const CBlockIndex *pindex, const std::vector<CTransactionRef> &txnConflicted) {}
+    virtual void BlockConnected(const std::shared_ptr<const CBlock> &block, const CBlockIndex *pindex, const std::vector<CTransactionRef> &vtxConflicted) {}
     virtual void BlockDisconnected(const std::shared_ptr<const CBlock> &block) {}
     virtual void SetBestChain(const CBlockLocator &locator) {}
     virtual void UpdatedTransaction(const uint256 &hash) {}
--- a/src/wallet/feebumper.h
+++ b/src/wallet/feebumper.h
@@ -23,7 +23,7 @@ enum class BumpFeeResult
 class CFeeBumper
 {
 public:
-    CFeeBumper(const CWallet *pWalletIn, const uint256 txidIn, int newConfirmTarget, bool specifiedConfirmTarget, CAmount totalFee, bool newTxReplaceable);
+    CFeeBumper(const CWallet *pWallet, const uint256 txidIn, int newConfirmTarget, bool specifiedConfirmTarget, CAmount totalFee, bool newTxReplaceable);
     BumpFeeResult getResult() const { return currentResult; }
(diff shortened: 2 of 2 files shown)


paveljanik:
utACK https://github.com/bitcoin/bitcoin/pull/10212/commits/72f406a475b20d8a3e284e437822009599f78dd1

mchrostowski:
How does Doxygen get confused? I could not find any warnings generated for these mismatches. Could you clarify?

I was checking into these changes and wanted to confirm there aren't other inconsistencies that can be fixed in the same pull.

practicalswift:
@mchrostowski The Doxygen auto-suggest search bar gets confused in cases like this. 
See @MarcoFalke's example in [this comment](https://github.com/bitcoin/bitcoin/pull/10029#issuecomment-287616869) :-)

mchrostowski:
@practicalswift The comment you referenced is a bit misleading as this Doxygen confusion only occurs with free functions. In fact one of the changes in #10029 was not affected by the issue, `CTxMemPool::UpdateTransactionsFromBloc`, but was changed anyway.

Doxygen does not get confused on any of the method signatures fixed by this PR; a query for 'BlockConnected' gives us:

> BlockConnected
> CMainSignals::BlockConnected()
> PeerLogicValidation::BlockConnected()
> ConnectTrace::BlockConnected()
> CValidationInterface::BlockConnected()
> CWallet::BlockConnected()
> CZMQNotificationInterface::BlockConnected()

It looks like Doxygen omits the arguments altogether unless you're looking up a free function (function outside of any class).

Also, @MarcoFalke's changes still behave oddly in Doxygen search, though slightly less so since his fix was applied and the parameter names match,

> PruneBlockFilesManual
> PruneBlockFilesManual(int nManualPruneHeight): validation.cpp
> PruneBlockFilesManual(int nManualPruneHeight): validation.cpp

This behavior has (allegedly) something to do with Doxygen not grouping the functions by namespace and instead grouping by file, though both say .cpp so that doesn't make sense. I don't know exactly what causes this or how to fix it.

Given this I don't believe there's a reason to make these changes as there are many more identical issues which we are arbitrarily choosing not to fix. Let me know if you think I missed something, otherwise I'll take a shot at making submitting a review.

practicalswift:
@mchrostowski Thanks for the review! I'm closing this PR - a more complete review would be great! :-)

PR #10214, opened 2017-04-14
Abortable rescans & rpc abortrescan command

rawodb:
* Added fAbortRescan flag to CWallet to break out of rescans loops.
* Added abortrescan RPC command to exit safely out of rescans

Feedback welcomed


sipa:
Seems similar to #10208.

rawodb:
ahhh! a case of work collision!

sorry; I was following #10183 for an abort command.

MarcoFalke:
@rawodb Since you independently wrote the code, you can help with review and provide feedback on the other pull.

PR #10220, opened 2017-04-17
Experiment: test: Disable fsync in travis tests

laanwj:
Install the package "eatmydata" which disables fsync and friends, and use this while running the functional tests.

Not sure whether it applies to travis, but at least locally on a VM this made some of the tests run three times as fast.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,15 +24,15 @@ env:
 # ARM
     - HOST=arm-linux-gnueabihf PACKAGES="g++-arm-linux-gnueabihf" DEP_OPTS="NO_QT=1" CHECK_DOC=1 GOAL="install" BITCOIN_CONFIG="--enable-glibc-back-compat --enable-reduce-exports"
 # Win32
-    - HOST=i686-w64-mingw32 DPKG_ADD_ARCH="i386" DEP_OPTS="NO_QT=1" PACKAGES="python3 nsis g++-mingw-w64-i686 wine1.6 bc eatmydata" RUN_TESTS=true GOAL="install" BITCOIN_CONFIG="--enable-reduce-exports"
+    - HOST=i686-w64-mingw32 DPKG_ADD_ARCH="i386" DEP_OPTS="NO_QT=1" PACKAGES="python3 nsis g++-mingw-w64-i686 wine1.6 bc" RUN_TESTS=true GOAL="install" BITCOIN_CONFIG="--enable-reduce-exports"
 # 32-bit + dash
-    - HOST=i686-pc-linux-gnu DPKG_ADD_ARCH="i386" PACKAGES="g++-multilib bc python3-zmq eatmydata:i386" DEP_OPTS="NO_QT=1" RUN_TESTS=true GOAL="install" BITCOIN_CONFIG="--enable-zmq --enable-glibc-back-compat --enable-reduce-exports LDFLAGS=-static-libstdc++" USE_SHELL="/bin/dash"
+    - HOST=i686-pc-linux-gnu DPKG_ADD_ARCH="i386" PACKAGES="g++-multilib bc python3-zmq eatmydata:i386" DEP_OPTS="NO_QT=1" RUN_TESTS=true GOAL="install" BITCOIN_CONFIG="--enable-zmq --enable-glibc-back-compat --enable-reduce-exports LDFLAGS=\"-static-libstdc++ -L/usr/lib/libeatmydata -leatmydata\"" USE_SHELL="/bin/dash" LD_LIBRARY_PATH="/usr/lib/libeatmydata"
 # Win64
     - HOST=x86_64-w64-mingw32 DPKG_ADD_ARCH="i386" DEP_OPTS="NO_QT=1" PACKAGES="python3 nsis g++-mingw-w64-x86-64 wine1.6 bc" RUN_TESTS=true GOAL="install" BITCOIN_CONFIG="--enable-reduce-exports"
 # bitcoind
-    - HOST=x86_64-unknown-linux-gnu PACKAGES="bc python3-zmq eatmydata" DEP_OPTS="NO_QT=1 NO_UPNP=1 DEBUG=1" RUN_TESTS=true GOAL="install" BITCOIN_CONFIG="--enable-zmq --enable-glibc-back-compat --enable-reduce-exports CPPFLAGS=-DDEBUG_LOCKORDER"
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -24,15 +24,15 @@ env:
...
 # Win64
     - HOST=x86_64-w64-mingw32 DPKG_ADD_ARCH="i386" DEP_OPTS="NO_QT=1" PACKAGES="python3 nsis g++-mingw-w64-x86-64 wine1.6 bc" RUN_TESTS=true GOAL="install" BITCOIN_CONFIG="--enable-reduce-exports"
 # bitcoind
-    - HOST=x86_64-unknown-linux-gnu PACKAGES="bc python3-zmq" DEP_OPTS="NO_QT=1 NO_UPNP=1 DEBUG=1" RUN_TESTS=true GOAL="install" BITCOIN_CONFIG="--enable-zmq --enable-glibc-back-compat --enable-reduce-exports CPPFLAGS=-DDEBUG_LOCKORDER"
+    - HOST=x86_64-unknown-linux-gnu PACKAGES="bc python3-zmq eatmydata" DEP_OPTS="NO_QT=1 NO_UPNP=1 DEBUG=1" RUN_TESTS=true GOAL="install" BITCOIN_CONFIG="--enable-zmq --enable-glibc-back-compat --enable-reduce-exports CPPFLAGS=-DDEBUG_LOCKORDER"
 # No wallet
-    - HOST=x86_64-unknown-linux-gnu PACKAGES="python3 xvfb" DEP_OPTS="NO_WALLET=1" RUN_TESTS=true GOAL="install" BITCOIN_CONFIG="--enable-glibc-back-compat --enable-reduce-exports"
+    - HOST=x86_64-unknown-linux-gnu PACKAGES="python3 xvfb eatmydata" DEP_OPTS="NO_WALLET=1" RUN_TESTS=true GOAL="install" BITCOIN_CONFIG="--enable-glibc-back-compat --enable-reduce-exports"
```
MarcoFalke: functional tests are disabled for windows and no_wallet, so I don't think we need this here.

MarcoFalke:
Concept ACK:

```
*** BEFORE ***
ALL                            | ✓ Passed  | 2648 s (accumulated) 
Runtime: 713 s

*** AFTER ***
ALL                            | ✓ Passed  | 1290 s (accumulated) 
Runtime: 337 s
```

As we ignore the return code of `fsync` in our code, this should not add additional debug hassle in case the fsync fails to write everything to disk (maybe due to hardware issues)?

.travis.yml
laanwj: My intent was to add it for all the lines that have `RUN_TESTS=true`

laanwj:
> As we ignore the return code of fsync in our code, this should not add additional debug hassle in case the fsync fails to write everything to disk (maybe due to hardware issues)?

I don't think so. 

It does remind me that we could use the same technique (library interposing) to inject failures for testing to test how bitcoind handles i/o failures.

laanwj:
Argh, it fails on 32-bit linux:
```
ERROR: ld.so: object '/usr/lib/libeatmydata/libeatmydata.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS64): ignored.
```

laanwj:
I can't get it to work on 32-bit: not with 32-bit and not with 64-bit `eatmydata`.

The problem, I think, is that python runs as 64-bit, and bitcoind as 32-bit. So with the 64-bit `eatmydata` it fails on `bitcoind`, with the 32-bit one it fails on `python`.

Not sure what would be a good way to fix this. I suppose a hack that sets the environment variable BITCOIND so that *only* bitcoind is started through `eatmydata` (and not the outer script) could work.

laanwj:
Hm another idea, instead of using the wrapper/interposer, which seems to be tricky on mixed architctures, couldn't we just link bitcoind against `libeatmydata.so` directly?
On Ubuntu 16.04 `libeatmydata1` is a separate package that can be installed, this isn't the case on trusty/14.04.

16.04:
```
$ dpkg -L libeatmydata1
...
/usr/lib/x86_64-linux-gnu/libeatmydata.so.1.1.2
/usr/lib/x86_64-linux-gnu/libeatmydata.so.1
/usr/lib/x86_64-linux-gnu/libeatmydata.so
```

14.04:
```
$ dpkg -L eatmydata
...
/usr/lib/libeatmydata/libeatmydata.so
```

laanwj:
With the wrapper most works, but it still fails in `p2p-versionbits-warning.py`, as this makes bitcoind invoke another process which apparently is 64 bit:
```
p2p-versionbits-warning.py failed, Duration: 8 s
stdout:
2017-04-17 17:09:13.351000 TestFramework (INFO): Initializing test directory /tmp/testmahvm5qh/275
2017-04-17 17:09:13.609000 TestFramework.mininode (INFO): Connecting to Bitcoin Node: 127.0.0.1:13200
2017-04-17 17:09:19.543000 TestFramework (INFO): Stopping nodes
2017-04-17 17:09:21.665000 TestFramework (INFO): Cleaning up
2017-04-17 17:09:21.666000 TestFramework (INFO): Tests successful
stderr:
ERROR: ld.so: object '/usr/lib/libeatmydata/libeatmydata.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
ERROR: ld.so: object '/usr/lib/libeatmydata/libeatmydata.so' from LD_PRELOAD cannot be preloaded (wrong ELF class: ELFCLASS32): ignored.
```
Sigh...

jnewbery:
> this makes bitcoind invoke another process

The only process I can think of that `p2p-versionbits-warning.py` specifically would invoke is echo:

```
        self.extra_args = [["-alertnotify=echo %s >> \"" + self.alert_filename + "\""]]
```

This causes bitcoind to launch a thread to run echo when it detects blocks signalling unknown versionbits.

laanwj:
> Hm another idea, instead of using the wrapper/interposer, which seems to be tricky on mixed architctures, couldn't we just link bitcoind against libeatmydata.so directly?

I've tried this locally with some success, but not entirely sure how to port it to travis. Trying...

laanwj:
Oh wow that worked.

But seems @MarcoFalke 's comparison was mistaken and this doesn't actually make a difference so closing...

MarcoFalke:
@laanwj Sorry for my confirmation bias. On a first glance those numbers seemed perfectly plausible.

laanwj:
Yes it would have been plausible; apparently Travis isn't using qemu/kvm or has already configured their VMs to not pass through fsync.

PR #10224, opened 2017-04-18
[test] Add test for getaddednodeinfo

jimmysong:
Test adds a node and sees if it's in the getaddednodeinfo call

### DIFF
--- a/test/functional/net.py
+++ b/test/functional/net.py
@@ -7,15 +7,15 @@
 Tests correspond to code in rpc/net.cpp.
 """
 
-from decimal import Decimal
 import time
 
 from test_framework.test_framework import BitcoinTestFramework
-from test_framework.authproxy import JSONRPCException
 from test_framework.util import (
     assert_equal,
-    start_nodes,
+    assert_raises_jsonrpc,
     connect_nodes_bi,
+    p2p_port,
+    start_nodes,
 )
 
 
@@ -33,7 +33,7 @@ class NetTest(BitcoinTestFramework):
 
     def run_test(self):
         assert_equal(self.nodes[0].getnetworkinfo()['networkactive'], True)
-        assert_equal(self.nodes[0].getnetworkinfo()['connections'], 2) # bilateral connection
+        assert_equal(self.nodes[0].getnetworkinfo()['connections'], 2)
 
         self.nodes[0].setnetworkactive(False)
         assert_equal(self.nodes[0].getnetworkinfo()['networkactive'], False)
@@ -48,6 +48,19 @@ class NetTest(BitcoinTestFramework):
         connect_nodes_bi(self.nodes, 0, 1)
         assert_equal(self.nodes[0].getnetworkinfo()['networkactive'], True)
         assert_equal(self.nodes[0].getnetworkinfo()['connections'], 2)
+
+        # test getaddednodeinfo
+        assert_equal(self.nodes[0].getaddednodeinfo(), [])
+        # add a node (node2) to node0
+        ip_port = "127.0.0.1:{}".format(p2p_port(2))
+        self.nodes[0].addnode(ip_port, 'add')
+        # check that the node has indeed been added
+        added_nodes = self.nodes[0].getaddednodeinfo(ip_port)
+        assert_equal(len(added_nodes), 1)
+        assert_equal(added_nodes[0]['addednode'], ip_port)
+        # check that a non-existant node returns an error
+        assert_raises_jsonrpc(-24, "Node has not been added",
+                              self.nodes[0].getaddednodeinfo, '1.1.1.1')
 
 
 if __name__ == '__main__':


jonasschnelli:
Instead of creating a new test (which slightly slows down the CI/test process), you could add this to an existing test?

MarcoFalke:
Agree, we already have two scripts [[1]](https://github.com/bitcoin/bitcoin/pull/10077#issuecomment-290542590) for net.cpp tests. Choose one of them.

laanwj:
> Instead of creating a new test (which slightly slows down the CI/test process), you could add this to an existing test?

Yes it's sometimes hard to decide. On one hand, a small test granularity is better to quickly find specific problems based on what fails, on the other hand grouping tests together is better for performance.

jnewbery:
> On one hand, a small test granularity is better to quickly find specific problems based on what fails

This can also be achieved by structuring the sub-tests intelligently and making sure there are no dependencies between them (ie trying to avoid anything that carries state between the sub-tests), and adding plenty of logging so it's immediately obvious where the test failed.

+1 for adding this to an existing test script. Even better (in my opinion) if nodehandling.py and net.py can be merged into a single test script, but that's possibly more controversial.

jimmysong:
@laanwj @jonasschnelli @jnewbery @MarcoFalke Thanks for the feedback. I'll put this into another test.

jimmysong:
Test was put into net.py. Please let me know if there's anything else I need to do.

test/functional/net.py
```diff
@@ -13,9 +13,7 @@
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.authproxy import JSONRPCException
 from test_framework.util import (
-    assert_equal,
-    start_nodes,
-    connect_nodes_bi,
+    assert_equal, connect_nodes_bi, p2p_port, start_nodes,
```
MarcoFalke: git solves conflicts on a line basis, not word basis, so I'd prefer to keep those in separate lines to make it easier to solve merge conflicts (in case they pop up).

jimmysong: Fixed.

jimmysong:
@MarcoFalke Fixed the code review item.

test/functional/net.py
```diff
@@ -49,6 +50,15 @@ def run_test(self):
         assert_equal(self.nodes[0].getnetworkinfo()['networkactive'], True)
         assert_equal(self.nodes[0].getnetworkinfo()['connections'], 2)
 
+        # test getaddednodeinfo
+        # add a node (node2) to node0
+        ip_port = "127.0.0.1:{}".format(p2p_port(2))
+        peers = self.nodes[0].addnode(ip_port, 'add')
```
jnewbery: nit: no need to assign this unused `peers` variable.

I recommend you run flake8 over PRs to catch nits like this.

jnewbery (approved):
Tested ACK 61d949fa308f556aca0dda45eb38196164dfdbc5. It'd be great if you could add test cases for:

- running getaddednodeinfo before any nodes are added and asserting that an empty array is returned
- running getaddednodeinfo with an address that isn't an added node and asserting that bitcoind raises a JSONRPC error (use `assert_raises_jsonrpc()`)

Even without those tests, this improves coverage, so ACK.

test/functional/net.py
jimmysong: Good idea. Let me fix this and all the stuff you mentioned.

jimmysong:
@jnewbery changed per your suggestions.

jnewbery:
Tested ACK https://github.com/bitcoin/bitcoin/pull/10224/commits/7c8babebf24bc5b1000035f2c9d1fc73d40a5a15. Please squash commits before merge.

jimmysong:
@MarcoFalke once it's ready for merge, I'll squash.

jimmysong:
squashed and rebased

MarcoFalke:
utACK bc53752

test/functional/net.py
```diff
@@ -49,6 +49,19 @@ def run_test(self):
...
+        # add a node (node2) to node0
+        ip_port = "127.0.0.1:{}".format(p2p_port(2))
+        self.nodes[0].addnode(ip_port, 'add')
+        # check that the node has indeed been added
+        added_nodes = self.nodes[0].getaddednodeinfo(ip_port)
+        assert_equal(len(added_nodes), 1)
+        assert_equal(added_nodes[0]['addednode'], ip_port)
+        # check that a non-existant node returns an error
+        assert_raises_jsonrpc(-24, "Node has not been added",
+                              self.nodes[0].getaddednodeinfo, '1.1.1.1')
```
MarcoFalke: nit: No need to limit lines at 80 chars because some tool tells you so. We don't have a strict guideline about the maximum length, so feel free to use any length that makes sense. 120 might be a rough rule of thumb.

PR #10225, opened 2017-04-18
[test] Add aborttrescan tests

kallewoof:
This PR adds tests for the new `abortrescan` RPC command.

A new function `ref_node` was added to `util.py`, which works like `start_node` except it never actually launches the process. This is used to get two `node` objects in python which work separately from each other, in order to make two simultaneous requests (`importprivkey` and `abortrescan`).

### DIFF
--- a/test/functional/test_runner.py
+++ b/test/functional/test_runner.py
@@ -108,6 +108,7 @@ BASE_SCRIPTS= [
     'rpcnamedargs.py',
     'listsinceblock.py',
     'p2p-leaktests.py',
+    'import-abort-rescan.py',
 ]
 
 ZMQ_SCRIPTS = [
--- a/test/functional/import-abort-rescan.py
+++ b/test/functional/import-abort-rescan.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Test wallet import RPCs.
+
+Test rescan behavior of importprivkey when aborted. The test ensures that:
+1. The abortrescan command indeed stops the rescan process.
+2. Subsequent rescan catches the aborted address UTXO
+"""
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import (assert_equal, get_rpc_proxy)
+from decimal import Decimal
+import threading # for bg importprivkey
+import time      # for sleep
+
+class ImportAbortRescanTest(BitcoinTestFramework):
+    def __init__(self):
+        super().__init__()
+        self.setup_clean_chain = True
+
+    def run_test(self):
+        # Generate for BTC
+        assert_equal(self.nodes[0].getbalance(), 0)
+        assert_equal(self.nodes[1].getbalance(), 0)
+        self.nodes[0].generate(300)
+        assert_equal(self.nodes[1].getbalance(), 0)
+        # Make blocks with spam to cause rescan delay
+        for i in range(5):
+            for j in range(5):
+                self.nodes[0].sendtoaddress(self.nodes[0].getnewaddress(), 0.1)
+            self.nodes[0].generate(10)
+        addr = self.nodes[0].getnewaddress()
+        privkey = self.nodes[0].dumpprivkey(addr)
+        self.nodes[0].sendtoaddress(addr, 0.123)
+        self.nodes[0].generate(10) # mature tx
+        self.sync_all()
+
+        # Import this address in the background ...
+        node1ref = get_rpc_proxy(self.nodes[1].url, 1, timeout=600)
+        importthread = threading.Thread(target=node1ref.importprivkey, args=[privkey])
+        importthread.start()
+        # ... then abort rescan; try a bunch until abortres becomes true,
+        # because we will start checking before above thread starts processing
+        for i in range(2000):
+            time.sleep(0.001)
+            abortres = self.nodes[1].abortrescan()
+            if abortres: break
+        assert abortres # if false, we failed to abort
+        # import should die soon
+        for i in range(10):
+            time.sleep(0.1)
+            deadres = not importthread.isAlive()
+            if deadres: break
+
+        assert deadres # if false, importthread did not die soon enough
+        assert_equal(self.nodes[1].getbalance(), 0.0)
+
+        # Import a different address and let it run
+        self.nodes[1].importprivkey(self.nodes[0].dumpprivkey(self.nodes[0].getnewaddress()))
+        # Expect original privkey to now also be discovered and added to balance
+        assert_equal(self.nodes[1].getbalance(), Decimal("0.123"))
+
+if __name__ == "__main__":
+    ImportAbortRescanTest().main()


test/functional/test_runner.py
```diff
@@ -108,6 +108,7 @@
     'rpcnamedargs.py',
     'listsinceblock.py',
     'p2p-leaktests.py',
+    'import-abort-rescan.py',           # ~17s
```
jonasschnelli: I think we should remove the time hint, otherwise this will become "the standard" and, with that, the "keep-it-updated" problem will follow.

jonasschnelli:
Concept ACK.
Nice way with the ref node & threading.

test/functional/test_runner.py
kallewoof: Hm, yeah, you're right. Removing.

jnewbery:
Thanks for opening this PR to cover abortrescan. Definitely worth doing.

I don't like the change you've made to `start_node()`. Adding an optional parameter to a function called `start_node()` which causes the function to *not* start a node is really an abuse of the function.

I'd much prefer us to move toward having an encapsulated class for a test node, where we could have multiple rpc connections, including asynchronous rpc connections if required. I've been trying to push towards that model in #10082, but I haven't had much luck in attracting reviewers for that PR or the linked PR. I think adding random functionality into util.py is unmaintainable and is moving in the wrong direction.

So concept ACK for covering this with a functional test, but please lets not add more to util.py.

kallewoof:
@jnewbery 

> I don't like the change you've made to start_node(). Adding an optional parameter to a function called start_node() which causes the function to not start a node is really an abuse of the function.

Yeah, I was wondering if I should have reversed it (i.e. ref_node takes a launch bool and start_node just proxies).

As for encapsulate with multiple rpc connections, that sounds great but the functionality is, as you say, not in there. My change is minimal and works, so maybe it's an okay for now until #10082 is ready? (I'll gladly review btw. I just gotta wake up first..)

kallewoof:
#10082 seems like a big project (I'd love to help btw), so I am proposing a solution until that is resolved here. `start_node` now calls `ref_node` which now has an optional `spawnproc` flag. This means `start_node` always starts a node, and `ref_node` does both, depending on the flag.

Unsquashed history: [1](https://github.com/kallewoof/bitcoin/commit/4dc2c402392cb516c32a7c06720a657a02a755a8) → [2](https://github.com/kallewoof/bitcoin/commit/28c2abf9ea8f86479dab9c7c76a380cf303b26cb) → [3⊰1](https://github.com/kallewoof/bitcoin/commit/04e8620b9f4586ee91688d5617588c0f7728ab05)

jnewbery:
@kallewoof thanks for being so accepting of my feedback! I don't like NACKing PRs, but I really want to try to not put any additional complexity in util.py if we can help it.

Have you had a look at getblocktemplate_longpoll.py ? That's doing something similar to this test where an additional asynchronous RPC thread is required. I've tried to rewrite your testcase in the same style here: https://github.com/jnewbery/bitcoin/tree/pr10225. Can you take a look and tell me what you think? I'd prefer to do it this way than adding more complexity to the mainline start_node() function. If we can get #10082 merged, then we can then look at adding an asynchronous RPC thread to the TestNode class so it's available more generally.

kallewoof:
@jnewbery not a worry at all -- you don't have to hold back the punches with me. I am learning a lot from the feedback I get from you guys. :)

Gotcha on the no-touching-util.py. I will look at getblocktemplate_longpoll and see if I can adapt. Worst case I have two options: I can drop this PR until #10082 or I can put the ref_node code into the test itself with a # TODO and we simply rip it out when it's time to replace.

As for the test, yes, I am trying to keep the time to run as low as possible and I may have put it a bit too low (upping the range in one or both of the top for loops should stabilize it, I think).

kallewoof:
@jnewbery Wow, the solution in getblocktemplate_longpoll.py was so much cleaner. I switched to that and dropped some commits. I also upped the range to hopefully address the intermittent fails you experienced.

Unsquashed history: [~~1~~](https://github.com/kallewoof/bitcoin/commit/4dc2c402392cb516c32a7c06720a657a02a755a8) → [2](https://github.com/kallewoof/bitcoin/commit/28c2abf9ea8f86479dab9c7c76a380cf303b26cb) → [~~3⊱1~~](https://github.com/kallewoof/bitcoin/commit/04e8620b9f4586ee91688d5617588c0f7728ab05) → [4⊱2](https://github.com/kallewoof/bitcoin/commit/c2321ae106ef4d38774f8764ef200a26a6d28a2c) → [5⊱2](https://github.com/kallewoof/bitcoin/commit/cf85ccf3769d699cece9cc37a0bec2d2a492eb90) → [6⊱2](https://github.com/kallewoof/bitcoin/commit/91ccad51b48979ee4d97625e7c38b426042c3135)

jnewbery:
Looks better, but the test is still failing more often than not for me. I ran the test 20 times (with 4 tests running in parallel):

```
TEST                                 | STATUS    | DURATION

import-abort-rescan.py --portseed=1  | ✓ Passed  | 11 s
import-abort-rescan.py --portseed=10 | ✖ Failed  | 11 s
import-abort-rescan.py --portseed=11 | ✖ Failed  | 11 s
import-abort-rescan.py --portseed=12 | ✖ Failed  | 11 s
import-abort-rescan.py --portseed=13 | ✖ Failed  | 11 s
import-abort-rescan.py --portseed=14 | ✖ Failed  | 11 s
import-abort-rescan.py --portseed=15 | ✖ Failed  | 11 s
import-abort-rescan.py --portseed=16 | ✖ Failed  | 11 s
import-abort-rescan.py --portseed=17 | ✖ Failed  | 11 s
import-abort-rescan.py --portseed=18 | ✓ Passed  | 11 s
import-abort-rescan.py --portseed=19 | ✖ Failed  | 11 s
import-abort-rescan.py --portseed=2  | ✖ Failed  | 12 s
import-abort-rescan.py --portseed=20 | ✖ Failed  | 11 s
import-abort-rescan.py --portseed=3  | ✖ Failed  | 12 s
import-abort-rescan.py --portseed=4  | ✖ Failed  | 13 s
import-abort-rescan.py --portseed=5  | ✖ Failed  | 11 s
import-abort-rescan.py --portseed=6  | ✖ Failed  | 11 s
import-abort-rescan.py --portseed=7  | ✖ Failed  | 11 s
import-abort-rescan.py --portseed=8  | ✖ Failed  | 11 s
import-abort-rescan.py --portseed=9  | ✖ Failed  | 11 s

ALL                                  | ✖ Failed  | 224 s (accumulated) 
```

(ignore the portseed argument - that's just a hack to allow instances of the same test to be run by the test_runner. The value of portseed is ignored).

I'm not why this fails so much for me, but seems to pass for you and Travis.

kallewoof:
@jnewbery Is there an easy way to run the test like that? E.g. 20 times 4 in parallel?

kallewoof:
Still seeing intermittent failures. Marking this WIP until this is fully resolved.

jnewbery:
> Is there an easy way to run the test like that? E.g. 20 times 4 in parallel?

You can do this by changing the `BASE_SCRIPTS` list in test_runner.py to be the same test multiple times. test_runner will automatically remove duplicates, but if you add `--portseed=x` to the test name, then it will run them as separate tests. The dummy portseed parameter is overridden by an actual portseed further down in test_runner.

> This all feels very flakey though. :(

Indeed! Is it true to say there's a tradeoff between making the test case last longer and making it more robust? If that's the case I think you should err towards making it run longer and perhaps add it as an extended script rather than a base script.

jnewbery:
This test is still failing intermittently for me in two different ways:

```
2017-05-02 18:51:29.922000 TestFramework (ERROR): Assertion failed
  File "/home/ubuntu/bitcoin/test/functional/test_framework/test_framework.py", line 146, in main
    self.run_test()
  File "./import-abort-rescan.py", line 50, in run_test
    assert abortres # if false, we failed to abort
AssertionError
```

and:

```
  File "/home/ubuntu/bitcoin/test/functional/test_framework/test_framework.py", line 146, in main
    self.run_test()
  File "./import-abort-rescan.py", line 58, in run_test
    assert_equal(self.nodes[1].getbalance(), 0.0)
  File "/home/ubuntu/bitcoin/test/functional/test_framework/util.py", line 408, in assert_equal
AssertionError: not(0.12300000 == 0.0)
```

This is also causing our Jenkins build machine to fail occasionally (2 times out of 30 builds)

@laanwj - was this merged accidentally? There haven't been any reviews/ACKs. Can we revert it and get it reviewed before remerging?

laanwj:
> @laanwj - was this merged accidentally? 

I think so, sorry @kallewoof, needs a new PR now.

jnewbery:
@kallewoof - this has been backed out by #10327 . Please open a new PR so this test can be reviewed before being merged back in. A few suggestions for making this less flakey:

- increase the number of generated blocks by at least an order of magnitude. Run the test many times on a fast machine, with bitcoind's datadir in /dev/shm. I think this passes on Travis because bitcoind runs slowly so you have a larger window for the `abortrescan` RPC to hit.
- check the debug logs to see how long the rescan actually takes so we're not guessing on what a safe window is.
- I think this test should be in the extended_test list rather than the base_tests list.
- alternatively, tack this test onto a longer test script which already has a long chain, for example pruning.py. Seems a bit untidy, but should guarantee that the rescan takes a long time.

kallewoof:
No problem - I will do proper testing and make a new PR once done. Sorry for the trouble!

PR #10226, opened 2017-04-18
wallet: Use boost to more portably ensure -wallet specifies only a filename

luke-jr:


laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/10226/commits/a4186dd23944b0b803523632e1578d5c4ebb9e57
We definitely shouldn't be hardcoding path characters.

jtimon:
utACK a4186dd

paveljanik:
utACK a4186dd

Issue #10209, opened 2017-04-14
Stalled shutdown


jonasschnelli:
Not sure if this is a real issue but just had a stalled shutdown on a node:

The node is running the master at `c7e73eafa139c29a38f73ab697e2e967a386908d`  from Wed Apr 5 11:29:19 2017.

As you can see, it took > 7 mins from a requested shutdown to a net thread exit.
Debug log:
```shell
2017-04-14 07:34:12 Shutdown: In progress...
2017-04-14 07:41:47 net thread exit
2017-04-14 07:41:47 sending version (103 bytes) peer=193
2017-04-14 07:41:47 msghand thread exit
2017-04-14 07:41:47 send version message: version 70015, blocks=461809, us=[::]:0, peer=193
2017-04-14 07:41:47 opencon thread exit
2017-04-14 07:41:47 Flushed 69008 addresses to peers.dat  109ms
2017-04-14 07:41:47 disconnecting peer=0
2017-04-14 07:41:47 disconnecting peer=1
2017-04-14 07:41:47 disconnecting peer=5
2017-04-14 07:41:47 disconnecting peer=7
2017-04-14 07:41:47 disconnecting peer=19
2017-04-14 07:41:47 disconnecting peer=24
2017-04-14 07:41:47 disconnecting peer=25
2017-04-14 07:41:47 disconnecting peer=38
2017-04-14 07:41:47 disconnecting peer=58
2017-04-14 07:41:47 disconnecting peer=137
2017-04-14 07:41:47 disconnecting peer=169
2017-04-14 07:41:47 disconnecting peer=178
2017-04-14 07:41:47 disconnecting peer=183
2017-04-14 07:41:47 disconnecting peer=185
2017-04-14 07:41:47 disconnecting peer=190
2017-04-14 07:41:47 disconnecting peer=193
2017-04-14 07:41:48 Dumped mempool: 0.046245s to copy, 0.174033s to dump
2017-04-14 07:41:48 Shutdown: done
```

The backtraces during the stalling are:

```cpp
(gdb) thread apply all bt

Thread 5 (Thread 0x7f777ffff700 (LWP 2034)):
#0  pthread_cond_wait@@GLIBC_2.3.2 () at ../nptl/sysdeps/unix/sysv/linux/x86_64/pthread_cond_wait.S:185
#1  0x00007f77b76419eb in BGThread (this=0x7f77b8dd94b0) at leveldb/util/env_posix.cc:574
[...]

Issue #10140, opened 2017-04-02
UI Hangs


achow101:
### Describe the issue

The UI frequently hangs for me, either requiring me to wait a very long time for it to become responsive again, or killing bitcoin-qt and restarting it. 

#### If so, please list the steps to reproduce below:
1. Start Bitcoin Core
2. Wait a few minutes
3. Minimize to the taskbar
4. Open the main window by clicking the icon in the taskbar and clicking "Show / Hide"
5. Attempt to do anything with the UI. I have found that opening the debug window and attempting to change tabs will usually cause it to hang.

### Expected behaviour

It shouldn't hang, and I shouldn't need to kill it.

### Actual behaviour

It hangs, and usually requires force killing in order make the node come back online (it seems to go offline when it hangs). I am unable to do anything with the GUI.

### What version of bitcoin-core are you using?

v0.14.99.0-f34cdcb

### Machine specs:
- OS: Ubuntu 16.04
- CPU: i7-6800k
- RAM: 32 GB

burnsd03:
Thanks

laanwj:
This unfortunately doesn't give enough information to reproduce or diagnose the issue.

- Is there anything being logged in debug.log when this happened? This might have the same cause as #10005 and #5851
- Is the process (any of its threads) still using CPU?
- If possible: attach a debugger when this happens and do `thread apply all bt` (in case of gdb) to get an idea what hangs

achow101:
> Is there anything being logged in debug.log when this happened? This might have the same cause as #10005 and #5851

No, nothing unusual.

> Is the process (any of its threads) still using CPU?

It doesn't seem like it. %CPU in the system monitor is 0 and top didn't even list the process.

> If possible: attach a debugger when this happens and do `thread apply all bt` (in case of gdb) to get an idea what hangs

Is it normal that after I attach gdb that gdb will pause immediately and let me enter commands?
Here is all of the output of that command after I attached it when it hung:
```
Thread 24 (Thread 0x7f12e37fe700 (LWP 25080)):
#0  0x00007f134cc9db5d in poll () at ../sysdeps/unix/syscall-template.S:84
#1  0x00007f134bde838c in ?? () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
#2  0x00007f134bde849c in g_main_context_iteration () from /lib/x86_64-linux-gnu/libglib-2.0.so.0
[...]

PR #10228, opened 2017-04-18
build: regenerate bitcoin-config.h as necessary

theuni:
This fixes issues like #10140 and #10209.

The cause is that #9921 introduced new defines, but without re-running autogen.sh, they wouldn't be found. This change makes sure that autoheader is re-run (as it's supposed to be) any time configure.ac is touched.

Also, this undoes #4322, as it would introduce an infinite loop otherwise. I suspect that these convenience targets get very little use anymore, but we can always add back hard-coded ones if necessary.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -76,9 +76,6 @@ $(BITCOIN_WIN_INSTALLER): all-recursive
 	  echo error: could not build $@
 	@echo built $@
 
-$(if $(findstring src/,$(MAKECMDGOALS)),$(MAKECMDGOALS), none): FORCE
-	$(MAKE) -C src $(patsubst src/%,%,$@)
-
 $(OSX_APP)/Contents/PkgInfo:
 	$(MKDIR_P) $(@D)
 	@echo "APPL????" > $@
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -462,6 +462,14 @@ DISTCLEANFILES = obj/build.h
 
 EXTRA_DIST = $(CTAES_DIST)
 
+
+config/bitcoin-config.h: config/stamp-h1
+	@$(MAKE) -C $(top_builddir) $(subdir)/$(@)
+config/stamp-h1: $(top_srcdir)/$(subdir)/config/bitcoin-config.h.in $(top_builddir)/config.status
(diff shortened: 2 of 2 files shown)


laanwj:
> Also, this undoes #4322, as it would introduce an infinite loop otherwise. I suspect that these convenience targets get very little use anymore, but we can always add back hard-coded ones if necessary.

Which ones does this prevent use of? I use `make src/bitcoind` a lot.

theuni:
@laanwj Only obscure targets. The bins are still listed there explicitly: https://github.com/bitcoin/bitcoin/blob/master/Makefile.am#L165, so ```make src/bitcoind``` still works.

laanwj:
> @laanwj Only obscure targets. The bins are still listed there explicitly: 

Sounds good to me then.

laanwj:
> This change makes sure that autoheader is re-run (as it's supposed to be) any time configure.ac is touched.

Something that makes me wonder: shouldn't the *entire* process re-run when configure.ac is touched? Or is that the effect of this - was autoheader the only step missing for that?

theuni:
Yes, that should be the case now. For some reason automake doesn't push the autoheader rules down to subdirs, so it was never re-triggered if you build from src/.

Thinking about it now, we should probably backport this for 0.14.2 as well, since switching branches is one of the most common ways this problem can show up.

laanwj:
> Thinking about it now, we should probably backport this for 0.14.2 as well, since switching branches is one of the most common ways this problem can show up.

Good point - I don't switch branches anymore for this reason (among others). I just have a git-worktree checkout per branch.

laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/10228/commits/91ab8f5a997dd1c838e2216fe1e676471bf680a3

luke-jr:
Any idea why we need to explicitly do this? Normally autotools takes care of it automagically...

practicalswift:
utACK 91ab8f5a997dd1c838e2216fe1e676471bf680a3

Thanks for fixing this @theuni

sipa:
Concept ACK

theuni:
@luke-jr I'm not sure. I suppose it's just an autotools bug that they don't generate the rules for subdirs.

PR #10229, opened 2017-04-19
Tests: Add test for getdifficulty

jimmysong:
Test added to blockchain.py as adding a new test to reduce test run time.

### DIFF
--- a/test/functional/blockchain.py
+++ b/test/functional/blockchain.py
@@ -6,6 +6,10 @@
 
 Test the following RPCs:
     - gettxoutsetinfo
+    - getdifficulty
+    - getbestblockhash
+    - getblockhash
+    - getblockheader
     - verifychain
 
 Tests correspond to code in rpc/blockchain.cpp.
@@ -40,6 +44,7 @@ class BlockchainTest(BitcoinTestFramework):
     def run_test(self):
         self._test_gettxoutsetinfo()
         self._test_getblockheader()
+        self._test_getdifficulty()
         self.nodes[0].verifychain(4, 0)
 
     def _test_gettxoutsetinfo(self):
@@ -57,7 +62,8 @@ class BlockchainTest(BitcoinTestFramework):
     def _test_getblockheader(self):
         node = self.nodes[0]
 
-        assert_raises_jsonrpc(-5, "Block not found", node.getblockheader, "nonsense")
+        assert_raises_jsonrpc(-5, "Block not found",
+                              node.getblockheader, "nonsense")
 
         besthash = node.getbestblockhash()
         secondbesthash = node.getblockhash(199)
@@ -78,6 +84,12 @@ class BlockchainTest(BitcoinTestFramework):
         assert isinstance(header['version'], int)
         assert isinstance(int(header['versionHex'], 16), int)
         assert isinstance(header['difficulty'], Decimal)
+
+    def _test_getdifficulty(self):
+        difficulty = self.nodes[0].getdifficulty()
+        # 1 hash in 2 should be valid, so difficulty should be 1/2**31
+        # binary => decimal => binary math is why we do this check
+        assert abs(difficulty * 2**31 - 1) < 0.0001
 
 if __name__ == '__main__':
     BlockchainTest().main()


test/functional/blockchain.py
```diff
@@ -57,7 +64,8 @@ def _test_gettxoutsetinfo(self):
     def _test_getblockheader(self):
         node = self.nodes[0]
 
-        assert_raises_jsonrpc(-5, "Block not found", node.getblockheader, "nonsense")
+        assert_raises_jsonrpc(-5, "Block not found",
```
paveljanik: Why this change?

test/functional/blockchain.py
```diff
@@ -40,8 +40,15 @@ def setup_network(self, split=False):
     def run_test(self):
         self._test_gettxoutsetinfo()
         self._test_getblockheader()
+        self._test_getdifficulty()
         self.nodes[0].verifychain(4, 0)
 
+    def _test_getdifficulty(self):
+        res = self.nodes[0].getdifficulty()
```
paveljanik: Mention the `getdifficulty` RPC test in the top level comment or remove the list of tested methods there.

test/functional/blockchain.py
```diff
@@ -40,8 +40,15 @@ def setup_network(self, split=False):
     def run_test(self):
         self._test_gettxoutsetinfo()
         self._test_getblockheader()
+        self._test_getdifficulty()
         self.nodes[0].verifychain(4, 0)
 
+    def _test_getdifficulty(self):
```
paveljanik: Can you move the def after the `_test_getblockheader` so it is ordered the same as it is called?

test/functional/blockchain.py
```diff
@@ -40,8 +40,15 @@ def setup_network(self, split=False):
...
+        res = self.nodes[0].getdifficulty()
+        # 1 hash in 2 should be valid, so difficulty should be 1/2**31
+        # binary => decimal => binary math is why we round
+        assert_equal(round(res * 2**31, 4), 1.0)
```
paveljanik: Ugh.

paveljanik (changes requested):

test/functional/blockchain.py
jimmysong: Yep, will do.

test/functional/blockchain.py
jimmysong: adding to the comment, thanks!

test/functional/blockchain.py
jimmysong: flake8 warning. I can change this back?

test/functional/blockchain.py
jimmysong: Is there a cleverer way to do this?

jimmysong:
Fixed issues, rebased to master.

paveljanik (approved):
ACK 821dd5e

test/functional/blockchain.py
```diff
@@ -6,6 +6,7 @@
 
 Test the following RPCs:
     - gettxoutsetinfo
+    - getdifficulty
     - verifychain
```
jnewbery: Might as well fix this docstring while you're at it. The following RPCs are also tested:

- getbestblockhash
- getblockhash
- getblockheader

test/functional/blockchain.py
jnewbery: My personal preference is to ignore the 'line too long' warning. Nothing wrong with going over 79 character lines in my opinion (but we don't have a style guide for functional tests, so entirely up to you)

test/functional/blockchain.py
```diff
@@ -79,5 +82,11 @@ def _test_getblockheader(self):
         assert isinstance(int(header['versionHex'], 16), int)
         assert isinstance(header['difficulty'], Decimal)
 
+    def _test_getdifficulty(self):
+        res = self.nodes[0].getdifficulty()
```
jnewbery: I'd prefer a more descriptive name than `res`, perhaps `difficulty`? (or even combine with the line below)

jnewbery (commented):
Looks good. Tested ACK ba2b295093d5f5cad4701f9ca86b2c83e60c3ef9.

A few nits which you can take or leave.

jnewbery (approved):

test/functional/blockchain.py
jimmysong: good idea.

test/functional/blockchain.py
jimmysong: Makes sense.

jimmysong:
Fixed nits and squashed.

PR #10239, opened 2017-04-20
Make Boost use std::atomic internally

sipa:
Boost has its own implementation of a spinlock and atomic counters inside its shared_ptr implementation, which is used by (at least) the signals2 implementation. By default, it prefers its own implementations over C++11 atomics.

Add two macro definitions to make it choose the std::atomics based implementations instead.

These macros are supported since Boost 1.56, and the behaviour is default since Boost 1.63.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -676,6 +676,10 @@ AX_BOOST_PROGRAM_OPTIONS
 AX_BOOST_THREAD
 AX_BOOST_CHRONO
 
+dnl Boost 1.56 through 1.62 allow using std::atomic instead of its own atomic
+dnl counter implementations. In 1.63 and later the std::atomic approach is default.
+m4_pattern_allow(DBOOST_AC_USE_STD_ATOMIC) dnl otherwise it's treated like a macro
+BOOST_CPPFLAGS="-DBOOST_SP_USE_STD_ATOMIC -DBOOST_AC_USE_STD_ATOMIC $BOOST_CPPFLAGS"
 
 if test x$use_reduce_exports = xyes; then
   AC_MSG_CHECKING([for working boost reduced exports])


configure.ac
```diff
@@ -236,7 +236,7 @@ if test "x$CXXFLAGS_overridden" = "xno"; then
   AX_CHECK_COMPILE_FLAG([-Wunused-local-typedef],[CXXFLAGS="$CXXFLAGS -Wno-unused-local-typedef"],,[[$CXXFLAG_WERROR]])
   AX_CHECK_COMPILE_FLAG([-Wdeprecated-register],[CXXFLAGS="$CXXFLAGS -Wno-deprecated-register"],,[[$CXXFLAG_WERROR]])
 fi
-CPPFLAGS="$CPPFLAGS -DHAVE_BUILD_INFO -D__STDC_FORMAT_MACROS"
+CPPFLAGS="$CPPFLAGS -DHAVE_BUILD_INFO -D__STDC_FORMAT_MACROS -DBOOST_SP_USE_STD_ATOMIC -DBOOST_\A\C_USE_STD_ATOMIC"
```
laanwj: `-DBOOST_\A\C_USE_STD_ATOMIC` that looks strange

sipa: Very much so, but I don't know how else to prevent autoconf from interpreting it as a macro.

laanwj: @theuni any idea ^^?

laanwj:
Concept ACK, I think using the same atomics everywhere (those of the compiler) is a good thing.

sipa:
Apparently Boost will use std::atomic by default since 1.63.

configure.ac
theuni: I think wrapping the whole line in [ ] might shut it up?

theuni:
Concept ACK. Are you sure these are the only two necessary?

sipa:
I grepped the source code for other BOOST_*_USE things, but nothing seems relevant.

configure.ac
sipa: I had tried that already.

sipa: The error is:

```
configure.ac:239: error: possibly undefined macro: DBOOST_AC_USE_STD_ATOMIC
      If this token and others are legitimate, please use m4_pattern_allow.
      See the Autoconf documentation.
autoreconf: /usr/bin/autoconf failed with exit status: 1
```

theuni: Ok, the recommended m4_pattern_allow hack works. Seems reasonable, there's no way that will ever become a macro.

Also, please stick this in BOOST_CPPFLAGS (after the AX_BOOST_* checks) rather than in the global one:
```
m4_pattern_allow(DBOOST_AC_USE_STD_ATOMIC) dnl otherwise it's treated like a macro
BOOST_CPPFLAGS="-DBOOST_SP_USE_STD_ATOMIC -DBOOST_AC_USE_STD_ATOMIC $BOOST_CPPFLAGS"
```

sipa:
Incorporated @theuni's suggestions.

theuni:
Thanks. utACK 394ccf76adb016b48908cd9a0fa08a1663a7ddba

I cloned the boost smart_ptr repo and verified that these are the only defines needed, and that there are no uses of them in any cpp files (duh, it's a header-only lib).

gmaxwell (approved):
utACK

PR #10240, opened 2017-04-20
Add HD wallet auto-restore functionality

jonasschnelli:
This PR ensures that we always check the keypool during SyncTransaction and, that the lookup window for HD restore is large enough (currently +20 keys).
Once the keypool size falls below the min gap limit (currently 20 keys) and the keypool can't be extended (locked wallet), the wallet then temporary pauses synchronisation.

### unencrypted wallets
For unencrypted wallets, the key pool will always be topped up. During sync (SyncTransaction), we check for used keypool keys and mark all keys up to the matched key as used. Additionally we topup the keypool to ensure we not fall below the gap limit (20 keys).

### encrypted wallets in non pruning mode
Same as above, but, If we hit the gap limit with an encrypted wallet, we can't topup the keypool. In that case, we just pause the sync (not the node, only the wallet).
Once the user unlocks the wallet over `walletpassphrase`, we rescan down to the point where we have stopped previously to ensure we don't miss possible funds.

### encrypted wallets in pruned mode
Same as above, but, we also stop requesting and connecting blocks. The full node will "pause" in this case until the user did unlock his wallet (== topup, rescan, etc.).

This has a little bit of complexity and requires careful reviews.

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -480,6 +480,9 @@ void FindNextBlocksToDownload(NodeId nodeid, unsigned int count, std::vector<con
     // Make sure pindexBestKnownBlock is up to date, we'll need it.
     ProcessBlockAvailability(nodeid);
 
+    if (isBlockRequestsPaused())
+        return;
+
     if (state->pindexBestKnownBlock == NULL || state->pindexBestKnownBlock->nChainWork < chainActive.Tip()->nChainWork) {
         // This peer has nothing interesting.
         return;
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -747,7 +747,7 @@ PeerLogicValidation::PeerLogicValidation(CConnman* connmanIn) : connman(connmanI
     recentRejects.reset(new CRollingBloomFilter(120000, 0.000001));
 }
 
-void PeerLogicValidation::BlockConnected(const std::shared_ptr<const CBlock>& pblock, const CBlockIndex* pindex, const std::vector<CTransactionRef>& vtxConflicted) {
+void PeerLogicValidation::BlockConnected(const std::shared_ptr<const CBlock>& pblock, const CBlockIndex* pindex, const std::vector<CTransactionRef>& vtxConflicted, bool &requestPause) {
     LOCK(cs_main);
 
     std::vector<uint256> vOrphanErase;
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -78,6 +78,38 @@ std::string COutput::ToString() const
     return strprintf("COutput(%s, %d, %d) [%s]", tx->GetHash().ToString(), i, nDepth, FormatMoney(tx->tx->vout[i].nValue));
 }
 
+class CAffectedKeysVisitor : public boost::static_visitor<void> {
+private:
+    const CKeyStore &keystore;
+    std::vector<CKeyID> &vKeys;
+
+public:
+    CAffectedKeysVisitor(const CKeyStore &keystoreIn, std::vector<CKeyID> &vKeysIn) : keystore(keystoreIn), vKeys(vKeysIn) {}
+
+    void Process(const CScript &script) {
+        txnouttype type;
+        std::vector<CTxDestination> vDest;
+        int nRequired;
+        if (ExtractDestinations(script, type, vDest, nRequired)) {
+            BOOST_FOREACH(const CTxDestination &dest, vDest)
+                boost::apply_visitor(*this, dest);
+        }
+    }
+
+    void operator()(const CKeyID &keyId) {
+        if (keystore.HaveKey(keyId))
+            vKeys.push_back(keyId);
+    }
+
+    void operator()(const CScriptID &scriptId) {
+        CScript script;
+        if (keystore.GetCScript(scriptId, script))
+            Process(script);
+    }
+
+    void operator()(const CNoDestination &none) {}
+};
+
 const CWalletTx* CWallet::GetWalletTx(const uint256& hash) const
 {
     LOCK(cs_wallet);
@@ -3385,38 +3417,6 @@ void CWallet::ListLockedCoins(std::vector<COutPoint>& vOutpts)
 }
 
 /** @} */ // end of Actions
-
-class CAffectedKeysVisitor : public boost::static_visitor<void> {
-private:
-    const CKeyStore &keystore;
-    std::vector<CKeyID> &vKeys;
-
-public:
-    CAffectedKeysVisitor(const CKeyStore &keystoreIn, std::vector<CKeyID> &vKeysIn) : keystore(keystoreIn), vKeys(vKeysIn) {}
-
-    void Process(const CScript &script) {
-        txnouttype type;
-        std::vector<CTxDestination> vDest;
-        int nRequired;
-        if (ExtractDestinations(script, type, vDest, nRequired)) {
-            BOOST_FOREACH(const CTxDestination &dest, vDest)
-                boost::apply_visitor(*this, dest);
-        }
-    }
-
-    void operator()(const CKeyID &keyId) {
-        if (keystore.HaveKey(keyId))
-            vKeys.push_back(keyId);
-    }
-
-    void operator()(const CScriptID &scriptId) {
-        CScript script;
-        if (keystore.GetCScript(scriptId, script))
-            Process(script);
-    }
-
-    void operator()(const CNoDestination &none) {}
-};
 
 void CWallet::GetKeyBirthTimes(std::map<CTxDestination, int64_t> &mapKeyBirth) const {
     AssertLockHeld(cs_wallet); // mapKeyMetadata
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -1171,7 +1171,6 @@ void CWallet::TransactionAddedToMempool(const CTransactionRef& ptx) {
     SyncTransaction(ptx);
 }
 
-void CWallet::BlockConnected(const std::shared_ptr<const CBlock>& pblock, const CBlockIndex *pindex, const std::vector<CTransactionRef>& vtxConflicted) {
     LOCK2(cs_main, cs_wallet);
     // TODO: Temporarily ensure that mempool removals are notified before
     // connected transactions.  This shouldn't matter, but the abandoned
@@ -1186,6 +1185,7 @@ void CWallet::BlockConnected(const std::shared_ptr<const CBlock>& pblock, const 
     }
     for (size_t i = 0; i < pblock->vtx.size(); i++) {
         SyncTransaction(pblock->vtx[i], pindex, i);
+void CWallet::BlockConnected(const std::shared_ptr<const CBlock>& pblock, const CBlockIndex *pindex, const std::vector<CTransactionRef>& vtxConflicted, bool &requestPause) {
     }
 }
 
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -3175,6 +3175,7 @@ void CWallet::ReserveKeyFromKeyPool(int64_t& nIndex, CKeyPool& keypool, bool int
                 setKeyPool.erase(id);
                 assert(keypool.vchPubKey.IsValid());
                 LogPrintf("keypool reserve %d\n", nIndex);
+                CheckKeypoolMinSize();
                 return;
             }
         }
@@ -3440,6 +3441,18 @@ void CReserveKey::ReturnKey()
     vchPubKey = CPubKey();
 }
 
+bool CWallet::CheckKeypoolMinSize() {
+    LOCK(cs_wallet);
+    size_t extKeypoolSize = KeypoolCountExternalKeys();
+    if (IsHDEnabled() && (extKeypoolSize < HD_RESTORE_KEYPOOL_SIZE_MIN || (setKeyPool.size()-extKeypoolSize) < HD_RESTORE_KEYPOOL_SIZE_MIN)) {
+        // if the remaining keypool size is below the gap limit, refuse to continue with the sync
+        fSyncPausedUntilKeypoolExt = true;
+        LogPrintf("%s: Keypool ran below min size, pause wallet sync\n", __func__);
+        return false;
+    }
+    return true;
+}
+
 void CWallet::MarkReserveKeysAsUsed(const CKeyID& keyId)
 {
     LOCK(cs_wallet);
@@ -3488,13 +3501,7 @@ void CWallet::MarkReserveKeysAsUsed(const CKeyID& keyId)
         LogPrintf("%s: Topping up keypool failed (locked wallet), pausing transaction processing\n", __func__);
     }
 
-    size_t extKeypoolSize = KeypoolCountExternalKeys();
-    if (IsHDEnabled() && (extKeypoolSize < HD_RESTORE_KEYPOOL_SIZE_MIN || (setKeyPool.size()-extKeypoolSize) < HD_RESTORE_KEYPOOL_SIZE_MIN)) {
-
-        // if the remaining keypool size is below the gap limit, refuse to cintinue with the sync
-        fSyncPausedUntilKeypoolExt = true;
-        LogPrintf("%s: Topping up keypool failed (locked wallet)\n", __func__);
-    }
+    CheckKeypoolMinSize();
 }
 
 void CWallet::GetAllReserveKeys(std::set<CKeyID>& setAddress) const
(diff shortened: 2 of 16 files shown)


ryanofsky:
I know there was discussion about this in IRC last week, but I didn't follow closely. Would it be better to review this PR now, or hold off for more changes?

jonasschnelli:
I'm currently implementing the changes we discussed in last weeks RIC meeting. A review makes more sense after the overhaul.

jonasschnelli:
Completely rewrote this PR. Keeping the WIP tag for now.
The current implementation takes care of encrypted and unencrypted wallets (also in conjunction with pruning).

### unencrypted wallets
For unencrypted wallets, the key pool will always be topped up. During sync (SyncTransaction), we check for used keypool keys and mark all keys up to the matched key as used. Additionally we topup the keypool to ensure we not fall below the gap limit (20 keys).

## encrypted wallets in non pruning mode
Same as above, but, If we hit the gap limit with an encrypted wallet, we can't topup the keypool. In that case, we just pause the sync (not the node, only the wallet).
Once the user unlocks the wallet over `walletpassphrase`, we rescan down to the point where we have stopped previously to ensure we don't miss possible funds.

## encrypted wallets in pruned mode
Same as above, but, we also stop requesting and connecting blocks. The full node will "pause" in this case until the user did unlock his wallet (== topup, rescan, etc.).

This has a little bit of complexity and requires careful reviews.
Thanks for early feedback.

jonasschnelli:
All tests are passing now.
Removed WIP tag.
Overhauled PR description

src/wallet/wallet.cpp
```diff
@@ -970,6 +1004,22 @@ bool CWallet::AddToWalletIfInvolvingMe(const CTransactionRef& ptx, const CBlockI
         if (fExisted && !fUpdate) return false;
         if (fExisted || IsMine(tx) || IsFromMe(tx))
         {
+            // check if we need to flag used keypool keys
+            std::set<CKeyID> setKeyPool;
+            GetAllReserveKeys(setKeyPool);
```
jonasschnelli: I guess this is very inefficient and something like #10235 should allow to speed up things.

test/functional/wallet-hd.py
```diff
@@ -75,10 +75,12 @@ def run_test (self):
 
         self.log.info("Restore backup ...")
         self.stop_node(1)
-        os.remove(self.options.tmpdir + "/node1/regtest/wallet.dat")
+        # we need to delete the complete regtest directory
+        # otherwise node1 would auto-recover all funds in flag the keypool keys as used
```
ryanofsky: In commit "Ensure we pause wallet sync when the keypool requires extension"

Is this change meant to be part of this commit? I don't think there's an interaction with hdignoregaplimit here.

src/wallet/wallet.cpp
```diff
@@ -1417,6 +1418,57 @@ bool CWallet::IsHDEnabled() const
     return !hdChain.masterKeyID.IsNull();
 }
 
+void CWallet::EventuallyRescanAfterKeypoolTopUp() {
+    if (fSyncPausedUntilKeypoolExt) {
+
+        // for now, enable block requests and tip updates via the states
+        // this will only be sufficient as long as only a single wallet adjusts these stats
+        // otherwise the net-/validation-logic needs to poll the state over a signal (or similar)
```
ryanofsky: In commit "Ensure we pause wallet sync when the keypool requires extension"

It should be possible to handle this without any polling or notifications by making `fPauseBlockRequests` and `fPauseTipUpdates` counters instead of bools. The counters would need to be incremented whenever `fSyncPausedUntilKeypoolExt` changed from false to true in any wallet, and decremented when it changed from true to false. Block requesting or tip updating would be paused whenever its respective counter was greater than 0.

src/wallet/wallet.cpp
```diff
@@ -1163,33 +1164,51 @@ void CWallet::SyncTransaction(const CTransactionRef& ptx, const CBlockIndex *pin
...
+        // the notification that the conflicted transaction was evicted.
+
+        for (const CTransactionRef& ptx : vtxConflicted) {
+            SyncTransaction(ptx);
+        }
+        for (size_t i = 0; i < pblock->vtx.size(); i++) {
+            SyncTransaction(pblock->vtx[i], pindex, i);
+        }
+        if (fSyncPausedUntilKeypoolExt) {
+            requestPause = true;
```
jonasschnelli: As you mention, it would be much simpler to call `setBlockRequestsPaused()`, but, wouldn't it be a massive layer violation? Communicating through the signals doesn't require that the wallet needs to know what the full node parts is really doing. But maybe im overcomplicating things here.

src/wallet/wallet.cpp
```diff
@@ -3764,6 +3764,7 @@ std::string CWallet::GetWalletHelpString(bool showDebug)
         strUsage += HelpMessageOpt("-flushwallet", strprintf("Run a thread to flush wallet periodically (default: %u)", DEFAULT_FLUSHWALLET));
         strUsage += HelpMessageOpt("-privdb", strprintf("Sets the DB_PRIVATE flag in the wallet db environment (default: %u)", DEFAULT_WALLET_PRIVDB));
         strUsage += HelpMessageOpt("-walletrejectlongchains", strprintf(_("Wallet will not create transactions that violate mempool chain limits (default: %u)"), DEFAULT_WALLET_REJECT_LONG_CHAINS));
+        strUsage += HelpMessageOpt("-hdignoregaplimit", strprintf(_("Ignores the minimum keypool-size warning for HD restore (default: %u)"), DEFAULT_IGNORE_HD_MIN_KEYPOOL_SIZE));
```
jonasschnelli: The `-hdignoregaplimit` is for debug purposes only. Our tests run with a keypool size of 1 and therefor would cause troubles running post this PR without a "ignore" option, and I don't think it matter how we "ignore" the gap limit and over a startup argument seems the be the simplest way.

src/wallet/wallet.cpp
jonasschnelli: Yes. Thats a good idea. Do you think this should be done in this PR or can we tackle this once we have multiple places where the sync may get paused?
Counters may also introduce the risk of unwillingly increasing it twice (while only decrease it once).

src/wallet/wallet.cpp
```diff
@@ -3349,14 +3365,62 @@ void CReserveKey::ReturnKey()
...
+    auto it = std::begin(setKeyPool);
+
+    bool foundInternal = false;
+    int64_t foundIndex = -1;
+    for (const int64_t& id : setKeyPool) {
+        CKeyPool keypool;
+        if (!walletdb.ReadPool(id, keypool))
+            throw std::runtime_error(std::string(__func__) + ": read failed");
+
+        if (keypool.vchPubKey.GetID() == keyId) {
```
jonasschnelli: It would also require to keep the `fInternal` flags in memory (for the address book internal/external change flagging). This is independently addresses in #10238.

src/wallet/wallet.cpp
```diff
@@ -1417,6 +1418,57 @@ bool CWallet::IsHDEnabled() const
...
+            if (fPruneMode)
+            {
+                CBlockIndex *block = chainActive.Tip();
+                while (block && block->pprev && (block->pprev->nStatus & BLOCK_HAVE_DATA) && block->pprev->nTx > 0 && pindexRescan != block)
+                    block = block->pprev;
+
+                if (pindexRescan != block) {
+                    const static std::string pruneError = "Prune: last wallet synchronisation goes beyond pruned data. You need to -reindex (download the whole blockchain again in case of pruned node)";
+                    uiInterface.ThreadSafeMessageBox(pruneError, "", CClientUIInterface::MSG_ERROR);
+                    StartShutdown();
```
jonasschnelli: I guess it already tries to avoid this situation by disabling the block downloads if the wallet is encrypted and the node runes in prune-mode.
But there is no guarantee that we won't prune bellow the wallets best block,.. this why this shutdown/protection is here.

jonasschnelli:
Needed rebase.
Addresses most of @ryanofsky points.

src/wallet/wallet.cpp
```diff
@@ -1002,6 +1002,22 @@ bool CWallet::AddToWalletIfInvolvingMe(const CTransactionRef& ptx, const CBlockI
         if (fExisted && !fUpdate) return false;
         if (fExisted || IsMine(tx) || IsFromMe(tx))
         {
+            // check if we need to flag used keypool keys
```
ryanofsky: >Would expand this comment

Done in 2a203b73fd2d7b016cd5f55977135a15a982414d

src/wallet/wallet.cpp
ryanofsky: >Do you think this should be done in this PR or can we tackle this once we have multiple places where the sync may get paused?

Maybe, this PR is already pretty complicated, so I wouldn't want to tack this on in extra commits, but it could be nice if implemented to replace the existing "Add option to pause..." commits.

src/wallet/wallet.cpp
```diff
@@ -970,6 +1004,22 @@ bool CWallet::AddToWalletIfInvolvingMe(const CTransactionRef& ptx, const CBlockI
...
+            std::set<CKeyID> setKeyPool;
```
ryanofsky: > Maybe rename this variable. (It shadows CWallet::setKeyPool, which makes that variable more annoying to grep for.)

Done in 2a203b73fd2d7b016cd5f55977135a15a982414d

src/wallet/wallet.cpp
```diff
@@ -1163,33 +1164,51 @@ void CWallet::SyncTransaction(const CTransactionRef& ptx, const CBlockIndex *pin
...
-    // connected transactions.  This shouldn't matter, but the abandoned
-    // state of transactions in our wallet is currently cleared when we
-    // receive another notification and there is a race condition where
-    // notification of a connected conflict might cause an outside process
-    // to abandon a transaction and then have it inadvertently cleared by
-    // the notification that the conflicted transaction was evicted.
-
-    for (const CTransactionRef& ptx : vtxConflicted) {
+    {
+        LOCK2(cs_main, cs_wallet);
```
ryanofsky: >No need to add the {} and extra indentation

(unchanged)

src/wallet/wallet.cpp
```diff
@@ -970,6 +1004,22 @@ bool CWallet::AddToWalletIfInvolvingMe(const CTransactionRef& ptx, const CBlockI
         if (fExisted && !fUpdate) return false;
         if (fExisted || IsMine(tx) || IsFromMe(tx))
         {
+            // check if we need to flag used keypool keys
+            std::set<CKeyID> setKeyPool;
+            GetAllReserveKeys(setKeyPool);
+            // loop though all outputs
+            BOOST_FOREACH(const CTxOut& txout, tx.vout) {
```
ryanofsky: >Maybe avoid BOOST_FOREACH here and below.

Done in 2a203b73fd2d7b016cd5f55977135a15a982414d

src/wallet/wallet.cpp
```diff
@@ -1163,33 +1164,51 @@ void CWallet::SyncTransaction(const CTransactionRef& ptx, const CBlockIndex *pin
...
 }
 
 void CWallet::BlockDisconnected(const std::shared_ptr<const CBlock>& pblock) {
-    LOCK2(cs_main, cs_wallet);
+    if (fSyncPausedUntilKeypoolExt) return;
 
-    for (const CTransactionRef& ptx : pblock->vtx) {
-        SyncTransaction(ptx);
+    {
+        LOCK2(cs_main, cs_wallet);
```
ryanofsky: >Again unnecessary {} and indentation make the diff messy here in BlockDisconnected

(unchanged)

src/wallet/wallet.h
```diff
@@ -1106,6 +1108,9 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
        caller must ensure the current wallet version is correct before calling
        this function). */
     bool SetHDMasterKey(const CPubKey& key);
+
+    bool IsSyncPausedUntilKeypoolExt() { return fSyncPausedUntilKeypoolExt; }
```
ryanofsky: >These functions are never called anywhere, and appear to be deleted in the next commit. Would remove them from this commit.

Removed in 5b70a2289e96f0ce8c49678e613845572d8203bc

sipa:
Needs rebase.

test/functional/wallet-hd-restore.py
```diff
@@ -0,0 +1,162 @@
...
+        self.sync_all()
+        self.nodes[0].generate(1)
+        self.sync_all()
+
+        assert_equal(self.nodes[1].getbalance(), 3)
+        assert_equal(self.nodes[1].listtransactions()[0]['category'], "receive")
+
+        print("Testing with unencrypted wallet")
+        self.stop_node(1)
+        shutil.rmtree(tmpdir + "/node1/regtest")
```
jnewbery: please don't do this! It erases the debug.log file. Are you able to be more targeted in what you're erasing?

test/functional/wallet-hd-restore.py
```diff
@@ -0,0 +1,162 @@
...
+        connect_nodes_bi(self.nodes,0,1)
+        for _ in range(10):
+            addr = self.nodes[1].getnewaddress()
+
+        self.nodes[0].generate(101)
+        addr = self.nodes[1].getnewaddress()
+        assert_equal(self.nodes[1].validateaddress(addr)['hdkeypath'], "m/0'/0'/11'")
+
+        rawch = self.nodes[1].getrawchangeaddress()
+        txid = self.nodes[0].sendtoaddress(addr, 1)
```
jnewbery: `txid` isn't used. No need to assign it here or below.

test/functional/wallet-hd.py
jnewbery: I'm also confused about why wallet-hd.py needs to be updated.

jnewbery (commented):
Needs rebase. I've added a lot of mainly nitty comments. The most important thing to address is to make this compatible with multiwallet.

gmaxwell:
Will begin code review post rebase.

Question about the parameters--  Why 20? that is very small,  I am not aware of any reason to not have it be more like-- say-- 10,000.   With a large number the pause is hopefully seldom/never hit.  Electrum uses a very small number because its a single user wallet and has to send all these addresses for processing to a remote server.   In a commercial application getting reordering beyond 20 would be very easy. (e.g hand out 30 keys to different users and only the first one makes a payment).

jonasschnelli:
Agree that 20 is probably way to low.
It's just a single constant and I worry more about getting the PR on a level where everything works as expected. Changing the 20 to 1000 then is simple... once the PR is stable, we can also test performance better (maybe between 20 and 10k).

test/functional/wallet-hd-restore.py
```diff
@@ -0,0 +1,162 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Test Hierarchical Deterministic wallet function."""
```
sipa: Please address.

test/functional/wallet-hd-restore.py
sipa: Please address.

src/wallet/wallet.cpp
```diff
@@ -3725,6 +3793,19 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
 
     RegisterValidationInterface(walletInstance);
 
+    // HD Restore: Make sure we always have a reasonable keypool size if HD is enabled
+    if (walletInstance->IsHDEnabled()) {
+        if (walletInstance->IsCrypted()) {
+            InitWarning(_("Your are using an encrypted HD wallet. In case you recover a HD wallet, you may miss incomming or outgoing funds."));
+        }
+        else {
```
sipa: Nit: else on the same line as `}`

src/net_processing.cpp
```diff
@@ -480,6 +480,9 @@ void FindNextBlocksToDownload(NodeId nodeid, unsigned int count, std::vector<con
     // Make sure pindexBestKnownBlock is up to date, we'll need it.
     ProcessBlockAvailability(nodeid);
 
+    if (isBlockRequestsPaused())
```
sipa: Please address.

sipa (commented):
Some comments already.

Overall question: why is any of this conditional on HD being enabled? I think we want to mark user and topup any key seen on the network at any time.

jonasschnelli:
Closing in favor of #10830

PR #10241, opened 2017-04-20
Allow tests to pass even when stderr got populated

jonasschnelli:


laanwj:
I thought about this a bit: apparently some people use the stderr output to determine whether a test emitted e.g. sanitizer warnings.
Possible idea: add a "WARNINGS" status, that is interpreted as PASS for the final tally but does show that something unexpected was printed.

jonasschnelli:
Added a new status "Passed with warnings" which gets set when the test passed but stderr was not empty. The test will be flagged as "successful" for that new status.

test/functional/test_runner.py
```diff
@@ -407,6 +409,9 @@ def __repr__(self):
         if self.status == "Passed":
             color = BLUE
             glyph = TICK
+        if self.status == "Passed with warnings":
+            color = ORANGE
```
```
NameError: name 'ORANGE' is not defined

Comment (jnewbery) [inline test/functional/test_runner.py]: `ORANGE` would need to be a constant defined at the top of this file (see around L47)

File: test/functional/test_runner.py [hunk 2]
```diff
@@ -407,6 +409,9 @@ def __repr__(self):
         if self.status == "Passed":
             color = BLUE
             glyph = TICK
+        if self.status == "Passed with warnings":
```
jnewbery: `elif` please

jnewbery (commented):
Couple of comments inline.

If you add a new status string, you'll need to update print justification (otherwise I think your new category will just be truncated). Justification is on L337 in `print_results()` and L422 in `TestResult.__repr__()`.

Now that we have 4 categories, it may make sense to define all the categories as constants at the top of the file and have a new `MAX_STATUS_WIDTH` constant as well.

One you've done that, can you add a screenshot of what this looks like?

MarcoFalke:
Concept ACK, but I will close this pull in case no one picks it up in the next couple of days.

MarcoFalke:
Closing for now. Feel free to pick this up again.

PR #10102, opened 2017-03-27
Multiprocess bitcoin

ryanofsky:
**This is based on # + #29409.** The non-base commits are:

- [`c0ae4dfa004` test: Increase feature_block.py and feature_taproot.py timeouts](https://github.com/bitcoin/bitcoin/pull/10102/commits/c0ae4dfa0048cc576ec4b112dd0275e2683f52d7)
- [`f274686c60e` test: Fix multiprocess test for unclean shutdown on kill](https://github.com/bitcoin/bitcoin/pull/10102/commits/f274686c60e783be469823bcaf32acee221def7b)
- [`32937a817c0` util: Add util::Result workaround to be compatible with libmultiprocess](https://github.com/bitcoin/bitcoin/pull/10102/commits/32937a817c00b419b965adc173d85fe48cfa109c)
- [`12ae360d84d` test: characterize btcsignals callback destruction timing](https://github.com/bitcoin/bitcoin/pull/10102/commits/12ae360d84d8fe4adadd51e7f6a6a89d7183ed7e)
- [`d5c694aeb15` btcsignals: destroy callbacks eagerly on disconnect](https://github.com/bitcoin/bitcoin/pull/10102/commits/d5c694aeb15b8d02d594239fe0bf0db21bea5015)
- [`66821528e4e` btcsignals: fall back to atomic_load/store without __cpp_lib_atomic_shared_ptr](https://github.com/bitcoin/bitcoin/pull/10102/commits/66821528e4ec431d38451c4c2dcdda2c19779d51)
- [`8fc97929d96` multiprocess: Add capnp serialization code for bitcoin types](https://github.com/bitcoin/bitcoin/pull/10102/commits/8fc97929d966468e77c8a57e5fc286375d1b5f66)
- [`897a9c3b3e3` interfaces, refactor: Change WalletLoader::restoreWallet parameter order](https://github.com/bitcoin/bitcoin/pull/10102/commits/897a9c3b3e3c3a170b14c1a7863915037a019424)
- [`d5bfd35ca8e` multiprocess: Add capnp wrapper for Wallet interface](https://github.com/bitcoin/bitcoin/pull/10102/commits/d5bfd35ca8e24ffd3b8fd0c939e6d65754676453)
- [`99442afacb2` multiprocess: Add capnp wrapper for Node interface](https://github.com/bitcoin/bitcoin/pull/10102/commits/99442afacb28a2f580a3bfd67fd4fae6b4a20841)
[...]

PR #10242, opened 2017-04-20
[qt] Don't call method on null WalletModel object

ryanofsky:
This doesn't crash currently because the method doesn't access any object
members, but this behavior is fragile and incompatible with #10102.


jonasschnelli:
Uh. Yes. That makes much more sense.
Thanks for fixing.

utACK 7ba55396a0b9e013bd08db0678eb917af5b1f0b2

luke-jr:
Ouch, utACK.

ryanofsky:
If anyone wants to add a tested ack, easiest way to test this is to start bitcoin-qt with `-walletrbf` flag, and make sure "Request-By-Fee" checkbox is checked, then start without the `-walletrbf` flag and make sure the checkbox is unchecked.

You can find the "Request-By-Fee" checkbox under "Send", "Transaction Fee", "Choose..."

mariodian:
tACK (OS X 10.12.4)

PR #9680, opened 2017-02-03
Unify CWalletTx construction

ryanofsky:
Two commits:

- `Construct CWalletTx objects in CommitTransaction` moves a bunch of CWalletTx initialization into CWallet::CommitTransaction to dedup some code and avoid future inconsistencies in how wallet transactions are created.
- `Get rid of CWalletTx default constructor` does what is described and eliminates the possibility of empty transaction entries being inadvertently created by mapWallet[hash] accesses.

Both of these changes were originally part of #9381

PR #10244, opened 2017-04-20
Refactor: separate gui from wallet and node

ryanofsky:
This is a refactoring PR that does not change behavior in any way. This change:

1. Creates abstract [`Node`](https://github.com/ryanofsky/bitcoin/blob/pr/ipc-local/src/interface/node.h) and [`Wallet`](https://github.com/ryanofsky/bitcoin/blob/pr/ipc-local/src/interface/wallet.h) interfaces in [`src/interface/`](https://github.com/ryanofsky/bitcoin/tree/pr/ipc-local/src/interface)
1. Updates Qt code to call the new interfaces. This largely consists of diffs of the form:

```diff
-    InitLogging();
-    InitParameterInteraction();
+    node.initLogging();
+    node.initParameterInteraction();
```

This change allows followup PR #10102 (makes `bitcoin-qt` control `bitcoind` over an IPC socket) to work without any significant updates to Qt code. Additionally:

* It provides a single place to describe the interface between GUI and daemon code.
* It can make better GUI testing possible, because Node and Wallet objects have virtual methods that can be overloaded for mocking.
* It can be used to help make the GUI more responsive (see https://github.com/bitcoin/bitcoin/issues/10504)

Other notes:

* I used python scripts [hide-globals.py](https://github.com/ryanofsky/home/blob/master/src/2017/hide-globals/hide-globals.py) and [replace-syms.py](https://github.com/ryanofsky/home/blob/master/src/2017/hide-globals/replace-syms.py) to identify all the places where Qt code was accessing libbitcoin global variables and calling functions accessing those global variables.
* These changes were originally part of #10102. Thanks to @JeremyRubin for the suggestion of splitting them out.

Commits:

- [`ea73b84d2d` Add src/interface/README.md](https://github.com/bitcoin/bitcoin/pull/10244/commits/ea73b84d2ddde22487dee0f71d7a619051e067f2)
- [`71e0d90876` Remove direct bitcoin calls from qt/bitcoin.cpp](https://github.com/bitcoin/bitcoin/pull/10244/commits/71e0d90876efd11e2a4aeb8f3f806c5a1fd54b42)
- [`c0f2756be5` Remove direct bitcoin calls from qt/optionsmodel.cpp](https://github.com/bitcoin/bitcoin/pull/10244/commits/c0f2756be517feddacd7c6b89b9faa888b3fef8e)
- [`3d619e9d36` Remove direct bitcoin calls from qt/bitcoingui.cpp](https://github.com/bitcoin/bitcoin/pull/10244/commits/3d619e9d3658e36cba4a19a5bed33e5538317b27)
- [`c2f672fb19` Remove direct bitcoin calls from qt/utilitydialog.cpp](https://github.com/bitcoin/bitcoin/pull/10244/commits/c2f672fb1960399389dea9cdd8f76d7156c2c88b)
- [`5fba3af21e` Remove direct bitcoin calls from qt/splashscreen.cpp](https://github.com/bitcoin/bitcoin/pull/10244/commits/5fba3af21e44ab7552c57782de430c1f4cfd6697)
- [`fe6f27e6ea` Remove direct bitcoin calls from qt/clientmodel.cpp](https://github.com/bitcoin/bitcoin/pull/10244/commits/fe6f27e6ea68a139d3a98b30a53706008ef8b132)
- [`d7c2c95948` Remove direct bitcoin calls from qt/intro.cpp](https://github.com/bitcoin/bitcoin/pull/10244/commits/d7c2c9594897c39df6739b92610dfb5a7a1cb3ec)
- [`e0b66a3b7c` Remove direct bitcoin calls from qt/peertablemodel.cpp](https://github.com/bitcoin/bitcoin/pull/10244/commits/e0b66a3b7c5d3a079636d61fcf611bb6b36c7bc1)
- [`3034a462a5` Remove direct bitcoin calls from qt/bantablemodel.cpp](https://github.com/bitcoin/bitcoin/pull/10244/commits/3034a462a5d30144cf0ec801d07f0c8c36d560f3)
- [`582daf6d22` Remove direct bitcoin calls from qt/rpcconsole.cpp](https://github.com/bitcoin/bitcoin/pull/10244/commits/582daf6d22da5394d02a12003b9542d9f5865ae2)
- [`90d4640b7e` Remove direct bitcoin calls from qt/optionsdialog.cpp](https://github.com/bitcoin/bitcoin/pull/10244/commits/90d4640b7eff3154a0750c5acb52d39bd41e0bbb)
- [`a0704a8996` Remove most direct bitcoin calls from qt/walletmodel.cpp](https://github.com/bitcoin/bitcoin/pull/10244/commits/a0704a8996bb950ae3c4d5b5a30e9dfe34cde1d3)
- [`827de038ab` Remove direct bitcoin calls from qt/coincontroldialog.cpp](https://github.com/bitcoin/bitcoin/pull/10244/commits/827de038ab6fa58aa3d46151eb2f8dc6add7743e)
- [`3ec2ebcd9b` Remove direct bitcoin calls from qt/addresstablemodel.cpp](https://github.com/bitcoin/bitcoin/pull/10244/commits/3ec2ebcd9b4beb4277f1f4791c6acbc538784f70)
- [`3cab2ce5f9` Remove direct bitcoin calls from qt/paymentserver.cpp](https://github.com/bitcoin/bitcoin/pull/10244/commits/3cab2ce5f9e159ad5a2e9ed682f28121b5248580)
- [`58845587e1` Remove direct bitcoin calls from qt transaction table files](https://github.com/bitcoin/bitcoin/pull/10244/commits/58845587e11140e81f087a74c3db76a4d1fc3a1a)
- [`e872c93ee8` Remove direct bitcoin access from qt/guiutil.cpp](https://github.com/bitcoin/bitcoin/pull/10244/commits/e872c93ee87477130fb877da1a536b4c693bbab9)
- [`56f33ca349` Remove direct bitcoin calls from qt/sendcoinsdialog.cpp](https://github.com/bitcoin/bitcoin/pull/10244/commits/56f33ca349b3721a15fce3bf0b6d4fd7fd788970)
- [`9a61eed1fc` Use WalletBalances struct in Qt](https://github.com/bitcoin/bitcoin/pull/10244/commits/9a61eed1fcc16ddcedc315045f470e1958b0760b)
- [`9960137697` Add developer notes about blocking GUI code](https://github.com/bitcoin/bitcoin/pull/10244/commits/996013769711bd507cdcd6dde88cbd59fcd4fbad)

### DIFF
--- a/src/interface/node.cpp
+++ b/src/interface/node.cpp
@@ -45,6 +45,7 @@ class NodeImpl : public Node
         Shutdown();
     }
     void startShutdown() override { StartShutdown(); }
+    bool shutdownRequested() override { return ShutdownRequested(); }
     void mapPort(bool use_upnp) override
     {
         if (use_upnp) {
@@ -58,6 +59,14 @@ class NodeImpl : public Node
     std::unique_ptr<Handler> handleInitMessage(InitMessageFn fn) override
     {
         return MakeHandler(::uiInterface.InitMessage.connect(fn));
+    }
+    std::unique_ptr<Handler> handleMessageBox(MessageBoxFn fn) override
+    {
+        return MakeHandler(::uiInterface.ThreadSafeMessageBox.connect(fn));
+    }
+    std::unique_ptr<Handler> handleQuestion(QuestionFn fn) override
+    {
+        return MakeHandler(::uiInterface.ThreadSafeQuestion.connect(fn));
     }
 };
 
--- a/src/interface/node.cpp
+++ b/src/interface/node.cpp
@@ -60,6 +60,7 @@ class NodeImpl : public Node
     void initLogging() override { InitLogging(); }
     void initParameterInteraction() override { InitParameterInteraction(); }
     std::string getWarnings(const std::string& type) override { return GetWarnings(type); }
+    uint32_t getLogCategories() override { return ::logCategories; }
     bool baseInitialize() override
     {
         return AppInitBasicSetup() && AppInitParameterInteraction() && AppInitSanityChecks() &&
@@ -227,6 +228,11 @@ class NodeImpl : public Node
     std::vector<std::string> listRpcCommands() override { return ::tableRPC.listCommands(); }
     void rpcSetTimerInterfaceIfUnset(RPCTimerInterface* iface) override { RPCSetTimerInterfaceIfUnset(iface); }
     void rpcUnsetTimerInterface(RPCTimerInterface* iface) override { RPCUnsetTimerInterface(iface); }
+    bool getUnspentOutput(const COutPoint& output, Coin& coin) override
+    {
+        LOCK(::cs_main);
+        return ::pcoinsTip->GetCoin(output, coin);
+    }
     std::vector<std::unique_ptr<Wallet>> getWallets() override
     {
 #ifdef ENABLE_WALLET
--- a/src/interface/node.cpp
+++ b/src/interface/node.cpp
@@ -7,6 +7,9 @@
 #include <chainparams.h>
 #include <init.h>
 #include <interface/handler.h>
+#include <net.h>
+#include <netaddress.h>
+#include <netbase.h>
 #include <scheduler.h>
 #include <ui_interface.h>
 #include <util.h>
@@ -24,6 +27,8 @@ class NodeImpl : public Node
         gArgs.ParseParameters(argc, argv);
     }
     void readConfigFile(const std::string& conf_path) override { gArgs.ReadConfigFile(conf_path); }
+    bool softSetArg(const std::string& arg, const std::string& value) override { return gArgs.SoftSetArg(arg, value); }
+    bool softSetBoolArg(const std::string& arg, bool value) override { return gArgs.SoftSetBoolArg(arg, value); }
     void selectParams(const std::string& network) override { SelectParams(network); }
     void initLogging() override { InitLogging(); }
     void initParameterInteraction() override { InitParameterInteraction(); }
@@ -40,6 +45,16 @@ class NodeImpl : public Node
         Shutdown();
     }
     void startShutdown() override { StartShutdown(); }
+    void mapPort(bool use_upnp) override
+    {
+        if (use_upnp) {
+            StartMapPort();
+        } else {
+            InterruptMapPort();
+            StopMapPort();
+        }
+    }
+    bool getProxy(Network net, proxyType& proxy_info) override { return GetProxy(net, proxy_info); }
     std::unique_ptr<Handler> handleInitMessage(InitMessageFn fn) override
     {
         return MakeHandler(::uiInterface.InitMessage.connect(fn));
--- a/src/interface/node.cpp
+++ b/src/interface/node.cpp
@@ -4,6 +4,7 @@
 
 #include <interface/node.h>
 
+#include <addrdb.h>
 #include <chain.h>
 #include <chainparams.h>
 #include <init.h>
@@ -101,6 +102,14 @@ class NodeImpl : public Node
                         GetNodeStateStats(std::get<0>(node_stats).nodeid, std::get<2>(node_stats));
                 }
             }
+            return true;
+        }
+        return false;
+    }
+    bool getBanned(banmap_t& banmap) override
+    {
+        if (g_connman) {
+            g_connman->GetBanned(banmap);
             return true;
         }
         return false;
--- a/src/interface/node.cpp
+++ b/src/interface/node.cpp
@@ -5,6 +5,7 @@
 #include <interface/node.h>
 
 #include <addrdb.h>
+#include <amount.h>
 #include <chain.h>
 #include <chainparams.h>
 #include <init.h>
@@ -28,6 +29,7 @@
 #include <config/bitcoin-config.h>
 #endif
 #ifdef ENABLE_WALLET
+#include <wallet/wallet.h>
 #define CHECK_WALLET(x) x
 #else
 #define CHECK_WALLET(x) throw std::logic_error("Wallet function called in non-wallet build.")
@@ -36,8 +38,6 @@
 #include <atomic>
 #include <boost/thread/thread.hpp>
 #include <univalue.h>
-
-class CWallet;
 
 namespace interface {
 namespace {
@@ -185,6 +185,8 @@ class NodeImpl : public Node
         }
     }
     bool getNetworkActive() override { return g_connman && g_connman->GetNetworkActive(); }
+    unsigned int getTxConfirmTarget() override { CHECK_WALLET(return ::nTxConfirmTarget); }
+    CAmount getMaxTxFee() override { return ::maxTxFee; }
     UniValue executeRpc(const std::string& command, const UniValue& params, const std::string& uri) override
     {
         JSONRPCRequest req;
@@ -196,6 +198,18 @@ class NodeImpl : public Node
     std::vector<std::string> listRpcCommands() override { return ::tableRPC.listCommands(); }
     void rpcSetTimerInterfaceIfUnset(RPCTimerInterface* iface) override { RPCSetTimerInterfaceIfUnset(iface); }
     void rpcUnsetTimerInterface(RPCTimerInterface* iface) override { RPCUnsetTimerInterface(iface); }
+    std::vector<std::unique_ptr<Wallet>> getWallets() override
+    {
+#ifdef ENABLE_WALLET
+        std::vector<std::unique_ptr<Wallet>> wallets;
+        for (CWalletRef wallet : ::vpwallets) {
+            wallets.emplace_back(MakeWallet(*wallet));
+        }
+        return wallets;
+#else
+        throw std::logic_error("Node::getWallets() called in non-wallet build.");
+#endif
+    }
     std::unique_ptr<Handler> handleInitMessage(InitMessageFn fn) override
     {
         return MakeHandler(::uiInterface.InitMessage.connect(fn));
--- a/src/interface/node.cpp
+++ b/src/interface/node.cpp
@@ -1,0 +1,53 @@
+// Copyright (c) 2018 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include <interface/node.h>
+
+#include <chainparams.h>
+#include <init.h>
+#include <interface/handler.h>
+#include <scheduler.h>
+#include <ui_interface.h>
+#include <util.h>
+#include <warnings.h>
+
+#include <boost/thread/thread.hpp>
+
+namespace interface {
+namespace {
+
+class NodeImpl : public Node
+{
+    void parseParameters(int argc, const char* const argv[]) override
+    {
+        gArgs.ParseParameters(argc, argv);
+    }
+    void readConfigFile(const std::string& conf_path) override { gArgs.ReadConfigFile(conf_path); }
+    void selectParams(const std::string& network) override { SelectParams(network); }
+    void initLogging() override { InitLogging(); }
+    void initParameterInteraction() override { InitParameterInteraction(); }
+    std::string getWarnings(const std::string& type) override { return GetWarnings(type); }
+    bool baseInitialize() override
+    {
+        return AppInitBasicSetup() && AppInitParameterInteraction() && AppInitSanityChecks() &&
+               AppInitLockDataDirectory();
+    }
+    bool appInitMain() override { return AppInitMain(); }
+    void appShutdown() override
+    {
+        Interrupt();
+        Shutdown();
+    }
+    void startShutdown() override { StartShutdown(); }
+    std::unique_ptr<Handler> handleInitMessage(InitMessageFn fn) override
+    {
+        return MakeHandler(::uiInterface.InitMessage.connect(fn));
+    }
+};
+
+} // namespace
+
+std::unique_ptr<Node> MakeNode() { return MakeUnique<NodeImpl>(); }
+
+} // namespace interface
--- a/src/interface/node.cpp
+++ b/src/interface/node.cpp
@@ -4,6 +4,7 @@
 
 #include <interface/node.h>
 
+#include <chain.h>
 #include <chainparams.h>
 #include <init.h>
 #include <interface/handler.h>
@@ -11,9 +12,13 @@
 #include <net.h>
 #include <netaddress.h>
 #include <netbase.h>
+#include <primitives/block.h>
 #include <scheduler.h>
+#include <sync.h>
+#include <txmempool.h>
 #include <ui_interface.h>
 #include <util.h>
+#include <validation.h>
 #include <warnings.h>
 
 #if defined(HAVE_CONFIG_H)
@@ -25,6 +30,7 @@
 #define CHECK_WALLET(x) throw std::logic_error("Wallet function called in non-wallet build.")
 #endif
 
+#include <atomic>
 #include <boost/thread/thread.hpp>
 
 class CWallet;
@@ -69,6 +75,56 @@ class NodeImpl : public Node
     }
     std::string helpMessage(HelpMessageMode mode) override { return HelpMessage(mode); }
     bool getProxy(Network net, proxyType& proxy_info) override { return GetProxy(net, proxy_info); }
+    size_t getNodeCount(CConnman::NumConnections flags) override
+    {
+        return g_connman ? g_connman->GetNodeCount(flags) : 0;
+    }
+    int64_t getTotalBytesRecv() override { return g_connman ? g_connman->GetTotalBytesRecv() : 0; }
+    int64_t getTotalBytesSent() override { return g_connman ? g_connman->GetTotalBytesSent() : 0; }
+    size_t getMempoolSize() override { return ::mempool.size(); }
+    size_t getMempoolDynamicUsage() override { return ::mempool.DynamicMemoryUsage(); }
+    bool getHeaderTip(int& height, int64_t& block_time) override
+    {
+        LOCK(::cs_main);
+        if (::pindexBestHeader) {
+            height = ::pindexBestHeader->nHeight;
+            block_time = ::pindexBestHeader->GetBlockTime();
+            return true;
+        }
+        return false;
+    }
+    int getNumBlocks() override
+    {
+        LOCK(::cs_main);
+        return ::chainActive.Height();
+    }
+    int64_t getLastBlockTime() override
+    {
+        LOCK(::cs_main);
+        if (::chainActive.Tip()) {
+            return ::chainActive.Tip()->GetBlockTime();
+        }
+        return Params().GenesisBlock().GetBlockTime(); // Genesis block's time of current network
+    }
+    double getVerificationProgress() override
+    {
+        const CBlockIndex* tip;
+        {
+            LOCK(::cs_main);
+            tip = ::chainActive.Tip();
+        }
+        return GuessVerificationProgress(::Params().TxData(), tip);
+    }
+    bool isInitialBlockDownload() override { return IsInitialBlockDownload(); }
+    bool getReindex() override { return ::fReindex; }
+    bool getImporting() override { return ::fImporting; }
+    void setNetworkActive(bool active) override
+    {
+        if (g_connman) {
+            g_connman->SetNetworkActive(active);
+        }
+    }
+    bool getNetworkActive() override { return g_connman && g_connman->GetNetworkActive(); }
     std::unique_ptr<Handler> handleInitMessage(InitMessageFn fn) override
     {
         return MakeHandler(::uiInterface.InitMessage.connect(fn));
@@ -89,6 +145,37 @@ class NodeImpl : public Node
     {
         CHECK_WALLET(
             return MakeHandler(::uiInterface.LoadWallet.connect([fn](CWallet* wallet) { fn(MakeWallet(*wallet)); })));
+    }
+    std::unique_ptr<Handler> handleNotifyNumConnectionsChanged(NotifyNumConnectionsChangedFn fn) override
+    {
+        return MakeHandler(::uiInterface.NotifyNumConnectionsChanged.connect(fn));
+    }
+    std::unique_ptr<Handler> handleNotifyNetworkActiveChanged(NotifyNetworkActiveChangedFn fn) override
+    {
+        return MakeHandler(::uiInterface.NotifyNetworkActiveChanged.connect(fn));
+    }
+    std::unique_ptr<Handler> handleNotifyAlertChanged(NotifyAlertChangedFn fn) override
+    {
+        return MakeHandler(::uiInterface.NotifyAlertChanged.connect(fn));
+    }
+    std::unique_ptr<Handler> handleBannedListChanged(BannedListChangedFn fn) override
+    {
+        return MakeHandler(::uiInterface.BannedListChanged.connect(fn));
+    }
+    std::unique_ptr<Handler> handleNotifyBlockTip(NotifyBlockTipFn fn) override
+    {
+        return MakeHandler(::uiInterface.NotifyBlockTip.connect([fn](bool initial_download, const CBlockIndex* block) {
+            fn(initial_download, block->nHeight, block->GetBlockTime(),
+                GuessVerificationProgress(::Params().TxData(), block));
+        }));
+    }
+    std::unique_ptr<Handler> handleNotifyHeaderTip(NotifyHeaderTipFn fn) override
+    {
+        return MakeHandler(
+            ::uiInterface.NotifyHeaderTip.connect([fn](bool initial_download, const CBlockIndex* block) {
+                fn(initial_download, block->nHeight, block->GetBlockTime(),
+                    GuessVerificationProgress(::Params().TxData(), block));
+            }));
     }
 };
 
--- a/src/interface/node.cpp
+++ b/src/interface/node.cpp
@@ -7,6 +7,7 @@
 #include <chainparams.h>
 #include <init.h>
 #include <interface/handler.h>
+#include <interface/wallet.h>
 #include <net.h>
 #include <netaddress.h>
 #include <netbase.h>
@@ -15,7 +16,18 @@
 #include <util.h>
 #include <warnings.h>
 
+#if defined(HAVE_CONFIG_H)
+#include <config/bitcoin-config.h>
+#endif
+#ifdef ENABLE_WALLET
+#define CHECK_WALLET(x) x
+#else
+#define CHECK_WALLET(x) throw std::logic_error("Wallet function called in non-wallet build.")
+#endif
+
 #include <boost/thread/thread.hpp>
+
+class CWallet;
 
 namespace interface {
 namespace {
@@ -68,6 +80,15 @@ class NodeImpl : public Node
     std::unique_ptr<Handler> handleQuestion(QuestionFn fn) override
     {
         return MakeHandler(::uiInterface.ThreadSafeQuestion.connect(fn));
+    }
+    std::unique_ptr<Handler> handleShowProgress(ShowProgressFn fn) override
+    {
+        return MakeHandler(::uiInterface.ShowProgress.connect(fn));
+    }
+    std::unique_ptr<Handler> handleLoadWallet(LoadWalletFn fn) override
+    {
+        CHECK_WALLET(
+            return MakeHandler(::uiInterface.LoadWallet.connect([fn](CWallet* wallet) { fn(MakeWallet(*wallet)); })));
     }
 };
 
--- a/src/interface/node.cpp
+++ b/src/interface/node.cpp
@@ -216,7 +216,9 @@ class NodeImpl : public Node
         return result;
     }
     CFeeRate getDustRelayFee() override { return ::dustRelayFee; }
+    CFeeRate getFallbackFee() override { CHECK_WALLET(return CWallet::fallbackFee); }
     CFeeRate getPayTxFee() override { CHECK_WALLET(return ::payTxFee); }
+    void setPayTxFee(CFeeRate rate) override { CHECK_WALLET(::payTxFee = rate); }
     UniValue executeRpc(const std::string& command, const UniValue& params, const std::string& uri) override
     {
         JSONRPCRequest req;
--- a/src/interface/node.cpp
+++ b/src/interface/node.cpp
@@ -113,7 +113,7 @@ class NodeImpl : public Node
             LOCK(::cs_main);
             tip = ::chainActive.Tip();
         }
-        return GuessVerificationProgress(::Params().TxData(), tip);
+        return GuessVerificationProgress(Params().TxData(), tip);
     }
     bool isInitialBlockDownload() override { return IsInitialBlockDownload(); }
     bool getReindex() override { return ::fReindex; }
@@ -166,7 +166,7 @@ class NodeImpl : public Node
     {
         return MakeHandler(::uiInterface.NotifyBlockTip.connect([fn](bool initial_download, const CBlockIndex* block) {
             fn(initial_download, block->nHeight, block->GetBlockTime(),
-                GuessVerificationProgress(::Params().TxData(), block));
+                GuessVerificationProgress(Params().TxData(), block));
         }));
     }
     std::unique_ptr<Handler> handleNotifyHeaderTip(NotifyHeaderTipFn fn) override
@@ -174,7 +174,7 @@ class NodeImpl : public Node
         return MakeHandler(
             ::uiInterface.NotifyHeaderTip.connect([fn](bool initial_download, const CBlockIndex* block) {
                 fn(initial_download, block->nHeight, block->GetBlockTime(),
-                    GuessVerificationProgress(::Params().TxData(), block));
+                    GuessVerificationProgress(Params().TxData(), block));
             }));
     }
 };
--- a/src/interface/node.cpp
+++ b/src/interface/node.cpp
@@ -55,6 +55,7 @@ class NodeImpl : public Node
             StopMapPort();
         }
     }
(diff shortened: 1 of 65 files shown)


laanwj:
ClientModel and WalletModel were already meant as abstraction layer for accessing the core from the GUI. What is your rationale for adding another layer?

ryanofsky:
ClientModel and WalletModel might have been intended to be an abstraction layer, but they are not functioning like one. There are libbitcoin functions and global variables accessed all over Qt code right now. With this change, all of these calls (there are around 200 of them) are stripped out of Qt code and moved into a one file: [src/ipc/local/interfaces.cpp](https://github.com/ryanofsky/bitcoin/blob/pr/ipc-local/src/ipc/local/interfaces.cpp).

jonasschnelli:
I once did a similar thing,.. but stopped at some point and now I know why.
It's an impressive code change and I kinda like a central point (your interfaces.cpp) where communication between the node, the wallet and the GUI happens.

I also agree with @laanwj that the clientmodel (node) and the walletmodal (wallet) are originally though to be that layer.
Though, there are many violations AFAIK.

What would be the downsides of using the exiting layers (clientmodel / walletmodel) better?

ryanofsky:
> What would be the downsides of using the exiting layers (clientmodel / walletmodel) better?

If you look at the `ClientModel` class, you can see it is doing a lot more work than the `ipc::local::Node` class is. Similarly with `WalletModel` and `ipc::local::Wallet.` The ipc classes are just simple shims around low-level node and wallet functionality, while Qt objects implement higher level logic specific to our current GUI. I still think `ClientModel` and `WalletModel` classes are still useful after this change. They will just have 1 job instead of 2. Instead of serving as both an abstraction layers and as MVC model classes, they will serve only as MVC model classes.

Also, and in more concrete terms, the reason these interfaces live outside the src/qt directory is that with #10102, they need to be accessed not only by `bitcoin-qt` but also by `bitcoind` (specifically inside the `StartServer` function in [src/ipc/server.cpp](https://github.com/ryanofsky/bitcoin/blob/pr/ipc/src/ipc/server.cpp) which is called here: https://github.com/ryanofsky/bitcoin/commit/ab0afba3a44255b3eec80f4eebe45a851ae23927#diff-6e30027c2045842fe842430d98d099fb

jonasschnelli:
The general IPC interface makes sense to me. The main problem I see for any type of low latency IPC/RPC is the missing asynchronity.
Take `getWalletTxDetails`. This IPC call may take 2-3 seconds depending on the communication protocol and database you are using. Ideally the GUI is design to handle it asynchronous (like an RPC call) otherwise this will lead to GUI thread freezes. Not sure if this would be solvable as a generic part in the IPC layer of if the wallet/GUI logic must handle it.

ryanofsky:
> The main problem I see for any type of low latency IPC/RPC is the missing asynchronity.

Not sure if you saw the comments about this in the other pr starting here: https://github.com/bitcoin/bitcoin/pull/10102#issuecomment-294094610

These changes are orthogonal to event processing / blocking issues in the UI. If UI blocked before, it will still block after these changes, if UI didn't block before, it won't start blocking now because of these changes. If remote calls are too slow because of socket/serialization overhead, we can process UI events in the background while they are being made. There are many ways to accomplish this, with one possible way described in that comment above.

ryanofsky:
Also would point out that Node and Wallet interfaces in [ipc/interfaces.h](https://github.com/ryanofsky/bitcoin/blob/pr/ipc-local/src/ipc/interfaces.h) were mainly designed with goal of changing existing Qt code as little as possible. They aren't in any way set in stone, and I would expect them to evolve over time. Probably some calls will get consolidated, others will get broken up, calls that currently return big chunks of data will be made streaming, etc.

jonasschnelli:
Thinking again and discussing this with @sipa / @laanwj, I think we should use the existing client-/walletmodal as node/wallet abstraction (including a possible IPC abstraction).

What's missing in the first place are better asynchronous messaging between the GUI and the wallet/node.

IMO using a thread with queue for general node/wallet communication (and eventual additional threads for procedures that usually take longer) seems after a low hanging fruit with direct benefits.

Using QT slots/signals for all(most?) communication would be required anyways would be beneficial even without IPC and therefor done first.

ryanofsky:
> What's missing in the first place are better asynchronous messaging between the GUI and the wallet/node.

Again I think this is (and should be an independent issue, but if you want to flesh out some more concrete suggestions and I would be happy to hear them.

> IMO using a thread with queue for general node/wallet communication (and eventual additional threads for procedures that usually take longer) seems after a low hanging fruit with direct benefits.

This is exactly what the change I was suggesting in https://github.com/bitcoin/bitcoin/pull/10102#issuecomment-294094610 does.

laanwj:
> Using QT slots/signals for all(most?) communication would be required anyways and would be beneficial even without IPC and therefor should be done first.

This was my point too. Making the GUI asynchronous would avoid ever hard-freezing the GUI. Modern operating systems assume that an application has crashed if its GUI thread is unresponsive. This is a priority for improving user experience. For example: Currently, if e.g. a transaction is sent while the cs_main lock is held the entire thing hangs for a moment. Ideally it would display a modal dialog with a status, or progress animation instead. There are similar issues at startup.

Sure, this is only partially related to IPC work: When the GUI already would communicate with Qt signals and slots with the core backend (similar to how RPCConsole and RPCThread communicate, for example), it could be mostly oblivious whether this backend exists in-process or communicates over a pipe.

Although it's laudable that you're working on this, it looks to me that what you are doing currently is simply replicating what we do now but replacing direct core calls with IPC calls. The drawback is that it calcifies some things that shouldn't have been designed that way in the first place (e.g. into multiple abstraction layers), making it harder to improve later.

ryanofsky:
> The drawback is that it calcifies some things

Could you be more concrete about this? I don't see how it is true. Direct calls before are still direct calls now. If we want to follow the RPCConsole / RPCExecutor model in other parts of Qt code, I don't see how any of the changes I've made for IPC make this more difficult.

ryanofsky:
I had a look at discussion in IRC (https://botbot.me/freenode/bitcoin-core-dev/msg/84348426/)

With respect, what I think you guys are missing on the `WalletModel`/`ClientModel` topic is that the `ipc::local::WalletImpl` and `ipc::local::NodeImpl` classes in [ipc/local/interfaces.cpp](https://github.com/ryanofsky/bitcoin/blob/pr/ipc-local/src/ipc/local/interfaces.cpp) are only **temporarily** being created and invoked within the `bitcoin-qt` process. In the next PR they are created and run in the `bitcoind` process instead of `bitcoin-qt`. That's the reason these classes do not reside in the `src/qt` directory and one reason why they don't really substitute for the `WalletModel`/`ClientModel` classes. See my previous comment for details and a code pointer: https://github.com/bitcoin/bitcoin/pull/10244#issuecomment-295887270. 

However, I do see that it is kind of silly to have cases where Qt code calls a `WalletModel`/`ClientModel` method that merely forwards to a `WalletImpl`/`NodeImpl` method. I can easily clear this up by inlining these `WalletModel`/`ClientModel` methods, which would make the classes more lean.

Also, if this PR will be too difficult to review because of its size (https://botbot.me/freenode/bitcoin-core-dev/msg/84348447/), I can easily decompose it into smaller PRs that could be gradually merged. It is already broken up into separate commits, and many of the individual commits could be further broken up (right now they try to group together related changes).

ryanofsky:
@laanwj and @jonasschnelli can you let me know if you still see issues with this approach?

On the Qt model class issue, I pulled out a bunch of model methods that were just wrapping IPC calls so it should be clearer what actual work [walletmodel.cpp](https://github.com/ryanofsky/bitcoin/blob/pr/ipc-local/src/qt/walletmodel.cpp) and [clientmodel.cpp](https://github.com/ryanofsky/bitcoin/blob/pr/ipc-local/src/qt/clientmodel.cpp) are doing. I also added a little blurb to the [IPC README](https://github.com/ryanofsky/bitcoin/tree/pr/ipc-local/src/ipc#does-having-an-ipc-layer-make-qt-model-classes-redundant) describing what the distinction between Qt model classes and the IPC interface classes is supposed to be.

On the asynchronous GUI issue, I created #10504 for more discussion, but think that issue is mostly unrelated to the changes in this PR, except as far as some changes here might potentially make it easier to identify blocking calls and make them asynchronous.

laanwj:
Concept ACK on the approach, moving things to interfaces instead of global calls is usually good as it makes it easier to see what the interface to the core is.

jonasschnelli:
Concept ACK.
I think we could try to rebase and get this in once 0.15 has been split off

ryanofsky:
@promag re: https://botbot.me/freenode/bitcoin-core-dev/msg/92986206/

> \<promag> to change the current wallet from widgets to qtquick would take several PR's (which would sit there for a long time)
> \<promag> dunno if the IPC PR from ryanofsky is going forward
> \<promag> both require lots of changes to the wallet ui code
> \<promag> that's why I think adding IPC to glue a new wallet ui is probably easier
> \<promag> we could have both bitcoin-qt and bitcoin-quick and later drop support for the first
> \<promag> anyway, moving to qtquick is something IMHO we should plan

I don't think you need any IPC stuff for this. This PR by itself, which is a refactoring, provides interfaces for GUI code to access node&wallet state while being less tied to their internals. The Node and Wallet interfaces here would probably have to be extended in various ways to do things you'd want in a more modern UI, but they do provide a base of functionality that someone could use to implement one.

ryanofsky:
Another IRC mention of this: https://botbot.me/freenode/bitcoin-core-dev/msg/93856775/

sipa:
After reading the discussion again, Concept ACK.

While WalletModel and ClientModel were designed as an abstraction for the "core", they're incomplete in that regard (many direct calls bypass the abstraction), and a bit too high level (having significant logic of their own).

I guess we can see this PR as introducing a perfect abstraction, with no logic of its own.

@ryanofsky What do you think the future responsibilities of WalletModel and ClientModel should be? Should they be turned into logic on the core signed, abstracted by the new interfaces, or just inlined into the calls sites inside the GUI?

ryanofsky:
> @ryanofsky What do you think the future responsibilities of WalletModel and ClientModel should be? Should they be turned into logic on the core signed, abstracted by the new interfaces, or just inlined into the calls sites inside the GUI?

I think if model code is just calling interface functions it should probably be inlined. But if model code is doing something useful to support the gui, like caching, it should be kept. If there is model code that makes too many assumptions about node/wallet internals, or implements functionality that could be useful for RPCs, I think it should be moved into the core and wrapped by new interfaces.

Example of code that could be inlined:

https://github.com/bitcoin/bitcoin/blob/9cdbd195ba85b1af12fe8c55db3880df73e107ff/src/qt/walletmodel.cpp#L354-L366

Example of code that should be kept:

https://github.com/bitcoin/bitcoin/blob/9cdbd195ba85b1af12fe8c55db3880df73e107ff/src/qt/clientmodel.cpp#L73-L86

An example of code that should be probably be moved into the core is a lot of the startup and config code in [qt/bitcoin.cpp](https://github.com/bitcoin/bitcoin/blob/9cdbd195ba85b1af12fe8c55db3880df73e107ff/src/qt/bitcoin.cpp).

src/interface/wallet.h
```diff
@@ -251,6 +288,38 @@ struct WalletBalances
...
 };
 
+// Wallet transaction information.
+struct WalletTx
+{
+    CTransactionRef tx;
+    std::vector<isminetype> txin_is_mine;
+    std::vector<isminetype> txout_is_mine;
+    std::vector<CTxDestination> txout_address;
+    std::vector<isminetype> txout_address_is_mine;
```
jamesob: This may not be practical to change for legacy reasons, but I find the naming here confusing; it makes me think bool and not vector. Something like `my_txout_addresses` (and likewise for two others above) seems clearer.

jamesob (commented):
Concept ACK

This represents a good deal of progress towards decomposing the code into its three major subsystems, regardless of how or if we approach actual process separation. Consolidating and making explicit the interfaces these subsystems expose and depend on will make it easier to reason about the best way to move forward. Evaluating whether to use this or that IPC mechanism or how to approach GUI asynchronicity will be framed nicely by the low-risk, low-overhead boundaries that @ryanofsky has defined in this changeset. On top of that, there are numerous readability improvements, e.g. consolidating lock acquisition, extracting functions, and cleaning up naming. I think this is great work.

I'll be testing this thoroughly in the next few days.

src/qt/bitcoin.cpp
```diff
@@ -27,16 +27,14 @@
 #endif
 
 #include <init.h>
```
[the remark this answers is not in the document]
jamesob: Oops, nevermind; the include should stay in some files: 
```
 $ for i in Interrupt Shutdown ShutdownRequested InitParameter AppInit HelpMessageMode HelpMessage LicenseInfo; do git grep "${i}*()" src/qt | cut -d':' -f 1 | sort -u; done

src/qt/bitcoin.cpp
src/qt/bitcoingui.cpp
src/qt/bitcoingui.h
src/qt/splashscreen.cpp
src/qt/winshutdownmonitor.cpp
src/qt/utilitydialog.cpp
```

ryanofsky (commented):
@jamesob, thanks for the review, really appreciate you looking at this so thoroughly!

---

Added 11 commits 9cdbd195ba85b1af12fe8c55db3880df73e107ff -> 7fcc6e43025fb2692acdd722dab2ef9598711377 ([pr/ipc-local.53](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.53) -> [pr/ipc-local.54](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.54), [compare](https://github.com/ryanofsky/bitcoin/compare/pr/ipc-local.53...pr/ipc-local.54))
Squashed 7fcc6e43025fb2692acdd722dab2ef9598711377 -> de2a1dca1a1e158b9352ff466cbd790ec9ef18ca ([pr/ipc-local.54](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.54) -> [pr/ipc-local.55](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.55))

src/interface/node.cpp
```diff
@@ -182,7 +186,29 @@ class NodeImpl : public Node
...
     unsigned int getTxConfirmTarget() override { CHECK_WALLET(return ::nTxConfirmTarget); }
+    CAmount getRequiredFee(unsigned int tx_bytes) override { CHECK_WALLET(return GetRequiredFee(tx_bytes)); }
+    CAmount getMinimumFee(unsigned int tx_bytes,
+        const CCoinControl& coin_control,
+        int* returned_target,
+        FeeReason* reason) override
+    {
+        FeeCalculation fee_calc;
+        CHECK_WALLET(return GetMinimumFee(tx_bytes, coin_control, ::mempool, ::feeEstimator, &fee_calc));
+        if (returned_target) *returned_target = fee_calc.returnedTarget;
```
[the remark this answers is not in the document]
jamesob: Ah sorry, what I meant was that, because of the way `CHECK_WALLET` works, it doesn't look like we'll ever make it to this line or the one below it.

src/interface/wallet.h
jamesob: Oh, you're totally right; I misunderstood the content of these vectors. Thanks for the clarification.

ryanofsky (commented):
Added 1 commits de2a1dca1a1e158b9352ff466cbd790ec9ef18ca -> 8aced6d22ddf6cf2af6ed02ad8344b583c238b9f ([pr/ipc-local.55](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.55) -> [pr/ipc-local.56](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.56), [compare](https://github.com/ryanofsky/bitcoin/compare/pr/ipc-local.55...pr/ipc-local.56))
Squashed 8aced6d22ddf6cf2af6ed02ad8344b583c238b9f -> 76ed998e79e21d1b2a524ec4850c91e03030f5fe ([pr/ipc-local.56](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.56) -> [pr/ipc-local.57](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.57))
Rebased 76ed998e79e21d1b2a524ec4850c91e03030f5fe -> aa57e3dafaee4ec9383483997768d7db5ce8880a ([pr/ipc-local.57](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.57) -> [pr/ipc-local.58](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.58)) due to conflict with #11041.
Rebased aa57e3dafaee4ec9383483997768d7db5ce8880a -> 84e80629d10cb48009b84e131ee61e0a8b9ad171 ([pr/ipc-local.58](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.58) -> [pr/ipc-local.59](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.59)) due to conflict with #9680.

src/interface/node.cpp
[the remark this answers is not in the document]
ryanofsky: https://github.com/bitcoin/bitcoin/pull/10244#discussion_r173528339

Previous comment looks a little odd in gmail. Github apparently tries to render `:man_facepalming:` as `['FACE PALM', 'ZERO WIDTH JOINER', 'MALE SIGN', 'VARIATION SELECTOR-16']` in unicode.

src/interface/node.cpp
```diff
@@ -192,6 +194,18 @@ class NodeImpl : public Node
     std::vector<std::string> listRpcCommands() override { return ::tableRPC.listCommands(); }
     void rpcSetTimerInterfaceIfUnset(RPCTimerInterface* iface) override { RPCSetTimerInterfaceIfUnset(iface); }
     void rpcUnsetTimerInterface(RPCTimerInterface* iface) override { RPCUnsetTimerInterface(iface); }
+    std::vector<std::unique_ptr<Wallet>> getWallets() override
+    {
+#ifdef ENABLE_WALLET
+        std::vector<std::unique_ptr<Wallet>> wallets;
+        for (CWalletRef wallet : ::vpwallets) {
+            wallets.emplace_back(MakeWallet(*wallet));
```
Sjors: Has anyone tested a rebased #11383 or #12610 with this?

src/interface/wallet.cpp
```diff
@@ -15,15 +29,211 @@
...
+};
+
 class WalletImpl : public Wallet
 {
 public:
     WalletImpl(CWallet& wallet) : m_wallet(wallet) {}
 
+    bool encryptWallet(const SecureString& wallet_passphrase) override
+    {
+        return m_wallet.EncryptWallet(wallet_passphrase);
```
Sjors: I tested encrypting a wallet and using it to send.

src/interface/wallet.cpp
```diff
@@ -15,15 +29,211 @@
...
+    {
+        LOCK2(cs_main, m_wallet.cs_wallet);
+        auto pending = MakeUnique<PendingWalletTxImpl>(m_wallet);
+        if (!m_wallet.CreateTransaction(recipients, pending->m_tx, pending->m_key, fee, change_pos,
+                fail_reason, coin_control, sign)) {
+            return {};
+        }
+        return std::move(pending);
+    }
+    bool transactionCanBeAbandoned(const uint256& txid) override { return m_wallet.TransactionCanBeAbandoned(txid); }
```
Sjors: I tested bumping the fee.

src/interface/wallet.cpp
```diff
@@ -15,15 +29,211 @@
...
+    bool abandonTransaction(const uint256& txid) override
+    {
+        LOCK2(cs_main, m_wallet.cs_wallet);
+        return m_wallet.AbandonTransaction(txid);
```
Sjors: I tested abandoning a transaction (after bumping it)

src/interface/wallet.cpp
```diff
@@ -15,15 +29,211 @@
...
+    bool changeWalletPassphrase(const SecureString& old_wallet_passphrase,
+        const SecureString& new_wallet_passphrase) override
+    {
+        return m_wallet.ChangeWalletPassphrase(old_wallet_passphrase, new_wallet_passphrase);
+    }
+    bool backupWallet(const std::string& filename) override { return m_wallet.BackupWallet(filename); }
+    bool getPubKey(const CKeyID& address, CPubKey& pub_key) override { return m_wallet.GetPubKey(address, pub_key); }
+    bool getPrivKey(const CKeyID& address, CKey& key) override { return m_wallet.GetKey(address, key); }
+    bool isSpendable(const CTxDestination& dest) override { return IsMine(m_wallet, dest) & ISMINE_SPENDABLE; }
+    bool haveWatchOnly() override { return m_wallet.HaveWatchOnly(); };
```
Sjors: I didn't check any of the watch-only and isMine stuff.

src/qt/sendcoinsdialog.cpp
```diff
@@ -193,7 +195,7 @@ void SendCoinsDialog::setModel(WalletModel *_model)
             settings.remove("nSmartFeeSliderPosition");
         }
         if (settings.value("nConfTarget").toInt() == 0)
-            ui->confTargetSelector->setCurrentIndex(getIndexForConfTarget(model->getDefaultConfirmTarget()));
+            ui->confTargetSelector->setCurrentIndex(getIndexForConfTarget(model->node().getTxConfirmTarget()));
```
Sjors: Are you sure this is correct? Not sure how `::nTxConfirmTarget` relates to `getDefaultConfirmTarget`

src/qt/walletmodel.cpp
```diff
@@ -502,21 +423,21 @@ static void NotifyWatchonlyChanged(WalletModel *walletmodel, bool fHaveWatchonly
 void WalletModel::subscribeToCoreSignals()
 {
     // Connect signals to wallet
-    wallet->NotifyStatusChanged.connect(boost::bind(&NotifyKeyStoreStatusChanged, this, _1));
-    wallet->NotifyAddressBookChanged.connect(boost::bind(NotifyAddressBookChanged, this, _1, _2, _3, _4, _5, _6));
```
Sjors: Why is `_6` dropped here?

src/interface/wallet.cpp
```diff
@@ -95,6 +99,10 @@ class WalletImpl : public Wallet
     {
         return m_wallet.SetAddressBook(dest, name, purpose);
     }
+    bool delAddressBook(const CTxDestination& dest) override
+    {
+        return m_wallet.DelAddressBook(dest);
```
Sjors: Tested using and deleting entries from address book.

src/qt/bitcoin.cpp
```diff
@@ -630,7 +630,7 @@ int main(int argc, char *argv[])
     }
 #ifdef ENABLE_WALLET
     // Parse URIs on command line -- this can affect Params()
-    PaymentServer::ipcParseCommandLine(argc, argv);
+    PaymentServer::ipcParseCommandLine(*node, argc, argv);
```
Sjors: Lightly tested BIP-21 URI.

Sjors (approved):
tACK 84e80629d to the degree that I understand the code.

This change makes code easier to understand for me. The interfaces could use more documentation, especially the various initialization steps, but that can wait. Thanks for splitting into incremental commits; that did make review less daunting.

Since QT doesn't have amazing test coverage, I left a bunch of comments on stuff I manually tested. Those are FYI or asking for clarification, unless clearly marked otherwise.

I encountered a situation once where I was unable to change tabs (macOS). I had to switch to master, compile and then switch back to your branch and compile again to make it work again. Could be unrelated, but I've never seen that before.

schermafbeelding 2018-03-14 om 16 10 27

MeshCollider:
Concept ACK, on my to-do list to review

ryanofsky (commented):
Thank you for the review and testing!

ryanofsky:
Going to hold off on testing this one until it's rebased.

> Rebased 84e80629d10cb48009b84e131ee61e0a8b9ad171 -> b39555ae6eeb205706922766463a80cfe09153be ([pr/ipc-local.59](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.59) -> [pr/ipc-local.60](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.60))

jamesob (approved):
Tested ACK https://github.com/bitcoin/bitcoin/pull/10244/commits/19a268c1f500e0c8793da6714a9d093d57bfe8f7

Left a few inline comments indicating things I manually tested, but quickly gave that up in favor of documenting a manual test plan:

- [x] *Send*: tested sending a transaction to self with automatic coin selection
  - [x] Select custom fee, minimum fee display (`getMinimumFee`)
  - [x] Use recommended fee for 2 block conf. time (`estimateSmartFee`)
- [x] *Transactions -> right click on row* -> Increase transaction fee: after sending with recommended fee, test bumping the fee 
- [x] *Send*: tested sending a transaction to self with manual coin selection, custom
  fee address, subtract fee from amount, RBF enabled
- [x] *Send -> Choose previously used address -> New*: append to the address book
  - Found (and subsequently abandoned) https://github.com/bitcoin/bitcoin/issues/12796
- [x] *Send -> Choose previously used address*: delete from the address book
- [x] *Settings -> Encrypt Wallet...*: encrypt existing wallet
- [x] *Send*: send from encrypted wallet
- [x] *Settings -> Change Passphrase...*: change wallet passphrase
- [x] *Send -> Coin Control Features -> Inputs*: lock/unlock UTXO from the coin selection interface
- [x] *Transactions -> right click on row -> Edit label*: edit transaction label after send (**failed**)
  - Fails on master as well, will investigate and maybe file a bug later

I've spent the better part of the day in QT (Ubuntu 17.10) with this changeset and I wasn't able to find any issues. Nice work!

jnewbery (commented):
Tested ACK 025e11b7e800de8265c33fa8efdc035282524dcd

There's too much code here for me to closely review all the changes, but I've had a look through all the commits and kicked the tires. We'd preferably get this PR merged early in a release cycle to get plenty of soaking.

This is definitely a big step in the right direction in clearly splitting qt from node code.

ryanofsky (commented):
Rebased 19a268c1f500e0c8793da6714a9d093d57bfe8f7 -> 6ddaaaabfec1df43857089cfd232548d0952cb21 ([pr/ipc-local.61](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.61) -> [pr/ipc-local.62](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.62))
Added 4 commits 6ddaaaabfec1df43857089cfd232548d0952cb21 -> a415fa2845f843db2ceb6512537cbb2e63b5b474 ([pr/ipc-local.62](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.62) -> [pr/ipc-local.63](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.63), [compare](https://github.com/ryanofsky/bitcoin/compare/pr/ipc-local.62...pr/ipc-local.63)) implementing jnewbery review suggestions
Squashed a415fa2845f843db2ceb6512537cbb2e63b5b474 -> d792df74338771174087480a70a7ad48e4e79ab7 ([pr/ipc-local.63](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.63) -> [pr/ipc-local.64](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.64))


---

Added 2 commits d792df74338771174087480a70a7ad48e4e79ab7 -> 9948724c84ba6e31d106c6dae2c901fdc4e85a9c ([pr/ipc-local.64](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.64) -> [pr/ipc-local.65](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.65), [compare](https://github.com/ryanofsky/bitcoin/compare/pr/ipc-local.64...pr/ipc-local.65)) to finish [WalletBalances suggestion](https://github.com/bitcoin/bitcoin/pull/10244#discussion_r177504284) from jnewbery and address [blocking GUI concerns](https://botbot.me/freenode/bitcoin-core-dev/msg/98388760/) from laanwj 
Squashed 9948724c84ba6e31d106c6dae2c901fdc4e85a9c -> 4bf295201d0f1059d9670329cb412fd074910319 ([pr/ipc-local.65](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.65) -> [pr/ipc-local.66](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.66))

jnewbery:
Needs rebase to fix unrelated travis failure

ryanofsky:
> Needs rebase to fix unrelated travis failure

Rebased 4bf295201d0f1059d9670329cb412fd074910319 -> e73a9f589f6ddeb0e350d887cd1f98fc55aac632 ([pr/ipc-local.66](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.66) -> [pr/ipc-local.67](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.67)). Looks like failure in mempool_persist test: https://travis-ci.org/bitcoin/bitcoin/jobs/361243719#L2810

jnewbery:
> Looks like failure in mempool_persist test

Ah. The new linter lint-include-guards was also failing, which is what required a rebase.

I think mempool_persist was broken by https://github.com/bitcoin/bitcoin/commit/cb1e319fe9e198c9c5cf5236fe9af5a3d748b9e8. Will open an issue.

ryanofsky:
Rebased e73a9f589f6ddeb0e350d887cd1f98fc55aac632 -> 4678ff464f03c09feb0e237343f2af329293343a ([pr/ipc-local.67](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.67) -> [pr/ipc-local.68](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.68)) due to conflict with #12846.

ryanofsky:
@laanwj or @jonasschnelli I wonder if it may be a good time to merge this given recent testing and discussion (https://botbot.me/freenode/bitcoin-core-dev/msg/98388625/). Review status:

Tested ACKs

- jnewbery https://github.com/bitcoin/bitcoin/pull/10244#pullrequestreview-107372769, partial code review
- jamesob https://github.com/bitcoin/bitcoin/pull/10244#pullrequestreview-107016963
- Sjors https://github.com/bitcoin/bitcoin/pull/10244#pullrequestreview-103872908

Concept ACKs

- MeshCollider https://github.com/bitcoin/bitcoin/pull/10244#issuecomment-373184202
- sipa https://github.com/bitcoin/bitcoin/pull/10244#issuecomment-371883024
- jonasschnelli https://github.com/bitcoin/bitcoin/pull/10244#issuecomment-308738280
- laanwj https://github.com/bitcoin/bitcoin/pull/10244#issuecomment-308731640

All the changes in this PR are limited to the GUI code and only affect bitcoin-qt (not bitcoind).

jamesob:
Tested ACK https://github.com/bitcoin/bitcoin/pull/10244/commits/4678ff464f03c09feb0e237343f2af329293343a

Ran through the same test plan I posted above earlier.

Sjors:
Some linter unhappiness: schermafbeelding 2018-04-04 om 15 11 55

src/interface/wallet.h
```diff
@@ -0,0 +1,348 @@
+#ifndef BITCOIN_INTERFACE_INTERFACES_H
+#define BITCOIN_INTERFACE_INTERFACES_H
```
MarcoFalke: "wrong" include guard. Also might want to add the usual (c) header?

ryanofsky:
Updated 4678ff464f03c09feb0e237343f2af329293343a -> 996013769711bd507cdcd6dde88cbd59fcd4fbad ([pr/ipc-local.68](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.68) -> [pr/ipc-local.69](https://github.com/ryanofsky/bitcoin/commits/pr/ipc-local.69)) from https://github.com/jnewbery/bitcoin/commits/pr10244_fix_guards_copyright (thanks @jnewbery!). Only change is fixing the broken include guard and adding copyrights.

ken2812221:
This PR broke the gitian build on windows
https://bitcoin.jonasschnelli.ch/builds/559/build_win.log

ryanofsky:
> This PR broke the gitian build on windows

Error is:

```c++
qt/guiutil.cpp:235:24: error: ‘Node’ in namespace ‘::’ does not name a type
 bool isDust(interface::Node& node, const QString& address, const CAmount& amount)
                        ^
qt/guiutil.cpp: In function ‘bool GUIUtil::isDust(int&, const QString&, const CAmount&)’:
qt/guiutil.cpp:240:31: error: request for member ‘getDustRelayFee’ in ‘node’, which is of non-class type ‘int’
     return IsDust(txOut, node.getDustRelayFee());
```

It seems like this is happening because microsoft has some headers that do `#define interface struct`: https://stackoverflow.com/questions/25234203/what-is-the-interface-keyword-in-msvc

I think I'll open a followup PR to rename `interface` directory & namespace to `interfaces` to avoid this problem.

src/qt/walletmodel.cpp
```diff
@@ -117,55 +77,34 @@ void WalletModel::updateStatus()
...
+    // holding the locks for a longer time - for example, during a wallet
+    // rescan.
+    interface::WalletBalances new_balances;
+    int numBlocks = -1;
+    if (!m_wallet->tryGetBalances(new_balances, numBlocks)) {
         return;
+    }
 
-    if(fForceCheckBalanceChanged || chainActive.Height() != cachedNumBlocks)
+    if(fForceCheckBalanceChanged || m_node.getNumBlocks() != cachedNumBlocks)
```
ryanofsky: In commit "Remove most direct bitcoin calls from qt/walletmodel.cpp" (a0704a8996bb950ae3c4d5b5a30e9dfe34cde1d3)

Bug here: `m_node.getNumBlocks()` should have been just `numBlocks`, see #18123

dodifirmansah (commented):
10000000

PR #10245, opened 2017-04-20
Minor fix in build documentation for FreeBSD 11

shigeya:
- Don't need libevent2 -- and it's not availble in pkg also
- Need to use GNU make

### DIFF
--- a/doc/build-unix.md
+++ b/doc/build-unix.md
@@ -321,8 +321,10 @@ Clang is installed by default as `cc` compiler, this makes it easier to get
 started than on [OpenBSD](build-openbsd.md). Installing dependencies:
 
     pkg install autoconf automake libtool pkgconf
-    pkg install boost-libs openssl libevent2
+    pkg install boost-libs openssl libevent
+    pkg install gmake
 
+You need to use GNU make (`gmake`) instead of `make`.
 (`libressl` instead of `openssl` will also work)
 
 For the wallet (optional):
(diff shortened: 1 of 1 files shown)


sipa:
Are you sure nothing is needed to install libevent?

shigeya:
Yes, I could successfully build without error on FreeBSD 11.0-RELEASE-p9. I briefly tested the code with Testnet3 and looks working (although, I haven't run until getting the latest block). 

I decided to shut down the server for now, due to poor CPU resource. If you want more confirmation, I can provide.

sipa:
@shigeya I don't doubt that you could build successfully, but wasn't that because somehow libevent was already installed?

shigeya:
@sipa I was partially wrong. It was named `libevent` and not `libevent2` in `pkg`.  It was installed part of other package already. That's why. I have pushed amended fix. thanks.

doc/build-unix.md
```diff
@@ -338,7 +340,7 @@ Then build using:
 
     ./autogen.sh
     ./configure --with-incompatible-bdb BDB_CFLAGS="-I/usr/local/include/db5" BDB_LIBS="-L/usr/local/lib -ldb_cxx-5"
-    make
+    gmake
```
laanwj: Oh no, how could I have made this heretic mistake.

shigeya: I know, I know... 😆  It's easy to guess due to the error output of `make`, but... Anyway, thanks for docs. It was short and easy.

doc/build-unix.md
```diff
@@ -321,8 +321,10 @@ Clang is installed by default as `cc` compiler, this makes it easier to get
 started than on [OpenBSD](build-openbsd.md). Installing dependencies:
 
     pkg install autoconf automake libtool pkgconf
-    pkg install boost-libs openssl libevent2
+    pkg install boost-libs openssl libevent
```
laanwj: Interesting. I upgraded from an older FreeBSD so my package was still called `libevent2`. But you're right that it has been renamed.

laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/10245/commits/0611bc3439c65e100db98914d18606f623048ba8

PR #10251, opened 2017-04-21
Add balances cache / GUI: use a signal instead of a poll thread

jonasschnelli:
There are two major parts in this PR.

## A) wallet: introduce a per-balance-type cache
Additionally to the per `CWalletTx` debit, etc. caches, this adds an atomic<CAmount> cache for each balance type (available, immature, watchonly, etc.).
If the balance has been cached, no lock will be acquired when calling `Get*Balance()`.
As always with caches, the problematic parts is where to invalidate it (that is why there are some calls to `MarkBalancesDirty()`).

## B) Signal for balance changes
This PR exposes a new wallet signal that will signal the GUI when a balance change had happened (instead of the 250ms polling timer).

### DIFF
--- a/src/qt/walletmodel.cpp
+++ b/src/qt/walletmodel.cpp
@@ -32,7 +32,6 @@
 #include <QDebug>
 #include <QMessageBox>
 #include <QSet>
-#include <QTimer>
 
 #include <boost/foreach.hpp>
 
@@ -45,16 +44,10 @@ WalletModel::WalletModel(const PlatformStyle *platformStyle, CWallet *_wallet, O
     cachedNumBlocks(0)
 {
     fHaveWatchOnly = wallet->HaveWatchOnly();
-    fForceCheckBalanceChanged = false;
 
     addressTableModel = new AddressTableModel(wallet, this);
     transactionTableModel = new TransactionTableModel(platformStyle, wallet, this);
     recentRequestsTableModel = new RecentRequestsTableModel(wallet, this);
-
-    // This timer will be fired repeatedly to update the balance
-    pollTimer = new QTimer(this);
-    connect(pollTimer, SIGNAL(timeout()), this, SLOT(pollBalanceChanged()));
-    pollTimer->start(MODEL_UPDATE_DELAY);
 
     subscribeToCoreSignals();
 }
@@ -112,31 +105,6 @@ void WalletModel::updateStatus()
         Q_EMIT encryptionStatusChanged(newEncryptionStatus);
 }
 
-void WalletModel::pollBalanceChanged()
-{
-    // Get required locks upfront. This avoids the GUI from getting stuck on
-    // periodical polls if the core is holding the locks for a longer time -
-    // for example, during a wallet rescan.
-    TRY_LOCK(cs_main, lockMain);
-    if(!lockMain)
-        return;
-    TRY_LOCK(wallet->cs_wallet, lockWallet);
-    if(!lockWallet)
-        return;
-
-    if(fForceCheckBalanceChanged || chainActive.Height() != cachedNumBlocks)
-    {
-        fForceCheckBalanceChanged = false;
-
-        // Balance and number of transactions might have changed
-        cachedNumBlocks = chainActive.Height();
-
-        checkBalanceChanged();
-        if(transactionTableModel)
-            transactionTableModel->updateConfirmations();
-    }
-}
-
 void WalletModel::checkBalanceChanged()
 {
     CAmount newBalance = getBalance();
@@ -164,12 +132,6 @@ void WalletModel::checkBalanceChanged()
         Q_EMIT balanceChanged(newBalance, newUnconfirmedBalance, newImmatureBalance,
                             newWatchOnlyBalance, newWatchUnconfBalance, newWatchImmatureBalance);
     }
-}
-
-void WalletModel::updateTransaction()
-{
-    // Balance and number of transactions might have changed
-    fForceCheckBalanceChanged = true;
 }
 
 void WalletModel::updateAddressBook(const QString &address, const QString &label,
@@ -367,7 +329,7 @@ WalletModel::SendCoinsReturn WalletModel::sendCoins(WalletModelTransaction &tran
         }
         Q_EMIT coinsSent(wallet, rcp, transaction_array);
     }
-    checkBalanceChanged(); // update balance immediately, otherwise there could be a short noticeable delay until pollBalanceChanged hits
+    checkBalanceChanged();
 
     return SendCoinsReturn(OK);
 }
@@ -476,14 +438,6 @@ static void NotifyAddressBookChanged(WalletModel *walletmodel, CWallet *wallet,
                               Q_ARG(int, status));
 }
 
-static void NotifyTransactionChanged(WalletModel *walletmodel, CWallet *wallet, const uint256 &hash, ChangeType status)
-{
-    Q_UNUSED(wallet);
-    Q_UNUSED(hash);
-    Q_UNUSED(status);
-    QMetaObject::invokeMethod(walletmodel, "updateTransaction", Qt::QueuedConnection);
-}
-
 static void ShowProgress(WalletModel *walletmodel, const std::string &title, int nProgress)
 {
     // emits signal "showProgress"
@@ -498,14 +452,19 @@ static void NotifyWatchonlyChanged(WalletModel *walletmodel, bool fHaveWatchonly
                               Q_ARG(bool, fHaveWatchonly));
 }
 
+static void NotifyBalanceChanged(WalletModel *walletmodel)
+{
+    QMetaObject::invokeMethod(walletmodel, "checkBalanceChanged", Qt::QueuedConnection);
+}
+
 void WalletModel::subscribeToCoreSignals()
 {
     // Connect signals to wallet
     wallet->NotifyStatusChanged.connect(boost::bind(&NotifyKeyStoreStatusChanged, this, _1));
     wallet->NotifyAddressBookChanged.connect(boost::bind(NotifyAddressBookChanged, this, _1, _2, _3, _4, _5, _6));
-    wallet->NotifyTransactionChanged.connect(boost::bind(NotifyTransactionChanged, this, _1, _2, _3));
     wallet->ShowProgress.connect(boost::bind(ShowProgress, this, _1, _2));
     wallet->NotifyWatchonlyChanged.connect(boost::bind(NotifyWatchonlyChanged, this, _1));
+    wallet->BalancesChanged.connect(boost::bind(NotifyBalanceChanged, this));
 }
 
 void WalletModel::unsubscribeFromCoreSignals()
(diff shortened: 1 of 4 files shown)


src/wallet/wallet.cpp
```diff
@@ -1902,95 +1909,123 @@ void CWallet::ResendWalletTransactions(int64_t nBestBlockTime, CConnman* connman
...
+{
+    fBalancesDirty = true;
+    BalancesDidChange();
+}
 
 CAmount CWallet::GetBalance() const
 {
-    CAmount nTotal = 0;
+    if (!fBalancesDirty) {
+        return nBalanceCache;
```
ryanofsky: I think think you need to acquire cs_wallet lock here, otherwise this could return a partial sum if `GetBalance` got called from two threads at the same time and one thread returned `nBalanceCache` while the other thread was in the middle of the `nBalanceCache +=` loop.

Alternately, you could bring back the nTotal local variable and just set `nBalanceCache = nTotal`  atomically.

src/wallet/wallet.cpp
```diff
@@ -1902,95 +1909,123 @@ void CWallet::ResendWalletTransactions(int64_t nBestBlockTime, CConnman* connman
...
+    }
+
+    nBalanceCache = 0;
```
ryanofsky: I think you need to move this line after the `LOCK2(cs_main, cs_wallet);` line. Otherwise if this method is called from two different threads simultaneously, this line could zero out a balance which is in the middle of being added up by the other thread.

Or alternately, delete this line and and set nBalanceCache = nTotal atomically after the loop.

laanwj:
Concept ACK, nice!

src/wallet/wallet.h
```diff
@@ -898,13 +899,26 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
...
+    mutable std::atomic<CAmount> nBalanceCache;
+    mutable std::atomic<CAmount> nUnconfirmedBalanceCache;
+    mutable std::atomic<CAmount> nImmatureBalanceCache;
+    mutable std::atomic<CAmount> nWatchOnlyBalanceCache;
+    mutable std::atomic<CAmount> nUnconfirmedWatchOnlyBalanceCache;
+    mutable std::atomic<CAmount> nImmatureWatchOnlyBalanceCache;
+    std::atomic<bool> fBalancesDirty;
+
+    /** Notification for balance changes */
+    boost::signals2::signal<void ()> BalancesDidChange;
```
ryanofsky: Maybe call it `BalancesChanged` instead of `BalancesDidChange` to be consistent with the naming of other signals (`StatusChanged`, `AddressBookChanged`, etc).

src/wallet/wallet.cpp
jonasschnelli: Oh.. good point.
I'll re-add the temp variable then.

src/wallet/wallet.h
jonasschnelli: Yes. I can do that. I kinda like the signal naming convention from apple. XYWillChange, XYDidChange, etc.: because it allows the listener to know (without reading to much code) if the event was synchronous or has triggered a process with a later update. But lets not overdo it here,.. will change therefore.

jonasschnelli:
Force pushed fixes for issues found by @ryanofsky.

src/qt/walletmodel.cpp
```diff
@@ -137,12 +136,6 @@ void WalletModel::updateBalance()
     }
 }
 
-void WalletModel::updateTransaction()
```
ryanofsky: In commit "[Qt] remove unused polling code"

I think it would be good to squash this commit into "[Qt] use the BalancesChanged signal instead of a 250ms poll timer." It's confusing to see code that checks the `fForceCheckBalanceChanged` variable and some code that sets it removed, while other code that sets it sticks around.

ryanofsky (commented):
See other comments, but it looks like the caching here is broken. The Qt changes seem fine but I think would be easier to understand as 1 commit instead of 3.

src/wallet/wallet.h
```diff
@@ -780,6 +780,7 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
         nRelockTime = 0;
         fAbortRescan = false;
         fScanningWallet = false;
+        fBalancesDirty = true;
```
jonasschnelli: Oh. Yes indeed. I now switched to a cache-invalidation-atomic per balance type and set it to `false` once the cache has been built.

src/qt/walletmodel.cpp
jonasschnelli: Good point. Squashed those two commits together.

jonasschnelli:
Force push fixed @ryanofsky points.

src/wallet/wallet.cpp
```diff
@@ -1914,13 +1935,18 @@ CAmount CWallet::GetBalance() const
             if (pcoin->IsTrusted())
                 nTotal += pcoin->GetAvailableCredit();
         }
+        nBalanceCache = nTotal;
+        fBalancesDirty = false;
```
ryanofsky: In commit "Add atomic cache for balances"

Maybe rename `fBalancesDirty` to `fBalanceDirty` since this only applies to the balance returned from CWallet::GetBalance and not the other balances.

ryanofsky (commented):
utACK a6903788aa105bd26dc6bd5a2784db5109467848

jonasschnelli:
Force push fixed @ryanofsky's nit (and also added the "other balances" atomics to `SetNull()`).

ryanofsky:
ACK 84904a9d8ac5e98558b366bc69338d4ad07dc2f7.

Changes since last review were those Jonas mentioned. Seeing the missing initializations in SetNull() was a little disconcerting, because I would have expected tests to catch errors they would cause. Actually one test did fail (wallet_tests/rescan) but I guess I didn't notice it.

jonasschnelli:
Needed a trivial rebase (`BOOST_FOREACH` -> `for` change).

src/wallet/wallet.cpp
```diff
@@ -812,11 +813,11 @@ bool CWallet::GetAccountPubkey(CPubKey &pubKey, std::string strAccount, bool bFo
...
 void CWallet::MarkDirty()
 {
-    {
-        LOCK(cs_wallet);
-        for (std::pair<const uint256, CWalletTx>& item : mapWallet)
-            item.second.MarkDirty();
-    }
+    MarkBalancesDirty();
+
+    LOCK(cs_wallet);
```
kallewoof: Am I mistaken when I think that the lock on `cs_wallet` will actually happen before the call to `MarkBalancesDirty()` due to scope? If so I would swap these lines for clarity.

kallewoof (commented):
utACK 9ca1140a19d3d39b86d7a6d6dc7ed907fcea2a88 (my QT skills are limited though)

sipa:
@kallewoof No, the lock on cs_wallet will happen after MarkBalancesDirty

src/wallet/wallet.cpp
jonasschnelli: I guess I should move the `LOCK()` up to the top. Because it RAII, I think it doesn't matter in this case (no extra block). The lock should take affect when we enter the function/scope (before `MarkBalancesDirty()`). But maybe I'm wrong.

sipa:
> But maybe I'm wrong

You're wrong.

jonasschnelli:
>> But maybe I'm wrong

> You're wrong.

Okay. Then I'll better be quite.
Moved up the `LOCK()`.

ryanofsky (commented):
utACK 187b3eea4a9da6e589099289632ce0425e269e37. Only change since last review was acquiring wallet lock before the one markdirty call.

laanwj:
I'm trying to understand whether this will introduce a hang in the GUI thread in some cases.

Right now, the recurrent poll timer (not thread! it's a qt timer which runs in the GUI thread) updates the balance *if* it's possible to get the locks using TRY_LOCK. Then it will get them, once, and process all balances.

After this change the locks are no longer TRY, so the `BalancesChanged` signal will always cause the GUI thread to take the wallet locks (through `CWallet::get*Balance()`). Not once, but up to once per type of balance, if they all turn out to be dirty (which is likely).
Taking the `cs_main` and `cs_wallet` lock up to 6 times can take ages when the `cs_main` lock is contended: during initial sync, catching up.

Maybe I'm misunderstanding the code, but if I'm right about this, I'm not convinced that this is an improvement in user experience.

jonasschnelli:
Some of the current heavy "freezes" are happening because of the balance poll thread... it seems to be a significant CPU eater.
Right now, we TRY_LOCK every 250ms (which often LOCKS successfully and then iterates (multiple times) through the complete mapWallet).

bildschirmfoto 2017-06-16 um 14 00 59

jonasschnelli:
Here the complete profiling (master):

It seems like that the pollBalanceThread took 50% of the complete CPU time for the measured run.
(30 seconds, opening debug window and 10-20 times sendtoaddress)

bildschirmfoto 2017-06-16 um 14 09 00

jonasschnelli:
I testes again with this PR and the responsiveness is much better. @laanwj points make sense to me.

Could it be, that the TRY_LOCK every 250ms will result in getting the LOCK very often and then do the heavy balance calculation on the main thread because `QTimer` runs on the Main/GUI thread?

ryanofsky:
> Could it be, that the TRY_LOCK every 250ms will result in getting the LOCK very often and then do the heavy balance calculation on the main thread because QTimer runs on the Main/GUI thread?

That's what I would expect. Have you thought about keeping the cache but bringing back the polling and try lock to address laanwj's point?

jonasschnelli:
@ryanofsky I profiled this PR with only the first commit (atomic caches, keep the polling).
The results are better, but not as good as with the signal/non-polling approach:

Test run: Startup, call `sendtoaddress(getnewaddress(), 1)` 20 times manually

## Master with only the first commit (pure atomic caches):
The polling thread gets measured (through significants) in my profiler with ~33% (very high IMO)
bildschirmfoto 2017-06-16 um 14 58 42

## Master with this PR
The balance updates seems no longer be significant:
bildschirmfoto 2017-06-16 um 15 04 14

ryanofsky:
Unclear to me if @laanwj's concern about losing the try lock is sufficiently addressed with the new testing and discussion at https://botbot.me/freenode/bitcoin-core-dev/msg/87346236/. If it is, maybe this PR is close to ready for merge (so far two utACKs from kallewoof and me). If not, it seems like there is some advantage (50% -> 30% poll thread cpu decrease) and no downside to just merging the first commit for now?

jonasschnelli:
I guess it would be great if someone with (g)perf experience on Linux/Ubuntu and maybe on Windows could benchmark this PR against master.

ryanofsky:
Linux and windows performance tests seem like they would be a good sanity check on the numbers you collected (50% cpu usage spent on balance computation on master, 33% with caching commit, and ~0% with full pr), but is there some reason to think this behavior would be dependent on the platform?

I guess the main thing I'm not clear on is whether the results you collected (assuming they do hold up on other platforms) address laanwj's concern above, or if there is some kind of different testing that needs to be done, or some other safeguard that should be added to prevent the possibility of new hangs.

promag:
@jonasschnelli if still relevant I do it.

jonasschnelli:
@promag: A benchmark would be very good to have.
Ideally you compare master against this PR **and** against only the first commit in this PR (3 versions).
Thanks!

jonasschnelli:
Closing for now...

PR #10254, opened 2017-04-21
[Qt] Remove shutdown poll timer and replace it with a signal

jonasschnelli:


laanwj:
I think we need to keep the polling-based path as well, to handle UNIX signals?

jonasschnelli:
> I think we need to keep the polling-based path as well, to handle UNIX signals?

Not sure. Just added a commit that calls `StartShutdown()` (instead of `fRequestShutdown = true;`).
I guess this should work?

src/init.cpp
```diff
@@ -273,7 +274,7 @@ void Shutdown()
  */
 static void HandleSIGTERM(int)
 {
-    fRequestShutdown = true;
+    StartShutdown();
```
laanwj: This is not a good idea, UNIX signal handlers are extremely limited in what they can do. Basically:
- Set atomic flags
- Send to/receive from pipe
This is why the polling was kept in #10125. It is there *in addition* to notification-based user initiated shutdown.

jonasschnelli:
I wasn't aware of the async-signal-safe limitations.
This seems indeed not simply solvable by a signal.
Closing.

PR #10257, opened 2017-04-22
[test] Add test for getmemoryinfo

jimmysong:
Checks memory before and after a transaction that requires a private key.
Each time, 32 bytes of memory for a private key should be used.
Tested in wallet.py instead of its own file to save testing time.

### DIFF
--- a/test/functional/wallet.py
+++ b/test/functional/wallet.py
@@ -57,8 +57,13 @@ class WalletTest(BitcoinTestFramework):
         assert_equal(len(self.nodes[2].listunspent()), 0)
 
         # Send 21 BTC from 0 to 2 using sendtoaddress call.
+        # Locked memory should use at least 32 bytes to sign each transaction
+        self.log.info("test getmemoryinfo")
+        memory_before = self.nodes[0].getmemoryinfo()
         self.nodes[0].sendtoaddress(self.nodes[2].getnewaddress(), 11)
         mempool_txid = self.nodes[0].sendtoaddress(self.nodes[2].getnewaddress(), 10)
+        memory_after = self.nodes[0].getmemoryinfo()
+        assert(memory_before['locked']['used'] + 64 <= memory_after['locked']['used'])
 
         self.log.info("test gettxout")
         # utxo spent in mempool should be visible if you exclude mempool
@@ -78,7 +83,6 @@ class WalletTest(BitcoinTestFramework):
         # but 10 will go to node2 and the rest will go to node0
         balance = self.nodes[0].getbalance()
         assert_equal(set([txout1['value'], txout2['value']]), set([10, balance]))
-
         walletinfo = self.nodes[0].getwalletinfo()
         assert_equal(walletinfo['immature_balance'], 0)
 


laanwj:
It's good to have a test for `getmemoryinfo`, but I'm not sure we should hardcode 32 bytes per key in the test. It should be acceptable for the wallet implementation to vary this, there's no reason to force a certain allocation policy.

jimmysong:
Made changes per @laanwj and squashed.

laanwj:
Thanks!
utACK https://github.com/bitcoin/bitcoin/pull/10257/commits/32eb3e7eb9cf9b242ec0dc097332e2210d86e0a6

test/functional/wallet.py
```diff
@@ -57,8 +57,12 @@ def run_test (self):
         assert_equal(len(self.nodes[2].listunspent()), 0)
 
         # Send 21 BTC from 0 to 2 using sendtoaddress call.
+        # Locked memory should use at least 32 bytes to sign each transaction
```
paveljanik: How does 64 below correspond to 32 bytes here?

paveljanik: Two `getnewaddress` calls?

jimmysong: Yes, two transactions get signed.

paveljanik: OK, but where is it guaranteed that [locked][used] is not re-used or freed after use? You test some other internal implementation behaviour here, not `getmemoryinfo`.

jimmysong: Good question. @laanwj, are there any guarantees to `getmemoryinfo` that can be tested? I was under the impression that the locked memory is only used for private keys and thus is really only reserved for wallet. But how do we know it won't get freed?

laanwj: Wallet keys are never freed, so this should be OK.

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/10257/commits/32eb3e7eb9cf9b242ec0dc097332e2210d86e0a6

jnewbery:
I'm on the fence about this. It's good to increase code coverage and exercise a new RPC, but I'm not sure this is the best way to go about this. getmemoryinfo isn't obviously linked to wallet functionality, so I don't think it naturally lives in the wallet.py test. In this case it probably makes more sense to create a new test (perhaps also exercising the other RPCs in misc.cpp).

Another concern, which I think you've addressed in this case after https://github.com/bitcoin/bitcoin/pull/10257#issuecomment-297001529 is that tests for getmemoryinfo shouldn't be too prescriptive on returned values. I can imagine this is the kind of thing that could harmlessly change between commits, and we don't want to have a test case that keeps on breaking for no good reason.

jimmysong:
@jnewbery I can write a separate test, but the only way to make getmemoryinfo to change is via usage of private keys which can only happen by spending coins from the wallet. In my mind, that doesn't seem to warrant a separate test. Maybe I'm missing something about how this RPC call works? Or if there's some other way to test this information, I'm open to ideas.

jnewbery:
Really this just me being overly picky. I think a test script should have a clearly defined objective (eg "this is the wallet test - it tests functions a user might want to use the wallet for"). Adding random one line tests to the script dilutes that objective.

To be honest, it's not really an issue. If we add any more tests for getmemoryinfo in future, I think they should be separated off into their own test script, but since this is such a small delta, it's probably ok to live in wallet.py for now.

jimmysong:
Rebased and ready. Thanks @MarcoFalke

paveljanik:
reACK https://github.com/bitcoin/bitcoin/pull/10257/commits/d4668f35ab230083e39ab05857ccae4fe77dcb50

PR #10265, opened 2017-04-23
[wallet] [moveonly] Check non-null pindex before potentially referencing

kallewoof:
The once-a-minute `LogPrintf` in `ScanForWalletTransactions` will do a null-ref if it happens to hit the 60s mark right at the last block (where `pindex` becomes `null`). This was moved up to the top of the block instead, where the `while (pindex && !fAbortRescan)` has checked `pindex` for us.

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -1544,6 +1544,10 @@ CBlockIndex* CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, bool f
         {
             if (pindex->nHeight % 100 == 0 && dProgressTip - dProgressStart > 0.0)
                 ShowProgress(_("Rescanning..."), std::max(1, std::min(99, (int)((GuessVerificationProgress(chainParams.TxData(), pindex) - dProgressStart) / (dProgressTip - dProgressStart) * 100))));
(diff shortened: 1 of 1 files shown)


jonasschnelli:
Nice catch!
utACK 6b816dc56b73025ca8fcb8b2a1e1dab8d18b4fd0

NicolasDorier:
utACK 6b816dc56b73025ca8fcb8b2a1e1dab8d18b4fd0. I am surprised there were no more bug report about that.

mchrostowski:
@kallewoof I would go with temptation and move the `GetTime() >= nNow` check to line 1554 as it is vaguely related to ShowProgress, this could help with readability too.

kallewoof:
@mchrostowski Did you mean [1545](https://github.com/kallewoof/bitcoin/blob/6b816dc56b73025ca8fcb8b2a1e1dab8d18b4fd0/src/wallet/wallet.cpp#L1545) by any chance?

paveljanik:
Concept ACK
@kallewoof Agree with the move to the top of the block.

kallewoof:
This PR is now a MOVEONLY. 4 lines moved up to L1547-1550 from L1560-1563.

laanwj:
utACK c36ea69

PR #10266, opened 2017-04-24
Update policy.h

1mbtxn:
Allow for push of 1mb txn.


gmaxwell:
NAK.  Provides no justification, does not do what it claims (it allows a push of a 250kb transaction, not 1MB), does not address any of the performance challenges with txn over 100KB, does not provide testing for the new behavior (and clearly wasn't tested since it's described incorrectly).

Issue #9374, opened 2016-12-19
Per-network bitcoin.conf files


rebroad:
When running with the -testnet option, I would expect the bitcoin.conf file in the testnet3 directory to be used, yet it uses the mainnet config file instead, yet the debug.log indicates the datadir is the testnet3 dir:-

```
2016-12-19 03:08:58.629380 Using data directory /home/rebroad/.bitcoin/testnet3
2016-12-19 03:08:58.629453 Using config file /home/rebroad/.bitcoin/bitcoin.conf
```

laanwj:
There has never been any functionality to read per-network config files. Config file is always read from the root of the data directory.

However this is more complicated to implement than you'd think, not only because of backward compatiblity, but because bitcoin.conf in the data directory can specify the network to use e.g. `testnet=1` or `regtest=1`. This is used by the tests. It is also useful in the case of specifying a .conf file with `-conf` that is outside any data directory (used for system-wide setups).

laanwj:
One idea of doing this is to keep `bitcoin.conf` as global configuration file, then after that is loaded (and thus the network determined) load an optional per-network configuration that can override these parameters, e.g.:

- `${DATADIR}/network.conf`
- `${DATADIR}/testnet3/network.conf`
- `${DATADIR}/regtest/network.conf`

rebroad:
@laanwj Hmmm. sounds messy; for example, how would 'addnode='s be overridden if there were none mentioned in the testnet config file? I often want to add nodes for the mainnet configuration but not for the testnet configuration. For backwards compatibility (which I agree is important), I'd suggest that if the conf file for testnet or regtest does not exist then it defaults back to the mainnet config file.

laanwj:
My proposal above allows for full flexibility here.

Bitcoin Core will look, in a second configuration stage, for a `network.conf` in the per-network data directory and process this.

If you wanted to add testnet-specific addnodes you'd put these in `${DATADIR}/testnet3/network.conf`. If you want to add mainnet-specific addnotes you'd put these in `${DATADIR}/network.conf`. In the "global" bitcoin.conf you'd not put any addnodes, you can put configuration there that you want to always apply.

morcos:
@laanwj +1 for network.conf idea

rebroad:
I suspect, I would just prefer to use one .conf file per network, so ideally this would allow me to abandon the bitcoin.conf file, and just use network.conf files.

jonasschnelli:
[...]

PR #10267, opened 2017-04-24
New -includeconf argument for including external configuration files

### LINKED ISSUE #10071: Support specifying rpcpassword by file
Original issue: https://github.com/namecoin/namecoin-core/issues/148.

Currently the `rpcpassword` must be specified in the configuration file. On some systems (for example [in NixOS](https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/services/networking/namecoind.nix#L14)) configuration files are stored in world-readable locations. It's a bad idea to store secrets in world-readable locations.

To better support these systems it would be nice if there was a `rpcpasswordFile` configuration parameter. Users can then keep the configuration file in a world-readable location but set  `rpcpasswordFile` to a file with restricted ownership and permissions.

kallewoof:
Fixes: #10071.

Done:
- adds `-includeconf=<path>`, where `<path>` is relative to `datadir` or to the path of the file being read, if in a file
- protects against circular includes
- updates help docs

~~~Thoughts:~~~
- ~~~I am not sure how to test this in a neat manner. Feedback on this would be nice. Will dig/think though.~~~

### DIFF
--- a/doc/release-notes-pr10267.md
+++ b/doc/release-notes-pr10267.md
@@ -1,0 +1,13 @@
+Changed command-line options
+----------------------------
+
+- `-includeconf=<file>` can be used to include additional configuration files.
+  Only works inside the `bitcoin.conf` file, not inside included files or from
+  command-line. Multiple files may be included. Can be disabled from command-
+  line via `-noincludeconf`. Note that multi-argument commands like
+  `-includeconf` will override preceding `-noincludeconf`, i.e.
+
+    noincludeconf=1
+    includeconf=relative.conf
+
+  as bitcoin.conf will still include `relative.conf`.
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -361,6 +361,7 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-debuglogfile=<file>", strprintf(_("Specify location of debug log file. Relative paths will be prefixed by a net-specific datadir location. (default: %s)"), DEFAULT_DEBUGLOGFILE));
     if (showDebug)
         strUsage += HelpMessageOpt("-feefilter", strprintf("Tell other nodes to filter invs to us by our mempool min fee (default: %u)", DEFAULT_FEEFILTER));
+    strUsage += HelpMessageOpt("-includeconf=<file>", _("Specify additional configuration file, relative to the -datadir path (only useable from configuration file, not command line)"));
     strUsage += HelpMessageOpt("-loadblock=<file>", _("Imports blocks from external blk000??.dat file on startup"));
     strUsage += HelpMessageOpt("-maxmempool=<n>", strprintf(_("Keep the transaction memory pool below <n> megabytes (default: %u)"), DEFAULT_MAX_MEMPOOL_SIZE));
     strUsage += HelpMessageOpt("-maxorphantx=<n>", strprintf(_("Keep at most <n> unconnectable transactions in memory (default: %u)"), DEFAULT_MAX_ORPHAN_TRANSACTIONS));
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -445,6 +445,17 @@ void ArgsManager::ParseParameters(int argc, const char* const argv[])
             m_override_args[key].push_back(val);
         }
     }
+
+    // we do not allow -includeconf from command line, so we clear it here
+    auto it = m_override_args.find("-includeconf");
+    if (it != m_override_args.end()) {
+        if (it->second.size() > 0) {
+            for (const auto& ic : it->second) {
+                fprintf(stderr, "warning: -includeconf cannot be used from commandline; ignoring -includeconf=%s\n", ic.c_str());
+            }
+            m_override_args.erase(it);
+        }
+    }
 }
 
 std::vector<std::string> ArgsManager::GetArgs(const std::string& strArg) const
@@ -706,18 +717,40 @@ void ArgsManager::ReadConfigStream(std::istream& stream)
     }
 }
 
-void ArgsManager::ReadConfigFile(const std::string& confPath)
+void ArgsManager::ReadConfigFiles()
 {
     {
         LOCK(cs_args);
         m_config_args.clear();
     }
 
+    const std::string confPath = GetArg("-conf", BITCOIN_CONF_FILENAME);
     fs::ifstream stream(GetConfigFile(confPath));
 
     // ok to not have a config file
     if (stream.good()) {
         ReadConfigStream(stream);
+        // if there is an -includeconf in the override args, but it is empty, that means the user
+        // passed '-noincludeconf' on the command line, in which case we should not include anything
+        if (m_override_args.count("-includeconf") == 0) {
+            std::vector<std::string> includeconf(GetArgs("-includeconf"));
+            {
+                // We haven't set m_network yet (that happens in SelectParams()), so manually check
+                // for network.includeconf args.
+                std::vector<std::string> includeconf_net(GetArgs(std::string("-") + GetChainName() + ".includeconf"));
+                includeconf.insert(includeconf.end(), includeconf_net.begin(), includeconf_net.end());
+            }
+
+            for (const std::string& to_include : includeconf) {
+                fs::ifstream include_config(GetConfigFile(to_include));
+                if (include_config.good()) {
+                    ReadConfigStream(include_config);
+                    LogPrintf("Included configuration file %s\n", to_include.c_str());
+                } else {
+                    fprintf(stderr, "Failed to include configuration file %s\n", to_include.c_str());
+                }
+            }
+        }
     }
 
     // If datadir is changed in .conf file:
--- a/test/functional/feature_includeconf.py
+++ b/test/functional/feature_includeconf.py
@@ -1,0 +1,78 @@
+#!/usr/bin/env python3
+# Copyright (c) 2018 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Tests the includeconf argument
+
+Verify that:
+
+1. adding includeconf to the configuration file causes the includeconf
+   file to be loaded in the correct order.
+2. includeconf cannot be used as a command line argument.
+3. includeconf cannot be used recursively (ie includeconf can only
+   be used from the base config file).
+4. multiple includeconf arguments can be specified in the main config
+   file.
+"""
+import os
+import tempfile
+
+from test_framework.test_framework import BitcoinTestFramework, assert_equal
+
+class IncludeConfTest(BitcoinTestFramework):
+    def set_test_params(self):
+        self.setup_clean_chain = False
+        self.num_nodes = 1
+
+    def setup_chain(self):
+        super().setup_chain()
+        # Create additional config files
+        # - tmpdir/node0/relative.conf
+        with open(os.path.join(self.options.tmpdir, "node0", "relative.conf"), "w", encoding="utf8") as f:
+            f.write("uacomment=relative\n")
+        # - tmpdir/node0/relative2.conf
+        with open(os.path.join(self.options.tmpdir, "node0", "relative2.conf"), "w", encoding="utf8") as f:
+            f.write("uacomment=relative2\n")
+        with open(os.path.join(self.options.tmpdir, "node0", "bitcoin.conf"), "a", encoding='utf8') as f:
+            f.write("uacomment=main\nincludeconf=relative.conf\n")
+
+    def run_test(self):
+        self.log.info("-includeconf works from config file. subversion should end with 'main; relative)/'")
+
+        subversion = self.nodes[0].getnetworkinfo()["subversion"]
+        assert subversion.endswith("main; relative)/")
+
+        self.log.info("-includeconf cannot be used as command-line arg. subversion should still end with 'main; relative)/'")
+        self.stop_node(0)
+        with tempfile.SpooledTemporaryFile(max_size=2**16) as log_stderr:
+            self.start_node(0, extra_args=["-includeconf=relative2.conf"], stderr=log_stderr)
+
+            subversion = self.nodes[0].getnetworkinfo()["subversion"]
+            assert subversion.endswith("main; relative)/")
+            log_stderr.seek(0)
+            stderr = log_stderr.read().decode('utf-8').strip()
+            assert_equal(stderr, 'warning: -includeconf cannot be used from commandline; ignoring -includeconf=relative2.conf')
+
+        self.log.info("-includeconf cannot be used recursively. subversion should end with 'main; relative)/'")
+        with open(os.path.join(self.options.tmpdir, "node0", "relative.conf"), "a", encoding="utf8") as f:
+            f.write("includeconf=relative2.conf\n")
+
+        self.restart_node(0)
+
+        subversion = self.nodes[0].getnetworkinfo()["subversion"]
+        assert subversion.endswith("main; relative)/")
+
+        self.log.info("multiple -includeconf args can be used from the base config file. subversion should end with 'main; relative; relative2)/'")
+        with open(os.path.join(self.options.tmpdir, "node0", "relative.conf"), "w", encoding="utf8") as f:
+            f.write("uacomment=relative\n")
+
+        with open(os.path.join(self.options.tmpdir, "node0", "bitcoin.conf"), "a", encoding='utf8') as f:
+            f.write("includeconf=relative2.conf\n")
+
+        self.restart_node(0)
+
+        subversion = self.nodes[0].getnetworkinfo()["subversion"]
+        assert subversion.endswith("main; relative; relative2)/")
+
+if __name__ == '__main__':
+    IncludeConfTest().main()
--- a/src/bitcoin-cli.cpp
+++ b/src/bitcoin-cli.cpp
@@ -107,7 +107,7 @@ static int AppInitRPC(int argc, char* argv[])
         return EXIT_FAILURE;
     }
     try {
-        gArgs.ReadConfigFile(gArgs.GetArg("-conf", BITCOIN_CONF_FILENAME));
+        gArgs.ReadConfigFiles();
     } catch (const std::exception& e) {
         fprintf(stderr,"Error reading configuration file: %s\n", e.what());
         return EXIT_FAILURE;
--- a/src/bitcoind.cpp
+++ b/src/bitcoind.cpp
@@ -92,7 +92,7 @@ bool AppInit(int argc, char* argv[])
         }
         try
         {
-            gArgs.ReadConfigFile(gArgs.GetArg("-conf", BITCOIN_CONF_FILENAME));
+            gArgs.ReadConfigFiles();
         } catch (const std::exception& e) {
             fprintf(stderr,"Error reading configuration file: %s\n", e.what());
             return false;
--- a/src/interfaces/node.cpp
+++ b/src/interfaces/node.cpp
@@ -52,7 +52,7 @@ class NodeImpl : public Node
     {
         gArgs.ParseParameters(argc, argv);
     }
-    void readConfigFile(const std::string& conf_path) override { gArgs.ReadConfigFile(conf_path); }
+    void readConfigFiles() override { gArgs.ReadConfigFiles(); }
     bool softSetArg(const std::string& arg, const std::string& value) override { return gArgs.SoftSetArg(arg, value); }
     bool softSetBoolArg(const std::string& arg, bool value) override { return gArgs.SoftSetBoolArg(arg, value); }
     void selectParams(const std::string& network) override { SelectParams(network); }
--- a/src/interfaces/node.h
+++ b/src/interfaces/node.h
@@ -48,7 +48,7 @@ public:
     virtual bool softSetBoolArg(const std::string& arg, bool value) = 0;
 
     //! Load settings from configuration file.
-    virtual void readConfigFile(const std::string& conf_path) = 0;
+    virtual void readConfigFiles() = 0;
 
     //! Choose network parameters.
     virtual void selectParams(const std::string& network) = 0;
--- a/src/qt/bitcoin.cpp
+++ b/src/qt/bitcoin.cpp
@@ -616,7 +616,7 @@ int main(int argc, char *argv[])
         return EXIT_FAILURE;
     }
     try {
-        node->readConfigFile(gArgs.GetArg("-conf", BITCOIN_CONF_FILENAME));
+        node->readConfigFiles();
     } catch (const std::exception& e) {
         QMessageBox::critical(0, QObject::tr(PACKAGE_NAME),
                               QObject::tr("Error: Cannot parse configuration file: %1. Only use key=value syntax.").arg(e.what()));
--- a/src/util.h
+++ b/src/util.h
@@ -140,7 +140,7 @@ public:
     void SelectConfigNetwork(const std::string& network);
 
     void ParseParameters(int argc, const char*const argv[]);
-    void ReadConfigFile(const std::string& confPath);
+    void ReadConfigFiles();
 
     /**
      * Log warnings for options in m_section_only_args when
--- a/test/functional/test_runner.py
+++ b/test/functional/test_runner.py
@@ -138,6 +138,7 @@ BASE_SCRIPTS = [
     'p2p_fingerprint.py',
     'feature_uacomment.py',
     'p2p_unrequested_blocks.py',
+    'feature_includeconf.py',
     'feature_logging.py',
     'p2p_node_network_limited.py',
     'feature_blocksdir.py',


jonasschnelli:
Yeah. Why not. This can be useful.

* I would recommend to use `-addconf=` (otherwise user may think it replaces the bitcoin.conf configuration file).
* If I follow `GetConfigFile()` correctly, you can also use absolut paths, right?

NicolasDorier:
Why not making the existing `-config` a repeatable argument.

kallewoof:
@jonasschnelli Good point - will switch to `-addconf=`. Yes, you can use absolute paths. My worry above is for when a user presumes the path is relative to the config file when it is in fact not.

@NicolasDorier I think `-config` simply tells what name to use and defaults to `bitcoin.conf` -- it doesn't actually load the file. This feature lets you load other files arbitrarily from within `bitcoin.conf`.

kallewoof:
[Unsquashed history](https://github.com/kallewoof/bitcoin/tree/feature-config-readconfig-unsquashed): [1](https://github.com/kallewoof/bitcoin/commit/26be3f7925cd9f7f25a1783f4b03362e3eb28183) → [2](https://github.com/kallewoof/bitcoin/commit/fe8850e405298127be3ddff6e009b56b57aeaf25) → [3⊱1](https://github.com/kallewoof/bitcoin/commit/637c01b8cf7e8f08b23cf4a5244361583e4aeaee) → [4⊱2](https://github.com/kallewoof/bitcoin/commit/e17a86e542a9e51d27771a9e572bae7dd2e8c852)

laanwj:
Concept ACK

> My worry above is for when a user presumes the path is relative to the config file when it is in fact not.

Yes, making it relative to the data directory is a good choice. I think we should handle all relative paths in `bitcoind` that way.

> Why not making the existing -config a repeatable argument.

That was also my first thought, but it may just be confusing as it changes the meaning of the option slightly. It's possible that some setups already use multiple `-conf` options, and rely on the overriding behavior.

So I'm good with making it an explicit option.  Another suggestion for the name would be `-includeconf`.

jtimon:
Concept ACK. Don't care much about the name, but what about -extraconf ?

kallewoof:
From the given suggestions I think `includeconf` is the most clear so I'll switch to that.

@laanwj:

> Yes, making it relative to the data directory is a good choice. I think we should handle all relative paths in bitcoind that way.

To clarify, you mean that the relative path inside /dir/file.conf should be /dir/, not <bitcoin datadir>, right? It will require some lines of code I bet but I think that makes sense too.

kallewoof:
The `includeconf` feature now defines the path relative to the file being read, if any. For command line, it is [datadir], for `/dir/abc.conf` it is `/dir/`. I tested this with
```
src/testreadconfig/bitcoin.conf: [...] includeconf=../global.conf
src/global.conf: includeconf=secrets.conf
src/secrets.conf: rpcpassword=foo
```
and ensured `bitcoin-cli` with password `foo` worked and password `bar` did not. 

[[...]](https://github.com/kallewoof/bitcoin/tree/feature-config-readconfig-unsquashed): → [5⊱1](https://github.com/kallewoof/bitcoin/commit/14501f2566ca056f27f9c45b3b288af6a426b224) → [6⊱2](https://github.com/kallewoof/bitcoin/commit/d12f84a762e0a7311b4b8ac467fce8ec7ba64bfb) → [7⊱1](https://github.com/kallewoof/bitcoin/commit/ea1e4595d9d46f6bed42c3e869f2d16fd70c3eec) → [8⊱2](https://github.com/kallewoof/bitcoin/commit/ef3e7ccc816b7e5f00aaca310b676c549f6f6640)

laanwj:
> To clarify, you mean that the relative path inside /dir/file.conf should be /dir/, not [bitcoin datadir], right? It > will require some lines of code I bet but I think that makes sense too.

Yes, seems good to me too. I wasn't thinking about relative includes in other includes.

kallewoof:
To clarify, the code now does what @laanwj suggested.

laanwj:
> I am not sure how to test this in a neat manner. Feedback on this would be nice. Will dig/think though.

Some ideas:
- RPC test that creates a tree of bitcoin config files including each other beneath the data directory
- Starts a node w/ `-includeconf=<path>`
- Then interrogate node over RPC to verify the files got included, in the right order

To achieve the latter the option `uacomment=` is useful, as these will be added to an array, then querying `getnetworkinfo` to see if the `(...)` part of the `subversion` matches expected content and order.

jnewbery:
@laanwj's suggested test method seems sensible. I'm happy to review that or lend a hand implementing it. Feel free to reach me on IRC.

kallewoof:
Thanks for the suggestion! I added a test that checks for load order and ensures circular include is guarded against. @jnewbery review would be wonderful :)

jnewbery:
> If anyone has ideas why travis is failing I'd appreciate it

You've made `ReadConfigFile()` recursive (through `ProcessSetting()`). `ReadConfigFile()` locks cs_args, and then at the end calls `ClearDatadirCache()`, which locks csPathCached. That means that the bottom-most `ReadConfigFile()` locks csPathCached while cs_args is still held.

There's already a function that locks in the other order: `GetDataDir()` locks csPathCached and then locks cs_args (in its call to `IsArgSet()`).

If those two functions are called in different threads, we'd have a deadlock.

There's a CPP_FLAG option that checks lock ordering `CPPFLAGS=-DDEBUG_LOCKORDER`, which is used in Travis build 5. That's why that build is failing. You can repro locally by running `configure` with that option.

You can fix this by not locking cs_args recursively.

kallewoof:
@jnewbery Thanks a lot for the explanation! I should've paid closer attention to locks considering the added recursiveness.

97ee63b fixes this by moving the conditionally-locked code into a new `ReadConfigStream` function which is called with locking/clearing in one case and without in the other, based on a `bool lockAndClear` added to `ReadConfigFile`.

(Also had to tweak tests a tiny bit; 8fb6511.)

jnewbery:
I think you've introduced a subtle bug here. If `-datadir` is configured in one of the additional config files, then the datadir cache won't be cleared, which means that bitcoind will continue to use the old datadir.

I think you should try to not make ReadConfigFile recursive. For me, it would be acceptable to only allow one level of redirection here (ie the "base" config file can specify `-includeconf` config file, but those included config files cannot themselves include other config files). I think that would be a simpler model and would remove a whole bunch of potential bugs (circular references, blowing the stack through too many `-includeconf` redirects, etc).

kallewoof:
@jnewbery Hm, no the datadir cache is cleared after any recursions happen, which means it is always cleared, just not directly after the config file has been parsed. There are two cases:
1. ParseParameters (https://github.com/kallewoof/bitcoin/blame/feature-config-readconfig-unsquashed/src/util.cpp#L407 called from https://github.com/kallewoof/bitcoin/blame/feature-config-readconfig-unsquashed/src/bitcoind.cpp#L75) -- this will work as normal and does not require cache clearing.
2. Nested ReadConfigs from the initial bitcoin.conf file: https://github.com/kallewoof/bitcoin/blame/feature-config-readconfig-unsquashed/src/bitcoind.cpp#L107 calls `ReadConfigFile` with the `lockAndClear` flag set; recursion then happens in ProcessSetting (L401) via ReadConfigStream (L622) with `lockAndClear` unset. Eventually this gets back to original caller which leaves ReadConfigStream and gets to https://github.com/kallewoof/bitcoin/blame/feature-config-readconfig-unsquashed/src/util.cpp#L649 which clears the datadir cache.
Or am I missing something?

As for forbidding multiple levels of recursion, I think the value outweighs the issues personally (and I addressed circular refs I believe), but if people think it's not worth it I'll restrict it to one include.

jnewbery:
@kallewoof yes you're right. datadir cache is cleared after all files are read. My mistake.

I still don't like the recursion and the fact that there can be multiple levels of imports. It means there are more edge cases and unexpected behaviour. For example, if `-includeconf` is included as a command line parameter, then the includeconf file is read *before* the regular conf file, and so takes precedence. If an includeconf line is included in the regular conf file, then it is read *during* the regular config file, and which settings are taken from the regular conf file and which are taken from the includeconf file depend on the ordering of settings in the regular conf file.

The new `warnOnFailure` and `lockAndClear` bool arguments to `ReadConfigFile()` seem pretty strange to me. They're only used when `ReadConfigFile()` is being called recursively, and they control a large chunk of the behaviour within `ReadConfigFile()`. That's a clue to me that maybe the functionality isn't split up correctly - perhaps the locking/clearing should be in an outer function which calls an inner function for each of the config files?

Finally, you've introduced a new crash bug. If `-conf` or `-includeconf` don't refer to a valid file, the bitcoind will crash on startup. Here's the backtrace:

```
Core was generated by `bitcoind -datadir=/tmp/user/1000/test6p7xn_xt/856/node0 -server -keypool=1 -dis'.
Program terminated with signal SIGABRT, Aborted.
#0  0x00007f3b77df0428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
54	../sysdeps/unix/sysv/linux/raise.c: No such file or directory.
(gdb) bt
#0  0x00007f3b77df0428 in __GI_raise (sig=sig@entry=6) at ../sysdeps/unix/sysv/linux/raise.c:54
#1  0x00007f3b77df202a in __GI_abort () at abort.c:89
#2  0x00007f3b7873284d in __gnu_cxx::__verbose_terminate_handler() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#3  0x00007f3b787306b6 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#4  0x00007f3b78730701 in std::terminate() () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007f3b78730919 in __cxa_throw () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#6  0x00007f3b79f29f82 in boost::filesystem::detail::canonical(boost::filesystem::path const&, boost::filesystem::path const&, boost::system::error_code*) () from /usr/lib/x86_64-linux-gnu/libboost_filesystem.so.1.58.0
#7  0x000055e5c58b2ae2 in boost::filesystem::canonical (base=..., p=...) at /usr/include/boost/filesystem/operations.hpp:459
#8  GetConfigFile (confPath="global.conf", relativePath="") at util.cpp:603
#9  0x000055e5c58b4353 in ArgsManager::ProcessSetting (this=this@entry=0x55e5c5db1fa0 <gArgs>, strKey="-includeconf", strValue="global.conf", relativePath="") at util.cpp:386
#10 0x000055e5c58b487d in ArgsManager::ParseParameters (this=0x55e5c5db1fa0 <gArgs>, argc=<optimized out>, argv=<optimized out>) at util.cpp:424
#11 0x000055e5c5653d4a in ParseParameters (argv=0x7fff5492e618, argc=14) at util.h:263
#12 AppInit (argc=14, argv=0x7fff5492e618) at bitcoind.cpp:75
#13 0x000055e5c5648bef in main (argc=14, argv=0x7fff5492e618) at bitcoind.cpp:196
(gdb) quit
```

If I use an invalid filename for `-conf` on master. I don't see this crash.

kallewoof:
@jnewbery Thanks a lot for all the feedback.

> I still don't like the recursion and the fact that there can be multiple levels of imports. It means there are more edge cases and unexpected behaviour. For example, if -includeconf is included as a command line parameter, then the includeconf file is read before the regular conf file, and so takes precedence. If an includeconf line is included in the regular conf file, then it is read during the regular config file, and which settings are taken from the regular conf file and which are taken from the includeconf file depend on the ordering of settings in the regular conf file.

That is the case without recursion as well, unless we forbid command line case. (Which I can do already, even while keeping recursion.)

> The new warnOnFailure and lockAndClear bool arguments to ReadConfigFile() seem pretty strange to me. They're only used when ReadConfigFile() is being called recursively, and they control a large chunk of the behaviour within ReadConfigFile(). That's a clue to me that maybe the functionality isn't split up correctly - perhaps the locking/clearing should be in an outer function which calls an inner function for each of the config files?

Hm.. the warnOnFailure was just a nice-to-have to inform the user when an explicitly included file didn't actually exist, but I can remove it for cleanliness. Since I moved most of ReadConfigFile into ReadConfigStream, the only remaining stuff was the caching stuff, which doesn't feel odd to me. I'm not actually sure why you consider this to be a problem: the ReadConfigFile is mostly there to do or not do the locking and data cache clearing, and the ReadConfigStream is there to do the actual reading/parsing part.

That said, I'm not overly attached to the idea of allowing recursion, so unless someone speaks *for* it I am going to try to simplify the code to only allow one single include and to only allow it in the file, i.e. not from command line. I believe that would address most of your concerns.

src/util.cpp
```diff
@@ -592,26 +592,41 @@ fs::path GetConfigFile(const std::string& confPath)
...
-            if (mapArgs.count(strKey) == 0)
-                mapArgs[strKey] = strValue;
-            mapMultiArgs[strKey].push_back(strValue);
+        ReadConfigStream(streamConfig);
+        if (mapArgs.count("-includeconf")) includeconf = mapArgs["-includeconf"];
+    }
+    if (includeconf != "") {
+        fs::path includeFile = GetConfigFile(includeconf);
+        fs::ifstream includeConfig(includeFile);
+        if (includeConfig.good()) {
```
jnewbery: Can we have an else clause that prints an error message if we fail to open an includeconfig file?

src/util.cpp
```diff
@@ -592,26 +592,41 @@ fs::path GetConfigFile(const std::string& confPath)
...
-            std::string strValue = it->value[0];
-            InterpretNegativeSetting(strKey, strValue);
-            if (mapArgs.count(strKey) == 0)
-                mapArgs[strKey] = strValue;
-            mapMultiArgs[strKey].push_back(strValue);
+        ReadConfigStream(streamConfig);
+        if (mapArgs.count("-includeconf")) includeconf = mapArgs["-includeconf"];
+    }
+    if (includeconf != "") {
+        fs::path includeFile = GetConfigFile(includeconf);
```
jnewbery: Why not directly:

```
fs::ifstream includeConfig(GetConfigFile(includeconf));
```

(like above)

src/util.cpp
```diff
@@ -592,26 +592,41 @@ fs::path GetConfigFile(const std::string& confPath)
     return pathConfigFile;
 }
 
+void ArgsManager::ReadConfigStream(fs::ifstream& streamConfig)
+{
+    std::set<std::string> setOptions;
```
jnewbery: Is it possible to add a log to this function to output which file it's reading, or is it too early to start logging?

Sorry - by my earlier comment, I didn't mean you should remove the logging on failure, just that the structure of the function suggested to me that it shouldn't be called recursively.

test/functional/includeconf.py
```diff
@@ -0,0 +1,43 @@
...
+        with open(os.path.join(self.options.tmpdir+"/node0", "bitcoin.conf"), 'a', encoding='utf8') as f:
+            f.write("uacomment=main\nincludeconf=relative.conf\n")
+        # subversion should end with "(main; relative)/"
+
+    def __init__(self):
+        super().__init__()
+        self.setup_clean_chain = False
+        self.num_nodes = 1
+
+    def setup_network(self):
```
jnewbery: not required. `setup_network()` in the base class just calls `setup_nodes()` when there's only one node.

test/functional/includeconf.py
```diff
@@ -0,0 +1,43 @@
...
+        self.setup_nodes()
+
+    def run_test (self):
```
jnewbery: no space before open parentheses please.

test/functional/includeconf.py
```diff
@@ -0,0 +1,43 @@
...
+import os
+from test_framework.test_framework import BitcoinTestFramework
+
+class IncludeConfTest(BitcoinTestFramework):
+
+    def setup_chain(self):
+        super().setup_chain()
+        # Create additional config file
+        # - tmpdir/node0/relative.conf
+        with open(os.path.join(self.options.tmpdir+"/node0", "relative.conf"), "w", encoding="utf8") as f:
```
jnewbery: nit: spaces around `+` please

test/functional/includeconf.py
```diff
@@ -0,0 +1,43 @@
...
+    def __init__(self):
+        super().__init__()
+        self.setup_clean_chain = False
+        self.num_nodes = 1
+
+    def setup_network(self):
+        self.setup_nodes()
+
+    def run_test (self):
+        '''
```
jnewbery: My personal preference is to have the description of the test in the module-level doc string (since that's the first thing people see when they open the file).

test/functional/includeconf.py
```diff
@@ -0,0 +1,43 @@
+#!/usr/bin/env python3
```
jnewbery: It'd be nice to have a couple of other subtests:

- including `--includeconf` in the command line argument has no effect
- including `includeconf` in an includeconf file has no effect.

jnewbery (commented):
This looks much much better. Thanks @kallewoof! I much prefer this implementation, which restricts the way `--includeconf` can be used and cuts out lots of edge cases.

If it turns out that people need more flexibility, we can extend this later to allow multiple includeconfs, including multiple levels. That's much easier than going the other way (from less restrictive to more restrictive).

I have a few nits, mainly around naming.

I think there's one bug, or at least confusing behaviour. You said in your comment that you don't allow `-includeconf` from the command line, but your new code *is* allowing that. Further, if an `-includeconf` is included on the command line and in the base config file, then the command line `-importconf` takes precedence. I don't think that's what we want. To fix that, I think the first thing `ReadConfigFile()` should do clear the `-includeconf` argument if it's set.

src/util.cpp
```diff
@@ -592,26 +592,41 @@ fs::path GetConfigFile(const std::string& confPath)
     return pathConfigFile;
 }
 
+void ArgsManager::ReadConfigStream(fs::ifstream& streamConfig)
```
kallewoof: I named if for the fact it got an`ifstream` as argument, but you're right that it's probably better to just name it `ReadConfigFile`.

src/util.cpp
kallewoof: I am logging before the call to this method now. That should cover it I think.

src/util.cpp
```diff
@@ -592,26 +592,41 @@ fs::path GetConfigFile(const std::string& confPath)
...
+        std::string strValue = it->value[0];
+        InterpretNegativeSetting(strKey, strValue);
+        if (mapArgs.count(strKey) == 0) {
+            mapArgs[strKey] = strValue;
+        }
+        mapMultiArgs[strKey].push_back(strValue);
+    }
+}
+
 void ArgsManager::ReadConfigFile(const std::string& confPath)
```
kallewoof: Done!

test/functional/includeconf.py
kallewoof: I'm not sure a neat and tidy way to test that. Should I just make multiple classes, one for each, and then call them one at a time in the bottom `if __name__ == '__main__':`?

kallewoof:
Updated and squashed. [Unsquashed history](https://github.com/kallewoof/bitcoin/tree/feature-config-readconfig-unsquashed).

test/functional/includeconf.py
jnewbery: These don't actually need to be separate tests. You can just have the following in `setup_chain()`:

```python
        with open(os.path.join(self.options.tmpdir + "/node0", "relative.conf"), "w", encoding="utf8") as f:
            f.write("uacomment=relative\nincludeconf=relativeofrelative.conf\n")
        with open(os.path.join(self.options.tmpdir + "/node0", "bitcoin.conf"), 'a', encoding='utf8') as f:
            f.write("uacomment=main\nincludeconf=relative.conf\n")
        with open(os.path.join(self.options.tmpdir + "/node0", "commandline.conf"), "w", encoding="utf8") as f:
            f.write("uacomment=commandline\n")
        with open(os.path.join(self.options.tmpdir + "/node0", "relativeofrelative.conf"), "w", encoding="utf8") as f:
            f.write("uacomment=relativeofrelative\n")
```

and the following line in `__init__()`:

```python
        self.extra_args = [['-includeconf=commandline.conf']]
```

and then keep the same assert:

```python
        assert subversion.endswith("main; relative)/")
```

to check that neither `comandline.conf` nor `relativeofrelative.conf` have been included.

(I recommend you also update the docstring to describe what's happening)

jnewbery:
Looks great. Tested ACK bc4f7a4b3f614dc2125c6af60da448606b622688

One suggestion for adding to the testcase. Up to you whether you want to take it.

jtimon:
This conflicts a little bit with https://github.com/bitcoin/bitcoin/pull/8994/commits/7246faea3e4421e609e6a24294bd23a45d2343de

There I use the old ArgsManager::ReadConfigFile(path) which this PR remove.
Could you conserve that method even if it's temporarily unused (although preferrably using it internally like "ArgsManager::ReadConfigFile(fs::ifstream& streamConfig)")?

```
void ArgsManager::ReadConfigFile(const std::string& confPath)
{
     fs::ifstream streamConfig(GetConfigFile(confPath));
      if (!streamConfig.good())
          return; // No bitcoin.conf file is OK
// ...
}
```
Or perhaps I should restore it later if this PR gets merged first.
utACK bc4f7a4b3f614dc2125c6af60da448606b622688 beyond that.

src/util.cpp
```diff
@@ -596,26 +596,47 @@ fs::path GetConfigFile(const std::string& confPath)
...
+        LogPrintf("Attempting to include configuration file %s\n", includeconf.c_str());
+        fs::ifstream includeConfig(GetConfigFile(includeconf));
+        if (includeConfig.good()) {
+            LOCK(cs_args);
+            ReadConfigFile(includeConfig);
+        } else {
+            LogPrintf("Failed to include configuration file %s\n", includeconf.c_str());
         }
     }
     // If datadir is changed in .conf file:
```
jtimon: Shouldn't ClearDatadirCache() be called from ArgsManager::ReadConfigFile(fs::ifstream& streamConfig) too?

kallewoof: ReadConfigFile is only called from ReadConfigFiles, which clears the data cache already, so it should be fine I think.

kallewoof:
@jtimon The method was not removed, it was renamed. If you change to `ReadConfigFiles` you should get the exact same result after this is merged. Let me know if that is not the case!

jtimon:
The method that is removed is the one that allows you to call with a path, ```ArgsManager::ReadConfigFile(const std::string& confPath)```.  The new ```ReadConfigFiles()``` will work just fine here, but in #8994 I cannot use it, because want I want is precisely to load from a different file (and not allow -includeconf or command line for "custom chainparams" configuration). If I use ```AgsManager::ReadConfigFile(fs::ifstream& streamConfig)``` directly from chainparams.cpp, not only I will duplicate code, but also call things like GetConfigFile(), which I would prefer not to call from chainparams.cpp. So my conclusion is that I would just restore ```ArgsManager::ReadConfigFile(const std::string& confPath)```. That's why I ask that you maintain it even if you don't need it for anything (but it can also be removed and then restored, it's not a big deal)

kallewoof:
Ohh, damn... It seems the removal of the path may have been premature. What do you think of `ReadConfigFiles()` → `ReadConfigFile(std::string path = "", bool allowIncludes = true)` whose defaults does exactly what `ReadConfigFiles` does now? @jtimon @jnewbery

jnewbery:
I think ReadConfigFiles() shouldn't take any arguments and should be responsible for finding and reading all config files. I don't understand why you'd want to read config files from other places in the codebase in #8994. It seems to me to be much simpler to reason about what config is loaded if it all happens in one place.

jtimon:
@kallewoof yeah, I think that would work too, and you could still call it without parameters. That solution is very simple for me to "restore" on #8994 if people don't like it here. There's no need to slow this down if other people don't like my request. Thank you for offering a good and simple solution to my concern.

@jnewbery I don't want the chain custom parameters to be perceived as "config". They select the chain you will be on, it is mostly intended to create new testnets, and sharing a "testnet config file" for a newly created one could be a useful thing. But perhaps that's something to discuss on #8994 rather than here. I still have it on a separated commit in case people prefer to allow consensus critical parameters to be passed from command line or the other config files that con be loaded.

jnewbery:
@jtimon 

> I don't want the chain custom parameters to be perceived as "config".

I actually think having the customchain config file contain general config could be useful. There seemed to be some enthusiasm for #9374 , which is similar in nature - it allows a separate config file for each separate chain.

This is a bit of a sidetrack from this PR though, which I think is a good and useful improvement. This needs rebasing because of a conflict in test_runner.py. Assuming just that needs changing, then I still ACK this. I'll give some more feedback in #9374.

jnewbery:
alternative: this PR could move the:

```
fs::ifstream streamConfig(GetConfigFile(confPath));
if (!streamConfig.good())
```
lines into the new `ReadConfigFile()` function instead of leaving them in `ReadConfigFiles()`, and have `ReadConfigFile()` take a `std::string` instead of a `fs::ifstream&`. That would remove some code duplication (since that's called for both the 'base' config file and the -includeconf file).

#9374 could then call `ReadConfigFile()` exactly as before.

kallewoof:
@jnewbery I would need to either change the return value to be a success flag for `streamConfig.good()`, or add an additional `bool warnOnFailure` flag, like I had before, as we currently don't warn for the main config missing, but we do warn for `includeconf`s.

jnewbery:
Required rebase after `gArgs` PR #10607. I've rebased here: https://github.com/jnewbery/bitcoin/tree/pr10267

I've also implemented my suggested change to make `ReadConfigFile` take a `std::string` here: https://github.com/jnewbery/bitcoin/tree/pr10267.1.

@kallewoof - can you let me know what you think? I think this resolves @jtimon's concerns so this doesn't conflict badly with his PR.

kallewoof:
@jnewbery Looks great! I rebased and adopted your changes. I tried to pull your commits into my branch but things exploded. Sorry about that. :/
Will squash if you think 9b70900 looks OK. (I picked slightly different varnames from you -- any reason for using snake case? I don't see it used elsewhere.)

jnewbery:
@kallewoof - no problem. My branch was just an example. Feel free to do whatever you want with it. (for future reference if you want to take the commits from my branch, easiest way is to add github.com/jnewbery/bitcoin as a remote, then `git fetch` from my remote, then `git reset --hard <sha of my branch>` from your branch).

For snake case convention, see 'symbol naming conventions' here: https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md#developer-notes . No need to change old code to use the conventions, but new PRs should follow them if possible.

kallewoof:
@jnewbery That's what I did, but when I tried rebasing master on top of it things exploded. I'll try again.

As for the symbol naming convention, I'm amazed I didn't catch that considering I'm a big fan of conventions myself. Will definitely start using it. The `m_foo` thing for class member vars looks ugly as hell to me, but ah well.

kallewoof:
@jnewbery Digging, but:
```
POTENTIAL DEADLOCK DETECTED
Previous lock order was:
 (1) cs_args  util.cpp:628
 (2) csPathCached  util.cpp:558
Current lock order is:
 (2) csPathCached  util.cpp:558
 (1) cs_args  util.cpp:432
Assertion failed: (false), function potential_deadlock_detected, file sync.cpp, line 98.
```

promag:
Needs rebase. Is this still valid?

kallewoof:
Rebased. It is valid, yes.

MeshCollider:
This could be very useful in the context of https://github.com/bitcoin/bitcoin/pull/11862, using `-regtest.includeconf` for example

kallewoof:
@laanwj Ping (since you self-assigned)

src/util.cpp
```diff
@@ -631,26 +631,47 @@ fs::path GetConfigFile(const std::string& confPath)
...
-        {
-            // Don't overwrite existing settings so command line settings override bitcoin.conf
-            std::string strKey = std::string("-") + it->string_key;
-            std::string strValue = it->value[0];
-            InterpretNegativeSetting(strKey, strValue);
-            if (mapArgs.count(strKey) == 0)
-                mapArgs[strKey] = strValue;
-            mapMultiArgs[strKey].push_back(strValue);
+        ReadConfigFile(streamConfig);
+        if (mapArgs.count("-includeconf")) includeconf = mapArgs["-includeconf"];
```
jnewbery: I think it'd be useful to be able to include multiple `-includeconf` arguments in the main config file (for example, if we want a `main.includeconf`, `test.includeconf` and `regtest.includeconf`). Can you make this a string vector and then iterate through the `includeconf`s?

test/functional/includeconf.py
```diff
@@ -0,0 +1,38 @@
...
```
jnewbery: This test now needs to be named `feature_includeconf`

jnewbery (commented):
A few comments inline.

I suggest you squash commits 1 and 2 (and please remove the #<issuenumber> from the commit message!)

I've made some suggested changes to the functional test here: https://github.com/jnewbery/bitcoin/commit/968c9158f0bffac8c91282f1789a5a1808ed13b9 . Feel free to take any of the code you want and squash into your commit.

Needs release notes.

src/init.cpp
```diff
@@ -338,6 +338,7 @@ std::string HelpMessage(HelpMessageMode mode)
     if (showDebug) {
         strUsage += HelpMessageOpt("-dbbatchsize", strprintf("Maximum database write batch size in bytes (default: %u)", nDefaultDbBatchSize));
     }
+    strUsage += HelpMessageOpt("-includeconf=<file>", _("Specify additional configuration file, relative to the -datadir path"));
```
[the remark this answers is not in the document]
kallewoof: Oh, coolness.

src/util.cpp
kallewoof: That makes sense, yeah.

kallewoof:
@jnewbery Thanks a lot for review / code suggestions. I'll add multiple includeconf support and then push updated code.

kallewoof:
This PR times out on job 2 without the timeout bump in 47848d0. I have not investigated the cause for this increase, and I find it unlikely that the new tests would be the actual cause.

jnewbery:
The travis job is timing out when running `make check` (the unit tests):

```
$ if [ "$RUN_TESTS" = "true" ]; then travis_wait 40 make $MAKEJOBS check VERBOSE=1; fi    1913.41s
```

Here's the end of the output from that command:

```
make[1]: Leaving directory `/home/travis/build/bitcoin/bitcoin/build/bitcoin-i686-w64-mingw32'
/home/travis/.travis/job_stages: line 169: 28077 Terminated              travis_jigger $! $timeout $cmd
```

@MarcoFalke / @laanwj - any idea why the unit tests would stall for half an hour here?

ryanofsky:
> @MarcoFalke / @laanwj - any idea why the unit tests would stall for half an hour here?

Something must have changed recently, but I think one reason why tests are slow is that they are running using wine to emulate windows. I'm also seeing the same timeouts with 11851.

test/functional/feature_includeconf.py
```diff
@@ -0,0 +1,71 @@
...
+
+Verify that:
+
+1. adding includeconf to the configuration file causes the includeconf
+   file to be loaded in the correct order.
+2. includeconf cannot be used as a command line argument.
+3. includeconf cannot be used recursively (ie includeconf can only
+   be used from the base config file).
+4. multiple includeconf arguments can be specified in the main config
+   file (Currently fails).
```
jnewbery: Can remove ` (Currently fails)`

jnewbery (commented):
utACK 47848d0fb0705facd7d4cf24c50d450386155f5a (although I'll defer to @MarcoFalke on whether the travis timeout bump is a good idea)

MarcoFalke:
I am fine with bumping to 40 or 50 minutes. In case we need that much time, it will time out due to the global limit anyway. The long term solution would be to somehow speed up the unit tests on Wine, but that doesn't have to block this (and other) pull requests.

Could cherry-pick that change to a separate pull request, though.

jnewbery:
> Could cherry-pick that change to a separate pull request, though.

Here you go: #12772

test/functional/feature_includeconf.py
kallewoof: Thanks, fixed.

kallewoof:
I'll just keep that commit in here until the other is merged and then drop it.

jnewbery:
#11862 is merged. Please rebase.

I'll commit to re-reviewing this once it's rebased.

kallewoof:
Rebased. I had to do some tweaking, because the `m_network` vector is empty when the check for `-includeconf` is done, but the new `bitcoin.conf` files start with `[regtest]`, so the parameter comes out as `-regtest.includeconf`, but `GetArgs("-includeconf")` is empty.

My workaround was to prepend the `-includeconf` call before `[regtest]`, but a better solution is to do the `-includeconf` stuff after `m_network` is populated. I'm a bit surprised it isn't, at this stage.

src/init.cpp
```diff
@@ -358,6 +358,7 @@ std::string HelpMessage(HelpMessageMode mode)
...
```
jnewbery: Help text could be expanded a bit to say that this can't be set from the command line - only from within the .conf file.

src/util.cpp
```diff
@@ -943,18 +943,40 @@ void ArgsManager::ReadConfigStream(std::istream& stream)
...
 
+    const std::string confPath = GetArg("-conf", BITCOIN_CONF_FILENAME);
     fs::ifstream stream(GetConfigFile(confPath));
 
     // ok to not have a config file
     if (stream.good()) {
         ReadConfigStream(stream);
+        std::vector<std::string> includeconf(GetArgs("-includeconf"));
+        {
+            std::vector<std::string> includeconf_net(GetArgs(std::string("-") + GetChainName() + ".includeconf"));
```
jnewbery: I was going to say that this isn't required because `ArgsManager::GetArgs()` already checks the network-specific config. However, we haven't yet set `m_network` (that only happens in `SelectParams()`) so you're right and you do need this manual check.

I don't know if there's a better way to avoid doing this manually. Perhaps @ajtowns can suggest something? At the least, I think you should add a comment like:

```
        // We haven't set m_network yet (that happens in SelectParms()), so manually check
        // for network.includeconf args.
```
You can also remove the block/braces.

jnewbery (commented):
Looks great. Thanks for sticking with this.

A couple of nits inline.

src/init.cpp
kallewoof: I thought I had done that already. Fixing.

src/util.cpp
kallewoof: Good point on adding comment. I added braces to throw `includeconf_net` out of scope as it was no longer used, and this makes that obvious.

src/init.cpp
kallewoof: Updated. Feedback welcome!

kallewoof:
Addressed @jnewbery nits (sans bracket removal).

src/util.cpp
ajtowns: I think this is subtly "wrong" fwiw: if you have `[main] includeconf=foo.conf` in bitcoin.conf, and then set `testnet=1` in foo.conf, you'll have foo.conf included, and be running testnet not mainnet, so you "shouldn't" have included foo.conf (and any other settings from the [main] section won't have been used). That might be too subtle to worry about though.

It's perhaps less subtly wrong in that if you said `includeconf=force-testnet.conf [testnet] includeconf=testnet-params.conf`, with force-testnet.conf setting `testnet=1`, you would not end up including testnet-params.conf, whereas if you set `-testnet` on the command line, you would.

The easiest approach might be to not allow `[regtest] includeconf=rt.conf` at all, but that would probably be annoying (particularly with custom chains).

A thought I had was that maybe `[regtest] includeconf=rt.conf` should treat all the options in rt.conf as being network-specific, so if rt.conf says `connect=10.10.10.10 [main] connect=10.10.10.20 [regtest] connect=10.10.10.30`, that would get interpreted as if it was `[regtest] connect=10.10.10.10 connect=10.10.10.30`. That would avoid the inconsistency above, in that it would turn `[main] includeconf=foo.conf` into essentially `[main] testnet=1` which would just be ignored.

I think it might be better to do it as four phases though: (1) read "-conf", (2) read the includeconfs that aren't under a section, (3) work out the chain based on everything noted so far, and (4) read the "-chain.includeconf" options from -conf (but not letting that change the chain).

test/functional/feature_includeconf.py
```diff
@@ -0,0 +1,72 @@
...
+        subversion = self.nodes[0].getnetworkinfo()["subversion"]
+        assert subversion.endswith("main; relative)/")
+
+        self.log.info("-includeconf cannot be used as command-line arg. subversion should still end with 'main; relative)/'")
+        self.restart_node(0, extra_args=["-includeconf=relative2.conf"])
+
+        subversion = self.nodes[0].getnetworkinfo()["subversion"]
+        assert subversion.endswith("main; relative)/")
+
+        self.log.info("-includeconf cannot be used recursively. subversion should end with 'main)/'")
```
ajtowns: I was expecting subversion to end with "main; relative", which would ensure relative.conf's options are being used despite it trying to recurse into relative2. No big deal either way.

ajtowns (approved):
Some nits and suggestions, but looks good.

doc/release-notes.md
```diff
@@ -96,6 +96,13 @@ Low-level RPC changes
   with any `-wallet=<path>` options, there is no change in behavior, and the
   name of any wallet is just its `<path>` string.
 
+Changed command-line options
+-----------------------------
```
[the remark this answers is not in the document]
kallewoof: Should that be a new file with only the changes, or a copy of `doc/release-notes.md`? Never seen that way of adding release notes before.

src/util.cpp
```diff
@@ -943,18 +943,42 @@ void ArgsManager::ReadConfigStream(std::istream& stream)
...
-void ArgsManager::ReadConfigFile(const std::string& confPath)
+void ArgsManager::ReadConfigFiles()
 {
     {
         LOCK(cs_args);
         m_config_args.clear();
+
+        // we do not allow -includeconf from command line, so we clear it here
+        auto it = m_override_args.find("-includeconf");
+        if (it != m_override_args.end()) m_override_args.erase(it);
```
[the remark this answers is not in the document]
kallewoof: A warning would probably be helpful, yeah. Will look into moving into `ParseParameters`.

test/functional/feature_includeconf.py
kallewoof: Good point -- I am readding `uacomment=relative\n` to `relative.conf` and checking that subversion ends with `main; relative)`.

src/init.cpp
```diff
@@ -358,7 +358,7 @@ std::string HelpMessage(HelpMessageMode mode)
     if (showDebug) {
         strUsage += HelpMessageOpt("-dbbatchsize", strprintf("Maximum database write batch size in bytes (default: %u)", nDefaultDbBatchSize));
     }
-    strUsage += HelpMessageOpt("-includeconf=<file>", _("Specify additional configuration file, relative to the -datadir path"));
+    strUsage += HelpMessageOpt("-includeconf=<file>", _("Specify additional configuration file, relative to the -datadir path (only useable from configuration file, not command line)"));
```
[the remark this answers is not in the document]
ajtowns: Err, yeah. You add the help message in the -includeconf commit, then change it in the release notes commit.

doc/release-notes.md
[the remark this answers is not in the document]
ajtowns: It's brand new! Just a new file with the section you want to add. There's a couple there already (12823, 12892) to use as examples.

src/init.cpp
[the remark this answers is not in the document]
kallewoof: OH. Fixed! Thanks...

src/util.cpp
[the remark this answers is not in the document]
kallewoof: Warning added -- I used `fprintf(stderr, ..)` though, cause it hasn't processed `-printtoconsole` or anything yet.

kallewoof:
@ajtowns Thanks for review! I believe I addressed your concerns, except for the discussion ^.

src/util.cpp
jnewbery: I think that these are good observations, but that we don't necessarily need to police such pathological edge-cases. The special casing for `[regtest] testnet=1` etc in https://github.com/bitcoin/bitcoin/pull/11862/commits/005ad266491f43d7a9bfd959396037416cb32a55 was useful because it's easy enough to have a config file like:

```
option1=0
option2=1
[regtest]
option3=1
```

and then append `testnet=1` to the end, without realising that you're actually adding `regtest.testnet=1`. (in fact, this PR does something similar here: https://github.com/bitcoin/bitcoin/pull/10267/files#diff-a3677de1e28a0e9b37d66c11cc9a760bR31 by appending an argument that becomes a network-specific argument). This kind of config error is easy to make, ambiguous, and ought to be policed.

The config that you're describing here:

> if you said includeconf=force-testnet.conf [testnet] includeconf=testnet-params.conf, with force-testnet.conf setting testnet=1, you would not end up including testnet-params.conf, whereas if you set -testnet on the command line, you would.

seems much more contorted. I don't think a user should reasonably expect that setting the network in an `includeconf` file should change how the network-specific `includeconf`s in the base config file are interpreted. That behaviour also feels a bit too multi-leveled in that we're reading a specific `includeconf` file based on the content of a different `includeconf` file (I argued against such behaviour here: https://github.com/bitcoin/bitcoin/pull/10267#issuecomment-301810263 because I think it leads to an overly complex model).

test/functional/feature_includeconf.py
```diff
@@ -0,0 +1,72 @@
...
+        with open(os.path.join(self.options.tmpdir + "/node0", "relative.conf"), "w", encoding="utf8") as f:
```
jnewbery: May be clearer to append to the config file, rather than rewrite the `uacomment` param:

```python
        with open(os.path.join(self.options.tmpdir + "/node0", "relative.conf"), "a", encoding="utf8") as f:
            f.write("includeconf=relative2.conf\n")
```

although you may disagree, in which case it's fine to leave as is!

You should certainly update the log above though:

```
self.log.info("-includeconf cannot be used recursively. subversion should end with 'main; relative)/'")
```

src/util.cpp
jnewbery: @kallewoof 

>  I added braces to throw includeconf_net out of scope

I don't know of any other places in the codebase that add blocks to throw local variables out of scope. Obviously functionally the same, but I'd personally have a preference to not add code blocks like this, since it makes the code slightly less clear.

jnewbery:
ACK f083a0e57ca7c991af56425e44cb5aeeaa383a50 with a couple of comments.

I agree with @kallewoof that the include precedence in the current implementation is fine. Perhaps more detailed documentation would satisfy @ajtowns ?

ajtowns:
I'd be happier if there weren't any weird edge cases, but I'm okay with not fixing them. I'll have another look over the latest commits tomorrow, but expect I'd beh appy to ack them as is.

src/util.cpp
```diff
@@ -682,6 +682,15 @@ void ArgsManager::ParseParameters(int argc, const char* const argv[])
             m_override_args[key].push_back(val);
         }
     }
+
+    // we do not allow -includeconf from command line, so we clear it here
+    auto it = m_override_args.find("-includeconf");
```
promag: This should be a loop, otherwise only the first `-includeconf` is removed from `m_override_args`. Also update comment above accordingly.

src/util.cpp
```diff
@@ -943,18 +952,38 @@ void ArgsManager::ReadConfigStream(std::istream& stream)
...
+        }
+
+        if (includeconf.size() > 0) {
+            for (const std::string& to_include : includeconf) {
+                LogPrintf("Attempting to include configuration file %s\n", to_include.c_str());
+                fs::ifstream include_config(GetConfigFile(to_include));
+                if (include_config.good()) {
+                    ReadConfigStream(include_config);
+                } else {
+                    LogPrintf("Failed to include configuration file %s\n", to_include.c_str());
```
promag: Maybe `Skipping invalid configuration file ...`?

src/util.cpp
```diff
@@ -943,18 +952,38 @@ void ArgsManager::ReadConfigStream(std::istream& stream)
...
+        {
+            // We haven't set m_network yet (that happens in SelectParams()), so manually check
+            // for network.includeconf args.
+            std::vector<std::string> includeconf_net(GetArgs(std::string("-") + GetChainName() + ".includeconf"));
+            includeconf.insert(includeconf.end(), includeconf_net.begin(), includeconf_net.end());
+        }
+
+        if (includeconf.size() > 0) {
+            for (const std::string& to_include : includeconf) {
+                LogPrintf("Attempting to include configuration file %s\n", to_include.c_str());
```
promag: Prefer to log the success case since you also log the failure case?

src/init.cpp
```diff
@@ -358,6 +358,7 @@ std::string HelpMessage(HelpMessageMode mode)
     if (showDebug) {
         strUsage += HelpMessageOpt("-dbbatchsize", strprintf("Maximum database write batch size in bytes (default: %u)", nDefaultDbBatchSize));
     }
+    strUsage += HelpMessageOpt("-includeconf=<file>", _("Specify additional configuration file, relative to the -datadir path (only useable from configuration file, not command line)"));
```
promag: Nit, keep sorted?

doc/release-notes-pr10267.md
```diff
@@ -0,0 +1,6 @@
+Changed command-line options
+-----------------------------
```
promag: Nit, remove extra `-`.

src/util.cpp
promag: Agree with @jnewbery, there is no reason to do this. At first sight I thought there was a lock somewhere.

src/util.cpp
promag: Right. Either log all or nothing?

src/util.cpp
```diff
@@ -943,18 +952,38 @@ void ArgsManager::ReadConfigStream(std::istream& stream)
...
         ReadConfigStream(stream);
+        std::vector<std::string> includeconf(GetArgs("-includeconf"));
+        {
+            // We haven't set m_network yet (that happens in SelectParams()), so manually check
+            // for network.includeconf args.
+            std::vector<std::string> includeconf_net(GetArgs(std::string("-") + GetChainName() + ".includeconf"));
+            includeconf.insert(includeconf.end(), includeconf_net.begin(), includeconf_net.end());
+        }
+
+        if (includeconf.size() > 0) {
```
promag: Remove this condition. If it's empty then there is nothing to loop below.

src/util.cpp
kallewoof: Makes sense. Added loop over vector.

src/util.cpp
kallewoof: If someone does `-noincludeconf`, I believe it will exist, but the vector will be empty. See line 682, which adds the key to the map, but clears it (the vector, not the map).

kallewoof: Sorry, line 950, not 682. 682 is for `m_override_args`.

kallewoof: Actually, nevermind, you're right. The for loop will just skip.

kallewoof:
After testing `-noincludeconf`, I did some tweaks so that you *can* do `-noincludeconf` from command line to disable `includeconf` stuff inside config file(s).

src/util.cpp
```diff
@@ -943,18 +954,38 @@ void ArgsManager::ReadConfigStream(std::istream& stream)
...
+                // for network.includeconf args.
+                std::vector<std::string> includeconf_net(GetArgs(std::string("-") + GetChainName() + ".includeconf"));
+                includeconf.insert(includeconf.end(), includeconf_net.begin(), includeconf_net.end());
+            }
+
+            for (const std::string& to_include : includeconf) {
+                fs::ifstream include_config(GetConfigFile(to_include));
+                if (include_config.good()) {
+                    ReadConfigStream(include_config);
+                    LogPrintf("Included configuration file %s\n", to_include.c_str());
```
promag: `printf`?

src/util.cpp
```diff
@@ -943,18 +954,38 @@ void ArgsManager::ReadConfigStream(std::istream& stream)
...
+                } else {
+                    LogPrintf("Failed to include configuration file %s\n", to_include.c_str());
```
promag: `fprintf(stderr`?

promag (commented):
Release notes should mention the argument precedence. For instance, if `printtoconsole=0` is specified in the main conf, then `printtoconsole=1` in included conf are ignored. But arguments with multiple values, like `-wallet, are concatenated.

kallewoof:
Addressed @promag nits.

src/util.cpp
```diff
@@ -943,18 +954,38 @@ void ArgsManager::ReadConfigStream(std::istream& stream)
...
+                // for network.includeconf args.
+                std::vector<std::string> includeconf_net(GetArgs(std::string("-") + GetChainName() + ".includeconf"));
+                includeconf.insert(includeconf.end(), includeconf_net.begin(), includeconf_net.end());
+            }
+
+            for (const std::string& to_include : includeconf) {
+                fs::ifstream include_config(GetConfigFile(to_include));
+                if (include_config.good()) {
+                    ReadConfigStream(include_config);
+                    printf("Included configuration file %s\n", to_include.c_str());
```
jnewbery: I don't understand why this is a print statement. It'll get printed every time bitcoind or bitcoin-qt is started, and each time bitcoin-cli is called.

I think just a normal log here is fine.

src/util.cpp
```diff
@@ -943,18 +954,38 @@ void ArgsManager::ReadConfigStream(std::istream& stream)
...
         m_config_args.clear();
     }
 
+    const std::string confPath = GetArg("-conf", BITCOIN_CONF_FILENAME);
     fs::ifstream stream(GetConfigFile(confPath));
 
     // ok to not have a config file
     if (stream.good()) {
         ReadConfigStream(stream);
+        if (m_override_args.count("-includeconf") == 0) {
```
jnewbery: Can you add a comment to explain why you're checking that count == 0?

src/util.cpp
```diff
@@ -682,6 +682,17 @@ void ArgsManager::ParseParameters(int argc, const char* const argv[])
             m_override_args[key].push_back(val);
         }
     }
+
+    // we do not allow -includeconf from command line, so we clear it here
+    auto it = m_override_args.find("-includeconf");
+    if (it != m_override_args.end()) {
+        if (it->second.size() > 0) {
```
jnewbery: It'd be nice if we could be more aggressive here and fail to start if a `-includeconf` argument is provided on the command line (since continuing to startup would be continuing with config that the user has specified that they don't want).

`ParseParameters()` doesn't currently have a way to return errors, and callers don't wrap calls with try-except (unlike `GetChainName()` for example). Perhaps a future PR could change the function to return a bool, so we could indicate an error to the caller and abort startup.

test/functional/feature_includeconf.py
```diff
@@ -0,0 +1,72 @@
...
+import os
+
+from test_framework.test_framework import BitcoinTestFramework
+
+class IncludeConfTest(BitcoinTestFramework):
+    def setup_chain(self):
+        super().setup_chain()
+        # Create additional config files
+        # - tmpdir/node0/relative.conf
+        with open(os.path.join(self.options.tmpdir + "/node0", "relative.conf"), "w", encoding="utf8") as f:
```
jnewbery: nit: prefer `os.path.join(self.options.tmpdir, "node0", "relative.conf")` for better portability.

jnewbery (commented):
Looks great. Just a few more nits!

Thanks for the release notes

src/util.cpp
kallewoof: Makes sense. I'm leaving it for now as it sounds like a future thing, I think.

src/util.cpp
kallewoof: Hm, you're right that it prints every time you call bitcoin-cli etc. That's ugly. I'll think of something.

kallewoof: I'm removing this entirely. As LogPrintf it is NOP and as printf it is ugly. I think we can live without being told the include was successful. The error case will be spammy, but I think that's okay.

src/util.cpp
```diff
@@ -445,6 +445,17 @@ void ArgsManager::ParseParameters(int argc, const char* const argv[])
...
+            for (const auto& ic : it->second) {
+                printf("warning: -includeconf cannot be used from commandline; ignoring -includeconf=%s\n", ic.c_str());
```
laanwj: `fprintf(stderr, ...`?

kallewoof: Makes sense. Done.

kallewoof: Actually, I remember now. The reason was because travis fails for certain tests when using fprintf.

laanwj: Should probably fix the test instead in that case :-)

kallewoof: I thought it was a travis internal thing to automatically assume error when stderr was printed to. I can't find any reference to this though... digging.

kallewoof: It's actually `test_runner.py`, not travis. Digging more.

laanwj:
utACK 8eebf2a55f43bbcf83b64051f5d1a6355b0d0145

laanwj:
@kallewoof Feel free to take over https://github.com/laanwj/bitcoin/commit/57b1b57cba84ea60e4de7cc0549aa448132e7573  to fix the stderr issue (and check stderr output)

src/util.cpp
ajtowns: LogPrintf gets stored and replayed when going to a logfile; we could fix it so the same thing happens with printtoconsole?

kallewoof: @ajtowns Really? I thought I checked the log file and didn't see it. If it is replayed, I should add it back.

kallewoof:
@laanwj Ohh, nice. Thanks for the fix. I squashed it into 360a4d8.

src/util.cpp
kallewoof: I restored LogPrintf line.

ajtowns: #13088 should add support replay for console logs as well as debug.log fwiw

jnewbery:
I'd prefer to not to have special stderr checking for this test. I have a PR open here: #12755 which adds stderr checking to the test framework.

I've also prepared a branch here: https://github.com/bitcoin/bitcoin/compare/master...jnewbery:pr10267.3?expand=1 which rebases this PR on #12755 and adds explicit stderr checking.

kallewoof:
@jnewbery Wouldn't it be simpler to fix this in a follow-up PR once #12755 is merged?

test/functional/feature_includeconf.py
```diff
@@ -0,0 +1,78 @@
...
+2. includeconf cannot be used as a command line argument.
+3. includeconf cannot be used recursively (ie includeconf can only
+   be used from the base config file).
+4. multiple includeconf arguments can be specified in the main config
+   file.
+"""
+import os
+
+from test_framework.test_framework import BitcoinTestFramework, assert_equal
+import tempfile
```
jnewbery: nit: PEP8 import ordering please (std library first, then local application imports)

test/functional/feature_includeconf.py
```diff
@@ -0,0 +1,78 @@
...
+
+class IncludeConfTest(BitcoinTestFramework):
+    def setup_chain(self):
```
jnewbery: micronit: all other tests have `set_test_params()` as the first method. I think that makes more logical sense since that method is called by `__init__()`. Do you mind doing the same here?

jnewbery (commented):
Tested ACK 6d5605517b477538f0447d41fe0bc734da27d10c. A couple of nits in the new test code, but nothing that should prevent merge.

> @jnewbery Wouldn't it be simpler to fix this in a follow-up PR once #12755 is merged?

Not 100% sure what you mean by 'fix this', but I'm guessing you mean merge this as is and then remove the specific stderr checking code from this test when the generic stderr checking is added in #12755? Sure - that's fine, as long as you don't mind reviewing #12755 for me :)

kallewoof:
@jnewbery Thanks for review! Addressed your nits.

kallewoof:
@jnewbery I'm a little confused. Should I close this PR in favor of #12755? I thought #12755 only added the error stuff, but it includes this PR as well now.

jnewbery:
Tested ACK 25b7ab9c02f691be6c1aa71a9cf51ac1a6ea9db4.

>  I'm a little confused. Should I close this PR in favor of #12755? I thought #12755 only added the error stuff, but it includes this PR as well now.

Sorry - I haven't been very clear in my updates. Now that this PR includes stderr testing, it conflicts with #12755. My initial suggestions were to remove the stderr testing from this test, or rebase on top of #12755. But now I think it would make more sense to rebase #12755 on this, since this PR has been around for so long and should be close to ready for merge.

I've rebased #12755 on this. As soon as this PR is merged, #12755 will be ready to go in.

jnewbery:
I think this is very nearly ready for merge. Are previous reviewers @laanwj, @ajtowns @promag, @ryanofsky, @MarcoFalke @MeshCollider, @jonasschnelli able to re-review this?

laanwj:
utACK 25b7ab9c02f691be6c1aa71a9cf51ac1a6ea9db4

practicalswift:
When working on the locking annotations (see #13126) I noticed that `ArgsManager::ReadConfigFiles()` introduced in this PR accesses `m_override_args` without first locking `cs_args`.

The PR #13126 adds the correct locking. FWIW Travis will catch this type of `cs_args` locking violations when #13126 is merged :-)

test/functional/feature_includeconf.py
```diff
@@ -0,0 +1,78 @@
...
+        self.log.info("-includeconf cannot be used as command-line arg. subversion should still end with 'main; relative)/'")
+        self.stop_node(0)
+        with tempfile.SpooledTemporaryFile(max_size=2**16) as log_stderr:
+            self.start_node(0, extra_args=["-includeconf=relative2.conf"], stderr=log_stderr)
+
+            subversion = self.nodes[0].getnetworkinfo()["subversion"]
+            assert subversion.endswith("main; relative)/")
+            log_stderr.seek(0)
+            stderr = log_stderr.read().decode('utf-8').strip()
+            assert_equal(stderr, 'warning: -includeconf cannot be used from commandline; ignoring -includeconf=relative2.conf')
```
MarcoFalke: nit: should probably be an (init)error, since it is clearly an invalid command line

src/util.cpp
```diff
@@ -706,18 +717,40 @@ void ArgsManager::ReadConfigStream(std::istream& stream)
...
+                includeconf.insert(includeconf.end(), includeconf_net.begin(), includeconf_net.end());
+            }
+
+            for (const std::string& to_include : includeconf) {
+                fs::ifstream include_config(GetConfigFile(to_include));
+                if (include_config.good()) {
+                    ReadConfigStream(include_config);
+                    LogPrintf("Included configuration file %s\n", to_include.c_str());
+                } else {
+                    fprintf(stderr, "Failed to include configuration file %s\n", to_include.c_str());
```
MarcoFalke: nit: return false on error?

maflcko (commented):
Post-merge ACK 25b7ab9 

See #13367 to fix my nit about the errors.

PR #10270, opened 2017-04-25
Remove Clang workaround for Boost 1.46

fanquake:
This was introduced in #1104. 

Now that we require at least Boost 1.47, we can remove it.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -88,19 +88,6 @@
 #include <openssl/rand.h>
 #include <openssl/conf.h>
 
-// Work around clang compilation problem in Boost 1.46:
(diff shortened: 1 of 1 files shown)


laanwj:
Yep: https://github.com/bitcoin/bitcoin/blob/master/configure.ac#L670

utACK

kallewoof:
ACK 93dbb15954679c1b2eb2ad59e601c5523272e810

macOS 10.12.4 on a MBP, Apple LLVM version 8.1.0 (clang-802.0.38)

Compiles. (The fix was a compiler issue, so no further testing seems necessary.)

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/10270/commits/93dbb15954679c1b2eb2ad59e601c5523272e810

PR #10273, opened 2017-04-25
[scripts] Minor improvements to `macdeployqtplus` script.

chrisgavin:


contrib/macdeploy/macdeployqtplus
```diff
@@ -302,7 +302,6 @@ def copyFramework(framework, path, verbose):
         if os.path.exists(fromContentsDir):
             toContentsDir = os.path.join(path, framework.destinationVersionContentsDirectory)
             shutil.copytree(fromContentsDir, toContentsDir, symlinks=True)
-            contentslinkfrom = os.path.join(path, framework.destinationContentsDirectory)
```
jonasschnelli: What's the reason to remove this line? We don't use dynamic linked custom Frameworks in the Bitcoin-Qt.app bundle.

chrisgavin: In af0bd5e, the majority of this code was removed but I think this line was left in by mistake. As far as I can tell it doesn't do anything - it just defines a variable that is never used.

jonasschnelli:
Yes. Makes sense (though a very tiny optimisation/cleanup near PR "dust-limit").
utACK e8babc4330cd7db82650f28147c388846a389c77

jonasschnelli:
Builds the DMG correctly (https://bitcoin.jonasschnelli.ch/build/113).
Tested ACK e8babc4330cd7db82650f28147c388846a389c77

PR #10275, opened 2017-04-25
[rpc] Allow fetching tx directly from specified block in getrawtransaction

kallewoof:
[Reviewer hint: use [?w=1](https://github.com/bitcoin/bitcoin/pull/10275/files?w=1) to avoid seeing a bunch of indentation changes.]

Presuming a user knows the block hash of the block containing a given transaction, this PR allows them to fetch the raw transaction, even without `-txindex`. It also enables support for getting transactions that are in orphaned blocks.

Note that supplying a block hash will override mempool and txindex support in `GetTransaction`. The rationale behind this is that a transaction may be in multiple places (orphaned blocks) and if the user supplies an explicit block hash it should be adhered to.

```Bash
$ # a41.. is a tx inside an orphan block ..3c6f.. -- first try getting it normally
$ ./bitcoin-cli getrawtransaction a41e66ee1341aa9fb9475b98cfdc1fe1261faa56c0a49254f33065ec90f7cd79 1
error code: -5
error message:
No such mempool transaction. Use -txindex to enable blockchain transaction queries. Use gettransaction for wallet transactions.
$ # now try with block hash
$ ./bitcoin-cli getrawtransaction a41e66ee1341aa9fb9475b98cfdc1fe1261faa56c0a49254f33065ec90f7cd79 1 0000000000000000003c6fe479122bfa4a9187493937af1734e1e5cd9f198ec7
{
  "hex": "01000000014e7e81144e42f6d65550e59b715d470c9301fd7ac189[...]90488ac00000000",
  "inMainChain": false,
  "txid": "a41e66ee1341aa9fb9475b98cfdc1fe1261faa56c0a49254f33065ec90f7cd79",
  "hash": "a41e66ee1341aa9fb9475b98cfdc1fe1261faa56c0a49254f33065ec90f7cd79",
  "size": 225,
[...]
}
$ # another tx 6c66... in block 462000
$ ./bitcoin-cli getrawtransaction 6c66b98191e9d6cc671f6817142152ebf6c5cab2ef008397b5a71ac13255a735 1 00000000000000000217f2c12922e321f6d4aa933ce88005a9a493c503054a40
{
  "hex": "0200000004d157[...]88acaf0c0700",
  "inMainChain": true,
  "txid": "6c66b98191e9d6cc671f6817142152ebf6c5cab2ef008397b5a71ac13255a735",
  "hash": "6c66b98191e9d6cc671f6817142152ebf6c5cab2ef008397b5a71ac13255a735",
  "size": 666,
[...]
}
$ 
```

### DIFF
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -141,19 +141,7 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
     // Accept either a bool (true) or a num (>=1) to indicate verbose output.
     bool fVerbose = false;
     if (!request.params[1].isNull()) {
-        if (request.params[1].isNum()) {
-            if (request.params[1].get_int() != 0) {
-                fVerbose = true;
-            }
-        }
-        else if(request.params[1].isBool()) {
-            if(request.params[1].isTrue()) {
-                fVerbose = true;
-            }
-        }
-        else {
-            throw JSONRPCError(RPC_TYPE_ERROR, "Invalid type provided. Verbose parameter must be a boolean.");
-        }
+        fVerbose = request.params[1].isNum() ? (request.params[1].get_int() != 0) : request.params[1].get_bool();
     }
 
     CTransactionRef tx;
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -64,12 +64,15 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
 
 UniValue getrawtransaction(const JSONRPCRequest& request)
 {
-    if (request.fHelp || request.params.size() < 1 || request.params.size() > 2)
+    if (request.fHelp || request.params.size() < 1 || request.params.size() > 3)
         throw std::runtime_error(
-            "getrawtransaction \"txid\" ( verbose )\n"
+            "getrawtransaction \"txid\" ( verbose \"blockhash\" )\n"
 
             "\nNOTE: By default this function only works for mempool transactions. If the -txindex option is\n"
-            "enabled, it also works for blockchain transactions.\n"
+            "enabled, it also works for blockchain transactions. If the block which contains the transaction\n"
+            "is known, its hash can be provided even for nodes without -txindex. Note that if a blockhash is\n"
+            "provided, only that block will be searched and if the transaction is in the mempool or other\n"
+            "blocks, or if this node does not have the given block available, the transaction will not be found.\n"
             "DEPRECATED: for now, it also works for transactions with unspent outputs.\n"
 
             "\nReturn the raw transaction data.\n"
@@ -78,13 +81,15 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
 
             "\nArguments:\n"
             "1. \"txid\"      (string, required) The transaction id\n"
-            "2. verbose       (bool, optional, default=false) If false, return a string, otherwise return a json object\n"
+            "2. verbose     (bool, optional, default=false) If false, return a string, otherwise return a json object\n"
+            "3. \"blockhash\" (string, optional) The block in which to look for the transaction\n"
 
             "\nResult (if verbose is not set or set to false):\n"
             "\"data\"      (string) The serialized, hex-encoded data for 'txid'\n"
 
             "\nResult (if verbose is set to true):\n"
             "{\n"
+            "  \"in_active_chain\": b, (bool) Whether specified block is in the active chain or not (only present with explicit \"blockhash\" argument)\n"
             "  \"hex\" : \"data\",       (string) The serialized, hex-encoded data for 'txid'\n"
             "  \"txid\" : \"id\",        (string) The transaction id (same as provided)\n"
             "  \"hash\" : \"id\",        (string) The transaction hash (differs from txid for witness transactions)\n"
@@ -132,11 +137,15 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
             + HelpExampleCli("getrawtransaction", "\"mytxid\"")
             + HelpExampleCli("getrawtransaction", "\"mytxid\" true")
             + HelpExampleRpc("getrawtransaction", "\"mytxid\", true")
+            + HelpExampleCli("getrawtransaction", "\"mytxid\" false \"myblockhash\"")
+            + HelpExampleCli("getrawtransaction", "\"mytxid\" true \"myblockhash\"")
         );
 
     LOCK(cs_main);
 
+    bool in_active_chain = true;
     uint256 hash = ParseHashV(request.params[0], "parameter 1");
+    CBlockIndex* blockindex = nullptr;
 
     // Accept either a bool (true) or a num (>=1) to indicate verbose output.
     bool fVerbose = false;
@@ -144,18 +153,42 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
         fVerbose = request.params[1].isNum() ? (request.params[1].get_int() != 0) : request.params[1].get_bool();
     }
 
-    CTransactionRef tx;
-    uint256 hashBlock;
-    if (!GetTransaction(hash, tx, Params().GetConsensus(), hashBlock, true))
-        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string(fTxIndex ? "No such mempool or blockchain transaction"
-            : "No such mempool transaction. Use -txindex to enable blockchain transaction queries") +
-            ". Use gettransaction for wallet transactions.");
+    if (!request.params[2].isNull()) {
+        uint256 blockhash = ParseHashV(request.params[2], "parameter 3");
+        if (!blockhash.IsNull()) {
+            BlockMap::iterator it = mapBlockIndex.find(blockhash);
+            if (it == mapBlockIndex.end()) {
+                throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Block hash not found");
+            }
+            blockindex = it->second;
+            in_active_chain = chainActive.Contains(blockindex);
+        }
+    }
 
-    if (!fVerbose)
+    CTransactionRef tx;
+    uint256 hash_block;
+    if (!GetTransaction(hash, tx, Params().GetConsensus(), hash_block, true, blockindex)) {
+        std::string errmsg;
+        if (blockindex) {
+            if (!(blockindex->nStatus & BLOCK_HAVE_DATA)) {
+                throw JSONRPCError(RPC_MISC_ERROR, "Block not available");
+            }
+            errmsg = "No such transaction found in the provided block";
+        } else {
+            errmsg = fTxIndex
+              ? "No such mempool or blockchain transaction"
+              : "No such mempool transaction. Use -txindex to enable blockchain transaction queries";
+        }
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, errmsg + ". Use gettransaction for wallet transactions.");
+    }
+
+    if (!fVerbose) {
         return EncodeHexTx(*tx, RPCSerializationFlags());
+    }
 
     UniValue result(UniValue::VOBJ);
-    TxToJSON(*tx, hashBlock, result);
+    if (blockindex) result.push_back(Pair("in_active_chain", in_active_chain));
+    TxToJSON(*tx, hash_block, result);
     return result;
 }
 
@@ -983,7 +1016,7 @@ UniValue sendrawtransaction(const JSONRPCRequest& request)
 static const CRPCCommand commands[] =
 { //  category              name                      actor (function)         argNames
   //  --------------------- ------------------------  -----------------------  ----------
-    { "rawtransactions",    "getrawtransaction",      &getrawtransaction,      {"txid","verbose"} },
+    { "rawtransactions",    "getrawtransaction",      &getrawtransaction,      {"txid","verbose","blockhash"} },
     { "rawtransactions",    "createrawtransaction",   &createrawtransaction,   {"inputs","outputs","locktime","replaceable"} },
     { "rawtransactions",    "decoderawtransaction",   &decoderawtransaction,   {"hexstring"} },
     { "rawtransactions",    "decodescript",           &decodescript,           {"hexstring"} },
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -926,47 +926,51 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
     return AcceptToMemoryPoolWithTime(chainparams, pool, state, tx, pfMissingInputs, GetTime(), plTxnReplaced, bypass_limits, nAbsurdFee);
 }
 
-/** Return transaction in txOut, and if it was found inside a block, its hash is placed in hashBlock */
-bool GetTransaction(const uint256 &hash, CTransactionRef &txOut, const Consensus::Params& consensusParams, uint256 &hashBlock, bool fAllowSlow)
+/**
+ * Return transaction in txOut, and if it was found inside a block, its hash is placed in hashBlock.
+ * If blockIndex is provided, the transaction is fetched from the corresponding block.
+ */
+bool GetTransaction(const uint256& hash, CTransactionRef& txOut, const Consensus::Params& consensusParams, uint256& hashBlock, bool fAllowSlow, CBlockIndex* blockIndex)
 {
-    CBlockIndex *pindexSlow = nullptr;
+    CBlockIndex* pindexSlow = blockIndex;
 
     LOCK(cs_main);
 
-    CTransactionRef ptx = mempool.get(hash);
-    if (ptx)
-    {
-        txOut = ptx;
-        return true;
-    }
-
-    if (fTxIndex) {
-        CDiskTxPos postx;
-        if (pblocktree->ReadTxIndex(hash, postx)) {
-            CAutoFile file(OpenBlockFile(postx, true), SER_DISK, CLIENT_VERSION);
-            if (file.IsNull())
-                return error("%s: OpenBlockFile failed", __func__);
-            CBlockHeader header;
-            try {
-                file >> header;
-                fseek(file.Get(), postx.nTxOffset, SEEK_CUR);
-                file >> txOut;
-            } catch (const std::exception& e) {
-                return error("%s: Deserialize or I/O error - %s", __func__, e.what());
-            }
-            hashBlock = header.GetHash();
-            if (txOut->GetHash() != hash)
-                return error("%s: txid mismatch", __func__);
+    if (!blockIndex) {
+        CTransactionRef ptx = mempool.get(hash);
+        if (ptx) {
+            txOut = ptx;
             return true;
         }
 
-        // transaction not found in index, nothing more can be done
-        return false;
-    }
+        if (fTxIndex) {
+            CDiskTxPos postx;
+            if (pblocktree->ReadTxIndex(hash, postx)) {
+                CAutoFile file(OpenBlockFile(postx, true), SER_DISK, CLIENT_VERSION);
+                if (file.IsNull())
+                    return error("%s: OpenBlockFile failed", __func__);
+                CBlockHeader header;
+                try {
+                    file >> header;
+                    fseek(file.Get(), postx.nTxOffset, SEEK_CUR);
+                    file >> txOut;
+                } catch (const std::exception& e) {
+                    return error("%s: Deserialize or I/O error - %s", __func__, e.what());
+                }
+                hashBlock = header.GetHash();
+                if (txOut->GetHash() != hash)
+                    return error("%s: txid mismatch", __func__);
+                return true;
+            }
 
-    if (fAllowSlow) { // use coin database to locate block that contains transaction, and scan it
-        const Coin& coin = AccessByTxid(*pcoinsTip, hash);
-        if (!coin.IsSpent()) pindexSlow = chainActive[coin.nHeight];
+            // transaction not found in index, nothing more can be done
+            return false;
+        }
+
+        if (fAllowSlow) { // use coin database to locate block that contains transaction, and scan it
+            const Coin& coin = AccessByTxid(*pcoinsTip, hash);
+            if (!coin.IsSpent()) pindexSlow = chainActive[coin.nHeight];
+        }
     }
 
     if (pindexSlow) {
--- a/test/functional/rawtransactions.py
+++ b/test/functional/rawtransactions.py
@@ -50,6 +50,30 @@ class RawTransactionsTest(BitcoinTestFramework):
         # This will raise an exception since there are missing inputs
         assert_raises_rpc_error(-25, "Missing inputs", self.nodes[2].sendrawtransaction, rawtx['hex'])
 
+        #####################################
+        # getrawtransaction with block hash #
+        #####################################
+
+        # make a tx by sending then generate 2 blocks; block1 has the tx in it
+        tx = self.nodes[2].sendtoaddress(self.nodes[1].getnewaddress(), 1)
+        block1, block2 = self.nodes[2].generate(2)
+        self.sync_all()
+        # We should be able to get the raw transaction by providing the correct block
+        gottx = self.nodes[0].getrawtransaction(tx, True, block1)
+        assert_equal(gottx['txid'], tx)
+        assert_equal(gottx['in_active_chain'], True)
+        # We should not have the 'in_active_chain' flag when we don't provide a block
+        gottx = self.nodes[0].getrawtransaction(tx, True)
+        assert_equal(gottx['txid'], tx)
+        assert 'in_active_chain' not in gottx
+        # We should not get the tx if we provide an unrelated block
+        assert_raises_rpc_error(-5, "No such transaction found", self.nodes[0].getrawtransaction, tx, True, block2)
+        # An invalid block hash should raise the correct errors
+        assert_raises_rpc_error(-8, "parameter 3 must be hexadecimal", self.nodes[0].getrawtransaction, tx, True, True)
+        assert_raises_rpc_error(-8, "parameter 3 must be hexadecimal", self.nodes[0].getrawtransaction, tx, True, "foobar")
+        assert_raises_rpc_error(-8, "parameter 3 must be of length 64", self.nodes[0].getrawtransaction, tx, True, "abcd1234")
+        assert_raises_rpc_error(-5, "Block hash not found", self.nodes[0].getrawtransaction, tx, True, "aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa")
+
         #########################
         # RAW TX MULTISIG TESTS #
         #########################
--- a/test/functional/rawtransactions.py
+++ b/test/functional/rawtransactions.py
@@ -188,13 +188,13 @@ class RawTransactionsTest(BitcoinTestFramework):
         assert_equal(self.nodes[0].getrawtransaction(txHash, True)["hex"], rawTxSigned['hex'])
 
         # 6. invalid parameters - supply txid and string "Flase"
-        assert_raises_rpc_error(-3,"Invalid type", self.nodes[0].getrawtransaction, txHash, "Flase")
+        assert_raises_rpc_error(-1,"not a boolean", self.nodes[0].getrawtransaction, txHash, "Flase")
 
         # 7. invalid parameters - supply txid and empty array
-        assert_raises_rpc_error(-3,"Invalid type", self.nodes[0].getrawtransaction, txHash, [])
+        assert_raises_rpc_error(-1,"not a boolean", self.nodes[0].getrawtransaction, txHash, [])
 
         # 8. invalid parameters - supply txid and empty dict
-        assert_raises_rpc_error(-3,"Invalid type", self.nodes[0].getrawtransaction, txHash, {})
+        assert_raises_rpc_error(-1,"not a boolean", self.nodes[0].getrawtransaction, txHash, {})
 
         inputs  = [ {'txid' : "1d1d4e24ed99057e84c3f80fd8fbec79ed9e1acee37da269356ecea000000000", 'vout' : 1, 'sequence' : 1000}]
         outputs = { self.nodes[0].getnewaddress() : 1 }
--- a/src/validation.h
+++ b/src/validation.h
@@ -273,7 +273,7 @@ void ThreadScriptCheck();
 /** Check whether we are doing an initial block download (synchronizing from disk or network) */
 bool IsInitialBlockDownload();
 /** Retrieve a transaction (from memory pool, or from disk, if possible) */
-bool GetTransaction(const uint256 &hash, CTransactionRef &tx, const Consensus::Params& params, uint256 &hashBlock, bool fAllowSlow = false);
+bool GetTransaction(const uint256& hash, CTransactionRef& tx, const Consensus::Params& params, uint256& hashBlock, bool fAllowSlow = false, CBlockIndex* blockIndex = nullptr);
 /** Find the best known block, and make it the tip of the block chain */
 bool ActivateBestChain(CValidationState& state, const CChainParams& chainparams, std::shared_ptr<const CBlock> pblock = std::shared_ptr<const CBlock>());
 CAmount GetBlockSubsidy(int nHeight, const Consensus::Params& consensusParams);


src/rpc/rawtransaction.cpp
```diff
@@ -192,11 +195,34 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
+            + HelpExampleCli("getrawtransaction", "\"myblockhash:mytxid\", true")
         );
 
     LOCK(cs_main);
 
-    uint256 hash = ParseHashV(request.params[0], "parameter 1");
+    uint256 hash;
+    int64_t blockHeight = -1;
+    bool inMainChain = true;
+    auto p = request.params[0].get_str().find(':');
```
laanwj: This functionality could come in useful.

As for the API I prefer to not do any string combining/parsing here, this makes the API less clean to work with at least in my experience. I'd prefer to add an optional (can be null) `fromBlock` argument.

gmaxwell (commented):
Concept ACK. I've really wanted this before.

Allowing it work on orphan blocks is an interesting idea. I'm a little less sure about that-- I think it could allow it to return transactions that have never been validated, which would be somewhat surprising.

jonasschnelli:
Nice feature!
I agree with @laanwj about the string parsing. The `<block>:<txid>` schematics looks good at first sight, but we are using JSON and should stay with JSON and don't add another form or key/value encoding.

kallewoof:
Fair enough, I'll add a blockhash argument instead. I was kind of toying with the idea of a new standard for referencing transactions which included the block height (not hash) so everyone could always find a tx presuming they had the block in question, and that thought sort of seeped in here.

kallewoof:
@gmaxwell 
> Allowing it work on orphan blocks is an interesting idea. I'm a little less sure about that-- I think it could allow it to return transactions that have never been validated, which would be somewhat surprising.

I just realized my reasoning was flawed on this. I am passing the block height only to the GetTransaction method, which means it will always pick the active chain at the given height. Either I throw when the block is not in the main chain (i.e. no support for orphaned blocks) or I move the height determine logic over to validation.cpp. I am leaning towards the latter, but feedback welcome.

kallewoof:
The code now works as advertised (see updated OP).

[History](https://github.com/kallewoof/bitcoin/commits/gettx-with-blockhash-unsquashed):
* 2017-04-25: [1 (block hash support in getrawtx)](https://github.com/kallewoof/bitcoin/commit/221e725814b9bba82c3f0f37ba3c4efea06038ea)
* 2017-04-26: → [2⊱1 (bh:txid to 2 params)](https://github.com/kallewoof/bitcoin/commit/9bd521c16a2bcdf58551bcca71809bedc2cbb553) → [3 (fix fVerbose parsing)](https://github.com/kallewoof/bitcoin/commit/7d891d534a3d1a5860889ec1f5f048a5b82371bd) → [4⊱1 (switch to block index in GetTx)](https://github.com/kallewoof/bitcoin/commit/eb2a9d237c3adea1aa6c90a1f6133783991b49ff)

src/rpc/rawtransaction.cpp
```diff
@@ -192,34 +195,44 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
 
     // Accept either a bool (true) or a num (>=1) to indicate verbose output.
     bool fVerbose = false;
     if (request.params.size() > 1) {
         if (request.params[1].isNum()) {
-            if (request.params[1].get_int() != 0) {
-                fVerbose = true;
-            }
+            fVerbose = (request.params[1].get_int() != 0);
+        } else if (request.params[1].isBool()) {
```
jnewbery: nit: You can replace this and the next three lines with:

```c++
} else {
    fVerbose = (request.params[1].get_bool());
}
```

since `get_bool()` does the type testing for you and throws the JSONRPCError if the type isn't a VBOOL.

Up to you whether you think that's clearer.

jnewbery:
utACK, but I think this deserves a new functional test case.

src/rpc/rawtransaction.cpp
kallewoof: Ohh, good point! Thanks.

kallewoof: So I end up with
```C++
        if (request.params[1].isNum()) {
            fVerbose = (request.params[1].get_int() != 0);
        } else if (!request.params[1].isNull()) {
            fVerbose = (request.params[1].get_bool());
        }
```

kallewoof:
@jnewbery I agree. Will get to work on that.

[[...]](https://github.com/kallewoof/bitcoin/commits/gettx-with-blockhash-unsquashed):
* 2017-05-04: [5⊱2 (improve fVerbose)](https://github.com/kallewoof/bitcoin/commit/2a921bf5585ac72e2154b41492ab70ecd3e46241)

kallewoof:
@jnewbery Added some tests to `rawtransactions.py` for the included blockhash variant (9f980f6).

[[...]](https://github.com/kallewoof/bitcoin/commits/gettx-with-blockhash-unsquashed):
* 2017-05-08: [6 (tests)](https://github.com/kallewoof/bitcoin/commit/99f115cc5a0d3dc8b93b5107ed2fe933488c85e6)

test/functional/rawtransactions.py
```diff
@@ -64,6 +64,20 @@ def run_test(self):
...
         assert_raises_jsonrpc(-25, "Missing inputs", self.nodes[2].sendrawtransaction, rawtx['hex'])
 
+        ############################
+        # getrawtx with block hash #
+        ############################
+
+        # make a tx by sending then generate 2 blocks; block1 has the tx in it,
+        # presumably
+        tx = self.nodes[2].sendtoaddress(self.nodes[1].getnewaddress(), 1)
+        [ block1, block2 ] = self.nodes[2].generate(2)
```
jnewbery: nit: no need for brackets here. The following should do:

`block1, block2 = self.nodes[2].generate(2)`

test/functional/rawtransactions.py
```diff
@@ -64,6 +64,20 @@ def run_test(self):
...
+        self.sync_all()
+        # We should be able to get the raw transaction by providing the correct block
+        assert self.nodes[0].getrawtransaction(tx, True, block1)
```
jnewbery: nit: I think it's better to assert on the actual value here (ie verify that the getrawtransaction returned the correct transaction rather than returned anything). The following should do that:

`assert_equal(self.nodes[0].getrawtransaction(tx, True, block1)['txid'], tx)`

jnewbery (commented):
tested ACK the integration test in a6b84618ad04dccc628aa4905fced4784a62895a with a couple of nits.

sipa:
Is this still needed after #8704?

kallewoof:
@sipa The use cases are quite different I think. #8704 lets you see all transactions in a given block. This lets you grab a transaction directly from a block without indexing.

sipa:
@kallewoof #8704 does not require indexing either

kallewoof:
Yeah, sorry, I meant that this is a way to get a specific transaction if you know the block hash, whereas #8704 shows you all transactions in the entire block. You get the info, but you have to wade through stuff to get it.

test/functional/rawtransactions.py
kallewoof: @jnewbery Good points, thanks. Addressed!

laanwj:
In both cases the whole block has to be loaded from disk, and parsed, and searched linearly. The difference is whether the linear search step happens on the server or client.

I think both #8704 and this can be useful, but have to agree there's only superficial difference.

maaku:
The difference between the two is the serialization and transmission and parsing of ~5MB of JSON data vs a few hundred bytes of hex encoded data. That's a 1,000x difference on the client side, and the same absolute improvement on the server -- although as a multiplier it'd be less since as you note the server still has to parse the block from disk. That's a nontrivial performance difference.

(Also, this would have greatly helped me in the past so another +1 from me.)

jnewbery:
Agree with @maaku - simply encoding the data into 5MB of json could be time-consuming, during which time the cs_main lock is held. Having a command to return just a single transaction from a block seems very useful.

src/rpc/rawtransaction.cpp
```diff
@@ -124,12 +124,13 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
...
-    if (request.fHelp || request.params.size() < 1 || request.params.size() > 2)
+    if (request.fHelp || request.params.size() < 1 || request.params.size() > 3)
         throw std::runtime_error(
-            "getrawtransaction \"txid\" ( verbose )\n"
+            "getrawtransaction \"txid\" ( verbose \"blockhash\" )\n"
 
             "\nNOTE: By default this function only works for mempool transactions. If the -txindex option is\n"
-            "enabled, it also works for blockchain transactions.\n"
+            "enabled, it also works for blockchain transactions. If the block hash is known, it can be provided\n"
+            "for nodes without -txindex.\n"
```
luke-jr: Should probably mention that it MUST be in that block in this case...

src/rpc/rawtransaction.cpp
```diff
@@ -192,34 +195,42 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
-            if(request.params[1].isTrue()) {
-                fVerbose = true;
+    }
+
+    if (request.params.size() > 2 && !request.params[2].isNull()) {
+        uint256 blockHash = ParseHashV(request.params[2], "parameter 3");
+        if (!blockHash.IsNull()) {
+            BlockMap::iterator it = mapBlockIndex.find(blockHash);
+            if (it == mapBlockIndex.end()) {
+                throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Block hash not found in chain");
```
luke-jr: Remove "in chain". Maybe "in database"?

src/rpc/rawtransaction.cpp
```diff
@@ -139,12 +140,14 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
             "1. \"txid\"      (string, required) The transaction id\n"
             "2. verbose       (bool, optional, default=false) If false, return a string, otherwise return a json object\n"
+            "3. \"blockhash\" (string, optional) The block in which to look for the transaction\n"
 
             "\nResult (if verbose is not set or set to false):\n"
             "\"data\"      (string) The serialized, hex-encoded data for 'txid'\n"
 
             "\nResult (if verbose is set to true):\n"
             "{\n"
+            "  \"inMainChain\": b,       (bool) Whether transaction is in the main chain or not. Only visible when specifying block hash\n"
```
luke-jr: Whether *the block specified* is in the main chain or not... This could be false with the tx being still in the main chain!

luke-jr (changes requested):

luke-jr:
Suggested message fix on my `gettx-with-blockhash-0.14` branch (cherry-pick).

sipa:
Needs rebase.

kallewoof:
@luke-jr Thanks for the review! I cherry-picked your commit. 

[[...]](https://github.com/kallewoof/bitcoin/commits/gettx-with-blockhash-unsquashed):
* 2017-06-05: [7 (nits (luke-jr))](https://github.com/kallewoof/bitcoin/commit/974e4ef131c79aa78b7f7eeb759b1d729823c33e)

@sipa Rebased.

src/rpc/rawtransaction.cpp
```diff
@@ -77,12 +79,14 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
             "1. \"txid\"      (string, required) The transaction id\n"
             "2. verbose       (bool, optional, default=false) If false, return a string, otherwise return a json object\n"
+            "3. \"blockhash\" (string, optional) The block in which to look for the transaction\n"
 
             "\nResult (if verbose is not set or set to false):\n"
             "\"data\"      (string) The serialized, hex-encoded data for 'txid'\n"
 
             "\nResult (if verbose is set to true):\n"
             "{\n"
+            "  \"inMainChain\": b,       (bool) Whether specified block is in the main chain or not\n"
```
TheBlueMatt: Hmm, maybe say "if blockhash is specified" or otherwise mention this wont appear unless a blockhash is provided. Even better, fill it out if GetTransaction returns the blockhash cause it found it via UTXO/txindex.

src/rpc/rawtransaction.cpp
```diff
@@ -62,12 +62,14 @@ void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry)
...
-    if (request.fHelp || request.params.size() < 1 || request.params.size() > 2)
+    if (request.fHelp || request.params.size() < 1 || request.params.size() > 3)
         throw std::runtime_error(
-            "getrawtransaction \"txid\" ( verbose )\n"
+            "getrawtransaction \"txid\" ( verbose \"blockhash\" )\n"
 
             "\nNOTE: By default this function only works for mempool transactions. If the -txindex option is\n"
-            "enabled, it also works for blockchain transactions.\n"
+            "enabled, it also works for blockchain transactions. If the block hash is known, it can be provided\n"
+            "for nodes without -txindex, in which case the transaction will only be found if it is in that\n"
```
TheBlueMatt: grammar nit: this reads funny to me, and could be a bit more explicit. Maybe:

"If the block which contains the transaction is known, its hash can be provided even for nodes without -txindex."
"Note that if a blockhash is provided, only it will be searched and if the transaction is in mempool, other blocks, or if this node does not have the given block available, the transaction will not be found."

src/rpc/rawtransaction.cpp
```diff
@@ -130,34 +134,42 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
+            inMainChain = (chainActive[blockIndex->nHeight] == blockIndex);
         }
     }
 
     CTransactionRef tx;
     uint256 hashBlock;
-    if (!GetTransaction(hash, tx, Params().GetConsensus(), hashBlock, true))
-        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string(fTxIndex ? "No such mempool or blockchain transaction"
+    if (!GetTransaction(hash, tx, Params().GetConsensus(), hashBlock, true, blockIndex))
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string(fTxIndex || blockIndex ? "No such mempool or blockchain transaction"
```
TheBlueMatt: Maybe further update this error message, eg if (blockIndex) "No such transaction found in the provided block".

TheBlueMatt (commented):
You may wish to squash "[rpc] Fix fVerbose parsing (remove excess if cases and ensure null is…" and "[test] Updated rawtransactions.py to assert for adjusted exception.".
Generally we try to make sure that after each individual commit, at least it builds and all tests pass.

src/rpc/rawtransaction.cpp
kallewoof: Thanks, that looks better yeah. Adding with minor tweaks.
```C++
            "\nNOTE: By default this function only works for mempool transactions. If the -txindex option is\n"
            "enabled, it also works for blockchain transactions. If the block which contains the transaction\n"
            "is known, its hash can be provided even for nodes without -txindex. Note that if a blockhash is\n"
            "provided, only that block will be searched and if the transaction is in the mempool or other\n"
            "blocks, or if this node does not have the given block available, the transaction will not be found.\n"
```

src/rpc/rawtransaction.cpp
kallewoof: I was sure I did, but guess not:
```C++
            "  \"inMainChain\": b,     (bool) Whether specified block is in the main chain or not (only present with explicit \"blockhash\" argument)\n"
```
I like the idea of including when able but will keep it out of this PR for now.

src/rpc/rawtransaction.cpp
```diff
@@ -130,34 +134,42 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
+        } else if (!request.params[1].isNull()) {
+            fVerbose = (request.params[1].get_bool());
         }
-        else if(request.params[1].isBool()) {
-            if(request.params[1].isTrue()) {
-                fVerbose = true;
+    }
+
+    if (request.params.size() > 2 && !request.params[2].isNull()) {
+        uint256 blockHash = ParseHashV(request.params[2], "parameter 3");
```
kallewoof: The general tendency seems to be to identify the parameter index so I stuck with that. I agree it may be better to be more descriptive though...

src/rpc/rawtransaction.cpp
kallewoof: Yeah, I wanted to avoid `?:?::`. Rewritten as:
```C++
    if (!GetTransaction(hash, tx, Params().GetConsensus(), hashBlock, true, blockIndex)) {
        std::string errmsg;
        if (blockIndex) {
            errmsg = "No such transaction found in the provided block";
        } else {
            errmsg = fTxIndex
              ? "No such mempool or blockchain transaction"
              : "No such mempool transaction. Use -txindex to enable blockchain transaction queries";
        }
        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, errmsg + ". Use gettransaction for wallet transactions.");
    }
```

kallewoof:
@TheBlueMatt Thanks for the review!

> Generally we try to make sure that after each individual commit, at least it builds and all tests pass.

We try to keep tests as separate commits though, so that would assume tests and code changes come in pairs (tests will fail before test commit or after test commit and before change commit, obv). That was my intention with the split here. I may have screwed up. I'll double check and/or squash as appropriate.

TheBlueMatt:
utACK 8f2ce52c9257392d7e6523b39a26e9697a30ca76

jonasschnelli:
Needs rebase.

kallewoof:
Rebased.

jonasschnelli:
a) Is there a reason why mainchain height is not supported as alternative for the blockhash? Eventually with a security of only accepting heights of a hundred blocks below the tip as a reorganisation protection (but I'd prefer to not add this protection).

b) @kallewoof the idea about the standard for a transaction reference has already been worked into a BIP: https://github.com/bitcoin/bips/pull/555 (maybe we can support this – if we agree on that BIP to be worth implementing – also via `getrawtransaction`).

kallewoof:
@jonasschnelli Regarding height, I chose not to include it as it could potentially cause unexpected results when a reorg happens, but if people don't think that's an issue it should be fairly straightforward to allow for both.

src/rpc/rawtransaction.cpp
```diff
@@ -76,12 +79,14 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
             "1. \"txid\"      (string, required) The transaction id\n"
             "2. verbose       (bool, optional, default=false) If false, return a string, otherwise return a json object\n"
+            "3. \"blockhash\" (string, optional) The block in which to look for the transaction\n"
 
             "\nResult (if verbose is not set or set to false):\n"
             "\"data\"      (string) The serialized, hex-encoded data for 'txid'\n"
 
             "\nResult (if verbose is set to true):\n"
             "{\n"
+            "  \"inMainChain\": b,     (bool) Whether specified block is in the main chain or not (only present with explicit \"blockhash\" argument)\n"
```
jtimon: Why not "chain" and return Params().NetworkIDString() like we do in other places (ie getblockchaininfo and getmininginfo)? This seems less generic, like we do with "testnet" in getinfo.

jtimon: This may be confused with Params().NetworkIDString() == "main" (kind of like we do with "testnet" in getinfo)
Can we rename to "inActiveChain" or something of the sort?

src/validation.cpp
```diff
@@ -900,43 +900,44 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
...
 
     LOCK(cs_main);
 
-    CTransactionRef ptx = mempool.get(hash);
-    if (ptx)
-    {
-        txOut = ptx;
-        return true;
-    }
+    if (!blockIndex) {
```
jtimon: Perhaps the diff can be less disruptive by moving everything inside ```if (!blockIndex) {...}``` to a new static function defined right above instead of indenting all of it?

jtimon:
utACK 3ec2d287e24c9cc573ead2898546499abf49db9d besides small nits.

src/validation.cpp
kallewoof: I'm not a huge fan of changing code just to make diffs smaller, and as I mention above, you can put `?w=1` to get diff without indentation changes.

kallewoof:
Addressed @jtimon nits.
@jonasschnelli I added support for supplying block height instead of block hash.

jtimon:
I'm not sure about allowing the height.
One can always call getblockhash if he knows the txid and block height without knowing the block hash somehow.
That's not the case for the BIP proposal @jonasschnelli refers to, since they don't know the txid anyway (only block height and tx position). I suggest we leave the BIP discussion out of this PR, which in my opinion makes sense on its own. And also leave the height option for later (or never).

kallewoof:
@jtimon It feels like a convenience thing that, IMO, getblock should have as well (i.e. allow both height number and block hash). I don't have a strong opinion on the subject though, and will drop the last 2 commits (height support) whenever this is ready for merging unless someone argues for block height support.

jtimon:
Well, don't have a strong opinion on adding the block height either, if more people like it, let's keep it.
Or perhaps it can be proposed after this for both getrawtransaction and getblock at the same time?

instagibbs:
concept ACK, needs rebase/squash, would review

kallewoof:
@instagibbs Rebased & squashed; also removed block height option for now. Will push as separate PR if desirable.

src/rpc/rawtransaction.cpp
```diff
@@ -76,12 +79,14 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
             "1. \"txid\"      (string, required) The transaction id\n"
             "2. verbose       (bool, optional, default=false) If false, return a string, otherwise return a json object\n"
+            "3. \"blockhash\" (string, optional) The block in which to look for the transaction\n"
 
             "\nResult (if verbose is not set or set to false):\n"
             "\"data\"      (string) The serialized, hex-encoded data for 'txid'\n"
 
             "\nResult (if verbose is set to true):\n"
             "{\n"
+            "  \"inActiveChain\": b,   (bool) Whether specified block is in the active chain or not (only present with explicit \"blockhash\" argument)\n"
```
instagibbs: maybe `in_active_chain`? I'm parsing this as "inactive chain"

src/rpc/rawtransaction.cpp
```diff
@@ -129,41 +134,53 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
             + HelpExampleCli("getrawtransaction", "\"mytxid\"")
             + HelpExampleCli("getrawtransaction", "\"mytxid\" true")
             + HelpExampleRpc("getrawtransaction", "\"mytxid\", true")
+            + HelpExampleCli("getrawtransaction", "\"mytxid\" false \"myblockhash\"")
+            + HelpExampleCli("getrawtransaction", "\"mytxid\" true \"myblockhash\"")
         );
 
     LOCK(cs_main);
 
+    bool inActiveChain = true;
```
instagibbs: `in_active_chain` to conform to new style guide and make easier to read.

src/validation.cpp
```diff
@@ -897,43 +897,44 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
...
-            if (txOut->GetHash() != hash)
-                return error("%s: txid mismatch", __func__);
-            return true;
         }
-    }
 
-    if (fAllowSlow) { // use coin database to locate block that contains transaction, and scan it
-        const Coin& coin = AccessByTxid(*pcoinsTip, hash);
-        if (!coin.IsSpent()) pindexSlow = chainActive[coin.nHeight];
+        if (fAllowSlow) { // use coin database to locate block that contains transaction, and scan it
```
instagibbs: should skip this logic if user has provided the blockhash?

instagibbs (changes requested):
Current behavior of GetTransaction means that even if you specify a block index, it still does all the other searches. This means that a user could specify an incorrect blockhash, and still return a valid result, with "inActive" == False(though it may actually be in the valid chain!).

As a user I would expect any call to fail if I simply gave the wrong blockhash, but that's simply my opinion.

src/validation.cpp
kallewoof: This is inside the `if (!blockIndex)` block so it should never execute if a block hash was provided, no?

instagibbs: Oops! Not good at counting brackets.

instagibbs:
I fail at counting brackets.

utACK https://github.com/bitcoin/bitcoin/pull/10275/commits/440123fb8c7196a7ee209495f02b2f7c0deaa0bb

jnewbery (commented):
A few nits inline.

I think you need to squash the first two commits to make the code and test changes atomic (otherwise the test fails on the intermediate 90e140862ac7ac172cf89ed88e8183984affe057 commit)

src/rpc/rawtransaction.cpp
```diff
@@ -129,42 +134,55 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
-    uint256 hashBlock;
-    if (!GetTransaction(hash, tx, Params().GetConsensus(), hashBlock, true))
-        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string(fTxIndex ? "No such mempool or blockchain transaction"
-            : "No such mempool transaction. Use -txindex to enable blockchain transaction queries") +
-            ". Use gettransaction for wallet transactions.");
+    uint256 hash_block;
+    if (!GetTransaction(hash, tx, Params().GetConsensus(), hash_block, true, blockindex)) {
+        std::string errmsg;
+        if (blockindex) {
+            errmsg = "No such transaction found in the provided block";
```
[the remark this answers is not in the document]
kallewoof: Good point!

I added code (https://github.com/bitcoin/bitcoin/pull/10275/commits/3c0cb434bde411d2daf8e4a440c85e80b1eb1f62) that checks if pruning is enabled, and if it is, checks if the block is actually available by reading it from disk and setting the error message based on the results of that.

test/functional/rawtransactions.py
```diff
@@ -52,6 +52,20 @@ def run_test(self):
...
+
+        # make a tx by sending then generate 2 blocks; block1 has the tx in it,
+        # presumably
+        tx = self.nodes[2].sendtoaddress(self.nodes[1].getnewaddress(), 1)
+        block1, block2 = self.nodes[2].generate(2)
+        self.sync_all()
+        # We should be able to get the raw transaction by providing the correct block
+        assert_equal(self.nodes[0].getrawtransaction(tx, True, block1)['txid'], tx)
+        # We should not get the tx if we provide an unrelated block
+        assert_raises_jsonrpc(-5, "No such", self.nodes[0].getrawtransaction, tx, True, block2)
```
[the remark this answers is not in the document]
kallewoof: Oops. Fixed, thanks.

kallewoof:
@jnewbery Thanks for the review! I think I've addressed all your comments.

src/rpc/rawtransaction.cpp
```diff
@@ -76,12 +79,14 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
             "1. \"txid\"      (string, required) The transaction id\n"
             "2. verbose       (bool, optional, default=false) If false, return a string, otherwise return a json object\n"
+            "3. \"blockhash\" (string, optional) The block in which to look for the transaction\n"
 
             "\nResult (if verbose is not set or set to false):\n"
             "\"data\"      (string) The serialized, hex-encoded data for 'txid'\n"
 
             "\nResult (if verbose is set to true):\n"
             "{\n"
+            "  \"in_active_chain\": b, (bool) Whether specified block is in the active chain or not (only present with explicit \"blockhash\" argument)\n"
```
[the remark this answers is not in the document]
jnewbery: Yes, of course you're correct. The `size`, `vsize`, `version`, ... fields below aren't aligned by the same logic, but that's beyond the scope of this PR. Please disregard my previous comment.

jnewbery:
> I try to keep tests and code changes as separate commits. Why would you want it to succeed between the test update? (Or am I confused about what you're asking?)

I think the principal is that each individual commit should be at least internally consistent (should build and pass its own tests). Your first commit updates the product code so the returned error changes, and your second commit updates the test code to pass with the new error string. That means that if I run the tests after the first commit they'll fail.

I understand the desire to structure the commits in a way that tells a story, but this breaks tools like git bisect that rely on commits to be internally consistent.

kallewoof:
@jnewbery You can `git bisect skip` if you land on a commit squeezed in between tests. I don't really agree with squashing test updates and code changes together, personally. That's how I've always done it, anyway.

promag:
I agree with @jnewbery, one commit should not break the code, tests including. If you revert that commit, codes stays green. IMO makes reviewing easier.

src/rpc/rawtransaction.cpp
```diff
@@ -129,42 +134,63 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
-    if (!GetTransaction(hash, tx, Params().GetConsensus(), hashBlock, true))
-        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string(fTxIndex ? "No such mempool or blockchain transaction"
-            : "No such mempool transaction. Use -txindex to enable blockchain transaction queries") +
-            ". Use gettransaction for wallet transactions.");
+    uint256 hash_block;
+    if (!GetTransaction(hash, tx, Params().GetConsensus(), hash_block, true, blockindex)) {
+        std::string errmsg;
+        if (blockindex) {
+            bool block_available = true;
+            if (gArgs.GetArg("-prune", 0)) {
```
luke-jr: Use `fHavePruned`

src/rpc/rawtransaction.cpp
```diff
@@ -129,42 +134,63 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
+                // we are pruning; see if block is available at all
+                CBlock block;
+                block_available = ReadBlockFromDisk(block, blockindex, Params().GetConsensus());
```
luke-jr: Use `!(pblockindex->nStatus & BLOCK_HAVE_DATA) && pblockindex->nTx > 0` (see `getblock`)

luke-jr (changes requested):

jtimon:
I agree individual commits should never break the tests ideally. That doesn't mean that you can't introduce new functionality and its new tests in separated commits when that's not the case.
Needs rebase again.

jnewbery:
Last two commits look good. For the pruning error, I think you can error out earlier by re-using the pruned test from `getblock`:

```
     if (fHavePruned && !(pblockindex->nStatus & BLOCK_HAVE_DATA) && pblockindex->nTx > 0)
        throw JSONRPCError(RPC_MISC_ERROR, "Block not available (pruned data)");
```

> I don't really agree with squashing test updates and code changes together, personally.

ok. I think both ways are valid, and I've come across both before. I'm not sure if this project has a preferred style.

kallewoof:
I'm merging those commits as it's not new functionality but just a fix that changes an error message.

kallewoof:
@luke-jr Nice!

src/rpc/rawtransaction.cpp
```diff
@@ -130,42 +135,58 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
-        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string(fTxIndex ? "No such mempool or blockchain transaction"
-            : "No such mempool transaction. Use -txindex to enable blockchain transaction queries") +
-            ". Use gettransaction for wallet transactions.");
+    uint256 hash_block;
+    if (!GetTransaction(hash, tx, Params().GetConsensus(), hash_block, true, blockindex)) {
+        std::string errmsg;
+        if (blockindex) {
+            bool block_pruned = fHavePruned && !(blockindex->nStatus & BLOCK_HAVE_DATA) && blockindex->nTx > 0;
+            errmsg = block_pruned
+              ? "Block not available (pruned data)"
```
jnewbery: supernit: Not finding a block because it's been pruned is not a `RPC_INVALID_ADDRESS_OR_KEY` error. It would more accurately be a `RPC_MISC_ERROR`.

My suggestion would be to do the pruned check above, immediately after the `throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Block hash not found");` line

kallewoof: Yeah, I was thinking about that too. Fixing.

src/rpc/rawtransaction.cpp
```diff
@@ -130,42 +135,58 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
     if (!request.params[1].isNull()) {
-        if (request.params[1].isNum()) {
-            if (request.params[1].get_int() != 0) {
-                fVerbose = true;
-            }
-        }
-        else if(request.params[1].isBool()) {
-            if(request.params[1].isTrue()) {
-                fVerbose = true;
+        fVerbose = request.params[1].isNum() ? (request.params[1].get_int() != 0) : request.params[1].get_bool();
```
jtimon: yay! +1-8

test/functional/rawtransactions.py
```diff
@@ -50,6 +50,19 @@ def run_test(self):
...
         assert_raises_jsonrpc(-25, "Missing inputs", self.nodes[2].sendrawtransaction, rawtx['hex'])
 
+        ############################
+        # getrawtx with block hash #
+        ############################
+
+        # make a tx by sending then generate 2 blocks; block1 has the tx in it
+        tx = self.nodes[2].sendtoaddress(self.nodes[1].getnewaddress(), 1)
+        block1, block2 = self.nodes[2].generate(2)
+        self.sync_all()
```
jtimon: Perhaps you need to sync after generating the first block?

jtimon:
re-utACK 300a5f15d54484691d47eb5ec23acde6d82bfda1

src/rpc/rawtransaction.cpp
```diff
@@ -130,42 +135,58 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
-    uint256 hashBlock;
-    if (!GetTransaction(hash, tx, Params().GetConsensus(), hashBlock, true))
-        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string(fTxIndex ? "No such mempool or blockchain transaction"
-            : "No such mempool transaction. Use -txindex to enable blockchain transaction queries") +
-            ". Use gettransaction for wallet transactions.");
+    uint256 hash_block;
+    if (!GetTransaction(hash, tx, Params().GetConsensus(), hash_block, true, blockindex)) {
+        std::string errmsg;
+        if (blockindex) {
+            if (fHavePruned && !(blockindex->nStatus & BLOCK_HAVE_DATA) && blockindex->nTx > 0) {
```
TheBlueMatt: Why check nTx here? Its just as useful to provide an error of "Block not available" if it has been pruned as if it has not yet been received in full.

TheBlueMatt (commented):
Needs trivial rebase. Note that first commit message is no longer correct (says "ensure null is accepted" but null is already accepted on master, commit contents are still good, though). Otherwise largely looks good.

src/rpc/rawtransaction.cpp
kallewoof: This is how the check is done in a lot of places. E.g. 
https://github.com/kallewoof/bitcoin/blame/gettx-with-blockhash/src/rest.cpp#L219

I'm not sure why an additional nTx check is needed, to be honest.

sipa: "nTx > 0" means we had the block's contents at some point.

TheBlueMatt: Yes, I'm saying remove the fHavePruned and nTx check here - if we have a block's header but not its body, then we'll throw a "No such transaction found in the provided block" error, which is nonsense. Easier to just make this error generic.

kallewoof: Ahh, OK.

kallewoof:
@jnewbery You wouldn't happen to have any hints on what's going on here? https://travis-ci.org/bitcoin/bitcoin/jobs/308238239
```
make  \
	  top_distdir="bitcoin-x86_64-unknown-linux-gnu" distdir="bitcoin-x86_64-unknown-linux-gnu" \
	  dist-hook
make[1]: Entering directory `/home/travis/build/bitcoin/bitcoin/build'
/usr/bin/git archive --format=tar HEAD -- src/clientversion.cpp | ${TAR-tar} -C bitcoin-x86_64-unknown-linux-gnu -xf -
fatal: current working directory is untracked
tar: This does not look like a tar archive
tar: Exiting with failure status due to previous errors
make[1]: [dist-hook] Error 2 (ignored)
make[1]: Leaving directory `/home/travis/build/bitcoin/bitcoin/build'
```

src/rpc/rawtransaction.cpp
```diff
@@ -132,42 +137,58 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
-    if (!GetTransaction(hash, tx, Params().GetConsensus(), hashBlock, true))
-        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string(fTxIndex ? "No such mempool or blockchain transaction"
-            : "No such mempool transaction. Use -txindex to enable blockchain transaction queries") +
-            ". Use gettransaction for wallet transactions.");
+    uint256 hash_block;
+    if (!GetTransaction(hash, tx, Params().GetConsensus(), hash_block, true, blockindex)) {
+        std::string errmsg;
+        if (blockindex) {
+            if (!(blockindex->nStatus & BLOCK_HAVE_DATA)) {
+                throw JSONRPCError(RPC_MISC_ERROR, "Block not available (pruned data)");
```
TheBlueMatt: Now the error message is wrong :/. You should drop the "(pruned data)" note, as this can happen if you've received the header but not (yet) the block.

src/validation.cpp
```diff
@@ -927,46 +927,44 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
...
-            if (txOut->GetHash() != hash)
-                return error("%s: txid mismatch", __func__);
+    if (!blockIndex) {
+        CTransactionRef ptx = mempool.get(hash);
+        if (ptx) {
+            txOut = ptx;
             return true;
         }
 
-        // transaction not found in index, nothing more can be done
```
TheBlueMatt: Looks like this line was lost in a rebase? Shouldn't remove it.

src/rpc/rawtransaction.cpp
kallewoof: Oh, damn. Sorry about that. Fixed.

src/validation.cpp
kallewoof: Fixed!

sipa:
I realize I'm very late with this, so this is just for consideration and not an objection to this PR.

I believe `getrawtransaction` is already a weird mixed bag of semantics (it works when a transaction is confirmed but at least one output is not yet spent by a confirmed transaction, or if the transaction is unconfirmed, or if txindex is available for everything but the genesis block). Adding yet another combination that works will only further the difficulty of explaining it.

Ideally, I think this RPC should be replaced with a `lookuprawtransaction` (or so) which only works when `-txindex` is enabled. In case you only know a transaction is unconfirmed, use `getmempoolentry` which already exists. This functionality provided here should IMHO just be a separate RPC.

src/rpc/rawtransaction.cpp
```diff
@@ -132,42 +137,58 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
+        if (!blockhash.IsNull()) {
+            BlockMap::iterator it = mapBlockIndex.find(blockhash);
+            if (it == mapBlockIndex.end()) {
+                throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Block hash not found");
             }
-        }
-        else {
-            throw JSONRPCError(RPC_TYPE_ERROR, "Invalid type provided. Verbose parameter must be a boolean.");
+            blockindex = it->second;
+            in_active_chain = (chainActive[blockindex->nHeight] == blockindex);
```
promag: Use `chainActive.Contains()`?

TheBlueMatt:
Dear God, this PR is 7 months old. The behavior changes are all well-documented and it's a very useful feature, there's gotta be a statue of limitations on concept objections.

src/rpc/rawtransaction.cpp
```diff
@@ -132,42 +137,58 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
-            }
-        }
-        else if(request.params[1].isBool()) {
-            if(request.params[1].isTrue()) {
-                fVerbose = true;
+        fVerbose = request.params[1].isNum() ? (request.params[1].get_int() != 0) : request.params[1].get_bool();
+    }
+
+    if (!request.params[2].isNull()) {
+        uint256 blockhash = ParseHashV(request.params[2], "parameter 3");
```
promag: Add tests for invalid hash?
 - must be string;
 - must be hexadecimal string;
 - must be of length.

src/rpc/rawtransaction.cpp
```diff
@@ -132,42 +137,58 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
...
+    }
 
-    if (!fVerbose)
+    if (!fVerbose) {
         return EncodeHexTx(*tx, RPCSerializationFlags());
+    }
 
     UniValue result(UniValue::VOBJ);
-    TxToJSON(*tx, hashBlock, result);
+    if (blockindex) result.push_back(Pair("in_active_chain", in_active_chain));
```
promag: Missing test for `in_active_chain`.

sipa:
I'm not objecting in any way, just offering a view to consider.

kallewoof:
I agree `getrawtransaction` is very confusing until you understand its limitations. I'll gladly work on streamlining it into potentially two or more RPC calls post-merge. Biggest concern is backwards compatibility and (alternatively) bloat.

test/functional/rawtransactions.py
```diff
@@ -50,6 +50,19 @@ def run_test(self):
         # This will raise an exception since there are missing inputs
         assert_raises_rpc_error(-25, "Missing inputs", self.nodes[2].sendrawtransaction, rawtx['hex'])
 
+        ############################
+        # getrawtx with block hash #
```
promag: `# getrawtransaction with block hash #`?

src/validation.cpp
```diff
@@ -927,46 +927,47 @@ bool AcceptToMemoryPool(CTxMemPool& pool, CValidationState &state, const CTransa
 }
 
 /** Return transaction in txOut, and if it was found inside a block, its hash is placed in hashBlock */
```
promag: Care to explain the new argument?

src/rpc/rawtransaction.cpp
```diff
@@ -79,12 +82,14 @@ UniValue getrawtransaction(const JSONRPCRequest& request)
             "\nArguments:\n"
             "1. \"txid\"      (string, required) The transaction id\n"
             "2. verbose       (bool, optional, default=false) If false, return a string, otherwise return a json object\n"
```
promag: Nit, mind to fix `verbose` alignment?

promag (commented):
IMO overloaded API's are harder to maintain, test and document. Anyway, some comments below.

kallewoof:
@promag Thanks for the review! I think I addressed all your comments.

test/functional/rawtransactions.py
```diff
@@ -50,6 +50,26 @@ def run_test(self):
...
+        self.sync_all()
+        # We should be able to get the raw transaction by providing the correct block
+        gottx = self.nodes[0].getrawtransaction(tx, True, block1)
+        assert_equal(gottx['txid'], tx)
+        assert_equal(gottx['in_active_chain'], True)
+        # We should not get the tx if we provide an unrelated block
+        assert_raises_rpc_error(-5, "No such transaction found", self.nodes[0].getrawtransaction, tx, True, block2)
+        # An invalid block hash should raise the correct errors
+        assert_raises_rpc_error(-8, "parameter 3 must be hexadecimal", self.nodes[0].getrawtransaction, tx, True, True)
+        assert_raises_rpc_error(-8, "parameter 3 must be hexadecimal", self.nodes[0].getrawtransaction, tx, True, "foobar")
```
promag: Can remove this.

promag (commented):
Missing a test checking that `in_active_chain` is not in the response when `blockhash` is not provided.

test/functional/rawtransactions.py
kallewoof: First test checks for non-string, this one for non-hexadecimal string. May be redundant but they check different things.

kallewoof:
@promag Added test for unset `in_active_chain`.

test/functional/rawtransactions.py
```diff
@@ -188,13 +212,13 @@ def run_test(self):
         assert_equal(self.nodes[0].getrawtransaction(txHash, True)["hex"], rawTxSigned['hex'])
 
         # 6. invalid parameters - supply txid and string "Flase"
-        assert_raises_rpc_error(-3,"Invalid type", self.nodes[0].getrawtransaction, txHash, "Flase")
+        assert_raises_rpc_error(-1,"not a boolean", self.nodes[0].getrawtransaction, txHash, "Flase")
```
promag: Nit, add space after `,`. Same below.

test/functional/rawtransactions.py
```diff
@@ -50,6 +50,30 @@ def run_test(self):
...
 
+        #####################################
+        # getrawtransaction with block hash #
+        #####################################
+
+        # make a tx by sending then generate 2 blocks; block1 has the tx in it
+        tx = self.nodes[2].sendtoaddress(self.nodes[1].getnewaddress(), 1)
+        block1, block2 = self.nodes[2].generate(2)
+        self.sync_all()
+        # We should be able to get the raw transaction by providing the correct block
```
promag: Nit, remaining comments are lowercase. Same below.

test/functional/rawtransactions.py
```diff
@@ -50,6 +50,30 @@ def run_test(self):
...
+        gottx = self.nodes[0].getrawtransaction(tx, True, block1)
+        assert_equal(gottx['txid'], tx)
+        assert_equal(gottx['in_active_chain'], True)
+        # We should not have the 'in_active_chain' flag when we don't provide a block
+        gottx = self.nodes[0].getrawtransaction(tx, True)
```
promag: Calling without blockhash in a *section* where the RPC is tested with a block hash :trollface: maybe move this elsewhere?

src/rpc/rawtransaction.cpp
promag: Could use parameter name instead of `parameter 3`? This is ugly for named args. (fix others in a new commit or in a follow up PR, whatever).

promag (commented):
Mind some nits.

TheBlueMatt:
utACK 434526aba680cb73208e018a02827d51a71cfff6

src/rpc/rawtransaction.cpp
```diff
@@ -983,7 +1016,7 @@ UniValue sendrawtransaction(const JSONRPCRequest& request)
 static const CRPCCommand commands[] =
 { //  category              name                      actor (function)         argNames
   //  --------------------- ------------------------  -----------------------  ----------
-    { "rawtransactions",    "getrawtransaction",      &getrawtransaction,      {"txid","verbose"} },
+    { "rawtransactions",    "getrawtransaction",      &getrawtransaction,      {"txid","verbose","blockhash"} },
```
MarcoFalke: The optional parameter doesn't work in the gui. Pls fix `src/rpc/client.cpp`

maflcko (commented):
utACK 434526aba680cb73208e018a02827d51a71cfff6

MarcoFalke:
Sorry for the sloppy review. I messed up the parameter order when testing. Also gcc complains about my nit-fix.

Tested ACK now:

```
Hash: SHA512

Tested ACK 434526aba680cb73208e018a02827d51a71cfff6
-----BEGIN PGP SIGNATURE-----

iQIcBAEBCgAGBQJaJ1PQAAoJENLqSFDnUoslonMQAJz8bVoePuyYUgOHkbGIGemj
4yUyywvp34A4BocYrczH9Im6q+/VBoLrxd+7BqcrhWclHC2G7tvVircgXS/O/RHE
SrEjox5dpLnQ5vW2Ik1hZkdlWvJvn6KBPRqLxZ8qLVFE2ym9BapDhZy33kePh4If
BPBalz7U/1rMz2p2a7yKLo26+MArSwt9kiGebuzgUeaUREjn3gQhaouo4mGSVN59
3Fc1zw3TS1zRuXek/EGL6uuRNQsZZkS08maHtZcsigtwS9drrekrPgx2LswvqGoN
HNkF3AkVK2RZ96cvyG1DbGPG8PkY+EQxO97de2Jlq8/8qQ7wu/BrwJvhjGpdoqGI
SL2pA9KgPfWDTDr13Tcavm+iGk+SbI99Hk+MgsyRArplPJYUtTi43WID/uRaydyE
mCykCPFHkDehVIb8FL/MxSCyBzqJxn4IGXL/Q4A+BbbEYgu1Yi8xtE1o/sCqHyvD
qoKlj1ns9JhNOIpdZNhaWvVb30VuugcZh1/I7CE4NJINdaP7oYKLS0JCiBjwxLKN
6lvBYfoJmNLOvEOiB5DGeQEtnU/mBeGrDgoRddNpnKWRQMOIKi53aKvmeJ4XNw0T
UcIAxaj6c+lZNKYeedePcuRKapqIW9fEQ9IAI8+duRisl6TBcn7/0bJHjWLgy062
NtvMJ3zBCmEhX3Vs4km5
=6+iY
-----END PGP SIGNATURE-----

File: test/functional/rawtransactions.py (hunk 37 above)
Comment (kallewoof) [reply test/functional/rawtransactions.py]: `in_active_chain` is a part of the blockhash stuff, though, so it sort of makes sense to keep the test together with the others.

File: src/rpc/rawtransaction.cpp (hunk 29 above)
Comment (kallewoof) [reply src/rpc/rawtransaction.cpp]: This is used in a ton of places. I never saw consensus on what exactly to do and chose to stick with it for now. As you said, different PR.

**Comment by MarcoFalke** [comment]
@jtimon @jnewbery 
Mind to re-ACK?

**Comment by jonasschnelli** [comment]
utACK 434526aba680cb73208e018a02827d51a71cfff6

File: src/rpc/rawtransaction.cpp (hunk 29 above)
Comment (jnewbery) [inline src/rpc/rawtransaction.cpp]: Nit: these lines are confusing for me:

```
        uint256 blockhash = ParseHashV(request.params[2], "parameter 3");
        if (!blockhash.IsNull()) {
```

`ParseHashV()` can either throw or return a uint256. The only way it'll return a null uint256 is if the input is a 64 length string of `0`s. That can't be a valid block (if someone finds a block that hashes to zero then they win Bitcoin The Game and we can all stop playing), so I think that this if condition is unnecessary - if parameter 3 is all zeroes, we should throw with the error that 0x0 is an invalid block.

I don't think this is a huge problem - if all zeroes is supplied as the block hash, we'll just drop through and search for the transaction in the {mempool , blockchain (if fTxIndex) , UTXO set} as if no block hash had been provided. However, it's a bit confusing for the code reader since it implies to me that we're expecting `ParseHashV()` to return a null value if it fails to parse the hash.

jnewbery (commented):
Post-merge tested ACK. I have one style nit, which I can open a tidy-up PR for if people agree.

promag:
For reference, This was proposed (and closed) in #2421.

kallewoof:
@promag I believe the concerns raised in #2421 are not relevant anymore / in this implementation.

PR #10276, opened 2017-04-25
contrib/verifybinaries: allow filtering by platform

knocte:
Downloading all the binaries of all platforms can take quite long,
especially for slow connections, which may deter people from using
this script and, therefore, to disregard security altogether.

This change introduces the new possibility of specifying the
platform along with the version number, so that only the binaries
that contain the platform name are downloaded.

### DIFF
--- a/contrib/verifybinaries/README.md
+++ b/contrib/verifybinaries/README.md
@@ -26,6 +26,14 @@ The script returns 0 if everything passes the checks. It returns 1 if either the
 ./verify.sh bitcoin-core-0.13.0-rc3
 ```
 
+If you only want to download the binaries of certain platform, add the corresponding suffix, e.g.:
+
+```sh
+./verify.sh bitcoin-core-0.11.2-osx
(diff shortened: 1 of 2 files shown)


laanwj:
Concept ACK, seems useful.

jonasschnelli:
Concept ACK

laanwj:
Just tested this. Found one issue: looks like the script hangs indefinitely when providing an argument that doesn't match any files.

knocte:
I'll fix it, thanks for the feedback.

knocte:
Updated: if a platform name is provided that doesn't match any files, it will fail now with an error instead of hanging.

NB: I don't know why the windows travis-CI build failed, looks unrelated to this...

MarcoFalke:
ACK 8d4dafd774436a355b6aa49a44b867cf78c38f6d

laanwj:
ACK 8d4dafd

PR #10278, opened 2017-04-25
[test] Add Unit Test for GetListenPort

jimmysong:
Add very basic unit test for GetListenPort in net_tests.cpp

### DIFF
--- a/src/test/net_tests.cpp
+++ b/src/test/net_tests.cpp
@@ -11,6 +11,7 @@
 #include "net.h"
 #include "netbase.h"
 #include "chainparams.h"
+#include "util.h"
 
 class CAddrManSerializationMock : public CAddrMan
 {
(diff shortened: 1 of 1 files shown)


jimmysong:
Please let me know if this is too basic a unit test. I'm using unit test writing as a way to get familiar with the code base.

laanwj:
This tests only one case of `GetListenPort`. Possibly this could be extended with a test for the `-port` argument as well, e.g. temporarily set the option to something else and see if the returned value changes to that.

Anyhow, thanks for adding a test.

laanwj:
Also might make sense to compare against `Params().GetDefaultPort()` instead of hardcoding the number.

jimmysong:
Thanks for the review @laanwj, I've changed the test per your suggestions and squashed.

src/test/net_tests.cpp
```diff
@@ -72,6 +73,18 @@ CDataStream AddrmanToStream(CAddrManSerializationMock& _addrman)
...
 BOOST_FIXTURE_TEST_SUITE(net_tests, BasicTestingSetup)
 
+BOOST_AUTO_TEST_CASE(cnode_listen_port)
+{
+    // test default
+    unsigned short port = GetListenPort();
+    BOOST_CHECK(port == Params().GetDefaultPort());
+    // test set port
+    unsigned short altPort = 12345;
+    SoftSetArg("-port", std::to_string(altPort));
```
MarcoFalke: This will modify the global state of test run after this test case. There are other places where this happnes, but imo, unit tests should really be unit test.

maflcko (commented):
Not sure if implementing unit tests for functions by effectively copy pasting the function into another cpp file and checking equality for a one or two input arguments is really worth it. In case GetListenPort breaks, I assume all code that depends on GetListenPort breaks as well.

Though, you mentioned that you wanted to get familiar with writing tests and this is already merged, so fine with me.

src/test/net_tests.cpp
jimmysong: Thanks for the feedback!

What should I do in the future? Unset the arg afterwards?

MarcoFalke: Yes, I think this is reasonable for tests.

PR #10281, opened 2017-04-26
doc: Add RPC interface guidelines

laanwj:
Could be useful. Anything that I forgot?

### DIFF
--- a/doc/developer-notes.md
+++ b/doc/developer-notes.md
@@ -495,3 +495,76 @@ Git and GitHub tips
   This will add an `upstream-pull` remote to your git repository, which can be fetched using `git fetch --all`
   or `git fetch upstream-pull`. Afterwards, you can use `upstream-pull/NUMBER/head` in arguments to `git show`,
   `git checkout` and anywhere a commit id would be acceptable to see the changes from pull request NUMBER.
+
+RPC interface guidelines
+--------------------------
+
+A few guidelines for introducing and reviewing new RPC interfaces:
+
+- Method naming: use consecutive lower-case names such as `getrawtransaction` and `submitblock`
+
+  - *Rationale*: Consistency with existing interface.
+
+- Argument naming: use snake case `fee_delta` (and not, e.g. camel case `feeDelta`)
+
+  - *Rationale*: Consistency with existing interface.
+
+- Use the JSON parser for parsing, don't manually parse integers or strings from
+  arguments unless absolutely necessary.
+
+  - *Rationale*: Introduces hand-rolled string manipulation code at both the caller and callee sites,
+    which is error prone, and it is easy to get things such as escaping wrong.
+    JSON already supports nested data structures, no need to re-invent the wheel.
+
+  - *Exception*: AmountToValue can parse amounts as string. This was introduced because many JSON
+    parsers and formatters hard-code handling decimal numbers as floating point
+    values, resulting in potential loss of precision. This is unacceptable for
+    monetary values. **Always** use `AmountToValue` and `ValueToAmount` when
+    inputting or outputting monetary values. The only exceptions to this are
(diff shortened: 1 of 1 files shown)


doc/developer-notes.md
```diff
@@ -495,3 +495,68 @@ Git and GitHub tips
...
+    new ones.
+
+- Try not to overload methods on argument type. E.g. don't make `getblock(true)` and `getblock("hash")`
+  do different things.
+
+  - *Rationale*: This is impossible to use with `bitcoin-cli`, and can be surprising to users.
+
+  - *Exception*: Some RPC calls can take both an `int` and `bool`, most notably when a bool was switched
+    to a multi-value, or due to other historical reasons. **Always** have false map to 0 and
+    true to 1 in this case.
```
kallewoof: This is not the case right now. At the moment, 0 maps to false and !0 maps to true. See e.g. `getrawtransaction`'s `verbose` argument handling.

doc/developer-notes.md
```diff
@@ -495,3 +495,68 @@ Git and GitHub tips
...
+
+- Don't forget to fill in the argument names correctly in the RPC command table.
+
+  - *Rationale*: If not, the call can not be used with name-based arguments.
+
+- Set okSafeMode in the RPC command table to a sensible value: safe mode is when the
+  block chain is regarded to be in a confused state, and the client deems it unsafe to
+  do anything irreversible such as send. Anything that just queries should be permitted.
```
kallewoof: Thanks for this! I vaguely presumed this was the case but never found anything explicitly stating it.

kallewoof (commented):
Nice.

jonasschnelli:
Great addition.
Concept ACK.

doc/developer-notes.md
```diff
@@ -495,3 +495,68 @@ Git and GitHub tips
...
+  - *Exception*: Some RPC calls can take both an `int` and `bool`, most notably when a bool was switched
+    to a multi-value, or due to other historical reasons. **Always** have false map to 0 and
+    true to 1 in this case.
+
+- Don't forget to fill in the argument names correctly in the RPC command table.
+
+  - *Rationale*: If not, the call can not be used with name-based arguments.
+
+- Set okSafeMode in the RPC command table to a sensible value: safe mode is when the
+  block chain is regarded to be in a confused state, and the client deems it unsafe to
```
fanquake: nit: s/block chain/blockchain

doc/developer-notes.md
laanwj: EH, I have that the wrong way around, thanks.

laanwj: Or not. What I meant here is:
```
false=0
true=1
```
E.g, the C++ mapping, not the bash mapping. 
If this wording is confusing, can you suggest another way.

kallewoof: I read it as "interpret 0 as false and 1 as true, and anything else is [??]".

I.e. if a user passes 2 or -1 to the new RPC command which maps as you said, how will it treat values beyond [0, 1]?

Am I misunderstanding?

laanwj: I don't say anything about values besides 0 and 1 on purpose. Those may either map to 1, or to something else entirely (referring to "when bool has switched to multi-value").

kallewoof: Ohh I was reading this backwards. You're talking about when a bool becomes an int, and I about when an int is interpreted as a bool. I'll think on wording to make this obvious.

kallewoof: Re-reading this, I think I was simply not reading it properly. I think it looks fine as is.

jnewbery:
Looks good. ACK 359f71d0fb243fa49884132c0fb0c24118d77842

sipa:
ACK

PR #10283, opened 2017-04-26
Cleanup: reduce to one GetMinimumFee call signature

morcos:

PR #10284, opened 2017-04-26
Always log debug information for fee calculation in CreateTransaction

morcos:
This is a first pass at always logging a single line of information whenever CreateTransaction is invoked which will help debug any problems in fee calculation or estimation.

Built on top of #10199 and #10283, only the last commit is new.

### DIFF
--- a/src/policy/fees.h
+++ b/src/policy/fees.h
@@ -74,6 +74,22 @@ enum FeeEstimateHorizon {
     LONG_HALFLIFE = 2
 };
 
+/* Enumeration of reason for returned fee estimate */
+enum class FeeReason {
+    NONE,
+    HALF_ESTIMATE,
+    FULL_ESTIMATE,
+    DOUBLE_ESTIMATE,
+    CONSERVATIVE,
+    MEMPOOL_MIN,
+    PAYTXFEE,
+    FALLBACK,
+    REQUIRED,
+    MAXTXFEE,
+};
+
+std::string StringForFeeReason(FeeReason reason);
+
 /* Used to return detailed information about a feerate bucket */
 struct EstimatorBucket
 {
@@ -90,8 +106,16 @@ struct EstimationResult
 {
     EstimatorBucket pass;
     EstimatorBucket fail;
-    double decay;
-    unsigned int scale;
+    double decay = 0;
+    unsigned int scale = 0;
+};
+
+struct FeeCalculation
+{
+    EstimationResult est;
+    FeeReason reason = FeeReason::NONE;
+    int desiredTarget = 0;
+    int returnedTarget = 0;
 };
 
 /**
@@ -173,7 +197,7 @@ public:
      *  the closest target where one can be given.  'conservative' estimates are
      *  valid over longer time horizons also.
      */
-    CFeeRate estimateSmartFee(int confTarget, int *answerFoundAtTarget, const CTxMemPool& pool, bool conservative = true) const;
+    CFeeRate estimateSmartFee(int confTarget, FeeCalculation *feeCalc, const CTxMemPool& pool, bool conservative = true) const;
 
     /** Return a specific fee estimate calculation with a given success
      * threshold and time horizon, and optionally return detailed data about
@@ -223,9 +247,9 @@ private:
     bool processBlockTx(unsigned int nBlockHeight, const CTxMemPoolEntry* entry);
 
     /** Helper for estimateSmartFee */
-    double estimateCombinedFee(unsigned int confTarget, double successThreshold, bool checkShorterHorizon) const;
+    double estimateCombinedFee(unsigned int confTarget, double successThreshold, bool checkShorterHorizon, EstimationResult *result) const;
     /** Helper for estimateSmartFee */
-    double estimateConservativeFee(unsigned int doubleTarget) const;
+    double estimateConservativeFee(unsigned int doubleTarget, EstimationResult *result) const;
     /** Number of blocks of data recorded while fee estimates have been running */
     unsigned int BlockSpan() const;
     /** Number of blocks of recorded fee estimate data represented in saved data file */
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -651,8 +651,8 @@ void SendCoinsDialog::updateSmartFeeLabel()
         return;
 
     int nBlocksToConfirm = ui->sliderSmartFee->maximum() - ui->sliderSmartFee->value() + 2;
-    int estimateFoundAtBlocks = nBlocksToConfirm;
-    CFeeRate feeRate = ::feeEstimator.estimateSmartFee(nBlocksToConfirm, &estimateFoundAtBlocks, ::mempool);
+    FeeCalculation feeCalc;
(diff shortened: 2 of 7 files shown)


src/wallet/feebumper.cpp
```diff
@@ -160,7 +160,7 @@ CFeeBumper::CFeeBumper(const CWallet *pWallet, const uint256 txidIn, int newConf
     } else {
         // if user specified a confirm target then don't consider any global payTxFee
         if (specifiedConfirmTarget) {
-            nNewFee = CWallet::GetMinimumFee(maxNewTxSize, newConfirmTarget, mempool, ::feeEstimator, true);
+            nNewFee = CWallet::GetMinimumFee(maxNewTxSize, newConfirmTarget, mempool, ::feeEstimator, nullptr, true);
```
jonasschnelli: Would logging the fee estimation details during bumpfee (all call to `GetMinimumFee`) be possible?

src/wallet/wallet.cpp
```diff
@@ -2777,23 +2787,32 @@ CAmount CWallet::GetRequiredFee(unsigned int nTxBytes)
...
     // User didn't set: use -txconfirmtarget to estimate...
     if (nFeeNeeded == 0 || ignoreUserSetFee) {
-        int estimateFoundTarget = nConfirmTarget;
-        nFeeNeeded = estimator.estimateSmartFee(nConfirmTarget, &estimateFoundTarget, pool).GetFee(nTxBytes);
+        nFeeNeeded = estimator.estimateSmartFee(nConfirmTarget, feeCalc, pool, true).GetFee(nTxBytes);
         // ... unless we don't have enough mempool data for estimatefee, then use fallbackFee
-        if (nFeeNeeded == 0)
+        if (nFeeNeeded == 0) {
             nFeeNeeded = fallbackFee.GetFee(nTxBytes);
+            if (feeCalc) feeCalc->reason = FeeReason::FALLBACK;
```
jonasschnelli: supernit: use brackets `if (feeCalc) { feeCalc->reason = FeeReason::FALLBACK; }`

sipa: No. From https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md:
```
If an if only has a single-statement then-clause, it can appear on the same line as the if, without braces. In every other case, braces are required, and the then and else clauses must appear correctly indented on a new line.
```

jonasschnelli: Okay. Fair point.

morcos:
rebased

src/policy/fees.h
```diff
@@ -74,6 +74,33 @@ enum FeeEstimateHorizon {
     LONG_HALFLIFE = 2
 };
 
+/* Enumeration of reason for returned fee estimate */
+enum FeeReason {
```
laanwj: shouldn't this be "enum class" to avoid the enum values from leaking into global scope?

src/wallet/wallet.cpp
```diff
@@ -2734,6 +2735,15 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CWalletT
             return false;
         }
     }
+
+    LogPrintf("Fee Calculation: Fee:%s Bytes:%u Tgt:%d (requested %d) Reason:\"%s\" Decay %.5f: Estimation: (%g - %g) %.2f%% %.1f/(%.1f %d mem %.1f out) Fail: (%g - %g) %.2f%% %.1f/(%.1f %d mem %.1f out)\n",
```
laanwj: Nice! That should give a lot of info for troubleshooting this.

ryanofsky (commented):
utACK d0fb0f69b91539166de98820251d3ac047be0de0, though needs rebase (conflicted with #10390).

src/qt/sendcoinsdialog.cpp
```diff
@@ -679,7 +679,7 @@ void SendCoinsDialog::updateSmartFeeLabel()
         ui->labelSmartFee->setText(BitcoinUnits::formatWithUnit(model->getOptionsModel()->getDisplayUnit(),
                                                                 std::max(feeRate.GetFeePerK(), CWallet::GetRequiredFee(1000))) + "/kB");
         ui->labelSmartFee2->hide();
-        ui->labelFeeEstimation->setText(tr("Estimated to begin confirmation within %n block(s).", "", estimateFoundAtBlocks));
+        ui->labelFeeEstimation->setText(tr("Estimated to begin confirmation within %n block(s).", "", feeCalc.desiredTarget));
```
morcos: damn, i made that same bug elsewhere.  thx

src/wallet/feebumper.cpp
morcos: I think this is possible, but it wasn't clear to me we wanted it here.
Perhaps that could be added later

src/wallet/wallet.cpp
jnewbery: Should this be `LogPrint(BCLog::ESTIMATEFEE...` (ie only log if estimatefee logs are enabled)?

morcos: It's meant to get logged even if you have not selected any debug log categories

morcos:
Addressed comments and squashed/rebased 
[debugEstimates.ver2](https://github.com/morcos/bitcoin/commits/debugEstimates.ver2) --> [debugEstimates.ver2.squash](https://github.com/morcos/bitcoin/commits/debugEstimates.ver2.squash)  --> [debugEstimates.ver2.rebased](https://github.com/morcos/bitcoin/commits/debugEstimates.ver2.rebased)

src/policy/fees.h
```diff
@@ -74,6 +74,22 @@ enum FeeEstimateHorizon {
...
+    DOUBLE_ESTIMATE,
+    CONSERVATIVE,
+    MEMPOOL_MIN,
+    PAYTXFEE,
+    FALLBACK,
+    REQUIRED,
+    MAXTXFEE,
+};
+
+const std::string StringForFeeReason(FeeReason reason);
```
ryanofsky: I don't think there's any benefit to making this const (cases where you do want this are pretty rare https://stackoverflow.com/questions/8716330/purpose-of-returning-by-const-value).

ryanofsky (commented):
utACK 436207ef833dfdd6dc2e664799a68516cb3249a8. Looks good. Changes since previous review were rebase, cleaned up feereason enum/strings, fixed ui target string.

src/policy/fees.h
morcos: done

morcos:
addressed nit

src/policy/fees.h
ryanofsky: Const is actually still here (you removed it from the definition but not declaration)

ryanofsky (commented):
utACK 7c287a7b5df6f33f3b7723496590e72cdd7fed2e. Only change is removed const.

morcos:
oops, actually addressed nit this time.

ryanofsky (commented):
utACK 1bebfc8d3aa615dfdd0221f21b87319e36821b71, const is gone

laanwj:
utACK 1bebfc8

PR #10287, opened 2017-04-27
[tests] Update Unit Test for addrman.h/addrman.cpp

jimmysong:
Add test for adding multiple addresses to address manager
Clean up unnecessary modulo operations
Add test for GetNewBucket's alternate method signature

### DIFF
--- a/src/test/addrman_tests.cpp
+++ b/src/test/addrman_tests.cpp
@@ -91,7 +91,7 @@ BOOST_AUTO_TEST_CASE(addrman_simple)
 
     // Test 2: Does Addrman::Add work as expected.
     CService addr1 = ResolveService("250.1.1.1", 8333);
-    addrman.Add(CAddress(addr1, NODE_NONE), source);
+    BOOST_CHECK(addrman.Add(CAddress(addr1, NODE_NONE), source));
     BOOST_CHECK(addrman.size() == 1);
     CAddrInfo addr_ret1 = addrman.Select();
     BOOST_CHECK(addr_ret1.ToString() == "250.1.1.1:8333");
@@ -99,14 +99,14 @@ BOOST_AUTO_TEST_CASE(addrman_simple)
     // Test 3: Does IP address deduplication work correctly.
     //  Expected dup IP should not be added.
     CService addr1_dup = ResolveService("250.1.1.1", 8333);
-    addrman.Add(CAddress(addr1_dup, NODE_NONE), source);
+    BOOST_CHECK(!addrman.Add(CAddress(addr1_dup, NODE_NONE), source));
     BOOST_CHECK(addrman.size() == 1);
 
 
     // Test 5: New table has one addr and we add a diff addr we should
     //  have two addrs.
     CService addr2 = ResolveService("250.1.1.2", 8333);
-    addrman.Add(CAddress(addr2, NODE_NONE), source);
+    BOOST_CHECK(addrman.Add(CAddress(addr2, NODE_NONE), source));
     BOOST_CHECK(addrman.size() == 2);
 
     // Test 6: AddrMan::Clear() should empty the new table.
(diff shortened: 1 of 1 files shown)


src/test/addrman_tests.cpp
```diff
@@ -398,8 +405,8 @@ BOOST_AUTO_TEST_CASE(addrman_getaddr)
     // Test 25: Ensure GetAddr still returns 23% when addrman has many addrs.
     for (unsigned int i = 1; i < (8 * 256); i++) {
         int octet1 = i % 256;
```
jnewbery: This code isn't at all doing what I expected. On first looking, I assumed it was slicing off 8 bits at a time in order to format a 32 bit int as an ipv4 address, ie I was expecting this:

```
int octet1 = i % 256;
int octet2 = i >> 8 % 256;
int octet3 = i >> 16 % 256;
```

but in fact it's effectively doing this:

```
int octet1 = i % 256;
int octet2 = i >> 8 % 256;
int octet3 = i >> 9 % 256;
```
ie the bits used for octet 2 and 3 are overlapping. Did you work out whether there was a reason for this?

I'd prefer to keep these using the modulos to indicate that each octet should be in the range 0-255. The only reason your suggested change would work is because i never exceeds 2^16. Perhaps replace with the following if there's no good reason to keep the strange octet3 behaviour:

```
int octet1 = i % 256;
int octet2 = i >> 8 % 256;
std::string strAddr = boost::to_string(octet1) + "." + boost::to_string(octet2) + ".0.23";
```
ie eliminate octet3 entirely.

jnewbery (commented):
ACK, with a question about the modulo behaviour in test 25.

src/test/addrman_tests.cpp
jimmysong: I was expecting the same until I tried to refactor it. I can't speak to the original intent, but @EthanHeilman can maybe give us a clue as to what it was?
Anyway it turns out that the addresses are going into a hash table and any adjustment of octet3 in any way changes the numbers in lines 415-417 (vAddr size and addrman size). Thankfully, the hash table is deterministic, but I wanted minimal disruption at least for this PR. I was thinking that this test could be refactored slightly so we can predict the number of collisions in another PR.

EthanHeilman: @jimmysong It just makes it less likely we will have a collision when adding addresses.

Changing
```cpp
int octet3 = (i / (256 * 2)) % 256;
```
to 
```cpp
int octet3 = (i / (256 * 256)) % 256;
```
would cause all of these generated addresses to map to the same source group in the new table as they would all have the same \16. 

Current behavior is intended to lower the chance that addresses are evicted when adding them by spreading them across more than one source group. A byproduct of this is also a more realistic test since the addresses returned by GetAddr() are now being drawn from several source groups.

jnewbery: Can this be achieved by fixing octet2 to zero and then setting octet3 to `i >> 8 % 256`?

EthanHeilman: Yes, that would be a better way of doing it.

jnewbery: Not sure if it's *better*, but I think it would be less confusing :)

EthanHeilman: Less confusing _is_ better. =)

jimmysong:
@EthanHeilman @jnewbery, I've tested a few combinations, and leaving out octet3 works for the current code I have. It's almost exactly the same as before. Let me know what you guys think.

EthanHeilman:
Removing octet3 seems reasonable and the code looks better especially when with the slick BOOST_CHECK_EQUAL.

I've been regretting numbering these tests, e.g. "Test 1: , Test: 2 ...", for a while since adding new tests requires renumbering them or using decimals as done here:
```cpp
// Test 6.5: AddrMan::Add multiple addresses works as expected
```
Could you do a commit which just removes the numbers so "Test 1:" becomes "Test: "?

jimmysong:
@EthanHeilman sure, will work on changing to BOOST_CHECK_EQUALS and taking out test numbers as soon as this one gets merged.

jnewbery:
> Could you do a commit which just removes the numbers so "Test 1:" becomes "Test: "?

Alternatively, you could replace the comments with checkpoints:
http://www.boost.org/doc/libs/1_64_0/libs/test/doc/html/boost_test/test_output/test_tools_support_for_logging/checkpoints.html . I'm not sure what other people would think about this. There aren't `BOOST_TEST_CHECKPOINT`s in the unit tests.

src/test/addrman_tests.cpp
```diff
@@ -398,9 +405,8 @@ BOOST_AUTO_TEST_CASE(addrman_getaddr)
...
-        int octet2 = (i / 256) % 256;
-        int octet3 = (i / (256 * 2)) % 256;
-        std::string strAddr = boost::to_string(octet1) + "." + boost::to_string(octet2) + "." + boost::to_string(octet3) + ".23";
+        int octet2 = i / 256;
```
jnewbery: nit: I still think this is clearer if you bitshift and modulo 256.

jimmysong:
nits addressed, squashed and rebased.

jnewbery:
tested ACK ed36de59e484ecdc22f464948d08a94a8b7fadbb

MarcoFalke:
utACK ed36de59e484ecdc22f464948d08a94a8b7fadbb

PR #10290, opened 2017-04-27
Add -stopatheight for benchmarking

sipa:
This is similar to -stopafterimport, but easier to use for repeated benchmarks.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -436,6 +436,8 @@ std::string HelpMessage(HelpMessageMode mode)
         strUsage += HelpMessageOpt("-dropmessagestest=<n>", "Randomly drop 1 of every <n> network messages");
         strUsage += HelpMessageOpt("-fuzzmessagestest=<n>", "Randomly fuzz 1 of every <n> network messages");
         strUsage += HelpMessageOpt("-stopafterblockimport", strprintf("Stop running after importing blocks from disk (default: %u)", DEFAULT_STOPAFTERBLOCKIMPORT));
+        strUsage += HelpMessageOpt("-stopatheight", strprintf("Stop running after reaching the given height in the main chain (default: %u)", DEFAULT_STOPATHEIGHT));
+
         strUsage += HelpMessageOpt("-limitancestorcount=<n>", strprintf("Do not accept transactions if number of in-mempool ancestors is <n> or more (default: %u)", DEFAULT_ANCESTOR_LIMIT));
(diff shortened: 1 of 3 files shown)


src/validation.cpp
```diff
@@ -2537,6 +2537,9 @@ bool ActivateBestChain(CValidationState &state, const CChainParams& chainparams,
         return false;
     }
 
+    int nStopAtHeight = GetArg("-stopatheight", DEFAULT_STOPATHEIGHT);
+    if (nStopAtHeight && pindexNewTip->nHeight >= nStopAtHeight) StartShutdown();
```
practicalswift: Is `pindexNewTip != NULL` guaranteed here?

fanquake:
Concept ACK. Will test.

src/validation.cpp
kallewoof: It is guaranteed non-null beyond line 2512, which is guaranteed to execute unless the function returns at line 2501 or 2506.

src/init.cpp
```diff
@@ -436,6 +436,8 @@ std::string HelpMessage(HelpMessageMode mode)
         strUsage += HelpMessageOpt("-dropmessagestest=<n>", "Randomly drop 1 of every <n> network messages");
         strUsage += HelpMessageOpt("-fuzzmessagestest=<n>", "Randomly fuzz 1 of every <n> network messages");
         strUsage += HelpMessageOpt("-stopafterblockimport", strprintf("Stop running after importing blocks from disk (default: %u)", DEFAULT_STOPAFTERBLOCKIMPORT));
+        strUsage += HelpMessageOpt("-stopatheight", strprintf("Stop running after reaching the given height in the main chain (default: %u)", DEFAULT_STOPATHEIGHT));
```
jtimon: Perhaps explain that with 0 it won't do anything?
Also, this is not only for the main chain, it will work with testnet3 and regtest right?

jtimon:
utACK b297426c9670fa83a12096f9db5666c15e1bb806

src/validation.cpp
TheBlueMatt: No, I dont believe so, in a crazy case you may have started shutdown (possibly due to stopatheight), and then gotten another call into ActivateBestChain(), which will break out of the main loop at the first chance it gets, and then blow up.

kallewoof: Yes, like 2486. I missed that somehow. `pindexNewTip` is not guaranteed null. Sorry about misleading response.

src/init.cpp
MarcoFalke: It's a debug option, so the user can figure out that -nostopatheight won't do anything?

Issue #10300, opened 2017-04-30
Feature Request (RPC): Specify a txid when calling getrawmempool in verbose mode


dabura667:
`bitcoin-cli getrawmempool true`

Currently this command returns all transactions in the mempool. it would be nice to have an extra arg dependent on verbose mode. (It would have no meaning in non-verbose mode)

ie.:
```
bitcoin-cli getrawmempool true 7dde270e7071667c5d53b073f1d7854319696ea10202e0fb8874c2361dcdd6da
```

Which would only return that single transaction if it is still unconfirmed and return empty if it has already confirmed or does not exist in your mempool.

ie.:
```
{
  "7dde270e7071667c5d53b073f1d7854319696ea10202e0fb8874c2361dcdd6da": {
    "size": 258,
    "fee": 0.00050000,
    "modifiedfee": 0.00050000,
    "time": 1493525525,
    "height": 464139,
    "startingpriority": 0,
    "currentpriority": 0,
    "descendantcount": 1,
    "descendantsize": 258,
    "descendantfees": 50000,
    "ancestorcount": 2,
    "ancestorsize": 516,
    "ancestorfees": 100000,
    "depends": [
      "8fb66a50a714462308fbaff262abe15d85a5823b7e8cbdefff23c2076685373f"
    ]
  }
}
```

Use case: Child Pays For Parent fee calculation by looking at ancestorsize and ancestorfees.

sipa:
Isn't that what `getmempoolentry` does?

paveljanik:
It is. @dabura667 Is your issue solved now? If it is, let's close this issue please.

dabura667:
#10310

This issue will continue on in the above issue.

PR #10304, opened 2017-05-01
[rpc] Allow a txid param in getrawmempool

kallewoof:
See #10300.

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -379,7 +379,7 @@ void entryToJSON(UniValue &info, const CTxMemPoolEntry &e)
     info.push_back(Pair("depends", depends));
 }
 
-UniValue mempoolToJSON(bool fVerbose)
+UniValue mempoolToJSON(bool fVerbose, uint256* txid)
 {
     if (fVerbose)
     {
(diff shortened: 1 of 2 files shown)


jnewbery:
utACK 0300d5e556cc1556b9cba1ec11ff3eef47bddca5.

Can you add a test to exercise the new functionality? Perhaps add some calls to `getrawmempool` with specified txids to the mempool_reorg test. Around here: https://github.com/bitcoin/bitcoin/blob/e2b99b13131b5d0fd6fae6d55892c2ebb327d438/test/functional/mempool_reorg.py#L88

sipa:
I believe this functionality already exists as `getmempoolentry`.

kallewoof:
@sipa Indeed, I had completely missed that one. I'm surprised there's a whole command for it, personally. (And it doesn't seem to be well known as several people (me, issue author, ..) did not realize it, so it may be an idea to mention it in getrawmempool help.)

paveljanik:
Thinking a bit more about it, `gettransaction` should be generalised to include not only in-wallet transactions, but also in-mempool txs. as we have in-wallet, in-blockchain/non-wallet and in-mempool txs (why entries?). But this is backward incompatible change in the API.

sipa:
@paveljanik We have `getrawtransaction` for that, which can query for mempool transactions (and without `-txindex`, that's pretty much the only thing it is useful for).

I disagree that `gettransaction` should be extended. Despite its name, `gettransaction` actually does not return information about a transaction, but is a wallet RPC that returns information about a ledger entry (i.e., the effects a transaction has on your personal balance).

PR #10310, opened 2017-05-02
[doc] Add hint about getmempoolentry to getrawmempool help.

kallewoof:
Lots of people seem to be unaware that getmempoolentry even exists (myself included, before today; see #10300 & #10304), so adding a hint to getrawmempool seems warranted.

I was tempted to start adding "See also: ..." strings to all RPC commands but wasn't sure that was generally desired.

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -413,6 +413,7 @@ UniValue getrawmempool(const JSONRPCRequest& request)
         throw std::runtime_error(
             "getrawmempool ( verbose )\n"
             "\nReturns all transaction ids in memory pool as a json array of string transaction ids.\n"
+            "\nHint: use getmempoolentry to fetch a specific transaction from the mempool.\n"
             "\nArguments:\n"
             "1. verbose (boolean, optional, default=false) True for a json object, false for array of transaction ids\n"
             "\nResult: (for verbose = false):\n"


junderw:
utACK 41bc5ba

TIL

jonasschnelli:
Yes. Adding a hint makes sense.
Though I'm not sure if we should add it below the RPC/curl info. What speaks against mentioning `getmempoolentry` in the description?

instagibbs:
agreed @jonasschnelli I think just putting that in the main body of help is best.

jnewbery:
I agree with the others. The hint should be further up, between the main text and the arguments listing.

Otherwise utACK

kallewoof:
Below is current output:
```
$ ./bitcoin-cli -regtest help getrawmempool
getrawmempool ( verbose )

Returns all transaction ids in memory pool as a json array of string transaction ids.

Hint: use getmempoolentry to fetch a specific transaction from the mempool.

Arguments:
1. verbose (boolean, optional, default=false) True for a json object, false for array of transaction ids

Result: (for verbose = false):
[                     (json array of string)
  "transactionid"     (string) The transaction id
  ,...
]

Result: (for verbose = true):
{                           (json object)
  "transactionid" : {       (json object)
[omitted]
}

Examples:
> bitcoin-cli getrawmempool true
> curl --user myusername --data-binary '{"jsonrpc": "1.0", "id":"curltest", "method": "getrawmempool", "params": [true] }' -H 'content-type: text/plain;' http://127.0.0.1:8332/

$ 
```

instagibbs:
utACK

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/10310/commits/06eb5a79c3e72a8804a54a64663e639034f27657

kallewoof:
Apologies; re-committed verified, same as 06eb5a7.

sdaftuar:
utACK

paveljanik:
reACK https://github.com/bitcoin/bitcoin/pull/10310/commits/3a0a5bc2341ea6a453306f6fe1c065b937294cfb

PR #10311, opened 2017-05-02
Remove unused args from GetFetchFlags()

btcdrak:


dcousens (approved):
utACK

kallewoof (approved):

practicalswift:
utACK

Nice find! Can't believe I didn't catch this one myself :-)

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/10311/commits/1ff2bb4e3e8edfa68797926be8aa3ab5949f2e0c

src/net_processing.cpp
```diff
@@ -1162,7 +1162,7 @@ void static ProcessGetData(CNode* pfrom, const Consensus::Params& consensusParam
     }
 }
 
-uint32_t GetFetchFlags(CNode* pfrom, const CBlockIndex* pprev, const Consensus::Params& chainparams) {
+uint32_t GetFetchFlags(CNode* pfrom) {
```
laanwj: Any idea *why* these were being passed?
Looked back in history and to me it seems that `CBlockIndex* pprev, const Consensus::Params& chainparams` were never used at all, since introduced in b8a97498df1e83f8dcc49bc3fa4344f9e9799242.
@sipa @NicolasDorier  maybe?

paveljanik:
super-micro nit: can you please correct the function name in the commit message?

src/net_processing.cpp
NicolasDorier: Before `GetFetchFlags` was using `IsWitnessEnabled`, not `fHaveWitness`and `GetLocalServices`

sipa:
Yeah, in older versions of the SegWit patch, fetch flags were based on the chain instead of explicit announcements. That led to too many edge cases to deal with, however. The dependency on pprev/params can safely go away.

NicolasDorier:
utACK

laanwj:
@sipa @NicolasDorier thanks for the explanation, so this change makes sense in that context
utACK 1ff2bb4

PR #10318, opened 2017-05-02
[tests] fix wait_for_inv()

jnewbery:
Fixes bug uncovered here: https://github.com/bitcoin/bitcoin/pull/10169#discussion_r114389770

### DIFF
--- a/test/functional/test_framework/mininode.py
+++ b/test/functional/test_framework/mininode.py
@@ -1604,7 +1604,12 @@ class NodeConnCB(object):
         assert wait_until(test_function, timeout=timeout)
 
     def wait_for_inv(self, expected_inv, timeout=60):
-        test_function = lambda: self.last_message.get("inv") and self.last_message["inv"] != expected_inv
+        """Waits for an INV message and checks that the first inv object in the message was as expected."""
+        if len(expected_inv) > 1:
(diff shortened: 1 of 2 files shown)


test/functional/test_framework/mininode.py
```diff
@@ -1604,7 +1604,10 @@ def wait_for_getheaders(self, timeout=60):
         assert wait_until(test_function, timeout=timeout)
 
     def wait_for_inv(self, expected_inv, timeout=60):
-        test_function = lambda: self.last_message.get("inv") and self.last_message["inv"] != expected_inv
+        """ Currently only checks the first inv object in an inv message"""
```
MarcoFalke: In which case you might want to `raise NotImplementedError` if `len(expected_inv) >= 2`?

jnewbery: Great idea. Thanks. Done.

sdaftuar (approved):
Thanks for catching this.  Looks like this was broken in two ways, which is why the test was succeeding in the first place:
- inequality instead of equality, which only didn't cause problems because
- we were comparing a p2p message with its payload (so the inequality condition would trigger, and the wait_until would return)

utACK

luke-jr:
Needs backport?

PR #10327, opened 2017-05-03
[tests] remove import-abort-rescan.py

jnewbery:
Reverts PR 10225

We should back out #10225 and then re-PR it with proper review.


laanwj:
I still think removing it completely instead of temporarily disabling it is over the top, but ok, utACK anyway.

jnewbery:
Thanks! :)

Issue #10247, opened 2017-04-21
cost too many fees?


JokerCatz:
Hi , I send a transaction , but it cost 5,209.456 sat/B ...... & used txin all over 50 confirms

    bitcoin-cli sendtoaddress 1846VsYbYJmqLteuMLdZAYxoVDsjWZyMRz 20

https://blockchain.info/tx/37f98737fe28f1f29840926edeb9cb3a85eb9496070048eb98f2523d612c3cfb

and my getinfo

    {
      "version": 120100,
      "protocolversion": 70012,
      "walletversion": 60000,
      "balance": SKIP,
      "blocks": 462806,
      "timeoffset": 0,
      "connections": 8,
      "proxy": "",
      "difficulty": 520808749422.1398,
      "testnet": false,
      "keypoololdest": 1485144781,
      "keypoolsize": 301,
      "paytxfee": 0.00000000,
      "relayfee": 0.00001000,
      "errors": "Warning: Unknown block versions being mined! It's possible unknown rules are in effect"
    }

& my config

    upnp=0
    server=1
    rpcuser=SKIP
    rpcpassword=SKIP
    rpcthreads=8
    rpckeepalive=false
    rpctimeout=30
    rpcallowip=127.0.0.1
    rpcport=8332
    keypool=300
    mintxfee=0.00000001
    checkblocks=16

did it just version too old & need to upgrade?

jonasschnelli:
Please use something like http://bitcoin.stackexchange.com for future user questions. This is the development issue tracker.

It looks like that your fee estimation spits out a relatively high value.
Try to upgrade to 0.14.0.
Maybe play with `txconfirmtarget=<your desired target in blocks>`.

NicolasDorier:
I think this question have its place here, it might have been a bug.

morcos:
You should definitely upgrade to 0.14.0, but it is still possible although rare for such a result to happen.

Do you know what your transaction confirm target was set to?  The default in 0.12 was changed to 2, but if you are running the GUI, then the GUI setting from a previous version may have persisted.  

This kind of result is much more likely with a target of 1 than with a target of 2.
Personally I would recommend increasing your transaction confirm target to at least 6 blocks.

JokerCatz:
[...]

PR #10333, opened 2017-05-03
[wallet] fee fixes: always create change, adjust value, and p…

instagibbs:
…rune later

This PR is an attempt to solve some of https://github.com/bitcoin/bitcoin/issues/10247 , first and foremost the case:

> // TODO: The case where there is no change output remains
>// to be addressed so we avoid creating too small an output

This is resolved by always assuming a change output exists, and rebalancing that value by the fee excess, and deleting when necessary. The logic is also simpler to me.

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -2597,68 +2597,59 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CWalletT
                 }
 
                 const CAmount nChange = nValueIn - nValueToSelect;
-                if (nChange > 0)
+                assert(nChange >= 0);
+
+                // Fill a vout to ourself, may be removed later on final fee adjustment
+                // TODO: pass in scriptChange instead of reservekey so
+                // change transaction isn't always pay-to-bitcoin-address
+                CScript scriptChange;
+
+                // coin control: send change to custom address
+                if (coinControl && !boost::get<CNoDestination>(&coinControl->destChange)) {
+                    scriptChange = GetScriptForDestination(coinControl->destChange);
+                }
+                // no coin control: send change to newly generated address
+                else
                 {
-                    // Fill a vout to ourself
-                    // TODO: pass in scriptChange instead of reservekey so
-                    // change transaction isn't always pay-to-bitcoin-address
-                    CScript scriptChange;
+                    // Note: We use a new key here to keep it from being obvious which side is the change.
+                    //  The drawback is that by not reusing a previous key, the change may be lost if a
+                    //  backup is restored, if the backup doesn't have the new private key for the change.
+                    //  If we reused the old key, it would be possible to add code to look for and
+                    //  rediscover unknown transactions that were written with keys of ours to recover
+                    //  post-backup change.
 
-                    // coin control: send change to custom address
-                    if (coinControl && !boost::get<CNoDestination>(&coinControl->destChange))
-                        scriptChange = GetScriptForDestination(coinControl->destChange);
-
-                    // no coin control: send change to newly generated address
-                    else
+                    // Reserve a new key pair from key pool
+                    CPubKey vchPubKey;
+                    bool ret;
+                    ret = reservekey.GetReservedKey(vchPubKey, true);
+                    if (!ret)
                     {
-                        // Note: We use a new key here to keep it from being obvious which side is the change.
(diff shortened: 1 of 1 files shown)


instagibbs:
Note: This is a "regression" due to `MIN_FINAL_CHANGE` being taken out of the calculation: https://github.com/bitcoin/bitcoin/pull/10333/files#diff-b2bb174788c7409b671c46ccc86034bdL2647

A bit confusing since the change is dropped when `IsDust` as before... this would be a good time to sync this behavior.

instagibbs:
rebased, and included logic to attempt change larger than `MIN_FINAL_CHANGE` when possible. In the end some cases cannot be avoided, and just-over-dust change may be kept.

instagibbs:
suggested new strategy: instead of "trying one more time" when change is in middle-ground, cache the current transaction, increase the amount you're attempting to send, and try again. If you then run out of available coins, you simply return the cached version of the transaction.

instagibbs:
Implemented the previously mentioned idea. Instead of doing "one more try" to get acceptable change, it caches the successful yet small change transaction and slowly grows the amount of coins it grabs until it creates an acceptable transaction, given reasonable chosen inputs this will result in a transaction with change larger than MIN_FINAL_CHANGE. If it fails and runs out of coins it returns the cached transaction.

src/wallet/wallet.cpp
```diff
@@ -2709,41 +2712,54 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CWalletT
...
+                    else if (IsDust(txNew.vout[nChangePosInOut], ::dustRelayFee)) {
+                        nFeeRet += txNew.vout[nChangePosInOut].nValue;
+                        txNew.vout.erase(txNew.vout.begin() + nChangePosInOut);
+                        nChangePosInOut = -1;
+                        reservekey.ReturnKey();
+                    // If larger than dust, but still small, increase fee target and try again
+                    } else if (txNew.vout[nChangePosInOut].nValue < MIN_FINAL_CHANGE) {
+                        // Save this transaction, use if we cannot get large-enough change
+                        if (!have_cached_txn) {
+                            tx_cached = txNew;
```
achow101: `have_cached_txn` should probably be set to true here as it is never actually set anywhere

src/wallet/wallet.cpp
```diff
@@ -2501,6 +2501,13 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CWalletT
     wtxNew.fTimeReceivedIsTxTime = true;
     wtxNew.BindWallet(this);
     CMutableTransaction txNew;
+    // This transaction is used to track the best transaction that has change
+    // falling between uneconomical dust and MIN_FINAL_CHANGE. If the wallet
+    // runs out of funds trying to find a transaction that has pruneable
+    // change dust or change larger than MIN_FINAL_CHANGE, it will
+    // return this transaction.
+    CMutableTransaction tx_cached;
+    bool have_cached_txn;
```
achow101: Shouldn't you assign `have_cached_txn` to false here as it is not assiged before it is used down below?

src/wallet/wallet.cpp
```diff
@@ -2592,73 +2599,69 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CWalletT
...
-                        std::vector<CTxOut>::iterator position = txNew.vout.begin()+nChangePosInOut;
-                        txNew.vout.insert(position, newTxOut);
-                    }
+                // TODO move this section near fee-adjustment area.
+                if (nChangePosInOut == -1)
+                {
+                    // Insert change txn at random position:
+                    nChangePosInOut = GetRandInt(txNew.vout.size()+1);
+                }
+                else if ((unsigned int)nChangePosInOut > txNew.vout.size())
```
achow101: I don't see how this case matters. Wouldn't `nChangePosInOut` never be set to an out of range index?

instagibbs: Caller may request a too-high change position. This is also a simple code move, not going to change this logic.

instagibbs:
fixed issues, thanks @achow101

gmaxwell (commented):
Concept ACK

morcos:
I think #10712 is preferable to this PR

instagibbs:
closing in favor of #10712

PR #10335, opened 2017-05-04
back-compat: add fallback getentropy implementation

theuni:
Yet another fallback needed for getentropy :(

When building on a newer linux machine (glibc >= 2.25), the resulting binary will fail to run on an older one (<2.25), due to missing symbols. This adds a weak symbol for our own implementation for use as a last resort.

Alternative solutions welcome.

### DIFF
--- a/src/compat/glibc_compat.cpp
+++ b/src/compat/glibc_compat.cpp
@@ -7,6 +7,7 @@
 #endif
 
 #include <cstddef>
+#include <errno.h>
 
 #if defined(HAVE_SYS_SELECT_H)
 #include <sys/select.h>
@@ -27,3 +28,11 @@ extern "C" FDELT_TYPE __fdelt_warn(FDELT_TYPE a)
     return a / __NFDBITS;
 }
 extern "C" FDELT_TYPE __fdelt_chk(FDELT_TYPE) __attribute__((weak, alias("__fdelt_warn")));
+
+/* getentropy was introduced with glibc 2.25 */
+extern "C" int int_getentropy(void *buf, size_t buflen)
+{
+    errno = ENOSYS;
+    return -1;
+}
+extern "C" int getentropy(void *buf, size_t buflen) __attribute__((weak, alias("int_getentropy")));
--- a/src/random.cpp
+++ b/src/random.cpp
@@ -102,12 +102,12 @@ static void RandAddSeedPerfmon()
 #endif
 }
 
-#ifndef WIN32
 /** Fallback: get 32 bytes of system entropy from /dev/urandom. The most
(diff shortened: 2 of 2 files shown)


dcousens:
> Alternative solutions welcome.

`return 0`?
I assume we are already mixing `/dev/urandom` elsewhere though.

sipa:
> Alternative solutions welcome.

How about not using `getentropy` on Linux? It's a BSD compatibility function. The Really Good (TM) solution is using `SYS_getrandom`, but if that's not available, just fallback to using `/dev/urandom` directly.

> I assume we are already mixing /dev/urandom elsewhere though.

No. `getrandom` / `getentropy` replace the need for /dev/urandom.

theuni:
Just had a quick chat about this on IRC. I'll fix this up as discussed.

laanwj:
Huh? we should never be using getrandom() on Linux, that's why the syscall is called directly. What is going wrong here exactly?

theuni:
Updated as discussed with @sipa on IRC.

@laanwj: specifically, the problem is building against a newer glibc, and older linux kernel headers for compatibility. So getentropy() is defined, but the syscall isn't. That's what I ran into here.

Would you like to explicitly disable getentropy for linux on the build side as well?

sipa:
@theuni This commit simplifies the fallback logic a bit: https://github.com/sipa/bitcoin/commit/0abac37a1b3481ed7ab9ea2d0f7b32d819ea9964

Both this PR or just disabling `getentropy` on Linux are fine solutions to me.

laanwj:
> Would you like to explicitly disable getentropy for linux on the build side as well?

I don't see why we'd need it. I made the syscall-based path for Linux to avoid having to rely on any function provided by glibc (e.g. to require a certain glibc version) for kernel randomness. But I don't care deeply, as long as it works...

ryanofsky:
utACK e596faf8efa211b3b17a555beba81a34dc68d55d. Unclear from discussion above whether there is agreement to continue allowing fallback to getentropy on linux, though IMO it makes the code cleaner to fall straight through all the available calls like this PR does (despite the need for the weak symbol workaround).

laanwj:
@theuni can you comment on @sipa's implementation?

theuni:
Looking at this again with fresh eyes, I think the most straightforward is to only enable getentropy for openbsd, where we know what it's doing:
```c++
#elif defined(HAVE_GETENTROPY) && defined(__OpenBSD__) 
```
Then we can drop all of the compat stuff. Sound reasonable?

laanwj:
Yes, sounds good to me.

theuni:
Ok, closing this and opening a new PR as this no long has anything to do with back-compat.

PR #10336, opened 2017-05-04
Get actual path for EUID instead of HOME dir

snvakula:
Let 'bitcoind' run by user 'bitcoin' in Linux.
Chown 'bitcoin-cli' to 'bitcoin:root' and set chmod '4750' (u+s flag).
Then every root-privileged user be able to run 'bitcoin-cli' even if 'bitcoind' runs by user 'bitcoin'.
For example: 'bitcoin-cli getinfo' or 'sudo bitcoin-cli getinfo'

We need to patch 'util.cpp' by this pull request to get actual path for EUID instead of HOME dir, when you use 'u+s' flag.

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -40,6 +40,7 @@
 #include <fcntl.h>
 #include <sys/resource.h>
 #include <sys/stat.h>
+#include <pwd.h>
 
 #else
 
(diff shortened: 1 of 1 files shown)


jonasschnelli:
Not sure about this.
The linux man page states:

`The pw_dir field contains the name of the initial working directory of the user. Login programs use the value of this field to initialize the HOME environment variable for the login shell. [====>] An application that wants to determine its user's home directory should inspect the value of HOME (rather than the value getpwuid(getuid())->pw_dir) since this allows the user to modify their notion of "the home directory" during a login session [<====]. To determine the (initial) home directory of another user, it is necessary to use getpwnam("username")->pw_dir or similar.`

snvakula:
jonasschnelli, when you are root, or run commands by root via sudo, your HOME env is '/root'. So when you try to execute 'sudo bitcoin-cli getinfo' with u+s flag, 'bitcoin-cli' will be ran by user 'bitcoin', but HOME env will be still '/root'. In this case, 'bitcoin-cli' cannot find the bitcoind without current patch.

laanwj:
I'd prefer not to introduce this dependency. Most straightforward way would be to make a script that sets $HOME correctly before calling the tools, or passes `-datadir` or `-conf` options explicitly.

OTOH:

Please don't set-UID `bitcoin-cli`. This **cannot** be guaranteed to be secure, there may be lots of ways to get the tool to do something with the permissions that it shouldn't.

There's also no need to:  `bitcoind` and `bitcoin-cli` can run as different users. Communication happens over TCP localhost, which all users on the local system can access. They just need the RPC credentials.

My suggestion would be to put bitcoin.conf in a place that's accessible to the users that need access.

laanwj:
Closing this for above reasons (and lack of other feedback)

PR #10341, opened 2017-05-05
rpc/wallet: Workaround older UniValue which returns a std::string temporary for get_str

luke-jr:
Without this, building against UniValue 1.0.2 results in:

```
wallet/rpcwallet.cpp: In function ‘UniValue getbalance(const JSONRPCRequest&)’:
wallet/rpcwallet.cpp:732:98: error: taking address of temporary [-fpermissive]
     const std::string* account = request.params[0].get_str() != "*" ? &request.params[0].get_str() : nullptr;
```

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -729,7 +729,8 @@ UniValue getbalance(const JSONRPCRequest& request)
     if (request.params.size() == 0)
         return  ValueFromAmount(pwallet->GetBalance());
 
-    const std::string* account = request.params[0].get_str() != "*" ? &request.params[0].get_str() : nullptr;
+    const std::string& account_param = request.params[0].get_str();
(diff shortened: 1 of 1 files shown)


ryanofsky:
utACK a637734ba22307f6e89857803070f566e736d4a6

jonasschnelli:
utACK a637734ba22307f6e89857803070f566e736d4a6

paveljanik:
utACK a637734

theuni:
This points out that we don't have a strong enough api contract to support an external UniValue :\

gmaxwell:
I recommend we instead threshold the univalue version.  They have been security issues in older versions... esp since as a library I don't believe it's actually maintained.

jgarzik:
It is maintained.

laanwj:
This makes the code slightly better readable so seems OK to me.

A minimum threshold for univalue would be a sane thing to do though. I didn't know we supported external univalue at all.

PR #10342, opened 2017-05-05
[tests] Improve mempool_persist test

jnewbery:
mempool_persist.py was failing very intermittently because after starting the node, there was a race between calling getrawmempool and the mempool being reloaded by a background thread.

This PR removes that raciness in the test. It also removes the is_network_split parameter and tidies up the start of the test (this test was written between #10198 being opened and merged)

### DIFF
--- a/test/functional/mempool_persist.py
+++ b/test/functional/mempool_persist.py
@@ -6,11 +6,11 @@
 
 By default, bitcoind will dump mempool on shutdown and
 then reload it on startup. This can be overridden with
-the -persistmempool=false command line option.
+the -persistmempool=0 command line option.
 
 Test is as follows:
 
-  - start node0, node1 and node2. node1 has -persistmempool=false
+  - start node0, node1 and node2. node1 has -persistmempool=0
   - create 5 transactions on node2 to its own address. Note that these
     are not sent to node0 or node1 addresses because we don't want
     them to be saved in the wallet.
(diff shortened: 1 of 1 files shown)


test/functional/mempool_persist.py
```diff
@@ -72,20 +66,24 @@ def run_test(self):
         self.nodes = []
         self.nodes.append(start_node(0, self.options.tmpdir))
         self.nodes.append(start_node(1, self.options.tmpdir))
-        assert_equal(len(self.nodes[0].getrawmempool()), 5)
-        assert_equal(len(self.nodes[1].getrawmempool()), 0)
+        # Give bitcoind a second to reload the mempool
+        time.sleep(1)
+        assert wait_until(lambda: len(self.nodes[0].getrawmempool()) == 5)
+        assert wait_until(lambda: len(self.nodes[1].getrawmempool()) == 0)
```
sdaftuar: Does the `wait_until` accomplish anything here, or at line 80 below?  Either the mempool will be empty when this RPC is first invoked, in which case the wait_until will immediately return, or else it will have something -- in which case waiting and then failing later doesn't seem useful?

jnewbery: No - this could just as well be a straight assert that node1's mempool is empty. The `time.sleep(1)` above is what actually guarantees that we're not just hitting a window before node1's mempool fills up.

I changed this to a `wait_until()` call for consistency with the line above, but could quite easily change it back if you think that would be clearer for the reader.

test/functional/mempool_persist.py
```diff
@@ -30,26 +30,20 @@
...
-    def setup_network(self):
-        # We need 3 nodes for this test. Node1 does not have a persistent mempool.
-        self.nodes = []
-        self.nodes.append(start_node(0, self.options.tmpdir))
-        self.nodes.append(start_node(1, self.options.tmpdir, ["-persistmempool=false"]))
-        self.nodes.append(start_node(2, self.options.tmpdir))
-        connect_nodes_bi(self.nodes, 0, 2)
-        connect_nodes_bi(self.nodes, 1, 2)
-        self.is_network_split = False
+        self.extra_args = [[], ["-persistmempool=false"], []]
```
MarcoFalke: Our parser does not support string representations of booleans. You should only use integers here...

MarcoFalke:
```vim
:%s/false/0/g

maflcko (commented):
utACK after global substitution.

test/functional/mempool_persist.py
sdaftuar: Yeah imo it would be clearer without the `wait_until`.

jnewbery:
Added fixup commit:

- replaces `persistmempool=false` with `-persistmempool=0`
- reverts the `assert wait_until(lambda: len(self.nodes[1].getrawmempool()) == 0)` to `assert_equal(len(self.nodes[1].getrawmempool()), 0)`
- adds a `sync_all()` call after `self.nodes[0].generate(1)` to ensure that we really are out of IBD.

Will squash once this is ACKed and ready for merge.

sdaftuar:
ACK

MarcoFalke:
utACK 6658e7fc4f3117beb82d02d71f2de2f0500d37c1, please squash.

jnewbery:
squashed

PR #10343, opened 2017-05-05
Remove redundant on-the-same-line-repetition of type names (DRY): RepeatedTypeName foo = static_cast<RepeatedTypeName>(bar)

practicalswift:
Use …

```c++
auto foo = static_cast<NonRepeatedTypeName>(bar);
```

… instead of …

```c++
RepeatedTypeName foo = static_cast<RepeatedTypeName>(bar);
```

See the [C++ Core Guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es11-use-auto-to-avoid-redundant-repetition-of-type-names) for a general discussion on the use of `auto` to avoid redundant repetition of type names.

### DIFF
--- a/src/bitcoin-cli.cpp
+++ b/src/bitcoin-cli.cpp
@@ -156,7 +156,7 @@ const char *http_errorstring(int code)
 
 static void http_request_done(struct evhttp_request *req, void *ctx)
 {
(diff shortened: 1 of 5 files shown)


practicalswift:
Closing due to clear lack of ACK consensus :-)

PR #10346, opened 2017-05-05
Use range-based for loops (C++11) when looping over vector elements

practicalswift:


src/wallet/wallet.cpp
```diff
@@ -1781,8 +1781,8 @@ bool CWalletTx::IsEquivalentTo(const CWalletTx& _tx) const
 {
         CMutableTransaction tx1 = *this->tx;
         CMutableTransaction tx2 = *_tx.tx;
-        for (unsigned int i = 0; i < tx1.vin.size(); i++) tx1.vin[i].scriptSig = CScript();
-        for (unsigned int i = 0; i < tx2.vin.size(); i++) tx2.vin[i].scriptSig = CScript();
+        for (auto txIn : tx1.vin) txIn.scriptSig = CScript();
+        for (auto txIn : tx2.vin) txIn.scriptSig = CScript();
```
ryanofsky: These need to be `auto&` to have an an effect. Probably all the `auto` declarations here should be replaced by `auto&` or `const auto&`

practicalswift: Was a bit too fast with submitting this - sorry about that. Closed this PR and cannot re-open. New correct PR submitted.

PR #10347, opened 2017-05-05
Use range-based for loops (C++11) when looping over vector elements

practicalswift:


src/script/interpreter.cpp
```diff
@@ -1142,24 +1142,24 @@ class CTransactionSignatureSerializer {
 
 uint256 GetPrevoutHash(const CTransaction& txTo) {
     CHashWriter ss(SER_GETHASH, 0);
-    for (unsigned int n = 0; n < txTo.vin.size(); n++) {
-        ss << txTo.vin[n].prevout;
+    for (const auto& txIn : txTo.vin) {
```
luke-jr: `txIn` indicates a `CTransaction` type. Better to make it lowercase `txin`.

luke-jr (approved):
utACK, with one minor nit.

src/wallet/wallet.cpp
```diff
@@ -1781,8 +1781,8 @@ bool CWalletTx::IsEquivalentTo(const CWalletTx& _tx) const
 {
         CMutableTransaction tx1 = *this->tx;
         CMutableTransaction tx2 = *_tx.tx;
-        for (unsigned int i = 0; i < tx1.vin.size(); i++) tx1.vin[i].scriptSig = CScript();
-        for (unsigned int i = 0; i < tx2.vin.size(); i++) tx2.vin[i].scriptSig = CScript();
+        for (const auto& txIn : tx1.vin) txIn.scriptSig = CScript();
```
ryanofsky: Needs to be non-const

src/script/interpreter.cpp
practicalswift: @luke-jr Likely I'm missing something here, but `txTo.vin` is `std::vector<CTxIn>` and not `CTransaction`. Shouldn't `txIn` then indicate `CTxIn`. What am I missing? :-)

luke-jr: Hungarian notation uses the lowercase type identifier (eg, "tx") followed by camel case variable name (eg, "In"). While it's not required to use this notation, your "txIn" variable name *looks* like it  is, while not matching the usage expectation.

practicalswift: @luke-jr I think I get it now. `CTransactionRef &txOut` and `CTransactionRef& txIn` if used in other parts of the code base. The suggestions is that I change all `txIn`/`txOut` to `txin`/`txout` in my PR?

luke-jr: Right. Might also change `transactionRef` to simply `tx` also, but it's less confusing.

practicalswift:
@luke-jr Nit addressed! Looks good? :-)

src/wallet/wallet.cpp
practicalswift: @ryanofsky Thanks! Fixed!

kallewoof (approved):

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/10347/commits/660b5237c4f7a4fb216ae5c9207e90156f843ea0

src/validation.cpp
```diff
@@ -3058,8 +3058,8 @@ bool ContextualCheckBlock(const CBlock& block, CValidationState& state, const Co
 
     // No witness data is allowed in blocks that don't commit to witness data, as this would otherwise leave room for spam
     if (!fHaveWitness) {
-        for (size_t i = 0; i < block.vtx.size(); i++) {
-            if (block.vtx[i]->HasWitness()) {
+      for (const auto& transactionRef : block.vtx) {
```
sipa: For consistency I would call this just `tx`.

src/wallet/wallet.cpp
```diff
@@ -2149,10 +2149,10 @@ bool CWallet::SelectCoinsMinConf(const CAmount& nTargetValue, const int nConfMin
 
     if (nTotalLower == nTargetValue)
     {
-        for (unsigned int i = 0; i < vValue.size(); ++i)
+        for (const auto& inputCoin : vValue)
```
sipa: Simply call this `input` or `coin`, to avoid making it look like Hungarian notation for a transaction input?

sipa (commented):
utACK, two nits

practicalswift:
@sipa Good points! Nits addressed.

Thanks for reviewing.

sipa:
utACK 83e8c9207733ad5d3013ee3009632514b15a9347

sipa:
Needs rebase.

practicalswift:
@sipa Thanks for the ping. Now rebased! :-)

practicalswift:
Rebased again!

ryanofsky (commented):
utACK 211adc074a662505f2b54f3f2755f4fefc167aac

PR #10351, opened 2017-05-07
removed unused code in INV message

Greg-Griffith:
vToFetch is never used after declaration within the INV message. When checked if not empty,
evaluation is always false. Best case scenario this is optimized by the
compiler, worst case it wastes cpu cycles.  It should be removed either
way.

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -1552,8 +1552,6 @@ bool static ProcessMessage(CNode* pfrom, const std::string& strCommand, CDataStr
 
         uint32_t nFetchFlags = GetFetchFlags(pfrom);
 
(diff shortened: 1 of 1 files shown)


laanwj:
Good catch, seems that this has been unused since #8872 / 037159cebf1eae4445050cec029986514ed4e9e2 (Remove block-request logic from INV message processing).

sipa:
utACK c707ca872dcce98892431d33c250372da714f15f

MarcoFalke:
utACK c707ca872dcce98892431d33c250372da714f15f

paveljanik:
ACK c707ca8

dcousens (approved):

PR #10353, opened 2017-05-08
Refactor script interpreter flags to use new CScriptFlags class for stricter checking

luke-jr:
I noticed that we've run out of bits in script flags, so I refactored the code so flags are an explicit class and cause compile failures instead of silent bugs, and can be safely extended to larger sizes. IMO it is also more readable, especially in regard to `IsAtLeastOneSet`.


jtimon:
Needs rebase

dcousens:
concept ACK

laanwj:
Closing, seems to be dormant, let me know if you pick this up again.

PR #10359, opened 2017-05-08
[tests] functional tests should call BitcoinTestFramework start/stop node methods

jnewbery:
Second step towards #10082

This PR changes the individual test cases to call the `BitcoinTestFramework.start_node` and `.stop_node` methods rather than the util.py functions. This is in preparation for moving the start/stop node functionality into BitcoinTestFramework (currently the BitcoinTestFramework start/stop node methods are just wrappers for the util.py functions).

The second commit changes the names of the functions in util.py to mark them as private. This is so no test cases are added/modified to call the functions directly before the next step of #10082 is merged.

### DIFF
--- a/test/functional/abandonconflict.py
+++ b/test/functional/abandonconflict.py
@@ -73,8 +73,8 @@ class AbandonConflictTest(BitcoinTestFramework):
 
         # Restart the node with a higher min relay fee so the parent tx is no longer in mempool
         # TODO: redo with eviction
-        stop_node(self.nodes[0],0)
-        self.nodes[0]=start_node(0, self.options.tmpdir, ["-minrelaytxfee=0.0001"])
+        self.stop_node(0)
+        self.nodes[0] = self.start_node(0, self.options.tmpdir, ["-minrelaytxfee=0.0001"])
 
         # Verify txs no longer in either node's mempool
         assert_equal(len(self.nodes[0].getrawmempool()), 0)
@@ -100,8 +100,8 @@ class AbandonConflictTest(BitcoinTestFramework):
         balance = newbalance
 
         # Verify that even with a low min relay fee, the tx is not reaccepted from wallet on startup once abandoned
-        stop_node(self.nodes[0],0)
-        self.nodes[0]=start_node(0, self.options.tmpdir, ["-minrelaytxfee=0.00001"])
+        self.stop_node(0)
+        self.nodes[0] = self.start_node(0, self.options.tmpdir, ["-minrelaytxfee=0.00001"])
         assert_equal(len(self.nodes[0].getrawmempool()), 0)
         assert_equal(self.nodes[0].getbalance(), balance)
 
@@ -120,8 +120,8 @@ class AbandonConflictTest(BitcoinTestFramework):
         balance = newbalance
 
         # Remove using high relay fee again
-        stop_node(self.nodes[0],0)
-        self.nodes[0]=start_node(0, self.options.tmpdir, ["-minrelaytxfee=0.0001"])
+        self.stop_node(0)
+        self.nodes[0] = self.start_node(0, self.options.tmpdir, ["-minrelaytxfee=0.0001"])
         assert_equal(len(self.nodes[0].getrawmempool()), 0)
         newbalance = self.nodes[0].getbalance()
         assert_equal(newbalance, balance - Decimal("24.9996"))
--- a/test/functional/blockchain.py
+++ b/test/functional/blockchain.py
@@ -24,7 +24,6 @@ from test_framework.util import (
     assert_raises_jsonrpc,
     assert_is_hex_string,
     assert_is_hash_string,
-    connect_nodes_bi,
 )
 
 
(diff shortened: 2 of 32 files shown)


MarcoFalke:
Concept ACK. Needs rebase.

jnewbery:
rebased

jnewbery:
rebased. It'd be really helpful to get this merged so we can make some progress on #10082 . This is the only disruptive changeset in that larger PR - the rest of the changes are restricted to the test_framework directory.

This PR touches a lot of files, but it should be a trivial review. It's simply changing calls from `stop_node()` to calls to `self.stop_node()` etc wrapper functions in the first commit. The second commit renames the inner functions so no-one calls them explicitly from the test cases (the inner functions will be removed entirely in a future PR).

@MarcoFalke , @TheBlueMatt , @ryanofsky , @sdaftuar - if I could convince any of you to review, I'd be very grateful!

test/functional/abandonconflict.py
```diff
@@ -73,8 +73,8 @@ def run_test(self):
 
         # Restart the node with a higher min relay fee so the parent tx is no longer in mempool
         # TODO: redo with eviction
-        stop_node(self.nodes[0],0)
-        self.nodes[0]=start_node(0, self.options.tmpdir, ["-minrelaytxfee=0.0001"])
+        self.stop_node(0)
+        self.nodes[0] = self.start_node(0, self.options.tmpdir, ["-minrelaytxfee=0.0001"])
```
jimmysong: The plan is to change `start_node` so assignment here would be unnecessary in the future?

test/functional/blockchain.py
```diff
@@ -24,6 +24,7 @@
     assert_raises_jsonrpc,
     assert_is_hex_string,
     assert_is_hash_string,
+    connect_nodes_bi,
```
jimmysong: Not sure why this was added?

ryanofsky: Change seems unrelated, maybe revert.

test/functional/fundrawtransaction.py
```diff
@@ -449,12 +449,12 @@ def run_test(self):
...
+        self.stop_node(0)
+        self.stop_node(2)
+        self.stop_node(3)
         self.nodes.pop(1)
-        stop_node(self.nodes[0], 0)
-        stop_node(self.nodes[1], 2)
-        stop_node(self.nodes[2], 3)
 
-        self.nodes = start_nodes(self.num_nodes, self.options.tmpdir)
+        self.nodes = self.start_nodes(self.num_nodes, self.options.tmpdir)
```
jimmysong: I assume that `self.start_nodes` would also do the assignment in the future?

test/functional/net.py
```diff
@@ -14,8 +14,7 @@
     assert_equal,
     assert_raises_jsonrpc,
     connect_nodes_bi,
-    p2p_port,
-)
+    p2p_port)
```
jimmysong: Nit: I personally like comma endings so adding new imports has cleaner diffs.

test/functional/fundrawtransaction.py
```diff
@@ -449,12 +449,12 @@ def run_test(self):
         ############################################################
         # locked wallet test
         self.nodes[1].encryptwallet("test")
+        self.stop_node(0)
```
ryanofsky: Maybe move these up one line so nodes[1].encrypt and pop(1) can be next to each other.

ryanofsky (commented):
Slightly tested ACK a433d8a15e165cfb42e44319d608365603f64e0c

jimmysong (approved):
ACK. testing now.

jimmysong:
Tested, ACK a433d8a15e165cfb42e44319d608365603f64e0c

A couple of really small nits. Excited to keep this moving!

test/functional/fundrawtransaction.py
kallewoof: Unless it should be possible to append nodes.

test/functional/net.py
kallewoof: Agreed

kallewoof (approved):
utACK a433d8a15e165cfb42e44319d608365603f64e0c

test/functional/blockchain.py
jnewbery: yes. not sure how that snuck in. Reverted.

test/functional/abandonconflict.py
jnewbery: yes - that's certainly something we can do in a future PR once the BitcoinTestFramework class is responsible for tracking and stop/starting the nodes.

test/functional/blockchain.py
jnewbery: me neither :flushed:. Removed

test/functional/fundrawtransaction.py
jnewbery: certainly something we can do in a future PR.

BitcoinTestFramework could include two functions - `start_nodes()` to start the nodes in `self.nodes`, and `add_nodes()` if you need to append nodes to that list.

test/functional/net.py
jnewbery: Sure. Reverted.

jnewbery:
Thanks @ryanofsky @jimmysong @kallewoof . Nits fixed in fixup commit. Will squash when this is ready for merge.

MarcoFalke:
utACK 53f6775

Issue #7664, opened 2016-03-09
Revisit Coin Selection


morcos:
There is concern that the current coin selection algorithm is too biased towards creating small inputs.  We should try to look at some data after the change for 0.12 and see if that is actually the case and regardless work at improving the coin selection algorithm to avoid that concern before the 0.13 release.

See https://github.com/bitcoin/bitcoin/issues/7657 and #4906 for back story.

Tag with 0.13 milestone please

jonasschnelli:
Agree with @morcos.
Something we might want to work towards is a flexible/modular Coinselection. Coinselection somehow depends on the use cases and could be modular therefore.

We could allow different coin selection classes (some OO magic / interfacing) and allow to select them in `fundrawtransaction` or over a global state only adjustable on startup `-coinselection=*class*`.

Bitcoinj has a similar approach.

gmaxwell:
In my experience the biggest impediment to doing anything with coin selection is the current testing approach that simply fail if anything changes at all. So when you intend to change it you just get a pile of failures and nothing that tells you if the change is behaving sanely.

I don't see any problem to having a selection, but I'm somewhat doubtful of the utility. Having a selection doesn't replace having a _good_ default-- and right now we don't even have a good default; and most of the time the user actually doesn't really much useful information here (nothing the user would call a "use case" probably directly translates into the right strategy)-- keep in mind that most users have no idea about the underlying "coins" behavior in the system.

Perhaps the primary strategy distinguisher would be the importance of privacy-- e.g. should sweeping be kept to "do no privacy harm" levels, or more aggressive-- but the user is often poorly equipt to weigh the privacy considerations (e.g. don't what their future risks are, and may not know how powerful transaction analysis is), nor are they usually able to weigh the privacy impacts on their counterparties or the ecosystem as a whole.

Another thing to keep in mind is that implementation of GROUPCHECKSIGVERIFY would make strategies that try to spend many inputs at once much more attractive. (beyond the argument I've made before that you generally want to pay fees sooner rather than later, since fees tend to go up over time)

laanwj:
#1643 should also be mentioned here. Bad coin selection with lots of small outputs has been a long-running issue.

MarcoFalke:
> which generally minimizes the input set

[...]

PR #10360, opened 2017-05-08
[WIP] [Wallet] Target effective value during transaction creation

instagibbs:
Previous fee-targeting behavior lead to needless looping, excessive fees, and surprising behavior. 

This PR changes the "fee targeting" algorithm by considering "effective value" of considered inputs instead of simply trying to hit an absolute fee, seeing if it failed, then trying again with the estimated total fee at the end of the loop.

The algorithm also doesn't select any coin with non-positive effective value.

In short:
`effectiveValue = nValue - feeRate*num_bytes_for_signed_input`

See https://github.com/bitcoin/bitcoin/issues/10247 , https://github.com/bitcoin/bitcoin/issues/7664 

To do:
1) Previously the wallet just kept stuffing fees until it succeeded. In this PR there is no strict sanity checks on the loop, so if there is some error where we are just shy and cannot re-balance change output to pay for it, it will look forever(I have no idea if this is possible since we should be over-estimating at worst). We should probably have something like https://github.com/bitcoin/bitcoin/pull/10333 for re-balancing, and then a sanity check to error out if a specified feerate cannot be hit for some reason post-coin selection, perhaps after a fixed number of tries. 
2) This breaks the `coinControl->nMinimumTotalFee` option. I can't conceive a use for it, so I want to remove this anyways.  https://github.com/bitcoin/bitcoin/pull/10390

### DIFF
--- a/src/bench/coin_selection.cpp
+++ b/src/bench/coin_selection.cpp
@@ -39,7 +39,7 @@ static void CoinSelection(benchmark::State& state)
 
     while (state.KeepRunning()) {
         // Empty wallet.
-        BOOST_FOREACH (COutput output, vCoins)
+        for (auto& output : vCoins)
             delete output.tx;
         vCoins.clear();
 
@@ -50,7 +50,7 @@ static void CoinSelection(benchmark::State& state)
 
         std::set<CInputCoin> setCoinsRet;
         CAmount nValueRet;
-        bool success = wallet.SelectCoinsMinConf(1003 * COIN, 1, 6, 0, vCoins, setCoinsRet, nValueRet);
+        bool success = wallet.SelectCoinsMinConf(1003 * COIN, 1, 6, 0, vCoins, setCoinsRet, nValueRet, CFeeRate(0));
         assert(success);
         assert(nValueRet == 1003 * COIN);
         assert(setCoinsRet.size() == 2);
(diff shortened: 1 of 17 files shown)


instagibbs:
Getting travis `RPC_WALLET_INSUFFICIENT_FUNDS` errors all over the place for one configuration. Can't reproduce locally...

morcos:
sort of concept ACK

I think there are a lot of issues to think about in order to make sure we don't accidentally introduce poor behavior.
One simple improvement would be to cache the number of bytes each input would require to spend to create a much simpler calculation of effective value each time through instead of dummy signing an ever growing virtual transaction.

But I'm more concerned with the interaction between the knapsack solver and the effective value inputs.  Since the knapsack solver aims for exact matches (either with or without change), I believe there will be a tendency to often include many very small effective inputs to try to get as close as possible to an arbitrary change number.

instagibbs:
>One simple improvement would be to cache the number of bytes each input would require to spend to create a much simpler calculation of effective value each time through instead of dummy signing an ever growing virtual transaction.

Yes, some additional refactoring would make sense especially for this. This was probably the most janky part of my implementation.

>I believe there will be a tendency to often include many very small effective inputs to try to get as close as possible to an arbitrary change number.

As we discussed on IRC, Core already does this, but you mean that previously slightly larger inputs would now fall under this bucket so it may pack even more. It's hard to say the net effect especially since we disallow negative effective value inputs, but something to think about. OTOH If we wait until everything is an unambiguous win, I am not sure any changes will get merged.

> some chatter about general refactoring on IRC

I really don't mind attempting to tackle more refactoring within this PR. I admittedly did just over minimal refactoring to make it work.

instagibbs:
@gmaxwell notes on IRC that we may want to consider unconfirmed change as smaller effective value by having it also need to pay for the ancestor(s) feerate delta to current confirmation target feerate.

instagibbs:
closing until BnB PR has been merged, then will revisit our transaction creation strategy as a whole.

PR #10361, opened 2017-05-08
qa: disablewallet: Check that wallet is really disabled

maflcko:
This test passes on current master even if the wallet is *not* disabled.


jnewbery:
Tested ACK fa7396d6d9adafdf1118f773a7f5cad04f31f734

PR #10366, opened 2017-05-09
[test] Remove all * imports

kallewoof:
Another take at #9876 (I didn't realize it existed, but it's closed and I really do think we should get rid of `import *`s if possible!).

Main difference that I can see: each commit corresponds to one file.

I attempted to pull imported imports out as much as possible, e.g. instead of importing `Decimal` from `util.py`, it is now imported directly from `decimal`. I may have missed some of these.

Options, please advice:
1. Discard this PR as redundant in favor of outcome in discussion in #9876 (`__all__`) -- personally I don't think this approach is ideal, as it breeds bad behavior from example (`import *` *is* considered bad behavior, even if `__all__` makes it more controllable) -- or in favor of reopening #9876.
2. Split into multiple PR's to spread impact out over time, e.g. "`p2p-*`, `mempool_*`, ...".
3. Take as is or squashed in some fashion (`p2p-*`, `mempool_*`, ...)

### DIFF
--- a/test/functional/abandonconflict.py
+++ b/test/functional/abandonconflict.py
@@ -11,7 +11,14 @@
  no effect on transactions which are already conflicted or abandoned.
 """
 from test_framework.test_framework import BitcoinTestFramework
-from test_framework.util import *
+from test_framework.util import (
+    assert_equal,
+    connect_nodes,
+    Decimal,
+    disconnect_nodes,
+    sync_blocks,
+    sync_mempools,
+)
 
 class AbandonConflictTest(BitcoinTestFramework):
     def __init__(self):
--- a/test/functional/bip65-cltv-p2p.py
+++ b/test/functional/bip65-cltv-p2p.py
@@ -17,7 +17,7 @@ Mine 1 old version block, see that the node rejects.
 """
 
 from test_framework.test_framework import ComparisonTestFramework
-from test_framework.util import *
+from test_framework.util import hex_str_to_bytes
 from test_framework.mininode import CTransaction, NetworkThread
 from test_framework.blocktools import create_coinbase, create_block
 from test_framework.comptool import TestInstance, TestManager
--- a/test/functional/bip65-cltv.py
+++ b/test/functional/bip65-cltv.py
@@ -5,7 +5,7 @@
 """Test the CHECKLOCKTIMEVERIFY (BIP65) soft-fork logic."""
 
 from test_framework.test_framework import BitcoinTestFramework
-from test_framework.util import *
+from test_framework.util import assert_raises_jsonrpc, connect_nodes
 
 class BIP65Test(BitcoinTestFramework):
     def __init__(self):
--- a/test/functional/bip68-112-113-p2p.py
+++ b/test/functional/bip68-112-113-p2p.py
@@ -44,11 +44,20 @@ bip112tx_special - test negative argument to OP_CSV
 """
 
 from test_framework.test_framework import ComparisonTestFramework
-from test_framework.util import *
+from test_framework.util import (
+    assert_equal,
+    Decimal,
+    get_bip9_status,
+    hex_str_to_bytes,
+)
 from test_framework.mininode import ToHex, CTransaction, NetworkThread
 from test_framework.blocktools import create_coinbase, create_block
 from test_framework.comptool import TestInstance, TestManager
-from test_framework.script import *
+from test_framework.script import (
+    CScript,
+    OP_CHECKSEQUENCEVERIFY,
+    OP_DROP,
+)
 from io import BytesIO
 import time
 
--- a/test/functional/bip68-sequence.py
+++ b/test/functional/bip68-sequence.py
@@ -4,9 +4,27 @@
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 """Test BIP68 implementation."""
 
+import time
+
 from test_framework.test_framework import BitcoinTestFramework
-from test_framework.util import *
-from test_framework.blocktools import *
+from test_framework.util import (
+    assert_equal,
+    assert_raises_jsonrpc,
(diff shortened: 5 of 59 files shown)


practicalswift:
utACK 6e43829

Very good! This will make Python linting easier!

sdaftuar:
I still question this goal, because I think having to (for example) manually add each script opcode as an import when you want to use one in a test is needlessly time-wasting for test writers, but for now I'd like to make the more important observation that merging this will cause substantive PRs to all need to be rebased (punishment for actually including or updating a test!).

For reference, my earlier comments on this idea: https://github.com/bitcoin/bitcoin/pull/9876#issuecomment-283057363

practicalswift:
@sdaftuar Do you have any examples of substantive PRs that would need to be rebased after merging this one?

MarcoFalke:
> This will make Python linting easier!

Even though I can understand the desire to have a single style or converge
the whole python code base to pep8, I tend to object based on the
maintenance overhead. The python syntax is already rather strict and
currently we should not make it more strict without clear reasoning. The
thread Suhas liked to, explains why wildcards could be useful to write
tests more productively and with less hassle.

Please note that we don't have a general coding guide for python code at
this point and refactoring for the sake of pep8 without previous discussion
might cause more hassle than it is worth.

So Concept NACK on this pull for now, as it would conflict with the ongoing
test_framework refactoring. More precisely, it would go in the wrong
direction, when i see `+ stop_node` in the diff, which is planned to be
made a private method of the test framework, so should not be added to the
list of imports.

kallewoof:
@sdaftuar I have to respectfully disagree. Adding opcodes as you use them makes perfect sense to me. It also gives you an overview of what opcodes are actually being checked, as an added bonus, just by looking at the imports and not the tests themselves (which can be a bit long-winded at times, unfortunately). I also have a hard time believing this PR would cause conflicts in a lot of test PRs, and I am fairly sure most of the test developers agree that * imports are just generally bad.

@MarcoFalke It's no so much about a style but about proper development. Importing * means you lose control over what is going where, just like rampant usage of using namespace xyz in C++. In core proper it can be flat out dangerous and lead to unpredictable behavior with version upgrades, which may not necessarily be the case with functional tests, but it's still a step in the right direction.

MarcoFalke:
I agree it is a step in the right direction, but right now is not a good
time to do global refactors that stall progress of other pulls. I think we
should wait until less refactoring happens in the core files of the test
framework (such as https://github.com/bitcoin/bitcoin/pull/10359/files#diff-
bf6de4c07d2e9114c8f01eea45c6cab3L11) and revisit this pull after the dust
has settled.

practicalswift:
@MarcoFalke When mentioning the linting improvements I was specifically thinking about finding unused imports (see below on why that matters), not full PEP-8 compliance :-)

Explicit imports in tests make it easier to get a quick feeling for what a specific test case covers. And perhaps more importantly, by using explicit imports while at the same time removing unused imports it makes it very clear what is **not** covered by a specific test case (not imported ⇔ not tested).

Strong concept ACK!

MarcoFalke:
The tests in /functional/ are not supposed to test the test framework,
but the bitcoind/qt api and internals, so test coverage can not be
inferred from which utility methods are (or are not) imported from the
test framework.

> by using explicit imports while at the same time removing unused imports

As mentioned earlier, removing unused imports is a cleanup that should
happen every couple of months, not on a regular basis... Another
reason against explicit imports is that people will start opening
daily pulls to sort all of those alphabetically (see #10302)

practicalswift:
Let's agree that sorting imports alphabetically daily is a waste of time :-)

ryanofsky:
I think this is a good change because `import *` can make unfamiliar python code really hard to read (you have no idea where identifiers are coming from).

I also think @sdaftuar has a good point about script opcodes, and if there could be a solution for referencing opcodes that doesn't rely on `import *` before making this change, that would also make this PR smaller and a little easier to review. Maybe the solution could be defining an `Op` class and replacing `OP_RETURN` with `Op.RETURN` etc. Or just accepting strings like "OP_RETURN".

kallewoof:
@ryanofsky There is the
```Python
import test_framework.script as script
[...]
        p2sh_program = CScript([script.OP_TRUE])
[...]
```
approach. A bit verbose perhaps, but better than import *.

kallewoof:
@MarcoFalke 
> I agree it is a step in the right direction, but right now is not a good
time to do global refactors that stall progress of other pulls. I think we
should wait until less refactoring happens in the core files of the test
framework (such as https://github.com/bitcoin/bitcoin/pull/10359/files#diff-
bf6de4c07d2e9114c8f01eea45c6cab3L11) and revisit this pull after the dust
has settled.

I don't see what would be stalled by replacing the import statements with explicit ones, aside from the cases of new imports. I'm perfectly content waiting on #10359 and other PR's that are considered high priority though, but I am also fairly certain @jnewbery (author of #10359) is all for this PR being, considering he tried the same thing in #9876.

practicalswift:
Needs rebase :-)

kallewoof:
Rebased!

practicalswift:
Thanks for rebasing.

It appears as if `test/functional/merkle_blocks.py` needs to import `assert_raises_jsonrpc` from `test_framework.util` :-)

The use of `assert_raises_jsonrpc` was introduced in @jnewbery:s commit 6294f3283a5b6919795621dc067ec80c0cd2a334.

jtimon:
Concept ACK, needs rebase

kallewoof:
@jtimon That might make sense, yeah. (As for rebasing, I will hold off on that until I know we are ready to move forward with this.)

laanwj:
Closing this, this seems to be somewhat controversial, and that's not desirable for something that makes changes all over the test framework and causes lots of necessary rebases of other commits.

PR #10368, opened 2017-05-09
[wallet] Remove helper conversion operator from wallet

kallewoof:
The `CTransaction&()` operator in `CMerkleTx` makes conversion into `CTransaction`s transparent, but was marked as to-be-removed in favor of explicitly getting the `tx` ivar, presumably as the operator can lead to ambiguous behavior and makes the code harder to follow.

This PR removes the operator and adapts callers. This includes some cases of `static_cast<CTransaction>(wtx)` → `*wtx.tx`, which is definitely an improvement.

### DIFF
--- a/src/qt/transactiondesc.cpp
+++ b/src/qt/transactiondesc.cpp
@@ -24,7 +24,7 @@
 QString TransactionDesc::FormatTxStatus(const CWalletTx& wtx)
 {
     AssertLockHeld(cs_main);
-    if (!CheckFinalTx(wtx))
+    if (!CheckFinalTx(*wtx.tx))
     {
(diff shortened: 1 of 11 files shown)


practicalswift:
concept ACK

laanwj:
utACK 5a5e4e9

jonasschnelli:
utACK 5a5e4e9cc19e30186ba0a15a1cf892516d419baf

laanwj:
This is an obvious TODO that was implemented, I think this was held up because we wanted to limit impact to the wallet code to be able to ease back-porting for 0.15.x segwit wallet release. But I think we just just push that for 0.16.0.

PR #10371, opened 2017-05-09
[tests] Clean up addrman_tests.cpp

jimmysong:
Cleanup request from #10287.
Change "Test #:" comments to "Test:"
Change BOOST_CHECK(... = ...) to BOOST_CHECK_EQUAL(..., ...)

Requested per @EthanHeilman

### DIFF
--- a/src/test/addrman_tests.cpp
+++ b/src/test/addrman_tests.cpp
@@ -84,43 +84,43 @@ BOOST_AUTO_TEST_CASE(addrman_simple)
 
     CNetAddr source = ResolveIP("252.2.2.2");
 
-    // Test 1: Does Addrman respond correctly when empty.
-    BOOST_CHECK(addrman.size() == 0);
+    // Test: Does Addrman respond correctly when empty.
+    BOOST_CHECK_EQUAL(addrman.size(), 0);
     CAddrInfo addr_null = addrman.Select();
-    BOOST_CHECK(addr_null.ToString() == "[::]:0");
+    BOOST_CHECK_EQUAL(addr_null.ToString(), "[::]:0");
 
-    // Test 2: Does Addrman::Add work as expected.
+    // Test: Does Addrman::Add work as expected.
     CService addr1 = ResolveService("250.1.1.1", 8333);
     BOOST_CHECK(addrman.Add(CAddress(addr1, NODE_NONE), source));
-    BOOST_CHECK(addrman.size() == 1);
+    BOOST_CHECK_EQUAL(addrman.size(), 1);
     CAddrInfo addr_ret1 = addrman.Select();
-    BOOST_CHECK(addr_ret1.ToString() == "250.1.1.1:8333");
+    BOOST_CHECK_EQUAL(addr_ret1.ToString(), "250.1.1.1:8333");
 
-    // Test 3: Does IP address deduplication work correctly.
+    // Test: Does IP address deduplication work correctly.
     //  Expected dup IP should not be added.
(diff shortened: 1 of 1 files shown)


src/test/addrman_tests.cpp
```diff
@@ -387,22 +387,22 @@ BOOST_AUTO_TEST_CASE(addrman_getaddr)
...
+    // Test: Ensure GetAddr works with new addresses.
     addrman.Add(addr1, source1);
     addrman.Add(addr2, source2);
     addrman.Add(addr3, source1);
     addrman.Add(addr4, source2);
     addrman.Add(addr5, source1);
 
     // GetAddr returns 23% of addresses, 23% of 5 is 1 rounded down.
-    BOOST_CHECK(addrman.GetAddr().size() == 1); 
+    BOOST_CHECK_EQUAL(addrman.GetAddr().size(), 1);
```
MarcoFalke: Please no trailing whitespace. This will cause issues in version control when used with editors that remove trailing whitespace by default.

jimmysong: sorry, not sure how that got there. fixing now.

MarcoFalke: It has been there before, though no need to drag along. :)

src/test/addrman_tests.cpp
```diff
@@ -296,23 +296,23 @@ BOOST_AUTO_TEST_CASE(addrman_find)
     addrman.Add(addr2, source2);
     addrman.Add(addr3, source1);
 
-    // Test 17: ensure Find returns an IP matching what we searched on.
+    // Test: ensure Find returns an IP matching what we searched on.
     CAddrInfo* info1 = addrman.Find(addr1);
     BOOST_CHECK(info1);
     if (info1)
```
jnewbery: I was going to suggest you update this if statement to follow our style guidelines (continuation if the block is one line, or braces), but really I don't think an if statement has any business being here at all. This is a test case - we should be asserting all our expectations (and indeed we do in the line above).

Suggest you remove the if line and run `BOOST_CHECK(info1->ToString() == "250.1.2.1:8333")` directly. Same for other 2 ifs below.

jnewbery (commented):
Looks good to me. Tested ACK b81103aad0189075e89de5626679ed57f563846d with one nit.

src/test/addrman_tests.cpp
jimmysong: Thanks for the review! Changed.

MarcoFalke: This was added as a guard against null pointer dereference

jimmysong: That makes sense. Still trying to figure this out. @jnewbery, what's the coding standard that I should change the if to?

jnewbery: Doesn't `BOOST_CHECK(info1)` check that info1 is non-null?

MarcoFalke: According to http://www.boost.org/doc/libs/1_64_0/libs/test/doc/html/boost_test/utf_reference/testing_tool_ref/assertion_boost_level.html you could use `BOOST_REQUIRE()`

jimmysong: attempt 3. Please let me know if this is satisfies everything @jnewbery @MarcoFalke

jnewbery:
looks great. Tested ACK a80f295666ba15d795be00773f768d7167599a63

MarcoFalke:
utACK a80f295666ba15d795be00773f768d7167599a63

PR #10374, opened 2017-05-09
qa: Warn when specified test is not found

maflcko:
This will warn on typos on tests provided manually on the command line.

Also, this comes with the new feature that the same test can be specified multiple times on the command line.


test/functional/test_runner.py
```diff
@@ -173,7 +173,7 @@ def main():
     args, unknown_args = parser.parse_known_args()
 
     # Create a set to store arguments and create the passon string
-    tests = set(arg for arg in unknown_args if arg[:2] != "--")
+    tests = [arg for arg in unknown_args if arg[:2] != "--"]
```
jimmysong: This is great functionality to allow tests to be run multiple times. It might be useful to edit https://github.com/MarcoFalke/bitcoin/blob/Mf1705-qaWarnTestNotFound/test/README.md#functional-tests  to specify that you can, in fact, run the same test a few times to catch intermittent errors.

test/functional/test_runner.py
```diff
@@ -219,9 +224,12 @@ def main():
 
     # Remove the test cases that the user has explicitly asked to exclude.
     if args.exclude:
-        for exclude_test in args.exclude.split(','):
-            if exclude_test + ".py" in test_list:
-                test_list.remove(exclude_test + ".py")
+        tests_excl = [re.sub("\.py$", "", t) + ".py" for t in args.exclude.split(',')]
+        for exclude_test in tests_excl:
+            if exclude_test in test_list:
+                test_list.remove(exclude_test)
```
jimmysong: Tthis will only remove the first instance of `exclude_test` from `test_list`. It's now possible that there are multiple instances of the script from line 176 above. Something like this will work:

    test_list = filter(lambda t: t == exclude_test, test_list)

jimmysong (commented):
utACK fa57826bbf14e16a959b945a1cf74d2c58f257df

Everything looks good other than a couple of nits. Will test once nits addressed.

test/functional/test_runner.py
```diff
@@ -173,7 +173,7 @@ def main():
     args, unknown_args = parser.parse_known_args()
 
     # Create a set to store arguments and create the passon string
```
jnewbery: nit: comment is now incorrect (since this is a list rather than a set)

jnewbery (commented):
Tested ACK fa57826bbf14e16a959b945a1cf74d2c58f257df

test/functional/test_runner.py
MarcoFalke: This is a feature. If you really want to specify the same test multiple times, and then exclude it again, you need to specify it multiple times as well. Though, I doubt this will ever be relevant.

MarcoFalke:
Thx for the feedback. Updated doc with --amend

laanwj:
utACK fac79e4

PR #10381, opened 2017-05-10
Shadowing warnings are not enabled by default, update doc accordingly

paveljanik:


jonasschnelli:
ACK f203ecc621c0af594c43c7fce8a765a8259e3b43

laanwj:
I've changed the title of the PR a bit, it sounded more like we're developing a 3D engine.

ACK otherwise

PR #10383, opened 2017-05-10
[logging] log system time and mock time

jnewbery:
Currently, if mocktime is set, the logs will be timestamped with the mocktime instead of the system time. It's often useful to have the system time in the logs (for example, when running an integration test, it's useful to know what bitcoind was doing at system time so you can see how it's interacting with the test framework or other bitcoind nodes).

This PR timestamps the log with the actual system time, and also prints out the mocktime if mocktime is being used.

@sdaftuar

### DIFF
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -310,10 +310,14 @@ static std::string LogTimestampStr(const std::string &str, std::atomic_bool *fSt
         return str;
 
     if (*fStartedNewLine) {
-        int64_t nTimeMicros = GetLogTimeMicros();
+        int64_t nTimeMicros = GetTimeMicros();
         strStamped = DateTimeStrFormat("%Y-%m-%d %H:%M:%S", nTimeMicros/1000000);
         if (fLogTimeMicros)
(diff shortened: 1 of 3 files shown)


laanwj:
An alternative suggestion would be to log changes to the mock time, instead of prepending it to every line, which causes vary long lines.

sdaftuar:
This looks reasonable to me, utACK.  This will fix the issues that `test/functional/combine_logs` has with tests using mocktime, right?

@laanwj I'd personally prefer to have the extra timestamp on each line when using mocktime -- one use case I have is for my simulation environment, where mocktime is set on every message (tx, block, etc) that is delivered to bitcoind.  Adding an extra line for each change in mocktime would add substantial clutter, and also make reading/parsing the actual debug output more difficult (the mocktime is usually more interesting to me than the system time).

laanwj:
OK as this is only used for testing (and only a subset of the tests, as that) I'm fine with adding it to every line.

kallewoof (approved):
utACK 761392db3a721728863f773d337d516125ddb23c

PR #10385, opened 2017-05-10
RPC: getblock returns coinbase scriptsig

pinheadmz:
Miners are putting lots of exciting information in coinbase scriptSig's these days... Currently to get that string via RPC two calls are required: first `getblock` to get the list of all transactions in the block, and then `getrawtransaction` with the coinbase transaction hash. Of course this only works if your node is running with `txindex=1`, which is a lot of overhead if all you're interested in is coinbase strings. This PR makes it easy:
```
$ bitcoin-cli getblock 000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f | grep coinbase
  "coinbase": "04ffff001d0104455468652054696d65732030332f4a616e2f32303039204368616e63656c6c6f72206f6e206272696e6b206f66207365636f6e64206261696c6f757420666f722062616e6b73",
$ echo 04ffff001d0104455468652054696d65732030332f4a616e2f32303039204368616e63656c6c6f72206f6e206272696e6b206f66207365636f6e64206261696c6f757420666f722062616e6b73 | xxd -r -p
??EThe Times 03/Jan/2009 Chancellor on brink of second bailout for banks
```

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -118,6 +118,7 @@ UniValue blockToJSON(const CBlock& block, const CBlockIndex* blockindex, bool tx
     result.push_back(Pair("version", block.nVersion));
     result.push_back(Pair("versionHex", strprintf("%08x", block.nVersion)));
     result.push_back(Pair("merkleroot", block.hashMerkleRoot.GetHex()));
+    result.push_back(Pair("coinbase", HexStr(block.vtx[0]->vin[0].scriptSig)));
     UniValue txs(UniValue::VARR);
     for(const auto& tx : block.vtx)
     {
@@ -705,6 +706,7 @@ UniValue getblock(const JSONRPCRequest& request)
             "  \"version\" : n,         (numeric) The block version\n"
             "  \"versionHex\" : \"00000000\", (string) The block version formatted in hexadecimal\n"
             "  \"merkleroot\" : \"xxxx\", (string) The merkle root\n"
+            "  \"coinbase\"   : \"xxxx\", (string) The scriptSig of the coinbase transaction\n"
             "  \"tx\" : [               (array of string) The transaction ids\n"
             "     \"transactionid\"     (string) The transaction id\n"
             "     ,...\n"


sdaftuar:
I believe this would be implemented by #8704?

pinheadmz:
Oh huh! You know I saw the extra "verbose" lines in the code and thought to myself "huh I didn't know that was a new feature" but didn't think that it actually solves my problem!

laanwj:
> but didn't think that it actually solves my problem!

What are you missing? After #8704:
```
$ src/bitcoin-cli getblock 000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f 2
...
      "vin": [
        {
          "coinbase": "04ffff001d0104455468652054696d65732030332f4a616e2f32303039204368616e63656c6c6f72206f6e206272696e6b206f66207365636f6e64206261696c6f757420666f722062616e6b73",
...
```

pinheadmz:
@laanwj I DIDN'T think, this is a facepalm. I saw the code from #8704, said "huh" out loud and went forward with this PR anyway, feel free to close.

laanwj:
It might still make sense to include the coinbase info at a lower verbosity level, as the verbosity=2 output can be huge, I don't know.

pinheadmz:
Right, well the motivation is that miners are mining with clients that use the coinbase scriptSig as an "extra version" field. My application is just trying to sort out what's going on out there, so I don't need the RPC call to return 1MB of json. I just want to see the version bits and the "extra version".

jnewbery:
There are a couple of ways to get the full coinbase transaction:

- `getblock()` with the verbosity level set to 2
- (once #10275 is merged) `getblock()` with verbosity level set to 1, and then `getrawtransaction()` with the txid and blockhash specified.

I don't think adding a new option to return only the coinbase transaction from `getblock()` is necessary, but I'm not completely opposed. Just please don't add another 'verbosity level' but instead make this an explicit option.

pinheadmz:
@jnewbery Isn't the problem with `getrawtransaction()` is that it requires `txindex=1`? You're right about `getblock` with `verbose=2`, though -- I didn't realize that was in the code already when I started my PR. However, since miners are putting messages in the coinbase scriptSig, it is useful at least to my own application to include that string (not even the whole transaction, just the scriptSig) as part of the top-level return values along with version and size when describing a block.

sipa:
I don't think this use case warrants an extra RPC/mode. You can just use verbose=2 and look at the coinbase tx of the output.

jnewbery:
> Isn't the problem with getrawtransaction() is that it requires txindex=1

Please see #10275

pinheadmz:
Cool I will close

PR #10386, opened 2017-05-11
[WIP] [wallet] Optional '-avoidreuse' flag which defaults to not reusing addresses in sends

kallewoof:
(Note: putting this on hold until #12257 is merged.)

#10065 brings up a privacy issue where a user can send a bunch of near-dust transactions to an address, which would be picked up by the coin select code when the owner funded transactions, connecting multiple transactions and addresses to the same user.

This adds a (by default turned off) flag `-avoidreuse`. When enabled, the wallet will mark any addresses that were used to fund a transaction as "dirty" and will avoid using them in funding additional transactions, unless an "allow dirty" flag is set.

It also adds support to allow dirty addresses in `sendtoaddress`. More tweaks to other RPC commands is necessary but I wanted to keep the PR as small as possible.

Retroactive flagging of dirty addresses can be done by rescanning the chain.

### DIFF
--- a/src/wallet/coincontrol.h
+++ b/src/wallet/coincontrol.h
@@ -34,6 +34,8 @@ public:
     boost::optional<bool> m_signal_bip125_rbf;
     //! Avoid partial use of funds sent to a given address
     bool m_avoid_partial_spends;
+    //! Allows inclusion of dirty (previously used) addresses
+    bool m_allow_dirty_addresses;
     //! Fee estimation mode to control arguments to estimateSmartFee
     FeeEstimateMode m_fee_mode;
 
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -511,9 +511,10 @@ static UniValue sendtoaddress(const JSONRPCRequest& request)
         return NullUniValue;
     }
 
-    if (request.fHelp || request.params.size() < 2 || request.params.size() > 8)
+    if (request.fHelp || request.params.size() < 2 || request.params.size() > 9)
         throw std::runtime_error(
-            "sendtoaddress \"address\" amount ( \"comment\" \"comment_to\" subtractfeefromamount replaceable conf_target \"estimate_mode\")\n"
+            std::string() +
+            "sendtoaddress \"address\" amount ( \"comment\" \"comment_to\" subtractfeefromamount replaceable conf_target \"estimate_mode\" allowdirty )\n"
             "\nSend an amount to a given address.\n"
             + HelpRequiringPassphrase(pwallet) +
             "\nArguments:\n"
@@ -532,12 +533,15 @@ static UniValue sendtoaddress(const JSONRPCRequest& request)
             "       \"UNSET\"\n"
             "       \"ECONOMICAL\"\n"
             "       \"CONSERVATIVE\"\n"
+            "9. allowdirty             (boolean, optional, " + (gArgs.GetBoolArg("-avoidreuse", DEFAULT_AVOIDREUSE) ? std::string("default=") + (DEFAULT_AVOIDREUSE ? "true" : "false") : "unavailable") + ") Allows spending from dirty addresses; addresses are considered\n"
+            "                             dirty if they have previously been used in a transaction.\n"
             "\nResult:\n"
             "\"txid\"                  (string) The transaction id.\n"
             "\nExamples:\n"
             + HelpExampleCli("sendtoaddress", "\"1M72Sfpbz1BPpXFHz9m3CdqATR44Jvaydd\" 0.1")
             + HelpExampleCli("sendtoaddress", "\"1M72Sfpbz1BPpXFHz9m3CdqATR44Jvaydd\" 0.1 \"donation\" \"seans outpost\"")
             + HelpExampleCli("sendtoaddress", "\"1M72Sfpbz1BPpXFHz9m3CdqATR44Jvaydd\" 0.1 \"\" \"\" true")
+            + HelpExampleCli("sendtoaddress", "\"1M72Sfpbz1BPpXFHz9m3CdqATR44Jvaydd\" 0.1 \"\" \"\" false true")
             + HelpExampleRpc("sendtoaddress", "\"1M72Sfpbz1BPpXFHz9m3CdqATR44Jvaydd\", 0.1, \"donation\", \"seans outpost\"")
         );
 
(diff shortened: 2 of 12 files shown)


luke-jr:
IMO this isn't complete unless incoming transactions to dirty addresses (even before being used!) are hidden as well (or at least flagged in some visible manner).

Also, shouldn't rescanning be sufficient?

kallewoof:
@luke-jr: Maybe I am misunderstanding you, but incoming transactions are irrelevant. The only thing that matters is when you spend from an address. Each time you do, that address is marked dirty and any UTXOs pointing to it are automatically considered dirty in this implementation. Am I missing a case?

> Also, shouldn't rescanning be sufficient?

Oops - yes, rescanning is sufficient. Updated OP, thanks.

src/wallet/wallet.cpp
```diff
@@ -846,6 +875,14 @@ bool CWallet::AddToWallet(const CWalletTx& wtxIn, bool fFlushOnClose)
 
     uint256 hash = wtxIn.GetHash();
 
+    if (GetBoolArg("-avoidreuse", false)) {
```
jonasschnelli: Maybe use a constant for the default value here.

src/wallet/wallet.cpp
```diff
@@ -3598,6 +3639,7 @@ std::string CWallet::GetWalletHelpString(bool showDebug)
     strUsage += HelpMessageOpt("-walletnotify=<cmd>", _("Execute command when a wallet transaction changes (%s in cmd is replaced by TxID)"));
     strUsage += HelpMessageOpt("-zapwallettxes=<mode>", _("Delete all wallet transactions and only recover those parts of the blockchain through -rescan on startup") +
                                " " + _("(1 = keep tx meta data e.g. account owner and payment request information, 2 = drop tx meta data)"));
+    strUsage += HelpMessageOpt("-avoidreuse", _("Mark addresses which have been used to fund transactions in the past, and avoid reusing these in future funding, except when explicitly requested"));
```
jonasschnelli: Missing default value (something like  `(strprintf(_("(default: %u)"), DEFAULT_AVOIDREUSE));`).

src/wallet/wallet.cpp
```diff
@@ -838,6 +838,35 @@ bool CWallet::MarkReplaced(const uint256& originalHash, const uint256& newHash)
...
 
+void CWallet::SetDirtyState(const uint256& hash, unsigned int n, bool dirty)
+{
+    const CWalletTx* srctx = GetWalletTx(hash);
+    if (srctx) {
+        CTxDestination dst;
+        if (ExtractDestination(srctx->tx->vout[n].scriptPubKey, dst)) {
+            if (::IsMine(*this, dst)) {
+                if (dirty && !GetDestData(dst, "dirty", NULL)) {
+                    AddDestData(dst, "dirty", "p"); // p for "present", opposite of absent (null)
```
jonasschnelli: Maybe add a `AssertLockHeld(cs_wallet);` somewhere above `GetDestData`?

src/wallet/rpcwallet.cpp
```diff
@@ -416,12 +419,15 @@ UniValue sendtoaddress(const JSONRPCRequest& request)
             "                             transaction, just kept in your wallet.\n"
             "5. subtractfeefromamount  (boolean, optional, default=false) The fee will be deducted from the amount being sent.\n"
             "                             The recipient will receive less bitcoins than you enter in the amount field.\n"
+            "6. allowdirty             (boolean, optional, " + (GetBoolArg("-avoidreuse", false) ? "default=false" : "unavailable") + ") Allows spending from dirty addresses; addresses are considered\n"
```
jonasschnelli: Why `unavailable`? Someone may just have enabled `-avoidreuse` but hasn't rescanned. Maybe keep a state somewhere if we have scanned with `-avoidreuse` up to the wallets bestblock.

kallewoof: It returns "unavailable" if you do not have the feature turned on. I don't want people to think they can avoid using dirty coins if they are running without the flag. I.e. even by explicitly saying `allowdirty=false`, it will still use dirty coins when `-avoidreuse` is not enabled.

You are right though, that it will still use dirty coins unless you rescan. Feels like that should be mentioned somewhere, but not sure where.

gmaxwell:
I support the goal, but:  If an address is paid 10 btc then 0.0001 btc  and then a transaction spends the latter, A would be dirty and the 10 BTC are stuck.  That seems sub-optimal.

The best idea I had to deal with that previously is that whenever you spend from A you make an effort to spend all payments to A or at least all non-dust payments to A.  Then the only time you get dirty funds is when someone pays a non-negligible amount to an address after you've already spent to it. 

If something is going to cause transactions to fail, we probably will need another kind of balance display to make the behavior explicable.

Independently from this (but also useful for it) we probably should have an icon in the transaction list for reused addresses. (Trefoil made of arrows? :P )

kallewoof:
You are able to spend by using the allowdirty flag in sendtoaddress, and you can always make a raw transaction yourself. The intention of this is to give expert users a way to plug the gaping security hole that exists in the system -- to make it intuitive and wonderful for the average user is not an aspiration at this point (as mentioned, in particular RPC commands need some work before this could ever be made a default-on feature).

The solutions you present are all solutions an expert user could make use of with this implementation. It would simply stop their clients from shooting them (privacy wise) in the foot automatically.

kallewoof:
Rebased (and code became slightly more clean thanks to new coin control being present).

src/wallet/coincontrol.h
```diff
@@ -27,6 +27,8 @@ class CCoinControl
     boost::optional<CFeeRate> m_feerate;
     //! Override the default confirmation target if set
     boost::optional<unsigned int> m_confirm_target;
+    //! Allows inclusion of dirty (previously used) addresses
+    bool fAllowDirtyAddresses;
```
promag: `bool allow_dirty_addresses`.

src/wallet/rpcwallet.cpp
```diff
@@ -500,6 +504,7 @@ UniValue sendtoaddress(const JSONRPCRequest& request)
         }
     }
 
+    coin_control.fAllowDirtyAddresses = !request.params[8].isNull() && request.params[8].get_bool();
```
promag: This does not default to `DEFAULT_AVOIDREUSE`.

promag (commented):
Should creating transactions to dirty addresses fail?

Needs rebase.

Nit, commit messages don't have period.

kallewoof:
@promag Thanks for the review! Fixed the default usage issue.

src/wallet/coincontrol.h
kallewoof: Missed this comment earlier; fixed.

PR #10391, opened 2017-05-12
OP_CHECKBLOCKATHEIGHT anti-replay (BIP 115; logic only)

luke-jr:


dcousens:
https://github.com/luke-jr/bips/blob/bip-0115/bip-0115.mediawiki - for others reference

src/script/interpreter.cpp
```diff
@@ -424,7 +425,36 @@ bool EvalScript(std::vector<std::vector<unsigned char> >& stack, const CScript&
...
+                        }
+                        break;
+                    }
+
+                    if (stack.size() < 2) {
+                        return set_error(serror, SCRIPT_ERR_INVALID_STACK_OPERATION);
+                    }
+
+                    // nHeight is a 32-bit signed integer field.
+                    const int32_t nHeight = CScriptNum(stacktop(-1), true, 4).getint();
```
aayanl: Can this cause a buffer overflow if a block height over 2147483647 occurs?

luke-jr: No? Why would it?

aayanl: nHeight is a int32_t, if someone makes a transaction with 2147483648 as the referenced block, it would cause an overflow

luke-jr: Not all overflows are buffer overflows. In this case, CScriptNum explicitly limits the size.

luke-jr:
The BIP needs revising for static analysis compat.

PR #10395, opened 2017-05-13
Replace boost::function with std::function (C++11)

practicalswift:


dcousens (approved):

MarcoFalke:
utACK 1b936f59264a4f4a867baece1e0ee4ec02f73cee

sipa:
utACK 1b936f59264a4f4a867baece1e0ee4ec02f73cee

paveljanik (approved):
ACK

src/bench/bench.cpp
```diff
@@ -5,6 +5,7 @@
 #include "bench.h"
 #include "perf.h"
 
+#include <assert.h>
```
practicalswift: @paveljanik That is a very good question!

Compling without `#include <assert.h>` in `src/bench/bench.cpp`:

```
  CXX      bench/bench_bench_bitcoin-bench.o
bench/bench.cpp: In member function ‘bool benchmark::State::KeepRunning()’:
bench/bench.cpp:95:86: error: ‘assert’ was not declared in this scope
     assert(count != 0 && "count == 0 => (now == 0 && beginTime == 0) => return above");
```

Compiling without `#include <limits>` in `src/bench/bench.h`:                                                                                      

```
  CXX      bench/bench_bench_bitcoin-bench_bitcoin.o
In file included from bench/bench_bitcoin.cpp:5:0:
bench/bench.h: In constructor ‘benchmark::State::State(std::__cxx11::string, double)’:
bench/bench.h:52:23: error: ‘numeric_limits’ is not a member of ‘std’
             minTime = std::numeric_limits<double>::max();
                       ^
```

Seems like `#include <boost/function.hpp>` gives you `assert.h` and `<limits>` too?

paveljanik: Perfect answer, thanks! I should have tested it myself 8)

fanquake:
utACK 1b936f5

PR #10400, opened 2017-05-14
[RPC] Add an uptime command that displays the amount of time (in seconds) bitcoind has been running

rvelhote:
Hello everyone,

I am working on a tool that displays information about a bitcoin node using RPC. One of the parameters I would like to include is the `uptime` of the bitcoin daemon. Currently, as far as I can tell, there is no way of obtaining such information.

After thinking a bit on how to perform this task, I concluded that the easiest way to do it is to check the modification time of the `bitcoind.pid` file. This has the disadvantage that it does not work in Windows because `bitcoind.pid` is not created.

I propose the following solution in this PR and submit it for your review and opinion concerning this new RPC command before I work further on this topic. Do you think this new command makes sense? Please consider it a WIP/PoC as it does not have any error checking or tests.

Here is a sample return from the RPC command:

```
$ ./bitcoin-cli uptime  
{
  "uptime": 696
}
```
and the help function:
```
$bitcoin-cli help uptime  
uptime

Display the uptime of the Bitcoin server.
```

Best regards,
Ricardo Velhote

### DIFF
--- a/src/rpc/server.cpp
+++ b/src/rpc/server.cpp
@@ -258,6 +258,22 @@ UniValue stop(const JSONRPCRequest& jsonRequest)
     return "Bitcoin server stopping";
 }
 
+UniValue uptime(const JSONRPCRequest& jsonRequest)
+{
+    if (jsonRequest.fHelp || jsonRequest.params.size() > 1)
+        throw std::runtime_error(
+                "uptime\n"
+                        "\nReturns the total uptime of the server.\n"
+                        "\nResult:\n"
+                        "ttt        (numeric) The number of seconds that the server has been running\n"
+                        "\nExamples:\n"
+                + HelpExampleCli("uptime", "")
+                + HelpExampleRpc("uptime", "")
+        );
+
+    return GetTime() - GetStartupTime();
+}
+
 /**
  * Call Table
  */
@@ -267,6 +283,7 @@ static const CRPCCommand vRPCCommands[] =
     /* Overall control/query calls */
     { "control",            "help",                   &help,                   true,  {"command"}  },
     { "control",            "stop",                   &stop,                   true,  {}  },
+    { "control",            "uptime",                 &uptime,                 true,  {}  },
 };
 
 CRPCTable::CRPCTable()
--- a/src/util.h
+++ b/src/util.h
@@ -5,7 +5,7 @@
 
 /**
  * Server/client environment: argument handling, config file parsing,
- * logging, thread wrappers
+ * logging, thread wrappers, startup time
  */
 #ifndef BITCOIN_UTIL_H
 #define BITCOIN_UTIL_H
@@ -28,6 +28,9 @@
 #include <vector>
 
 #include <boost/signals2/signal.hpp>
+
+// Application startup time (used for uptime calculation)
+int64_t GetStartupTime();
 
 static const bool DEFAULT_LOGTIMEMICROS = false;
 static const bool DEFAULT_LOGIPS        = false;
--- a/test/functional/test_runner.py
+++ b/test/functional/test_runner.py
@@ -113,6 +113,7 @@ BASE_SCRIPTS= [
     'listsinceblock.py',
     'p2p-leaktests.py',
     'wallet-encryption.py',
+    'uptime.py',
 ]
 
 EXTENDED_SCRIPTS = [
--- a/src/qt/clientmodel.cpp
+++ b/src/qt/clientmodel.cpp
@@ -26,7 +26,6 @@
 
 class CBlockIndex;
 
-static const int64_t nClientStartupTime = GetTime();
 static int64_t nLastHeaderTipUpdateNotification = 0;
 static int64_t nLastBlockTipUpdateNotification = 0;
 
@@ -238,7 +237,7 @@ bool ClientModel::isReleaseVersion() const
 
 QString ClientModel::formatClientStartupTime() const
 {
-    return QDateTime::fromTime_t(nClientStartupTime).toString();
+    return QDateTime::fromTime_t(GetStartupTime()).toString();
 }
 
 QString ClientModel::dataDir() const
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -84,6 +84,8 @@
 #include <openssl/rand.h>
 #include <openssl/conf.h>
 
+// Application startup time (used for uptime calculation)
+const int64_t nStartupTime = GetTime();
 
 const char * const BITCOIN_CONF_FILENAME = "bitcoin.conf";
 const char * const BITCOIN_PID_FILENAME = "bitcoind.pid";
@@ -888,4 +890,10 @@ std::string CopyrightHolders(const std::string& strPrefix)
         strCopyrightHolders += "\n" + strPrefix + "The Bitcoin Core developers";
     }
     return strCopyrightHolders;
+}
+
+// Obtain the application startup time (used for uptime calculation)
+int64_t GetStartupTime()
+{
+    return nStartupTime;
 }
--- a/test/functional/uptime.py
+++ b/test/functional/uptime.py
@@ -1,0 +1,32 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Test the RPC call related to the uptime command.
+
+Test corresponds to code in rpc/server.cpp.
+"""
+
+import time
+
+from test_framework.test_framework import BitcoinTestFramework
+
+
+class UptimeTest(BitcoinTestFramework):
+    def __init__(self):
+        super().__init__()
+
+        self.num_nodes = 1
+        self.setup_clean_chain = True
+
+    def run_test(self):
+        self._test_uptime()
+
+    def _test_uptime(self):
+        wait_time = 10
+        self.nodes[0].setmocktime(int(time.time() + wait_time))
+        assert(self.nodes[0].uptime() >= wait_time)
+
+
+if __name__ == '__main__':
+    UptimeTest().main()


laanwj:
> After thinking a bit on how to perform this task, I concluded that the easiest way to do it is to check the modification time of the bitcoind.pid file. 

There is no need to look at the timestamp of any files - the server could simply remember its own startup timestamp and use that as a reference. This would work just as well on windows.

rvelhote:
Thank you for replying Wladimir. You are correct of course. Checking the PID file was the most simple and with the least changes to the codebase I could think of - while still being useful (I am not familiar with the Bitcoin core and have not worked with C++ since 2010).

So, if you think this command is useful for inclusion in the core, I could work on improving it and then squash the commits in the end.

Should this information be its own command or should it be included in another existing command - if so, which one?

Thank you,
Ricardo Velhote

jonasschnelli:
It seems to be useful.
Concept ACK.

Agree with @laanwj. Just set a variable during startup with `GetTime()` (now) and calculate the delta to "`now()`" when someone want's to know the current uptime.

paveljanik:
Concept ACK

The other way could be adding starttime to `getinfo` RPC. But I prefer your solution, separate simple approach, new RPC. Please also add a test case for this.

laanwj:
> The other way could be adding starttime to getinfo RPC. But I prefer your solution, separate simple approach, new RPC. Please also add a test case for this.

A gentle reminder that `getinfo` should not be extended, see https://github.com/bitcoin/bitcoin/blob/master/src/rpc/misc.cpp#L34:
```c++
/**
 * @note Do not add or change anything in the information returned by this
 * method. `getinfo` exists for backwards-compatibility only. It combines
 * information from wildly different sources in the program, which is a mess,
 * and is thus planned to be deprecated eventually.
 *
 * Based on the source of the information, new information should be added to:
 * - `getblockchaininfo`,
 * - `getnetworkinfo` or
 * - `getwalletinfo`
 *
 * Or alternatively, create a specific query method for the information.
 **
```
One of the other `get*info` would be an option but I don't see where it'd clearly fit in. So agree a separate call is OK.

achow101:
There's already a variable in the GUI for getting the startup time: https://github.com/bitcoin/bitcoin/blob/86ea3c2ff247bb2ba0fb50013c8ecdbaf8a9fe8f/src/qt/clientmodel.cpp#L29

You could probably just move it so that it can be used for both the GUI and for this RPC call.

rvelhote:
Thank you for your feedback. I went ahead and made some modifications according to your suggestions.

I have pushed a new commit that removes `nClientStartupTime` from being specific to the QT interface and moved it to `util.h` where it can be used by both the QT interface and the RPC server.

I called it `nStartupTime` instead of `nClientStartupTime` because now it related to both the client and the daemon. I also decided on `util.h` instead of `init.h`. The reason is that the startup time is more of an utility/informational feature rather than a core initialization feature.

I am not sure of the naming conventions. Should I use `g_nStartupTime` to symbolize that it's a global variable or is that used for specific objects like `g_connman`?

I have build the QT client (QT5 in my case) and the daemon and from my testing every is working well. I have also started the QT client with the `-server` flag and both are returning the correct value. The QT client displays the startup date in an ISO format (Debug Window » Information) and the RPC server displays the amount of seconds that have passed since that startup date.

I will add a functional test to the RPC interface for this command. I'm not sure how the testing is done for the QT client.

src/util.h
```diff
@@ -29,6 +29,9 @@
 #include <boost/signals2/signal.hpp>
 #include <boost/thread/exceptions.hpp>
 
+// Bitcoin startup time
```
paveljanik: // Application startup time (used for uptime calculation)

src/rpc/server.cpp
```diff
@@ -258,6 +258,27 @@ UniValue stop(const JSONRPCRequest& jsonRequest)
...
+UniValue uptime(const JSONRPCRequest& jsonRequest)
+{
+    if (jsonRequest.fHelp || jsonRequest.params.size() > 1)
+        throw std::runtime_error(
+                "uptime\n"
+                        "\nDisplay the uptime of the Bitcoin server.");
+
+    int64_t uptime = GetTime() - nStartupTime;
+
+    if (uptime < 0)
```
paveljanik: Do we need this at all?

src/rpc/server.cpp
```diff
@@ -258,6 +258,27 @@ UniValue stop(const JSONRPCRequest& jsonRequest)
...
+    {
+        throw std::runtime_error(
+                "stop\n"
+                        "\nUptime value for the Bitcoin server is a negative value.");
```
paveljanik: Started in the future!? ;-)

src/rpc/server.cpp
```diff
@@ -258,6 +258,27 @@ UniValue stop(const JSONRPCRequest& jsonRequest)
     return "Bitcoin server stopping";
 }
 
+UniValue uptime(const JSONRPCRequest& jsonRequest)
+{
+    if (jsonRequest.fHelp || jsonRequest.params.size() > 1)
+        throw std::runtime_error(
+                "uptime\n"
+                        "\nDisplay the uptime of the Bitcoin server.");
```
paveljanik: Let's document output as Result: as other RPCs. Add unit specification (seconds).

src/rpc/server.cpp
rvelhote: I imagined that for some weird reason the clock would change to before the start and the user would get a negative value. Maybe I'm over-thinking this? ;)

paveljanik: Yup, I do not have a problem with this. Just return negative value and do not crash.

rvelhote:
I have made a new commit with some changes suggested by Pavel.

While searching for examples of documentation I noticed `getconnectioncount` which only returns a single value (the node count). Previously the `uptime` command was returning an object with an `uptime` key so I followed the example of `getconnectioncount` and changed the return value to a single integer that represents the amount of seconds instead of an object.

Here is a sample of the documentation when running `bitcoin-cli help uptime`

```
$ bitcoin-cli help uptime
uptime

Returns the total uptime of the Bitcoin server.

Result:
ttt        (numeric) The number of seconds that the Bitcoin server has been running

Examples:
> bitcoin-cli uptime 
> curl --user myusername --data-binary '{"jsonrpc": "1.0", "id":"curltest", "method": "uptime", "params": [] }' -H 'content-type: text/plain;' http://127.0.0.1:8332/
```

I used the `ttt` designation because I've seen it in other areas to indicate a time-based value (e.g. `getpeerinfo`).

rvelhote:
I've added a functional test to check for the uptime value. In this test I've placed a `time.sleep` before calling the `uptime` command. The test then checks if the current uptime value reported by the RPC command is greater or equal the value used in `time.sleep`.

As I've noted in the test comment:
> Depending on how fast the setup executes, the RPC server starts and this test
> executes the check for the uptime value might not deterministic. At the very
> least we will check if the reported uptime is the amount of seconds we waited.

test/functional/server.py
```diff
@@ -0,0 +1,37 @@
...
+        self._test_uptime()
+
+    def _test_uptime(self):
+        # The number of seconds to wait before checking the uptime value
+        wait_time = 6
+
+        # Depending on how fast the setup executes, the RPC server starts and this test
+        # executes the check for the uptime value might not deterministic. At the very
+        # least we will check if the reported uptime is the amount of seconds we waited.
+        time.sleep(wait_time)
```
paveljanik: You can use `setmocktime` to not really sleep/slow down tests for 6 seconds.

test/functional/test_runner.py
```diff
@@ -110,6 +110,7 @@
     'rpcnamedargs.py',
     'listsinceblock.py',
     'p2p-leaktests.py',
+    'server.py',
```
paveljanik: Please use the name `uptime.py` or so, not generic `server.py`.

rvelhote: I took as an example the `net.py` test file that referred to tests in `rpc/net.cpp`. Since the uptime command is in the `server.cpp` - although there are no tests for this file - I chose to create a `server.py` test file. Do you still think it's best to use `uptime.py`?

test/functional/server.py
rvelhote: Arghh I didn't know about this `setmocktime`. Didn't think of searching either. Will get to it.

rvelhote:
@paveljanik I made the change to `setmocktime` in the test. I held off on the name change of the test file until I get your final feedback after my reply to your comment. Thank you.

test/functional/test_runner.py
paveljanik: @rvelhote Yup, OK.

paveljanik:
OK, let's squash please. The travis failure is unrelated.

src/util.h
```diff
@@ -29,6 +29,9 @@
 #include <boost/signals2/signal.hpp>
 #include <boost/thread/exceptions.hpp>
 
+// Application startup time (used for uptime calculation)
+static const int64_t nStartupTime = GetTime();
```
jnewbery: This shouldn't be static (a static variable in a header file means that any .cpp file that includes the header will have its own local variable called `nStartupTime`).

Instead, I think you should define this variable in util.cpp and then have an extern declaration here. See ` BITCOIN_CONF_FILENAME` below for example.

I don't know if you just coppied the `DEFAULT_LOG...` constants below, but I believe those should also be defined in util.cpp.

jnewbery (commented):
A few nits. The only important one is `nStartupTime` shouldn't be defined as static in the util.h header file.

src/rpc/server.cpp
```diff
@@ -258,6 +258,22 @@ UniValue stop(const JSONRPCRequest& jsonRequest)
     return "Bitcoin server stopping";
 }
 
+UniValue uptime(const JSONRPCRequest& jsonRequest)
```
rvelhote: Thank you very much for your feedback John.

I placed it on `src/rpc/server.php` because the command was information related to the server rather than a utility command (although it can certainly be considered an utility command). That is my only argument for using `server.cpp`. Maybe let's wait for other people's opinion on this?

src/rpc/server.cpp
```diff
@@ -258,6 +258,22 @@ UniValue stop(const JSONRPCRequest& jsonRequest)
...
+{
+    if (jsonRequest.fHelp || jsonRequest.params.size() > 1)
+        throw std::runtime_error(
+                "uptime\n"
+                        "\nReturns the total uptime of the Bitcoin server.\n"
+                        "\nResult:\n"
+                        "ttt        (numeric) The number of seconds that the Bitcoin server has been running\n"
```
rvelhote: I used `ttt` because I saw it referenced in other commands for usage with time related values. For example in `getpeerinfo` you have `conntime` for a timestamp and also `timeoffset` for a time-based value in seconds. I think `ttt` will keep it consistent with other commands.

src/util.h
rvelhote: The `DEFAULT_LOG` constants are [already in master](https://github.com/bitcoin/bitcoin/blob/master/src/util.h#L33). Might be something to improve then.

Also, I don't know if you noticed, this variable was [moved](https://github.com/bitcoin/bitcoin/blob/master/src/qt/clientmodel.cpp#L29) from `src/qt/clientmodel.cpp` to a more global area so that it can be reused by both the server and the QT client. 

I will make this change as you suggested and force-push the commit.

src/rpc/server.cpp
jnewbery: > Maybe let's wait for other people's opinion on this?

certainly. Just my opinion that this should live in misc.cpp. I'm happy to defer to the consensus opinion.

src/rpc/server.cpp
jnewbery: sure. Again, just my personal preference. `ttt` is fine.

src/util.h
jnewbery: Yes, sorry I wasn't clear - I didn't expect you to change the DEFAULT_LOG constants. I do however think you need to move the nStartupTime definition to avoid the following build warnings:

```
In file included from addrman.h:14:0,
                 from addrdb.cpp:8:
util.h:33:22: warning: ‘nStartupTime’ defined but not used [-Wunused-variable]
 static const int64_t nStartupTime = GetTime();
                      ^   
  CXX      libbitcoin_server_a-bloom.o
In file included from addrman.h:14:0,
                 from addrman.cpp:6:
util.h:33:22: warning: ‘nStartupTime’ defined but not used [-Wunused-variable]
 static const int64_t nStartupTime = GetTime();
                      ^   
  CXX      libbitcoin_server_a-blockencodings.o
 /bin/bash ./config.status
In file included from bitcoind.cpp:18:0:
util.h:33:22: warning: ‘nStartupTime’ defined but not used [-Wunused-variable]
 static const int64_t nStartupTime = GetTime();
                      ^
...(repeated many times)
```

rvelhote: No problem John, I understood what you meant :) When I mentioned I would make the change I was referring to moving `nStartupTime` to `util.cpp`. I pushed a new commit with this change.

rvelhote:
@jnewbery @paveljanik When you have the opportunity, please let me know if you have any further feedback on this PR? The TravisCI build is failing but it's because of #10429 and I have not updated my branch since I started the PR. Thank you.

ipoptika:

paveljanik:
Needs rebase

booxie:
``<a href="https://minecloud.io/?in=Booxie" target="_blank" title="MineCloud.io">MineCloud.io</a>

booxie:
Yes uptime for how long in sec Bitcoin has been running

src/util.h
```diff
@@ -29,6 +29,9 @@
 #include <boost/signals2/signal.hpp>
 #include <boost/thread/exceptions.hpp>
 
+// Application startup time (used for uptime calculation)
+extern const int64_t nStartupTime;
```
laanwj: Please add a `GetStartupTime` function instead of exporting this variable.

src/rpc/server.cpp
laanwj: I understand both sides so don't have a strong opinion. We want to have multiple JSON-RPC endpoints at some point, but it seems harmless if they all export `uptime`.

However: `server.cpp` should be independent of bitcoin, so if you plan on keeping it there, please remove the explicit reference to Bitcoin below in the documentation.

laanwj:
Tested ACK c074752

PR #10403, opened 2017-05-15
Fix importmulti failure to return rescan errors

ryanofsky:
An off-by-one-block bug in importmulti rescan logic could cause it to return
success in an edge case even when a rescan was not successful. The case where
this would happen is if there were multiple blocks in a row with the same
GetBlockTimeMax() value, and the last block was scanned successfully, but one
or more of the earlier blocks was not readable.

This is fixed by changing the `CWallet::ScanForWalletTransactions` return value to point to the last failed block that could not be rescanned, and then updating the `ScanForWalletTransactions` call in `importmulti` to use the new return value.

### DIFF
--- a/src/wallet/rpcdump.cpp
+++ b/src/wallet/rpcdump.cpp
@@ -1121,13 +1121,13 @@ UniValue importmulti(const JSONRPCRequest& mainRequest)
 
     if (fRescan && fRunScan && requests.size()) {
         CBlockIndex* pindex = nLowestTimestamp > minimumTimestamp ? chainActive.FindEarliestAtLeast(std::max<int64_t>(nLowestTimestamp - TIMESTAMP_WINDOW, 0)) : chainActive.Genesis();
-        CBlockIndex* scannedRange = nullptr;
+        CBlockIndex* scanFailed = nullptr;
         if (pindex) {
-            scannedRange = pwallet->ScanForWalletTransactions(pindex, true);
+            scanFailed = pwallet->ScanForWalletTransactions(pindex, true);
             pwallet->ReacceptWalletTransactions();
         }
 
-        if (!scannedRange || scannedRange->nHeight > pindex->nHeight) {
+        if (scanFailed) {
             std::vector<UniValue> results = response.getValues();
             response.clear();
             response.setArray();
@@ -1137,12 +1137,23 @@ UniValue importmulti(const JSONRPCRequest& mainRequest)
                 // range, or if the import result already has an error set, let
                 // the result stand unmodified. Otherwise replace the result
                 // with an error message.
-                if (GetImportTimestamp(request, now) - TIMESTAMP_WINDOW >= scannedRange->GetBlockTimeMax() || results.at(i).exists("error")) {
+                if (GetImportTimestamp(request, now) - TIMESTAMP_WINDOW > scanFailed->GetBlockTimeMax() || results.at(i).exists("error")) {
                     response.push_back(results.at(i));
                 } else {
                     UniValue result = UniValue(UniValue::VOBJ);
                     result.pushKV("success", UniValue(false));
-                    result.pushKV("error", JSONRPCError(RPC_MISC_ERROR, strprintf("Failed to rescan before time %d, transactions may be missing.", scannedRange->GetBlockTimeMax())));
+                    result.pushKV(
+                        "error",
+                        JSONRPCError(
+                            RPC_MISC_ERROR,
+                            strprintf("Rescan failed for key with creation timestamp %d. There was an error reading a "
+                                      "block from time %d, which is after or within %d seconds of key creation, and "
+                                      "could contain transactions pertaining to the key. As a result, transactions "
+                                      "and coins using this key may not appear in the wallet. This error could be "
+                                      "caused by pruning or data corruption (see bitcoind log for details) and could "
+                                      "be dealt with by downloading and rescanning the relevant blocks (see -reindex "
+                                      "and -rescan options).",
+                                GetImportTimestamp(request, now), scanFailed->GetBlockTimeMax(), TIMESTAMP_WINDOW)));
                     response.push_back(std::move(result));
                 }
                 ++i;
--- a/src/wallet/test/wallet_tests.cpp
+++ b/src/wallet/test/wallet_tests.cpp
@@ -364,6 +364,7 @@ BOOST_FIXTURE_TEST_CASE(rescan, TestChain100Setup)
     LOCK(cs_main);
 
     // Cap last block file size, and mine new block in a new block file.
+    CBlockIndex* const nullBlock = nullptr;
     CBlockIndex* oldTip = chainActive.Tip();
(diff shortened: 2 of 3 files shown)


src/wallet/rpcdump.cpp
```diff
@@ -1137,12 +1137,12 @@ UniValue importmulti(const JSONRPCRequest& mainRequest)
...
                 // the result stand unmodified. Otherwise replace the result
                 // with an error message.
-                if (GetImportTimestamp(request, now) - TIMESTAMP_WINDOW >= scannedRange->GetBlockTimeMax() || results.at(i).exists("error")) {
+                if (GetImportTimestamp(request, now) - TIMESTAMP_WINDOW > scanFailed->GetBlockTimeMax() || results.at(i).exists("error")) {
                     response.push_back(results.at(i));
                 } else {
                     UniValue result = UniValue(UniValue::VOBJ);
                     result.pushKV("success", UniValue(false));
-                    result.pushKV("error", JSONRPCError(RPC_MISC_ERROR, strprintf("Failed to rescan before time %d, transactions may be missing.", scannedRange->GetBlockTimeMax())));
+                    result.pushKV("error", JSONRPCError(RPC_MISC_ERROR, strprintf("Failed to rescan before time %d, transactions may be missing.", scanFailed->GetBlockTimeMax() + 1)));
```
laanwj: As it scans forward (increasing date / block numbers), and scanFailed returns the first block that could not be scanned, shouldn't this be "after time"?

src/wallet/rpcdump.cpp
```diff
@@ -1137,12 +1137,12 @@ UniValue importmulti(const JSONRPCRequest& mainRequest)
                 // range, or if the import result already has an error set, let
                 // the result stand unmodified. Otherwise replace the result
                 // with an error message.
-                if (GetImportTimestamp(request, now) - TIMESTAMP_WINDOW >= scannedRange->GetBlockTimeMax() || results.at(i).exists("error")) {
+                if (GetImportTimestamp(request, now) - TIMESTAMP_WINDOW > scanFailed->GetBlockTimeMax() || results.at(i).exists("error")) {
```
laanwj: So this checks that the birthday of the imported key is after most recent block that could not be scanned, in which case the current result will be returned?
This seems to disagree with the comment "key creation date is within the successfully scanned range".
What am I misunderstanding here?

src/wallet/rpcdump.cpp
ryanofsky: > As it scans forward (increasing date / block numbers), and scanFailed returns the first block that could not be scanned, shouldn't this be "after time"?

scanFailed returns the last block that could not be scanned, not the first block that couldn't be scanned. To take a concrete example, say there are two blocks in a row with GetBlockTimeMax() == 1000, and the first block fails to scan, and the second block succeeds, and later blocks also succeed. If there are any keys created at time 1000 or earlier, there could be transactions missing that pertain to those keys, so those keys should return errors. If there are any keys created at time 1001 or higher, they should technically be ok (even though they will still return errors for birthdays up to 1000 + TIMESTAMP_WINDOW).

The error returned for the problem keys will be "Failed to rescan before time 1001, transactions may be missing." which I think makes sense if "before" implies "less than" rather than "less than or equal".

src/wallet/rpcdump.cpp
ryanofsky: > So this checks that the birthday of the imported key is after most recent block that could not be scanned, in which case the current result will be returned?

Yes, if the key birthday is after the GetBlockTimeMax of the most recent block that could not be scanned + TIMESTAMP_WINDOW, the rescan for that key's transactions is considered successful, and the result is not replaced.

> This seems to disagree with the comment "key creation date is within the successfully scanned range".
What am I misunderstanding here?

I may be misunderstanding something, but the comment seems right to me. If the most recent block that could not be scanned has GetBlockTimeMax == 1000, then any key created at timestamp 1001 or higher is in the "successfully scanned range", because all the blocks that were bad or missing are older than the key, and therefore not relevant.

src/wallet/rpcdump.cpp
jnewbery: I spent some time thinking about edge cases here, and I've basically convinced myself that this is ok, because:

- the `TIMESTAMP_WINDOW` means that we have a good safety window between the address's birthday and the block timestamp (ie we're safe even if an address appears in a block with a timestamp before its birthday).
- the monotonically increasing `GetBlockTimeMax()` means that this error message gives the most conservative value (ie if a block B's time has moved back from its parent block A and we couldn't scan block B then the error will say that we couldn't scan before Block A's timestamp. That's fine)

I think it might be useful for the user if there was a bit more information in the error message. Up to you how verbose you want to go, but a couple of things that might be good to include:

- how does the user fix this problem (reindex, which will cause a full chain download if this is a pruning node)
- what if the `Failed to rescan before time <time>` is for a time before the user specified as the address birthday? eg what if the user set the birthday to 10am, and we failed to scan a block at 9:55am. We'd return an error  saying `Failed to rescan before time 9:55am` and the user might thing "that's ok, my key's birthday is 10am". In fact it's not ok, because the 9:55am block may include a transaction for that address, since block time is not canonical or monotonic.

ryanofsky: Thanks, expanded error message in 2e041dd247c19a47700b1e1b181c7f8a1b5b109b

ryanofsky (commented):
Squashed 2e041dd247c19a47700b1e1b181c7f8a1b5b109b -> 4d2d6045a4784d576d56299244b9f76a5909904b ([pr/scansame.3](https://github.com/ryanofsky/bitcoin/commits/pr/scansame.3) -> [pr/scansame.4](https://github.com/ryanofsky/bitcoin/commits/pr/scansame.4))

jnewbery:
Looks good. Tested ACK https://github.com/bitcoin/bitcoin/commit/2e041dd247c19a47700b1e1b181c7f8a1b5b109b. It'd be nice to have tests covering the edge cases (block time the same for two consecutive blocks and block time decreasing), but that can be left for a future PR.

jonasschnelli:
utACK 4d2d6045a4784d576d56299244b9f76a5909904b
Very verbose error description texts (but great explanation).

src/wallet/rpcdump.cpp
laanwj: > because all the blocks that were bad or missing are older than the key, and therefore not relevant.

That's the part I'm not sure about: aren't the blocks *at and after* that point missing, instead of before that point?

ryanofsky: > That's the part I'm not sure about: aren't the blocks at and after that point missing, instead of before that point?

Because it returns a pointer to the last block that couldn't be scanned,  it means that any blocks after the pointer were scanned successfully.

> After all it's the point where it failed in a linear forward scan?

The scan does keep going even if individual blocks couldn't be read. There might be a smarter way to handle this, but the behavior has been around a while. Only the error reporting is fairly new.

laanwj:
utACK 4d2d604

PR #10404, opened 2017-05-15
doc: Add logging to FinalizeNode()

sdaftuar:
Disconnecting a peer doesn't update the variables that affect p2p logic until `FinalizeNode()` is called, so it's helpful for debugging to have a log message in this function.

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -303,6 +303,7 @@ void FinalizeNode(NodeId nodeid, bool& fUpdateConnectionTime) {
         assert(nPreferredDownload == 0);
         assert(nPeersWithValidatedDownloads == 0);
     }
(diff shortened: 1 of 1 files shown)


jnewbery:
Tested ACK 1530bfc72d6e3a57cc81980faaf52d9b084353f2.

Minor suggestion: `Cleared nodestate for peer` is a little cryptic without the context that the state is being cleared after disconnect. Perhaps better wording would be something like: `Peer %d fully disconnected. NodeState cleared.`

Either way, this is useful logging, so ACK.

sipa:
utACK 1530bfc72d6e3a57cc81980faaf52d9b084353f2

jonasschnelli:
Seems useful.
utACK 1530bfc72d6e3a57cc81980faaf52d9b084353f2

PR #10405, opened 2017-05-15
tests: Correct testcase in script_tests.json for large number OP_EQUAL

laanwj:
Fix a test case that was passing correctly by accident, but not testing the right thing. Reported by helo on IRC.


jonasschnelli:
utACK 2f84cf66542a34db5bbfacaf9125474d85038bf1

fanquake:
utACK 2f84cf6

PR #10408, opened 2017-05-16
Net: Improvements to Tor control port parser

str4d:
Cherry-picked from the following Zcash PRs:

- https://github.com/zcash/zcash/pull/2177
- https://github.com/zcash/zcash/pull/2251

### DIFF
--- a/src/torcontrol.cpp
+++ b/src/torcontrol.cpp
@@ -330,6 +330,10 @@ static std::pair<bool,std::string> ReadBinaryFile(const fs::path &filename, size
     char buffer[128];
     size_t n;
     while ((n=fread(buffer, 1, sizeof(buffer), f)) > 0) {
+        // Check for reading errors so we don't return any data if we couldn't
+        // read the entire file (or up to maxsize)
+        if (ferror(f))
+            return std::make_pair(false,"");
         retval.append(buffer, buffer+n);
         if (retval.size() > maxsize)
             break;
--- a/src/torcontrol.cpp
+++ b/src/torcontrol.cpp
@@ -277,17 +277,19 @@ static std::map<std::string,std::string> ParseTorReplyMapping(const std::string 
     size_t ptr=0;
     while (ptr < s.size()) {
         std::string key, value;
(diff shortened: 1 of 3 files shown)


src/torcontrol.cpp
```diff
@@ -323,6 +376,10 @@ static std::pair<bool,std::string> ReadBinaryFile(const fs::path &filename, size
     char buffer[128];
     size_t n;
     while ((n=fread(buffer, 1, sizeof(buffer), f)) > 0) {
+        // Check for reading errors so we don't return any data if we couldn't
+        // read the entire file (or up to maxsize)
+        if (ferror(f))
+            return std::make_pair(false,"");
```
laanwj: Wouldn't the fread already return `<=0` in this case? (making it necessary to move the check outside of the loop)

laanwj:
Thanks for the improvements; I still have to review the unescaping loop in detail but overall it looks good to me.

src/torcontrol.cpp
str4d: No, there is [no such guarantee](http://www.cplusplus.com/reference/cstdio/fread/#return):

> If [the return value] differs from the _count_ parameter, either a reading error occurred or the _end-of-file_ was reached while reading. In both cases, the proper indicator is set, which can be checked with ferror and feof, respectively.

So it is entirely possible for `fread` to return a number `0 < retval < count` if a reading error occurs.

str4d: The reason we added this check was the possibility that a read error could occur at byte 33, resulting in 32 bytes of a larger file being returned and used as the cookie (which checks for exact length), in violation of `control-spec`. Thus we only need to defend against the case where `fread` returns a value greater than zero on error; since the buffer is larger than the cookie size, the only remaining way to return a 32-byte read is by reaching EOF.

laanwj: Interesting, I didn't know that, I though that only the EOF case would cause it to return a smaller amount larger or equal to zero (which according to the manual page is the case for `read`, but apparently not `fread`).

laanwj:
utACK 49a199b

PR #10410, opened 2017-05-16
Fix importwallet edge case rescan bug

ryanofsky:
Start importwallet rescans at the first block with timestamp greater or equal
to the wallet birthday instead of the last block with timestamp less or equal.
This fixes an edge case bug where importwallet could fail to start the rescan
early enough if there are blocks with decreasing timestamps or multiple blocks
with the same timestamp.

### DIFF
--- a/src/wallet/test/wallet_tests.cpp
+++ b/src/wallet/test/wallet_tests.cpp
@@ -19,6 +19,8 @@
 #include <univalue.h>
 
 extern UniValue importmulti(const JSONRPCRequest& request);
+extern UniValue dumpwallet(const JSONRPCRequest& request);
+extern UniValue importwallet(const JSONRPCRequest& request);
 
 // how many times to run all the tests to have a chance to catch errors that only show up with particular random shuffles
 #define RUN_TESTS 100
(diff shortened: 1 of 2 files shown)


sipa:
utACK 2a8e35a11d4bd4828631654fc7b8b8fe8f0a2460

laanwj:
Good catch. Makes the code more readable too, nice!

utACK https://github.com/bitcoin/bitcoin/commit/2a8e35a11d4bd4828631654fc7b8b8fe8f0a2460

laanwj:
Adding "needs backport" tag as this is a clear bugfix.

ryanofsky:
@sipa, @laanwj, thanks for reviewing. I also have similar bugfix for `importmulti` in https://github.com/bitcoin/bitcoin/pull/10403 if you want to take a look.

gmaxwell:
oops Guess I should have gone and looked for other cases that needed the earliest at least. Concept ACK.

src/wallet/test/wallet_tests.cpp
```diff
@@ -437,6 +439,66 @@ BOOST_FIXTURE_TEST_CASE(rescan, TestChain100Setup)
...
+{
+    CWallet *pwalletMainBackup = ::pwalletMain;
+    LOCK(cs_main);
+
+    // Create two blocks with same timestamp to verify that importwallet rescan
+    // will pick up both blocks, not just the first.
+    const int64_t BLOCK_TIME = chainActive.Tip()->GetBlockTimeMax() + 5;
+    SetMockTime(BLOCK_TIME);
+    coinbaseTxns.emplace_back(*CreateAndProcessBlock({}, GetScriptForRawPubKey(coinbaseKey.GetPubKey())).vtx[0]);
+    coinbaseTxns.emplace_back(*CreateAndProcessBlock({}, GetScriptForRawPubKey(coinbaseKey.GetPubKey())).vtx[0]);
```
jnewbery: This is a good test, but why not set the mocktime to `BLOCK_TIME - 1` for the second block, or add a third block with block time going backwards? That would make it clearer that the `importwallet()` code can deal with non-monotonic block times.

jnewbery (commented):
Looks good. Tested ACK 2a8e35a11d4bd4828631654fc7b8b8fe8f0a2460. New test fails without the `importwallet()` change.

PR #10413, opened 2017-05-17
Fix docs (there's no rpc command setpaytxfee)

RHavar:


laanwj:
utACK - not only does `settxfee` exist, it seems to be what is intended here, as it sets the `payTxFee` variable.

fanquake:
utACK 0f1b26a

PR #10420, opened 2017-05-18
Add Qt tests for wallet spends & bumpfee

ryanofsky:
A few code changes were needed to accompany the test:

* Adding setObjectName() calls for a few Qt controls to make them easily
  accessible from the test.

* Calling contextMenu->popup() instead of contextMenu->exec() to open
  the transaction list context menu without blocking the test thread.

* Opening the context menu at the contextualMenu event point rather than
  the cursor position (this change was not strictly needed to make the test
  work, but is more correct).

* Updating the bumped transaction row with showTransaction=true instead of
  false. This is needed to prevent the bumped tx from being hidden, so the last
  part of the test which attempts to bump the bumped tx can work. (Technically
  this change is a more general bugfix not limited to the testing environment,
  but the bug doesn't happen outside of the testing environment because in the
  full Qt client, a queued NotifyTransactionChanged notification causes the row
  to be updated twice, first with showTransaction=false, then immediately after
  with showTransaction=true.)

### DIFF
--- a/src/qt/test/wallettests.cpp
+++ b/src/qt/test/wallettests.cpp
@@ -8,26 +8,46 @@
 #include "qt/sendcoinsdialog.h"
 #include "qt/sendcoinsentry.h"
 #include "qt/transactiontablemodel.h"
+#include "qt/transactionview.h"
 #include "qt/walletmodel.h"
 #include "test/test_bitcoin.h"
 #include "validation.h"
 #include "wallet/wallet.h"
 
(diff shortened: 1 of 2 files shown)


laanwj:
utACK  - thanks!

jonasschnelli:
Thanks!
Tested ACK 5749a48821f6a83d17b4f4626d09d6abdada5135

pstratem:
breaks qt4

PR #10421, opened 2017-05-18
[qt] Remove excess logic: Prefer "return foo;" to "if (foo) { return true; } else { return false; }"

practicalswift:
Replace …

```
if (foo) { return true; } else { return false; }
```

… with the equivalent …

```
return foo;
```

The code was introduced in fbf385cc830df2aec7cdcbab0c2b09b46569e8c1 which was merged into `master` earlier today.

Friendly ping @jonasschnelli :-)

### DIFF
--- a/src/qt/walletmodel.cpp
+++ b/src/qt/walletmodel.cpp
@@ -702,9 +702,7 @@ bool WalletModel::transactionSignalsRBF(uint256 hash) const
 {
     LOCK2(cs_main, wallet->cs_wallet);
     const CWalletTx *wtx = wallet->GetWalletTx(hash);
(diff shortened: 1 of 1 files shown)


laanwj:
Can you please add *somewhat* more description to your commit message and PR? Please don't just drop PRs without any description or motivation.

practicalswift:
@laanwj Fixed :-)

jonasschnelli:
ACK e49b868cfe9391ec61b4fedef9811a4be265525d

paveljanik:
utACK e49b868

fanquake:
utACK e49b868

PR #10423, opened 2017-05-18
[tests] skipped tests should clean up after themselves

jnewbery:
If a test is skipped, it doesn't clean up its temp directory. This PR fixes that so skipped tests' temp directories are cleaned up after the test is skipped. The temp directory can be left intact by using the `--nocleanup` argument as usual.

Implementation: Define a SkipTest exception type. Catch it in the test_framework and go through the normal test teardown functionality. Similar to the [SkipTest exception class in Python's standard unittest library](https://docs.python.org/3/library/unittest.html#unittest.SkipTest)

@sdaftuar

### DIFF
--- a/test/functional/test_framework/test_framework.py
+++ b/test/functional/test_framework/test_framework.py
@@ -5,6 +5,7 @@
 """Base class for RPC testing."""
 
 from collections import deque
+from enum import Enum
 import logging
 import optparse
 import os
@@ -41,6 +42,15 @@ from .util import (
 )
 from .authproxy import JSONRPCException
 
+class TestStatus(Enum):
+    PASSED = 1
+    FAILED = 2
+    SKIPPED = 3
+
+TEST_EXIT_PASSED = 0
+TEST_EXIT_FAILED = 1
+TEST_EXIT_SKIPPED = 77
+
 class BitcoinTestFramework(object):
     """Base class for a bitcoin test script.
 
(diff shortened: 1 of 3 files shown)


laanwj:
Concept ACK

jnewbery:
This runs fine on my local pc but something is causing the tests to hang on Travis. I've pushed a temporary commit to increase travis logging.

sdaftuar:
Concept ACK, though I'm also puzzled by the travis failures.

jnewbery:
930deb9b2c5a8234d430f0a739c97e1fc961ffad passes travis (at least on my fork). I think the earlier failure was perhaps spurious.

This should now be ready for review.

sdaftuar:
ACK

MarcoFalke:
utACK 930deb9b2c5a8234d430f0a739c97e1fc961ffad

test/functional/test_framework/test_framework.py
```diff
@@ -139,15 +144,18 @@ def main(self):
...
         try:
             self.setup_chain()
             self.setup_network()
             self.run_test()
-            success = True
+            success = TestStatus.PASSED
         except JSONRPCException as e:
             self.log.exception("JSONRPC error")
+        except SkipTest as e:
+            self.log.warning("Test Skipped: %s" % e.message)
```
jimmysong: This is info elsewhere, should it really be warning?

jimmysong (approved):
Tested ACK. 930deb9b2c5a8234d430f0a739c97e1fc961ffad

test/functional/test_framework/test_framework.py
jnewbery: Yes - this log is replacing the logs in the individual test cases (rpcbind_test.py and zmq_test.py), which are currently warnings.

The reason is that we always want the user to know if a test has been skipped.

jimmysong: OK, I see. I thought the log.info below was also logging something similar. This is fine.

PR #10426, opened 2017-05-18
Replace bytes_serialized with bogosize

sipa:
Current result:

```
  "bogosize": 3846482678,
  "disk_size": 2272310093,
```

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -781,11 +781,12 @@ struct CCoinsStats
     uint256 hashBlock;
     uint64_t nTransactions;
     uint64_t nTransactionOutputs;
+    uint64_t nBogoSize;
     uint256 hashSerialized;
     uint64_t nDiskSize;
     CAmount nTotalAmount;
 
-    CCoinsStats() : nHeight(0), nTransactions(0), nTransactionOutputs(0), nTotalAmount(0) {}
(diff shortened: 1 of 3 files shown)


fanquake:
Looks like you'll need to update the blockchain RPC test [here](https://github.com/bitcoin/bitcoin/blob/master/test/functional/blockchain.py#L52).

sipa:
Fixed.

MarcoFalke:
Breaking changes need to be mentioned in the release notes. Mind to do it
in this pull?

sipa:
I've rebased this PR on top of #10396, as both touch the same code. A single combined release-notes.md entry was added.

sipa:
Rebased, and updated the release notes a bit.

sipa:
Rebased on top of #10195.

src/rpc/blockchain.cpp
```diff
@@ -904,6 +907,7 @@ UniValue gettxoutsetinfo(const JSONRPCRequest& request)
             "  \"bestblock\": \"hex\",   (string) the best block hash hex\n"
             "  \"transactions\": n,      (numeric) The number of transactions\n"
             "  \"txouts\": n,            (numeric) The number of output transactions\n"
+            "  \"bogosize\": n,          (numeric) A meaningless metric for UTXO set size\n"
             "  \"hash_serialized\": \"hash\",   (string) The serialized hash\n"
```
MarcoFalke: Could you suffix the name here with `2` while touching the file anyway?

sipa: Done.

laanwj:
tested ACK 8b22af3

PR #10429, opened 2017-05-19
tests: fix spurious addrman test failure

theuni:
There have been a few PRs lately that have failed checks due to a bug introduced in #10287. The first check happened to work already because the chosen addresses didn't collide. The ```addrman.Clear()``` after that check causes addrman's salt to be reset, making all future tests non-deterministic.

When inserting two addresses of the same class, from the same source, there's a 1/64 chance that their hashes will collide. So it's faulty logic to be checking for an exact size. @sipa, mind verifying that?

I'd like to fix this properly by ensuring determinism in the test as a next step, but I'd rather fix the spurious failures asap.

### DIFF
--- a/src/test/addrman_tests.cpp
+++ b/src/test/addrman_tests.cpp
@@ -104,10 +104,14 @@ BOOST_AUTO_TEST_CASE(addrman_simple)
 
 
     // Test: New table has one addr and we add a diff addr we should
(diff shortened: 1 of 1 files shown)


laanwj:
utACK 6b51b0a, had noticed this one as well, thanks for fixing

PR #9279, opened 2016-12-05
Consensus: Move CFeeRate out of libconsensus

jtimon:
CTxOut and CTransaction shouldn't depend on CFeeRate, which is something for policy checks and not consensus checks.
This replaces #7820

PR #10431, opened 2017-05-19
Prevent shadowing the global dustRelayFee

paveljanik:
@jtimon @morcos As discussed on IRC: `dustRelayFee` shadows the global `::dustRelayFee` (origin in #9279).

It is always called with `::dustRelayFee`, but we want to keep it is an argument for later wallet use so it was renamed (using `In` suffix) instead of removed. Comments are unchanged though because `dustRelayFee ` is a name of the concept there.

### DIFF
--- a/src/policy/policy.cpp
+++ b/src/policy/policy.cpp
@@ -15,7 +15,7 @@
 
 #include <boost/foreach.hpp>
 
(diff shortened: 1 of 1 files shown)


dcousens (approved):

morcos:
utACK 43c5877

jtimon:
utACK 43c5877

fanquake:
utACK 43c5877

PR #10432, opened 2017-05-19
[Trivial] Add BITCOIN_FS_H endif footer in fs.h

ghost:
Adding BITCOIN_FS_H as a footer as per header conventions


laanwj:
utACK 2300a5e

PR #10433, opened 2017-05-19
[tests] improve tmpdir structure

jnewbery:
tl;dr:

- temp directories now include the name of the test when run through the test_runner
- passing a `--tmpdirprefix` to the test_runner allows test_runner to be called by cron/build jobs without risk of directory name collision

==================

By default, the functional tests create their temp directories in the $TMPDIR as follows:

```
$TMPDIR
   |
   -- test<rand_chars>
        |
        -- <port seed>
            |
            -- temp files
            -- ...
```

a `--tmpdir` argument can be passed into the individual test case, which causes the temp directory to be written to:

```
<specified --tmpdir>
   |
   -- <port seed>
        |
        -- temp files
        -- ...
```

If `--tmpdir` is passed to the test_runner, then it gets passed down to the indivdual tests and the directory structure is:

```
<specified --tmpdir>
   |
   -- <port seed 1>
        |
        -- temp files
        -- ...
   -- <port seed 2>
        |
        -- temp files
        -- ...
   -- ...

```

There are a few problems:

- the directory names give no indication of which test was being run, so it's difficult to quickly find the temp directory for an indvidual test if using the test_runner
- if a `--tmpdir` directory is passed to the test_runner and there are already directories in that tmpdir with names that clash with the port seed, then tests will fail when they try to use that subdirectory. This can be a problem when running test_runner as a cron job or build job with a specified tmpdir. Test failures in one build will cause subsequent runs to fail because the subdirectories are not cleaned up
- even if a tmpdir is specified, test_framework will still create an empty temp directory `$TMPDIR/test<rand_chars>` (this is because `tempfile.mkdtemp(prefix="test")` is called even when `--tmpdir`  is used).

This PR changes the temp dir structure so that:

- if running the test directly:
    - if no `--tmpdir` is specified
        - temp directory is `$TMPDIR/test<rand_chars>` (as now, except without the `<port seed>` subdirectory)
    - if a `--tmpdir` is specified
        - temp directory is the specified tmpdir (as now, except without the `<port seed>` subdirectory)
- if running through the test_runner
    - if no `--tmpdirprefix` is specified
        - temp directory is `$TMPDIR/bitcoin_test_runner_<date>_<time>/<test_name>_portseed`.
    - if a `--tmpdirprefix` is specified
        - temp directory is `<tmpdirprefix>/bitcoin_test_runner_<date>_<time>/<test_name>_portseed`.

`<tmpdirprefix>` will be cleaned up by the test_runner if it is empty at the end of all tests (ie if all tests have cleaned up their own temp directories).

### DIFF
--- a/test/functional/test_framework/test_framework.py
+++ b/test/functional/test_framework/test_framework.py
@@ -109,8 +109,7 @@ class BitcoinTestFramework(object):
                           help="Source directory containing bitcoind/bitcoin-cli (default: %default)")
         parser.add_option("--cachedir", dest="cachedir", default=os.path.normpath(os.path.dirname(os.path.realpath(__file__))+"/../../cache"),
                           help="Directory for caching pregenerated datadirs")
-        parser.add_option("--tmpdir", dest="tmpdir", default=tempfile.mkdtemp(prefix="test"),
-                          help="Root directory for datadirs")
+        parser.add_option("--tmpdir", dest="tmpdir", help="Root directory for datadirs")
         parser.add_option("-l", "--loglevel", dest="loglevel", default="INFO",
                           help="log events at this level and higher to the console. Can be set to DEBUG, INFO, WARNING, ERROR or CRITICAL. Passing --loglevel DEBUG will output all logs to console. Note that logs at all levels are always written to the test_framework.log file in the temporary test directory.")
         parser.add_option("--tracerpc", dest="trace_rpc", default=False, action="store_true",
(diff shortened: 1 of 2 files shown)


sdaftuar:
ACK, this looks great!

laanwj:
Concept ACK

kallewoof:
utACK b040243cc3dfd69f5a8b6e217d1a658d59b15809

MarcoFalke:
utACK b040243cc3dfd69f5a8b6e217d1a658d59b15809

PR #10434, opened 2017-05-20
[WIP] 3072-bit MuHash based hash_serialized

sipa:
This implements a 3072-bit MuHash discussed on https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2017-May/014337.html as a replacement for the hash_serialized field in `gettxoutsetinfo`.

This is an order-independent hash, allowing it to be computed either by iterating over the UTXO set in non-sorted order. It also supports incremental addition and deletion of entries from the hash, allowing it to be updated on the fly for each block. Neither of these approaches is currently implemented.

### DIFF
--- a/src/crypto/muhash.cpp
+++ b/src/crypto/muhash.cpp
@@ -28,6 +28,54 @@ namespace {
     c1 = 0; \
 }
 
+#if defined(__amd64__) || defined(__x86_64__)
+
+/** [c0,c1] = a * b */
+#define mul(c0,c1,a,b) { \
+    __asm__ ("mulq %3" : "=d"(c1), "=a"(c0) : "a"(a), "g"(b) : "cc"); \
+}
+
+/** [c0,c1,c2] += a * b */
+#define muladd3(c0,c1,c2,a,b) { \
+    uint64_t tl, th; \
+    __asm__ ("mulq %3" : "=a"(tl), "=d"(th) : "a"(a), "g"(b) : "cc"); \
+    __asm__ ("addq %3,%0; adcq %4,%1; adcq $0,%2" : "+r"(c0), "+r"(c1), "+r"(c2) : "a"(tl), "d"(th) : "cc"); \
+}
+
+/** [c0,c1,c2] += 2 * a * b */
+#define muldbladd3(c0,c1,c2,a,b) { \
+    uint64_t tl, th; \
+    __asm__ ("mulq %3" : "=a"(tl), "=d"(th) : "a"(a), "g"(b) : "cc"); \
+    __asm__ ("addq %3,%0; adcq %4,%1; adcq $0,%2" : "+r"(c0), "+r"(c1), "+r"(c2) : "a"(tl), "d"(th) : "cc"); \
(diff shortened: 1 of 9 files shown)


src/crypto/muhash.cpp
```diff
@@ -0,0 +1,329 @@
+#include "muhash.h"
```
paveljanik: Copyright header missing

src/test/crypto_tests.cpp
```diff
@@ -507,4 +508,58 @@ BOOST_AUTO_TEST_CASE(countbits_tests)
...
+}
+
+BOOST_AUTO_TEST_CASE(muhash_tests)
+{
+    unsigned char out[384];
+
+    for (int i = 0; i < 10; ++i) {
+        unsigned char res[384];
+        int table[4];
+        for (int i = 0; i < 4; ++i) {
```
paveljanik: `i` shadowing.

src/test/crypto_tests.cpp
```diff
@@ -507,4 +508,58 @@ BOOST_AUTO_TEST_CASE(countbits_tests)
...
+            table[i] = insecure_rand_ctx.randbits(3);
+        }
+        for (int order = 0; order < 4; ++order) {
+            MuHash3072 acc;
+            for (int i = 0; i < 4; ++i) {
```
paveljanik: `i` shadowing

src/test/crypto_tests.cpp
```diff
@@ -507,4 +508,58 @@ BOOST_AUTO_TEST_CASE(countbits_tests)
...
+
+        MuHash3072 x = FromInt(insecure_rand_ctx.randbits(4)); // x=X
+        MuHash3072 y = FromInt(insecure_rand_ctx.randbits(4)); // x=X
+        MuHash3072 z; // x=X, y=Y, z=1
+        z *= x; // x=X, y=Y, z=X
+        z /= y; // x=X, y=Y, z=X/Y
+        y /= x; // x=X, y=Y/X, z=X/Y
+        z *= y; // x=X, y=Y/X, z=1
+        z.Finalize(out);
+        for (int i = 0; i < 384; ++i) {
```
paveljanik: `i` shadowing

paveljanik (commented):
Release notes entry needed.

src/crypto/muhash.cpp
sipa: Fixed.

src/test/crypto_tests.cpp
sipa: Fixed.

src/test/crypto_tests.cpp
sipa: Fixed.

n1bor:
@sipa - did this POC of chainstate only sync:
https://gist.github.com/n1bor/d5b0330a9addb0bf5e0f869518883522
https://botbot.me/freenode/bitcoin-core-dev/2016-10-24/?msg=75398575&page=3

Worked well - syncs as fast at you can download chainstate db (30mins for 2.8gig chainstate). Network was the bottleneck.
Should be able to rework to use MuHash easily - and rethink the snapshot interval/process.
But if consensus is to NEVER commit chain-state to blockchain - than is all a little pointless. gmaxwell did suggest an expiring consensus rule so if want to change from MuHash would be easy - which would do.

sipa:
Closing as stale.

PR #10437, opened 2017-05-21
[WIP] Implement BIP135 (generalized versionbits)

sanch0panza:
This is an implementation of https://github.com/bitcoin/bips/blob/master/bip-0135.mediawiki .

It extends the BIP9 state machine processing with configurable per-bit window size, activation threshold and grace period parameters (minlockedblocks and minlockedtime).

The built-in deployments are parameterized to maintain backward compatibility, and the existing BIP9 tests are retained unmodified except for a minor fix made to `versionbits_tests` to improve the disjointness checks for configured bits).

A `bip135_forks` section has been added to the `getblockchaininfo` RPC output, leaving the existing `bip9_softforks` section for backward compatibility.

The check for 'unknown versions being mined' has been altered to take into account that for unknown bits, we can no longer rely on a 95% activation threshold. The `p2p-versionbits-warning` test has been modified accordingly - the new logic warns when an unknown bit exceeds 50/100 of the last blocks.

NOTE: This implementation contains a specific feature which is not covered by the specification (and thus not strictly required for BIP135): the optional loading of fork configuration from a CSV file (using `-forks=datafile` command line option) and the ability to dump out the built-in configuration in CSV format (`-dumpforks` option). This has been retained from the original reference implementation since it makes testing and adaptation of the configuration easier.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -1,10 +1,11 @@
 // Copyright (c) 2010 Satoshi Nakamoto
-// Copyright (c) 2009-2016 The Bitcoin Core developers
+// Copyright (c) 2009-2017 The Bitcoin Core developers
 // Distributed under the MIT software license, see the accompanying
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
 #include "chainparams.h"
 #include "consensus/merkle.h"
+#include "versionbits.h"  // bip135 added
 
 #include "tinyformat.h"
 #include "util.h"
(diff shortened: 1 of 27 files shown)


sanch0panza:
Could someone advise why the new subtree would not be visible to the Travis build?
I don't have a problem building and testing this locally.

jonasschnelli:
Why use CSV?
We already have an internal JSON parser (UniValue) or a flexible configuration file reader...

sanch0panza:
@jonasschnelli : I used CSV as it seemed more suited to the type of configuration, with all data for one deployment fitting compactly on a single line, and most parsers can parse a comment header so it met the "self-documenting" request received during BIP135 review.

But I agree, JSON could well be used.

In fact, the loading of data from a file could be removed entirely to streamline this implementation - it is something that an implementation can choose to do whichever way it wants. Others have suggested to be able to specify it using multiple `-bip135=<data>` arguments, which would be another option.

I did not really want to mix this fork configuration with the users existing config file, as it seems something which which ordinary users should not really tinker.

sanch0panza:
I would like to include some more improvement commits based on code review feedback I received so far (on code parts which are common to https://github.com/BitcoinUnlimited/BitcoinUnlimited/pull/458) .

I'll keep these as separate commits on top of this branch for now, for easier separate review of these additional changes.

jtimon (changes requested):

src/chainparams.h
```diff
@@ -58,6 +58,8 @@ class CChainParams
     };
 
     const Consensus::Params& GetConsensus() const { return consensus; }
+    /** Modifiable consensus parameters added by bip135 */
+    Consensus::Params& GetModifiableConsensus() { return consensus; }
```
sanch0panza: I agree, this can probably be improved by re-using existing code.
Will look into that - I'm not so familiar with the recent Core code and so used some methods from existing implementation.

src/versionbits.h
```diff
@@ -16,39 +16,47 @@ static const int32_t VERSIONBITS_TOP_BITS = 0x20000000UL;
...
 /** Total bits available for versionbits */
 static const int32_t VERSIONBITS_NUM_BITS = 29;
+/** Size of window to use for assessing warning of unknown bits */
+static const int BIT_WARNING_WINDOW = 100;
+/** Threshold to use for assessing warning of unknown bits */
+static const int BIT_WARNING_THRESHOLD = 50;
 
+// bip135: assigned numbers to these enum values
 enum ThresholdState {
-    THRESHOLD_DEFINED,
```
sanch0panza: Agreed - will remove.
I might have wanted to set these to known values for debugging purposes.

src/consensus/params.h
```diff
@@ -1,5 +1,5 @@
 // Copyright (c) 2009-2010 Satoshi Nakamoto
-// Copyright (c) 2009-2016 The Bitcoin Core developers
+// Copyright (c) 2009-2017 The Bitcoin Core developers
```
sanch0panza: I have a habit of updating the copyright dates whenever I modify a file.
What do you mean by done separately - should I not update them?

src/chainparams.cpp
```diff
@@ -339,6 +357,25 @@ const CChainParams &Params() {
     return *globalChainParams;
 }
 
+CChainParams& Params(const std::string& chain)
```
sanch0panza: You're right, this function should not be needed and I will re-use CreateChainParams.
The params created in Params() are not explicitly destroyed.

sanch0panza:
Changing this to [WIP] ahead of a rebasing which will contain GetStateStatisticsFor() from https://github.com/bitcoin/bitcoin/commit/557c9a68fb72aeb535f2efe3cc82d3f5e66c6ad3 which needs to be checked, maybe it requires further adaptation to BIP135 .

sanch0panza:
Rebased.

The review comments addressing the "Modifiable" functions are still work in progress, as is revising the BIP9 statistics that were introduced by the rebase.

PR #10443, opened 2017-05-22
Add fee_est tool for debugging fee estimation code

ryanofsky:
This PR adds an `-estlog` option for saving live fee estimation data from a bitcoin node, and a `fee_est` command line tool for processing the data and testing fee estimation code.

The idea is to make it easier to test improvements to fee estimation like https://github.com/bitcoin/bitcoin/pull/10199 in a more systematic and reproducible way.

Some documentation is in [`src/test/fee_est/README.md`](https://github.com/ryanofsky/bitcoin/blob/pr/fee/src/test/fee_est/README.md)

Sample log file: [`est.log.xz`](https://storage.googleapis.com/ryanofsky/est.log.xz) (65M)

Sample [`fee_est.cpp`](https://github.com/ryanofsky/bitcoin/blob/pr/fee/src/test/fee_est/fee_est.cpp)  graph output: 

graph

### DIFF
--- a/src/policy/fees.cpp
+++ b/src/policy/fees.cpp
@@ -527,8 +527,7 @@ bool CBlockPolicyEstimator::_removeTx(const uint256& hash, bool inBlock)
     }
 }
 
-CBlockPolicyEstimator::CBlockPolicyEstimator(const fs::path& estimation_filepath)
-    : m_estimation_filepath{estimation_filepath}
+CBlockPolicyEstimator::CBlockPolicyEstimator()
 {
     static_assert(MIN_BUCKET_FEERATE > 0, "Min feerate must be nonzero");
     size_t bucketIndex = 0;
@@ -544,21 +543,17 @@ CBlockPolicyEstimator::CBlockPolicyEstimator(const fs::path& estimation_filepath
     feeStats = std::unique_ptr<TxConfirmStats>(new TxConfirmStats(buckets, bucketMap, MED_BLOCK_PERIODS, MED_DECAY, MED_SCALE));
     shortStats = std::unique_ptr<TxConfirmStats>(new TxConfirmStats(buckets, bucketMap, SHORT_BLOCK_PERIODS, SHORT_DECAY, SHORT_SCALE));
     longStats = std::unique_ptr<TxConfirmStats>(new TxConfirmStats(buckets, bucketMap, LONG_BLOCK_PERIODS, LONG_DECAY, LONG_SCALE));
-
-    // If the fee estimation file is present, read recorded estimations
-    AutoFile est_file{fsbridge::fopen(m_estimation_filepath, "rb")};
-    if (est_file.IsNull() || !Read(est_file)) {
-        LogPrintf("Failed to read fee estimates from %s. Continue anyway.\n", fs::PathToString(m_estimation_filepath));
-    }
 }
 
 CBlockPolicyEstimator::~CBlockPolicyEstimator() = default;
 
-void CBlockPolicyEstimator::processTransaction(const CTxMemPoolEntry& entry, bool validFeeEstimate)
+void CBlockPolicyEstimator::processTx(const uint256& hash,
+    unsigned int txHeight,
+    CAmount fee,
+    uint32_t size,
+    bool validFeeEstimate)
 {
     LOCK(m_cs_fee_estimator);
-    unsigned int txHeight = entry.GetHeight();
-    uint256 hash = entry.GetTx().GetHash();
     if (mapMemPoolTxs.count(hash)) {
         LogPrint(BCLog::ESTIMATEFEE, "Blockpolicy error mempool tx %s already being tracked\n",
                  hash.ToString());
@@ -582,7 +577,7 @@ void CBlockPolicyEstimator::processTransaction(const CTxMemPoolEntry& entry, boo
     trackedTxs++;
 
     // Feerates are stored and reported as BTC-per-kb:
-    CFeeRate feeRate(entry.GetFee(), entry.GetTxSize());
+    CFeeRate feeRate(fee, size);
 
     mapMemPoolTxs[hash].blockHeight = txHeight;
     unsigned int bucketIndex = feeStats->NewTx(txHeight, (double)feeRate.GetFeePerK());
@@ -593,10 +588,14 @@ void CBlockPolicyEstimator::processTransaction(const CTxMemPoolEntry& entry, boo
     assert(bucketIndex == bucketIndex3);
 }
 
-bool CBlockPolicyEstimator::processBlockTx(unsigned int nBlockHeight, const CTxMemPoolEntry* entry)
+bool CBlockPolicyEstimator::processBlockTx(unsigned int nBlockHeight,
+    const uint256& hash,
+    unsigned int height,
+    CAmount fee,
+    uint32_t size)
 {
     AssertLockHeld(m_cs_fee_estimator);
-    if (!_removeTx(entry->GetTx().GetHash(), true)) {
+    if (!_removeTx(hash, true)) {
         // This transaction wasn't being tracked for fee estimation
         return false;
     }
@@ -604,7 +603,7 @@ bool CBlockPolicyEstimator::processBlockTx(unsigned int nBlockHeight, const CTxM
     // How many blocks did it take for miners to include this transaction?
     // blocksToConfirm is 1-based, so a transaction included in the earliest
     // possible block has confirmation count of 1
-    int blocksToConfirm = nBlockHeight - entry->GetHeight();
+    int blocksToConfirm = nBlockHeight - height;
     if (blocksToConfirm <= 0) {
         // This can't happen because we don't process transactions from a block with a height
         // lower than our greatest seen height
@@ -613,7 +612,7 @@ bool CBlockPolicyEstimator::processBlockTx(unsigned int nBlockHeight, const CTxM
     }
 
     // Feerates are stored and reported as BTC-per-kb:
-    CFeeRate feeRate(entry->GetFee(), entry->GetTxSize());
+    CFeeRate feeRate(fee, size);
 
     feeStats->Record(blocksToConfirm, (double)feeRate.GetFeePerK());
     shortStats->Record(blocksToConfirm, (double)feeRate.GetFeePerK());
@@ -621,8 +620,7 @@ bool CBlockPolicyEstimator::processBlockTx(unsigned int nBlockHeight, const CTxM
     return true;
 }
 
-void CBlockPolicyEstimator::processBlock(unsigned int nBlockHeight,
-                                         std::vector<const CTxMemPoolEntry*>& entries)
+void CBlockPolicyEstimator::processBlock(unsigned int nBlockHeight, const AddTxsFn& add_txs)
 {
     LOCK(m_cs_fee_estimator);
     if (nBlockHeight <= nBestSeenHeight) {
@@ -650,11 +648,13 @@ void CBlockPolicyEstimator::processBlock(unsigned int nBlockHeight,
     longStats->UpdateMovingAverages();
 
     unsigned int countedTxs = 0;
+    unsigned int block_txs = 0;
     // Update averages with data points from current block
-    for (const auto& entry : entries) {
-        if (processBlockTx(nBlockHeight, entry))
-            countedTxs++;
-    }
+    add_txs([&](const uint256& hash, unsigned int height, CAmount fee, uint32_t size)
+        EXCLUSIVE_LOCKS_REQUIRED(m_cs_fee_estimator) {
+            if (processBlockTx(nBlockHeight, hash, height, fee, size)) ++countedTxs;
+            ++block_txs;
+        });
 
     if (firstRecordedHeight == 0 && countedTxs > 0) {
         firstRecordedHeight = nBestSeenHeight;
@@ -663,7 +663,7 @@ void CBlockPolicyEstimator::processBlock(unsigned int nBlockHeight,
 
 
     LogPrint(BCLog::ESTIMATEFEE, "Blockpolicy estimates updated by %u of %u block txs, since last block %u of %u tracked, mempool map size %u, max target %u from %s\n",
-             countedTxs, entries.size(), trackedTxs, trackedTxs + untrackedTxs, mapMemPoolTxs.size(),
+             countedTxs, block_txs, trackedTxs, trackedTxs + untrackedTxs, mapMemPoolTxs.size(),
              MaxUsableEstimate(), HistoricalBlockSpan() > BlockSpan() ? "historical" : "current");
 
     trackedTxs = 0;
@@ -754,6 +754,29 @@ unsigned int CBlockPolicyEstimator::MaxUsableEstimate() const
 {
     // Block spans are divided by 2 to make sure there are enough potential failing data points for the estimate
     return std::min(longStats->GetMaxConfirms(), std::max(BlockSpan(), HistoricalBlockSpan()) / 2);
+}
+
+unsigned int CBlockPolicyEstimator::getMaxTarget() const
+{
+    LOCK(m_cs_fee_estimator);
+    return MaxUsableEstimate();
+}
+
+// static
+std::vector<unsigned int> CBlockPolicyEstimator::GetUniqueTargets()
+{
+    std::set<unsigned int> targets;
+    auto addTargets = [&](unsigned int numTargets, unsigned int scale) {
+        int target = scale;
+        for (unsigned int i = 1; i < numTargets; ++i) {
+            targets.emplace(target);
+            target += scale;
+        }
+    };
+    addTargets(SHORT_BLOCK_PERIODS, SHORT_SCALE);
+    addTargets(MED_BLOCK_PERIODS, MED_SCALE);
+    addTargets(LONG_BLOCK_PERIODS, LONG_SCALE);
+    return {targets.begin(), targets.end()};
 }
 
 /** Return a fee estimate at the required successThreshold from the shortest
@@ -899,15 +922,6 @@ CFeeRate CBlockPolicyEstimator::estimateSmartFee(int confTarget, FeeCalculation 
     if (median < 0) return CFeeRate(0); // error condition
 
     return CFeeRate(llround(median));
-}
-
-void CBlockPolicyEstimator::Flush() {
-    FlushUnconfirmed();
-
-    AutoFile est_file{fsbridge::fopen(m_estimation_filepath, "wb")};
-    if (est_file.IsNull() || !Write(est_file)) {
-        LogPrintf("Failed to write fee estimates to %s. Continue anyway.\n", fs::PathToString(m_estimation_filepath));
-    }
 }
 
 bool CBlockPolicyEstimator::Write(AutoFile& fileout) const
--- a/src/policy/fees.h
+++ b/src/policy/fees.h
@@ -14,6 +14,7 @@
 #include <util/fs.h>
 
 #include <array>
+#include <functional>
 #include <map>
 #include <memory>
 #include <set>
@@ -80,6 +81,9 @@ struct FeeCalculation
     int desiredTarget = 0;
     int returnedTarget = 0;
 };
+
+using AddTxFn = std::function<void(const uint256& hash, unsigned int height, CAmount fee, uint32_t size)>;
+using AddTxsFn = std::function<void(const AddTxFn&)>;
 
 /** \class CBlockPolicyEstimator
  * The BlockPolicyEstimator is used for estimating the feerate needed
@@ -179,20 +183,17 @@ private:
      * Therefore it makes sense to exponentially space the buckets
      */
     static constexpr double FEE_SPACING = 1.05;
-
-    const fs::path m_estimation_filepath;
 public:
     /** Create new BlockPolicyEstimator and initialize stats tracking classes with default values */
-    CBlockPolicyEstimator(const fs::path& estimation_filepath);
+    CBlockPolicyEstimator();
     ~CBlockPolicyEstimator();
 
     /** Process all the transactions that have been included in a block */
-    void processBlock(unsigned int nBlockHeight,
-                      std::vector<const CTxMemPoolEntry*>& entries)
+    void processBlock(unsigned int nBlockHeight, const AddTxsFn& add_txs)
         EXCLUSIVE_LOCKS_REQUIRED(!m_cs_fee_estimator);
 
     /** Process a transaction accepted to the mempool*/
-    void processTransaction(const CTxMemPoolEntry& entry, bool validFeeEstimate)
+    void processTx(const uint256& hash, unsigned int txHeight, CAmount fee, uint32_t size, bool validFeeEstimate)
         EXCLUSIVE_LOCKS_REQUIRED(!m_cs_fee_estimator);
 
     /** Remove a transaction from the mempool tracking stats*/
@@ -231,12 +232,18 @@ public:
     void FlushUnconfirmed()
         EXCLUSIVE_LOCKS_REQUIRED(!m_cs_fee_estimator);
 
-    /** Calculation of highest target that estimates are tracked for */
-    unsigned int HighestTargetTracked(FeeEstimateHorizon horizon) const
+    /** Get highest target that reasonable estimate can be provided for. */
+    unsigned int getMaxTarget() const
         EXCLUSIVE_LOCKS_REQUIRED(!m_cs_fee_estimator);
 
-    /** Drop still unconfirmed transactions and record current estimations, if the fee estimation file is present. */
-    void Flush()
+    /**
+     * Get sorted list of targets the estimator can directly calculate feerates
+     * for (without rounding up the nearest supported target).
+     */
+    static std::vector<unsigned int> GetUniqueTargets();
+
+    /** Calculation of highest target that estimates are tracked for */
+    unsigned int HighestTargetTracked(FeeEstimateHorizon horizon) const
         EXCLUSIVE_LOCKS_REQUIRED(!m_cs_fee_estimator);
 
 private:
@@ -269,7 +276,7 @@ private:
     std::map<double, unsigned int> bucketMap GUARDED_BY(m_cs_fee_estimator); // Map of bucket upper-bound to index into all vectors by bucket
 
     /** Process a transaction confirmed in a block*/
-    bool processBlockTx(unsigned int nBlockHeight, const CTxMemPoolEntry* entry) EXCLUSIVE_LOCKS_REQUIRED(m_cs_fee_estimator);
+    bool processBlockTx(unsigned int nBlockHeight, const uint256& hash, unsigned int height, CAmount fee, uint32_t size) EXCLUSIVE_LOCKS_REQUIRED(m_cs_fee_estimator);
 
     /** Helper for estimateSmartFee */
     double estimateCombinedFee(unsigned int confTarget, double successThreshold, bool checkShorterHorizon, EstimationResult *result) const EXCLUSIVE_LOCKS_REQUIRED(m_cs_fee_estimator);
--- a/src/policy/fees_input.cpp
+++ b/src/policy/fees_input.cpp
@@ -1,0 +1,236 @@
+// Copyright (c) 2018 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include <policy/fees_input.h>
+
+#include <chainparams.h>
+#include <clientversion.h>
+#include <policy/fees.h>
+#include <streams.h>
+#include <util/strencodings.h>
+#include <util/system.h>
+#include <util/time.h>
+#include <univalue.h>
+
+#include <fstream>
+
+namespace {
+UniValue TxLog(const uint256& hash, unsigned int height, CAmount fee, uint32_t size)
+{
+    UniValue tx(UniValue::VOBJ);
+    tx.pushKV("hash", hash.ToString());
+    tx.pushKV("height", int(height));
+    tx.pushKV("fee", fee);
+    tx.pushKV("size", uint64_t(size));
+    return tx;
+}
+} // namespace
+
+FeeEstInput::FeeEstInput(CBlockPolicyEstimator& estimator) : m_estimator(estimator) {}
+
+bool FeeEstInput::open(const fs::path& estimation_filepath, const fs::path& log_filepath) {
+    if (!writeLog(log_filepath)) return false;
+    m_estimation_filepath = estimation_filepath;
+
+    // If the fee estimation file is present, read recorded estimations
+    if (!readData(m_estimation_filepath)) {
+        LogPrintf("Failed to read fee estimates from %s. Continue anyway.\n", fs::PathToString(m_estimation_filepath));
+    }
+
+    return true;
+}
+
+bool FeeEstInput::close()
+{
+    m_estimator.FlushUnconfirmed();
+
+    if (!writeData(m_estimation_filepath)) {
+        LogPrintf("Failed to write fee estimates to %s. Continue anyway.\n", fs::PathToString(m_estimation_filepath));
+    }
+    return writeLog({});
+}
+
+void FeeEstInput::processTx(const uint256& hash, unsigned int height, CAmount fee, uint32_t size, bool valid)
+{
+    m_estimator.processTx(hash, height, fee, size, valid);
+    if (m_log) {
+        UniValue value(UniValue::VOBJ);
+        value.pushKV("tx", TxLog(hash, height, fee, size));
+        value.pushKV("valid", UniValue(valid));
+        value.pushKV("time", GetTime());
+        *m_log << value.write() << std::endl;
+    }
+}
+
+void FeeEstInput::processBlock(unsigned int block_height, const AddTxsFn& add_txs)
+{
+    UniValue json_txs{UniValue::VARR};
+    m_estimator.processBlock(block_height, [&](const AddTxFn& add_tx) {
+        add_txs([&](const uint256& hash, unsigned int height, CAmount fee, uint32_t size) {
+            add_tx(hash, height, fee, size);
+            if (m_log) json_txs.push_back(TxLog(hash, height, fee, size));
+        });
+    });
+    if (m_log) {
+        UniValue json(UniValue::VOBJ);
+        UniValue json_block(UniValue::VOBJ);
+        json_block.pushKV("height", int64_t(block_height));
+        json.pushKV("block", json_block);
+        json.pushKV("txs", std::move(json_txs));
+        json.pushKV("time", GetTime());
+        *m_log << json.write() << std::endl;
+    }
+}
+
+void FeeEstInput::removeTx(const uint256& hash, bool in_block)
+{
+    m_estimator.removeTx(hash, in_block);
+
+    if (m_log) {
+        UniValue value(UniValue::VOBJ);
+        UniValue removeTx(UniValue::VOBJ);
+        removeTx.pushKV("hash", hash.ToString());
+        removeTx.pushKV("inBlock", UniValue(in_block));
+        value.pushKV("removeTx", removeTx);
+        value.pushKV("time", GetTime());
+        *m_log << value.write() << std::endl;
+    }
+}
+
+bool FeeEstInput::writeData(const fs::path& filepath)
+{
+    AutoFile file{fsbridge::fopen(filepath, "wb")};
+    if (file.IsNull()) return false;
+    m_estimator.Write(file);
+
+    if (m_log) {
+        UniValue value(UniValue::VOBJ);
+        UniValue flush(UniValue::VARR);
+        value.pushKV("flush", flush);
+        value.pushKV("time", GetTime());
+        *m_log << value.write() << std::endl;
+    }
+    return true;
+}
+
+bool FeeEstInput::readData(const fs::path& filepath)
+{
+    AutoFile file{fsbridge::fopen(filepath, "rb")};
+    if (file.IsNull() || !m_estimator.Read(file)) return false;
+
+    if (m_log) {
+        UniValue value(UniValue::VOBJ);
+        std::ifstream data(filepath, std::ifstream::binary);
+        value.pushKV(
+            "read", HexStr(std::string(std::istreambuf_iterator<char>(data), std::istreambuf_iterator<char>())));
+        value.pushKV("time", GetTime());
+        *m_log << value.write() << std::endl;
+    }
+    return true;
+}
+
+bool FeeEstInput::writeLog(const fs::path& filepath)
+{
+    if (m_log) {
+        UniValue value(UniValue::VOBJ);
+        value.pushKV("stop", Params().NetworkIDString());
+        value.pushKV("time", GetTime());
+        *m_log << value.write() << std::endl;
+    }
+
+    if (filepath.empty()) {
+        m_log.reset();
+    } else {
+        m_log = std::make_unique<std::ofstream>(filepath, std::ofstream::out | std::ofstream::app);
+        if (!*m_log) {
+            m_log.reset();
+            return false;
+        }
+    }
+
+    if (m_log) {
+        UniValue value(UniValue::VOBJ);
+        value.pushKV("start", Params().NetworkIDString());
+        value.pushKV("time", GetTime());
+        *m_log << value.write() << std::endl;
+    }
+
+    return true;
+}
+
+bool FeeEstInput::readLog(const fs::path& filepath, const std::function<bool(UniValue&)>& filter)
+{
+    std::ifstream file(filepath);
+    if (!file) {
+        LogPrintf("%s: Failed to open log file %s\n", __func__, fs::PathToString(filepath));
+        return false;
+    }
+
+    std::string line;
+    while (std::getline(file, line)) {
+        UniValue value;
+        if (!value.read(line)) {
+            throw std::runtime_error("Failed to parse fee estimate log line.");
+        }
+
+        if (filter && !filter(value)) {
+            continue;
+        }
+
+        const UniValue& tx = value["tx"];
+        if (tx.isObject()) {
+            m_estimator.processTx(uint256S(tx["hash"].get_str()), tx["height"].getInt<int>(), tx["fee"].getInt<int64_t>(),
+                tx["size"].getInt<uint32_t>(), value["valid"].get_bool());
+            continue;
+        }
+
+        const UniValue& block = value["block"];
+        if (block.isObject()) {
+            int height = block["height"].getInt<int>();
+            m_estimator.processBlock(height, [&](const AddTxFn& add_tx) {
+                const auto& block_txs = value["txs"].getValues();
+                for (const UniValue& block_tx : block_txs) {
+                    add_tx(uint256S(block_tx["hash"].get_str()), block_tx["height"].getInt<int>(),
+                        block_tx["fee"].getInt<int64_t>(), block_tx["size"].getInt<uint32_t>());
+                }
+                return block_txs.size();
+            });
+            continue;
+        }
+
+        const UniValue& removeTx = value["removeTx"];
+        if (removeTx.isObject()) {
+            m_estimator.removeTx(uint256S(removeTx["hash"].get_str()), removeTx["inBlock"].get_bool());
+            continue;
+        }
+
+        const UniValue& flush = value["flush"];
+        if (flush.isArray()) {
+            m_estimator.FlushUnconfirmed();
+            continue;
+        }
+
+        const UniValue& read = value["read"];
+        if (read.isStr()) {
+            std::vector<unsigned char> data = ParseHex(read.get_str());
+            uint16_t randv = 0;
+            GetRandBytes(Span{(unsigned char*)&randv, sizeof(randv)});
+            fs::path data_filepath = fs::PathFromString(strprintf("fee_estimates.tmp.%04x", randv));
+            CAutoFile(fsbridge::fopen(data_filepath, "wb"), SER_DISK, CLIENT_VERSION).write(MakeByteSpan(data));
+            {
+                CAutoFile data(fsbridge::fopen(data_filepath, "rb"), SER_DISK, CLIENT_VERSION);
+                m_estimator.Read(data);
+            }
+            fs::remove(data_filepath);
+            continue;
+        }
+    }
+
+    if (file.bad()) {
+        LogPrintf("%s: Failure reading log file %s\n", __func__, fs::PathToString(filepath));
+        return false;
+    }
+
+    return true;
+}
--- a/src/policy/fees_input.h
+++ b/src/policy/fees_input.h
@@ -1,0 +1,61 @@
+// Copyright (c) 2009-2010 Satoshi Nakamoto
+// Copyright (c) 2009-2016 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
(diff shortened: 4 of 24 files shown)


paveljanik:
This is interesting but still no comment.

Concept ACK

I'll play around with this a bit.

```
Wshadow statistics: 
   1 policy/fees_input.cpp:127:23: warning: declaration shadows a local variable [-Wshadow]
   1 policy/fees_input.cpp:197:38: warning: declaration shadows a local variable [-Wshadow]
   1 policy/fees_input.cpp:229:22: warning: declaration shadows a local variable [-Wshadow]
   1 policy/fees_input.cpp:233:27: warning: declaration shadows a local variable [-Wshadow]
   1 policy/fees_input.cpp:27:61: warning: declaration shadows a field of 'CBlockPolicyInput' [-Wshadow]
   1 test/fee_est/fee_est.cpp:301:34: warning: declaration shadows a local variable [-Wshadow]
   1 test/fee_est/fee_est.cpp:79:20: warning: declaration shadows a field of 'TxData' [-Wshadow]
   1 test/fee_est/fee_est.cpp:79:29: warning: declaration shadows a field of 'TxData' [-Wshadow]
   1 test/fee_est/fee_est.cpp:79:39: warning: declaration shadows a field of 'TxData' [-Wshadow]
```

paveljanik:
When estimate log doesn't have enough data, `fee_est` is generating empty data in the HTML file. Emit some warning in such cases?

paveljanik:
For perfect output, there should be 3rd dimension - time or current block height when the tx was first seen :-)

jnewbery:
Concept ACK. I think this could be a useful tool.

Could this be split into two PRs to aid reviewers? The first PR would cover the `-estlog` option and writing the fee estimation data to disk. The second PR would be for a tool to read and graph the logs.

Parsing and graphing json files seems like a problem that has probably been solved many times before. If I was approaching this, I'd look at implementing this as a script in /contrib or a separate repository. Was there a particular reason you chose to implement this as a new C++ program?

laanwj:
> Could this be split into two PRs to aid reviewers? The first PR would cover the -estlog option and writing the fee estimation data to disk. The second PR would be for a tool to read and graph the logs.

I think this is useful.

However I'm not sure the analysis tool belongs in this repository. As it's specific to developers debugging the fee estimation code it's not something we want to ship with the distribution, or install by default. One place it could be is e.g. https://github.com/bitcoin-core/bitcoin-maintainer-tools . You could still document or refer to it from here.

ryanofsky:
Agree that fee_est tool shouldn't be installed, since it's a tool specifically for made for modifying and debugging fee estimation code.

But it would be awkward to use and maintain from a separate repository because it links and calls into the fee estimation code. (The tool works by piping historical data into the fee estimator so it's possible to make experimental changes to fee estimation and see how those changes affect output and internal state.)

As far as build / distribution is concerned I think it makes sense to think of it more like a unit test or benchmark than a maintainer tool.

laanwj:
> But it would be awkward to use and maintain from a separate repository because it links and calls into the fee estimation code. 

Yes, I agree, that's unfortunately true.

> Parsing and graphing json files seems like a problem that has probably been solved many times before. If I was approaching this, I'd look at implementing this as a script in /contrib or a separate repository.

I agree - especially with the references to cloudflare CDN and such. It feels a bit ugly to have that in the C++ code. Better to spit out e.g. JSON or CSV or whatever is convenient, then use a separate script for pretty formatting. This would also keep open the option of using different (non-web) visualization tools.

src/test/fee_est/README.md
```diff
@@ -0,0 +1,37 @@
+src/test/fee_est -- Fee estimation offline testing tool
+=======================================================
+
+The `fee_est` tool is intended to help debug and test changes in bitcoin fee
+estimation code using transaction data gathered from live bitcoin nodes.
+
+Transaction data can be collected by running bitcoind with `-estlog` parameter
+which will produce newline-delimited json file (http://ndjson.org/,
```
practicalswift:
````
2018-09-11 16:31:38 mdl(pr=10443): src/test/fee_est/README.md:8: MD034 Bare URL used
2018-09-11 16:31:38 mdl(pr=10443): src/test/fee_est/README.md:9: MD034 Bare URL used
```

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,332 @@
...
+    std::vector<const TxMap::value_type*> sample;
+    sample.reserve(SAMPLE_TXS);
+    int i = 0;
+    for (const auto& entry : txMap) {
+        if (entry.second.actualBlocks > int(maxTarget)) {
+            continue;
+        } else if (sample.size() < SAMPLE_TXS) {
+            sample.emplace_back(&entry);
+        } else {
+            int j = rand() % i;
```
practicalswift: Mod by zero here.

src/policy/fees.h
```diff
@@ -190,11 +191,12 @@ class CBlockPolicyEstimator
...
 
     /** Process all the transactions that have been included in a block */
+    struct Block;
     void processBlock(unsigned int nBlockHeight,
-                      std::vector<const CTxMemPoolEntry*>& entries);
+                      const std::function<void(Block&)> process_txs);
 
     /** Process a transaction accepted to the mempool*/
-    void processTransaction(const CTxMemPoolEntry& entry, bool validFeeEstimate);
+    void processTx(const uint256& hash, unsigned int height, CAmount fee, size_t size, Block* block, bool valid);
```
practicalswift:
```
2018-09-20 04:48:56 clang-tidy(pr=10443): src/policy/fees.h:199:10: warning: function 'CBlockPolicyEstimator::processTx' has a definition with different parameter names [readability-inconsistent-declaration-parameter-name]
```

src/policy/fees.cpp
```diff
@@ -9,11 +9,17 @@
 #include <clientversion.h>
 #include <primitives/transaction.h>
 #include <streams.h>
-#include <txmempool.h>
 #include <util.h>
 
 static constexpr double INF_FEERATE = 1e99;
 
+struct CBlockPolicyEstimator::Block {
+    Block(int height) : height(height) {}
```
practicalswift:
```
2018-09-20 04:48:56 clang-tidy(pr=10443): src/policy/fees.cpp:17:5: warning: single-argument constructors must be marked explicit to avoid unintentional implicit conversions [google-explicit-constructor]
```

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,332 @@
...
+
+// Drop 1/20th of sampled transactions with highest feerates if they are
+// outliers that would change the scale of the graph.
+const int DROP_FRAC = 20;
+
+void WriteGraph(const std::string& filename,
+    const TxMap& txMap,
+    const std::vector<unsigned int>& targets,
+    CBlockPolicyEstimator& estimator,
+    CTxMemPool& pool)
```
practicalswift:
```
2018-09-20 04:48:56 clang-tidy(pr=10443): src/test/fee_est/fee_est.cpp:149:17: warning: parameter 'pool' is unused [misc-unused-parameters]
```

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,332 @@
...
+    const unsigned int maxTarget = estimator.getMaxTarget();
+    for (bool conservative : {false, true}) {
+        for (unsigned int target : targets) {
+            if (target > maxTarget) continue;
+            UniValue fee(UniValue::VOBJ);
+            fee.pushKV("mode", conservative ? "conservative" : "default");
+            fee.pushKV("blocks", int(target));
+            int64_t feeRate = estimator.estimateSmartFee(target, nullptr /* confTarget */, conservative).GetFeePerK();
+            fee.pushKV("feeRate", feeRate);
+            fees.push_back(std::move(fee));
```
practicalswift:
```
2018-09-20 04:48:56 clang-tidy(pr=10443): src/test/fee_est/fee_est.cpp:163:28: warning: passing result of std::move() as a const reference argument; no move will actually happen [hicpp-move-const-arg]
```

src/test/fee_est/fee_est.cpp
practicalswift:
```
2018-09-20 04:48:56 clang-tidy(pr=10443): src/test/fee_est/fee_est.cpp:177:21: warning: rand() has limited randomness; use C++11 random library instead [cert-msc30-c]
```

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,332 @@
...
+        } else {
+            int j = rand() % i;
+            if (j < SAMPLE_TXS) {
+                sample[j] = &entry;
+            }
+        }
+        ++i;
+    }
+
+    if (sample.size() > 0) {
```
practicalswift:
```
2018-09-20 04:48:56 clang-tidy(pr=10443): src/test/fee_est/fee_est.cpp:185:9: warning: the 'empty' method should be used to check for emptiness instead of 'size' [readability-container-size-empty]
```

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,332 @@
...
+    UniValue txs(UniValue::VARR);
+    for (const auto& entry : sample) {
+        if (entry->second.actualBlocks > 0) {
+            int64_t feeRate = CFeeRate(entry->second.fee, entry->second.size).GetFeePerK();
+            if (feeRate <= maxFeeRate) {
+                UniValue tx(UniValue::VOBJ);
+                tx.pushKV("feeRate", feeRate);
+                tx.pushKV("blocks", entry->second.actualBlocks);
+                tx.pushKV("height", entry->second.height);
+                txs.push_back(std::move(tx));
```
practicalswift:
```
2018-09-20 04:48:56 clang-tidy(pr=10443): src/test/fee_est/fee_est.cpp:205:31: warning: passing result of std::move() as a const reference argument; no move will actually happen [hicpp-move-const-arg]
```

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,332 @@
...
+            return estFee <= 1 || tx.second.fee < estFee;
+        });
+        tx.second.expectedBlocks = it == targets.end() ? std::numeric_limits<int>::max() : *it;
+        return false;
+    }
+    return true;
+}
+
+void PrintCross(const TxMap& txMap,
+    const std::vector<unsigned int>& targets,
```
practicalswift:
```
2018-09-20 04:48:56 clang-tidy(pr=10443): src/test/fee_est/fee_est.cpp:229:38: warning: parameter 'targets' is unused [misc-unused-parameters]
```

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,332 @@
...
+    CBlockPolicyEstimator& estimator,
```
practicalswift:
```
2018-09-20 04:48:56 clang-tidy(pr=10443): src/test/fee_est/fee_est.cpp:230:28: warning: parameter 'estimator' is unused [misc-unused-parameters]
```

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,332 @@
...
+    CTxMemPool& pool)
```
practicalswift:
```
2018-09-20 04:48:56 clang-tidy(pr=10443): src/test/fee_est/fee_est.cpp:231:17: warning: parameter 'pool' is unused [misc-unused-parameters]
```

src/policy/fees_input.h
```diff
@@ -0,0 +1,55 @@
...
+
+    /** Process all the transactions that have been included in a block */
+    struct Block;
+    void processBlock(int height, const std::function<void(Block&)>& process_txs);
+
+    /** Process a transaction added the mempool or a block */
+    void processTx(const uint256& hash, unsigned int height, CAmount fee, size_t size, Block* block, bool valid);
+
+    /** Remove a transaction from the mempool tracking stats */
+    void removeTx(const uint256& hash, bool in_block);
```
practicalswift:
```
2018-09-20 04:48:56 clang-tidy(pr=10443): src/policy/fees_input.h:33:10: warning: function 'FeeEstInput::removeTx' has a definition with different parameter names [readability-inconsistent-declaration-parameter-name]
```

src/policy/fees_input.cpp
```diff
@@ -0,0 +1,218 @@
...
+{
+    estimator.processTx(hash, height, fee, size, block ? block->est_block : nullptr, valid);
+    if (log) {
+        UniValue tx(UniValue::VOBJ);
+        tx.pushKV("hash", hash.ToString());
+        tx.pushKV("height", int(height));
+        tx.pushKV("fee", fee);
+        tx.pushKV("size", uint64_t(size));
+        if (block) {
+            block->json_txs.push_back(std::move(tx));
```
practicalswift:
```
2018-09-20 04:48:56 clang-tidy(pr=10443): src/policy/fees_input.cpp:30:39: warning: passing result of std::move() as a const reference argument; no move will actually happen [hicpp-move-const-arg]
```

src/policy/fees_input.cpp
```diff
@@ -0,0 +1,218 @@
...
+        } else {
+            UniValue value(UniValue::VOBJ);
+            value.pushKV("tx", std::move(tx));
```
practicalswift:
```
2018-09-20 04:48:56 clang-tidy(pr=10443): src/policy/fees_input.cpp:33:32: warning: passing result of std::move() as a const reference argument; no move will actually happen [hicpp-move-const-arg]
```

src/policy/fees_input.cpp
```diff
@@ -0,0 +1,218 @@
...
+        UniValue value(UniValue::VOBJ);
+        value.pushKV("stop", Params().NetworkIDString());
+        value.pushKV("time", GetTime());
+        *log << value.write() << std::endl;
+    }
+
+    if (filename.empty()) {
+        log.reset();
+    } else {
+        log.reset(new std::ofstream(filename, std::ofstream::out | std::ofstream::app));
```
practicalswift:
```
2018-09-20 04:48:56 clang-tidy(pr=10443): src/policy/fees_input.cpp:126:13: warning: use std::make_unique instead [modernize-make-unique]
```

`MakeUnique` in our case :-)

src/policy/fees_input.cpp
```diff
@@ -0,0 +1,218 @@
...
+        UniValue value(UniValue::VOBJ);
+        value.pushKV("start", Params().NetworkIDString());
+        value.pushKV("time", GetTime());
+        *log << value.write() << std::endl;
+    }
+
+    return true;
+}
+
+bool FeeEstInput::readLog(const std::string& filename, std::function<bool(UniValue&)> filter)
```
practicalswift:
```
2018-09-20 04:48:56 clang-tidy(pr=10443): src/policy/fees_input.cpp:143:87: warning: the parameter 'filter' is copied for each invocation but only used as a const reference; consider making it a const reference [performance-unnecessary-value-param]
```

src/policy/fees_input.cpp
```diff
@@ -0,0 +1,218 @@
...
+        const UniValue& flush = value["flush"];
+        if (flush.isArray()) {
+            estimator.FlushUnconfirmed();
+            continue;
+        }
+
+        const UniValue& read = value["read"];
+        if (read.isStr()) {
+            std::vector<unsigned char> data = ParseHex(read.get_str());
+            unsigned short randv = 0;
```
practicalswift:
```
2018-09-20 04:48:56 clang-tidy(pr=10443): src/policy/fees_input.cpp:198:13: warning: consider replacing 'unsigned short' with 'uint16' [google-runtime-int]
```

src/policy/fees_input.h
```diff
@@ -0,0 +1,55 @@
...
+
+class CAutoFile;
+class CBlockPolicyEstimator;
+class UniValue;
+class uint256;
+
+class FeeEstInput
+{
+public:
+    FeeEstInput(CBlockPolicyEstimator& estimator);
```
practicalswift:
```
2018-09-20 05:13:44 cppcheck(pr=10443): [src/policy/fees_input.h:23]: (style) Class 'FeeEstInput' has a constructor with 1 argument that is not explicit.
```

src/policy/fees.cpp
practicalswift:
```
2018-09-20 05:13:44 cppcheck(pr=10443): [src/policy/fees.cpp:17]: (style) Struct 'Block' has a constructor with 1 argument that is not explicit.
```

src/policy/fees_input.cpp
practicalswift:
```
2018-09-22 22:51:38 cpplint(pr=10443): src/policy/fees_input.cpp:198:  Use int16/int64/etc, rather than the C type short  [runtime/int] [4]
```

src/policy/fees_input.h
practicalswift:
```
2018-09-22 22:51:38 cpplint(pr=10443): src/policy/fees_input.h:23:  Single-parameter constructors should be marked explicit.  [runtime/explicit] [5]
```

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,332 @@
+#include <policy/fees.h>
```
practicalswift:
```
2018-09-22 22:51:38 cpplint(pr=10443): src/test/fee_est/fee_est.cpp:0:  No copyright message found.
```

src/policy/fees_input.h
```diff
@@ -0,0 +1,55 @@
...
+class uint256;
+
+class FeeEstInput
+{
+public:
+    explicit FeeEstInput(CBlockPolicyEstimator& estimator);
+
+    /** Process all the transactions that have been included in a block */
+    struct Block;
+    void processBlock(int height, const std::function<void(Block&)>& process_txs);
```
practicalswift: Isn't `height` guaranteed to be non-negative? Should be `unsigned int`?

src/policy/fees_input.cpp
```diff
@@ -0,0 +1,222 @@
...
+
+bool FeeEstInput::readData(const fs::path& filename)
+{
+    CAutoFile file(fsbridge::fopen(filename, "rb"), SER_DISK, CLIENT_VERSION);
+    if (file.IsNull() || !estimator.Read(file)) {
+        return false;
+    }
+    if (log) {
+        UniValue value(UniValue::VOBJ);
+        std::ifstream file(filename.string(), std::ifstream::binary);
```
practicalswift: The variable `file` shadows an already existing local variable. Please choose another name :-)

src/policy/fees_input.cpp
```diff
@@ -0,0 +1,222 @@
...
+                tx["size"].get_int(), nullptr /* block */, value["valid"].get_bool());
+            continue;
+        }
+
+        const UniValue& block = value["block"];
+        if (block.isObject()) {
+            int height = block["height"].get_int();
+            estimator.processBlock(height, [&](CBlockPolicyEstimator::Block& est_block) {
+                const auto& txs = value["txs"].getValues();
+                for (const UniValue& tx : txs) {
```
practicalswift: `tx` shadows a local variable. Please choose another name :-)

src/policy/fees_input.cpp
```diff
@@ -0,0 +1,222 @@
...
+            estimator.FlushUnconfirmed();
+            continue;
+        }
+
+        const UniValue& read = value["read"];
+        if (read.isStr()) {
+            std::vector<unsigned char> data = ParseHex(read.get_str());
+            uint16_t randv = 0;
+            GetRandBytes((unsigned char*)&randv, sizeof(randv));
+            fs::path filename = strprintf("fee_estimates.tmp.%04x", randv);
```
practicalswift: `filename` shadows another variable. Please use another name :-)

src/policy/fees_input.cpp
```diff
@@ -0,0 +1,222 @@
...
+            CAutoFile(fsbridge::fopen(filename, "wb"), SER_DISK, CLIENT_VERSION)
+                .write((const char*)data.data(), data.size());
+            {
+                CAutoFile file(fsbridge::fopen(filename, "rb"), SER_DISK, CLIENT_VERSION);
```
practicalswift: Same here: shadows another local variable :-)

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+    int64_t maxFeeRate = 0;
+    const unsigned int maxTarget = estimator.getMaxTarget();
+    for (bool conservative : {false, true}) {
+        for (unsigned int target : targets) {
+            if (target > maxTarget) continue;
+            UniValue fee(UniValue::VOBJ);
+            fee.pushKV("mode", conservative ? "conservative" : "default");
+            fee.pushKV("blocks", int(target));
+            int64_t feeRate = estimator.estimateSmartFee(target, nullptr /* confTarget */, conservative).GetFeePerK();
```
practicalswift: An implicit signedness changing conversion takes place here. Avoid or make it explicit :-)

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,333 @@
...
+    sample.reserve(SAMPLE_TXS);
+    int i = 0;
+    std::minstd_rand randint;
+    for (const auto& entry : txMap) {
+        if (entry.second.actualBlocks > int(maxTarget)) {
+            continue;
+        } else if (sample.size() < SAMPLE_TXS) {
+            sample.emplace_back(&entry);
+        } else {
+            int j = randint() % i;
```
practicalswift: An implicit signedness changing conversion takes place here. Avoid or make it explicit :-)

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,333 @@
...
+            if (j < SAMPLE_TXS) {
+                sample[j] = &entry;
```
practicalswift: An implicit signedness changing conversion takes place here. Avoid or make it explicit :-)

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,333 @@
...
+            }
+        }
+        ++i;
+    }
+
+    if (!sample.empty()) {
+        std::vector<int64_t> feeRates;
+        feeRates.reserve(sample.size());
+        for (const auto& entry : sample) {
+            feeRates.emplace_back(CFeeRate(entry->second.fee, entry->second.size).GetFeePerK());
```
practicalswift: An implicit signedness changing conversion takes place here. Avoid or make it explicit :-)

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,333 @@
...
+        }
+        auto nth = feeRates.begin() + sample.size() / DROP_FRAC;
+        std::nth_element(feeRates.begin(), nth, feeRates.end(), std::greater<int64_t>());
+        maxFeeRate = std::max(maxFeeRate, *nth);
+    }
+
+    UniValue txs(UniValue::VARR);
+    for (const auto& entry : sample) {
+        if (entry->second.actualBlocks > 0) {
+            int64_t feeRate = CFeeRate(entry->second.fee, entry->second.size).GetFeePerK();
```
practicalswift: An implicit signedness changing conversion takes place here. Avoid or make it explicit :-)

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,333 @@
...
+
+    std::ofstream file(filename);
+    file << strprintf(GRAPH_HTML, txs.write(), fees.write());
+}
+
+bool UpdateCross(TxMap::value_type& tx, const std::vector<unsigned int>& targets, CBlockPolicyEstimator& estimator)
+{
+    if (*tx.first.begin() == 0) {
+        auto it = std::lower_bound(targets.begin(), targets.end(), nullptr, [&](unsigned int target, std::nullptr_t) {
+            CAmount estFee = estimator.estimateSmartFee(target, nullptr /* feeCalc */, false /* conservative */)
```
practicalswift: An implicit signedness changing conversion takes place here. Avoid or make it explicit :-)

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,333 @@
...
+                                 .GetFee(tx.second.size);
```
practicalswift: An implicit signedness changing conversion takes place here. Avoid or make it explicit :-)

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,333 @@
...
+            return estFee <= 1 || tx.second.fee < estFee;
+        });
+        tx.second.expectedBlocks = it == targets.end() ? std::numeric_limits<int>::max() : *it;
```
practicalswift: An implicit signedness changing conversion takes place here. Avoid or make it explicit :-)

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,333 @@
...
+        return false;
+    }
+    return true;
+}
+
+void PrintCross(const TxMap& txMap,
+    CBlockPolicyEstimator& estimator)
```
practicalswift: The unused parameter `estimator` can be removed?

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,333 @@
...
+            if (options.cross && !UpdateCross(*inserted.first, targets, estimator)) {
+                keep = false;
+            }
+        }
+
+        const UniValue& block = value["block"];
+        if (block.isObject()) {
+            int blockHeight = block["height"].get_int();
+            const auto& txs = value["txs"].getValues();
+            for (const UniValue& tx : txs) {
```
practicalswift: `tx` here shadows an already existing local variable. Please choose another name :-)

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,330 @@
...
+                sample[j] = &entry;
+            }
+        }
+        ++i;
+    }
+
+    if (!sample.empty()) {
+        std::vector<CAmount> feeRates;
+        feeRates.reserve(sample.size());
+        for (const auto& entry : sample) {
```
practicalswift: Could use `std::transform`? :-)

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,330 @@
...
+            ++unconfirmedTestTxs;
+        } else if (entry.second.expectedBlocks == std::numeric_limits<int>::max()) {
+            ++outlierTestTxs;
+        } else {
+            ++keptTestTxs;
+            int err = entry.second.expectedBlocks - entry.second.actualBlocks;
+            errsq += err * err;
+        }
+    }
+    printf("Non-test txs: %i\n", nonTestTxs);
```
practicalswift: Avoid `printf(…)` – it is not used in the rest of the project:

```
$ git grep "[^frns]printf" -- "*.cpp" "*.h" ":(exclude)src/leveldb/" ":(exclude)src/secp256k1/" ":(exclude)src/tinyformat.h" ":(exclude)src/univalue/"
$
```

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,330 @@
...
+    UniValue txs(UniValue::VARR);
+    for (const auto& entry : sample) {
+        if (entry->second.actualBlocks > 0) {
+            CAmount feeRate = CFeeRate(entry->second.fee, entry->second.size).GetFeePerK();
+            if (feeRate <= maxFeeRate) {
+                UniValue tx(UniValue::VOBJ);
+                tx.pushKV("feeRate", feeRate);
+                tx.pushKV("blocks", entry->second.actualBlocks);
+                tx.pushKV("height", entry->second.height);
+                txs.push_back(std::move(tx));
```
practicalswift: This `move` is not needed?

src/policy/fees_input.cpp
```diff
@@ -0,0 +1,227 @@
...
+        UniValue tx(UniValue::VOBJ);
+        tx.pushKV("hash", hash.ToString());
+        tx.pushKV("height", int(height));
+        tx.pushKV("fee", fee);
+        tx.pushKV("size", uint64_t(size));
+        if (block) {
+            block->json_txs.push_back(std::move(tx));
+        } else {
+            UniValue value(UniValue::VOBJ);
+            value.pushKV("tx", std::move(tx));
```
practicalswift: `move` not needed?

src/policy/fees_input.cpp
```diff
@@ -0,0 +1,227 @@
...
+{
+    estimator.processTx(hash, height, fee, size, block ? block->est_block : nullptr, valid);
+    if (log) {
+        UniValue tx(UniValue::VOBJ);
+        tx.pushKV("hash", hash.ToString());
+        tx.pushKV("height", int(height));
+        tx.pushKV("fee", fee);
+        tx.pushKV("size", uint64_t(size));
+        if (block) {
+            block->json_txs.push_back(std::move(tx));
```
practicalswift: `move` not needed?

src/policy/fees_input.cpp
```diff
@@ -0,0 +1,227 @@
...
+        const UniValue& read = value["read"];
+        if (read.isStr()) {
+            std::vector<unsigned char> data = ParseHex(read.get_str());
+            uint16_t randv = 0;
+            GetRandBytes((unsigned char*)&randv, sizeof(randv));
+            fs::path data_filename = strprintf("fee_estimates.tmp.%04x", randv);
+            CAutoFile(fsbridge::fopen(data_filename, "wb"), SER_DISK, CLIENT_VERSION)
+                .write((const char*)data.data(), data.size());
+            {
+                CAutoFile data(fsbridge::fopen(data_filename, "rb"), SER_DISK, CLIENT_VERSION);
```
practicalswift: This `data` shadows the `data` on line 206 above :-)

ryanofsky:
Rebased 51163cee78f32fb6311bf620266c42fd115ba98f -> 8b9c3da3cf077051d6b1112ba4adbadd60096d57 ([pr/fee.16](https://github.com/ryanofsky/bitcoin/commits/pr/fee.16) -> [pr/fee.17](https://github.com/ryanofsky/bitcoin/commits/pr/fee.17)) due to conflict with #13341
Rebased 8b9c3da3cf077051d6b1112ba4adbadd60096d57 -> 4bcecdb666d1dbc90201dccbfee27a234f88c960 ([pr/fee.17](https://github.com/ryanofsky/bitcoin/commits/pr/fee.17) -> [pr/fee.18](https://github.com/ryanofsky/bitcoin/commits/pr/fee.18)) due to conflict with #13786
Updated 4bcecdb666d1dbc90201dccbfee27a234f88c960 -> aaa4b687c5fdeb2c75246e23077c984a0f133e81 ([pr/fee.18](https://github.com/ryanofsky/bitcoin/commits/pr/fee.18) -> [pr/fee.19](https://github.com/ryanofsky/bitcoin/commits/pr/fee.19)) fixing travis size_t univalue errors
Rebased aaa4b687c5fdeb2c75246e23077c984a0f133e81 -> a5faeb1cedd52b2886a1d5c320802874436fe2f6 ([pr/fee.19](https://github.com/ryanofsky/bitcoin/commits/pr/fee.19) -> [pr/fee.20](https://github.com/ryanofsky/bitcoin/commits/pr/fee.20)) to fix broken lint error caused by #13705
Updated a5faeb1cedd52b2886a1d5c320802874436fe2f6 -> 61ab4c17eebf9b73267c992fb664d603620f9e81 ([pr/fee.20](https://github.com/ryanofsky/bitcoin/commits/pr/fee.20) -> [pr/fee.21](https://github.com/ryanofsky/bitcoin/commits/pr/fee.21)) adding missing lock annotation.
Rebased 61ab4c17eebf9b73267c992fb664d603620f9e81 -> a9eb5cc5ceace3e567a0f7ba44c6350b2746436f ([pr/fee.21](https://github.com/ryanofsky/bitcoin/commits/pr/fee.21) -> [pr/fee.22](https://github.com/ryanofsky/bitcoin/commits/pr/fee.22)) due to conflict with #14100
Updated a9eb5cc5ceace3e567a0f7ba44c6350b2746436f -> 8cd1bb684bfec22fc54e091dc96c782a26847393 ([pr/fee.22](https://github.com/ryanofsky/bitcoin/commits/pr/fee.22) -> [pr/fee.23](https://github.com/ryanofsky/bitcoin/commits/pr/fee.23)) to fix `G_TRANSLATION_FUN` link error after #13961
Rebased 8cd1bb684bfec22fc54e091dc96c782a26847393 -> a60d26ac2bceea2c67ef34c8641705b16f656d07 ([pr/fee.23](https://github.com/ryanofsky/bitcoin/commits/pr/fee.23) -> [pr/fee.24](https://github.com/ryanofsky/bitcoin/commits/pr/fee.24)) due to conflict with #13311
Updated a60d26ac2bceea2c67ef34c8641705b16f656d07 -> 1e7ec287c927437b5a183bc807c49eba4a16d18a ([pr/fee.24](https://github.com/ryanofsky/bitcoin/commits/pr/fee.24) -> [pr/fee.25](https://github.com/ryanofsky/bitcoin/commits/pr/fee.25)) with suggested lint fixes
Updated 1e7ec287c927437b5a183bc807c49eba4a16d18a -> 56446378f5d0c2bdeb7a45fa7382ccd58c9b994c ([pr/fee.25](https://github.com/ryanofsky/bitcoin/commits/pr/fee.25) -> [pr/fee.26](https://github.com/ryanofsky/bitcoin/commits/pr/fee.26)) with more lint fixes
Updated 56446378f5d0c2bdeb7a45fa7382ccd58c9b994c -> 8d99281859540689b55285bd0acdf8c1529c4b37 ([pr/fee.26](https://github.com/ryanofsky/bitcoin/commits/pr/fee.26) -> [pr/fee.27](https://github.com/ryanofsky/bitcoin/commits/pr/fee.27)) with more lint fixes
Updated 8d99281859540689b55285bd0acdf8c1529c4b37 -> 68944a5120c651772fd7f3c028c446e38f9a3f57 ([pr/fee.27](https://github.com/ryanofsky/bitcoin/commits/pr/fee.27) -> [pr/fee.28](https://github.com/ryanofsky/bitcoin/commits/pr/fee.28)) with more lint fixes
Rebased 68944a5120c651772fd7f3c028c446e38f9a3f57 -> 50cd9b1cb7d03a81140d8a59698eba8d06b89612 ([pr/fee.28](https://github.com/ryanofsky/bitcoin/commits/pr/fee.28) -> [pr/fee.29](https://github.com/ryanofsky/bitcoin/commits/pr/fee.29)) due to conflict with #14555
Rebased 50cd9b1cb7d03a81140d8a59698eba8d06b89612 -> a0148157c57e91b299e6b450c7ee758f606488e6 ([pr/fee.29](https://github.com/ryanofsky/bitcoin/commits/pr/fee.29) -> [pr/fee.30](https://github.com/ryanofsky/bitcoin/commits/pr/fee.30)) due to conflict with #13128
Updated a0148157c57e91b299e6b450c7ee758f606488e6 -> 6f9c07f0a72ff0d95bfee9f37e4ae3dc9ed9c0e4 ([pr/fee.30](https://github.com/ryanofsky/bitcoin/commits/pr/fee.30) -> [pr/fee.31](https://github.com/ryanofsky/bitcoin/commits/pr/fee.31)) restructuring to work around clang lock annotation errors from #13128

src/policy/fees.cpp
```diff
@@ -546,11 +545,13 @@ CBlockPolicyEstimator::~CBlockPolicyEstimator()
 {
 }
 
-void CBlockPolicyEstimator::processTransaction(const CTxMemPoolEntry& entry, bool validFeeEstimate)
```
Empact: nit: this renaming is succinct, yes, but also less communicative.

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,330 @@
...
+    CAmount fee;
+    size_t size;
+    int height;
+    int expectedBlocks = -1;
+    int actualBlocks = -1;
+};
+
+using TxMap = std::map<uint256, TxData>;
+
+const char* const GRAPH_HTML = R"(<!DOCTYPE html>
```
Empact: nit: maybe load this from a separate file?

sipa:
Concept ACK, but it seems the HTML code (or even the conversion to HTML) could live outside the repo?

ryanofsky:
This PR has some overlap with @kallewoof's [mff](https://github.com/kallewoof/mff) utility. One difference is that this creates logs in [json-lines](http://jsonlines.org/) format and only logs mempool data needed to run fee estimation code reproducibly, while mff uses a custom format and logs additional information.

jonatack:
Discovered this PR thanks to DrahtBot's "needs rebase" notifications.

Concept ACK, interesting idea for dev tooling in the tests or contrib.

src/policy/fees_input.cpp
```diff
@@ -0,0 +1,222 @@
...
+        UniValue value(UniValue::VOBJ);
+        value.pushKV("stop", Params().NetworkIDString());
+        value.pushKV("time", GetTime());
+        *log << value.write() << std::endl;
+    }
+
+    if (filename.empty()) {
+        log.reset();
+    } else {
+        log = MakeUnique<std::ofstream>(filename, std::ofstream::out | std::ofstream::app);
```
fanquake, line 130: Please [use `std::make_unique` in new code](https://github.com/bitcoin/bitcoin/blob/master/src/util/memory.h#L13).

ryanofsky:
Rebased 2c512ff0729f89aa47019009534d3958ec18f42b -> de8a2d80f74d33067a3b532368cca877fc65e7d2 ([`pr/fee.60`](https://github.com/ryanofsky/bitcoin/commits/pr/fee.60) -> [`pr/fee.61`](https://github.com/ryanofsky/bitcoin/commits/pr/fee.61), [compare](https://github.com/ryanofsky/bitcoin/compare/pr/fee.60-rebase..pr/fee.61)) due to conflicts with #18766

aureleoules:
Concept ACK
CI is failing though.

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,331 @@
...
+const size_t SAMPLE_TXS = 50000;
+
+// Drop 1/20th of sampled transactions with highest feerates if they are
+// outliers that would change the scale of the graph.
+const int DROP_FRAC = 20;
+
+void WriteGraph(const std::string& filename,
+    const TxMap& txMap,
+    const std::vector<unsigned int>& targets,
+    CBlockPolicyEstimator& estimator)
```
aureleoules, line 151: 538c780d078ecf400eb7b509f9fbcd176cdd290c
```suggestion
    const CBlockPolicyEstimator& estimator)
```

src/policy/fees.cpp
```diff
@@ -527,8 +526,8 @@ bool CBlockPolicyEstimator::_removeTx(const uint256& hash, bool inBlock)
     }
 }
 
-CBlockPolicyEstimator::CBlockPolicyEstimator(const fs::path& estimation_filepath)
-    : m_estimation_filepath{estimation_filepath}, nBestSeenHeight{0}, firstRecordedHeight{0}, historicalFirst{0}, historicalBest{0}, trackedTxs{0}, untrackedTxs{0}
+CBlockPolicyEstimator::CBlockPolicyEstimator()
+    : nBestSeenHeight{0}, firstRecordedHeight{0}, historicalFirst{0}, historicalBest{0}, trackedTxs{0}, untrackedTxs{0}
```
aureleoules, line 530: 55016bfa39f496c08ea0c18260a3b6abbdb23550
nit: use class initialization

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,331 @@
...
+                                 .GetFee(tx.second.size);
+            return estFee <= 1 || tx.second.fee < estFee;
+        });
+        tx.second.expectedBlocks = it == targets.end() ? std::numeric_limits<int>::max() : int(*it);
+        return false;
+    }
+    return true;
+}
+
+void PrintCross(const TxMap& txMap, CBlockPolicyEstimator&)
```
aureleoules, line 231: 538c780d078ecf400eb7b509f9fbcd176cdd290c
```suggestion
void PrintCross(const TxMap& txMap)
```

src/test/fee_est/fee_est.cpp
```diff
@@ -0,0 +1,331 @@
...
+                txs.push_back(std::move(tx));
+            }
+        }
+    }
+
+    std::ofstream file(filename);
+    file << strprintf(GRAPH_HTML, txs.write(), fees.write());
+}
+
+bool UpdateCross(TxMap::value_type& tx, const std::vector<unsigned int>& targets, CBlockPolicyEstimator& estimator)
```
aureleoules, line 217: 538c780d078ecf400eb7b509f9fbcd176cdd290c
```suggestion
bool UpdateCross(TxMap::value_type& tx, const std::vector<unsigned int>& targets, const CBlockPolicyEstimator& estimator)
```

josibake:
Haven't dug into the code, but I'm leaning toward Concept ACK, Approach NACK

As someone who is planning to do fee estimation research in the near future, I don't think I'd use something like this the way it's currently written. I would much prefer something that logs Bitcoin Core's live _fee rate_ estimations, or some option that would allow me to tell Bitcoin Core it's at a certain chain height, feed it some mempool data, and then ask it for fee estimations.

I also agree that the tool for analyzing the data shouldn't be in this repo. Much better if we can figure out what data is missing for people who want to do fee estimation analysis and provide an option for logging that data.

Perhaps a good next step would be to talk to people who are currently doing fee estimation work and see what data would help with their projects/PRs? @ClaraShk @darosior @Xekyo ?

ryanofsky:
@josie thanks for looking at this. I don't think I understand what the idea behind an approach NACK would be based on what you are saying, though.

It sounds more like you're asking for an additional feature which would be easy to implement on top of the first commit, not pointing to a drawback of the approach. If you want to gather live fee estimates, the first commit outputs a structured [jsonl](https://jsonlines.org/) log of fee estimation events, so it would be easy to add the actual fee estimates to this log, or add internal metrics used to produce the fee estimates, or add any other data points that might be interesting. It is also possible to take the existing log files this PR already generates, and feed them to the fee estimation algorithm by modifying the tool in the second commit to produce live fee estimates or a history of estimates over time.

If you just think I should drop the second commit, or trim it down, I think that would be reasonable. The point of the second commit is mainly to provide example code that demonstrates how to read the JSONL log file and how to call the fee estimator API. The idea is to avoid needing reinvent the wheel each time you want to monitor fee estimation behavior, or reproduce a bug, or determine the effects of a change.

> I would much prefer [...] some option that would allow me to tell Bitcoin Core it's at a certain chain height, feed it some mempool data, and then ask it for fee estimations.

This is actually what the second commit provides. The .jsonl log contains chain and mempool data. You can feed it to the test program in the second commit, and it calls fee estimation code to produce fee estimates. It can produce a graph of the fee estimates, and calculate how accurate they were by looking at all the transactions that got confirmed and outputting the mean squared error of how many blocks it took the transactions to be confirmed vs how many blocks the fee estimation algorithm predicted it would take to get them confirmed based on their feerate.

This PR is a few years old, so I would probably implement the second commit a little differently today. Instead of adding a standalone c++ executable that calls fee estimator code and produces graphs and metrics, I might want to just add a python module that exposes the `CBlockPolicyEstimator` class, and generate the graphs and metrics from a python notebook.

josibake:
Thanks for the thorough response @ryanofsky. I realize I didn't explain myself very well in my initial comment.

> It sounds more like you're asking for an additional feature which would be easy to implement on top of the first commit, not pointing to a drawback of the approach

Since we already record mempool events in the debug log file (this is what @jamesob and I are using to log mempool events in [bmon](https://github.com/chaincodelabs/bmon)), it feels a bit strange to have special code for logging transactions again here for fee estimation. It's possible I'm not understanding something here, though: is there some special information in the log you are writing that we can't get from the debug log file?

> This is actually what the second commit provides. The .jsonl log contains chain and mempool data. You can feed it to the test program in the second commit, and it calls fee estimation code to produce fee estimates

I think what I was envisioning was even simpler: have a flag to make Bitcoin Core log `fee_rate` estimates per block and then have some way to walk the node back to some point in the past, have it log `fee_rate` estimates (perhaps with a code change) and then leave it up to the individual on how to analyze the diff between those logs. Perhaps this is easier said than done since the fee estimation code in Bitcoin Core today seems to work on a per transaction basis?

ryanofsky:
> is there some special information in the log you are writing that we can't get from the debug log file?

I think so. The json log captures all inputs to fee estimator to used update its internal state. It includes not just mempool events but also information about which transactions are included in blocks, flush events, and contents of `fee_estimates.dat` blobs that are loaded. It makes fee estimation code reproducible, so if you make a change to the fee estimation code or parameters, you can evaluate the effects of the change with real world inputs offline. [^1]

[^1]: I also think in practice parsing a json log file is cleaner and more straightforward than parsing a debug.log file. Compare
[readLog](https://github.com/ryanofsky/bitcoin/blob/26240e1a5f320e8a9da71cf4b7e6313a9b494aff/src/policy/fees_input.cpp#L162-L228) vs [logparse](https://github.com/chaincodelabs/bmon/blob/26032dbec86f14044c9af0af0e06528750ad6fc8/bmon/logparse.py). But I recognize there could be other advantages to using debug.log files, especially since debug.log files are omnipresent and already exist.

> I think what I was envisioning was even simpler: have a flag to make Bitcoin Core log fee_rate estimates per block and then have some way to walk the node back to some point in the past, have it log fee_rate estimates (perhaps with a code change) and then leave it up to the individual on how to analyze the diff between those logs. Perhaps this is easier said than done since the fee estimation code in Bitcoin Core today seems to work on a per transaction basis?

I think I don't actually understand the proposal or don't understand the intended use-case. Use-cases for PR this are to:

- Be able to make changes to the fee estimation algorithm and be able to see how the changes affect fee estimates in a quick feedback loop
- Be able to evaluate quality of fee estimates. The example program compares how quickly fee estimator expects transactions get to get confirmed (based on feerate) vs how quickly they actually get confirmed. This could be one metric for comparing quality of different fee estimation algorithms, or seeing if one algorithm starts to perform better or worse over time. It should also be possible to implement other metrics besides this one and do similar things.

Is your proposal basically to save a `fee_estimates.dat` file each time a block is connected and then use those files for something? I'm guess I'm not sure how you would use those files to improve fee estimates. This PR does capture a superset of that information, too. For example, it would be possible to generate those files offline from the json log file produced in this PR by adding a few lines to the [`filter`](https://github.com/ryanofsky/bitcoin/blob/26240e1a5f320e8a9da71cf4b7e6313a9b494aff/src/test/fee_est/fee_est.cpp#L285) callback:

```c++
const UniValue& block = value["block"];
if (block.isObject()) {
    AutoFile file{fsbridge::fopen(strprintf("fee_estimates-%07d.dat", block["height"].getInt<int>());, "wb")};
    estimator.Write(file);
}
```

Another advantage of this change is that it decouples `CBlockPolicyEstimator` from the mempool, so it's possible to call it offline. The external API changes from:

https://github.com/bitcoin/bitcoin/blob/7b45d171f549595a831489827c28e8493f36c00c/src/policy/fees.h#L189-L200

to:

https://github.com/bitcoin/bitcoin/blob/26240e1a5f320e8a9da71cf4b7e6313a9b494aff/src/policy/fees.h#L191-L201

There are probably lots of changes that could be made to this PR to improve it, or trim it down if it is doing too much, but in general I think takes a good first step of pulling things out of `CBlockPolicyEstimator` that don't belong there (mempool access, filesystem access), and then builds some useful features on top of that.

ryanofsky:
Going to close this because I think @josibake is working on a different approach which will try replace current fee estimator, instead of just instrumenting it like this PR.

There's also another PR #28368 which decouples the fee estimator from the mempool, and might make it easier to replay fee estimation events like the changes in this PR, even though it only provides an API, not a CLI tool.

PR #10444, opened 2017-05-23
Implement BIP91 Reduced threshold Segwit MASF

jameshilliard:
This is an implementation of BIP91.
BIP91 details: https://github.com/bitcoin/bips/blob/master/bip-0091.mediawiki
Previous discussion:
https://lists.linuxfoundation.org/pipermail/bitcoin-dev/2017-May/014380.html

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -96,6 +96,11 @@ public:
         consensus.vDeployments[Consensus::DEPLOYMENT_SEGWIT].nStartTime = 1479168000; // November 15th, 2016.
         consensus.vDeployments[Consensus::DEPLOYMENT_SEGWIT].nTimeout = 1510704000; // November 15th, 2017.
 
+        // Deployment of SEGSIGNAL
+        consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].bit = 2;
+        consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].nStartTime = 1496275200; // June 1st, 2017.
+        consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].nTimeout = 1510704000; // November 15th, 2017.
+
         // The best chain should have at least this much work.
         consensus.nMinimumChainWork = uint256S("0x0000000000000000000000000000000000000000003f94d1ad391682fe038bf5");
 
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -97,7 +97,7 @@ public:
         consensus.vDeployments[Consensus::DEPLOYMENT_SEGWIT].nTimeout = 1510704000; // November 15th, 2017.
 
         // Deployment of SEGSIGNAL
-        consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].bit = 2;
+        consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].bit = 4;
         consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].nStartTime = 1496275200; // June 1st, 2017.
         consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].nTimeout = 1510704000; // November 15th, 2017.
 
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -291,6 +291,11 @@ public:
         consensus.vDeployments[Consensus::DEPLOYMENT_SEGWIT].bit = 1;
         consensus.vDeployments[Consensus::DEPLOYMENT_SEGWIT].nStartTime = 0;
         consensus.vDeployments[Consensus::DEPLOYMENT_SEGWIT].nTimeout = 999999999999ULL;
+        consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].bit = 4;
+        consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].nStartTime = 0;
+        consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].nTimeout = 999999999999ULL;
+        consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].nOverrideMinerConfirmationWindow = 48;
+        consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].nOverrideRuleChangeActivationThreshold = 29; // 60%
 
         // The best chain should have at least this much work.
         consensus.nMinimumChainWork = uint256S("0x00");
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -100,6 +100,8 @@ public:
         consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].bit = 4;
         consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].nStartTime = 1496275200; // June 1st, 2017.
         consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].nTimeout = 1510704000; // November 15th, 2017.
+        consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].nOverrideMinerConfirmationWindow = 672; // ~4.67 days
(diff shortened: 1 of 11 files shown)


jonasschnelli:
`bip68-112-113-p2p.py` are falling...

src/validation.cpp
```diff
@@ -2918,6 +2930,13 @@ bool IsWitnessEnabled(const CBlockIndex* pindexPrev, const Consensus::Params& pa
     return (VersionBitsState(pindexPrev, params, Consensus::DEPLOYMENT_SEGWIT, versionbitscache) == THRESHOLD_ACTIVE);
 }
 
+// Check if Segregated Witness is Locked In
+bool IsWitnessLockedIn(const CBlockIndex* pindexPrev, const Consensus::Params& params)
```
jtimon: This function could be static

src/versionbits.cpp
```diff
@@ -148,7 +152,14 @@ class VersionBitsConditionChecker : public AbstractThresholdConditionChecker {
     int64_t BeginTime(const Consensus::Params& params) const { return params.vDeployments[id].nStartTime; }
     int64_t EndTime(const Consensus::Params& params) const { return params.vDeployments[id].nTimeout; }
     int Period(const Consensus::Params& params) const { return params.nMinerConfirmationWindow; }
-    int Threshold(const Consensus::Params& params) const { return params.nRuleChangeActivationThreshold; }
+    int Threshold(const Consensus::Params& params) const {
+        if (params.nRuleChangeActivationThreshold == 1916 && params.nMinerConfirmationWindow == 2016 &&
```
earonesty: can these magic numbers - when not being used as actual thresholds - be constants?

src/versionbits.cpp
```diff
@@ -148,7 +152,13 @@ class VersionBitsConditionChecker : public AbstractThresholdConditionChecker {
     int64_t BeginTime(const Consensus::Params& params) const { return params.vDeployments[id].nStartTime; }
     int64_t EndTime(const Consensus::Params& params) const { return params.vDeployments[id].nTimeout; }
     int Period(const Consensus::Params& params) const { return params.nMinerConfirmationWindow; }
-    int Threshold(const Consensus::Params& params) const { return params.nRuleChangeActivationThreshold; }
+    int Threshold(const Consensus::Params& params) const {
+        if (params.vDeployments[id].bit == params.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].bit &&
+            params.vDeployments[id].nStartTime == params.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].nStartTime) {
+            return 1311; // 65% threshold for SEGSIGNAL only
```
jtimon: Perhaps this can be a field on Consensus::Params? or at least a properly defined constant

src/versionbits.cpp
```diff
@@ -148,7 +152,13 @@ class VersionBitsConditionChecker : public AbstractThresholdConditionChecker {
     int64_t BeginTime(const Consensus::Params& params) const { return params.vDeployments[id].nStartTime; }
     int64_t EndTime(const Consensus::Params& params) const { return params.vDeployments[id].nTimeout; }
     int Period(const Consensus::Params& params) const { return params.nMinerConfirmationWindow; }
-    int Threshold(const Consensus::Params& params) const { return params.nRuleChangeActivationThreshold; }
+    int Threshold(const Consensus::Params& params) const {
+        if (params.vDeployments[id].bit == params.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].bit &&
```
jtimon: why not just ```if (id == Consensus::DEPLOYMENT_SEGSIGNAL)``` ?

src/chainparams.cpp
```diff
@@ -97,7 +97,7 @@ class CMainParams : public CChainParams {
         consensus.vDeployments[Consensus::DEPLOYMENT_SEGWIT].nTimeout = 1510704000; // November 15th, 2017.
 
         // Deployment of SEGSIGNAL
-        consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].bit = 2;
+        consensus.vDeployments[Consensus::DEPLOYMENT_SEGSIGNAL].bit = 4;
```
jtimon: why bit 4 ? what's wrong with bit 2 ?

src/rpc/blockchain.cpp
```diff
@@ -1135,6 +1135,7 @@ UniValue getblockchaininfo(const JSONRPCRequest& request)
     softforks.push_back(SoftForkDesc("bip65", 4, tip, consensusParams));
     BIP9SoftForkDescPushBack(bip9_softforks, "csv", consensusParams, Consensus::DEPLOYMENT_CSV);
     BIP9SoftForkDescPushBack(bip9_softforks, "segwit", consensusParams, Consensus::DEPLOYMENT_SEGWIT);
+    BIP9SoftForkDescPushBack(bip9_softforks, "segsignal", consensusParams, Consensus::DEPLOYMENT_SEGSIGNAL);
```
jtimon: I would squash this commit and the previous one.

src/versionbits.cpp
jtimon: This commit seems unnecessary, just don't use Consensus::DEPLOYMENT_SEGSIGNAL on chains that shouldn't use it (like regtest and testnet3).

src/chainparams.cpp
jameshilliard: This was chosen because it was what was listed in the NYC agreement, I made a separate BIP that uses bit 2 however.

src/versionbits.cpp
jameshilliard: This is now defined in chainparams

src/versionbits.cpp
jameshilliard: Fixed

src/validation.cpp
jameshilliard: Any reason IsWitnessEnabled isn't also static?

jameshilliard:
I've [modified](https://github.com/bitcoin/bips/pull/547/files) the BIP91 spec slightly with a different threshold and immediate enforcement upon lock-in.

src/versionbits.cpp
jameshilliard: Is the behavior for that identical?

jameshilliard:
I've [updated](https://github.com/bitcoin/bips/pull/548/files) BIP91 to use an even smaller confirmation window and changed it so enforcement isn't until activation in order to give a ~2 day grace period between lock-in and enforcement.

luke-jr:
NACK, this does nothing for users, and is unnecessary with BIP148 around the corner.

earonesty:
ACK: avoiding a chain split in a coordinated manner is certainly not "doing nothing for users".  Since the -bip148 option is not available, a BIP91 feature is the best alternative.

dooglus:
Apparently BIP91 was locked in today.

How can I make sure my node follows the correct chain? Is this pull request on a branch somewhere or do I need to merge it in my own local repository?

And what about the thousands of other nodes out there? Shouldn't this PR be merged to allow them to track the correct chain?

jameshilliard:
@dooglus BIP91 is mostly a miner enforced soft fork, you can use [this tree](https://github.com/segsignal/bitcoin).

dooglus:
As I understand Bitcoins miners don't enforce anything. They mine blocks which are either valid or invalid. It is the users who decide whether to accept their blocks or not, and thereby enforce the rules.

I don't see any argument here against merging this PR other than Luke's inaccurate complaint that BIP91 does nothing for users. What am I missing?

jameshilliard:
@dooglus Both miners and users have the ability to enforce rules, a majority of miners can add a new rule without creating a chain split(assuming other nodes don't change their rules). Since a majority of miners are adding a rule with BIP91 the blocks will be valid to existing nodes. There's mostly 2 arguments I'm seeing, one is that more users should enforce the BIP91 rules to strengthen the rule enforcement, the other is that users should not so that there is more consistency among rule enforcement. I somewhat agree with the consistency argument, however any businesses accepting transactions without many confirmations should be running both BIP91 and non-BIP91 nodes and ensuring that the transactions are confirmed on both.

dooglus:
> however any businesses accepting transactions without many confirmations should be running both BIP91 and non-BIP91 nodes and ensuring that the transactions are confirmed on both

Exactly. That seems like an argument for merging this PR so that I can run a BIP91 node without having to trust or code review a non-core repository.

achow101:
Can this be squashed and possibly updated with the latest from the segsignal repo?

jameshilliard:
Closing this as enforcement of BIP91 is no longer required.

PR #10445, opened 2017-05-23
Fix and add test for empty chain and reorg consistency for gettxoutsetinfo.

gmaxwell:


gmaxwell:
This will crash on master. Fix is currently in the per-txout PR.

sipa:
utACK 4b3dd975a303a0d7efaec790c387f0285a4bb8c4

jonasschnelli (approved):
utACK 4b3dd975a303a0d7efaec790c387f0285a4bb8c4

paveljanik:
ACK https://github.com/bitcoin/bitcoin/commit/4b3dd975a303a0d7efaec790c387f0285a4bb8c4

FWIW: The new test crashes on master with:

```
2017-05-24 06:50:24.034000 TestFramework (INFO): Initializing test directory /var/folders/65/fn0h49r55k7779vg1b_h461r0000gn/T/test0wr7ofxj
Assertion failed: (valid_), function key, file leveldb/db/db_iter.cc, line 67.
```

Fixed in this PR.

test/functional/blockchain.py
```diff
@@ -53,6 +53,25 @@ def _test_gettxoutsetinfo(self):
         assert_equal(len(res['bestblock']), 64)
         assert_equal(len(res['hash_serialized']), 64)
 
+        b1hash = node.getblockhash(1)
```
jnewbery: Can you add a comment (or info log) here explaining what this test is for:

```py
self.log.info("Test that gettxoutsetinfo() works for blockchain with just the genesis block")
```

test/functional/blockchain.py
```diff
@@ -53,6 +53,25 @@ def _test_gettxoutsetinfo(self):
...
+        node.invalidateblock(b1hash)
+
+        res2 = node.gettxoutsetinfo()
+        assert_equal(res2['transactions'], 0)
+        assert_equal(res2['total_amount'], Decimal('0'))
+        assert_equal(res2['height'], 0)
+        assert_equal(res2['txouts'], 0)
```
jnewbery: Why not test all fields:

```py
assert_equal(res2['bestblock'], node.getblockhash(0)))
assert_equal(len(res2['hash_serialized']), 64)
```

test/functional/blockchain.py
```diff
@@ -53,6 +53,25 @@ def _test_gettxoutsetinfo(self):
...
+
+        node.reconsiderblock(b1hash)
+
```
jnewbery: Again, a comment/info log would be nice here:

```py
self.log.info("Test that gettxoutsetinfo() returns the same result after invalidate/reconsider block")
```

test/functional/blockchain.py
```diff
@@ -53,6 +53,25 @@ def _test_gettxoutsetinfo(self):
...
+        res3 = node.gettxoutsetinfo()
+        assert_equal(res['total_amount'], res3['total_amount'])
```
jnewbery: Consider just testing equality of res and res3 for succinctness:

```py
assert_equal(res, res3)
```

jnewbery (commented):
I can't review the code change in src/txdb.cpp since I don't know that code, but I can confirm that this test causes bitcoind to crash without the code change, and passes successfully with the code change.

A few nits inline. Also consider changing line 35 of the test from:

```py
        self.num_nodes = 2
```
to:

```py
        self.num_nodes = 1
```

I know that wsan't changed by this PR, but 2 nodes are unnecessary for this test, and changing it to 1 cuts the test time in half (from 5s to 2.5s on my pc) due to node startup/shutdown time.

sdaftuar:
ACK

test/functional/blockchain.py
gmaxwell: Equality will be the wrong comparison with the disk usage field which is about to be added.

test/functional/blockchain.py
gmaxwell: done

test/functional/blockchain.py
gmaxwell: done (and also added that to the first check)

test/functional/blockchain.py
gmaxwell: done

gmaxwell:
updated for @jnewbery 's nits.  (Thanks!)

jnewbery:
tested reACK https://github.com/bitcoin/bitcoin/pull/10445/commits/513da90cdd8463964e8c430b25f4326defc0bc96

sdaftuar:
utACK 513da90

paveljanik:
reACK 513da90

Issue #10402, opened 2017-05-15
error: make sure server is running and you are connecting to the correct RPC port


zottirik:
### Describe the issue
I installed bitcoin-cli but i can't run it.
I got an error : "make sure server is running and you are connecting to the correct RPC port"

### What version of bitcoin-core are you using?
0.14

### Machine specs:
- OS: Ubuntu Server 14
- CPU: Intel® Xeon® E3-1270
- RAM: 16 GB
- Disk size: 600 GB
- Disk Type (HD/SDD): SSD

sipa:
Is `bitcoind` running?

zottirik:
When I try to start bitcoind, I get this;

root@server:~# bitcoind -daemon start
Error: There is no RPC client functionality in bitcoind anymore. Use the bitcoin-cli utility instead.

MarcoFalke:
Try without the trailing "start".

You can use `bitcoind -h` or look at the [manpages](https://github.com/bitcoin/bitcoin/tree/master/doc/man) for futher information about the usage and command line options.

fanquake:
@zottirik is this still an issue? If not can you please close this.

laanwj:
> Error: There is no RPC client functionality in bitcoind anymore. Use the bitcoin-cli utility instead.

Eeh we should probably change that message at some point to something like `Error: Command line contains unexpected token *thing*, see bitcoind -h for a list of options`. Hardly anyone even remembers that bitcoind used to be a cli utility, let alone new users.

vinyyoo:
i face same problem,
can anyone suggest me how to solved it.
![errrorrrr](https://user-images.githubusercontent.com/32484213/31169540-953c1250-a916-11e7-865c-6269a99b675e.png)

MarcoFalke:
It might be more comfortable for you to use the GUI. Make sure to start with `bitcoin-qt -server -testnet`, so that you can connect with the `bitcoin-cli -testnet getblockchaininfo`.

Otherwise, you'd have to start the daemon first: `bitcoind -testnet`. (Should work without `-server` specified)

songtaoM:
just run `bitcoind --daemon`. It works for me

PR #10447, opened 2017-05-24
Make bitcoind invalid argument error message specific

laanwj:
The current message is not helpful. Hardly anyone even remembers that bitcoind used to be a cli utility, let alone new users. Print what the actual problem is.

Inspired by discussion here: https://github.com/bitcoin/bitcoin/issues/10402#issuecomment-303639064

Also remove the check for `bitcoin:` URIs. The expected output when accidentally passing a `bitcoin:` URI to `bitcoind` instead of `bitcoin-qt` would be an error not silent acceptance.

### DIFF
--- a/src/bitcoind.cpp
+++ b/src/bitcoind.cpp
@@ -117,17 +117,14 @@ bool AppInit(int argc, char* argv[])
             return false;
         }
 
-        // Command-line RPC
-        bool fCommandLine = false;
-        for (int i = 1; i < argc; i++)
-            if (!IsSwitchChar(argv[i][0]) && !boost::algorithm::istarts_with(argv[i], "bitcoin:"))
-                fCommandLine = true;
-
-        if (fCommandLine)
-        {
-            fprintf(stderr, "Error: There is no RPC client functionality in bitcoind anymore. Use the bitcoin-cli utility instead.\n");
-            exit(EXIT_FAILURE);
+        // Error out when loose non-argument tokens are encountered on command line
+        for (int i = 1; i < argc; i++) {
+            if (!IsSwitchChar(argv[i][0])) {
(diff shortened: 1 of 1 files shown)


jonasschnelli:
I guess most users already interpret `[...]There is no RPC client functionality [...]` as "oh I missed a dash somewhere".... 

utACK 8ec3a6bebb7c7d2a739e8827e3e0509638172504

laanwj:
> I guess most users already interpret [...]There is no RPC client functionality [...] as "oh I missed a dash somewhere"....

Yeah.... "somewhere" :-)

fanquake:
utACK 8ec3a6b

src/bitcoind.cpp
```diff
@@ -117,17 +117,14 @@ bool AppInit(int argc, char* argv[])
...
-            if (!IsSwitchChar(argv[i][0]) && !boost::algorithm::istarts_with(argv[i], "bitcoin:"))
-                fCommandLine = true;
-
-        if (fCommandLine)
-        {
-            fprintf(stderr, "Error: There is no RPC client functionality in bitcoind anymore. Use the bitcoin-cli utility instead.\n");
-            exit(EXIT_FAILURE);
+        // Error out when loose non-argument tokens are encountered on command line
+        for (int i = 1; i < argc; i++) {
+            if (!IsSwitchChar(argv[i][0]) && !boost::algorithm::istarts_with(argv[i], "bitcoin:")) {
```
laanwj: another realization - why are we checking for "bitcoin:" URLs at all bitcoind.cpp? Shouldn't they be treated as invalid tokens as well?

sipa:
utACK cd5622dcaf55ba08d33574e36bbd21be76aa3513

jtimon:
utACK cd5622dcaf55ba08d33574e36bbd21be76aa3513

PR #10450, opened 2017-05-24
Fix bumpfee rpc "errors" return value

ryanofsky:


sdaftuar:
utACK

jonasschnelli:
utACK 1b6602fbe66c0205acd4a33b9183080ec500fc72
I guess this is a tiny API change and maybe deserves a short mentioning in the release notes?

ryanofsky:
> I guess this is a tiny API change and maybe deserves a short mentioning in the release notes?

It appears the bug was introduced after the last release (in https://github.com/bitcoin/bitcoin/commit/0337a39d31603eb2c723a560410357f186763dc2#diff-df7d84ff2f53fcb2a0dc15a3a51e55ceL3093), so I don't think the release notes need to be updated.

jonasschnelli:
Oh! I see, this is an refactoring-bug fix. Sorry for the slouchiness and this certainly doesn't require to be in the RN then.

laanwj:
utACK 1b6602f

PR #10454, opened 2017-05-25
Fix broken q4 test build

ryanofsky:
Error reported by @pstratem https://github.com/bitcoin/bitcoin/pull/10420#issuecomment-303908782


jonasschnelli:
Meh adding Qt4 (which is EOL) hacks for new (test) code, but I guess we still want to support it.
So utACK 620444bd591355b7aeeed9d023d19cc665367b35

gmaxwell (approved):
tested ACK.

Issue #7463, opened 2016-02-03
salvagewallet fails verification


MarcoFalke:
This is somewhat rare but still worth to take a look sometime. I can reproduce locally and [travis complains](https://travis-ci.org/bitcoin/bitcoin/jobs/106719585) in `wallet.py`:

```
check -salvagewallet

start_node: bitcoind started, calling bitcoin-cli -rpcwait getblockcount

No output has been received in the last 10 minutes, this potentially indicates a stalled build or something wrong with the build itself.

The build has been terminated
```

---

Edit:

This will now show up as

```
check -salvagewallet
Unexpected exception caught during testing: Exception('bitcoind exited with status 1 during initialization',)
```

laanwj:
Another one at https://travis-ci.org/bitcoin/bitcoin/jobs/109329790

laanwj:
I've noticed that this mostly happens on the 32 bit travis build (.3)

laanwj:
I don't think we're even sure whether -salvagewallet terminates. In this case every time the last line seem to call some bitcoin-cli command with -rpcwait. In case the client would crash, or exit, it would just wait forever.

MarcoFalke:
> In case the client would crash, or exit, it would just wait forever.

The client does not crash nor exit, iIrc. salvagewallet returns some sort of error and the client tries to shutdown. Somehow this fails and the "write-peers.dat-thread" just continues to run. Haven't had the time  to look into that yet...

laanwj:
So you managed to reproduce this locally?

MarcoFalke:
You can just run

```sh
$ for i in {0..10000}; do qa/pull-tester/rpc-tests.py wallet --nocleanup;if [ $? -ne 0 ];then echo; echo "fail"; sleep 1d;fi ;done
```

and come back 20 minutes later to dig through the dumps.

laanwj:
I did a similar thing, I put the part of wallet.py that does the salvage in an infinite loop.

Managed to make it hang after ~100 iterations.

Looking at the debug.log it didn't hang during the salvage, but manages to go into normal operation. And the node didn't start shutting down at all. I could even succesfully send it a stop command and it stopped running. This won't set the python script in motion again though. Curious.

laanwj:
Ok, I was right in my gist above. I can reproduce this 100% of the time now with a specific wallet:
After the salvage goes wrong (for whatever reason) the node will just exit.

[...]

PR #10457, opened 2017-05-26
Don't use fixed "wallet.bak"-filename during salvagewallet

jonasschnelli:
Simple ~~fix~~ change.

At the moment, during salvagewallet/recover, we rename the wallet to `wallet.<timestamp>.bak` regardless of a possible custom wallet filename (`-wallet=`).
This fix is "required" for multiwallet via #8694.

### DIFF
--- a/src/wallet/db.cpp
+++ b/src/wallet/db.cpp
@@ -142,7 +142,7 @@ void CDBEnv::MakeMock()
     fMockDb = true;
 }
 
-CDBEnv::VerifyResult CDBEnv::Verify(const std::string& strFile, recoverFunc_type recoverFunc, std::string& out_backup_filename)
+CDBEnv::VerifyResult CDBEnv::Verify(const std::string& strFile, bool (*recoverFunc)(const std::string& strFile))
 {
     LOCK(cs_db);
     assert(mapFileUseCount.count(strFile) == 0);
@@ -155,11 +155,11 @@ CDBEnv::VerifyResult CDBEnv::Verify(const std::string& strFile, recoverFunc_type
         return RECOVER_FAIL;
 
     // Try to recover:
-    bool fRecovered = (*recoverFunc)(strFile, out_backup_filename);
+    bool fRecovered = (*recoverFunc)(strFile);
     return (fRecovered ? RECOVER_OK : RECOVER_FAIL);
 }
 
-bool CDB::Recover(const std::string& filename, void *callbackDataIn, bool (*recoverKVcallback)(void* callbackData, CDataStream ssKey, CDataStream ssValue), std::string& newFilename)
+bool CDB::Recover(const std::string& filename, void *callbackDataIn, bool (*recoverKVcallback)(void* callbackData, CDataStream ssKey, CDataStream ssValue))
 {
     // Recovery procedure:
     // move wallet file to walletfilename.timestamp.bak
@@ -169,7 +169,7 @@ bool CDB::Recover(const std::string& filename, void *callbackDataIn, bool (*reco
     // Set -rescan so any missing transactions will be
     // found.
     int64_t now = GetTime();
-    newFilename = strprintf("%s.%d.bak", filename, now);
+    std::string newFilename = strprintf("wallet.%d.bak", now);
 
     int result = bitdb.dbenv->dbrename(NULL, filename.c_str(), NULL,
                                        newFilename.c_str(), DB_AUTO_COMMIT);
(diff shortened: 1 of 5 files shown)


laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/10457/commits/f4823abad5bd2c9a556f3ebf2b87160c94f693ed.

If I read correctly this will change the name to `<something>.dat.<timestamp>.bak` but that's okay.

laanwj (approved):

jnewbery:
Concept ACK. Not having this blocks multi-wallet.

I'd be a lot happier if the -salvagewallet test was being run. It's currently commented out of wallet.py because of intermittent failure (#7463). I'll spend some time seeing if I can fix those failures.

sipa:
Concept ACK

laanwj:
> It's currently commented out of wallet.py because of intermittent failure (#7463)

Yes I narrowed it down to a berkeleydb issue last time, and wasn't sure how to proceed:

https://github.com/bitcoin/bitcoin/issues/7463#issuecomment-201319622

MarcoFalke:
utACK f4823abad5bd2c9a556f3ebf2b87160c94f693ed (uncommented the salvagewallet check and run in a loop without issues)

jnewbery:
> uncommented the salvagewallet check and run in a loop without issues

I can reproduce this by running `wallet.py` 50 times through test_runner.py. From debug.log:

```
2017-06-05 15:46:54.510473 CDBEnv::Salvage: Database salvage found errors, all data may not be recoverable.
2017-06-05 15:46:54.510639 CDBEnv::Salvage: WARNING: Number of keys in data does not match number of values.
2017-06-05 15:46:54.510648 CDBEnv::Salvage: WARNING: Unexpected end of file while reading salvage output.
2017-06-05 15:46:54.510657 Salvage(aggressive) found 102 records
```

jnewbery:
Tested ACK with one small nit:

Please update the comment at the top of CDB::Recover() from:
`    // move wallet file to wallet.timestamp.bak`

to:
`    // move wallet file to <wallet_name>.<timestamp>.bak`

Otherwise looks good. I've tested f4823abad5bd2c9a556f3ebf2b87160c94f693ed and it renames the wallet backups correctly.

laanwj:
Needs rebase after #8694

jnewbery:
I think this is a much cleaner solution than was implemented in #8694. There's no need to pass around the wallet name just so it can be printed in a warning message. I think it's worth rebasing this on master and reverting the changes in https://github.com/bitcoin/bitcoin/pull/8694/commits/b823a4c9f6836c802803dbd265cb7451fb93b8e7

jonasschnelli:
Changed:
This now reverts https://github.com/bitcoin/bitcoin/commit/b823a4c9f6836c802803dbd265cb7451fb93b8e7 and adds the identical simple fix for the static wallet name.

ryanofsky:
PR title and summary are misleading because now this PR only does code cleanup without actually changing the filename.

Also, I prefer b823a4c9f6836c802803dbd265cb7451fb93b8e7 to this PR, because it prints a more meaningful log message, and I don't think it is a big deal that the recover function now takes 2 arguments instead of 1.

I think if you wanted to clean up this code, you would just call the recover function directly or through std::function instead through a raw function pointer.

jnewbery:
Tested ACK e9d09f23e3a097605730988f2dcb818a57dee745. I think this is definitely an improvement.

src/wallet/db.cpp
```diff
@@ -259,19 +259,18 @@ bool CDB::VerifyEnvironment(const std::string& walletFile, const fs::path& dataD
...
-        CDBEnv::VerifyResult r = bitdb.Verify(walletFile, recoverFunc, backup_filename);
+        CDBEnv::VerifyResult r = bitdb.Verify(walletFile, recoverFunc);
         if (r == CDBEnv::RECOVER_OK)
         {
             warningStr = strprintf(_("Warning: Wallet file corrupt, data salvaged!"
                                      " Original %s saved as %s in %s; if"
                                      " your balance or transactions are incorrect you should"
                                      " restore from a backup."),
-                                   walletFile, backup_filename, dataDir);
+                                   walletFile, "{walletfilename}.{timestamp}.bak", dataDir);
```
luke-jr: Why show a filename at all, if it's not even going to tell you which one?

jonasschnelli: I think it won't hurt. But no strong opinion.

src/wallet/db.h
```diff
@@ -55,8 +55,7 @@ class CDBEnv
     enum VerifyResult { VERIFY_OK,
                         RECOVER_OK,
                         RECOVER_FAIL };
-    typedef bool (*recoverFunc_type)(const std::string& strFile, std::string& out_backup_filename);
```
laanwj: Not sure what the state of this PR is, but I just saw this change: isn't it clearer to define a type here, instead of make the function type definition in-line in the parameter definition?

ryanofsky:
PR description and title are misleading and need to be updated to say this is only changing the `Wallet file corrupt, data salvaged!` error message (by removing the recovered filename), and making a slight code simplification (especially now that PR descriptoin will go into git history).

This PR is **not** fixing any current bug. It's just temporarily reintroducing a bug that was fixed a while ago, and then applying a slightly simpler fix. IMO, squashing commit history in this PR would be a good way to make it less confusing and avoid temporarily reintroducing a bug.

src/wallet/db.h
laanwj: And maybe use a std::function, that's more c++11 than a function pointer.

luke-jr:
I think it's better to show the actual filename, but so long as this is refactored cleanly (ie, keep the typedef), I don't care to argue against it.

jnewbery:
I'm marginally for this change because it's a worthwhile simplification to not pass the wallet name around simply for printing in an error message. I agree with @ryanofsky that PR comment should be changed and commits squashed.

laanwj:
Closing this for now. I think the change in itself is worthwhile, but the diff itself seems to be controversial, and it seems low priority.

PR #10459, opened 2017-05-26
I found I needed to use gmake on FreeBSD 10.3

dbolser:
Not sure how universal this is (i.e. FreeBSD 11).

Just running through this now.


dbolser:
See also: https://www.freshports.org/net-p2p/bitcoin/ which is old, but shows that gmake is required

laanwj:
Thanks, but this was already fixed in master, see #10245, which was backported recently and will be in the next 0.14 release.

PR #10460, opened 2017-05-26
Broadcast address every day, not 9 hours

sipa:
Discovered by @gmaxwell.


dcousens (approved):

gmaxwell:
Doesn't actually matter,  9 hours would also be a fine time-- but it wasn't what was intended there.  utACK.

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/10460/commits/010d68e8023a07ff85d58cd1aff80c6a63e8249d

practicalswift:
ACK 010d68e

laanwj:
utACK, though I agree that:

> Doesn't actually matter, 9 hours would also be a fine time

PR #10461, opened 2017-05-26
Update style guide

sipa:
As discussed in this week's IRC meeting.

### DIFF
--- a/doc/developer-notes.md
+++ b/doc/developer-notes.md
@@ -3,41 +3,64 @@ Developer Notes
 
 Various coding styles have been used during the history of the codebase,
 and the result is not very consistent. However, we're now trying to converge to
-a single style, so please use it in new code. Old code will be converted
-gradually and you are encouraged to use the provided
-[clang-format-diff script](/contrib/devtools/README.md#clang-format-diffpy)
-to clean up the patch automatically before submitting a pull request.
+a single style, which is specified below. When writing patches, favor the new
+style over attempting to mimick the surrounding style, except for move-only
+commits.
 
-- Basic rules specified in [src/.clang-format](/src/.clang-format).
+Do not submit patches solely to modify the style of existing code.
+
+- **Indentation and whitespace rules** as specified in
+[src/.clang-format](/src/.clang-format). You can use the provided
+[clang-format-diff script](/contrib/devtools/README.md#clang-format-diffpy)
+tool to clean up patches automatically before submission.
   - Braces on new lines for namespaces, classes, functions, methods.
   - Braces on the same line for everything else.
   - 4 space indentation (no tabs) for every block except namespaces.
   - No indentation for `public`/`protected`/`private` or for `namespace`.
   - No extra spaces inside parenthesis; don't do ( this )
   - No space after function names; one space after `if`, `for` and `while`.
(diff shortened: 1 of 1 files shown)


dcousens (approved):

paveljanik (approved):

practicalswift:
ACK 47d8441

Very good! Thanks for clarifying!

doc/developer-notes.md
```diff
@@ -3,41 +3,64 @@ Developer Notes
...
+- **Symbol naming conventions**. These are preferred in new code, but are not
+required when doing so would need changes to significant pieces of existing
+code.
+  - Variable and namespace names are all lowercase, and may use `_` to
+    separate words.
+    - Class member variables have a `m_` prefix.
+    - Global variables have a `g_` prefix.
+  - Constant names are all uppercase, and use `_` to separate words.
+  - Class names, function names and method names are CamelCase. Do not prefix
+    class names with `C`.
```
luke-jr: Why not prefix class names with `C`?

sipa:
It is always obvious from the usage that a name refers to a class and not a function/method, so no need to repeat that information in the name.

luke-jr:
Is it? `function(0, class())` seems fairly non-obvious.

gmaxwell:
Seems fine to me.

doc/developer-notes.md
```diff
@@ -3,41 +3,64 @@ Developer Notes
...
+
+- **Symbol naming conventions**. These are preferred in new code, but are not
+required when doing so would need changes to significant pieces of existing
+code.
+  - Variable and namespace names are all lowercase, and may use `_` to
+    separate words.
+    - Class member variables have a `m_` prefix.
+    - Global variables have a `g_` prefix.
+  - Constant names are all uppercase, and use `_` to separate words.
+  - Class names, function names and method names are CamelCase. Do not prefix
```
jonasschnelli: Not sure if we had a discussion about that already:
I often saw projects where the static class methods and global space functions where in `UpperCamelCase()` while the class methods where in `lowerCamelCase()` and I think this is useful.
But no strong opinion.

jonasschnelli:
ACK 47d84414662c09bc08cfc96f78457ddd602a20ac

doc/developer-notes.md
laanwj: Except for the qt code, this just documents the style already in use.
So while this sounds sensible for a new project, I don't think we should change this now.

doc/developer-notes.md
```diff
@@ -3,41 +3,64 @@ Developer Notes
...
 and the result is not very consistent. However, we're now trying to converge to
-a single style, so please use it in new code. Old code will be converted
-gradually and you are encouraged to use the provided
-[clang-format-diff script](/contrib/devtools/README.md#clang-format-diffpy)
-to clean up the patch automatically before submitting a pull request.
+a single style, which is specified below. When writing patches, favor the new
+style over attempting to mimick the surrounding style, except for move-only
+commits.
+
+Do not submit patches solely to modify the style of existing code.
```
laanwj: Thanks for adding this explicitly :+1:

laanwj:
utACK 47d8441

jtimon:
ACK

PR #10464, opened 2017-05-27
Introduce static DoWarning (simplify UpdateTip)

jtimon:
Simplifies things with DRY by the book. Better is better too or something. Just look at the code I guess...
Also translate "%d of last 100 blocks have unexpected version"

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -1814,6 +1814,16 @@ void PruneAndFlush() {
     FlushStateToDisk(state, FLUSH_STATE_NONE);
 }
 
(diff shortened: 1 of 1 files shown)


ryanofsky:
utACK f285145764bb0495539daeb59fc651bd5e4d844b. Can't argue with "better is better"

paveljanik:
utACK f285145

Great, we were already talking about this change earlier, but I was not able to find the corresponding issue.

sipa:
utACK f285145764bb0495539daeb59fc651bd5e4d844b

PR #10467, opened 2017-05-27
add SeedNodes to CConnman::Options

benma:
Start of a series of changes to clean up the instantiation of connman
by decoupling the command line arguments.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1386,11 +1386,6 @@ bool AppInitMain(boost::thread_group& threadGroup, CScheduler& scheduler)
         }
     }
 
-    if (gArgs.IsArgSet("-seednode")) {
-        BOOST_FOREACH(const std::string& strDest, gArgs.GetArgs("-seednode"))
-            connman.AddOneShot(strDest);
-    }
-
 #if ENABLE_ZMQ
     pzmqNotificationInterface = CZMQNotificationInterface::Create();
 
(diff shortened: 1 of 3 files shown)


benma:
Pinging @theuni

benma:
@fanquake I don't think the P2P label is accurate. It should be `Refactoring`, like #9897, which this PR is a subset of.

src/net.cpp
```diff
@@ -2212,6 +2212,10 @@ bool CConnman::Start(CScheduler& scheduler, std::string& strNodeError, Options c
 
     SetBestHeight(connOptions.nBestHeight);
 
+    BOOST_FOREACH(const std::string& strDest, connOptions.vSeedNodes) {
```
fanquake: Can you change this to not introduce new Boost code.

src/init.cpp
```diff
@@ -1659,6 +1654,11 @@ bool AppInitMain(boost::thread_group& threadGroup, CScheduler& scheduler)
     connOptions.nMaxOutboundTimeframe = nMaxOutboundTimeframe;
     connOptions.nMaxOutboundLimit = nMaxOutboundLimit;
 
+    if (gArgs.IsArgSet("-seednode")) {
+        BOOST_FOREACH(const std::string& strDest, gArgs.GetArgs("-seednode"))
```
fanquake: Even though your only moving this, if we're going to touch these lines, can you remove Boost.

src/net.cpp
benma: Done

src/init.cpp
benma: Done

theuni (approved):
Nice. Ridding CConnman of the argument parsing has been on my todo for a long time.

utACK 5d67526026cd41845e06e93e5ae216bc1a7b9596

PR #8547, opened 2016-08-19
Update btcdrak key with new expiry dates

btcdrak:

PR #10476, opened 2017-05-29
Update luke-jr's PGP key

luke-jr:
Same key, extended expiration date


laanwj:
Can't find the discussion quickly but I remember we decided not to do this, to only update keys when they actually change, not for expiration dates. This is why the instructions for getting the gpg keys in [contrib/gitian-keys/README.md](https://github.com/bitcoin/bitcoin/blob/master/contrib/gitian-keys/README.md) include a `gpg --refresh-keys` step.

luke-jr:
Hmm, I thought the discussion was about coordinating it so all of the keys were updated once a year together, but nothing ever materialised further...

fanquake:
The discussion was in #8547/#8603.

MarcoFalke:
I assume it comes with less review and maintenance overhead if there was a script that does this for all keys in one go.

PR #10478, opened 2017-05-30
rpc: Add listen address to incoming connections in `getpeerinfo`

laanwj:
This adds the listening address on which incoming connections were received to the CNode and CNodeStats structures.

The address is reported in `getpeerinfo`.

This can be useful for distinguishing connections received on different listening ports (e.g. when using a different listening port for Tor hidden service connections) or different networks.

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -340,6 +340,22 @@ bool CConnman::CheckIncomingNonce(uint64_t nonce)
     return true;
 }
 
+/** Get the bind address for a socket as CAddress */
+static CAddress GetBindAddress(SOCKET sock)
+{
+    CAddress addr_bind;
+    struct sockaddr_storage sockaddr_bind;
+    socklen_t sockaddr_bind_len = sizeof(sockaddr_bind);
+    if (sock != INVALID_SOCKET) {
+        if (!getsockname(sock, (struct sockaddr*)&sockaddr_bind, &sockaddr_bind_len)) {
+            addr_bind.SetSockAddr((const struct sockaddr*)&sockaddr_bind);
+        } else {
+            LogPrint(BCLog::NET, "Warning: getsockname failed\n");
+        }
+    }
+    return addr_bind;
+}
+
 CNode* CConnman::ConnectNode(CAddress addrConnect, const char *pszDest, bool fCountFailure)
 {
     if (pszDest == NULL) {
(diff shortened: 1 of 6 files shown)


src/net.cpp
```diff
@@ -1032,13 +1033,22 @@ void CConnman::AcceptConnection(const ListenSocket& hListenSocket) {
...
+    CAddress addr, addr_listen;
     int nInbound = 0;
     int nMaxInbound = nMaxConnections - (nMaxOutbound + nMaxFeeler);
 
-    if (hSocket != INVALID_SOCKET)
-        if (!addr.SetSockAddr((const struct sockaddr*)&sockaddr))
+    if (hSocket != INVALID_SOCKET) {
+        if (!addr.SetSockAddr((const struct sockaddr*)&sockaddr)) {
             LogPrintf("Warning: Unknown socket family\n");
+	}
```
ryanofsky: Should replace tab with spaces here.

src/rpc/net.cpp
```diff
@@ -133,6 +133,8 @@ UniValue getpeerinfo(const JSONRPCRequest& request)
         obj.push_back(Pair("addr", stats.addrName));
         if (!(stats.addrLocal.empty()))
             obj.push_back(Pair("addrlocal", stats.addrLocal));
+        if (stats.addrListen.IsValid())
+            obj.push_back(Pair("addrlisten", stats.addrListen.ToString()));
```
ryanofsky: getpeerinfo help string needs to be updated

src/net.h
```diff
@@ -506,6 +506,8 @@ class CNodeStats
     double dMinPing;
     std::string addrLocal;
     CAddress addr;
+    // Bind address that this connection came in on (for incoming connections)
```
ryanofsky: I think it would be helpful if you could add comments for `addr` and `addrLocal` above, they contain "ip address and port of the peer" and "local address" according to RPC documention, and say specifically how `addrListen` is different from `addrLocal`.

ryanofsky (commented):
Slightly tested ACK 63c71582352f62d1ba674f09a5c00b93dd6f9284.

I would maybe consider renaming addrlisten to addrbind and not just restricting it to incoming connections.

laanwj:
> I would maybe consider renaming addrlisten to addrbind and not just restricting it to incoming connections.

Good point. We could also call getsockname for for outgoing connections, though the meaning is somewhat more obscure.

src/net.h
laanwj: Yes, addrLocal is pretty much useless in most cases, as it's what is sent by the peer. Wish we could rename that one. But yes it needs to be documented.

gmaxwell: bogoip :P

src/rpc/net.cpp
```diff
@@ -133,8 +134,8 @@ UniValue getpeerinfo(const JSONRPCRequest& request)
         obj.push_back(Pair("addr", stats.addrName));
         if (!(stats.addrLocal.empty()))
             obj.push_back(Pair("addrlocal", stats.addrLocal));
-        if (stats.addrListen.IsValid())
-            obj.push_back(Pair("addrlisten", stats.addrListen.ToString()));
+        if (stats.addrBind.IsValid())
+            obj.push_back(Pair("addrlisten", stats.addrBind.ToString()));
```
ryanofsky: "addrlisten" string needs to be updated here

ryanofsky (commented):
utACK a501d01491745f5b78d4ae484d50c2bcdb5d4700 with one comment. Thanks for taking some of my suggestions.

laanwj:
@ryanofsky yes, should have gotten most of them, I still intend to add a python tests. Now that both incoming and outgoing connections have the `addrbind` it's possible to correlate 100% the connection ids between local nodes.

src/rpc/net.cpp
laanwj: Eh yes

laanwj:
Rebased, and added a basic python test.

PR #10480, opened 2017-05-30
Improve commit-check-script.sh

sipa:
This makes an improvement to commit-check-script.sh (which was introduced in #10189):
* Fail when a VERIFY SCRIPT is present in the commit body, but no `scripted-diff` exists in its title. This would have caught a bug in #10321 (which had `script-diff` instead of `scripted-diff`) by reducing the reviewer burden from noticing a correct `scripted-diff` to that or a correct VERIFY SCRIPT (using fairly fuzzy rules).

### DIFF
--- a/contrib/devtools/commit-script-check.sh
+++ b/contrib/devtools/commit-script-check.sh
@@ -20,20 +20,27 @@ RET=0
 PREV_BRANCH=`git name-rev --name-only HEAD`
 PREV_HEAD=`git rev-parse HEAD`
 for i in `git rev-list --reverse $1`; do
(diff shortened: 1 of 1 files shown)


jtimon:
Concept ACK!

theuni:
utACK 5432fc3d59ce013d6c27f440ad86fc1f16d6e384. I don't believe the second commit is necessary, though. The diff is already shown regardless of success, this should make it print twice.

sipa:
@theuni Removed the last commit. I somehow assumed that what it was printing out was the whole commit, not just the unscripted diff...

jtimon:
Perhaps it would be interesting to rebase #10502 just to check this (and #10502 ) slightly more (although 1238f13 has 2 utACKs )

PR #9522, opened 2017-01-12
[RPC] Fix decoderawtransaction decoding of segwit txs

achow101:
Makes decoderawtransaction check for the marker and flag bytes present in segwit transactions and decodes them as such in order to avoid situations where segwit transactions are decoded as non-segwit transactions.

As per IRC discussion: https://botbot.me/freenode/bitcoin-core-dev/2017-01-12/?msg=79232896&page=1

Edit: Instead of checking for the bytes, it will just decode the transactions with the segwit formatting as that is the same behavior.

PR #10481, opened 2017-05-30
Decodehextx scripts sanity check 

achow101:
This PR adds some sanity checking to `DecodeHexTx` to make sure that the scripts contain valid opcodes. The main purpose of doing this is to avoid decoding witness transactions as 0-input 1-output non-witness transactions. The scripts are likely to be garbage and contain invalid opcodes when decoded as non-witness.

There is an issue where other invalid opcodes (not just OP_INVALIDOPCODE) are not handled but I am not sure how to go about doing that.

This is related to #9522.

### DIFF
--- a/src/core_read.cpp
+++ b/src/core_read.cpp
@@ -88,10 +88,32 @@ CScript ParseScript(const std::string& s)
     return result;
 }
 
+// Check that all of the input and output scripts of a transaction contains valid opcodes
+bool CheckTxScriptsSanity(const CMutableTransaction& tx)
+{
+    // Check input scripts for non-coinbase txs
+    if (!CTransaction(tx).IsCoinBase()) {
+        for (unsigned int i = 0; i < tx.vin.size(); i++) {
+            if (!tx.vin[i].scriptSig.HasValidOps() || tx.vin[i].scriptSig.size() > MAX_SCRIPT_SIZE) {
+                return false;
+            }
+        }
+    }
+    // Check output scripts
+    for (unsigned int i = 0; i < tx.vout.size(); i++) {
+        if (!tx.vout[i].scriptPubKey.HasValidOps() || tx.vout[i].scriptPubKey.size() > MAX_SCRIPT_SIZE) {
+            return false;
+        }
+    }
+    
+    return true;
+}
+
 bool DecodeHexTx(CMutableTransaction& tx, const std::string& strHexTx, bool fTryNoWitness)
 {
-    if (!IsHex(strHexTx))
+    if (!IsHex(strHexTx)) {
         return false;
+    }
 
     std::vector<unsigned char> txData(ParseHex(strHexTx));
 
@@ -99,7 +121,7 @@ bool DecodeHexTx(CMutableTransaction& tx, const std::string& strHexTx, bool fTry
         CDataStream ssData(txData, SER_NETWORK, PROTOCOL_VERSION | SERIALIZE_TRANSACTION_NO_WITNESS);
         try {
             ssData >> tx;
-            if (ssData.eof()) {
+            if (ssData.eof() && CheckTxScriptsSanity(tx)) {
                 return true;
             }
         }
@@ -111,8 +133,9 @@ bool DecodeHexTx(CMutableTransaction& tx, const std::string& strHexTx, bool fTry
     CDataStream ssData(txData, SER_NETWORK, PROTOCOL_VERSION);
     try {
         ssData >> tx;
-        if (!ssData.empty())
+        if (!ssData.empty()) {
             return false;
+        }
     }
     catch (const std::exception&) {
         return false;
--- a/src/script/script.cpp
+++ b/src/script/script.cpp
@@ -273,7 +273,8 @@ bool CScript::HasValidOps() const
     CScript::const_iterator it = begin();
     while (it < end()) {
         opcodetype opcode;
-        if (!GetOp(it, opcode) || opcode > 0xb9) {
+        std::vector<unsigned char> item;
+        if (!GetOp(it, opcode, item) || opcode > MAX_OPCODE || item.size() > MAX_SCRIPT_ELEMENT_SIZE) {
             return false;
         }
     }
--- a/src/script/script.cpp
+++ b/src/script/script.cpp
@@ -267,3 +267,15 @@ std::string CScriptWitness::ToString() const
     }
     return ret + ")";
 }
+
+bool CScript::HasValidOps() const
+{
+    CScript::const_iterator it = begin();
+    while (it < end()) {
+        opcodetype opcode;
+        if (!GetOp(it, opcode) || opcode > 0xb9) {
+            return false;
+        }
+    }
+    return true;
+}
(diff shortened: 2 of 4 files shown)


src/core_read.cpp
```diff
@@ -88,18 +88,38 @@ CScript ParseScript(const std::string& s)
     return result;
 }
 
+// Check that all of the input and output scripts of a transaction contains valid opcodes
+bool CheckTxScriptsSanity(CMutableTransaction& tx)
+{
+    // Check input scripts
+    for(unsigned int i = 0; i < tx.vin.size(); i++) {
```
sipa: You should probably skip this loop for coinbase transactions, as those don't have meaningful input scripts.

Is testing the inputs even useful, as the function of this test is to distinguish between transactions that may errorneously be decoded as 0-input transaction?

Nit: space after `for` (and elsewhere).

dcousens (approved):

sipa:
utACK 6909865869d9f2582e33e936e770538dfc541cf7

sipa:
Something else you could consider is checking whether the total script is not too large (MAX_SCRIPT_SIZE) or the stack elements aren't too large (MAX_SCRIPT_ELEMENT_SIZE).

src/core_read.cpp
```diff
@@ -88,18 +88,40 @@ CScript ParseScript(const std::string& s)
     return result;
 }
 
+// Check that all of the input and output scripts of a transaction contains valid opcodes
+bool CheckTxScriptsSanity(CMutableTransaction& tx)
```
achow101: done

achow101:
addressed @sipa's comments.

src/script/script.cpp
```diff
@@ -267,3 +267,16 @@ std::string CScriptWitness::ToString() const
...
     return ret + ")";
 }
+
+bool CScript::HasValidOps() const
+{
+    CScript::const_iterator it = begin();
+    while (it < end()) {
+        opcodetype opcode;
+        CScript::const_iterator prev_it = it;
+        if (!GetOp(it, opcode) || opcode > 0xb9 || it - prev_it > MAX_SCRIPT_ELEMENT_SIZE) {
```
sipa: I think this isn't actually correct. MAX_SCRIPT_ELEMENT_SIZE is the maximum length of the data, but its encoding as a CScript push may be up to 2 bytes more. Perhaps you can use GetOp2, which returns the data being pushed?

achow101: done

src/core_read.cpp
```diff
@@ -88,18 +88,40 @@ CScript ParseScript(const std::string& s)
...
+    if (!CTransaction(tx).IsCoinBase()) {
+        for (unsigned int i = 0; i < tx.vin.size(); i++) {
+            if (!tx.vin[i].scriptSig.HasValidOps()) {
+                return false;
+            }
+        }
+    }
+    // Check output scripts
+    for (unsigned int i = 0; i < tx.vout.size(); i++) {
+        if (!tx.vout[0].scriptPubKey.HasValidOps() || tx.vout[0].scriptPubKey.size() > MAX_SCRIPT_SIZE) {
```
sipa: vout[i]

src/core_read.cpp
```diff
@@ -88,18 +88,40 @@ CScript ParseScript(const std::string& s)
     return result;
 }
 
+// Check that all of the input and output scripts of a transaction contains valid opcodes
+bool CheckTxScriptsSanity(const CMutableTransaction& tx)
+{
+    // Check input scripts for non-coinbase txs
+    if (!CTransaction(tx).IsCoinBase()) {
+        for (unsigned int i = 0; i < tx.vin.size(); i++) {
+            if (!tx.vin[i].scriptSig.HasValidOps()) {
```
sipa: No MAX_SCRIPT_SIZE test here?

achow101: done

sipa:
reutACK e76d516a19f09f546d6fb5f16f8b4c758c5287ed

src/script/script.cpp
```diff
@@ -267,3 +267,16 @@ std::string CScriptWitness::ToString() const
...
     return ret + ")";
 }
+
+bool CScript::HasValidOps() const
+{
+    CScript::const_iterator it = begin();
+    while (it < end()) {
+        opcodetype opcode;
+        std::vector<unsigned char> item;
+        if (!GetOp(it, opcode, item) || opcode > 0xb9 || item.size() > MAX_SCRIPT_ELEMENT_SIZE) {
```
laanwj: Let's define a constant for 0xb9, this is kind of a magic number here. `OP_NOP10` would work, but maybe it's better to explicitly mark what the maximum valid opcode is.

achow101: Done

roconnor-blockstream:
This sanity check doesn't work.  invalid opcodes are not illegal, and can even appear in real transactions.  For an example see 95f7832b6449348bd1f3bc2e6eb110785248427bbe13853f2d61da9a6e810800 on testnet.

sipa:
@roconnor-blockstream This test is only a heuristic to determine whether to prefer a non-witness interpretation of a transaction over a witness interpretation when it has 0 inputs. It's also only relevant for interpreting inputs to `decoderawtransaction` and `fundrawtransaction` (which are the only things that accept transactions with 0 inputs in the first place).

roconnor-blockstream:
We cannot use heuristics for deciding how to interpret ambiguous parameters to API calls such as `fundrawtransaction`.  It is a security hazard.

Imagine someone is using Bitcoin Core as part of some protocol.  They have assembled data from various parties, some of them generated adversarially.  The protocol safely validates the adversarial inputs and composes a valid, partial transaction, and passes it to `fundrawtransaction`, then sends that output to `signrawtransaction` and then passes it around or even broadcasts it with `sendrawtransaction`.

The system has validated the construction of their transaction, and the designers have tested it many times and believes it work.  However, it isn't secure.  After deployment an adversary tickles their data, perhaps by creating an output with invalid opcodes.  These adversarially provided output scripts are legal and pass protocol's validation engine.  However this causes the data passed to `fundrawtransaction` to be interpreted differently, either as a segwit transaction when the protocol thinks it is generating a legacy transaction, or vice versa.  Now `fundrawtransaciton` is generating a transaction that is different than the protocol intends, and it is probably a transaction sending the funds somewhere else.  Nobody expected that fine details about the contents of output scripts to change how the transaction being passed is parsed, that would be very surprising, so the software just assumes `fundrawtransaction` output is the transaction it thinks it has created, and blindly sends it onto `signrawtransaction`. With that transaction now signed, who knows where the money will end up.

sipa:
@roconnor-blockstream Then what do you suggest? The only full solution is not using the same format for partial transactions (something that @achow101's PSBT format accomplishes). Anything will inherently risk being ambiguous.

roconnor-blockstream:
Yes, ultimately this format for partial transactions must be retired.

In the mean time I suggest either adding a, preferably mandatory, flag to fundrawtransaction to tell it whether the parameter is a partially signed segwit transaction or a partially signed legacy transaction.
If you choose not to make the flag mandatory, then you must parse the parameter both ways and fail of both parses are successful.  For consistency, and safety, decoderawtransaction should operate the same way.

(instead of a flag, you could also create two new API calls for fundrawlegacytransaction and fundrawsegwittransaction).

Perhaps there are some other ideas.

PR #10486, opened 2017-05-31
devtools: Retry after signing fails in github-merge

laanwj:
When signing fails, go back to the sign/exit prompt instead of exiting the script.

Nowadays I use a signing token for my GPG key, which can sometimes fail (especially when it's accidentally not plugged in :') ). This avoids having to re-do the merge in that case.


MarcoFalke:
utACK 1983c8772cdd4ba4389f7f9880ec394da73d4a3b

fanquake:
utACK 1983c87

PR #10489, opened 2017-05-31
build: silence gcc7's implicit fallthrough warning

theuni:
This is a well-intentioned but realistically annoying warning. Unfortunately, it's too easy for a warning in one header to cause dozens of repeated warnings.

See [here](https://developers.redhat.com/blog/2017/03/10/wimplicit-fallthrough-in-gcc-7) for details.

If it were just our code, I'd add in the (non-standard) annotations. But it's also an issue in leveldb and tinyformat, and would require substantial autoconf hackery to avoid.

Fwiw, None of the existing warnings turned up are real bugs.


laanwj:
Yes, this one seems more trouble than it's worth (with Wshadow still fresh in the mind). utACK

fanquake:
utACK cf390df

PR #10493, opened 2017-06-01
Use range-based for loops (C++11) when looping over map elements

practicalswift:
Before this commit:

```c++
for (std::map<T1, T2>::iterator x = y.begin(); x != y.end(); ++x) {
    T1 z = (*x).first;
    …
}
```

After this commit:

```c++
for (auto& x : y) {
    T1 z = x.first;
    …
}
```

### DIFF
--- a/src/addrman.cpp
+++ b/src/addrman.cpp
@@ -390,9 +390,9 @@ int CAddrMan::Check_()
     if (vRandom.size() != nTried + nNew)
         return -7;
 
-    for (std::map<int, CAddrInfo>::iterator it = mapInfo.begin(); it != mapInfo.end(); it++) {
-        int n = (*it).first;
-        CAddrInfo& info = (*it).second;
+    for (const auto& entry : mapInfo) {
+        int n = entry.first;
+        const CAddrInfo& info = entry.second;
         if (info.fInTried) {
             if (!info.nLastSuccess)
                 return -1;
--- a/src/qt/bantablemodel.cpp
+++ b/src/qt/bantablemodel.cpp
@@ -55,11 +55,11 @@ public:
 #if QT_VERSION >= 0x040700
         cachedBanlist.reserve(banMap.size());
 #endif
-        for (banmap_t::iterator it = banMap.begin(); it != banMap.end(); it++)
+        for (const auto& entry : banMap)
         {
             CCombinedBan banEntry;
-            banEntry.subnet = (*it).first;
-            banEntry.banEntry = (*it).second;
+            banEntry.subnet = entry.first;
+            banEntry.banEntry = entry.second;
             cachedBanlist.append(banEntry);
         }
 
--- a/src/rpc/net.cpp
+++ b/src/rpc/net.cpp
@@ -569,11 +569,11 @@ UniValue listbanned(const JSONRPCRequest& request)
     g_connman->GetBanned(banMap);
 
     UniValue bannedAddresses(UniValue::VARR);
-    for (banmap_t::iterator it = banMap.begin(); it != banMap.end(); it++)
+    for (const auto& entry : banMap)
     {
-        CBanEntry banEntry = (*it).second;
+        const CBanEntry& banEntry = entry.second;
         UniValue rec(UniValue::VOBJ);
-        rec.push_back(Pair("address", (*it).first.ToString()));
+        rec.push_back(Pair("address", entry.first.ToString()));
         rec.push_back(Pair("banned_until", banEntry.nBanUntil));
         rec.push_back(Pair("ban_created", banEntry.nCreateTime));
         rec.push_back(Pair("ban_reason", banEntry.banReasonToString()));
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -3249,8 +3249,8 @@ void PruneOneBlockFile(const int fileNumber)
 {
     LOCK(cs_LastBlockFile);
 
-    for (BlockMap::iterator it = mapBlockIndex.begin(); it != mapBlockIndex.end(); ++it) {
-        CBlockIndex* pindex = it->second;
+    for (const auto& entry : mapBlockIndex) {
+        CBlockIndex* pindex = entry.second;
         if (pindex->nFile == fileNumber) {
             pindex->nStatus &= ~BLOCK_HAVE_DATA;
             pindex->nStatus &= ~BLOCK_HAVE_UNDO;
@@ -3802,8 +3802,8 @@ bool RewindBlockIndex(const CChainParams& params)
     // Reduce validity flag and have-data flags.
     // We do this after actual disconnecting, otherwise we'll end up writing the lack of data
     // to disk before writing the chainstate, resulting in a failure to continue if interrupted.
-    for (BlockMap::iterator it = mapBlockIndex.begin(); it != mapBlockIndex.end(); it++) {
-        CBlockIndex* pindexIter = it->second;
+    for (const auto& entry : mapBlockIndex) {
+        CBlockIndex* pindexIter = entry.second;
 
         // Note: If we encounter an insufficiently validated block that
         // is on chainActive, it must be because we are a pruning node, and
@@ -4078,8 +4078,8 @@ void static CheckBlockIndex(const Consensus::Params& consensusParams)
 
     // Build forward-pointing map of the entire block tree.
     std::multimap<CBlockIndex*,CBlockIndex*> forward;
-    for (BlockMap::iterator it = mapBlockIndex.begin(); it != mapBlockIndex.end(); it++) {
-        forward.insert(std::make_pair(it->second->pprev, it->second));
+    for (auto& entry : mapBlockIndex) {
+        forward.insert(std::make_pair(entry.second->pprev, entry.second));
     }
 
     assert(forward.size() == mapBlockIndex.size());
--- a/src/addrman.h
+++ b/src/addrman.h
@@ -313,9 +313,9 @@ public:
         s << nUBuckets;
         std::map<int, int> mapUnkIds;
         int nIds = 0;
-        for (std::map<int, CAddrInfo>::const_iterator it = mapInfo.begin(); it != mapInfo.end(); it++) {
-            mapUnkIds[(*it).first] = nIds;
-            const CAddrInfo &info = (*it).second;
+        for (const auto& entry : mapInfo) {
+            mapUnkIds[entry.first] = nIds;
+            const CAddrInfo &info = entry.second;
             if (info.nRefCount) {
                 assert(nIds != nNew); // this means nNew was wrong, oh ow
                 s << info;
@@ -323,8 +323,8 @@ public:
             }
         }
         nIds = 0;
-        for (std::map<int, CAddrInfo>::const_iterator it = mapInfo.begin(); it != mapInfo.end(); it++) {
-            const CAddrInfo &info = (*it).second;
+        for (const auto& entry : mapInfo) {
+            const CAddrInfo &info = entry.second;
             if (info.fInTried) {
                 assert(nIds != nTried); // this means nTried was wrong, oh ow
                 s << info;
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -110,13 +110,13 @@ bool GetLocal(CService& addr, const CNetAddr *paddrPeer)
     int nBestReachability = -1;
     {
         LOCK(cs_mapLocalHost);
-        for (std::map<CNetAddr, LocalServiceInfo>::iterator it = mapLocalHost.begin(); it != mapLocalHost.end(); it++)
+        for (const auto& entry : mapLocalHost)
         {
-            int nScore = (*it).second.nScore;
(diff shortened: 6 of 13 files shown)


dcousens (approved):
utACK

src/addrman.cpp
```diff
@@ -390,7 +390,7 @@ int CAddrMan::Check_()
     if (vRandom.size() != nTried + nNew)
         return -7;
 
-    for (std::map<int, CAddrInfo>::iterator it = mapInfo.begin(); it != mapInfo.end(); it++) {
+    for (auto& it : mapInfo) {
         int n = (*it).first;
```
benma: Should be `it.first` and `it.second`.

practicalswift:
@benma Thanks for reviewing! Fixed! :-)

benma:
utACK cf02bb1a6c3c0077f1cf37f60e13d6cb71fd9187

practicalswift:
Rebased!

sipa:
Overall comment: the loop variables should probably not keep the name `it`, as that may confuse someone into thinking it refers to an iterators, while it is a container element reference.

practicalswift:
@sipa Good point! Any suggestion on an appropriate generic variable name to use instead? Perhaps `pair` in the case of maps or `o` otherwise?

sipa:
@practicalswift Perhaps `entry` for maps?

practicalswift:
@sipa Ah, yes, `entry` is better. Fixing it now!

practicalswift:
@sipa Fixed! :-)

src/rpc/net.cpp
```diff
@@ -567,11 +567,11 @@ UniValue listbanned(const JSONRPCRequest& request)
     g_connman->GetBanned(banMap);
 
     UniValue bannedAddresses(UniValue::VARR);
-    for (banmap_t::iterator it = banMap.begin(); it != banMap.end(); it++)
+    for (auto& entry : banMap)
```
MarcoFalke: Since you are touching all those lines anyway, you can change them to const references for free. Something like the following diff should do:


```diff
diff --git a/src/addrdb.h b/src/addrdb.h
index c3d509b..e0a49c8 100644
--- a/src/addrdb.h
+++ b/src/addrdb.h
@@ -61,7 +61,7 @@ public:
         banReason = BanReasonUnknown;
     }
 
-    std::string banReasonToString()
+    std::string banReasonToString() const
     {
         switch (banReason) {
         case BanReasonNodeMisbehaving:
diff --git a/src/addrman.cpp b/src/addrman.cpp
index 0edf111..2f7c014 100644
--- a/src/addrman.cpp
+++ b/src/addrman.cpp
@@ -390,9 +390,9 @@ int CAddrMan::Check_()
     if (vRandom.size() != nTried + nNew)
         return -7;
 
-    for (auto& entry : mapInfo) {
+    for (const auto& entry : mapInfo) {
         int n = entry.first;
-        CAddrInfo& info = entry.second;
+        const CAddrInfo& info = entry.second;
         if (info.fInTried) {
             if (!info.nLastSuccess)
                 return -1;
diff --git a/src/net.cpp b/src/net.cpp
index e6c5da8..c071fec 100644
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -106,7 +106,7 @@ bool GetLocal(CService& addr, const CNetAddr *paddrPeer)
     int nBestReachability = -1;
     {
         LOCK(cs_mapLocalHost);
-        for (auto& entry : mapLocalHost)
+        for (const auto& entry : mapLocalHost)
         {
             int nScore = entry.second.nScore;
             int nReachability = entry.first.GetReachabilityFrom(paddrPeer);
@@ -472,10 +472,10 @@ bool CConnman::IsBanned(CNetAddr ip)
     bool fResult = false;
     {
         LOCK(cs_setBanned);
-        for (auto& entry : setBanned)
+        for (const auto& entry : setBanned)
         {
-            CSubNet subNet = entry.first;
-            CBanEntry banEntry = entry.second;
+            const CSubNet& subNet = entry.first;
+            const CBanEntry& banEntry = entry.second;
 
             if(subNet.Match(ip) && GetTime() < banEntry.nBanUntil)
                 fResult = true;
diff --git a/src/qt/transactiontablemodel.cpp b/src/qt/transactiontablemodel.cpp
index 5b75886..99c2475 100644
--- a/src/qt/transactiontablemodel.cpp
+++ b/src/qt/transactiontablemodel.cpp
@@ -82,7 +82,7 @@ public:
         cachedWallet.clear();
         {
             LOCK2(cs_main, wallet->cs_wallet);
-            for (auto& entry : wallet->mapWallet)
+            for (const auto& entry : wallet->mapWallet)
             {
                 if (TransactionRecord::showTransaction(entry.second))
                     cachedWallet.append(TransactionRecord::decomposeTransaction(wallet, entry.second));
diff --git a/src/rpc/net.cpp b/src/rpc/net.cpp
index 85df33a..345befd 100644
--- a/src/rpc/net.cpp
+++ b/src/rpc/net.cpp
@@ -570,9 +570,9 @@ UniValue listbanned(const JSONRPCRequest& request)
     g_connman->GetBanned(banMap);
 
     UniValue bannedAddresses(UniValue::VARR);
-    for (auto& entry : banMap)
+    for (const auto& entry : banMap)
     {
-        CBanEntry banEntry = entry.second;
+        const CBanEntry& banEntry = entry.second;
         UniValue rec(UniValue::VOBJ);
         rec.push_back(Pair("address", entry.first.ToString()));
         rec.push_back(Pair("banned_until", banEntry.nBanUntil));
diff --git a/src/test/coins_tests.cpp b/src/test/coins_tests.cpp
index db39ac6..6ee1f7d 100644
--- a/src/test/coins_tests.cpp
+++ b/src/test/coins_tests.cpp
@@ -88,7 +88,7 @@ public:
         // Manually recompute the dynamic usage of the whole data, and compare it.
         size_t ret = memusage::DynamicUsage(cacheCoins);
         size_t count = 0;
-        for (auto& entry : cacheCoins) {
+        for (const auto& entry : cacheCoins) {
             ret += entry.second.coin.DynamicMemoryUsage();
             ++count;
         }
@@ -182,7 +182,7 @@ BOOST_AUTO_TEST_CASE(coins_cache_simulation_test)
 
         // Once every 1000 iterations and at the end, verify the full cache.
         if (insecure_rand() % 1000 == 1 || i == NUM_SIMULATION_ITERATIONS - 1) {
-            for (auto& entry : result) {
+            for (const auto& entry : result) {
                 bool have = stack.back()->HaveCoin(entry.first);
                 const Coin& coin = stack.back()->AccessCoin(entry.first);
                 BOOST_CHECK(have == !coin.IsSpent());
@@ -413,7 +413,7 @@ BOOST_AUTO_TEST_CASE(updatecoins_simulation_test)
 
         // Once every 1000 iterations and at the end, verify the full cache.
         if (insecure_rand() % 1000 == 1 || i == NUM_SIMULATION_ITERATIONS - 1) {
-            for (auto& entry : result) {
+            for (const auto& entry : result) {
                 bool have = stack.back()->HaveCoin(entry.first);
                 const Coin& coin = stack.back()->AccessCoin(entry.first);
                 BOOST_CHECK(have == !coin.IsSpent());
diff --git a/src/wallet/rpcwallet.cpp b/src/wallet/rpcwallet.cpp
index a5143aa..e85732e 100644
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -1263,7 +1263,7 @@ UniValue ListReceived(CWallet * const pwallet, const UniValue& params, bool fByA
 
     if (fByAccounts)
     {
-        for (auto& entry : mapAccountTally)
+        for (const auto& entry : mapAccountTally)
         {
             CAmount nAmount = entry.second.nAmount;
             int nConf = entry.second.nConf;

**Comment by practicalswift** [comment]
@MarcoFalke Good idea! You're suggestions are now part of the PR :-)

**Comment by practicalswift** [comment]
Rebased!

**Comment by sipa** [comment]
Needs rebase

**Comment by practicalswift** [comment]
Rebased!

**Comment by jtimon** [comment]
utACK 875ebee20e048892d999c41acb7100d0fdac1f91

File: src/qt/bantablemodel.cpp [hunk 3]
```diff
@@ -55,11 +55,11 @@ class BanTablePriv
 #if QT_VERSION >= 0x040700
         cachedBanlist.reserve(banMap.size());
 #endif
-        for (banmap_t::iterator it = banMap.begin(); it != banMap.end(); it++)
+        for (auto& entry : banMap)
```
sipa: I think this can be `const`

src/validation.cpp
```diff
@@ -3120,8 +3120,8 @@ static uint64_t CalculateCurrentUsage()
 /* Prune a block file (modify associated database entries)*/
 void PruneOneBlockFile(const int fileNumber)
 {
-    for (BlockMap::iterator it = mapBlockIndex.begin(); it != mapBlockIndex.end(); ++it) {
-        CBlockIndex* pindex = it->second;
+    for (auto& entry : mapBlockIndex) {
```
sipa: Can be `const` (the loop modifies the entries pointed to by `mapBlockIndex`, not the entries directly.

src/validation.cpp
```diff
@@ -3570,8 +3570,8 @@ bool RewindBlockIndex(const CChainParams& params)
     // Reduce validity flag and have-data flags.
     // We do this after actual disconnecting, otherwise we'll end up writing the lack of data
     // to disk before writing the chainstate, resulting in a failure to continue if interrupted.
-    for (BlockMap::iterator it = mapBlockIndex.begin(); it != mapBlockIndex.end(); it++) {
-        CBlockIndex* pindexIter = it->second;
+    for (auto& entry : mapBlockIndex) {
```
sipa: Same, can be `const`.

sipa (commented):
I believe there may be a few more possible `const`s missing.

practicalswift:
@sipa Thanks! `const`:s added :-)

promag (commented):
You are missing reverse iteration. For instance, https://github.com/bitcoin/bitcoin/blob/master/src/wallet/wallet.cpp#L3722.

Edit, fwiw, see [reverse_iterator.h](https://github.com/bitcoin/bitcoin/blob/master/src/reverse_iterator.h).

practicalswift:
@promag Not part of the scope of this PR :-)

promag:
@practicalswift commit message suggests otherwise, range based loop over map entries regardless of direction.

practicalswift:
@promag Feel free to provide an unified diff and I'll incorporate it :-)

practicalswift:
@promag Is it these two you are referring to?

```
src/wallet/rpcwallet.cpp:    for (CWallet::TxItems::const_reverse_iterator it = txOrdered.rbegin(); it != txOrdered.rend(); ++it)
src/wallet/wallet.cpp:            for (auto it = txOrdered.rbegin(); it != txOrdered.rend(); ++it) {
```

promag:
Yes, they are multimap, but still a map. IMO it makes sense to also replace those for range loop.

jtimon:
You can use reverse_iterate() for those.

promag:
@jtimon that was my [suggestion](https://github.com/bitcoin/bitcoin/pull/10493#pullrequestreview-509249509).

practicalswift:
Rebased!

jtimon:
fast review ACK

practicalswift:
Rebased!

practicalswift:
Rebased again! Anyone willing to re-review? I thinks this one should be ready for merge :-)

jtimon:
fast review re-ACK

promag (commented):
Check this:

http://github.com/practicalswift/bitcoin/blob/680bc2cbb34d6bedd0e64b17d0555216572be4c8/src/wallet/wallet.cpp#L827

practicalswift:
@promag As an example of a missing coversion? :-)

practicalswift:
@promag What about it? :-)

promag:
@practicalswift use range-based there too?

practicalswift:
@promag Yes, but with an additional condition (`account.vchPubKey.IsValid()`). The scope of this PR is to cover the simplest possible case only (`for (std::map<T1, T2>::iterator x = y.begin(); x != y.end(); ++x) {`) to make reviewing trivial.

ajtowns (approved):
TestedACK 680bc2cbb34d6bedd0e64b17d0555216572be4c8. Looks good to me. There's a bunch of untouched for()'s with vector::iterator and set::iterator's that could presumably have similar refactors in a later commit.

practicalswift:
@ajtowns Thanks for the review and the tested ACK! This PR is intentionally only touching `std::map` to limit the scope and make it easy to review :-)

practicalswift:
Does this PR stand a chance of being merged? If not I'll close it :-)

sipa:
utACK 680bc2cbb34d6bedd0e64b17d0555216572be4c8

MarcoFalke:
utACK 680bc2cbb34d6bedd0e64b17d0555216572be4c8

practicalswift:
Ready for merge? :-)

Issue #10163, opened 2017-04-07
Downloading sipa's seeds.txt yields a 404 error


JeremyRand:
### Describe the issue

Downloading `seeds.txt` from sipa's DNS seed yields a 404 error.

### Can you reliably reproduce the issue?

Yes.

#### If so, please list the steps to reproduce below:

1. Run the following command from a terminal: `curl -s http://bitcoin.sipa.be/seeds.txt`.
2. (My OS is Whonix 13; I have no reason to expect different results on other OS's but I haven't tried.)
3. (I tried replacing `http://` with `https://` since I was wondering if a Tor exit was interfering with the response; the results were unchanged.)

### Expected behaviour

A valid seeds.txt file should be displayed by curl.

### Actual behaviour

I get the following output:

```
$ curl -s http://bitcoin.sipa.be/seeds.txt
<?xml version="1.0" encoding="iso-8859-1"?>
<!DOCTYPE html PUBLIC "-//W3C//DTD XHTML 1.0 Transitional//EN"
         "http://www.w3.org/TR/xhtml1/DTD/xhtml1-transitional.dtd">
<html xmlns="http://www.w3.org/1999/xhtml" xml:lang="en" lang="en">
 <head>
  <title>404 - Not Found</title>
 </head>
 <body>
  <h1>404 - Not Found</h1>
 </body>
</html>
```

### Machine specs:

- OS: Whonix 13.  (That means all traffic is tunneled through Tor.)
- Other specs: believed to not be relevant.

laanwj:
Would make sense to tag @sipa  (or maybe even mail him privately or ask on IRC). Not sure why this is a bitcoin core issue.

jonasschnelli:
As far as I know is there no promise to provide the `seed.txt` as a 24/7 API. I guess this is not a Bitcoin-Core issue. The only timeframe where the seed.txt is used (can be used) is when updating the static internal ip list.
You can run your own seeder and create your own seed.txt.

JeremyRand:
-----BEGIN PGP SIGNED MESSAGE-----
Hash: SHA512

I don't think I (or anyone else) alleged that the seeds.txt API (or
any other non-critical Bitcoin Core development API) needs to have
perfect uptime.  That said, for transparency and educational purposes,
it is beneficial for average bystanders to be able to learn how the
release process works.  This would mean either making an effort to
keep the seeds.txt API running where feasible (in which case it's
unclear to me why pointing out that it's down would be problematic on
[...]

PR #10495, opened 2017-06-01
contrib: Update location of seeds.txt

laanwj:
Update the steps for updating the hardcoded seed nodes to point to the new filename on @sipa's server, and add command to decompress it.

Ref: #10163

### DIFF
--- a/contrib/seeds/README.md
+++ b/contrib/seeds/README.md
@@ -8,7 +8,7 @@ and remove old versions as necessary.
 
 The seeds compiled into the release are created from sipa's DNS seed data, like this:
 
-    curl -s http://bitcoin.sipa.be/seeds.txt > seeds_main.txt
+    curl -s http://bitcoin.sipa.be/seeds.txt.gz | gzip -dc > seeds_main.txt
     python3 makeseeds.py < seeds_main.txt > nodes_main.txt
     python3 generate-seeds.py . > ../../src/chainparamsseeds.h
 


jonasschnelli:
Tested ACK ac9df130a507177ec77ef3f5f4fa43c0c5ef57c3

contrib/seeds/README.md
```diff
@@ -8,7 +8,7 @@ and remove old versions as necessary.
 
 The seeds compiled into the release are created from sipa's DNS seed data, like this:
 
-    curl -s http://bitcoin.sipa.be/seeds.txt > seeds_main.txt
+    curl -s http://bitcoin.sipa.be/seeds.txt.gz | gunzip > seeds_main.txt
```
practicalswift: Nit: Isn't `| gzip -dc` the preferred over `| gunzip`?

laanwj: For what reason?

practicalswift: @laanwj Good question! :-)  Perhaps it is a legacy thing, but for some reason I have it hard wired that `gzip -dc` is used when operating on stdin/stdout and `gunzip` is for unpacking `file.gz`, removing the original file and leaving `file`. They might be totally equivalent nowadays. If so, ignore my minor nit :-)

paveljanik:
IIRC on some systems, gunzip symlink was not available in the default install.

contrib/seeds/README.md
practicalswift: Found this thead: https://lists.gnu.org/archive/html/bug-automake/2009-10/msg00012.html

laanwj:
> IIRC on some systems, gunzip symlink was not available in the default install.

I tried on FreeBSD, OpenBSD, Linux (Ubuntu) and `gunzip` worked out of the box at least... Also these are not general instructions for end-users, but for ourselves. I personally find `gunzip` much easier to remember. But meh I'm fine with changing it.

fanquake:
utACK https://github.com/bitcoin/bitcoin/pull/10495/commits/ac9cd953d91e2448d24851b2f4973c7be8f4c3cd

practicalswift:
ACK ac9cd95

PR #10496, opened 2017-06-01
Add Binds, WhiteBinds, Whitelistedrange to CConnman::Options

benma:
Subset of #9897, with the goal to decouple command line arg parsing and make AppInitMain() smaller.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1289,16 +1289,6 @@ bool AppInitMain(boost::thread_group& threadGroup, CScheduler& scheduler)
         }
     }
 
-    if (gArgs.IsArgSet("-whitelist")) {
-        for (const std::string& net : gArgs.GetArgs("-whitelist")) {
-            CSubNet subnet;
-            LookupSubNet(net.c_str(), subnet);
-            if (!subnet.IsValid())
-                return InitError(strprintf(_("Invalid netmask specified in -whitelist: '%s'"), net));
-            connman.AddWhitelistedRange(subnet);
-        }
-    }
-
     // Check for host lookup allowed before parsing any network related parameters
     fNameLookup = GetBoolArg("-dns", DEFAULT_NAME_LOOKUP);
 
@@ -1660,6 +1650,16 @@ bool AppInitMain(boost::thread_group& threadGroup, CScheduler& scheduler)
 
     connOptions.nMaxOutboundTimeframe = nMaxOutboundTimeframe;
     connOptions.nMaxOutboundLimit = nMaxOutboundLimit;
+
+    if (gArgs.IsArgSet("-whitelist")) {
+        for (const auto& net : gArgs.GetArgs("-whitelist")) {
+            CSubNet subnet;
+            LookupSubNet(net.c_str(), subnet);
+            if (!subnet.IsValid())
+                return InitError(strprintf(_("Invalid netmask specified in -whitelist: '%s'"), net));
+            connOptions.vWhitelistedRange.push_back(subnet);
+        }
+    }
 
     if (gArgs.IsArgSet("-seednode")) {
         connOptions.vSeedNodes = gArgs.GetArgs("-seednode");
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -88,14 +88,6 @@ static CZMQNotificationInterface* pzmqNotificationInterface = NULL;
 #define MIN_CORE_FILEDESCRIPTORS 150
 #endif
 
-/** Used to pass flags to the Bind() function */
-enum BindFlags {
-    BF_NONE         = 0,
-    BF_EXPLICIT     = (1U << 0),
-    BF_REPORT_ERROR = (1U << 1),
-    BF_WHITELIST    = (1U << 2),
-};
-
 static const char* FEE_ESTIMATES_FILENAME="fee_estimates.dat";
 
 //////////////////////////////////////////////////////////////////////////////
@@ -296,17 +288,6 @@ static void registerSignalHandler(int signal, void(*handler)(int))
 }
 #endif
 
-bool static Bind(CConnman& connman, const CService &addr, unsigned int flags) {
-    if (!(flags & BF_EXPLICIT) && IsLimited(addr))
-        return false;
-    std::string strError;
-    if (!connman.BindListenPort(addr, strError, (flags & BF_WHITELIST) != 0)) {
-        if (flags & BF_REPORT_ERROR)
-            return InitError(strError);
-        return false;
-    }
-    return true;
-}
 void OnRPCStarted()
 {
     uiInterface.NotifyBlockTip.connect(&RPCNotifyBlockChange);
@@ -900,10 +881,16 @@ bool AppInitParameterInteraction()
             return InitError(_("Prune mode is incompatible with -txindex."));
     }
 
+    // -bind and -whitebind can't be set when not listening
+    size_t nUserBind =
+        (gArgs.IsArgSet("-bind") ? gArgs.GetArgs("-bind").size() : 0) +
+        (gArgs.IsArgSet("-whitebind") ? gArgs.GetArgs("-whitebind").size() : 0);
+    if (nUserBind != 0 && !gArgs.GetBoolArg("-listen", DEFAULT_LISTEN)) {
+        return InitError("Cannot set -bind or -whitebind together with -listen=0");
+    }
+
     // Make sure enough file descriptors are available
-    int nBind = std::max(
-                (gArgs.IsArgSet("-bind") ? gArgs.GetArgs("-bind").size() : 0) +
-                (gArgs.IsArgSet("-whitebind") ? gArgs.GetArgs("-whitebind").size() : 0), size_t(1));
+    int nBind = std::max(nUserBind, size_t(1));
     nUserMaxConnections = GetArg("-maxconnections", DEFAULT_MAX_PEER_CONNECTIONS);
     nMaxConnections = std::max(nUserMaxConnections, 0);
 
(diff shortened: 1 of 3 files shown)


benma:
Pinging @theuni

src/net.cpp
```diff
@@ -64,6 +64,14 @@
 #endif
 #endif
 
+/** Used to pass flags to the Bind() function */
+enum BindFlags {
```
practicalswift: Minor nit, but what about switching to C++11 strongly typed enumerations (`enum class`) while we are at it?

benma: Great idea,  but unfortunately, bitwise operations don't work with those (unless you manually add those functions to the class) :(

See http://blog.bitwigglers.org/using-enum-classes-as-type-safe-bitmasks/

practicalswift: Ah, of course. Sorry!

ryanofsky (commented):
utACK 7313bfec5354f44571929249ee3672b8c38675a3. This seems like an improvement, though I also think it might be a little better if the connman options were vectors of strings, so AppInit would only be dealing with translating arguments (and the loops could be eliminated), and connman would deal with everything network related (lookups).

benma:
@ryanofsky thanks. I actually did it like this in a first iteration, but changed it because I figured that the arguments should evaluated as much as possible before entering ConnMan.

It was a bit weird having the error messages that relate to `-whitelist`, `-bind`, `-whitebind` still in ConnMan as opposed to where the arguments are actually parsed.

theuni:
I believe this is OK. I have a PR coming up which (after lots of thought) moves resolving into its own class, of which CConnman is one user. This way other layers are able to use the resolver without necessarily having to go through CConnman.

So, concept ACK. Reviewing.

src/init.cpp
```diff
@@ -1644,6 +1594,28 @@ bool AppInitMain(boost::thread_group& threadGroup, CScheduler& scheduler)
     connOptions.nMaxOutboundTimeframe = nMaxOutboundTimeframe;
     connOptions.nMaxOutboundLimit = nMaxOutboundLimit;
 
+    if (gArgs.IsArgSet("-bind")) {
```
theuni: This looks like a change in behavior when setting "-bind=foo -listen=0". Granted, that combination makes no sense.

I think we should probably just check for that combination, and exit with an error in that case.

src/net.cpp
```diff
@@ -2183,7 +2191,41 @@ NodeId CConnman::GetNewNodeId()
...
+        if ((flags & BF_REPORT_ERROR) && clientInterface) {
+            clientInterface->ThreadSafeMessageBox(strError, "", CClientUIInterface::MSG_ERROR);
+        }
+        return false;
+    }
+    return true;
+}
+
+bool CConnman::InitBinds(const std::vector<CService>& binds, const std::vector<CService>& whiteBinds) {
+    if (fListen) {
```
theuni: See other comment. It doesn't make sense to have binds/whitebinds set at all if !fListen.

src/net.cpp
```diff
@@ -2190,7 +2191,41 @@ NodeId CConnman::GetNewNodeId()
...
+        bool fBound = false;
+        for (const auto& addrBind : binds) {
```
theuni: Mm, this moves the logic away from the caller and into CConnman, which is what we're trying to avoid by adding the heaps of options.

I think I'd prefer to see flags passed in along with CServices, so that CConnman can just do what it's told. So either a tiny struct, or just a pair.

Then, CConnman will try to bind the BF_EXPLICIT addresses, and if it fails, tries to bind the rest.

Also, in that case, the last failed bind wouldn't need to have BF_REPORT_ERROR set, we'd just notice the false return and act on it accordingly.

Sound reasonable?

theuni (changes requested):
Concept ACK. Thanks for working on this!

src/net.cpp
benma: > Mm, this moves the logic away from the caller and into CConnman, which is what we're trying to avoid by adding the heaps of options.
> I think I'd prefer to see flags passed in along with CServices, so that CConnman can just do what it's told. So either a tiny struct, or just a pair.

The logic would be in `CConnman` all the same, just the data structure you pass the info with is changed from two vectors to one vector of pairs, which is equivalent. The code looks more complicated from what I can tell. Or did I misunderstand your request?

I added this in a separate commit to see the diff, I drop it again if needed.

> Then, CConnman will try to bind the BF_EXPLICIT addresses, and if it fails, tries to bind the rest.

At the moment, the rest is only attempted if no explicit binds are specified. If they fail, all fails, which makes sense to me.

> Also, in that case, the last failed bind wouldn't need to have BF_REPORT_ERROR set, we'd just notice the false return and act on it accordingly.

Act how, apart from reporting the error (which this flag does)?

src/init.cpp

src/net.cpp
benma: Done.

sipa:
@theuni Can you check whether your requested changes are addressed?

src/net.cpp
theuni: >>Mm, this moves the logic away from the caller and into CConnman, which is what we're trying to avoid by adding the heaps of options.
    I think I'd prefer to see flags passed in along with CServices, so that CConnman can just do what it's told. So either a tiny struct, or just a pair.

> The logic would be in CConnman all the same, just the data structure you pass the info with is changed from two vectors to one vector of pairs, which is equivalent. The code looks more complicated from what I can tell. Or did I misunderstand your request?

The logic would still be in CConnman, but the caller would have the ability to dictate exactly how the binds are setup. For example, there's no way to specify "listen on this address as well as the localhost fallbacks".

I think this is ok for now, though. We can add options as things come up.

src/net.cpp
```diff
@@ -2230,7 +2231,38 @@ NodeId CConnman::GetNewNodeId()
...
 }
 
-bool CConnman::Start(CScheduler& scheduler, std::string& strNodeError, Options connOptions)
+
+bool CConnman::Bind(const CService &addr, unsigned int flags) {
+    if (!(flags & BF_EXPLICIT) && IsLimited(addr))
+        return false;
+    std::string strError;
+    if (!BindListenPort(addr, strError, (flags & BF_WHITELIST) != 0)) {
+        if ((flags & BF_REPORT_ERROR) && clientInterface) {
```
theuni: Well done! Thanks for using the local instance.

src/net.cpp
```diff
@@ -2230,7 +2231,38 @@ NodeId CConnman::GetNewNodeId()
...
+
+bool CConnman::InitBinds(const std::vector<std::pair<CService, Options::BindType> >& binds) {
+    bool fBound = false;
+    for (const auto& addrBind : binds) {
+        unsigned int flags = BF_EXPLICIT | BF_REPORT_ERROR;
+        if (addrBind.second == Options::BindType::WhiteBind)
+            flags |= BF_WHITELIST;
+        fBound |= Bind(addrBind.first, flags);
+    }
+    if (binds.empty()) {
```
theuni: Looks like there's no way to run without listening now.

Sorry for not being more clear here: https://github.com/bitcoin/bitcoin/pull/10496#discussion_r120199183. I was suggesting that we shouldn't get into InitBinds in the first place if binding isn't allowed.

We'll either need to change this so that its caller uses the ugly global as before:
``` c++
    if(fListen && !InitBinds(...)) {
        ...
    }
```
or add a flag for fallback connections, which InitBinds() only uses if nothing else is already bound.

src/net.cpp
```diff
@@ -2252,11 +2284,19 @@ bool CConnman::Start(CScheduler& scheduler, std::string& strNodeError, Options c
 
     SetBestHeight(connOptions.nBestHeight);
 
+    clientInterface = connOptions.uiInterface;
+
+    if (!InitBinds(connOptions.vBinds)) {
+        strError = _("Failed to listen on any port. Use -listen=0 if you want this.");
```
theuni: This strError param is clunky (my fault!). We should do away with it and just use clientInterface->InitError() directly. That can be done as a follow-up though.

src/init.cpp
```diff
@@ -898,10 +879,16 @@ bool AppInitParameterInteraction()
             return InitError(_("Prune mode is incompatible with -txindex."));
     }
 
+    // -bind and -whitebind can't be set when not listening
+    size_t nUserBind =
+        (gArgs.IsArgSet("-bind") ? gArgs.GetArgs("-bind").size() : 0) +
+        (gArgs.IsArgSet("-whitebind") ? gArgs.GetArgs("-whitebind").size() : 0);
+    if (nUserBind != 0 && !GetBoolArg("-listen", DEFAULT_LISTEN)) {
```
theuni: s/GetBoolArg/gArgs.GetBoolArg/

theuni:
looks good aside from those few issues.

src/net.cpp
benma: Oops, was a bit too eager here. Re-added the `fListen` check as you suggested.

src/net.cpp
benma: Done.

src/init.cpp
benma: GetBoolArg() is simply wrapping gArgs.GetBoolArg(), and both variants are already used in the function. Could make that consistent in a separate PR.

benma:
@theuni thanks for the feedback. I added a new commit for easier review, I will squash in the end.

src/net.cpp
benma: > For example, there's no way to specify "listen on this address as well as the localhost fallbacks".

But that wasn't possible before either.

Do you prefer the commit which passes the whitelist flag in the pair (https://github.com/bitcoin/bitcoin/pull/10496/commits/768c7f72d9fac542a45ad788713faa1a3f46897c), or should I drop it? I don't mind either way, but the new code seems less readable than the code before, but it is semantically exactly the same.

src/init.cpp
theuni: gArgs is the new (as of very recently) way of handling these args, it'd be helpful if new code didn't reintroduce the former usage as I assume the wrappers are intended to be temporary.

src/net.cpp
theuni: > But that wasn't possible before either.

Sure, but we've been slowly moving these config options out to the caller so that less is hard-coded.

In the near-future, I'm looking forward to creating 2 CConnman instances and running them against each-other for testing. The more we can control via setup options, the better. I have no problem with adding those options incrementally, I'd just like to make sure we're not making future changes tougher.

As for the whitelist flag, either way is fine for this PR. Whichever you think is more readable. We'll expand on the options when needed.

benma: :+1: I reverted it as I find the previous version more readable.

benma:
@theuni addressed the GetBoolArgs issue and squashed.

benma:
Rebased.

theuni (approved):

theuni:
utACK 07b2afef10bb6366a270e325fd41a8bc526c9ef3

ryanofsky (commented):
utACK 07b2afef10bb6366a270e325fd41a8bc526c9ef3. There were many changes since last review, all discussed above.

laanwj:
utACK 07b2afe

PR #10497, opened 2017-06-01
remove the PAIRTYPE macro

benma:
Replace all BOOST_FOREACH and Q_FOREACH loops using it with range
based loops.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -611,7 +611,7 @@ void CleanupBlockRevFiles()
     // keeping a separate counter.  Once we hit a gap (or if 0 doesn't exist)
     // start removing block files.
     int nContigCounter = 0;
-    BOOST_FOREACH(const PAIRTYPE(std::string, fs::path)& item, mapBlockFiles) {
+    for (const auto& item : mapBlockFiles) {
         if (atoi(item.first) == nContigCounter) {
             nContigCounter++;
             continue;
--- a/src/qt/addresstablemodel.cpp
+++ b/src/qt/addresstablemodel.cpp
@@ -81,7 +81,7 @@ public:
         cachedAddressTable.clear();
         {
             LOCK(wallet->cs_wallet);
-            BOOST_FOREACH(const PAIRTYPE(CTxDestination, CAddressBookData)& item, wallet->mapAddressBook)
+            for (const auto& item : wallet->mapAddressBook)
             {
                 const CBitcoinAddress& address = item.first;
                 bool fMine = IsMine(*wallet, address.Get());
--- a/src/qt/coincontroldialog.cpp
+++ b/src/qt/coincontroldialog.cpp
@@ -626,7 +626,7 @@ void CoinControlDialog::updateView()
     std::map<QString, std::vector<COutput> > mapCoins;
     model->listCoins(mapCoins);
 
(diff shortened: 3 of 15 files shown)


practicalswift:
Concept ACK

dcousens (approved):

fanquake:
This overlaps with #10193. However given it's a smaller change-set, and not solely about nuking Boost, maybe we can merge it first? Depends on how many other nearly/ready-to-merge PRs it's going to break.

benma:
@fanquake thanks for pointing it out, I wasn't aware of the other PR. This PR removes the macro and some Q_FOREACH loops over the other one, so it's a small difference.

@jtimon can decide, I guess constantly rebasing such a big PR is a pain, so I don't mind for his to go first.

ryanofsky (commented):
utACK 2ffa099308e2f3566531e30e5754f83800adbd7d. Nice change. I think it would be good to merge this before #10193 because this change is more targeted and will only make that PR easier to review. I would expect that rebasing #10193 should not be very difficult because most of the commits are scripted.

benma:
@ryanofsky thanks, I addressed all of your feedback.

ryanofsky (commented):
utACK 6b50bb066a55d073f870afb12ab57482da725735

jtimon:
Rebasing #10193 isn't hard at all, once I wrote the sed scripts once it's just running them again. The problem with #10193 is that the reverse iterator to replace BOOST_REVERSE_FOREACH isn't working for some reason. So I separated #10502 with the first 3 commits that could be merged right away.
I would really prefer to do this using scripted-diff to make sure the removal is complete, on the other hand, it isn't adding const in the cases where it can like this one.

MarcoFalke:
Needs rebase

benma:
@MarcoFalke rebased.

practicalswift:
utACK b3489576c28e8e38f1321c1090f7342b75f485c0

Very nice!

sipa:
Can you write this as a scripted-diff?

jtimon:
> Can you write this as a scripted-diff?

Can't we just merge https://github.com/bitcoin/bitcoin/pull/10502 instead?
Did anybody read https://github.com/bitcoin/bitcoin/pull/10497#issuecomment-305564105 ?

sipa:
@jtimon Cool!

benma:
As I said before, I don't mind at all to merge #10502 first. Then I don't have to repeat the effort of scripting the diff and after rebase this will simply be removing the macro.

laanwj:
> Can't we just merge #10502 instead?

So: close this in favor of #10502, as it encompasses more and is scripted?

MarcoFalke:
The only difference is that this one puts `auto` in place of
`std::pair<type1, type2>`.

benma:
I think when this PR was started, the other one didn't deal with the Q_FOREACH and removing the pairtype macro (or I was blind ;)). Now that it does, I'll close this PR.

(I don't mind the explicit std::pair annotation).

jtimon:
https://github.com/bitcoin/bitcoin/pull/10193 didn't remove PAIRTYPE until @tjps suggested it in https://github.com/bitcoin/bitcoin/pull/10193#discussion_r111326802

PR #10500, opened 2017-06-01
Avoid CWalletTx copies in GetAddressBalances and GetAddressGroupings

ryanofsky:


src/wallet/wallet.cpp
```diff
@@ -3261,9 +3261,9 @@ std::map<CTxDestination, CAmount> CWallet::GetAddressBalances()
 
     {
         LOCK(cs_wallet);
-        BOOST_FOREACH(PAIRTYPE(uint256, CWalletTx) walletEntry, mapWallet)
+        for (const auto& walletEntry : mapWallet)
         {
-            CWalletTx *pcoin = &walletEntry.second;
+            const CWalletTx *pcoin = &walletEntry.second;
```
benma: personal preference, just in case you agree: I'd use `const CWalletTX& pcoin = walletEntry.second;` instead.

benma:
utACK b9b814a38ee768ec8a902544722b81c4b9866834

MarcoFalke:
utACK b9b814a38ee768ec8a902544722b81c4b9866834

src/wallet/wallet.cpp
ryanofsky: > personal preference, just in case you agree: I'd use const CWalletTX& pcoin = walletEntry.second; instead.

I'd be fine with this change, but I'm not so motivated to implement it myself, and kind of think it would sidetrack the pr which right now is very simple.

benma: Yeah, no biggie ;)

sipa:
utACK b9b814a38ee768ec8a902544722b81c4b9866834

PR #10511, opened 2017-06-02
[Tests] Include branch coverage info in coverage test

achow101:
This PR  adds an option to configure to include branch coverage statistics be gathered during `make cov`. It also disables logging when coverage is enabled so that the logging branches are not included in the coverage stats. This has a side effect of causing `-Wunused-variable` warnings during compile time when coverage is enabled.

### DIFF
--- a/contrib/filter-lcov.py
+++ b/contrib/filter-lcov.py
@@ -1,0 +1,27 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017 Bitcoin Core Developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+import argparse
+
+parser = argparse.ArgumentParser(description='Remove the coverage data from a tracefile for all files matching the pattern.')
+parser.add_argument('pattern', help='the pattern of files to remove')
+parser.add_argument('tracefile', help='the tracefile to remove the coverage data from')
+parser.add_argument('outfile', help='filename for the output to be written to')
+
+args = parser.parse_args()
+tracefile = args.tracefile
+pattern = args.pattern
+outfile = args.outfile
+
+in_remove = False
+with open(tracefile, 'r') as f:
+    with open(outfile, 'w') as wf:
+        for line in f:
+            if line.startswith("SF:") and pattern in line:
+                in_remove = True
+            if not in_remove:
+                wf.write(line)
+            if line == 'end_of_record\n':
+                in_remove = False
--- a/Makefile.am
+++ b/Makefile.am
@@ -171,13 +171,15 @@ baseline.info:
 	$(LCOV) -c -i -d $(abs_builddir)/src -o $@
 
 baseline_filtered.info: baseline.info
-	$(LCOV) -r $< "/usr/include/*" $(LCOV_OPTS) -o $@
+	$(abs_builddir)/contrib/filter-lcov.py "/usr/include/" $< $@
+	$(LCOV) -a $@ $(LCOV_OPTS) -o $@
 
 leveldb_baseline.info: baseline_filtered.info
 	$(LCOV) -c -i -d $(abs_builddir)/src/leveldb -b $(abs_builddir)/src/leveldb -o $@
 
 leveldb_baseline_filtered.info: leveldb_baseline.info
-	$(LCOV) -r $< "/usr/include/*" $(LCOV_OPTS) -o $@
+	$(abs_builddir)/contrib/filter-lcov.py "/usr/include/" $< $@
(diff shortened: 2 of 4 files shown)


sipa:
Here is a possible hack to force a variable list of arguments to be treated as 'used' for the purpose of `-Wunused-variable`:

```c++
static inline void MarkUsed() {}
template<typename T, typename... Args> static inline void MarkUsed(const T& t, const Args&... args)
{
    (void)t;
    MarkUsed(args...);
}

#ifdef USE_COVERAGE
#define LogPrintf(...) do { MarkUsed(__VA_ARGS__); } while(0)
#define LogPrint(category, ...) do { MarkUsed(__VA_ARGS__); } while(0)
#else
```

achow101:
Used @sipa's hack to shut up the warnings.

gmaxwell:
Cool. Concept ACK.

contrib/filter-lcov.py
```diff
@@ -0,0 +1,24 @@
...
+args = parser.parse_args()
+tracefile = args.tracefile
+pattern = args.pattern
+outfile = args.outfile
+
+in_remove = False
+with open(tracefile, 'r') as f:
+    with open(outfile, 'w') as wf:
+        for line in f:
+            if pattern in line:
```
sipa: Perhaps instead use `line.startswith("SF:" + pattern)` instead? That would prevent matching on variables that accidentally match the pattern name.

contrib/filter-lcov.py
```diff
@@ -0,0 +1,24 @@
...
+                in_remove = True
+            if not in_remove:
+                wf.write(line)
+            if 'end_of_record' in line:
```
sipa: Perhaps use `line == 'end_of_record'` instead? Same reason as above.

contrib/filter-lcov.py
achow101: Done

contrib/filter-lcov.py
achow101: done

sipa:
utACK 405b86a92aee4f2ddb6710bfe07ff714f2afcfa2

contrib/filter-lcov.py
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env python3
```
MarcoFalke: explicit license is missing:
```py
# Copyright (c) 2017 Bitcoin Core Developers
# Distributed under the MIT software license, see the accompanying
# file COPYING or http://www.opensource.org/licenses/mit-license.php.

File: contrib/filter-lcov.py [hunk 4]
```diff
@@ -0,0 +1,24 @@
...
+tracefile = args.tracefile
+pattern = args.pattern
+outfile = args.outfile
+
+in_remove = False
+with open(tracefile, 'r') as f:
+    with open(outfile, 'w') as wf:
+        for line in f:
+            if line.startswith("SF:") and pattern in line:
+                in_remove = True
```
MarcoFalke: Not sure why you need the var `in_remove`. Couldn't you write `wf.write(line);continue;` here?

maflcko (commented):
Concept ACK

contrib/filter-lcov.py
achow101: The format of the file is

```
SF:<path to file>
<bunch of lines>
end_of_record
<repeat>
```
In order to know that we are in a `<bunch of lines>` section that should be removed (because the `SF:...` line matches our removal pattern, we need to have a variable to remember to remember that we shouldn't be writing these lines to the output file. That is what `in_remove` is for.

contrib/filter-lcov.py
achow101: done

theuni:
The lcov stuff was added _ages_ ago, when lcov's --no-external was too new to rely on. I think it'd be ok now to require >=1.10 and forego all of the filtering.

MarcoFalke:
@achow101 Are you still working on this?

achow101:
@MarcoFalke Yes

achow101:
Would it be okay with everyone if I just merged #10565 into this PR? It all falls under "improvements to coverage data".

sipa:
@achow101 Have you tried the approach @theuni suggested above (using --no-external, and removing all filtering)?

achow101:
@sipa I am thinking of doing that and combining #10565 into this PR. If I remove filtering here, it would just be added back in by #10565.

MarcoFalke:
Needs rebase.

achow101:
Closing this as #10565 contains this PR (I totally forgot) except for the copyright header I added last commit :/

PR #10512, opened 2017-06-02
Rework same-chain from abusing DoS banning, to explicit checks

luke-jr:
This fixes bugs where we end up banning peers running different consensus rules (including old nodes following a softfork). Since we rely on having peers enforcing the same rules as us, it also enforces that outbound non-feeler connections must have and maintain the same tip block; if they diverge from us, we disconnect them (and find a replacement).

### DIFF
--- a/src/chain.h
+++ b/src/chain.h
@@ -357,6 +357,11 @@ public:
     //! Efficiently find an ancestor of this block.
     CBlockIndex* GetAncestor(int height);
     const CBlockIndex* GetAncestor(int height) const;
+
+    /** (Somewhat) efficiently check whether a block is present in this chain. */
+    const bool Contains(const CBlockIndex * const needle) const {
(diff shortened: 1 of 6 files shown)


gmaxwell:
bleh please don't use reject for this.

luke-jr:
@gmaxwell Do you have a better solution? It doesn't seem like we have any other way to determine that case?

luke-jr:
```
p2p-compactblocks.py failed, Duration: 83 s
```

But it passes locally... :/

luke-jr:
This breaks because compact blocks allows relaying invalid blocks without penalty, but there is no indication at all even later, whether the peer sending you the compact block has accepted it himself or not... so basically we can almost never know if a compact-block peer is on a different tip than us.

Ideas for how to solve this?

luke-jr:
Closed due to impracticality, and superceded by #10593

PR #6962, opened 2015-11-06
translations: Don't translate markup or force English grammar

maflcko:

PR #10513, opened 2017-06-02
Trivial: grammar fix to CONTRIBUTING.md

ABISprotocol:


paveljanik:
If `"pull requests."` is good grammar, the grammar is bad.

NACK

fanquake:
Agree with @paveljanik, neither change seems to be fixing any grammar.

ABISprotocol:
Although this is marked Trivial, I request that you reopen this [because of the following reasons](https://www.quora.com/Should-the-quotation-mark-be-placed-before-or-after-a-period-comma).

Alternatively, please retract statements you have made that "grammar is bad" - @paveljanik - and that "neither change seems to be fixing any grammar" - @fanquake - and then propose a new issue dealing with the grammar issues which come up when dealing with the different ways that different countries handle periods (as in some countries the period concluding the sentence, for example, is placed inside the quotation marks, and in others place it is placed outside the quotation marks).

However, it would be improper to close this issue due to your claims that "the grammar is bad" or that "neither change seems to be fixing any grammar."


This merged pull request -- https://github.com/bitcoin/bitcoin/pull/6962/commits/6dd3a44ce2cb4480054ae0ca0960f588a9ad8d25 -- was provided so as to not translate markdown or force English grammar, but it certainly did not clarify what to do about differences in English grammar with respect to how people in different countries handle commas and periods with respect to quotation marks (in determining whether to place them inside or outside quotation marks).

Suggested solution:  

Reopen this issue, or open a new issue to discuss further the inconsistency identified above.

@laanwj

paveljanik:
I can't retract my opinion. I'm non English native speaker, do not care about en_US and en_UK differences and do not care about your side in that "fight".

ABISprotocol:
@paveljanik It's an interesting position, not to care about someone else's thoughts or views.  It's not a fight, but it is a legitimate question.  As said, trivial, but look at it like this.  What logic is there in imposing UK grammar on everybody who is using English?  

It strikes me that (again, as pointed out) that although this is trivial, it is worth a bit of discussion.

laanwj:
Please don't tag me in this, I have no duck in this fight. Life is too short to have long arguments about minor language issues IMO. I try to focus on changes affecting performance, user experience, stability and testing.

ABISprotocol:
Not intended to be a fight, as mentioned, but if you are focusing on changes affecting user experience, this would be one of them.

PR #10515, opened 2017-06-03
[test] Add test for getchaintxstats

jimmysong:
Simple test for getchaintxstats. The default is a month and that's not tested as there's 200 blocks and not 2016.


sipa:
utACK cc36b5e3d7ae9eebcc0768b6707e8928e3cd69aa

paveljanik:
ACK cc36b5e

MarcoFalke:
utACK cc36b5e

PR #10517, opened 2017-06-03
Factor out CCoinsView based AreInputsStandard/IsWitnessStandard

jonasschnelli:
It's move-onlyish.
This will allow IsStandard checks without depending on the mempool, UTXO set.

This may be useful for libbitcoinconsensus and is certainly required for compiling a wallet-tool (next to some other things that are required).


src/policy/inputs.h
```diff
@@ -0,0 +1,26 @@
...
+
+#ifndef BITCOIN_POLICY_INPUTS_H
+#define BITCOIN_POLICY_INPUTS_H
+
+#include <string>
+
+class CCoinsViewCache;
+class CTransaction;
+
+    /**
```
benma: I know it's just a move, but maybe you could fix the indentation here and below?

src/policy/inputs.cpp
```diff
@@ -0,0 +1,101 @@
+// Copyright (c) 2017 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+// NOTE: This file is intended to be customised by the end user, and includes only local node policy logic
+
+#include "policy/policy.h"
```
benma: I think it would be beneficial to complete the split, i.e.:

- include `script/standard.h` instead
- move `MAX_P2SH_SIGOPS`, `MAX_STANDARD_P2WSH_STACK_ITEMS`, `MAX_STANDARD_P2WSH_STACK_ITEM_SIZE` and `MAX_STANDARD_P2WSH_SCRIPT_SIZE` from `policy.h` to `inputs.h`.

benma (changes requested):

jonasschnelli:
Followed @benma's advice (force push amend commit):
* moved the `input.cpp` relevant constants to from `policy.h` to `input.h`
* fixed indentation

I kept the name `policy.*`. Renaming it to `standard.*` makes little sense as long as we have tx size in there as well. Can split later.

src/Makefile.am
```diff
@@ -314,14 +317,21 @@ libbitcoin_common_a_SOURCES = \
...
   scheduler.cpp \
   script/sign.cpp \
   script/standard.cpp \
   warnings.cpp \
   $(BITCOIN_CORE_H)
 
+# policy: shared between all executables that require policy rules.
+libbitcoin_policy_a_CPPFLAGS = $(AM_CPPFLAGS) $(BITCOIN_INCLUDES)
+libbitcoin_policy_a_CXXFLAGS = $(AM_CXXFLAGS) $(PIE_FLAGS)
+libbitcoin_policy_a_SOURCES = \
```
benma: This is a bit confusing to me.

The sources list doesn't contain all of policy/*.cpp (`fees.cpp` and `inputs.cpp` could be in there?), and the ones listed below are not used together anywhere (i.e. `feerate.cpp` is used in `libbitcoin_common` which doesn't need `policy.cpp`, and `libbitcoin_server` relies on `policy.cpp` but not `feerate.cpp`).

Why is this shared library needed, and why this particular combination?

src/policy/inputs.cpp
benma: @jonasschnelli You can include `script/standard.h` here instead of `policy/policy.h` to cut all ties :)

Mabye move the new include further down next to `script/interpreter.h`.

benma (changes requested):
(see individual comments)

src/Makefile.am
jonasschnelli: It's an in-between step and allows to grow the independent policy code/module. `libbitcoin_policy_a` does only contain the complete independent policy classes. But we could also completely avoid the new "package" (aka library). Thoughts @theuni?

ryanofsky (commented):
utACK 2fc36f0d929b6fbcefcae8d986efe38a4adbeec3

I don't know all the reasoning behind libconsensus code organization, but I can see how this would be useful for a wallet tool, and confirmed the change is move only.

jonasschnelli:
Closing for now... I'll maybe pickup later.

PR #8680, opened 2016-09-07
Address Travis spurious failures

theuni:
As discussed in last week's meeting. The failures are annoying.

I'm sure that this isn't the best approach, but these rpcs seem generally useful, and I believe this fixes the current race issue.

The problem is that the wallet is not yet finished dealing with transactions before getblockcount/getbestblockhash return new values, so assertions about balances are racy. The approach I've taken here is to wait for a signal that the processing is complete, then cache the new height/hash on the rpc side. That allows for blocking in the call rather than polling, and it ensures that we're synced up before it returns. It also means no cs_main locking.

To fix sync_blocks in particular, wait for height 0 from all nodes. That returns the current height/hash instantly, since it should never actually have to wait. Then, grab the current height from all nodes, take the max, and wait on that height for all nodes. Repeat until all heights match.

@sdaftuar Mentioned that it may be useful to a wait_for_wallet() call, which would presumably work similarly, though I'm afraid it may introduce new race issues. Suggestions on that approach are also welcome.

PR #10519, opened 2017-06-03
[RPC] Remove waitforblock and waitfornewblock

achow101:
Removed because these RPCs are for tests only and are now unused

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -202,86 +202,6 @@ void RPCNotifyBlockChange(bool ibd, const CBlockIndex * pindex)
     cond_blockchange.notify_all();
 }
 
-UniValue waitfornewblock(const JSONRPCRequest& request)
-{
-    if (request.fHelp || request.params.size() > 1)
-        throw std::runtime_error(
-            "waitfornewblock (timeout)\n"
-            "\nWaits for a specific new block and returns useful info about it.\n"
-            "\nReturns the current block on timeout or exit.\n"
-            "\nArguments:\n"
-            "1. timeout (int, optional, default=0) Time in milliseconds to wait for a response. 0 indicates no timeout.\n"
-            "\nResult:\n"
-            "{                           (json object)\n"
-            "  \"hash\" : {       (string) The blockhash\n"
-            "  \"height\" : {     (int) Block height\n"
-            "}\n"
-            "\nExamples:\n"
-            + HelpExampleCli("waitfornewblock", "1000")
-            + HelpExampleRpc("waitfornewblock", "1000")
-        );
-    int timeout = 0;
-    if (!request.params[0].isNull())
-        timeout = request.params[0].get_int();
-
-    CUpdatedBlock block;
-    {
-        std::unique_lock<std::mutex> lock(cs_blockchange);
-        block = latestblock;
-        if(timeout)
-            cond_blockchange.wait_for(lock, std::chrono::milliseconds(timeout), [&block]{return latestblock.height != block.height || latestblock.hash != block.hash || !IsRPCRunning(); });
-        else
-            cond_blockchange.wait(lock, [&block]{return latestblock.height != block.height || latestblock.hash != block.hash || !IsRPCRunning(); });
-        block = latestblock;
-    }
-    UniValue ret(UniValue::VOBJ);
-    ret.pushKV("hash", block.hash.GetHex());
-    ret.pushKV("height", block.height);
-    return ret;
-}
-
-UniValue waitforblock(const JSONRPCRequest& request)
-{
-    if (request.fHelp || request.params.size() < 1 || request.params.size() > 2)
-        throw std::runtime_error(
-            "waitforblock <blockhash> (timeout)\n"
-            "\nWaits for a specific new block and returns useful info about it.\n"
-            "\nReturns the current block on timeout or exit.\n"
-            "\nArguments:\n"
-            "1. \"blockhash\" (required, string) Block hash to wait for.\n"
-            "2. timeout       (int, optional, default=0) Time in milliseconds to wait for a response. 0 indicates no timeout.\n"
-            "\nResult:\n"
-            "{                           (json object)\n"
-            "  \"hash\" : {       (string) The blockhash\n"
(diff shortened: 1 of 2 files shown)


sipa:
@theuni These RPCs were added in #8680, and never used (only `waitforblockatheight` was ever used). Is there a plan or expectation that they'll ever be needed?

NicolasDorier:
I did not know it existed, but I think it is very useful, I would use it if I am sure it stays.

Easier to use than configuring blocknotify or ZMQ.

laanwj:
Would be nice to have a test for them, if we keep them.

theuni:
@sipa If ever used, I figured it'd only be to make some python tests less cumbersome.

@jnewbery Do you have any use for these? Otherwise, I'd be fine with removing.

jnewbery:
> @jnewbery Do you have any use for these? Otherwise, I'd be fine with removing.

Not currently, although they may be useful for some tests. On the other hand, it's just as easy to call `getbestblockhash` or `getblock` in a loop until the desired block has been accepted.

I have no issue with these being removed or staying (as long as they remain hidden and not part of the public API).

MarcoFalke:
I am ok with removing or keeping and adding tests. Though, it seems the functionality is redundant:

* waitfornewblock() ---> waitforblockheight(current_height + 1)
* waitforblock(hash) ---> waitforblockheight(hash_height) && getblockhash(hash_height) == hash

MarcoFalke:
utACK 5fc1253bb8b95d34880e565cb2aa193caab06d71

theuni:
@MarcoFalke They're intended to be atomic operations as we were dealing with some test race issues a while back. They're indeed redundant if atomicity isn't needed.

NicolasDorier:
well, I think they are useful. Instead of polling for new block I would definitively use that.

sipa:
utACK b69194916c5fc8ebdaebb04e3cc6c4aab37e1c49

achow101:
rebased

promag:
utACK b691949.

IMO blocking or long pool RPC/REST calls should be avoided.

NicolasDorier:
I think this should not be removed, but documented. Right now the only way to get block notification is to connect to to the P2P layer or use ZeroMQ. Both requires additional configuration. (also, I have never managed to make ZeroMQ notifications works with a .NET Client)
I end up connecting to my node by P2P to get new blocks.

I would have known that such RPC method existed, I would have definitively used it in place of my current solution.

I think a `WaitWalletTransaction` would also be higly helpful.

luke-jr:
@NicolasDorier FYI there is also `blocknotify`

NicolasDorier:
Same problem as ZeroMQ. It requires additional configuration on the bitcoin node, and need additional code, the program called by `blocknotify` need a way to notify the program interested into this event by another mechanism.... This is a pain.

jnewbery:
@NicolasDorier - Be careful about using these RPCs in product code. They were introduced as test tools only and there is no guarantee that they will work as you expect. For example, I just managed to break bitcoind's RPC server by running 4 `waitfornewblock` requests in parallel. Any request after that fails, and I don't think there's any way to cancel the `waitfornewblock` requests.

There are also no tests for them and they are hidden debug RPCs, so it's likely you'd be the only one using them, which increases the risk of regressions.

I'm slightly averse to long-polling RPCs in general, but I'm also not too concerned if these stay (with the provisos above).

promag:
Agree with @jnewbery. You can have a small utility to listen zmq messages and forward the most suitable transport for your .NET app.

And I think doing that to some kind of queue would benefit both parts, specially if the queue has means to easily notify you app.

NicolasDorier:
@jnewbery I know there is a maximum number of concurrent requests, this can be notified to the user when it happens.

@promag I enumerated the problems of using ZMQ above. This just make it harder to build stuff on top of Core by making it require more configuration and dependency.
It is easier to connect to the P2P layer and listen for new block than it is to setup ZMQ, and listening through it with a cross plateform library.

promag:
> making it require more configuration

@NicolasDorier IMO that´s fine. Even if there was websocket support, it should have/require configuration before it could be used. You also configure RPC right?

NicolasDorier:
@promag no I do not configure RPC. If bitcoin is running with default parameters, I can connect to it through the wellknown port and wellknown cookiefile.

achow101:
Rebased

NicolasDorier:
If https://github.com/bitcoin/bitcoin/pull/10519 would be merged, I would see no problem removing this rpc method.

laanwj:
Needs rebase, if still relevant.

NicolasDorier:
Seems like nobody is using it, and on my side I am not at ease using something that might disappear tomorrow, so feel free to delete it.

But I still think it is a shame, this is the most useful RPC call ever! :(

achow101:
Rebased

nopara73:
NACK.

`waitfor...` RPC calls are very useful. They should be documented rather than removed. If they ruin some coding best practices, then remove it.

Imagine a wallet daemon that gives you p2pkh only through RPC, it gives you p2wpkh only through ZMQ and it gives you p2wpkh over p2sh only through websockets. Sure that works, but you must have 3 different communication channels properly implemented and configured in order to get all three of them. **The more communication protocols a software uses the less reliable it is.** Networking may be the trickiest coding challenge, where so many little things can go wrong.

https://github.com/MetacoSA/NBitcoin/pull/363
https://github.com/nopara73/MagicalCryptoWallet/issues/32

laanwj:
As both @NicolasDorier and @nopara73 are apparently using these calls (or would if they aren't goin to be removed) and think they are useful, I don't think they should be removed, then.
Closing.

nopara73:
Personally we were using but removed them, due to this change, so it's fine by me.

jnewbery:
> Although I'd still not be happy if the user space breaks (like ever.)

These RPC methods were introduced as hidden calls for testing only. They shouldn't be used in production systems. They should either be removed (since they're unused in testing now), or made unhidden and fully supported. I'd vote for removing them.

MarcoFalke:
Marked up for grabs for someone to address my feedback: https://github.com/bitcoin/bitcoin/pull/10519#discussion_r173045365

PR #10521, opened 2017-06-03
Limit variable scope

practicalswift:
Limit variable scope as much as possible in order to:
* improve code readability,
* reduce complexity, and/or
* avoid unnecessary initialisations.

### DIFF
--- a/src/policy/fees.cpp
+++ b/src/policy/fees.cpp
@@ -855,13 +855,13 @@ bool CBlockPolicyEstimator::Read(CAutoFile& filein)
     try {
         LOCK(cs_feeEstimator);
         int nVersionRequired, nVersionThatWrote;
-        unsigned int nFileBestSeenHeight, nFileHistoricalFirst, nFileHistoricalBest;
         filein >> nVersionRequired >> nVersionThatWrote;
         if (nVersionRequired > CLIENT_VERSION)
             return error("CBlockPolicyEstimator::Read(): up-version (%d) fee estimate file", nVersionRequired);
 
         // Read fee estimates file into temporary variables so existing data
         // structures aren't corrupted if there is an exception.
+        unsigned int nFileBestSeenHeight;
         filein >> nFileBestSeenHeight;
 
         if (nVersionThatWrote < 149900) {
@@ -890,6 +890,7 @@ bool CBlockPolicyEstimator::Read(CAutoFile& filein)
             }
         }
         else { // nVersionThatWrote >= 149900
+            unsigned int nFileHistoricalFirst, nFileHistoricalBest;
             filein >> nFileHistoricalFirst >> nFileHistoricalBest;
             if (nFileHistoricalFirst > nFileHistoricalBest || nFileHistoricalBest > nFileBestSeenHeight) {
                 throw std::runtime_error("Corrupt estimates file. Historical block range for estimates is invalid");
--- a/src/qt/trafficgraphwidget.cpp
+++ b/src/qt/trafficgraphwidget.cpp
@@ -47,13 +47,14 @@ int TrafficGraphWidget::getGraphRangeMins() const
 
 void TrafficGraphWidget::paintPath(QPainterPath &path, QQueue<float> &samples)
 {
-    int h = height() - YMARGIN * 2, w = width() - XMARGIN * 2;
-    int sampleCount = samples.size(), x = XMARGIN + w, y;
+    int sampleCount = samples.size();
     if(sampleCount > 0) {
+        int h = height() - YMARGIN * 2, w = width() - XMARGIN * 2;
+        int x = XMARGIN + w;
         path.moveTo(x, YMARGIN + h);
         for(int i = 0; i < sampleCount; ++i) {
             x = XMARGIN + w - w * i / DESIRED_SAMPLES;
-            y = YMARGIN + h - (int)(h * samples.at(i) / fMax);
+            int y = YMARGIN + h - (int)(h * samples.at(i) / fMax);
             path.lineTo(x, y);
         }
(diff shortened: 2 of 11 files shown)


sipa:
utACK 5e6c94531046135d406ea755851fd614a76fb543

src/policy/fees.cpp
```diff
@@ -855,7 +855,7 @@ bool CBlockPolicyEstimator::Read(CAutoFile& filein)
     try {
         LOCK(cs_feeEstimator);
         int nVersionRequired, nVersionThatWrote;
-        unsigned int nFileBestSeenHeight, nFileHistoricalFirst, nFileHistoricalBest;
+        unsigned int nFileBestSeenHeight;
```
benma: move down to the before L865, to move it closer to its first use.

src/qt/trafficgraphwidget.cpp
```diff
@@ -48,12 +48,12 @@ int TrafficGraphWidget::getGraphRangeMins() const
 void TrafficGraphWidget::paintPath(QPainterPath &path, QQueue<float> &samples)
 {
     int h = height() - YMARGIN * 2, w = width() - XMARGIN * 2;
```
benma: `h`, `w`, `x` could be moved into the if clause

src/rpc/mining.cpp
```diff
@@ -98,13 +98,12 @@ UniValue getnetworkhashps(const JSONRPCRequest& request)
...
 {
     static const int nInnerLoopCount = 0x10000;
-    int nHeightStart = 0;
     int nHeightEnd = 0;
     int nHeight = 0;
 
     {   // Don't keep cs_main locked
         LOCK(cs_main);
-        nHeightStart = chainActive.Height();
+        int nHeightStart = chainActive.Height();
```
benma: is this var even needed? could do `nHeight = chainActive.Height();` and use `nHeight` instead of `nHeightStart` on the line after.

src/script/interpreter.cpp
```diff
@@ -263,11 +263,11 @@ bool EvalScript(std::vector<std::vector<unsigned char> >& stack, const CScript&
     set_error(serror, SCRIPT_ERR_UNKNOWN_ERROR);
     if (script.size() > MAX_SCRIPT_SIZE)
         return set_error(serror, SCRIPT_ERR_SCRIPT_SIZE);
-    int nOpCount = 0;
     bool fRequireMinimal = (flags & SCRIPT_VERIFY_MINIMALDATA) != 0;
```
benma: why did you move only `nOpCount` but not this or some of the other vars above?

benma (changes requested):

src/script/interpreter.cpp
practicalswift: Yes, there are more vars that could be moved into the tighter `try` scope. I'll take all `interpreter.cpp` changes in a separate PR since they'll need a more thorough review. I've now removed the `nOpCount` move from this PR.

practicalswift:
@benma Feedback items addressed! Thanks for the review.

src/qt/trafficgraphwidget.cpp
```diff
@@ -47,13 +47,14 @@ int TrafficGraphWidget::getGraphRangeMins() const
 
 void TrafficGraphWidget::paintPath(QPainterPath &path, QQueue<float> &samples)
 {
-    int h = height() - YMARGIN * 2, w = width() - XMARGIN * 2;
-    int sampleCount = samples.size(), x = XMARGIN + w, y;
+    int sampleCount = samples.size();
     if(sampleCount > 0) {
+        int w = width() - XMARGIN * 2;
+        int x = XMARGIN + w, h = height() - YMARGIN * 2;
```
benma: a super-nit, but I preferred `w` and `h` on the same line as before, because of the symmetry.

benma:
utACK 4adc8389af964e84c1f97bc28707904341bbbc74

practicalswift:
@benma Good point. `w` and `h` now on the same line again :-)

src/qt/trafficgraphwidget.cpp
```diff
@@ -47,13 +47,14 @@ int TrafficGraphWidget::getGraphRangeMins() const
 
 void TrafficGraphWidget::paintPath(QPainterPath &path, QQueue<float> &samples)
 {
-    int h = height() - YMARGIN * 2, w = width() - XMARGIN * 2;
-    int sampleCount = samples.size(), x = XMARGIN + w, y;
+    int sampleCount = samples.size();
     if(sampleCount > 0) {
+        int h = height() - YMARGIN * 2, w = width() - XMARGIN * 2;
+        int x = XMARGIN + w;
```
benma: Sorry for another nit, but why not inline this `x` (`YMARGIN + h` is also inlined`)? Then you can declare `int x` on line 56 (or inline those two as well).

Currently, there something like shadowing going on, where `x` is not modified inside the loop, as they are independent.

jtimon:
utACK 90593ed92cfd5d49900a6fb6c2c10a482ab9fdbb

paveljanik:
utACK 90593ed

src/qt/trafficgraphwidget.cpp
practicalswift: @benma I'm not sure I follow. The value of `x` assigned in the looped is used outside of the loop, so declaring it on line 56 wouldn't maintain the same logic as before, would it? Please post a gist with your suggested formulation :-)

benma: You are correct, sir! Apologies :)

practicalswift: @benma Apology accepted! 😄

PR #10524, opened 2017-06-03
[tests] Remove printf(...)

practicalswift:


sipa:
These should just be changed to use `LogPrintf` (or nothing at all, it's debug code!) instead of `printf`.

Historically, `printf` was a macro in the codebase that invoked the logging code. Since then we have 1) switched to tinyformat (where %d is legal for unsigned values) and 2) replaced the `printf` calls with `LogPrintf`. It seems these two are somehow copied from an old version of the code.

practicalswift:
@sipa Thanks for the review!

Turns out that these two were the last remaining instances of `printf`:

```
$ git grep "[^a-z]printf(" -- "*.cpp" "*.h" | egrep -v "(univalue|tinyformat|secp256k1)"
src/test/sighash_tests.cpp:        printf("ERROR: SignatureHash(): nIn=%d out of range\n", nIn);
src/test/sighash_tests.cpp:            printf("ERROR: SignatureHash(): nOut=%d out of range\n", nOut);
```

Looks good now?

laanwj:
The logging output of the tests is ignored, I don't think there's any flag to show/store it. So I think it'd be better to just remove these lines then?

MarcoFalke:
As they indicate an error, the alternative would be to replace them with `BOOST_CHECK_MESSAGE`, which also works with the operator<<

laanwj:
> As they indicate an error

Sure, though I guess if it fails, I'd hope the test will fail later on anyhow because of the return value.

practicalswift:
PR updated. Now simply removing the `printf(…)`:s :-)

sipa:
utACK 0abc58890c1a6600fa5097603febd3990a071d03

PR #10526, opened 2017-06-04
Force on-the-fly compaction during pertxout upgrade

sipa:
It seems that LevelDB tends to leave the old "per txid" UTXO entries in the database lying around for a significant amount of time during and after the per-txout upgrade. This introduces a `CompactRange` function in the database wrapper, and invokes it after every batch of updates in `CCoinsViewDB::Upgrade()`. This lowers temporary disk usage during and after the upgrade.

### DIFF
--- a/src/txdb.cpp
+++ b/src/txdb.cpp
@@ -375,12 +375,13 @@ bool CCoinsViewDB::Upgrade() {
     CDBBatch batch(db);
     uiInterface.SetProgressBreakAction(StartShutdown);
     int reportDone = 0;
+    std::pair<unsigned char, uint256> key;
+    std::pair<unsigned char, uint256> prev_key = {DB_COINS, uint256()};
     while (pcursor->Valid()) {
         boost::this_thread::interruption_point();
         if (ShutdownRequested()) {
             break;
         }
-        std::pair<unsigned char, uint256> key;
         if (pcursor->GetKey(key) && key.first == DB_COINS) {
             if (count++ % 256 == 0) {
                 uint32_t high = 0x100 * *key.second.begin() + *(key.second.begin() + 1);
@@ -409,6 +410,8 @@ bool CCoinsViewDB::Upgrade() {
             if (batch.SizeEstimate() > batch_size) {
                 db.WriteBatch(batch);
                 batch.Clear();
+                db.CompactRange(prev_key, key);
+                prev_key = key;
             }
             pcursor->Next();
         } else {
@@ -416,6 +419,7 @@ bool CCoinsViewDB::Upgrade() {
         }
     }
     db.WriteBatch(batch);
+    db.CompactRange({DB_COINS, uint256()}, key);
     uiInterface.SetProgressBreakAction(std::function<void(void)>());
     LogPrintf("[%s].\n", ShutdownRequested() ? "CANCELLED" : "DONE");
     return true;
--- a/src/dbwrapper.h
+++ b/src/dbwrapper.h
@@ -321,6 +321,23 @@ public:
         pdb->GetApproximateSizes(&range, 1, &size);
         return size;
     }
+
+    /**
+     * Compact a certain range of keys in the database.
+     */
+    template<typename K>
+    void CompactRange(const K& key_begin, const K& key_end) const
+    {
+        CDataStream ssKey1(SER_DISK, CLIENT_VERSION), ssKey2(SER_DISK, CLIENT_VERSION);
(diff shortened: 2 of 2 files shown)


laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/10526/commits/fe84334135afa79845d18e472b71536cca72fd0d

gmaxwell:
Concept ACK

laanwj:
Original chainstate (up to height=430234).
```
2.4G    testbtc.olddb/chainstate
```
After converting using master (and shutting down immediately):
```
2017-06-07 12:54:30 Upgrading database...
2017-06-07 13:09:39 LoadBlockIndexDB: last block file = 628

4.5G    chainstate
```
Converting using this pull (and shutting down immediately):
```
2017-06-07 12:29:37 Upgrading database...
2017-06-07 12:41:50 LoadBlockIndexDB: last block file = 628

4.4G    testbtc.newdb/chainstate
```
In this test, it doesn't seem to make a significant difference.

sipa:
I wonder what to do here. Without any forced compaction, it's likely that many nodes will remain with a database that still contains table files for the old rows for an extended period of time.

TheBlueMatt:
Looks like this needed a 15 tag... If it's not too late let's call it a bugfix, tag it 15, and try to get it in this week.

TheBlueMatt:
Also, needs rebase.

laanwj:
After converting up to block height=430234 (same dataset as last time), hardwiring shutdown immediately after the block database processing.

Tested with rebased version here: https://github.com/laanwj/bitcoin/tree/2017_07_sipa_onthefly
```
2.1G    chainstate/
```
With master:
```
4.3G    chainstate/
```
So it seems the negative result last time was a fluke. utACK.

laanwj:
I repeated a few times, and it seems that it's like a coin toss. Sometimes it ends up as 4.4G, sometimes as 2.1G. 
When the former, *every time* after running bitcoind again, even terminating immediately after chainstate processing, it was at 2.1G after shutting down.
It might be that a cleanup is re-triggered when re-opening the database, or when closing it.

sipa:
Rebased. I've made the final compaction require for the entire per-tx range of keys, rather than just the last group updated. Maybe that triggers compaction sooner. Or not.

src/txdb.cpp
```diff
@@ -375,6 +375,8 @@ bool CCoinsViewDB::Upgrade() {
     CDBBatch batch(db);
     uiInterface.SetProgressBreakAction(StartShutdown);
     int reportDone = 0;
+    std::pair<unsigned char, uint256> key;
```
morcos: Is this meant to be shadowed by the key defined 7 lines down?

morcos:
If someone was running with master before this commit, how does their database ever get compacted?

TheBlueMatt:
@morcos leveldb does automatic background compaction.

morcos:
>@morcos leveldb does automatic background compaction.

when?  I have nodes running for a week that have 5.8G chainstate despite recent repeated stopping and starting.

gmaxwell:
@morcos  Yea, I have one two weeks now.   Perhaps we should also have some general compaction thing to clean up existing databases too....

TheBlueMatt:
I got some strange testing results...current master:
 * Convert time 26:30, final chainstate size 3.5G
 * Convert time 28:05, final chainstate size 3.5G

This PR:
 * Convert time: 16:51, final chainstate size 5.9G (down to 2.7 immediately upon restart)
 * Convert time: 16:11, final chainstate size 5.9G (down to 2.7 immediately upon restart)

Either way, I dont care about the excess usage if its faster and gets cleared upon restart (might get cleared in the background, too, i just made bitcoind do a clean exit right after chain state gets loaded).

TheBlueMatt:
@morcos hmm, I guess its just done when tables fill up, it should /eventually/ happen, but may take a long time, I suppose.

morcos:
@TheBlueMatt wow, that's some really long convert times
My convert times with or without this PR were on the order of between 4 mins and 7 mins with some variation.

TheBlueMatt:
Heh, thats what putting a database on spinning rust on a CoW filesystem
does to you....I've seen it much faster under other conditions.

src/txdb.cpp
sipa: Fixed.

TheBlueMatt:
utACK efeb273305e3e4d2c42e1e153de83c1cb6f0a28c, I tested an earlier version of the patch which should be equivalent, however.

PR #10527, opened 2017-06-04
Use parentheses to clarify intended precedence when using bitwise operations

practicalswift:
Use parentheses to clarify intended precedence when using bitwise operations.

### DIFF
--- a/src/crypto/aes.cpp
+++ b/src/crypto/aes.cpp
@@ -137,7 +137,7 @@ static int CBCDecrypt(const T& dec, const unsigned char iv[AES_BLOCKSIZE], const
         // If used, padding size is the value of the last decrypted byte. For
         // it to be valid, It must be between 1 and AES_BLOCKSIZE.
         padsize = *--out;
-        fail = !padsize | (padsize > AES_BLOCKSIZE);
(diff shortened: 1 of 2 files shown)


benma:
Sometimes bitwise precedences can be tricky, but those two instances you changed are very clear without the parens. I.e. confusing `!a | b` as `!(a | b)` looks impossible to me, even if only due to the spacing.

The second example has only two operands, so you can't even read it wrong.

src/crypto/aes.cpp
```diff
@@ -137,7 +137,7 @@ static int CBCDecrypt(const T& dec, const unsigned char iv[AES_BLOCKSIZE], const
         // If used, padding size is the value of the last decrypted byte. For
         // it to be valid, It must be between 1 and AES_BLOCKSIZE.
         padsize = *--out;
-        fail = !padsize | (padsize > AES_BLOCKSIZE);
+        fail = (!padsize) | (padsize > AES_BLOCKSIZE);
```
MarcoFalke: nack per @benma

practicalswift:
Closing due to NACK consensus :-)

PR #10530, opened 2017-06-05
Fix invalid instantiation and possibly unsafe accesses of array in class base_uint<BITS>

pavlosantoniou:
The implementation of base_uint<BITS>::operator++(int) and base_uint<BITS>::operator--(int) is now safer.
Array pn is accessed via index i after bounds checking has been performed on the index, rather than before.

The logic of the while loops has also been made more clear.

### DIFF
--- a/src/arith_uint256.h
+++ b/src/arith_uint256.h
@@ -31,12 +31,16 @@ public:
 
     base_uint()
     {
+        static_assert(BITS/32 > 0 && BITS%32 == 0, "Template parameter BITS must be a positive multiple of 32.");
+
         for (int i = 0; i < WIDTH; i++)
             pn[i] = 0;
     }
 
     base_uint(const base_uint& b)
     {
+        static_assert(BITS/32 > 0 && BITS%32 == 0, "Template parameter BITS must be a positive multiple of 32.");
+
         for (int i = 0; i < WIDTH; i++)
             pn[i] = b.pn[i];
     }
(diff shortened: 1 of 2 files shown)


src/arith_uint256.h
```diff
@@ -174,7 +174,7 @@ class base_uint
     {
         // prefix operator
         int i = 0;
-        while (++pn[i] == 0 && i < WIDTH-1)
+        while (i < WIDTH && ++pn[i] == 0)
```
laanwj: As I see it, this code can cause an out-of-bounds access only in the case of WIDTH=0, which would be pointless.
Or am I missing something?
I do agree that this change makes the code somewhat more readable.

pavlosantoniou: Yes, the out-of-bounds access occurs in the edge case of WIDTH == 0.
This template class can, although it probably won't, be instantiated with BITS < 32, something that would lead to WIDTH == 0. 
I was mainly interested in ensuring the out-of-bounds-check before array access regardless of such assumptions.
Making it easier to follow the code logic is an extra gain.

laanwj: > This template class can, although it probably won't, be instantiated with BITS < 32

Indeed - as the code throughout the entire class is implemented, it can only be instantiated for a positive multiple of 32. Might make sense to add a compile-time assertion for that.

pavlosantoniou: Do you think adding:
`static_assert(BITS/32 > 0 && BITS%32 == 0, "Template parameter BITS must be a positive multiple of 32."");`
in the class constructor would be fine?

laanwj: Yes, that'd be nice!

pavlosantoniou:
Reopened with addition of static_assert for BITS value.

src/arith_uint256.h
pavlosantoniou: Is there something more I can do for this PR?

pavlosantoniou: I have added the assertions, is there something more I can do for this PR?

sipa:
utACK

PR #10531, opened 2017-06-05
Increased startup timeout.

colgreen:
In my experience a two second timeout will cause the startup to always fail, causing systemd to attempt shutdown while bitcoind is still starting up.


laanwj:
This conflicts with #10529.

laanwj:
This seems inactive, closing for now

PR #10533, opened 2017-06-05
[tests] Use cookie auth instead of rpcuser and rpcpassword

achow101:
Since rpcuser and rpcpassword are now deprecated, replace them with cookie auth.

### DIFF
--- a/test/functional/multi_rpc.py
+++ b/test/functional/multi_rpc.py
@@ -16,16 +16,21 @@ class HTTPBasicsTest (BitcoinTestFramework):
     def __init__(self):
         super().__init__()
         self.setup_clean_chain = False
-        self.num_nodes = 1
+        self.num_nodes = 2
 
     def setup_chain(self):
         super().setup_chain()
         #Append rpcauth to bitcoin.conf before initialization
         rpcauth = "rpcauth=rt:93648e835a54c573682c2eb19f882535$7681e9c5b74bdd85e78166031d2058e1069b3ed7ed967c93fc63abba06f31144"
         rpcauth2 = "rpcauth=rt2:f8607b1a88861fac29dfccf9b52ff9f$ff36a0c23c8c62b4846112e50fa888416e94c17bfd4c42f88fd8f55ec6a3137e"
+        rpcuser = "rpcuser=rpcuser💻"
+        rpcpassword = "rpcpassword=rpcpassword🔑"
         with open(os.path.join(self.options.tmpdir+"/node0", "bitcoin.conf"), 'a', encoding='utf8') as f:
             f.write(rpcauth+"\n")
             f.write(rpcauth2+"\n")
+        with open(os.path.join(self.options.tmpdir+"/node1", "bitcoin.conf"), 'a', encoding='utf8') as f:
+            f.write(rpcuser+"\n")
+            f.write(rpcpassword+"\n")
 
     def run_test(self):
 
@@ -50,7 +55,7 @@ class HTTPBasicsTest (BitcoinTestFramework):
         conn.connect()
         conn.request('POST', '/', '{"method": "getbestblockhash"}', headers)
         resp = conn.getresponse()
-        assert_equal(resp.status==401, False)
+        assert_equal(resp.status, 200)
         conn.close()
         
         #Use new authpair to confirm both work
@@ -60,7 +65,7 @@ class HTTPBasicsTest (BitcoinTestFramework):
         conn.connect()
         conn.request('POST', '/', '{"method": "getbestblockhash"}', headers)
         resp = conn.getresponse()
-        assert_equal(resp.status==401, False)
+        assert_equal(resp.status, 200)
         conn.close()
 
         #Wrong login name with rt's password
@@ -71,7 +76,7 @@ class HTTPBasicsTest (BitcoinTestFramework):
         conn.connect()
         conn.request('POST', '/', '{"method": "getbestblockhash"}', headers)
         resp = conn.getresponse()
-        assert_equal(resp.status==401, True)
+        assert_equal(resp.status, 401)
         conn.close()
 
         #Wrong password for rt
@@ -82,7 +87,7 @@ class HTTPBasicsTest (BitcoinTestFramework):
         conn.connect()
         conn.request('POST', '/', '{"method": "getbestblockhash"}', headers)
         resp = conn.getresponse()
-        assert_equal(resp.status==401, True)
+        assert_equal(resp.status, 401)
         conn.close()
 
         #Correct for rt2
@@ -93,7 +98,7 @@ class HTTPBasicsTest (BitcoinTestFramework):
         conn.connect()
         conn.request('POST', '/', '{"method": "getbestblockhash"}', headers)
         resp = conn.getresponse()
-        assert_equal(resp.status==401, False)
+        assert_equal(resp.status, 200)
         conn.close()
 
         #Wrong password for rt2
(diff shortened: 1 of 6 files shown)


test/functional/test_framework/util.py
```diff
@@ -181,21 +181,24 @@ def initialize_datadir(dirname, n):
...
-    return 'rpcuser💻' + str(n), 'rpcpass🔑' + str(n)
-
-def rpc_url(i, rpchost=None):
-    rpc_u, rpc_p = rpc_auth_pair(i)
+    
+def get_datadir_path(dirname, n):
+    return os.path.join(dirname, "node"+str(n))
+    
+def get_auth_cookie(datadir, n):
+    with open(os.path.join(os.path.join(datadir, "regtest"), ".cookie"), 'r') as f:
```
sipa: can be written as `os.path.join(datadir, "regtest", ".cookie")` (and elsewhere)

test/functional/p2p-segwit.py
```diff
@@ -1502,14 +1502,14 @@ def test_upgrade_after_activation(self, node, node_id):
         sync_blocks(self.nodes)
 
         # Make sure that this peer thinks segwit has activated.
-        assert(get_bip9_status(node, 'segwit')['status'] == "active")
+        assert(get_bip9_status(self.nodes[node_id], 'segwit')['status'] == "active")
```
sipa: Would a `node = self.nodes[node_id]` work as well?

laanwj:
Agree on doing this by default.

I do think we should keep one regression test around for rpcuser/rpcpasswd as long as that functionality is still supported, though.

test/functional/p2p-segwit.py
achow101: It would, but I don't see why that would be any better.

achow101:
I added a test for rpcuser and rpcpassword.

sipa:
utACK; the first two commits should probably be squashed together

sdaftuar:
Concept ACK, though the pruning.py test fails for me -- I assume it needs an update similar to p2p-segwit.py.

test/functional/test_framework/util.py
```diff
@@ -215,7 +231,9 @@ def wait_for_bitcoind_start(process, url, i):
         if process.poll() is not None:
             raise Exception('bitcoind exited with status %i during initialization' % process.returncode)
         try:
-            rpc = get_rpc_proxy(url, i)
+            # Check if .cookie file to be created
+            rpc = get_rpc_proxy(rpc_url(datadir, i), i)
+            logger.info(rpc.url)
```
jnewbery: This should be a debug log (`logger.debug()`)

test/functional/test_framework/util.py
```diff
@@ -181,21 +181,37 @@ def initialize_datadir(dirname, n):
...
+            userpass = f.read()
+            split_userpass = userpass.split(':')
+            return split_userpass[0], split_userpass[1]
+    else:
+        with open(os.path.join(datadir, "bitcoin.conf"), 'r') as f:
+            user = None
+            password = None
+            for line in f:
+                if line.startswith("rpcuser="):
+                    user = line.split("=")[1].strip("\n")
```
jnewbery: Perhaps assert if there are multiple "rpcuser=" lines? Same for "rpcpassword=" lines

test/functional/multi_rpc.py
```diff
@@ -106,6 +111,45 @@ def run_test(self):
...
+        # rpcuser and rpcpassword authpair
+        rpcuserauthpair = "rpcuser:rpcpassword"
+
+        headers = {"Authorization": "Basic " + str_to_b64str(rpcuserauthpair)}
+
+        conn = http.client.HTTPConnection(url.hostname, url.port)
+        conn.connect()
+        conn.request('POST', '/', '{"method": "getbestblockhash"}', headers)
+        resp = conn.getresponse()
+        assert_equal(resp.status==401, False)
```
jnewbery: It'd be better to assert directly on the status, ie:

`assert_equal(resp.status, 200)`

test/functional/multi_rpc.py
```diff
@@ -106,6 +111,45 @@ def run_test(self):
...
+        
+        #Wrong login name with rpcuser's password
+        rpcuserauthpair = "rpcuserwrong:rpcpassword"
+        headers = {"Authorization": "Basic " + str_to_b64str(rpcuserauthpair)}
+
+        conn = http.client.HTTPConnection(url.hostname, url.port)
+        conn.connect()
+        conn.request('POST', '/', '{"method": "getbestblockhash"}', headers)
+        resp = conn.getresponse()
+        assert_equal(resp.status==401, True)
```
jnewbery: Better to assert directly on the status, ie:

`assert_equal(resp.status, 401)`

This will cause the received status to be written out in the assert if the assert fails.

test/functional/multi_rpc.py
```diff
@@ -16,16 +16,21 @@ class HTTPBasicsTest (BitcoinTestFramework):
...
         self.setup_clean_chain = False
-        self.num_nodes = 1
+        self.num_nodes = 2
 
     def setup_chain(self):
         super().setup_chain()
         #Append rpcauth to bitcoin.conf before initialization
         rpcauth = "rpcauth=rt:93648e835a54c573682c2eb19f882535$7681e9c5b74bdd85e78166031d2058e1069b3ed7ed967c93fc63abba06f31144"
         rpcauth2 = "rpcauth=rt2:f8607b1a88861fac29dfccf9b52ff9f$ff36a0c23c8c62b4846112e50fa888416e94c17bfd4c42f88fd8f55ec6a3137e"
+        rpcuser = "rpcuser=rpcuser"
```
jnewbery: Recommend that you add some unicode characters to the username/password to test that we don't regress unicode support at some point.

jnewbery (commented):
Concept ACK. It definitely makes sense to have our test framework use the mainline expected authentication method (and have a regression test for username/password as before).

However, I'd very much prefer if this happens after #10082. That introduces a TestNode class that will encapsulate all the properties and state for a node under test. I think that makes the test framework a lot clearer and prevents having to pass variables into functions in utils.py.

I've added a few review comments inline.

test/functional/test_framework/util.py
achow101: done

test/functional/test_framework/util.py
achow101: that's a mistake from my earlier debugging. removed.

test/functional/multi_rpc.py
achow101: done

test/functional/multi_rpc.py
achow101: done

test/functional/multi_rpc.py
achow101: done

laanwj: Yes, this was the reason why we use the funny unicode characters in the rpcuser/pass in the current code. It needs to be tested at some point.

sipa:
utACK 3ec5ad88e67bba74c795575d52c97fd2c7e880c1

test/functional/test_framework/util.py
```diff
@@ -181,21 +181,39 @@ def initialize_datadir(dirname, n):
...
 
-def rpc_auth_pair(n):
-    return 'rpcuser💻' + str(n), 'rpcpass🔑' + str(n)
+def get_datadir_path(dirname, n):
+    return os.path.join(dirname, "node"+str(n))
 
-def rpc_url(i, rpchost=None):
-    rpc_u, rpc_p = rpc_auth_pair(i)
+def get_auth_cookie(datadir, n):
+    if os.path.isfile(os.path.join(datadir, "regtest", ".cookie")):
```
MarcoFalke: I think you can not check for the cookie file first. This file is always created and thus, the code in the `else` block is dead code.

You might want to check for `rpcuser` first?

test/functional/test_framework/util.py
```diff
@@ -181,21 +181,39 @@ def initialize_datadir(dirname, n):
...
+            user = None
+            password = None
+            for line in f:
+                if line.startswith("rpcuser="):
+                    assert user is None # Ensure that there is only one rpcuser line
+                    user = line.split("=")[1].strip("\n")
+                if line.startswith("rpcpassword="):
+                    assert password is None # Ensure that there is only one rpcpassword line
+                    password = line.split("=")[1].strip("\n")
+            if user is None and password is None:
```
MarcoFalke: boolean `or` instead of `and`?

test/functional/p2p-segwit.py
```diff
@@ -1944,7 +1944,7 @@ def run_test(self):
         self.test_signature_version_1()
         self.test_non_standard_witness()
         sync_blocks(self.nodes)
-        self.test_upgrade_after_activation(self.nodes[2], 2)
+        self.test_upgrade_after_activation(2)
```
MarcoFalke: Please name primitive types. If you don't want to clutter the scope with variable names, you could always just use named args.

maflcko (changes requested):
utACK 3ec5ad88e67bba74c795575d52c97fd2c7e880c1, but needs feedback addressed.

achow101:
Addressed @MarcoFalke's comments

test/functional/test_framework/util.py
```diff
@@ -181,21 +181,40 @@ def initialize_datadir(dirname, n):
...
+                    password = line.split("=")[1].strip("\n")
+    if os.path.isfile(os.path.join(datadir, "regtest", ".cookie")):
+        with open(os.path.join(datadir, "regtest", ".cookie"), 'r') as f:
+            userpass = f.read()
+            split_userpass = userpass.split(':')
+            user = split_userpass[0]
+            password = split_userpass[1]
+    if user is None or password is None:
+        raise ValueError("No RPC credentials")
+    return user, password
```
MarcoFalke: I still don't understand how this could pass travis. As mentioned earlier, the `.cookie` file is always created, even when rpcuser, rpcpassword is in the `bitcoin.conf`. Thus, the current function should always "fall through" to read the cookie file (even if the rpcuser and rpcpassword was read). Maybe travis has just slow IO and the cookie is not yet created, but we shouldn't rely on this hardware detail for the test to pass.

I'd suggest you move the return statement into each body of the if blocks.

Specifically,

```py
with open('conf'):
 for line in f:
    # Read user and password ...
 if user is None or password is None:
    raise ValueError("No RPC credentials")
 return user, password

# ...

with open('cookie'):
    # Read and split
    return split_userpass[0], split_userpass[1]
```

MarcoFalke:
Please squash after addressing my feedback.

test/functional/test_framework/util.py
achow101: The `.cookie` file is never created if rpcuser and rpcpassword are specified in the `.conf` file. You can check the behavior yourself. All of the tests pass locally for me too.

MarcoFalke: Guess you are right. I must have messed up something locally when testing this. Let me check ...

maflcko (approved):

test/functional/test_framework/util.py
MarcoFalke: Sorry, I thought that only specifying `rpcuser` in the conf is enough for authentication, when in fact only specifying `rpcpassword` is enough. Cf. https://github.com/bitcoin/bitcoin/blob/1ad3d4e1261f4a444d982a1470c257c78233bda3/src/httprpc.cpp#L226

So your previous code is right.

MarcoFalke:
Tested ACK 3ec5ad88e67bba74c795575d52c97fd2c7e880c1 (not current HEAD)

PR #10535, opened 2017-06-05
[qa] fundrawtx: Fix shutdown race

maflcko:
We need to wait for node 1 to shut down completely. Otherwise we might not be able to restart the nodes and get a error:

```
Error: Cannot obtain a lock on data directory /tmp/testc_zc8wco/node1/regtest. Bitcoin Core is probably already running.


laanwj:
Thanks, I noticed this error and didn't understand why it happened (assumed it was a local problem with zombie bitcoinds). utACK https://github.com/bitcoin/bitcoin/pull/10535/commits/fa62dadabed9e7f3f2322961806a0fddb9d8f85a

gmaxwell (approved):
utACK

PR #10539, opened 2017-06-06
doc: Mention update manpages in release process

laanwj:
Before minor and major versions, update the man pages.


MarcoFalke:
This is mention in the section "Before every release candidate:"

laanwj:
OK

Issue #7379, opened 2016-01-19
salvagewallet drops valid keys


MarcoFalke:
Need to look into this next week but I will disable the check in travis in the meantime.

``` python
check -salvagewallet
Assertion failed: [Decimal('33.99911140'), Decimal('5549.99796160'), Decimal('65.99974900')] != [Decimal('0E-8'), Decimal('5549.99796160'), Decimal('65.99974900')]
```

Edit: Will submit a proper fix tomorrow...

laanwj:
"Hopefully" fixed by #7381

laanwj:
Another problem in salvagewallet, from https://travis-ci.org/bitcoin/bitcoin/jobs/107074798
```
check -salvagewallet
start_node: bitcoind started, calling bitcoin-cli -rpcwait getblockcount
start_node: calling bitcoin-cli -rpcwait getblockcount returned
start_node: bitcoind started, calling bitcoin-cli -rpcwait getblockcount
start_node: calling bitcoin-cli -rpcwait getblockcount returned
start_node: bitcoind started, calling bitcoin-cli -rpcwait getblockcount
start_node: calling bitcoin-cli -rpcwait getblockcount returned
Assertion failed: [Decimal('33.99911140'), Decimal('5549.99796180'), Decimal('59.99974800')] != [Decimal('0E-8'), Decimal('5549.99796180'), Decimal('59.99974800')]
  File "/home/travis/build/bitcoin/bitcoin/bitcoin-x86_64-unknown-linux-gnu/qa/rpc-tests/test_framework/test_framework.py", line 135, in main
    self.run_test()
  File "/home/travis/build/bitcoin/bitcoin/bitcoin-x86_64-unknown-linux-gnu/qa/rpc-tests/wallet.py", line 274, in run_test
    assert_equal(balance_nodes, [self.nodes[i].getbalance() for i in range(3)])
  File "/home/travis/build/bitcoin/bitcoin/bitcoin-x86_64-unknown-linux-gnu/qa/rpc-tests/test_framework/util.py", line 415, in assert_equal
    raise AssertionError("%s != %s"%(str(thing1),str(thing2)))
Stopping nodes
Cleaning up
Failed
```

MarcoFalke:
I am aware of that. Need to look into this soon™

adamjonas:
`-salvagewallet` was removed in #18918

PR #10540, opened 2017-06-06
[WIP] Salvage wallet should not set the aggressive flag on Db::verify()

### LINKED ISSUE #7463: salvagewallet fails verification

jnewbery:
Currently, the salvagewallet function sets the `DB_AGGRESSIVE` flag when calling BDB's `Db::verify()` and then tries to load the output back into a database. Here's the documentation for that flag:

> Output all the key/data pairs in the file that can be found. By default, Db::verify() does not assume corruption. For example, if a key/data pair on a page is marked as deleted, it is not then written to the output file. When DB_AGGRESSIVE is specified, corruption is assumed, and any key/data pair that can be found is written. In this case, key/data pairs that are corrupted or have been deleted may appear in the output (even if the file being salvaged is in no way corrupt), and **the output will almost certainly require editing before being loaded into a database.**

We should not be setting the DB_AGGRESSIVE flag and then immediately trying to load the output into a database.

For more technical details see https://github.com/bitcoin/bitcoin/issues/7463#issuecomment-306336149.

Running with aggressive mode can lead to full database corruption and loss of private keys (#7379). The wallet is backed up to wallet.timestamp.bak, but the original wallet.dat file will lose some or all of its key-value pairs. That's obviously a very serious usability issue for users. Many will not know about the existence of the wallet.timestamp.bak backup files.

I think in almost all cases it's better to run `Db::verify()` without the aggressive flag set. There perhaps should be a tool to run Db::verify() with the aggressive flag. If the database has been badly corrupted, it may be the only way to recover key-value pairs. However, this should be done in an external tool as the output will almost certainly require editing before being reloaded. #8745 (bitcoin-wallet-tool) is the right place for this.

This PR removes the DB_AGGRESSIVE flag from calls to `Db::verify()`. It also re-enables the `-salvagewallet` incidental test in wallet.py. I've also included a commit which cleans up that test, which can be omitted if it's not wanted.

I've run wallet.py 100 times and not yet hit the salvagewallet issue in #7463. That was intermittent, so I may just be getting lucky, but I think this is an improvement.

Fixes #7463

### DIFF
--- a/src/wallet/db.cpp
+++ b/src/wallet/db.cpp
@@ -164,8 +164,7 @@ bool CDB::Recover(const std::string& filename, void *callbackDataIn, bool (*reco
 {
     // Recovery procedure:
     // move wallet file to wallet.timestamp.bak
-    // Call Salvage with fAggressive=true to
-    // get as much data as possible.
+    // Call Salvage to get as much data as possible.
     // Rewrite salvaged data to fresh wallet file
     // Set -rescan so any missing transactions will be
     // found.
@@ -183,13 +182,13 @@ bool CDB::Recover(const std::string& filename, void *callbackDataIn, bool (*reco
     }
 
     std::vector<CDBEnv::KeyValPair> salvagedData;
-    bool fSuccess = bitdb.Salvage(newFilename, true, salvagedData);
+    bool fSuccess = bitdb.Salvage(newFilename, salvagedData);
     if (salvagedData.empty())
     {
-        LogPrintf("Salvage(aggressive) found no records in %s.\n", newFilename);
+        LogPrintf("Salvage found no records in %s.\n", newFilename);
         return false;
     }
-    LogPrintf("Salvage(aggressive) found %u records\n", salvagedData.size());
+    LogPrintf("Salvage found %u records\n", salvagedData.size());
 
     std::unique_ptr<Db> pdbCopy(new Db(bitdb.dbenv, 0));
     int ret = pdbCopy->open(NULL,               // Txn pointer
@@ -289,14 +288,12 @@ static const char *HEADER_END = "HEADER=END";
 /* End of key/value data */
 static const char *DATA_END = "DATA=END";
 
-bool CDBEnv::Salvage(const std::string& strFile, bool fAggressive, std::vector<CDBEnv::KeyValPair>& vResult)
+bool CDBEnv::Salvage(const std::string& strFile, std::vector<CDBEnv::KeyValPair>& vResult)
 {
     LOCK(cs_db);
     assert(mapFileUseCount.count(strFile) == 0);
 
     u_int32_t flags = DB_SALVAGE;
-    if (fAggressive)
-        flags |= DB_AGGRESSIVE;
 
     std::stringstream strDump;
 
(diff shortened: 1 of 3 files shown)


src/wallet/db.cpp
```diff
@@ -289,27 +288,20 @@ static const char *HEADER_END = "HEADER=END";
...
-    if (fAggressive)
-        flags |= DB_AGGRESSIVE;
 
     std::stringstream strDump;
 
     Db db(dbenv, 0);
     int result = db.verify(strFile.c_str(), NULL, &strDump, flags);
     if (result == DB_VERIFY_BAD) {
         LogPrintf("CDBEnv::Salvage: Database salvage found errors, all data may not be recoverable.\n");
-        if (!fAggressive) {
```
jnewbery: Note for reviewers: This branch was impossible to hit since Salvage() was never called with fAggressive set to false

laanwj:
`-salvagewallet` is meant to "salvage" - do everything it can - to get something usable from corrupted wallets.
I think a better approach would be to try with non-aggressive first, and *only if that fails* try aggressively.
But disabling aggressive mode completely is throwing away the baby with the bathwater and neutering this mode completely.

jnewbery:
> neutering this mode completely.

That doesn't seem to be true from my testing. Running verify without DB_AGGRESSIVE does remove corruption from the wallet.dat file and make the balance available. Perhaps that's only because -salvagewallet is also causing a rescan to happen.

Do you have experience of -salvagewallet recovering a wallet where running a non-aggressive verify would not have recovered it? I have evidence of -salvagewallet making the situation worse, but we don't have any directed tests for salvagewallet, so I haven't seen any examples of it making the situation better. The documentation "the output will almost certainly require editing before being loaded into a database." makes me think that -salvagewallet is very unlikely to actually work.

gmaxwell:
I was going to make pretty much exactly the same post that Wladimir made:   Salvage is supposed to be used on a corrupted file-- thats the whole point, perhaps we should do both, if you think that we shouldn't be working on corrupted files there perhaps we should remove it. (I'd be pretty game to remove it: I think it's mistakenly used by users FAR too often even though it's something of a nuclear option.)

> Do you have experience of -salvagewallet recovering a wallet where running a non-aggressive verify would not have recovered it?

Yes, I have had wallets that were corrupted that needed agressive mode to extract the keys.   I've also seen aggressive mode fail to read keys in totally non-corrupted wallets...

Keys in the wallet are stored with this terrible der private key encoding that is obvious from orbit... if we had an extra tool it could also scan just looking for that encoding as https://bitcointalk.org/index.php?topic=25091.0 does-- which would allow recovering files that BDB won't do anything with too.

But in all cases the recovery should probably try all reading methods available and union them.

laanwj:
> But in all cases the recovery should probably try all reading methods available and union them.

Excellent comment. Completely agree with this. Would make sense for `-salvagewallet` to try various techniques, then combine the result.

> Keys in the wallet are stored with this terrible der private key encoding that is obvious from orbit... 

Yes, that would be an additional strategy for non-encrypted keys.

jnewbery:
Thanks for the great feedback @laanwj and @gmaxwell . Sounds like there are plenty of improvements we can make here. There are at least three recovery techniques we can use:
- attempt to salvage non-aggressively
- attempt to salvage aggressively
- do a memory scan for DER encoded private keys in the .dat file.

Those operations should be done by a separate tool and not loaded immediately into bitcoind, since the output could be worse than simply reading the .dat file, and can lead to loss of keys. I'll put this on hold until we have #8745 merged.

laanwj:
@jnewbery Sounds good to me

PR #10547, opened 2017-06-07
[tests] Use FastRandomContext instead of boost::random::{mt19937,uniform_int_distribution}

practicalswift:
Use `FastRandomContext` instead of `boost::random::{mt19937,uniform_int_distribution}`.

Was:
> Use `std::random::mt19937`/`uniform_int_distribution` (C++11) instead of `boost::random` equivalents.

### DIFF
--- a/src/test/scheduler_tests.cpp
+++ b/src/test/scheduler_tests.cpp
@@ -8,8 +8,6 @@
 #include "test/test_bitcoin.h"
 
 #include <boost/bind.hpp>
-#include <boost/random/mersenne_twister.hpp>
-#include <boost/random/uniform_int_distribution.hpp>
 #include <boost/thread.hpp>
 #include <boost/test/unit_test.hpp>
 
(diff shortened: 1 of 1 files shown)


laanwj:
Why not use our own fast_random_context?

practicalswift:
@laanwj Good idea! Now using `FastRandomContext`. Looks OK now?

laanwj:
Yes, that's a cool minimal-impact solution using the lambda expressions.

src/test/scheduler_tests.cpp
```diff
@@ -56,10 +54,10 @@ BOOST_AUTO_TEST_CASE(manythreads)
...
     boost::mutex counterMutex[10];
     int counter[10] = { 0 };
-    boost::random::mt19937 rng(42);
-    boost::random::uniform_int_distribution<> zeroToNine(0, 9);
-    boost::random::uniform_int_distribution<> randomMsec(-11, 1000);
-    boost::random::uniform_int_distribution<> randomDelta(-1000, 1000);
+    FastRandomContext rng(42);
+    auto zeroToNine = [](FastRandomContext& rc) { return rc.randrange(10); }; // [0, 9]
+    auto randomMsec = [](FastRandomContext& rc) { return -11 + rc.randrange(1012); }; // [-11, 1000]
+    auto randomDelta = [](FastRandomContext& rc) { return -1000 + rc.randrange(2001); }; // [-1000, 1000]
```
MarcoFalke: This lambda returns an unsigned expression and such can never be negative.

maflcko (changes requested):

src/test/scheduler_tests.cpp
laanwj: Good catch, hadn't realized `randrange` returns an unsigned value. This is a risk with `auto`.

practicalswift: @MarcoFalke Oh, good catch! Now explicitly setting the return type. Thanks a lot for noticing!

gmaxwell (approved):
utACK

jonasschnelli:
utACK 227ae9b34dce632e42ca9f22139e6efad0485df1.
Great use of lambdas.

PR #9306, opened 2016-12-09
Make CCoinsViewCache::Cursor() return latest data

ryanofsky:
Change CCoinsViewCache::Cursor() method to return a cursor yielding the latest CCoins entries, instead of just previous entries prior to the last cache flush.

The CCoinsViewCache::Cursor method is not currently used. This change just enables new features that rely on scanning the UXTO set to work correctly (for example https://github.com/bitcoin/bitcoin/pull/9152, which adds a sweepprivkeys RPC, and https://github.com/bitcoin/bitcoin/pull/9137, which improves handling of imported keys for nodes with pruning enabled.)

PR #10550, opened 2017-06-07
Don't return stale data from CCoinsViewCache::Cursor()

ryanofsky:
CCoinsViewCache doesn't actually support cursor iteration returning the
current contents of the cache, so raise an error when the cursor method is
called instead of returning a cursor that iterates over stale data.

Also update the gettxoutsetinfo RPC which was relying on the old behavior to be
explicit about which view it is returning data about.

### DIFF
--- a/src/coins.h
+++ b/src/coins.h
@@ -212,6 +212,9 @@ public:
     uint256 GetBestBlock() const;
     void SetBestBlock(const uint256 &hashBlock);
     bool BatchWrite(CCoinsMap &mapCoins, const uint256 &hashBlock);
+    CCoinsViewCursor* Cursor() const {
+        throw std::logic_error("CCoinsViewCache cursor iteration not supported.");
+    }
 
     /**
      * Check if we have the given utxo already loaded in this cache.
--- a/src/coins.h
+++ b/src/coins.h
@@ -207,12 +207,12 @@ public:
     CCoinsViewCache(CCoinsView *baseIn);
 
     // Standard CCoinsView methods
-    bool GetCoin(const COutPoint &outpoint, Coin &coin) const;
-    bool HaveCoin(const COutPoint &outpoint) const;
-    uint256 GetBestBlock() const;
+    bool GetCoin(const COutPoint &outpoint, Coin &coin) const override;
+    bool HaveCoin(const COutPoint &outpoint) const override;
+    uint256 GetBestBlock() const override;
     void SetBestBlock(const uint256 &hashBlock);
-    bool BatchWrite(CCoinsMap &mapCoins, const uint256 &hashBlock);
-    CCoinsViewCursor* Cursor() const {
+    bool BatchWrite(CCoinsMap &mapCoins, const uint256 &hashBlock) override;
+    CCoinsViewCursor* Cursor() const override {
         throw std::logic_error("CCoinsViewCache cursor iteration not supported.");
     }
 
--- a/src/validation.h
+++ b/src/validation.h
@@ -36,6 +36,7 @@ class CBlockIndex;
 class CBlockTreeDB;
 class CBloomFilter;
 class CChainParams;
+class CCoinsViewDB;
 class CInv;
 class CConnman;
 class CScriptCheck;
@@ -481,6 +482,9 @@ bool ResetBlockFailureFlags(CBlockIndex *pindex);
 
 /** The currently-connected chain of blocks (protected by cs_main). */
 extern CChain chainActive;
+
+/** Global variable that points to the coins database (protected by cs_main) */
+extern CCoinsViewDB *pcoinsdbview;
 
 /** Global variable that points to the active CCoinsView (protected by cs_main) */
 extern CCoinsViewCache *pcoinsTip;
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -162,7 +162,6 @@ public:
     // Writes do not need similar protection, as failure to write is handled by the caller.
 };
 
-static CCoinsViewDB *pcoinsdbview = NULL;
 static CCoinsViewErrorCatcher *pcoinscatcher = NULL;
 static std::unique_ptr<ECCVerifyHandle> globalVerifyHandle;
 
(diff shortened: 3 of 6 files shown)


src/coins.h
```diff
@@ -212,6 +212,9 @@ class CCoinsViewCache : public CCoinsViewBacked
     uint256 GetBestBlock() const;
     void SetBestBlock(const uint256 &hashBlock);
     bool BatchWrite(CCoinsMap &mapCoins, const uint256 &hashBlock);
+    CCoinsViewCursor* Cursor() const override {
+        throw std::logic_error("CCCoinsViewCache cursor iteration not supported.");
```
sipa: CCC?

ryanofsky: Fixed in fe76b30c838a4de4afb8382c4f11948d91536d22

ryanofsky (commented):
Squashed fe76b30c838a4de4afb8382c4f11948d91536d22 -> c9d5b100209988ae2ffe3339c8b30dea28449ac1 ([pr/coins-stale.2](https://github.com/ryanofsky/bitcoin/commits/pr/coins-stale.2) -> [pr/coins-stale.3](https://github.com/ryanofsky/bitcoin/commits/pr/coins-stale.3))

jonasschnelli:
utACK c9d5b100209988ae2ffe3339c8b30dea28449ac1

src/coins.h
```diff
@@ -212,6 +212,9 @@ class CCoinsViewCache : public CCoinsViewBacked
     uint256 GetBestBlock() const;
     void SetBestBlock(const uint256 &hashBlock);
     bool BatchWrite(CCoinsMap &mapCoins, const uint256 &hashBlock);
+    CCoinsViewCursor* Cursor() const override {
```
laanwj: This 'override' is causing tons of warnings here.
```
/home/orion/projects/bitcoin/bitcoin/src/coins.h:210:10: warning: 'GetCoin' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
    bool GetCoin(const COutPoint &outpoint, Coin &coin) const;
```
Some compilers (clang5, in this case) demand that a whole class is annotated with it if it is used anywhere. So I'd argue against adding it here, or in a separate commit add it to all appropriate CCoinsViewCache methods.

laanwj:
This makes a lot of sense, the use of API here has always bothered me but I wasn't sure how to fix it. Seems a good stopgap until we need #9306.
utACK apart from nit.

src/validation.h
```diff
@@ -482,6 +483,9 @@ bool ResetBlockFailureFlags(CBlockIndex *pindex);
 /** The currently-connected chain of blocks (protected by cs_main). */
 extern CChain chainActive;
 
+/** Global variable that points to the coins database (protected by cs_main) */
+extern CCoinsViewDB *pcoinsdbview;
```
paveljanik: Coins View DB vs coins db view?

ryanofsky: > Coins View DB vs coins db view?

Are you suggesting to rename the variable? If so I would slightly prefer not to, because this is a preexisting variable which this PR is just making public.

src/coins.h
ryanofsky: Removed in c2a7640d2ac5e6b1c8a238cfffa2b6f86fc9e490.

ryanofsky (commented):
Squashed c2a7640d2ac5e6b1c8a238cfffa2b6f86fc9e490 -> 24e44c354d5b9174cb9cb7d157a985e197aa7886 ([pr/coins-stale.4](https://github.com/ryanofsky/bitcoin/commits/pr/coins-stale.4) -> [pr/coins-stale.6](https://github.com/ryanofsky/bitcoin/commits/pr/coins-stale.6))

src/validation.h
paveljanik: Agreed. Thanks for clarification.

src/coins.h
ryanofsky: Added back overrides in a51ff25b6ea93ecb3dccbd37ca3a11c9d8fc0693

src/coins.h
```diff
@@ -207,11 +207,14 @@ class CCoinsViewCache : public CCoinsViewBacked
...
-    bool HaveCoin(const COutPoint &outpoint) const;
-    uint256 GetBestBlock() const;
+    bool GetCoin(const COutPoint &outpoint, Coin &coin) const override;
+    bool HaveCoin(const COutPoint &outpoint) const override;
+    uint256 GetBestBlock() const override;
     void SetBestBlock(const uint256 &hashBlock);
     bool BatchWrite(CCoinsMap &mapCoins, const uint256 &hashBlock);
+    CCoinsViewCursor* Cursor() const override {
+        throw std::logic_error("CCoinsViewCache cursor iteration not supported.");
+    }
```
laanwj: Thanks! Though I'm still getting one warning after this:
```
In file included from /home/.../bitcoin/src/rpc/rawtransaction.cpp:8:
/home/.../bitcoin/src/coins.h:214:10: warning: 'BatchWrite' overrides a member function but is not marked 'override' [-Winconsistent-missing-override]
    bool BatchWrite(CCoinsMap &mapCoins, const uint256 &hashBlock);
         ^
/home/.../bitcoin/src/coins.h:186:10: note: overridden virtual function is here
    bool BatchWrite(CCoinsMap &mapCoins, const uint256 &hashBlock) override;
```

ryanofsky: Fixed 36a14a9bceeafd7f1565c10579f698ec135d8bc6

ryanofsky (commented):
Squashed 36a14a9bceeafd7f1565c10579f698ec135d8bc6 -> 3ff1fa8c4a6198ebc277bbf814b4e07bf78eb98a ([pr/coins-stale.8](https://github.com/ryanofsky/bitcoin/commits/pr/coins-stale.8) -> [pr/coins-stale.9](https://github.com/ryanofsky/bitcoin/commits/pr/coins-stale.9))

PR #10551, opened 2017-06-07
[Tests] Wallet encryption functional tests

achow101:
Added a functional test which tests the encryptwallet, walletpassphrase, walletpassphrasechange, and walletlock RPCs

### DIFF
--- a/test/functional/wallet-encryption.py
+++ b/test/functional/wallet-encryption.py
@@ -1,0 +1,66 @@
+#!/usr/bin/env python3
+# Copyright (c) 2016 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Test Wallet encryption"""
+
+import time
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import (
+    assert_equal,
+    assert_raises_jsonrpc,
+    bitcoind_processes,
+    BITCOIND_PROC_WAIT_TIMEOUT,
+)
+
+class WalletEncryptionTest(BitcoinTestFramework):
+
+    def __init__(self):
+        super().__init__()
+        self.setup_clean_chain = True
+        self.num_nodes = 1
+
+    def run_test(self):
+        passphrase = "WalletPassphrase"
+        passphrase2 = "SecondWalletPassphrase"
+
+        # Make sure the wallet isn't encrypted first
+        address = self.nodes[0].getnewaddress()
+        privkey = self.nodes[0].dumpprivkey(address)
+        assert_equal(privkey[:1], "c")
+        assert_equal(len(privkey), 52)
+
+        # Encrypt the wallet
+        self.nodes[0].encryptwallet(passphrase)
+        bitcoind_processes[0].wait(timeout=BITCOIND_PROC_WAIT_TIMEOUT)
+        self.nodes[0] = self.start_node(0, self.options.tmpdir)
+
+        # Test that the wallet is encrypted
+        assert_raises_jsonrpc(-13, "Please enter the wallet passphrase with walletpassphrase first", self.nodes[0].dumpprivkey, address)
+
+        # Check that walletpassphrase works
+        self.nodes[0].walletpassphrase(passphrase, 2)
+        assert_equal(privkey, self.nodes[0].dumpprivkey(address))
+
+        # Check that the timeout is right
+        time.sleep(2)
+        assert_raises_jsonrpc(-13, "Please enter the wallet passphrase with walletpassphrase first", self.nodes[0].dumpprivkey, address)
+
+        # Test wrong passphrase
+        assert_raises_jsonrpc(-14, "wallet passphrase entered was incorrect", self.nodes[0].walletpassphrase, passphrase + "wrong", 10)
+
+        # Test walletlock
+        self.nodes[0].walletpassphrase(passphrase, 84600)
+        assert_equal(privkey, self.nodes[0].dumpprivkey(address))
+        self.nodes[0].walletlock()
+        assert_raises_jsonrpc(-13, "Please enter the wallet passphrase with walletpassphrase first", self.nodes[0].dumpprivkey, address)
+
+        # Test passphrase changes
+        self.nodes[0].walletpassphrasechange(passphrase, passphrase2)
+        assert_raises_jsonrpc(-14, "wallet passphrase entered was incorrect", self.nodes[0].walletpassphrase, passphrase, 10)
+        self.nodes[0].walletpassphrase(passphrase2, 10)
+        assert_equal(privkey, self.nodes[0].dumpprivkey(address))
+
+if __name__ == '__main__':
+    WalletEncryptionTest().main()
--- a/test/functional/test_runner.py
+++ b/test/functional/test_runner.py
@@ -111,6 +111,7 @@ BASE_SCRIPTS= [
     'rpcnamedargs.py',
     'listsinceblock.py',
     'p2p-leaktests.py',
+    'wallet-encryption.py',
 ]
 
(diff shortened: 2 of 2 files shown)


jonasschnelli:
Meh.
Those code parts are already tested in bumpfee / keypool / fundrawtransaction / wallet-dump.
IMO we don't need another test for this.

Or where are the improvements?

achow101:
The improvements are that walletpassphrasechange is now tested and that these functionality are explicitly being tested. The usage of these RPCs in other tests are more for doing stuff with testing other stuff rather than actually testing the behavior of these RPCs themselves.

jonasschnelli:
I would recommend to add the missing test to an existing wallet test and maybe rename the test file (so the name may make more sense).
It's just my opinion, I think we should try to combine tests (if it makes sense) to avoid unnecessary test-bootstraps/runs.

test/functional/wallet-encryption.py
```diff
@@ -0,0 +1,64 @@
...
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Test Wallet encryption"""
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import (
+    assert_equal,
+    assert_raises_jsonrpc,
+)
+import os
```
jnewbery: `os` and `shutil` are imported but not used.

test/functional/wallet-encryption.py
```diff
@@ -0,0 +1,64 @@
+#!/usr/bin/env python3
+# Copyright (c) 2016 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Test Wallet encryption"""
+
```
jnewbery: nit: prefer [PIP8](https://www.python.org/dev/peps/pep-0008/) import order (std library, followed by 3rd party library, followed by project imports)

test/functional/wallet-encryption.py
```diff
@@ -0,0 +1,64 @@
...
+import time
+
+class WalletEncryptionTest(BitcoinTestFramework):
+
+    def __init__(self):
+        super().__init__()
+        self.setup_clean_chain = True
+        self.num_nodes = 1
+
+    def run_test (self):
```
jnewbery: no space before parens for function parameters please.

test/functional/wallet-encryption.py
```diff
@@ -0,0 +1,64 @@
...
+        tmpdir = self.options.tmpdir
```
jnewbery: I don't think this local variable is needed. You can just use `self.options.tmpdir` directly in the call to `start_node()` below.

test/functional/wallet-encryption.py
```diff
@@ -0,0 +1,64 @@
...
+        passphrase = "WalletPassphrase"
+        passphrase2 = "SecondWalletPassphrase"
+        
+        # Make sure the wallet isn't encrypted first
+        address = self.nodes[0].getnewaddress()
+        privkey = self.nodes[0].dumpprivkey(address)
```
jnewbery: Is it possible to assert that privkey is a legitimate privkey here? You're asserting everything against it further down, but you never assert that privkey is actually a private key. This test would still pass if, for example, `dumpprivkey()` always returns an empty string.

test/functional/wallet-encryption.py
```diff
@@ -0,0 +1,64 @@
...
+        
+        # Encrypt the wallet
+        self.nodes[0].encryptwallet(passphrase)
+        self.stop_node(0)
```
jnewbery: This call to `stop_node()` is not needed (and may be dangerous). the encryptwallet RPC causes bitcoind to stop. Really what you need to do here is wait for the bitcoind process to terminate rather than call `stop_node()`. See https://github.com/bitcoin/bitcoin/blob/9c248e39f2807a7f89e555e99cc55cb2de1ad335/test/functional/fundrawtransaction.py#L455 for example.

`stop_node()` will assert if the bitcoind process no longer exists, so your current code is a bit racy. If the call to encryptwallet causes the node to stop before stop_node() is called, then the test will assert and fail.

test/functional/wallet-encryption.py
```diff
@@ -0,0 +1,64 @@
...
+        
+        # Test that the wallet is encrypted
+        assert_raises_jsonrpc(-13, "Please enter the wallet passphrase with walletpassphrase first", self.nodes[0].dumpprivkey, address)
+        
+        # Check that walletpassphrase works
+        self.nodes[0].walletpassphrase(passphrase, 10)
+        assert_equal(privkey, self.nodes[0].dumpprivkey(address))
+        
+        # Check that the timeout is right
+        time.sleep(10)
```
jnewbery: Consider reducing the timeout in the call to walletpassphrase to shorten this test. Alternatively, you could use mocktime to avoid waiting for 10 seconds.

jnewbery (commented):
Looks good. I suggest you use a tool like flake8 or pylint to catch common style issues like unused imports, trailing whitespace, etc. There's a bunch of trailing whitespace in the new test which I haven't commented on individually.

Longer term, I think wallet encryption should be handled by an external tool (#8745), but even if that happens, it still makes sense to have tests for bitcoind unlocking the encrypted walled.

jnewbery:
> Those code parts are already tested in bumpfee / keypool / fundrawtransaction / wallet-dump. IMO we don't need another test for this.

Only just saw this comment. I agree with @achow101 that the calls to encryptwallet in those tests is incidental and it makes sense to have directed tests of encrypting/decrypting and changing passphrases. It may make sense to add it to an existing test, but on the other hand I'm not too concerned about adding a new test file.

There's a balance here: adding new test cases conceptually separates the different test objectives, and also allows tests to be run in parallel, but test setup/teardown is a non-negligable part of the time taken to run the tests.

test/functional/wallet-encryption.py
achow101: I suppose it would be possible to do a prefix and length check on it. Alternatively a private key could be hard-coded and imported into the wallet. But actually checking that the private key is legitimate, I'm not sure that that can be done.

test/functional/wallet-encryption.py
achow101: done

test/functional/wallet-encryption.py
achow101: done

test/functional/wallet-encryption.py
achow101: done

test/functional/wallet-encryption.py
achow101: done

test/functional/wallet-encryption.py
achow101: done

test/functional/wallet-encryption.py
achow101: done

test/functional/wallet-encryption.py
jnewbery: Presumably it'd be possible to check that it's a base58 string of the correct length? That should be enough. I think we should at least detect the case where the return value is obviously *not* a private key.

achow101: I added a prefix check and length check since there isn't actually anything to do a base58 check in the tests.

jonasschnelli:
> There's a balance here: adding new test cases conceptually separates the different test objectives, and also allows tests to be run in parallel, but test setup/teardown is a non-negligable part of the time taken to run the tests.

Fair point.

jnewbery:
Looks good to me! Tested ACK ec98b78e1e1ce6bd9e6189788f7689aebfbf343c

laanwj:
ACK ec98b78

PR #10554, opened 2017-06-08
ZMQ: add publishers for wallet transactions.

somdoron:
There is no way to only get real time notifications of transaction that affect the wallet.
You have to do that manually by enabling zmqrawtx and filter out transactions.

I'm suggesting adding two new publisers, both for hash and raw wallet transactions.

Also topic will indicate if transaction came from mempool or block so developers can handle the transaction accordingly without a RPC round trip to bitcoind.

Tests and documentation are updated.

### DIFF
--- a/doc/zmq.md
+++ b/doc/zmq.md
@@ -58,8 +58,10 @@ Currently, the following notifications are supported:
 
     -zmqpubhashtx=address
     -zmqpubhashblock=address
+    -zmqpubhashwallettx=address
     -zmqpubrawblock=address
     -zmqpubrawtx=address
+    -zmqpubrawwallettx=address
 
 The socket type is PUB and the address must be a valid ZeroMQ socket
 address. The same address can be used in more than one notification.
@@ -74,6 +76,15 @@ corresponds to the notification type. For instance, for the
 notification `-zmqpubhashtx` the topic is `hashtx` (no null
 terminator) and the body is the transaction hash (32
 bytes).
+
+For wallet transaction notifications (both hash and tx), the
+topic also indicate if the transaction came from a block
+or mempool. If originated from mempool `-mempool` postfix
+will be added to the topic, for block `-block` postfix will
+be added. Because zeromq is using prefix matching for topics
+you can subscribe to `rawwallettx` (or `hashwallettx`) to get
+both notifications. If you only want one type of notification
+subscribe to either `rawwallettx-mempool` or `rawwallettx-block`.
 
 These options can also be provided in bitcoin.conf.
 
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -959,6 +959,9 @@ bool CWallet::AddToWallet(const CWalletTx& wtxIn, bool fFlushOnClose)
     // Notify UI of new or updated transaction
     NotifyTransactionChanged(this, hash, fInsertedNew ? CT_NEW : CT_UPDATED);
 
+    // Notify listeners on new wallet transaction
+    TransactionAddedToWallet(wtx.tx, wtx.hashBlock);
+
     // notify an external script when a wallet transaction comes in or is updated
     std::string strCmd = gArgs.GetArg("-walletnotify", "");
 
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -58,6 +58,10 @@ CFeeRate CWallet::minTxFee = CFeeRate(DEFAULT_TRANSACTION_MINFEE);
  * Override with -fallbackfee
  */
 CFeeRate CWallet::fallbackFee = CFeeRate(DEFAULT_FALLBACK_FEE);
+/*
+ * Signal when transactions are added to wallet
+ */
+boost::signals2::signal<void (const CTransactionRef &ptxn, const uint256 &blockHash)> CWallet::TransactionAddedToWallet;
 
 CFeeRate CWallet::m_discard_rate = CFeeRate(DEFAULT_DISCARD_FEE);
 
@@ -960,7 +964,7 @@ bool CWallet::AddToWallet(const CWalletTx& wtxIn, bool fFlushOnClose)
     NotifyTransactionChanged(this, hash, fInsertedNew ? CT_NEW : CT_UPDATED);
 
     // Notify listeners on new wallet transaction
-    TransactionAddedToWallet(wtx.tx, wtx.hashBlock);
+    CWallet::TransactionAddedToWallet(wtx.tx, wtx.hashBlock);
 
     // notify an external script when a wallet transaction comes in or is updated
     std::string strCmd = gArgs.GetArg("-walletnotify", "");
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -1060,7 +1060,7 @@ public:
     boost::signals2::signal<void (CWallet *wallet, const uint256 &hashTx,
             ChangeType status)> NotifyTransactionChanged;
 
-    boost::signals2::signal<void (const CTransactionRef &ptxn,
+    static boost::signals2::signal<void (const CTransactionRef &ptxn,
                                   const uint256 &blockHash)> TransactionAddedToWallet;
 
     /** Show progress e.g. for rescan */
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -1060,6 +1060,9 @@ public:
     boost::signals2::signal<void (CWallet *wallet, const uint256 &hashTx,
             ChangeType status)> NotifyTransactionChanged;
 
+    boost::signals2::signal<void (const CTransactionRef &ptxn,
+                                  const uint256 &blockHash)> TransactionAddedToWallet;
+
     /** Show progress e.g. for rescan */
     boost::signals2::signal<void (const std::string &title, int nProgress)> ShowProgress;
 
--- a/src/zmq/zmqabstractnotifier.cpp
+++ b/src/zmq/zmqabstractnotifier.cpp
@@ -20,3 +20,7 @@ bool CZMQAbstractNotifier::NotifyTransaction(const CTransaction &/*transaction*/
 {
     return true;
 }
+
+bool CZMQAbstractNotifier::NotifyWalletTransaction(const CTransaction &transaction, const uint256 &hashBlock){
+    return true;
+}
--- a/src/zmq/zmqnotificationinterface.cpp
+++ b/src/zmq/zmqnotificationinterface.cpp
@@ -81,6 +81,10 @@ bool CZMQNotificationInterface::Initialize()
     LogPrint(BCLog::ZMQ, "zmq: Initialize notification interface\n");
     assert(!pcontext);
 
+#ifdef ENABLE_WALLET
+    CWallet::TransactionAddedToWallet.connect(boost::bind(&CZMQNotificationInterface::TransactionAddedToWallet, this, _1, _2));
+#endif
+
     pcontext = zmq_init(1);
 
     if (!pcontext)
@@ -118,9 +122,7 @@ void CZMQNotificationInterface::Shutdown()
     LogPrint(BCLog::ZMQ, "zmq: Shutdown notification interface\n");
 
 #ifdef ENABLE_WALLET
-    for (CWalletRef pwallet : vpwallets) {
-        pwallet->TransactionAddedToWallet.disconnect(boost::bind(&CZMQNotificationInterface::TransactionAddedToWallet, this, _1, _2));
-    }
+    CWallet::TransactionAddedToWallet.disconnect(boost::bind(&CZMQNotificationInterface::TransactionAddedToWallet, this, _1, _2));
 #endif
 
     if (pcontext)
@@ -136,14 +138,6 @@ void CZMQNotificationInterface::Shutdown()
         pcontext = nullptr;
     }
 }
-
-#ifdef ENABLE_WALLET
-void CZMQNotificationInterface::ConnectToWalletSignals() {
-     for (CWalletRef pwallet : vpwallets) {
-        pwallet->TransactionAddedToWallet.connect(boost::bind(&CZMQNotificationInterface::TransactionAddedToWallet, this, _1, _2));
-    }
-}
-#endif
 
 void CZMQNotificationInterface::UpdatedBlockTip(const CBlockIndex *pindexNew, const CBlockIndex *pindexFork, bool fInitialDownload)
 {
--- a/src/zmq/zmqnotificationinterface.cpp
+++ b/src/zmq/zmqnotificationinterface.cpp
@@ -10,6 +10,10 @@
 #include "streams.h"
 #include "util.h"
 
+#ifdef ENABLE_WALLET
+#include "../wallet/wallet.h"
+#endif
+
 void zmqError(const char *str)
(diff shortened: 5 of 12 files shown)


src/validationinterface.h
```diff
@@ -52,6 +53,8 @@ struct CMainSignals {
     boost::signals2::signal<void (const CBlockIndex *, const CBlockIndex *, bool fInitialDownload)> UpdatedBlockTip;
     /** Notifies listeners of a transaction having been added to mempool. */
     boost::signals2::signal<void (const CTransactionRef &)> TransactionAddedToMempool;
+    /** Notifies listeners of a transaction having been added to the wallet. */
+    boost::signals2::signal<void (const CTransactionRef &, uint256 hashBlock)> TransactionAddedWallet;
```
ryanofsky: Probably s/TransactionAddedWallet/TransactionAddedToWallet/ for consistency

ryanofsky: Throughout the PR, should use `const uint256&` instead of `uint256` arg type to avoid unnecessary copies.

doc/zmq.md
```diff
@@ -75,6 +77,11 @@ notification `-zmqpubhashtx` the topic is `hashtx` (no null
 terminator) and the body is the hexadecimal transaction hash (32
 bytes).
 
+For wallet transaction notifications (both hash and tx), the topic also indicate if the transaction came from a block or mempool.
```
ryanofsky: Maybe wrap these lines for consistency with rest of the file.

ryanofsky (commented):
utACK 76d3e3bb9e2d5fd5e0c570489cc377ce95cc370f. Change looks good. Left some minor comments.

laanwj:
Concept ACK - however some comments:

- This should respect `ENABLE_WALLET`, for compiling without wallet support
- It should respect runtime `-wallet=0`
- Wallet events should not go through `CValidationInterface`, it is for core events only.  To do this correctly, zmq should subscribe to wallet events through `CWallet` directly.

somdoron:
@ryanofsky fixed the type, wrapped the docs and using const uint256 &hashBlock
@laanwj using wallet directly (without going through CValidationInterface)  and respecting  `ENABLE_WALLET`. Respecting `-wallet=0` by checking if `pwalletMain` is not null.

somdoron:
@ryanofsky @laanwj rebased the pull request on top of @jnewbery pull request #10555 and now all tests are passing

sipa:
Does ZMQ have any authentication? I believe originally nothing wallet-related was exposed through it, as there may at least by privacy issues from publishing this.

somdoron:
zeromq does have authentication, but it is not being used within bitcoind.
Do you think it is needed here? my only counter argument is that it should be exposed to internal network and trusted peers only.

But if you feel that it is needed I can add authentication for the wallet publishers.

laanwj:
> Do you think it is needed here? my only counter argument is that it should be exposed to internal network and trusted peers only.

We could just add a warning that the API is unauthenticated, and thus wallet notifications should not be used when the zmq endpoints are accessible to other users - they can be restricted by other means, e.g. binding locally, binding to UNIX socket, firewall, etc.

> But if you feel that it is needed I can add authentication for the wallet publishers.

I'd insist on that only when adding control of the wallet to the ZMQ interface.

sipa:
I have no strong opinion on the need for authentication; I just wanted to bring up that I believed that was the reason for not having wallet specific notifications in ZMQ before.

MarcoFalke:
Needs rebase.

somdoron:
@MarcoFalke rebased

src/zmq/zmqnotificationinterface.cpp
```diff
@@ -110,6 +116,12 @@ bool CZMQNotificationInterface::Initialize()
 void CZMQNotificationInterface::Shutdown()
 {
     LogPrint(BCLog::ZMQ, "zmq: Shutdown notification interface\n");
+
+#ifdef ENABLE_WALLET
+    if (pwalletMain)
```
MarcoFalke: nit: missing braces `{` and `}` for the if block, also this need adjustment for multiwallet.

somdoron: fixed both

somdoron:
@MarcoFalke fixed and rebased

src/wallet/wallet.h
```diff
@@ -1069,6 +1069,9 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
     boost::signals2::signal<void (CWallet *wallet, const uint256 &hashTx,
             ChangeType status)> NotifyTransactionChanged;
 
+    boost::signals2::signal<void (const CTransactionRef &ptxn,
```
ryanofsky: I think you should make this a static variable to make the zmq code simpler and less fragile and remove the dependency on ::vpwallets. Advantages of making this static:

* It will let you get rid of the `ConnectToWalletSignals` method and all the ifdefed code around it and simply connect to the signal in `CZMQNotificationInterface::Initialize` consistent with you how currently disconnect in `CZMQNotificationInterface::Shutdown`
* If will make it possible in the future for new wallets to be added to ::vpwallets at runtime without having to modify  zmq code and make it register for new notifications.

ryanofsky (commented):
utACK 2c393441b8dc55322625670a7aaf37e2df583e29, but I definitely think you should consider making `CWallet::TransactionAddedToWallet` static as described below to make init and shutdown less fragile.

Changes since previous review were moving the signal from validation interface to wallet, passing txids by const reference, wrapping markdown documentation, rebasing the test.

src/wallet/wallet.h
somdoron: thanks, will make the change

ryanofsky (commented):
utACK 123c3e75e96e56a78387353d10dd05b8d73e51c1. Change since last review was a new commit making the wallet signal static.

src/wallet/wallet.cpp
```diff
@@ -56,6 +56,12 @@ CFeeRate CWallet::minTxFee = CFeeRate(DEFAULT_TRANSACTION_MINFEE);
  * Override with -fallbackfee
  */
 CFeeRate CWallet::fallbackFee = CFeeRate(DEFAULT_FALLBACK_FEE);
+/*
+ * Signal when transactions are added to wallet
+ */
+boost::signals2::signal<void (const CTransactionRef &ptxn,
+const uint256 &blockHash)> CWallet::TransactionAddedToWallet = boost::signals2::signal<void (const CTransactionRef &ptxn,
```
somdoron: fixed and rebased last commit. Do you want me to rebase everything into one commit?

ryanofsky: > fixed and rebased last commit. Do you want me to rebase everything into one commit?

No preference, either way seems fine.

ryanofsky (commented):
utACK d358230d10b00ef7e1a284d1b855fa1604bf1d28. Only change since last review was removing unneeded init assignment.

doc/zmq.md
```diff
@@ -75,6 +77,15 @@ notification `-zmqpubhashtx` the topic is `hashtx` (no null
...
 bytes).
 
+For wallet transaction notifications (both hash and tx), the
+topic also indicate if the transaction came from a block
+or mempool. If originated from mempool `-mempool` postfix
+will be added to the topic, for block `-block` postfix will
+be added. Because zeromq is using prefix matching for topics
+you can subscribe to `rawwallettx` (or `hashwallettx`) to get
+both notifications. If you only want one type of notification
+subscribe to either `rawwallettx-mempool` or `rawwallettx-block`.
```
luke-jr: No documentation on how to determine *which* wallet this transaction involves (is it even possible?)

src/wallet/wallet.cpp
```diff
@@ -56,6 +56,10 @@ CFeeRate CWallet::minTxFee = CFeeRate(DEFAULT_TRANSACTION_MINFEE);
  * Override with -fallbackfee
  */
 CFeeRate CWallet::fallbackFee = CFeeRate(DEFAULT_FALLBACK_FEE);
+/*
+ * Signal when transactions are added to wallet
+ */
+boost::signals2::signal<void (const CTransactionRef &ptxn, const uint256 &blockHash)> CWallet::TransactionAddedToWallet;
```
luke-jr: IMO `blockHash` should be `nullptr` when not in a block.

luke-jr: Probably `CWalletTx` should be passed instead of `CTransactionRef`?

doc/zmq.md
somdoron: It was actually submitted before the multi-wallet was merged. Wallet name (or identifier) can be part of the topic.

src/wallet/wallet.cpp
somdoron: actually if CWalletTX is used the blockhash is not needed anymore. I will take a look.

ryanofsky:
@somdoron are you still working on this? IIUC, luke's improvements could be implemented later without breaking backwards compatibility, if they are what's holding this up.

somdoron:
@ryanofsky yes, will rebase today and send a PR

somdoron:
@ryanofsky rebased and all tests passed. Doesn't yet include @luke-jr comments.

src/zmq/zmqpublishnotifier.cpp
```diff
@@ -166,6 +170,23 @@ bool CZMQPublishHashTransactionNotifier::NotifyTransaction(const CTransaction &t
     return SendMessage(MSG_HASHTX, data, 32);
 }
 
+bool CZMQPublishHashWalletTransactionNotifier::NotifyWalletTransaction(const CTransaction &transaction, const uint256 &hashBlock){
+    uint256 hash = transaction.GetHash();
+    LogPrint(BCLog::ZMQ, "zmq: Publish hashwallettx %s\n", hash.GetHex());
+    char data[32];
+    for (unsigned int i = 0; i < 32; i++)
```
jonasschnelli: memcpy?

jonasschnelli: Nevermind. I see we do reverse the byte order here...

jonasschnelli:
I'm still unsure about this, if wtxs (protected by http auth) should be something we broadcast via ZMQ.
Alternative would be HTTP based long polling (#7949).

Also, how does this handle multiwallet? Should we somehow integrate the wallet identifier in the notifications?

src/zmq/zmqnotificationinterface.cpp
```diff
@@ -180,3 +195,22 @@ void CZMQNotificationInterface::BlockDisconnected(const std::shared_ptr<const CB
...
     }
 }
+
+void CZMQNotificationInterface::TransactionAddedToWallet(const CTransactionRef& ptx, const uint256 &hashBlock) {
+    const CTransaction& tx = *ptx;
+
+    for (std::list<CZMQAbstractNotifier*>::iterator i = notifiers.begin(); i!=notifiers.end(); )
+    {
+        CZMQAbstractNotifier *notifier = *i;
+        if (notifier->NotifyWalletTransaction(tx, hashBlock))
```
jnewbery: style nit: braces on same line please.

src/zmq/zmqnotificationinterface.cpp
```diff
@@ -180,3 +195,22 @@ void CZMQNotificationInterface::BlockDisconnected(const std::shared_ptr<const CB
...
+        {
+            i++;
+        }
+        else
```
jnewbery: style nit:

`} else {`

src/zmq/zmqpublishnotifier.cpp
```diff
@@ -166,6 +170,23 @@ bool CZMQPublishHashTransactionNotifier::NotifyTransaction(const CTransaction &t
...
+        data[31 - i] = hash.begin()[i];
+
+    const char *command;
+
+    if (!hashBlock.IsNull())
```
jnewbery: nit: braces please.

src/zmq/zmqpublishnotifier.cpp
jnewbery: nit: braces please

src/zmq/zmqabstractnotifier.cpp
```diff
@@ -20,3 +20,7 @@ bool CZMQAbstractNotifier::NotifyTransaction(const CTransaction &/*transaction*/
 {
     return true;
 }
+
+bool CZMQAbstractNotifier::NotifyWalletTransaction(const CTransaction &transaction, const uint256 &hashBlock){
```
jnewbery: nit: function brace on newline please.

test/functional/zmq_test.py
```diff
@@ -22,16 +23,17 @@ def __init__(self, socket, topic):
         import zmq
         self.socket.setsockopt(zmq.SUBSCRIBE, self.topic)
 
-    def receive(self):
+    def receive(self, specific_topic = None):
```
jnewbery: nit: no spaces around the `=` for named arguments.

test/functional/test_framework/util.py
```diff
@@ -37,6 +37,10 @@ def assert_equal(thing1, thing2, *args):
     if thing1 != thing2 or any(thing1 != arg for arg in args):
         raise AssertionError("not(%s)" % " == ".join(str(arg) for arg in (thing1, thing2) + args))
 
+def assert_not_equal(thing1, thing2):
```
jnewbery: You've added this but not used it. If it's not being used, can you remove it from this PR?

jnewbery (commented):
Apologies for all the style nits. Please see https://github.com/bitcoin/bitcoin/blob/6157e8ce3937af3f46d3e7dd922d19d6dc272145/doc/developer-notes.md for the full style guide.

I agree with the other reviewers that this needs a bit more work for multiwallet. Can you either:
1. add documentation saying that multiwallet is supported but that notifications will not indicate which wallet the notification is for; or
2. add the wallet name to the notification topic

Both are fine. You could do (1) now and a future PR could add the name of the wallet to the notification.

Can you also update the zmq_test.py to verify that this notifications are received for all wallets when multiwallet is being used? I have a branch here that does that: https://github.com/jnewbery/bitcoin/tree/pr10554.1 . Feel free to take that and squash into your commit if you like it.

Also, I think this could use some extra documentation warning users about security implications:
- zmq is unauthenticated
- notifications are received for all wallets and can't be enabled/disabled on a per-wallet basis.

ryanofsky (commented):
utACK ed4fd266f78f9fea8dd0e4cb94553e55171fb7d2. Only change since last review is rebase. Agree with John it'd be good to document the multiwallet limitation if you don't want to add wallet names to the notifications right now. Also I don't see any reason not to squash the two commits. Would make understanding the PR a little simpler.

ryanofsky:
For some reason travis is failing with `contrib/devtools/lint-python.sh: 10: contrib/devtools/lint-python.sh: flake8: not found`. It's possible a rebase might fix this.

promag:
> Alternative would be HTTP based long polling (#7949).

I tend to agree with @jonasschnelli alternative.

breitsmiley:
What about this PR? Is it going to be added?

jnewbery:
No update from PR author @somdoron since October 2017. Are you still working on this @somdoron ?

If not, I suggest we close this as 'Up for grabs'

somdoron:
You can close.

I actually used the code in production as part of payments solution.

PR #10552, opened 2017-06-08
[Test] Tests for zmqpubrawtx and zmqpubrawblock

achow101:

PR #10555, opened 2017-06-08
[tests] various improvements to zmq_test.py

jnewbery:
This PR contains various improvements to zmq_test.py, including stopping it from hanging and causing travis builds to time out. Individual commits are:

- update zmq test to use correct config.ini file - follows #10331, which moved the config.ini file from /bitcoin/test/functional to /bitcoin/test. zmq_test.py was not updated to find the config file in the correct location. This isn't a problem when running through test_runner.py, but would prevent zmq_test.py from being run directly.
- tidy up zmq_test.py - general code tidy-up. Cuts run time from ~12s to ~6s
- in zmq test, timeout if message not received - stops a race condition where we could try to receive a message on the zmq socket before it's been sent
- destroy zmq context in zmq_tests.py - fixes an issue where the  zmq contest would not be close properly at the end of the test, causing `sys.exit()` to fail. This would cause travis builds to hang. See https://stackoverflow.com/questions/17140417/termination-of-python-script-while-using-zeromq-with-dead-server for more details.
- timeout integration tests on travis after 20 minutes - causes travis builds to timeout individual tests if they run for more than 20 minutes. That means we'll get logs from hung tests, rather than travis itself timing out after ~50 minutes (in which case we don't get any useful diagnostics).

I need this for the next step of #10082, since this test was causing the next PR to hang in travis.

I believe this will also fix #10552 once that is rebased on top of this.

### DIFF
--- a/test/functional/zmq_test.py
+++ b/test/functional/zmq_test.py
@@ -8,15 +8,15 @@ import os
 import struct
 
 from test_framework.test_framework import BitcoinTestFramework, SkipTest
-from test_framework.util import *
+from test_framework.util import (assert_equal,
+                                 bytes_to_hex_str,
+                                 )
 
 class ZMQTest (BitcoinTestFramework):
 
     def __init__(self):
         super().__init__()
-        self.num_nodes = 4
-
-    port = 28332
+        self.num_nodes = 2
 
     def setup_nodes(self):
         # Try to import python3-zmq. Skip this test if the import fails.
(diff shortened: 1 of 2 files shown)


achow101:
utACK

jimmysong (approved):
tested ACK c4f6b7f2bc540c368e2ba1a5ab332432b0e93889

jnewbery:
Oops. Accidentally squashed two commits together in my last fixup. I've divided this PR back into the correct number of commits.

https://github.com/jnewbery/bitcoin/tree/pr10555.1 == https://github.com/jnewbery/bitcoin/tree/pr10555.2

test/functional/zmq_test.py
```diff
@@ -38,72 +52,71 @@ def setup_nodes(self):
...
+        assert_equal(topic, b"hashtx")
+        hashZMQ = bytes_to_hex_str(body)
+        msgSequence = struct.unpack('<I', msg[-1])[-1]
+        assert_equal(msgSequence, blockcount + 1)
 
-        assert_equal(hashRPC, hashZMQ) #blockhash from generate must be equal to the hash received over zmq
+        assert_equal(hashRPC, hashZMQ)  # txid from sendtoaddress must be equal to the hash received over zmq
 
+        # Destroy the zmq context
+        self.zmqContext.destroy(linger=None)
```
somdoron: when test failed this won't be called and the process will hang, might be good to call this in finally block

jnewbery: Good point. Added a fixup commit 6cdfd7020915356bd82c1f538c0b7817665fda20. Can you take a look and let me know what you think. If you're happy I'll squash with the previous commit.

somdoron: looks good

test/functional/zmq_test.py
```diff
@@ -6,29 +6,43 @@
...
+import time
 
 from test_framework.test_framework import BitcoinTestFramework, SkipTest
-from test_framework.util import *
+from test_framework.util import (assert_equal,
+                                 bytes_to_hex_str,
+                                 )
+
+def wait_for_multipart(socket, timeout=60):
+    interval = 0.1
```
somdoron: You can also call socket.SET(zmq.RCVTIMEO, 60000) when creating the socket instead of the NOBLOCK and the loop. It will raise zmq.error.Again when time is over

jnewbery: Thanks! I didn't know about `zmq.RCVTIMEO`. Updated to use that instead of the timeout loop in https://github.com/bitcoin/bitcoin/pull/10555/commits/b1bac1cb7ea1560b46ff758864660a21a228ed67

somdoron (approved):

MarcoFalke:
Concept ACK 0a4912e46aecef6f04c6287508020ec60424ed24

PR #10556, opened 2017-06-08
Move stop/start functions from utils.py into BitcoinTestFramework

jnewbery:
Final preparation step in #10082 before introducing the TestNode class.

This builds on top of #10555. The first 5 commits are from that PR. Without those commits, the travis build times out.

This PR moves the stop/start functions from utils.py into BitcoinTestFramework. It also moves stateful functions/variables from utils.py into BitcoinTestFramework (coverage variables, bitcoind_processes dict, mocktime functions). It also does some general tidyup of test_framework.py and utils.py.

This touches a few individual test cases, but those changes are very minor. The important changes are in util.py and test_framework.py. The final commit is a code move only.

### DIFF
--- a/test/functional/bip9-softforks.py
+++ b/test/functional/bip9-softforks.py
@@ -15,6 +15,10 @@ mine a further 143 blocks (LOCKED_IN)
 test that enforcement has not triggered (which triggers ACTIVE)
 test that enforcement has triggered
 """
+from io import BytesIO
+import shutil
+import time
+import itertools
 
 from test_framework.test_framework import ComparisonTestFramework
 from test_framework.util import *
@@ -22,9 +26,6 @@ from test_framework.mininode import CTransaction, NetworkThread
 from test_framework.blocktools import create_coinbase, create_block
 from test_framework.comptool import TestInstance, TestManager
 from test_framework.script import CScript, OP_1NEGATE, OP_CHECKSEQUENCEVERIFY, OP_DROP
-from io import BytesIO
-import time
-import itertools
 
 class BIP9SoftForksTest(ComparisonTestFramework):
 
--- a/test/functional/fundrawtransaction.py
+++ b/test/functional/fundrawtransaction.py
@@ -4,7 +4,7 @@
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 """Test the fundrawtransaction RPC."""
 
-from test_framework.test_framework import BitcoinTestFramework
+from test_framework.test_framework import BitcoinTestFramework, BITCOIND_PROC_WAIT_TIMEOUT
 from test_framework.util import *
 
 
@@ -452,7 +452,7 @@ class RawTransactionsTest(BitcoinTestFramework):
         self.stop_node(2)
         self.stop_node(3)
         self.nodes[1].encryptwallet("test")
-        bitcoind_processes[1].wait(timeout=BITCOIND_PROC_WAIT_TIMEOUT)
+        self.bitcoind_processes[1].wait(timeout=BITCOIND_PROC_WAIT_TIMEOUT)
 
         self.nodes = self.start_nodes(self.num_nodes, self.options.tmpdir)
         # This test is not meant to test fee estimation and we'd like
--- a/test/functional/test_framework/test_framework.py
+++ b/test/functional/test_framework/test_framework.py
@@ -26,10 +26,7 @@ from .util import (
     assert_equal,
     check_json_precision,
     connect_nodes_bi,
-    disable_mocktime,
     disconnect_nodes,
-    enable_mocktime,
-    get_mocktime,
     get_rpc_proxy,
     initialize_datadir,
     get_datadir_path,
@@ -73,6 +70,7 @@ class BitcoinTestFramework(object):
         self.setup_clean_chain = False
         self.nodes = []
         self.bitcoind_processes = {}
+        self.mocktime = 0
 
     def add_options(self, parser):
         pass
@@ -211,7 +209,7 @@ class BitcoinTestFramework(object):
         datadir = os.path.join(dirname, "node" + str(i))
         if binary is None:
             binary = os.getenv("BITCOIND", "bitcoind")
-        args = [binary, "-datadir=" + datadir, "-server", "-keypool=1", "-discover=0", "-rest", "-logtimemicros", "-debug", "-debugexclude=libevent", "-debugexclude=leveldb", "-mocktime=" + str(get_mocktime()), "-uacomment=testnode%d" % i]
+        args = [binary, "-datadir=" + datadir, "-server", "-keypool=1", "-discover=0", "-rest", "-logtimemicros", "-debug", "-debugexclude=libevent", "-debugexclude=leveldb", "-mocktime=" + str(self.mocktime), "-uacomment=testnode%d" % i]
         if extra_args is not None:
             args.extend(extra_args)
         self.bitcoind_processes[i] = subprocess.Popen(args, stderr=stderr)
@@ -312,6 +310,21 @@ class BitcoinTestFramework(object):
             sync_blocks(group)
             sync_mempools(group)
 
+    def enable_mocktime(self):
+        """Enable mocktime for the script.
+
+        mocktime may be needed for scripts that use the cached version of the
+        blockchain.  If the cached version of the blockchain is used without
+        mocktime then the mempools will not sync due to IBD.
+
+        For backwared compatibility of the python scripts with previous
+        versions of the cache, this helper function sets mocktime to Jan 1,
+        2014 + (201 * 10 * 60)"""
+        self.mocktime = 1388534400 + (201 * 10 * 60)
+
+    def disable_mocktime(self):
+        self.mocktime = 0
+
     # Private helper methods. These should not be accessed by the subclass test scripts.
 
     def _start_logging(self):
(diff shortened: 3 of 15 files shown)


jnewbery:
@jimmysong @kallewoof @sdaftuar - you've all mentioned that you're happy to help move #10082 forward. Any review of this and/or #10555 would be appreciated.

test/functional/zmq_test.py
```diff
@@ -38,57 +38,55 @@ def setup_nodes(self):
...
             if topic == b"hashblock":
                 zmqHashes.append(bytes_to_hex_str(body))
                 msgSequence = struct.unpack('<I', msg[-1])[-1]
-                assert_equal(msgSequence, blockcount+1)
+                assert_equal(msgSequence, blockcount + 1)
                 blockcount += 1
 
-        for x in range(0,n):
-            assert_equal(genhashes[x], zmqHashes[x]) #blockhash from generate must be equal to the hash received over zmq
+        for x in range(0, n):
```
jimmysong: nit: `range(n)` is better than `range(0, n)`

test/functional/zmq_test.py
```diff
@@ -6,12 +6,23 @@
...
 import struct
+import time
 
 from test_framework.test_framework import BitcoinTestFramework, SkipTest
 from test_framework.util import (assert_equal,
                                  bytes_to_hex_str,
                                  )
 
+def wait_for_multipart(socket, timeout=60):
+    while True:
```
jimmysong: This might be easier to read if made into a for loop:

    interval = 0.1
    for _ in range(int(timeout/interval)):
        try:
            return socket.recv_multipart(zmq.NOBLOCK)
        except zmq.error.Again:
            time.sleep(interval)
    raise AssertionError("Timed out waiting for zmq message")

test/functional/fundrawtransaction.py
```diff
@@ -452,7 +452,7 @@ def run_test(self):
         self.stop_node(2)
         self.stop_node(3)
         self.nodes[1].encryptwallet("test")
-        bitcoind_processes[1].wait(timeout=BITCOIND_PROC_WAIT_TIMEOUT)
+        self.bitcoind_processes[1].wait(60)
```
jimmysong: Why take out the constant here?

test/functional/test_framework/test_framework.py
```diff
@@ -14,6 +14,7 @@
 import sys
 import tempfile
 import time
+import traceback
```
jimmysong: Is this being used?

test/functional/bip9-softforks.py
```diff
@@ -15,16 +15,17 @@
 test that enforcement has not triggered (which triggers ACTIVE)
 test that enforcement has triggered
 """
+from io import BytesIO
+import shutil
```
jimmysong: How did this work before? Was it one of the import *'s?

jimmysong (commented):
utAck f7acc7adea731589aec8a7d9e16cc9069e24b5a9

Added a few nits and questions. Testing next.

test/functional/zmq_test.py
jnewbery: Yes, that's better. I've made the change in #10555, and will rebase this PR off that once you've finished your review.

test/functional/zmq_test.py
jnewbery: Agree. Changed in #10555

test/functional/fundrawtransaction.py
jnewbery: Because I've moved the `BITCOIND_PROC_WAIT_TIMEOUT` constant into test_framework.py. The next PR (introduce TestNode) will remove `bitcoind_processes` dict entirely, so for this intermediate PR rather than add an import which I'll revert in the next PR, I've just changed this to a hard-coded value.

test/functional/test_framework/test_framework.py
jnewbery: Here: https://github.com/bitcoin/bitcoin/blob/9c248e39f2807a7f89e555e99cc55cb2de1ad335/test/functional/test_framework/test_framework.py#L194

I must have missed out this import in a previous commit. Makes sense to fix it in this cleanup commit.

test/functional/bip9-softforks.py
jnewbery: Yes, this was being indirectly imported from utils.py :( there are still lots of indirect imports in functional tests, which makes refactors like this seem more disruptive than they actually are.

jnewbery:
@jimmysong - Thanks for the quick review! Responses inline.

jimmysong:
Great, please let me know when you rebase to #10555 and I'll test it.

jnewbery:
rebased on #10555

jimmysong:
Tested ACK e1e3dd074b71503d8f74651bf2a3cfd9e55a4d2b

achow101:
utACK e1e3dd074b71503d8f74651bf2a3cfd9e55a4d2b

test/functional/test_framework/test_framework.py
```diff
@@ -207,16 +203,82 @@ def main(self):
...
+            binary = [None] * num_nodes
+        assert_equal(len(extra_args), num_nodes)
+        assert_equal(len(binary), num_nodes)
+        rpcs = []
+        try:
+            for i in range(num_nodes):
+                rpcs.append(self.start_node(i, dirname, extra_args[i], rpchost, timewait=timewait, binary=binary[i]))
+        except:
+            # If one node failed to start, stop the others
+            self.stop_nodes()
```
kallewoof: I have a hard time understanding how this will work. `self.nodes` seems to be set to `self.start_nodes(...)` which means at this point, `self.nodes` is not set at all, and `self.stop_nodes()` will do nothing. Am I missing something?

kallewoof (approved):
Tested ACK e1e3dd0

test/functional/test_framework/test_framework.py
achow101: I don't think you are. Instead of returning `rpcs`, I think this function should appent to `self.nodes` directly.

jnewbery: Good spot @kallewoof - you're not missing anything at all. This is obviously wrong :(

Once #10082 is fully merged, I'd like to improve the way the TestNode objects are handled by the BitcoinTestFramework class. There should be separate methods `add_nodes()`, which would instantiate new TestNode objects, and `start_node()`, which would start the bitcoind process. I don't think `start_node()` or `start_nodes()` should be responsible for updating the `self.nodes` list, and individual tests shouldn't normally have to concern themselves with updating the list.

I also think that eventually `self.nodes` should be a dictionary.

For now, I've added a fixup commit that appends the newly created nodes to self.nodes, then calls self.stop_nodes(), and then clears the list. It's a little hacky, but it should restore the functionality here until #10082 is fully merged. Can you take a look and let me know what you think?

kallewoof: @jnewbery Fix looks good to me!

jnewbery:
Needed rebase because the new wallet-encryption test used functions that are moved into the BitcoinTestFramework class by the PR.

test/functional/fundrawtransaction.py
MarcoFalke: Please keep the arg named. I'd really prefer if all primitive types are named in our python code.

jnewbery: Done. I've also restored the use of a constant here (at least until the next PR in this series is merged).

jnewbery:
Rebased

test/functional/fundrawtransaction.py
MarcoFalke: Are you sure this was fixed? GitHub didn't collapse the comments.

jnewbery: Oops. Fixed but not pushed.

Should be good now.

laanwj:
Needs rebase after #10533

test/functional/test_framework/test_framework.py
```diff
@@ -187,7 +183,7 @@ def main(self):
                 for fn in filenames:
                     try:
                         with open(fn, 'r') as f:
-                            print("From" , fn, ":")
+                            print("From", fn, ":")
```
MarcoFalke: Not sure how to feel about the whitespace changes. This might discourage or distract reviewers. Maybe we could make those in a separate somewhat larger pull?

jnewbery:
rebased (twice - since master changed while I was rebasing 🙁 )

> Not sure how to feel about the whitespace changes. This might discourage or distract reviewers.

I include style fixes for the modules I'm touching in PRs for a few reasons:
- I like my commits to run clean through a linter. It's the best way I know to make sure I'm not introducing new style nits.
- no-one likes reviewing style-change-only PRs, so unless they're included in larger PRs, style tends to degrade over time
- it's less disruptive than project-wide style PRs in terms of rebasing, since the module is being updated anyway.

It doesn't appear that this has discouraged reviewers - this PR has three ACKs and a comment from another reviewer (you).

The style changes are contained within separate commits so if you feel really strongly about this I can remove them. However, I'd prefer not to since I've already rebased this PR five times.

MarcoFalke:
Sorry, this needs rebase again. :(

jnewbery:
rebased

MarcoFalke:
Well, the issue with including whitespace fixes is that it makes it harder
to backport and also harder to keep mergeable. So the rebases will
introduce errors (see below). Actually I would be fine with a cleanup in
test that fixes whitespace and is a `git diff --ignore-all-space` empty
diff.

utACK with 05b8c081b435e87b08335e9f9b62a55fa1d48ecc, lets leave this as is
and fix two nits:

* rpc_auth_pair does not exist previously and is never used, please remove
* You want to move the `assert_equal(return_code, 0)` down by one line,
such that the process is deleted even if the return code does not match
* wait_for_node_exit should probably come with a default value for timeout,
but this should be done i a follow a later pull.

jnewbery:
Fixed the first two nits. Thanks for catching

> wait_for_node_exit should probably come with a default value for timeout,
but this should be done i a follow a later pull.

Right - I think that's completely orthogonal to this pull. It was introduced in 176c021d085f5a45bc9e038e760942aa648dd797

PR #10560, opened 2017-06-08
Remove unused constants

practicalswift:
* Remove unused constant MEMPOOL_GD_VERSION
* Comment out unused constant REJECT_DUST

### DIFF
--- a/src/consensus/validation.h
+++ b/src/consensus/validation.h
@@ -14,7 +14,7 @@ static const unsigned char REJECT_INVALID = 0x10;
 static const unsigned char REJECT_OBSOLETE = 0x11;
 static const unsigned char REJECT_DUPLICATE = 0x12;
 static const unsigned char REJECT_NONSTANDARD = 0x40;
-static const unsigned char REJECT_DUST = 0x41;
+// static const unsigned char REJECT_DUST = 0x41; // part of BIP 61
 static const unsigned char REJECT_INSUFFICIENTFEE = 0x42;
 static const unsigned char REJECT_CHECKPOINT = 0x43;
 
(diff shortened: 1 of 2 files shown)


achow101:
NACK `-testsafemode` is not unused. It is part of GetWarnings and will set a warning for strStatusBar, strRPC, and strGUI which will then trigger the safe mode handling of RPCs. Safe mode is still used in the RPCs to disable certain RPCs if safe mode is enabled.

practicalswift:
@achow101 Ah, sorry. I missed that the safe mode logic in `OnRPCPreCommand` is triggered by `strRPC != nullptr` (via `GetWarnings("rpc")` :-)  Thanks for clarifying. Reverting that part of this PR.

practicalswift:
@achow101 Updated :-)

src/consensus/validation.h
```diff
@@ -14,7 +14,7 @@ static const unsigned char REJECT_INVALID = 0x10;
 static const unsigned char REJECT_OBSOLETE = 0x11;
 static const unsigned char REJECT_DUPLICATE = 0x12;
 static const unsigned char REJECT_NONSTANDARD = 0x40;
-static const unsigned char REJECT_DUST = 0x41;
+// static const unsigned char REJECT_DUST = 0x41;
```
achow101: Since this is part of BIP 61, I think it should remain as is even though it is unused.

practicalswift: @achow101 Yes, I left it commented out instead of removing it :-)

practicalswift:
@achow101 Thanks for reviewing! Added a comment about BIP 61. Looks good?

sipa:
utACK dce82397c1540769a8b39a9a475c366cb9438469

achow101:
utACK dce82397c1540769a8b39a9a475c366cb9438469

fanquake:
utACK dce8239

PR #10565, opened 2017-06-09
[coverage] Remove subtrees and benchmarks from coverage report

achow101:
Removes stats about leveldb, univalue, and benchmark coverage from the coverage reports. Since we don't test these three items specifically and thus don't care about test coverage for these items.

This builds on top of #10511 to use the faster `lcov -r` replacement.

### DIFF
--- a/Makefile.am
+++ b/Makefile.am
@@ -171,13 +171,15 @@ baseline.info:
 	$(LCOV) -c -i -d $(abs_builddir)/src -o $@
 
 baseline_filtered.info: baseline.info
-	$(LCOV) -r $< "/usr/include/*" $(LCOV_OPTS) -o $@
+	$(abs_builddir)/contrib/filter-lcov.py "/usr/include/" $< $@
+	$(LCOV) -a $@ $(LCOV_OPTS) -o $@
 
 leveldb_baseline.info: baseline_filtered.info
 	$(LCOV) -c -i -d $(abs_builddir)/src/leveldb -b $(abs_builddir)/src/leveldb -o $@
 
 leveldb_baseline_filtered.info: leveldb_baseline.info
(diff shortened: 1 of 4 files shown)


sipa:
Perhaps remove all the subtrees... including src/crypto/ctaes and src/secp256k1?

achow101:
Removed ctaes and secp256k1 too.

contrib/filter-lcov.py
```diff
@@ -0,0 +1,25 @@
...
+outfile = args.outfile
+
+in_remove = False
+with open(tracefile, 'r') as f:
+    with open(outfile, 'w') as wf:
+        for line in f:
+            for p in pattern:
+                if line.startswith("SF:") and p in line:
+                    in_remove = True
+                if not in_remove:
```
sipa: This line should probably not be inside the `for p in pattern:` loop, as it may result in multiple writes of the same input line.

laanwj:
Concept ACK

Makefile.am
```diff
@@ -171,44 +171,32 @@ baseline.info:
 	$(LCOV) -c -i -d $(abs_builddir)/src -o $@
 
 baseline_filtered.info: baseline.info
-	$(abs_builddir)/contrib/filter-lcov.py "/usr/include/" $< $@
+	$(abs_builddir)/contrib/filter-lcov.py -p "/usr/include/" -p "src/leveldb/" -p "src/bench/" -p "src/univalue" -p "src/crypto/ctaes" -p "src/secp256k1" $< $@
```
sipa: Perhaps abstract out the list of things to filter into a variable, as it's used more than once?

achow101: done

sipa:
Tested ACK d5711f4a2d59adc45755b13e3776b9d36e1c55f5

PR #10566, opened 2017-06-09
[docs] Use the "domain name setup" image (previously unused) in the gitian docs

practicalswift:
Prior to this commit the file `debian_install_6_domain_name.png` was unused.


fanquake:
utACK 7631066

practicalswift:
Wow, a short hash with only digits!

practicalswift:
A 3.725 % probably for such a short hash (before the fact, obviously – 100 % after the fact) :-)

PR #10569, opened 2017-06-09
Fix stopatheight

achow101:
Moves stopatheight check into the ActivateBestChain loop so that the block height is actually checked and stopped at the appropriate time.

Noticed and reported here: https://github.com/bitcoin/bitcoin/pull/10339#issuecomment-307161260

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -2294,6 +2294,7 @@ bool ActivateBestChain(CValidationState &state, const CChainParams& chainparams,
 
     CBlockIndex *pindexMostWork = NULL;
     CBlockIndex *pindexNewTip = NULL;
+    int nStopAtHeight = GetArg("-stopatheight", DEFAULT_STOPATHEIGHT);
     do {
         boost::this_thread::interruption_point();
         if (ShutdownRequested())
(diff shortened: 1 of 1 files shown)


luke-jr (approved):
utACK

src/validation.cpp
```diff
@@ -2343,6 +2343,9 @@ bool ActivateBestChain(CValidationState &state, const CChainParams& chainparams,
         if (pindexFork != pindexNewTip) {
             uiInterface.NotifyBlockTip(fInitialDownload, pindexNewTip);
         }
+
+        int nStopAtHeight = GetArg("-stopatheight", DEFAULT_STOPATHEIGHT);
```
sipa: Can you move this line out of the loop? No need to parse the command line argument for every block.

luke-jr:
What's up with the random optimisation?

achow101:
sorry, checked out from the wrong worktree

sipa:
utACK c45cbaf69f4fe6c8dabd48205da3bf302ed99c8a

paveljanik:
utACK c45cbaf

laanwj:
I thought I was doing something wrong :$
Thanks for fixing this.
Tested ACK c45cbaf
```
$ src/bitcoind -datadir=/store/tmp/testbtc -reindex-chainstate -stopatheight=200000 -printtoconsole
2017-06-11 12:00:02 UpdateTip: new best=000000000000034a7dedef4a161fa058a2d67a173a90155f3a2fe6fc132e0ebf height=200000 version=0x00000002 log2_work=68.741562 tx=7316696 date='2012-09-22 10:45:59' progress=0.032039 cache=359.0MiB(2636343txo)
...
2017-06-11 12:00:18 Shutdown: done
```
As `stopatheight` is not in any releases this doesn't need backport.

Issue #10570, opened 2017-06-10
[RPC] Split signrawtransaction into multiple distinct RPCs


achow101:
The current signrawtransaction rpc command is kind of a mess. It contains both wallet code and non-wallet code as well as an undocumented feature to combine multiple transactions. This issue is to track progress on this and to get some commentary on this split.

The goal of this is to result in the following (and probably in this order):
- [x] Split combination of transactions into a new RPC that also handles the combination (#10571)
- [x] Split `signrawtransaction` into wallet and non-wallet RPCs. (#10579)
- [x] Mark `signrawtransaction` as deprecated (#10579)
- [x] Remove `signrawtransaction` (#12490)

sipa:
One idea would be to combine this with a new partial transaction format (using header byte 0xff rather than 0x00?), which can contain redundant information:
* The UTXOs being spent (so that the signer doesn't need access to the UTXO set)
* The redeemscript in case of P2SH/witness outputs.
* Which output(s) to send change to (before it's known how much that change will be).
* ...

achow101:
@sipa that would be out of scope for this. Such a format could be added later. This issue is really only for the cleanup of the mess that is `signrawtransaction`

achow101:
Closing now that #12490 has been merged

PR #10571, opened 2017-06-10
[RPC]Move transaction combining from signrawtransaction to new RPC

achow101:
Create a combinerawtransaction RPC which accepts a json array of hex raw transactions to combine them into one transaction. Signrawtransaction is changed to no longer combine transactions and only accept one transaction at a time.

The tests have been updated to test this. Tests for the signrawtransaction merge have also been removed.

This is part of #10570

### DIFF
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -554,6 +554,93 @@ static void TxInErrorToJSON(const CTxIn& txin, UniValue& vErrorsRet, const std::
     vErrorsRet.push_back(entry);
 }
 
+UniValue combinerawtransaction(const JSONRPCRequest& request)
+{
+
+    if (request.fHelp || request.params.size() != 1)
+        throw std::runtime_error(
+            "combinerawtransaction [\"hexstring\",...]\n"
+            "\nCombine multiple partially signed transactions into one transaction.\n"
+            "The combined transaction may be another partially signed transaction or a \n"
+            "fully signed transaction."
+
+            "\nArguments:\n"
+            "1. \"txs\"         (string) A json array of hex strings of partially signed transactions\n"
+            "    [\n"
+            "      \"hexstring\"     (string) A transaction hash\n"
+            "      ,...\n"
+            "    ]\n"
+
+            "\nResult:\n"
+            "\"hex\" : \"value\",           (string) The hex-encoded raw transaction with signature(s)\n"
+
+            "\nExamples:\n"
+            + HelpExampleCli("combinerawtransaction", "[\"myhex1\", \"myhex2\", \"myhex3\"]")
+        );
+
+
+    UniValue txs = request.params[0].get_array();
+    std::vector<CMutableTransaction> txVariants(txs.size());
+
+    for (unsigned int idx = 0; idx < txs.size(); idx++) {
+        if (!DecodeHexTx(txVariants[idx], txs[idx].get_str(), true)) {
+            throw JSONRPCError(RPC_DESERIALIZATION_ERROR, strprintf("TX decode failed for tx %d", idx));
+        }
+    }
+
+    if (txVariants.empty()) {
+        throw JSONRPCError(RPC_DESERIALIZATION_ERROR, "Missing transactions");
+    }
+
+    // mergedTx will end up with all the signatures; it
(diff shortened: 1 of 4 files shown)


sipa:
Concept ACK. However, this does break backward compatibility with an undocumented but apparently intentional and tested feature. Does this require deprecation first?

jnewbery:
Concept feedback: combining transactions seems like a utility. Is it possible to add it to bitcoin-tx rather than add a utility-only RPC?

sipa:
It is not a utility-only RPC. It needs access to the UTXO set.

sipa:
utACK c710eb9eaffa50daddfedb50a93036085b3296c7

gmaxwell (approved):
ACK

test/functional/rawtransactions.py
```diff
@@ -124,6 +124,57 @@ def run_test(self):
...
+
+        txId = self.nodes[0].sendtoaddress(mSigObj, 2.2)
+        decTx = self.nodes[0].gettransaction(txId)
+        rawTx2 = self.nodes[0].decoderawtransaction(decTx['hex'])
+        self.sync_all()
+        self.nodes[0].generate(1)
+        self.sync_all()
+
+        #THIS IS A INCOMPLETE FEATURE
+        #NODE2 HAS TWO OF THREE KEY AND THE FUNDS SHOULD BE SPENDABLE AND COUNT AT BALANCE CALCULATION
```
morcos: This is copy-paste error leading to incorrect documentation of the test.  This test is a 2of2 multisig for which node 2 has only 1 key.

morcos (approved):
utACK c710eb9

although at least the documentation in the RPC test should probably be corrected

achow101:
Nits addressed and rebased.

morcos:
@achow101 It's better if you address the nits separately from rebasing (was rebasing needed?) otherwise its hard to review just the differential and reACK.

test/functional/rawtransactions.py
```diff
@@ -124,6 +124,57 @@ def run_test(self):
...
+
+        txId = self.nodes[0].sendtoaddress(mSigObj, 2.2)
+        decTx = self.nodes[0].gettransaction(txId)
+        rawTx2 = self.nodes[0].decoderawtransaction(decTx['hex'])
+        self.sync_all()
+        self.nodes[0].generate(1)
+        self.sync_all()
+
+        #THIS IS A INCOMPLETE FEATURE
+        #NODE2 HAS ONE OF TWO KEY AND THE FUNDS SHOULD BE SPENDABLE AND COUNT AT BALANCE CALCULATION
```
morcos: This comment is still incorrect.  The funds should not be spendable there is nothing incomplete here.

achow101:
@morcos sorry, I rebased out of habit.

achow101:
Now the comment should be right.

morcos:
heh thanks... 
and i think its ok to squash, especially if the old commit can still be found, b/c then a diff btwn old and new can show the differences.  it's the rebase that makes it difficult

utACK f293282

morcos:
utACK 3cd7a5b

sipa:
utACK 3cd7a5bc31e1a3cdd1cc740a884f69d9c6689e2f

src/rpc/rawtransaction.cpp
```diff
@@ -554,6 +554,93 @@ static void TxInErrorToJSON(const CTxIn& txin, UniValue& vErrorsRet, const std::
...
+            + HelpExampleCli("combinerawtransaction", "[\"myhex1\", \"myhex2\", \"myhex3\"]")
+        );
+
+
+    UniValue txs = request.params[0].get_array();
+    std::vector<CMutableTransaction> txVariants(txs.size());
+
+    for (unsigned int idx = 0; idx < txs.size(); idx++) {
+        if (!DecodeHexTx(txVariants[idx], txs[idx].get_str(), true)) {
+            throw JSONRPCError(RPC_DESERIALIZATION_ERROR, "TX decode failed for tx " + idx);
```
laanwj: This won't work, you're adding a string and an integer. Use sprintf?

achow101: Fixed

PR #10573, opened 2017-06-11
[WIP, far from finished] Remove unnecessary includes

practicalswift:
Remove unnecessary includes.

_WIP: This PR is *far* from finished. Submitting this version to trigger a Travis CI build to get testing on multiple platforms. Let's discuss which of these includes that should still be kept around despite being "technically redundant" and which we really want to remove._


sipa:
NACK.

You're removing includes of direct dependencies because they're already included indirectly. That results in fragile code that breaks when indirect dependencies change.

In general, a .cpp and .h file should `#include` every header that it directly uses function/classes/definitions from, even if that's already implied by other dependencies. One exception is that a .cpp file does not need to re-include things that its corresponding .h already included.

practicalswift:
This PR is far from complete - it is a rough first version. Updated PR description.

I'll try to find the subset of these includes that are not only "technically redundant" but also unnecessary in the sense that they don't reflect use.

Let me know if you see any such instances right away from the list.

practicalswift:
I look into setting up my own Travis CI account so that I don't pollute the main project with WIP PR:s to trigger Travis. Sorry about that!

src/base58.cpp
```diff
@@ -4,16 +4,8 @@
 
 #include "base58.h"
 
-#include "hash.h"
-#include "uint256.h"
 
-#include <assert.h>
-#include <stdint.h>
 #include <string.h>
-#include <vector>
```
luke-jr: `std::vector` is used in this file, so the include is necessary. I imagine this is true of many other removed includes...

practicalswift: Yes, that's true! Please disregard this WIP PR (it shouldn't have been submitted in the first place - I should have found another way to trigger Travis CI testing), see the final PR "Remove includes in .cpp files for things the corresponding .h file already included" instead :-)

Sorry about taking up reviewer time with this WIP stuff.

PR #10574, opened 2017-06-11
Remove includes in .cpp files for things the corresponding .h file already included

practicalswift:
Remove includes in .cpp files for things the corresponding .h file already included.

Example case:
* `addrdb.cpp` includes `addrdb.h` and `fs.h`
* `addrdb.h` includes `fs.h`

Then remove the direct inclusion of `fs.h` in `addrman.cpp` and rely on the indirect inclusion of `fs.h` via the included `addrdb.h`.

In line with the header include guideline (see #10575).

### DIFF
--- a/src/qt/platformstyle.cpp
+++ b/src/qt/platformstyle.cpp
@@ -8,10 +8,8 @@
 
 #include <QApplication>
 #include <QColor>
-#include <QIcon>
 #include <QImage>
 #include <QPalette>
-#include <QPixmap>
 
 static const struct {
     const char *platformId;
--- a/src/addrdb.cpp
+++ b/src/addrdb.cpp
@@ -8,7 +8,6 @@
 #include <addrman.h>
 #include <chainparams.h>
 #include <clientversion.h>
-#include <fs.h>
 #include <hash.h>
 #include <random.h>
 #include <streams.h>
--- a/src/checkpoints.cpp
+++ b/src/checkpoints.cpp
@@ -8,7 +8,6 @@
 #include <chainparams.h>
 #include <reverse_iterator.h>
 #include <validation.h>
-#include <uint256.h>
 
 #include <stdint.h>
 
--- a/src/clientversion.cpp
+++ b/src/clientversion.cpp
@@ -6,7 +6,6 @@
 
 #include <tinyformat.h>
 
-#include <string>
 
 /**
  * Name of client reported in the 'version' message. Report the same name
--- a/src/coins.cpp
+++ b/src/coins.cpp
@@ -5,10 +5,7 @@
 #include <coins.h>
 
 #include <consensus/consensus.h>
-#include <memusage.h>
 #include <random.h>
-
-#include <assert.h>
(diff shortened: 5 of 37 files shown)


practicalswift:
Candidates for removal found by:

```bash
#!/bin/bash

for H in $(git grep "" -- "*.h" | cut -f1 -d: | uniq); do
    CPP=${H/%\.h/.cpp}
    if [[ ! -e $CPP ]]; then
        continue
    fi
    cat "$H" "$CPP" | grep -E "^#include " | sort | uniq -c | grep " 2 #include" && echo "$CPP" && echo
done
```

practicalswift:
Whoops, force pushed to the wrong branch. Correction coming up!

practicalswift:
Rebased!

practicalswift:
Updated!

practicalswift:
Anyone willing to review? :-)

razamobin:
Hi, I'm new to the bitcoin source code, but I looked for redundant includes and got the same changeset as you. Also I was able to run unit tests and they all passed (although I guess Travis did this already?). Either way, looks good to me.

luke-jr:
Soft NACK. IMO it's better code quality to explicitly include everything needed in each source file using them.

practicalswift:
Rebased! :-)

danra:
I agree with luke-jr

sipa:
@luke-jr @danra Then you should argue to change the guidelines in developer-notes.md:

> Every .cpp and .h file should #include every header file it directly uses classes, functions or other definitions from, even if those headers are already included indirectly through other headers. One exception is that a .cpp file does not need to re-include the includes already included in its corresponding .h file.

I personally think that rule makes sense, as you can see the .cpp file as the continuation of the corresponding .h file.

danra:
@sipa Yes, I'm for changing this rule as it creates an unnecessary dependency of the implementation (.cpp) on the interface (.h). If the interface no longer uses some other include directly, but the implementation still uses that include directly, removing the include from the interface would require adding it to the implementation.

(Note I'm not for having the implementation re-include *every* include of the interface - only those which it uses directly)

sipa:
@danra I understand that point of view, but I think it's a non-issue. When the interface changes, almost certainly the implementation needs to change anyway.

practicalswift:
Rebased! :-)

practicalswift:
Rebased! Getting this merged would get rid of 76 lines :-)

laanwj:
utACK a720b928c80f18d340173f39f63e7ef9cfb367c1

src/qt/platformstyle.cpp
```diff
@@ -8,10 +8,8 @@
 
 #include <QApplication>
 #include <QColor>
-#include <QIcon>
 #include <QImage>
 #include <QPalette>
-#include <QPixmap>
```
promag: This is in the header BUT I believe it's not necessary there. IMO the first exercise should be to reduce to a minimal the includes in a header file.

PR #10575, opened 2017-06-11
Header include guideline

sipa:


gmaxwell (approved):
ACK

practicalswift:
ACK!

fanquake:
ACK a090d1c

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/10575/commits/a090d1c1c9c379bce7345ba5b78aa9491e350d0b

jonasschnelli:
Good!
ACK a090d1c1c9c379bce7345ba5b78aa9491e350d0b

laanwj:
Makes sense.

Is there a good way to check/review this, though? Would be useful to add suggestion for reviewing.

I'm also afraid this will generate lots of 'mess around with headers' PRs, so we probably want to add that this holds for new code.

practicalswift:
@laanwj I wrote the following script ...

```bash
#!/bin/bash

for H in $(git grep "" -- "*.h" | cut -f1 -d: | uniq); do
    CPP=${H/%\.h/.cpp}
    if [[ ! -e $CPP ]]; then
        continue
    fi
    cat "$H" "$CPP" | grep -E "^#include " | sort | uniq -c | \
      grep " 2 #include" && echo "$CPP" && echo
done
```

... to identify `.cpp` files that re-include the includes already included in its corresponding `.h` file. See PR #10575.

laanwj:
Well yes but I'm more interested in the first point:

> Every `.cpp` and `.h` file should `#include` every header file it directly uses classes, functions or other
 definitions from, even if those headers are already included indirectly through other headers.

So how to check that a .cpp or .h is using functionality that is only indirectly included? Seems like *a lot* of work to do manually, at least.

practicalswift:
@laanwj Ah, [`iwyu`](https://github.com/include-what-you-use/include-what-you-use) is best-in-class there!

ryanofsky:
Iwyu is great and it would be really nice if it works on bitcoin source. It manages forward declarations as well as includes and will automatically add/remove/convert between them as appropriate.

sipa:
Perhaps we should have iwyu included in our build system, and then refer to it in developer-notes.md?

laanwj:
> Iwyu is great and it would be really nice if it works on bitcoin source

Has anyone tried?

PR #10577, opened 2017-06-11
Add an explanation of quickly hashing onto a non-power of two range.

gmaxwell:
In Olaoluwa Osuntokun's recent protocol proposal they were using a
 mod in an inner loop.  I wanted to suggest a normative protocol
 change to use the trick we use here, but to find an explanation
 of it I had to dig up the PR on github.  After I posted about it
 several other developers commented that it was very interesting
 and they were unaware of it.

I think ideally the code should be self documenting and help
 educate other contributors about non-obvious techniques that
 we use.  So I've written a description of the technique with
 citations for future reference.

### DIFF
--- a/src/cuckoocache.h
+++ b/src/cuckoocache.h
@@ -206,6 +206,37 @@ private:
     /** compute_hashes is convenience for not having to write out this
      * expression everywhere we use the hash values of an Element.
      *
+     * We need to map the 32-bit input hash onto a hash bucket in a range [0, size) in a
+     *  manner which preserves as much of the hash's uniformity as possible.  Ideally
+     *  this would be done by bitmasking but the size is usually not a power of two.
+     *
+     * The naive approach would be to use a mod -- which isn't perfectly uniform but so
+     *  long as the hash is much larger than size it is not that bad.  Unfortunately,
+     *  mod/division is fairly slow on ordinary microprocessors (e.g. 90-ish cycles on
+     *  haswell, ARM doesn't even have an instruction for it.); when the divisor is a
+     *  constant the compiler will do clever tricks to turn it into a multiply+add+shift,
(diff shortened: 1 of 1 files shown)


laanwj:
math-review-ACK https://github.com/bitcoin/bitcoin/pull/10577/commits/3dedf120f2fa51833e6587365d1ad94990189343, thanks for adding documentation.

src/cuckoocache.h
```diff
@@ -206,6 +206,37 @@ class cache
     /** compute_hashes is convenience for not having to write out this
      * expression everywhere we use the hash values of an Element.
      *
+     * We need to map the 32-bit input hash onto a hash bucket in a range [0, size) in a
+     *  manner which preserves as much of the hashe's uniformity as possible.  Ideally
```
sipa: typo: hashe's

src/cuckoocache.h
```diff
@@ -206,6 +206,37 @@ class cache
...
+     *  would bitmask but the size is usually not a power of two.
```
sipa: typo: "Ideally would bitmask" is not grammatical

src/cuckoocache.h
```diff
@@ -206,6 +206,37 @@ class cache
...
+     *
+     * The naive approach would be to use a mod -- which isn't perfectly uniform but so
+     *  long as the hash is much larger than size its not that bad.  Unfortunately,
```
sipa: typo: its

sipa (commented):
ACK, but some typos

laanwj (approved):

sipa:
utACK dd869c60ca069fa3eea3dd1aab977b8a10e05f2f

MarcoFalke:
utACK dd869c60ca069fa3eea3dd1aab977b8a10e05f2f

PR #10578, opened 2017-06-12
Add missing include for atomic in db.h

morcos:
I was getting compile errors without this.

### DIFF
--- a/src/wallet/db.h
+++ b/src/wallet/db.h
@@ -13,6 +13,7 @@
 #include "sync.h"
 #include "version.h"
 
+#include <atomic>
 #include <map>
 #include <string>
 #include <vector>


jnewbery:
Tested ACK 8e69adc6651ed533a014d883f06be33c468cf2db

paveljanik:
trivial ACK https://github.com/bitcoin/bitcoin/pull/10578/commits/8e69adc6651ed533a014d883f06be33c468cf2db

ryanofsky (commented):
utACK 8e69adc6651ed533a014d883f06be33c468cf2db

Compile error doesn't happen for me, but include should be added because an atomic was introduced here in #8694

sipa:
utACK 8e69adc6651ed533a014d883f06be33c468cf2db

PR #10579, opened 2017-06-12
[RPC] Split signrawtransaction into wallet and non-wallet RPC command

achow101:
This PR is part of #10570. It also builds on top of #10571.

This PR splits `signrawtransaction` into two commands, `signrawtransactionwithkey` and `signrawtransactionwithwallet`. `signrawtransactionwithkey` requires private keys to be passed in and does not use the wallet for any signing. `signrawtransactionwithwallet` uses the wallet to sign a raw transaction and does not have any parameters to take private keys.

The `signrawtransaction` RPC has been marked as deprecated and will call the appropriate RPC command based upon the parameters given. A test was added to check this behavior is still consistent with the original behavior.

All tests that used `signrawtransaction` have been updated to use one of the two new RPCs. Most uses were changed to `signrawtransactionwithwallet`. These were changed via a scripted diff.


sipa:
Incorrect scripted diff.

jonasschnelli:
I'm not entirely sure if this is a good long term strategy.
`signrawtransactionwithwallet` seems okay(ish) but I don't see a reason to pass around a private key (though RPC, TCP into the same process that runs the p2p/validation/node).

Where are the differences between`bitcoin-tx`'s `sign` command and the here proposed `signrawtransactionwithkey`?
Wouldn't it make more sense to focus on `bitcoin-tx` for (offline) rawtx signing?

achow101:
@jonasschnelli `signrawtransactionwithkey` will lookup the UTXOs in order to sign whereas `bitcoin-tx`'s `sign' command requires you to supply them. This is much easier to use.

laanwj:
This has missed the 0.15 feature freeze, moving to 0.16.

jnewbery:
needs rebase

achow101:
rebased

src/rpc/rawtransaction.cpp
```diff
@@ -641,86 +641,13 @@ UniValue combinerawtransaction(const JSONRPCRequest& request)
     return EncodeHexTx(mergedTx);
 }
 
-UniValue signrawtransaction(const JSONRPCRequest& request)
+UniValue signtransaction(CMutableTransaction& mtx, const UniValue& prevTxsUnival, CBasicKeyStore *keystore, bool tempKeystore, const UniValue& hashType)
```
jnewbery: nit: new function parameters should be snake_case

jnewbery (changes requested):
Some review comments inline. It's a shame that this doesn't remove the server->wallet dependency, but it's a good first step in that direction.

I felt bad about asking you to rebase and then not actually reviewing before it got stale again, so I rebased it myself here: https://github.com/jnewbery/bitcoin/tree/pr10579 . Feel free to grab that branch if it helps.

src/rpc/client.cpp
```diff
@@ -95,6 +95,9 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "createrawtransaction", 3, "replaceable" },
     { "signrawtransaction", 1, "prevtxs" },
     { "signrawtransaction", 2, "privkeys" },
+    { "signrawtransactionwithwallet", 1, "prevtxs" },
+    { "signrawtransactionwithkey", 2, "prevtxs" },
+    { "signrawtransactionwithkey", 1, "privkeys" },
```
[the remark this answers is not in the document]
promag: @jnewbery for me it's more "Sort please" :trollface:

src/Makefile.am
```diff
@@ -132,6 +132,7 @@ BITCOIN_CORE_H = \
   rpc/protocol.h \
   rpc/server.h \
   rpc/register.h \
+  rpc/rawtransaction.h \
```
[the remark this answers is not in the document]
promag: ... and sorted.

test/functional/signrawtransactions.py
```diff
@@ -34,6 +34,16 @@ def successful_signing_test(self):
         outputs = {'mpLQjfK79b7CCV4VMJWEWAj5Mpx8Up5zxB': 0.1}
 
         rawTx = self.nodes[0].createrawtransaction(inputs, outputs)
+        rawTxSigned = self.nodes[0].signrawtransactionwithkey(rawTx, privKeys, inputs)
+
+        # 1) The transaction has a complete set of signatures
+        assert 'complete' in rawTxSigned
```
promag: Remove key asserts since below you assert the value, which fail in case the key is not defined? cc @jnewbery.

src/rpc/rawtransaction.h
```diff
@@ -0,0 +1,14 @@
...
+
+#ifndef BITCOIN_RPC_RAWTRANSACTION_H
+#define BITCOIN_RPC_RAWTRANSACTION_H
+
+class CBasicKeyStore;
+
+/** Sign a transaction with the given keystore and previous transactions */
+UniValue signtransaction(CMutableTransaction& mtx, const UniValue& prevTxs, CBasicKeyStore *keystore, bool tempKeystore, const UniValue& hashType);
+
+#endif
```
promag: Missing comment.

src/rpc/rawtransaction.h
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) 2017 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef BITCOIN_RPC_RAWTRANSACTION_H
+#define BITCOIN_RPC_RAWTRANSACTION_H
+
+class CBasicKeyStore;
```
promag: Unknown types below: `CMutableTransaction` and `UniValue`.

src/wallet/rpcwallet.cpp
```diff
@@ -18,6 +18,7 @@
 #include "policy/rbf.h"
 #include "rpc/mining.h"
 #include "rpc/server.h"
+#include "rpc/rawtransaction.h"
```
promag: Sort.

src/rpc/rawtransaction.cpp
```diff
@@ -888,6 +786,185 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
     return result;
 }
 
+UniValue signrawtransactionwithkey(const JSONRPCRequest& request)
+{
+
```
promag: Remove newline.

src/rpc/rawtransaction.cpp
```diff
@@ -888,6 +786,185 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
...
+    RPCTypeCheck(request.params, {UniValue::VSTR, UniValue::VARR, UniValue::VARR, UniValue::VSTR}, true);
+
+    CMutableTransaction mtx;
+    if (!DecodeHexTx(mtx, request.params[0].get_str(), true))
+        throw JSONRPCError(RPC_DESERIALIZATION_ERROR, "TX decode failed");
+
+    CBasicKeyStore keystore;
+    UniValue keys = request.params[1].get_array();
+    for (unsigned int idx = 0; idx < keys.size(); idx++) {
```
promag: `++idx`.

src/rpc/rawtransaction.cpp
```diff
@@ -888,6 +786,185 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
...
+        UniValue k = keys[idx];
+        CBitcoinSecret vchSecret;
+        bool fGood = vchSecret.SetString(k.get_str());
+        if (!fGood)
+            throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid private key");
+        CKey key = vchSecret.GetKey();
+        if (!key.IsValid())
```
promag: Braces.

src/rpc/rawtransaction.cpp
```diff
@@ -888,6 +786,185 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
...
+    if (!DecodeHexTx(mtx, request.params[0].get_str(), true))
+        throw JSONRPCError(RPC_DESERIALIZATION_ERROR, "TX decode failed");
+
+    CBasicKeyStore keystore;
+    UniValue keys = request.params[1].get_array();
+    for (unsigned int idx = 0; idx < keys.size(); idx++) {
+        UniValue k = keys[idx];
+        CBitcoinSecret vchSecret;
+        bool fGood = vchSecret.SetString(k.get_str());
+        if (!fGood)
```
promag: Take the opportunity to kill `fGood`:
```cpp
if (!vchSecret.SetString(k.get_str())) {
```

src/rpc/rawtransaction.cpp
```diff
@@ -888,6 +786,185 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
...
+    // Otherwise sign with the wallet
+#ifdef ENABLE_WALLET
+    else if (request.params[2].isNull()) {
+        newRequest.params.push_back(request.params[0]);
+        newRequest.params.push_back(request.params[1]);
+        newRequest.params.push_back(request.params[3]);
+        return signrawtransactionwithwallet(newRequest);
+    }
+#endif
+    else
```
promag:
```cpp
else {
```

src/rpc/rawtransaction.cpp
```diff
@@ -970,7 +1047,8 @@ static const CRPCCommand commands[] =
     { "rawtransactions",    "decodescript",           &decodescript,           true,  {"hexstring"} },
     { "rawtransactions",    "sendrawtransaction",     &sendrawtransaction,     false, {"hexstring","allowhighfees"} },
     { "rawtransactions",    "combinerawtransaction",  &combinerawtransaction,  true,  {"txs"} },
-    { "rawtransactions",    "signrawtransaction",     &signrawtransaction,     false, {"hexstring","prevtxs","privkeys","sighashtype"} }, /* uses wallet if enabled */
+    { "rawtransactions",    "signrawtransaction",     &signrawtransaction,     true, {"hexstring","prevtxs","privkeys","sighashtype"} },
```
promag: Ops, align this and above lines? 😞

src/rpc/rawtransaction.cpp
```diff
@@ -641,86 +641,13 @@ UniValue combinerawtransaction(const JSONRPCRequest& request)
...
-
-    CMutableTransaction mtx;
-    if (!DecodeHexTx(mtx, request.params[0].get_str(), true))
-        throw JSONRPCError(RPC_DESERIALIZATION_ERROR, "TX decode failed");
-
     // Fetch previous transactions (inputs):
     CCoinsView viewDummy;
     CCoinsViewCache view(&viewDummy);
     {
+        LOCK(cs_main);
```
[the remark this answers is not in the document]
achow101: since `cs_main` isn't necessary in `signrawtransactionwithwallet`, I've removed it.

src/rpc/rawtransaction.h
achow101: Huh?

src/wallet/rpcwallet.cpp
achow101: It is sorted..

src/rpc/rawtransaction.h
promag:
```cpp
#endif // BITCOIN_RPC_RAWTRANSACTION_H
```

src/rpc/rawtransaction.cpp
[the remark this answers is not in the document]
achow101: Actually it is needed, but locking twice in the same thread is fine since it is a `recursive_mutex`.

achow101:
Rebased and addressed many comments

test/functional/txn_clone.py
```diff
@@ -154,5 +154,4 @@ def run_test(self):
         assert_equal(self.nodes[1].getbalance("from0", 0), -(tx1["amount"] + tx2["amount"]))
 
 if __name__ == '__main__':
-    TxnMallTest().main()
-
+    TxnMallTest().main()
```
MarcoFalke: Whitespace issue causing the scripted diff to fail

test/functional/signrawtransactions.py
MarcoFalke: Right. The line below should raise KeyError in that case.

test/functional/signrawtransactions.py
```diff
@@ -85,11 +95,37 @@ def script_verification_error_test(self):
...
+        # 3) The transaction has no complete set of signatures
+        assert 'complete' in rawTxSigned
+        assert not rawTxSigned['complete']
+
+        # 4) Two script verification errors occurred
+        assert 'errors' in rawTxSigned
+        assert_equal(len(rawTxSigned['errors']), 2)
+
+        # 5) Script verification errors have certain properties
+        assert 'txid' in rawTxSigned['errors'][0]
```
jnewbery: As above, you don't need to test presence of these keys if you're going to test the values below. Recommend you flip `# 5` and `# 6` sections, and only test presence of keys where you haven't tested the values.

src/rpc/rawtransaction.h
```diff
@@ -0,0 +1,13 @@
...
+
+/** Sign a transaction with the given keystore and previous transactions */
+UniValue sign_transaction(CMutableTransaction& mtx, const UniValue& prevTxs, CBasicKeyStore *keystore, bool tempKeystore, const UniValue& hashType);
+
+#endif
```
promag: Missing comment
```cpp
#endif // BITCOIN_RPC_RAWTRANSACTION_H
```

test/functional/signrawtransactions.py
achow101: These were copied from a previously existing test.

src/rpc/rawtransaction.cpp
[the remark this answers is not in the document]
jnewbery: looking at this some more, I don't think `cs_main` is required here. You could turn this back into `LOCK(mempool.cs)` and change the lock in `signrawtransactionwithwallet` to `LOCK(pwallet->cs_wallet)`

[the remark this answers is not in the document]
achow101: @jnewbery that's what I originally thought. Then when I removed it I got a bunch of `AssertLockHeld` problems that went away after I added it back in.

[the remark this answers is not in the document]
jnewbery: You're right. Suhas points out to me that `pcoinsTip` requires `cs_main`. This should be left as `LOCK2(cs_main, mempool.cs)`

test/functional/signrawtransactions.py
jnewbery: ok, I don't think there's any need to copy bad patterns, but I'm not too concerned either way.

src/rpc/rawtransaction.h
jnewbery: Still not addressed in f5a3e0d4a3b4b359714b7d380d7784b7ca0524c0

src/wallet/rpcwallet.cpp
```diff
@@ -2962,6 +2963,75 @@ UniValue fundrawtransaction(const JSONRPCRequest& request)
...
+            "    ,...\n"
+            "  ]\n"
+            "}\n"
+
+            "\nExamples:\n"
+            + HelpExampleCli("signrawtransactionwithwallet", "\"myhex\"")
+            + HelpExampleRpc("signrawtransactionwithwallet", "\"myhex\"")
+        );
+
+    LOCK2(cs_main, pwallet->cs_wallet);
```
jnewbery: suggest you move this lock down to immediately before the call to `sign_transaction()`. No need to hold the lock if the parameters are invalid and we're just going to throw.

src/wallet/rpcwallet.cpp
```diff
@@ -3162,6 +3232,7 @@ static const CRPCCommand commands[] =
 { //  category              name                        actor (function)           okSafeMode
     //  --------------------- ------------------------    -----------------------    ----------
     { "rawtransactions",    "fundrawtransaction",       &fundrawtransaction,       false,  {"hexstring","options"} },
+    { "wallet",             "signrawtransactionwithwallet",       &signrawtransactionwithwallet,       false, {"hexstring","prevtxs","sighashtype"} },
```
jnewbery: alignment

src/rpc/rawtransaction.cpp
```diff
@@ -885,6 +786,185 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
...
+        new_request.params.push_back(request.params[0]);
+        new_request.params.push_back(request.params[1]);
+        new_request.params.push_back(request.params[3]);
+        return signrawtransactionwithwallet(new_request);
+    }
+#endif
+    // If we have made it this far, then wallet is disabled and no private keys were given, so fail here.
+    throw JSONRPCError(RPC_INVALID_PARAMETER, "No private keys available.");
+
+    return NullUniValue;
```
jnewbery: I don't think this is required (we can never get this far since we throw two lines above.

test/functional/signrawtransactions.py
promag: Agree with @jnewbery.

jnewbery (commented):
A few more nits inline. I'm still not entirely convinced that this is the best approach. Would it be better to have a `signrawtransaction` RPC in the wallet and a different `signrawtransaction` RPC in the node? If `signrawtransaction` is called on the node without a `prevkeys` argument, then an error is returned telling the user to call `signrawtransaction` on the wallet endpoint.

That would be a breaking API change, but we need to break the API at some point (to disallow calling `signrawtransaction` without keys). Having different `signrawtransaction` RPCs for wallet and node would mean that clients who call `signrawtransaction` with `privkeys` wouldn't need to change at all, and clients who call `signrawtransaction` for the wallet would simply need to change the endpoint that they send the RPC to. This change will eventually require all users to change the method they call (and the order/number of parameters)

The downside is that having two different `signrawtransaction` RPCs could be confusing.

src/wallet/rpcwallet.cpp
achow101: Aligning this is hard

jnewbery: why? Can be done in a whitespace only commit at the end if you're worried about the diff

src/rpc/rawtransaction.h
achow101: Done

src/wallet/rpcwallet.cpp
achow101: Done

src/wallet/rpcwallet.cpp
achow101: Done

src/rpc/rawtransaction.cpp
achow101: Done

test/functional/txn_clone.py
```diff
@@ -45,11 +45,11 @@ def run_test(self):
         # Coins are sent to node1_address
         node1_address = self.nodes[1].getnewaddress("from0")
 
-        # Send tx1, and another transaction tx2 that won't be cloned
```
sipa: This change line and a few others aren't supposed to happen in a scripted diff.

achow101: Oops

achow101:
Rebased onto #11031 and put `signrawtransaction` behind `IsDeprecatedRPCEnabled`.

achow101:
I don't know why travis is failing here

src/rpc/rawtransaction.cpp
```diff
@@ -893,6 +793,190 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
     return result;
 }
 
+UniValue signrawtransactionwithkey(const JSONRPCRequest& request)
```
fanquake: Whitespace above this line seems to be causing the travis failure.

fanquake:
@achow101 Looks like the whitespace linter.
```
diff --git a/src/rpc/rawtransaction.cpp b/src/rpc/rawtransaction.cpp
@@ -895,0 +796,184 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
+
^---- failure generated from contrib/devtools/lint-whitespace.sh
```
Have commented inline.

achow101:
@fanquake I figured that much. I think I found the line with the error and it is definitely not the one which the linter points to. That error message is super unhelpful.

fanquake:
@achow101 is that the first time you've seen the linter masking/hiding another error? We definitely don't want a situation where the linter ends up being ignored, because it's being triggering miscellaneously (the whitespace here wasn't touched or introduced), and in actual fact it's hiding errors.

src/qt/rpcconsole.cpp
```diff
@@ -72,6 +72,8 @@ const QStringList historyFilter = QStringList()
     << "importmulti"
     << "signmessagewithprivkey"
     << "signrawtransaction"
+    << "signrawtransactionwithwallet"
```
jnewbery: `signrawtransactionwithwallet` doesn't handle private keys, so I don't think it needs to be included here.

src/rpc/rawtransaction.cpp
jnewbery: still not removed in ac4f385

achow101:
@fanquake Yes.

test/functional/signrawtransactions.py
```diff
@@ -2,7 +2,7 @@
 # Copyright (c) 2015-2016 The Bitcoin Core developers
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
-"""Test transaction signing using the signrawtransaction RPC."""
+"""Test transaction signing using the signrawtransactionwithwallet RPC."""
```
jnewbery: Test transaction signing using the signrawtransactionwithkey and signrawtransactionwithwallet RPCs.

test/functional/signrawtransactions.py
```diff
@@ -33,11 +34,19 @@ def successful_signing_test(self):
         outputs = {'mpLQjfK79b7CCV4VMJWEWAj5Mpx8Up5zxB': 0.1}
 
         rawTx = self.nodes[0].createrawtransaction(inputs, outputs)
+        rawTxSigned = self.nodes[0].signrawtransactionwithkey(rawTx, privKeys, inputs)
```
jnewbery: I don't think you need to repeat all these tests twice. Just add something like the following to the end of each section:

```python
rawTxSignedWithKey = self.nodes[0].signrawtransactionwithkey(rawTx, privKeys, inputs)
assert_equal(rawTxSigned, rawTxSignedWithKey)
```

src/rpc/rawtransaction.cpp
achow101: I thought I did that already.. Should be done now.

src/qt/rpcconsole.cpp
achow101: Done.

test/functional/signrawtransactions.py
achow101: Done.

test/functional/signrawtransactions.py
achow101: Done

achow101:
Rebased to master. Addressed @jnewbery's comments.

jnewbery:
scripted-diff failure

achow101:
> scripted-diff failure

Doesn't look like it...

jnewbery:
> Doesn't look like it...

Sorry, I must've been seeing a cached result.

src/rpc/rawtransaction.cpp
```diff
@@ -893,6 +793,188 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
...
+            "2. \"privkeys\"     (string, required) A json array of base58-encoded private keys for signing\n"
+            "    [                  (json array of strings)\n"
+            "      \"privatekey\"   (string) private key in base58-encoding\n"
+            "      ,...\n"
+            "    ]\n"
+            "3. \"prevtxs\"       (string, optional) An json array of previous dependent transaction outputs\n"
+            "     [               (json array of json objects, or 'null' if none provided)\n"
+            "       {\n"
+            "         \"txid\":\"id\",             (string, required) The transaction id\n"
+            "         \"vout\":n,                  (numeric, required) The output number\n"
```
promag: Nit, fix alignment.

src/rpc/rawtransaction.cpp
```diff
@@ -893,6 +793,188 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
...
+            "         \"scriptPubKey\": \"hex\",   (string, required) script key\n"
+            "         \"redeemScript\": \"hex\",   (string, required for P2SH or P2WSH) redeem script\n"
+            "         \"amount\": value            (numeric, required) The amount spent\n"
```
promag: Nit, fix alignment.

src/rpc/rawtransaction.cpp
```diff
@@ -893,6 +793,188 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
...
+
+    RPCTypeCheck(request.params, {UniValue::VSTR, UniValue::VARR, UniValue::VARR, UniValue::VSTR}, true);
+
+    CMutableTransaction mtx;
+    if (!DecodeHexTx(mtx, request.params[0].get_str(), true)) {
+        throw JSONRPCError(RPC_DESERIALIZATION_ERROR, "TX decode failed");
+    }
+
+    CBasicKeyStore keystore;
+    UniValue keys = request.params[1].get_array();
```
promag:
```cpp
const UniValue& keys = ...
```

src/rpc/rawtransaction.cpp
```diff
@@ -893,6 +793,188 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
...
+#ifdef ENABLE_WALLET
+            + HelpRequiringPassphrase(pwallet) + "\n"
+#endif
+            "\nArguments:\n"
+            "1. \"hexstring\"     (string, required) The transaction hex string\n"
+            "2. \"prevtxs\"       (string, optional) An json array of previous dependent transaction outputs\n"
+            "     [               (json array of json objects, or 'null' if none provided)\n"
+            "       {\n"
+            "         \"txid\":\"id\",             (string, required) The transaction id\n"
+            "         \"vout\":n,                  (numeric, required) The output number\n"
```
promag: Fix alignment.

src/rpc/rawtransaction.cpp
```diff
@@ -893,6 +793,188 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
...
+            "         \"scriptPubKey\": \"hex\",   (string, required) script key\n"
+            "         \"redeemScript\": \"hex\",   (string, required for P2SH or P2WSH) redeem script\n"
+            "         \"amount\": value            (numeric, required) The amount spent\n"
```
promag: Fix alignment.

promag (commented):
Some comments, I'll test later today.

src/rpc/rawtransaction.h
achow101: Hmm. It must have gotten lost somewhere.

src/rpc/rawtransaction.cpp
achow101: Alignment fixed

src/rpc/rawtransaction.cpp
achow101: Done

jnewbery:
ACK 03ca3abb6963bbf621337f2de59407b6dfb6c8f0

achow101:
Rebased.

achow101:
n'th rebase

achow101:
Rebased.

src/wallet/rpcwallet.cpp
```diff
@@ -3439,61 +3509,62 @@ extern UniValue importmulti(const JSONRPCRequest& request);
...
-    { "wallet",             "settxfee",                 &settxfee,                 {"amount"} },
-    { "wallet",             "signmessage",              &signmessage,              {"address","message"} },
-    { "wallet",             "walletlock",               &walletlock,               {} },
-    { "wallet",             "walletpassphrasechange",   &walletpassphrasechange,   {"oldpassphrase","newpassphrase"} },
-    { "wallet",             "walletpassphrase",         &walletpassphrase,         {"passphrase","timeout"} },
-    { "wallet",             "removeprunedfunds",        &removeprunedfunds,        {"txid"} },
-    { "wallet",             "rescanblockchain",         &rescanblockchain,         {"start_height", "stop_height"} },
-
-    { "generating",         "generate",                 &generate,                 {"nblocks","maxtries"} },
+ { //  category              name                                actor (function)                argNames
```
jnewbery: bad rebase: leading space character introduced.

jnewbery (commented):
utACK 3f4351dbd43a86c9197c30f13ce8b409c4a1548e

Confirm rebase is good except for a stray space character introduced.

src/rpc/client.cpp
```diff
@@ -95,6 +95,9 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "createrawtransaction", 3, "replaceable" },
     { "signrawtransaction", 1, "prevtxs" },
     { "signrawtransaction", 2, "privkeys" },
+    { "signrawtransactionwithwallet", 1, "prevtxs" },
```
jnewbery: nit: alphabetical sort please

src/wallet/rpcwallet.cpp
```diff
@@ -3439,61 +3509,62 @@ extern UniValue importmulti(const JSONRPCRequest& request);
...
+    //  --------------------- ------------------------          -----------------------         ----------
+    { "rawtransactions",    "fundrawtransaction",               &fundrawtransaction,            {"hexstring","options"} },
+    { "wallet",             "signrawtransactionwithwallet",     &signrawtransactionwithwallet,  {"hexstring","prevtxs","sighashtype"} },
```
jnewbery: nit: alphabetical sort please

achow101: Done.

achow101:
Addressed @jnewbery's nits and rebased.

achow101:
Fixed that test failure.

achow101:
Rebased

achow101:
Rebased

achow101:
Rebased

achow101:
Rebased yet again

src/Makefile.am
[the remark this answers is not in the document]
instagibbs: following lines aren't sorted either, not sure if it matters to clean this up

src/wallet/rpcwallet.cpp
instagibbs: I don't see the changes suggested?

src/wallet/rpcwallet.h
```diff
@@ -22,4 +23,6 @@ std::string HelpRequiringPassphrase(CWallet *);
 void EnsureWalletIsUnlocked(CWallet *);
 bool EnsureWalletIsAvailable(CWallet *, bool avoidException);
 
+UniValue signrawtransactionwithwallet(const JSONRPCRequest& request);
```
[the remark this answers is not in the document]
instagibbs: Although it's slightly more verbose, parallel structure with other version is nice.

`keysignrawtransaction` doesn't quite roll off the tongue

jnewbery:
Needed rebase again. Rebased here: https://github.com/jnewbery/bitcoin/tree/pr10579_rebased_again_so_achow_doesnt_want_to_kill_himself . You're welcome :)

src/wallet/rpcwallet.cpp
achow101: It probably got lost in one of the billion rebases. I added it back in.

achow101:
Thanks @jnewbery for the rebase!

src/wallet/rpcwallet.cpp
achow101: Fixed.

src/rpc/rawtransaction.cpp
```diff
@@ -672,88 +672,13 @@ UniValue combinerawtransaction(const JSONRPCRequest& request)
     return EncodeHexTx(mergedTx);
 }
 
-UniValue signrawtransaction(const JSONRPCRequest& request)
+UniValue sign_transaction(CMutableTransaction& mtx, const UniValue& prevTxsUnival, CBasicKeyStore *keystore, bool tempKeystore, const UniValue& hashType)
```
instagibbs: non blocking nit: if this stuff is being renamed we should follow snake case

src/rpc/rawtransaction.cpp
```diff
@@ -834,8 +738,8 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
             }
 
             // if redeemScript given and not using the local wallet (private keys
-            // given), add redeemScript to the tempKeystore so it can be signed:
-            if (fGivenKeys && (scriptPubKey.IsPayToScriptHash() || scriptPubKey.IsPayToWitnessScriptHash())) {
+            // given), add redeemScript to the keystore so it can be signed:
+            if (tempKeystore && (scriptPubKey.IsPayToScriptHash() || scriptPubKey.IsPayToWitnessScriptHash())) {
```
instagibbs: this is a bit confusing since previously `tempKeystore` was an actual keystore, now it's a bool. `is_temp_keystore` or something please

instagibbs (commented):
looking pretty good, will do some testing

src/rpc/rawtransaction.cpp
achow101: I believe you meant CamelCase, not snake_case. I changed it to CanelCase.

src/rpc/rawtransaction.cpp
achow101: Done

src/rpc/rawtransaction.cpp
instagibbs: referring to "variable and namespace names are all lowercase, and may use _ to separate words (snake_case)." in https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md

jnewbery:
reACK 3808129837f12482fcadd030d4e720ff2498b452. Only changes are:

- renaming suggested by instagibbs
- removing stray space character
- moving `LOCK2()` down a few lines in `signrawtransactionwithwallet()`

src/rpc/rawtransaction.cpp
achow101: Ah. Most of these aren't renamed (I think).

jnewbery: Please no more renames! :)

achow101:
Rebased again.

instagibbs:
re-utACK https://github.com/bitcoin/bitcoin/pull/10579/commits/d60234885bcc07b1a7f85ded7731549ec2fcfefa

src/rpc/rawtransaction.h
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2017 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef BITCOIN_RPC_RAWTRANSACTION_H
+#define BITCOIN_RPC_RAWTRANSACTION_H
+
+class CBasicKeyStore;
+class CMutableTransaction;
```
promag: Nit, gives compiler warning:
```
In file included from rpc/rawtransaction.cpp:20:
./rpc/rawtransaction.h:9:1: warning: class 'CMutableTransaction' was previously declared as a struct [-Wmismatched-tags]
class CMutableTransaction;
^
./primitives/transaction.h:362:8: note: previous use is here
struct CMutableTransaction
       ^
./rpc/rawtransaction.h:9:1: note: did you mean struct here?
class CMutableTransaction;
^~~~~
struct
```

promag (commented):
utACK d602348.

jnewbery:
reACK d60234885bcc07b1a7f85ded7731549ec2fcfefa

sipa:
utACK eefff65a4b0852587e0b06d93c33e2ad439d34e7

src/rpc/rawtransaction.h
achow101: I don't see this on my machine..

PR #10580, opened 2017-06-13
Prefer gpg2 to gpg

drizzt:
Many distributions (like Debian, Ubuntu LTS or Fedora) installs gpg 2.x as gpg2,
so prefer gpg2 when available

Moreover add gpg 1.4.20 to warning message since it supports --weak-digest

### DIFF
--- a/contrib/verify-commits/gpg.sh
+++ b/contrib/verify-commits/gpg.sh
@@ -9,7 +9,7 @@ REVSIG=false
 IFS='
 '
 if [ "$BITCOIN_VERIFY_COMMITS_ALLOW_SHA1" = 1 ]; then
-	GPG_RES="$(echo "$INPUT" | gpg --trust-model always "$@" 2>/dev/null)"
+	GPG_RES="$(echo "$INPUT" | "$GPG" --trust-model always "$@" 2>/dev/null)"
 else
 	# Note how we've disabled SHA1 with the --weak-digest option, disabling
 	# signatures - including selfsigs - that use SHA1. While you might think that
@@ -20,11 +20,11 @@ else
 	# they've created a collision for. Not the most likely attack, but preventing
 	# it is pretty easy so we do so as a "belt-and-suspenders" measure.
 	GPG_RES=""
-	for LINE in "$(gpg --version)"; do
+	for LINE in "$("$GPG" --version)"; do
 		case "$LINE" in
 			"gpg (GnuPG) 1.4.1"*|"gpg (GnuPG) 2.0."*)
-				echo "Please upgrade to at least gpg 2.1.10 to check for weak signatures" > /dev/stderr
-				GPG_RES="$(echo "$INPUT" | gpg --trust-model always "$@" 2>/dev/null)"
+				echo "Please upgrade to at least gpg 1.4.20 or 2.1.10 to check for weak signatures" > /dev/stderr
+				GPG_RES="$(echo "$INPUT" | "$GPG" --trust-model always "$@" 2>/dev/null)"
 				;;
 			# We assume if you're running 2.1+, you're probably running 2.1.10+
 			# gpg will fail otherwise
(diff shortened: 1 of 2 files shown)


MarcoFalke:
I suspect you need to update to gpg2 in travis.yml as well, as trusty comes with gpg2?

https://github.com/bitcoin/bitcoin/blob/303c171b949be2c050f6f52e03b74ff1ad3dea63/.travis.yml#L54

laanwj:
Concept ACK.

Though I think the `gpg2`/`gpg` thing is temporary and distribution-dependent. Let's first figure out how various Linux distributions (as well as BSDs) stand on this.

https://superuser.com/questions/763724/how-to-set-gpg2-as-default-implementation-of-gpg-on-debian/852994:
> Background: At DebConf15 the Debian GnuPG Package Maintainers announced that in the future, GnuPG 2.x will be /usr/bin/gpg in Debian and the GnuPG 1.x command will be renamed to /usr/bin/gpg1.

That's from 2015: I don't know what the plans are now.

In any case ACK on making the tool configurable with an environment variable.

drizzt:
@laanwj I have an alternative implementation idea.
What do you think about using the "standard" git config gpg.program?

Something like:
```
diff --git a/contrib/verify-commits/gpg.sh b/contrib/verify-commits/gpg.sh
index ad26e9bf0..4d2380021 100755
--- a/contrib/verify-commits/gpg.sh
+++ b/contrib/verify-commits/gpg.sh
@@ -8,9 +8,6 @@ VALID=false
 REVSIG=false
 IFS='
 '
-GPG=$(which gpg2)
-[ x"$GPG" = "x" ] && GPG=$(which gpg)
-
 if [ "$BITCOIN_VERIFY_COMMITS_ALLOW_SHA1" = 1 ]; then
 	GPG_RES="$(echo "$INPUT" | "$GPG" --trust-model always "$@" 2>/dev/null)"
 else
diff --git a/contrib/verify-commits/verify-commits.sh b/contrib/verify-commits/verify-commits.sh
index 74b7f3837..86451ff74 100755
--- a/contrib/verify-commits/verify-commits.sh
+++ b/contrib/verify-commits/verify-commits.sh
@@ -17,6 +17,12 @@ HAVE_FAILED=false
 HAVE_GNU_SHA512=1
 [ ! -x "$(which sha512sum)" ] && HAVE_GNU_SHA512=0
 
+GPG=$(git config gpg.program)
+if [ x"$GPG" = "x" ]; then
+	GPG=gpg
+fi
+export GPG
+
 if [ x"$1" = "x" ]; then
 	CURRENT_COMMIT="HEAD"
 else
```

laanwj:
Yes, using git's preferred gpg seems like a good idea!

sipa:
Concept ACK

contrib/verify-commits/gpg.sh
```diff
@@ -9,7 +9,7 @@ REVSIG=false
 IFS='
 '
 if [ "$BITCOIN_VERIFY_COMMITS_ALLOW_SHA1" = 1 ]; then
-	GPG_RES="$(echo "$INPUT" | gpg --trust-model always "$@" 2>/dev/null)"
+	GPG_RES="$(echo "$INPUT" | "$GPG" --trust-model always "$@" 2>/dev/null)"
```
laanwj: Is this script ever called separately?
We should document in a comment that this script requires environment variable GPG to be set.
Also maybe it should fail if it is not.

TheBlueMatt:
Somehow I recall there being one or two Debian releases where gpg 1 was updated to support weak digests but gpg2 was not. I don't think we have any specific reason to prefer any gpg otherwise, so if we're gonna use something other than the default maybe we should check which supports --weak-digest?

laanwj:
@TheBlueMatt Mind that the PR title is wrong (should be updated) - it no longer prefers `gpg2`, but whatever has been set as preferred gpg in `git` config.

TheBlueMatt:
@laanwj Fair, if the user has manually set a gpg program in git we could use that (and maybe update the warning docs to inform the user they can do so), but I'd think thats rare, and we might as well check both gpg and gpg2 to see which, if any, has weak-digest support?

laanwj:
> Fair, if the user has manually set a gpg program in git we could use that (and maybe update the warning docs to inform the user they can do so), but I'd think thats rare

Rare or not, there should be a way to override the gpg command used.

> might as well check both gpg and gpg2 to see which, if any, has weak-digest support?

Sounds good to me.

laanwj:
Closoing this, as there is no clear agreement in the way forward.

PR #10582, opened 2017-06-13
Pass in smart fee slider value to coin control dialog

morcos:
Since cfe77ef41 the global nTxConfirmTarget wasn't being updated by the smart
fee slider and thus the coin control dialog and labels were not being updated.

I'm not sure this is the ideal way to pass this information around, I'm open to suggestions.

### DIFF
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -822,6 +822,12 @@ void SendCoinsDialog::coinControlUpdateLabels()
     // set pay amounts
     CoinControlDialog::payAmounts.clear();
     CoinControlDialog::fSubtractFeeFromAmount = false;
+    if (ui->radioSmartFee->isChecked()) {
+        CoinControlDialog::coinControl->nConfirmTarget = ui->sliderSmartFee->maximum() - ui->sliderSmartFee->value() + 2;
+    } else {
+        CoinControlDialog::coinControl->nConfirmTarget = model->getDefaultConfirmTarget();
+    }
+
     for(int i = 0; i < ui->entries->count(); ++i)
     {
         SendCoinsEntry *entry = qobject_cast<SendCoinsEntry*>(ui->entries->itemAt(i)->widget());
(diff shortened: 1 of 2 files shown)


morcos:
Forgot to mention this should definitely be tagged 0.14
Not sure if it's worth holding up 0.14.2
@laanwj 

It's broken in the entire 0.14 branch

src/qt/sendcoinsdialog.cpp
```diff
@@ -822,6 +822,12 @@ void SendCoinsDialog::coinControlUpdateLabels()
     // set pay amounts
     CoinControlDialog::payAmounts.clear();
     CoinControlDialog::fSubtractFeeFromAmount = false;
+    if (ui->radioSmartFee->isChecked()) {
+        CoinControlDialog::coinControl->nConfirmTarget = ui->sliderSmartFee->maximum() - ui->sliderSmartFee->value() + 2;
```
ryanofsky: Seems strange that all these CoinControlDialog members are static, but of course the pr isn't changing that. This seems like a clean fix given the current organization.

laanwj:
I've tagged it for 0.14.3. If there's any issues with 0.14.2 we can include it in the next rc, but it seems that rc2 can go straight to release.

src/qt/sendcoinsdialog.cpp
laanwj: Yes the use of static here has always annoyed me too, but indeed that's not the focus here.

jonasschnelli:
Thanks for the fix.
Tested ACK: 3e7cbeb23aa97c55d2fd9656662a593a2769ec5c

morcos:
oops, forgot to replace the global nTxConfirmTarget in one more place

MarcoFalke:
@morcos As far as I see, 0.14.2 is not tagged, so you are still free to
fixup the release notes. Just make sure to ping wumpus on irc when you have
opened the pull.

I haven't looked at the code, but if this is just a display issue, it is
probably fine to only mention in the release notes. If this is actually
ignoring the value from the slider, and users that chose a slow confirm
target are effectively overpaying, we should hold back 0.14.2 for a few
days.

morcos:
@MarcoFalke yes, just display

opened #10588

laanwj:
utACK e9cd778

Issue #7965, opened 2016-04-28
Remaining instances of ENABLE_WALLET in `libbitcoin_server.a`


laanwj:
We've been quite successful in eliminating these lately (#7905, #7691). 

It would be nice to disentangle the wallet and non-wallet initialization completely so that init.cpp no longer depends on the wallet (and thus, `libbitcoin_server.a` no longer depends on `libbitcoin_wallet.a` - the other way is fine, a circular dependency is not).

Ignoring the GUI for now (which still needs some work in this direction), the following are left:
- `init.cpp`: these are mostly self-contained and short. (#10762)
  - [x] print BerkeleyDB version should probably be moved to ~~`CWallet::InitLoadWallet`~~ `CWallet::Verify` (fixed in #8036).
  - [x] wallet-specfic option interaction (`-walletbroadcast`, `-prune`+`-rescan`) should move to wallet module (possibly a new call that can be called in addition to `InitParameterInteraction()`)
  - [x] printing setKeyPool size etc should probably be moved to `CWallet::InitLoadWallet`.
  - [x] wallet initialization/verify/shutdown - maybe this could be replaced by a general registration system for initialization/interrupt/shutdown functions, called as signals in the appropriate places. The same could be used by other modules
  - [x] `-disablewallet` could result in the wallet not being registered at all, and that can all be handled from outside init.
- RPC still has some calls that vary depending on wallet support. We should split these up. This is the more annoying part as it will involve API changes. No non-wallet RPC call should make an assumption about "a wallet".
  - `rpcmisc.cpp`
    - [x] `getinfo` - this call is already on the list to be deprecated for a long time (#10838)
    - [x] `validateaddress` - should be split up into an utility-only `validateaddress` and a wallet-specific call `getaddressinfo` (name open for discussion) (#10583)
    - [x] `createmultisig` - same. `createmultisig` and `walletcreatemultisig` or so (#11415)
  - [x] `rpcblockchain.cpp`
    - [x] `signrawtransaction` - should be split into an utility-only `signrawtransaction` and a wallet-specific call `walletsignrawtransaction` (#10579)
  - [x] remove deprecated `validateaddress` and `signrawtransaction` wallet dependency completely (#12490)
- [x] Mining: no `ENABLE_WALLET` but an implicit assumption and a circuitous registration system. The `generate` call should be a wallet-specific call, whereas `generatetoaddress` is core. See discussion in #6481. (done in #10683).
[...]

PR #10583, opened 2017-06-13
[RPC] Split part of validateaddress into getaddressinfo

achow101:
This PR makes a new RPC command called `getaddressinfo` which relies on the wallet. It contains all of `validateaddress`'s address info stuff. Those parts in `validateaddress` have been marked as deprecated. The tests have been updated to use `getaddressinfo` except the `disablewallet` test which is the only test that actually uses `validateaddress` to validate an address.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -135,10 +135,10 @@ BITCOIN_CORE_H = \
   rpc/register.h \
   rpc/util.h \
   scheduler.h \
+  script/ismine.h \
   script/sigcache.h \
   script/sign.h \
   script/standard.h \
-  script/ismine.h \
   streams.h \
   support/allocators/secure.h \
   support/allocators/zeroafterfree.h \
@@ -215,9 +215,7 @@ libbitcoin_server_a_SOURCES = \
   rpc/rawtransaction.cpp \
   rpc/safemode.cpp \
   rpc/server.cpp \
-  rpc/util.cpp \
   script/sigcache.cpp \
-  script/ismine.cpp \
   timedata.cpp \
   torcontrol.cpp \
   txdb.cpp \
@@ -334,6 +332,7 @@ libbitcoin_common_a_SOURCES = \
   policy/feerate.cpp \
   protocol.cpp \
   scheduler.cpp \
+  script/ismine.cpp \
   script/sign.cpp \
   script/standard.cpp \
   warnings.cpp \
@@ -390,10 +389,10 @@ endif
 
 bitcoind_LDADD = \
   $(LIBBITCOIN_SERVER) \
+  $(LIBBITCOIN_WALLET) \
   $(LIBBITCOIN_COMMON) \
   $(LIBUNIVALUE) \
   $(LIBBITCOIN_UTIL) \
-  $(LIBBITCOIN_WALLET) \
   $(LIBBITCOIN_ZMQ) \
   $(LIBBITCOIN_CONSENSUS) \
   $(LIBBITCOIN_CRYPTO) \
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -215,6 +215,7 @@ libbitcoin_server_a_SOURCES = \
   rpc/rawtransaction.cpp \
   rpc/safemode.cpp \
   rpc/server.cpp \
+  rpc/util.cpp \
   script/sigcache.cpp \
   script/ismine.cpp \
   timedata.cpp \
--- a/src/rpc/misc.cpp
+++ b/src/rpc/misc.cpp
@@ -33,13 +33,62 @@
 
 #include <univalue.h>
 
-#ifdef ENABLE_WALLET
 class DescribeAddressVisitor : public boost::static_visitor<UniValue>
 {
 public:
-    CWallet * const pwallet;
+    explicit DescribeAddressVisitor() {}
 
-    explicit DescribeAddressVisitor(CWallet *_pwallet) : pwallet(_pwallet) {}
+    UniValue operator()(const CNoDestination &dest) const { return UniValue(UniValue::VOBJ); }
+
+    UniValue operator()(const CKeyID &keyID) const {
+        UniValue obj(UniValue::VOBJ);
+        obj.pushKV("isscript", false);
+        obj.pushKV("iswitness", false);
+        return obj;
+    }
+
+    UniValue operator()(const CScriptID &scriptID) const {
+        UniValue obj(UniValue::VOBJ);
+        obj.pushKV("isscript", true);
+        obj.pushKV("iswitness", false);
+        return obj;
+    }
+
+    UniValue operator()(const WitnessV0KeyHash& id) const
+    {
+        UniValue obj(UniValue::VOBJ);
+        obj.pushKV("isscript", false);
+        obj.pushKV("iswitness", true);
+        obj.pushKV("witness_version", 0);
+        obj.pushKV("witness_program", HexStr(id.begin(), id.end()));
+        return obj;
+    }
+
+    UniValue operator()(const WitnessV0ScriptHash& id) const
+    {
+        UniValue obj(UniValue::VOBJ);
+        obj.pushKV("isscript", true);
+        obj.pushKV("iswitness", true);
+        obj.pushKV("witness_version", 0);
+        obj.pushKV("witness_program", HexStr(id.begin(), id.end()));
+        return obj;
+    }
+
+    UniValue operator()(const WitnessUnknown& id) const
+    {
+        UniValue obj(UniValue::VOBJ);
+        obj.pushKV("iswitness", true);
+        obj.pushKV("witness_version", (int)id.version);
+        obj.pushKV("witness_program", HexStr(id.program, id.program + id.length));
+        return obj;
+    }
+};
+
+#ifdef ENABLE_WALLET
+class DescribeWalletAddressVisitor : public boost::static_visitor<UniValue>
+{
+public:
+    CWallet * const pwallet;
 
     void ProcessSubScript(const CScript& subscript, UniValue& obj, bool include_addresses = false) const
     {
@@ -54,7 +103,11 @@ public:
         UniValue a(UniValue::VARR);
         if (ExtractDestination(subscript, embedded)) {
             // Only when the script corresponds to an address.
-            UniValue subobj = boost::apply_visitor(*this, embedded);
+            UniValue subobj(UniValue::VOBJ);
+            UniValue detail = boost::apply_visitor(DescribeAddressVisitor(), embedded);
+            subobj.pushKVs(detail);
+            UniValue wallet_detail = boost::apply_visitor(*this, embedded);
+            subobj.pushKVs(wallet_detail);
             subobj.pushKV("address", EncodeDestination(embedded));
             subobj.pushKV("scriptPubKey", HexStr(subscript.begin(), subscript.end()));
             // Always report the pubkey at the top level, so that `getnewaddress()['pubkey']` always works.
@@ -81,13 +134,13 @@ public:
         if (include_addresses) obj.pushKV("addresses", std::move(a));
     }
 
+    explicit DescribeWalletAddressVisitor(CWallet *_pwallet) : pwallet(_pwallet) {}
+
     UniValue operator()(const CNoDestination &dest) const { return UniValue(UniValue::VOBJ); }
 
     UniValue operator()(const CKeyID &keyID) const {
         UniValue obj(UniValue::VOBJ);
         CPubKey vchPubKey;
-        obj.pushKV("isscript", false);
-        obj.pushKV("iswitness", false);
         if (pwallet && pwallet->GetPubKey(keyID, vchPubKey)) {
             obj.pushKV("pubkey", HexStr(vchPubKey));
             obj.pushKV("iscompressed", vchPubKey.IsCompressed());
@@ -98,8 +151,6 @@ public:
     UniValue operator()(const CScriptID &scriptID) const {
         UniValue obj(UniValue::VOBJ);
         CScript subscript;
-        obj.pushKV("isscript", true);
-        obj.pushKV("iswitness", false);
         if (pwallet && pwallet->GetCScript(scriptID, subscript)) {
             ProcessSubScript(subscript, obj, true);
         }
@@ -110,10 +161,6 @@ public:
     {
         UniValue obj(UniValue::VOBJ);
         CPubKey pubkey;
-        obj.pushKV("isscript", false);
-        obj.pushKV("iswitness", true);
-        obj.pushKV("witness_version", 0);
-        obj.pushKV("witness_program", HexStr(id.begin(), id.end()));
         if (pwallet && pwallet->GetPubKey(CKeyID(id), pubkey)) {
             obj.pushKV("pubkey", HexStr(pubkey));
         }
@@ -124,10 +171,6 @@ public:
     {
         UniValue obj(UniValue::VOBJ);
         CScript subscript;
-        obj.pushKV("isscript", true);
-        obj.pushKV("iswitness", true);
-        obj.pushKV("witness_version", 0);
-        obj.pushKV("witness_program", HexStr(id.begin(), id.end()));
         CRIPEMD160 hasher;
         uint160 hash;
         hasher.Write(id.begin(), 32).Finalize(hash.begin());
@@ -137,15 +180,7 @@ public:
         return obj;
     }
 
-    UniValue operator()(const WitnessUnknown& id) const
-    {
-        UniValue obj(UniValue::VOBJ);
-        CScript subscript;
-        obj.pushKV("iswitness", true);
-        obj.pushKV("witness_version", (int)id.version);
-        obj.pushKV("witness_program", HexStr(id.program, id.program + id.length));
-        return obj;
-    }
+    UniValue operator()(const WitnessUnknown& id) const { return UniValue(UniValue::VOBJ); }
 };
 #endif
 
@@ -219,8 +254,10 @@ UniValue validateaddress(const JSONRPCRequest& request)
         isminetype mine = pwallet ? IsMine(*pwallet, dest) : ISMINE_NO;
         ret.pushKV("ismine", bool(mine & ISMINE_SPENDABLE));
         ret.pushKV("iswatchonly", bool(mine & ISMINE_WATCH_ONLY));
-        UniValue detail = boost::apply_visitor(DescribeAddressVisitor(pwallet), dest);
+        UniValue detail = boost::apply_visitor(DescribeAddressVisitor(), dest);
         ret.pushKVs(detail);
+        UniValue wallet_detail = boost::apply_visitor(DescribeWalletAddressVisitor(pwallet), dest);
+        ret.pushKVs(wallet_detail);
         if (pwallet && pwallet->mapAddressBook.count(dest)) {
             ret.pushKV("account", pwallet->mapAddressBook[dest].name);
         }
--- a/src/rpc/misc.cpp
+++ b/src/rpc/misc.cpp
@@ -33,57 +33,6 @@
 
 #include <univalue.h>
 
-class DescribeAddressVisitor : public boost::static_visitor<UniValue>
-{
-public:
-    explicit DescribeAddressVisitor() {}
-
-    UniValue operator()(const CNoDestination &dest) const { return UniValue(UniValue::VOBJ); }
-
-    UniValue operator()(const CKeyID &keyID) const {
-        UniValue obj(UniValue::VOBJ);
-        obj.pushKV("isscript", false);
-        obj.pushKV("iswitness", false);
-        return obj;
-    }
-
-    UniValue operator()(const CScriptID &scriptID) const {
-        UniValue obj(UniValue::VOBJ);
-        obj.pushKV("isscript", true);
-        obj.pushKV("iswitness", false);
-        return obj;
-    }
-
-    UniValue operator()(const WitnessV0KeyHash& id) const
-    {
-        UniValue obj(UniValue::VOBJ);
-        obj.pushKV("isscript", false);
-        obj.pushKV("iswitness", true);
-        obj.pushKV("witness_version", 0);
-        obj.pushKV("witness_program", HexStr(id.begin(), id.end()));
-        return obj;
-    }
-
-    UniValue operator()(const WitnessV0ScriptHash& id) const
-    {
-        UniValue obj(UniValue::VOBJ);
-        obj.pushKV("isscript", true);
-        obj.pushKV("iswitness", true);
-        obj.pushKV("witness_version", 0);
-        obj.pushKV("witness_program", HexStr(id.begin(), id.end()));
-        return obj;
-    }
-
-    UniValue operator()(const WitnessUnknown& id) const
-    {
-        UniValue obj(UniValue::VOBJ);
-        obj.pushKV("iswitness", true);
-        obj.pushKV("witness_version", (int)id.version);
-        obj.pushKV("witness_program", HexStr(id.program, id.program + id.length));
-        return obj;
-    }
-};
-
 #ifdef ENABLE_WALLET
 class DescribeWalletAddressVisitor : public boost::static_visitor<UniValue>
 {
@@ -182,6 +131,15 @@ public:
 
     UniValue operator()(const WitnessUnknown& id) const { return UniValue(UniValue::VOBJ); }
 };
+
+UniValue DescribeWalletAddress(CWallet* pwallet, const CTxDestination& dest)
+{
+    UniValue ret(UniValue::VOBJ);
+    UniValue detail = DescribeAddress(dest);
+    ret.pushKVs(detail);
+    ret.pushKVs(boost::apply_visitor(DescribeWalletAddressVisitor(pwallet), dest));
+    return ret;
+}
 #endif
 
 UniValue validateaddress(const JSONRPCRequest& request)
@@ -254,10 +212,8 @@ UniValue validateaddress(const JSONRPCRequest& request)
         isminetype mine = pwallet ? IsMine(*pwallet, dest) : ISMINE_NO;
         ret.pushKV("ismine", bool(mine & ISMINE_SPENDABLE));
         ret.pushKV("iswatchonly", bool(mine & ISMINE_WATCH_ONLY));
-        UniValue detail = boost::apply_visitor(DescribeAddressVisitor(), dest);
+        UniValue detail = DescribeWalletAddress(pwallet, dest);
         ret.pushKVs(detail);
-        UniValue wallet_detail = boost::apply_visitor(DescribeWalletAddressVisitor(pwallet), dest);
-        ret.pushKVs(wallet_detail);
         if (pwallet && pwallet->mapAddressBook.count(dest)) {
             ret.pushKV("account", pwallet->mapAddressBook[dest].name);
         }
@@ -284,6 +240,8 @@ UniValue validateaddress(const JSONRPCRequest& request)
                 }
             }
         }
+#else
+        ret.pushKvs = DescribeAddress(dest);
 #endif
     }
     return ret;
--- a/src/rpc/misc.cpp
+++ b/src/rpc/misc.cpp
@@ -33,167 +33,32 @@
 
 #include <univalue.h>
 
-#ifdef ENABLE_WALLET
-class DescribeWalletAddressVisitor : public boost::static_visitor<UniValue>
-{
-public:
-    CWallet * const pwallet;
-
-    void ProcessSubScript(const CScript& subscript, UniValue& obj, bool include_addresses = false) const
-    {
-        // Always present: script type and redeemscript
-        txnouttype which_type;
-        std::vector<std::vector<unsigned char>> solutions_data;
-        Solver(subscript, which_type, solutions_data);
-        obj.pushKV("script", GetTxnOutputType(which_type));
-        obj.pushKV("hex", HexStr(subscript.begin(), subscript.end()));
-
-        CTxDestination embedded;
-        UniValue a(UniValue::VARR);
-        if (ExtractDestination(subscript, embedded)) {
-            // Only when the script corresponds to an address.
-            UniValue subobj(UniValue::VOBJ);
-            UniValue detail = boost::apply_visitor(DescribeAddressVisitor(), embedded);
-            subobj.pushKVs(detail);
-            UniValue wallet_detail = boost::apply_visitor(*this, embedded);
-            subobj.pushKVs(wallet_detail);
-            subobj.pushKV("address", EncodeDestination(embedded));
-            subobj.pushKV("scriptPubKey", HexStr(subscript.begin(), subscript.end()));
-            // Always report the pubkey at the top level, so that `getnewaddress()['pubkey']` always works.
-            if (subobj.exists("pubkey")) obj.pushKV("pubkey", subobj["pubkey"]);
-            obj.pushKV("embedded", std::move(subobj));
-            if (include_addresses) a.push_back(EncodeDestination(embedded));
-        } else if (which_type == TX_MULTISIG) {
-            // Also report some information on multisig scripts (which do not have a corresponding address).
-            // TODO: abstract out the common functionality between this logic and ExtractDestinations.
-            obj.pushKV("sigsrequired", solutions_data[0][0]);
-            UniValue pubkeys(UniValue::VARR);
-            for (size_t i = 1; i < solutions_data.size() - 1; ++i) {
-                CPubKey key(solutions_data[i].begin(), solutions_data[i].end());
-                if (include_addresses) a.push_back(EncodeDestination(key.GetID()));
-                pubkeys.push_back(HexStr(key.begin(), key.end()));
-            }
-            obj.pushKV("pubkeys", std::move(pubkeys));
-        }
-
-        // The "addresses" field is confusing because it refers to public keys using their P2PKH address.
-        // For that reason, only add the 'addresses' field when needed for backward compatibility. New applications
-        // can use the 'embedded'->'address' field for P2SH or P2WSH wrapped addresses, and 'pubkeys' for
-        // inspecting multisig participants.
-        if (include_addresses) obj.pushKV("addresses", std::move(a));
-    }
-
-    explicit DescribeWalletAddressVisitor(CWallet *_pwallet) : pwallet(_pwallet) {}
-
-    UniValue operator()(const CNoDestination &dest) const { return UniValue(UniValue::VOBJ); }
-
-    UniValue operator()(const CKeyID &keyID) const {
-        UniValue obj(UniValue::VOBJ);
-        CPubKey vchPubKey;
-        if (pwallet && pwallet->GetPubKey(keyID, vchPubKey)) {
-            obj.pushKV("pubkey", HexStr(vchPubKey));
-            obj.pushKV("iscompressed", vchPubKey.IsCompressed());
-        }
-        return obj;
-    }
-
-    UniValue operator()(const CScriptID &scriptID) const {
-        UniValue obj(UniValue::VOBJ);
-        CScript subscript;
-        if (pwallet && pwallet->GetCScript(scriptID, subscript)) {
-            ProcessSubScript(subscript, obj, true);
-        }
-        return obj;
-    }
-
-    UniValue operator()(const WitnessV0KeyHash& id) const
-    {
-        UniValue obj(UniValue::VOBJ);
-        CPubKey pubkey;
-        if (pwallet && pwallet->GetPubKey(CKeyID(id), pubkey)) {
-            obj.pushKV("pubkey", HexStr(pubkey));
-        }
-        return obj;
-    }
-
-    UniValue operator()(const WitnessV0ScriptHash& id) const
-    {
-        UniValue obj(UniValue::VOBJ);
-        CScript subscript;
-        CRIPEMD160 hasher;
-        uint160 hash;
-        hasher.Write(id.begin(), 32).Finalize(hash.begin());
-        if (pwallet && pwallet->GetCScript(CScriptID(hash), subscript)) {
-            ProcessSubScript(subscript, obj);
-        }
-        return obj;
-    }
-
-    UniValue operator()(const WitnessUnknown& id) const { return UniValue(UniValue::VOBJ); }
-};
-
-UniValue DescribeWalletAddress(CWallet* pwallet, const CTxDestination& dest)
-{
-    UniValue ret(UniValue::VOBJ);
-    UniValue detail = DescribeAddress(dest);
-    ret.pushKVs(detail);
-    ret.pushKVs(boost::apply_visitor(DescribeWalletAddressVisitor(pwallet), dest));
-    return ret;
-}
-#endif
-
 UniValue validateaddress(const JSONRPCRequest& request)
 {
     if (request.fHelp || request.params.size() != 1)
         throw std::runtime_error(
             "validateaddress \"address\"\n"
             "\nReturn information about the given bitcoin address.\n"
+            "DEPRECATION WARNING: Parts of this command have been deprecated and moved to getaddressinfo. Clients must\n"
+            "transition to using getaddressinfo to access this information before upgrading to v0.18. The following deprecated\n"
+            "fields have moved to getaddressinfo and will only be shown here with -deprecatedrpc=validateaddress: ismine, iswatchonly,\n"
+            "script, hex, pubkeys, sigsrequired, pubkey, addresses, embedded, iscompressed, account, timestamp, hdkeypath, kdmasterkeyid.\n"
             "\nArguments:\n"
-            "1. \"address\"     (string, required) The bitcoin address to validate\n"
+            "1. \"address\"                    (string, required) The bitcoin address to validate\n"
             "\nResult:\n"
             "{\n"
             "  \"isvalid\" : true|false,       (boolean) If the address is valid or not. If not, this is the only property returned.\n"
             "  \"address\" : \"address\",        (string) The bitcoin address validated\n"
             "  \"scriptPubKey\" : \"hex\",       (string) The hex encoded scriptPubKey generated by the address\n"
-            "  \"ismine\" : true|false,        (boolean) If the address is yours or not\n"
-            "  \"iswatchonly\" : true|false,   (boolean) If the address is watchonly\n"
-            "  \"isscript\" : true|false,      (boolean, optional) If the address is P2SH or P2WSH. Not included for unknown witness types.\n"
-            "  \"iswitness\" : true|false,     (boolean) If the address is P2WPKH, P2WSH, or an unknown witness version\n"
-            "  \"witness_version\" : version   (number, optional) For all witness output types, gives the version number.\n"
-            "  \"witness_program\" : \"hex\"     (string, optional) For all witness output types, gives the script or key hash present in the address.\n"
-            "  \"script\" : \"type\"             (string, optional) The output script type. Only if \"isscript\" is true and the redeemscript is known. Possible types: nonstandard, pubkey, pubkeyhash, scripthash, multisig, nulldata, witness_v0_keyhash, witness_v0_scripthash, witness_unknown\n"
-            "  \"hex\" : \"hex\",                (string, optional) The redeemscript for the P2SH or P2WSH address\n"
-            "  \"addresses\"                   (string, optional) Array of addresses associated with the known redeemscript (only if \"iswitness\" is false). This field is superseded by the \"pubkeys\" field and the address inside \"embedded\".\n"
-            "    [\n"
-            "      \"address\"\n"
-            "      ,...\n"
-            "    ]\n"
-            "  \"pubkeys\"                     (string, optional) Array of pubkeys associated with the known redeemscript (only if \"script\" is \"multisig\")\n"
-            "    [\n"
-            "      \"pubkey\"\n"
-            "      ,...\n"
-            "    ]\n"
-            "  \"sigsrequired\" : xxxxx        (numeric, optional) Number of signatures required to spend multisig output (only if \"script\" is \"multisig\")\n"
-            "  \"pubkey\" : \"publickeyhex\",    (string, optional) The hex value of the raw public key, for single-key addresses (possibly embedded in P2SH or P2WSH)\n"
-            "  \"embedded\" : {...},           (object, optional) information about the address embedded in P2SH or P2WSH, if relevant and known. It includes all validateaddress output fields for the embedded address, excluding \"isvalid\", metadata (\"timestamp\", \"hdkeypath\", \"hdmasterkeyid\") and relation to the wallet (\"ismine\", \"iswatchonly\", \"account\").\n"
-            "  \"iscompressed\" : true|false,  (boolean) If the address is compressed\n"
-            "  \"account\" : \"account\"         (string) DEPRECATED. The account associated with the address, \"\" is the default account\n"
-            "  \"timestamp\" : timestamp,      (number, optional) The creation time of the key if available in seconds since epoch (Jan 1 1970 GMT)\n"
-            "  \"hdkeypath\" : \"keypath\"       (string, optional) The HD keypath if the key is HD and available\n"
-            "  \"hdmasterkeyid\" : \"<hash160>\" (string, optional) The Hash160 of the HD master pubkey\n"
+            "  \"isscript\" : true|false,      (boolean) If the key is a script\n"
+            "  \"iswitness\" : true|false,     (boolean) If the address is a witness address\n"
+            "  \"witness_version\" : version   (numeric, optional) The version number of the witness program\n"
+            "  \"witness_program\" : \"hex\"     (string, optional) The hex value of the witness program\n"
             "}\n"
             "\nExamples:\n"
             + HelpExampleCli("validateaddress", "\"1PSSGeFHDnKNxiEyFrD1wcEaHr9hrQDDWc\"")
             + HelpExampleRpc("validateaddress", "\"1PSSGeFHDnKNxiEyFrD1wcEaHr9hrQDDWc\"")
         );
-
-#ifdef ENABLE_WALLET
-    CWallet * const pwallet = GetWalletForJSONRPCRequest(request);
-
-    LOCK2(cs_main, pwallet ? &pwallet->cs_wallet : nullptr);
-#else
-    LOCK(cs_main);
-#endif
 
     CTxDestination dest = DecodeDestination(request.params[0].get_str());
     bool isValid = IsValidDestination(dest);
@@ -202,47 +67,22 @@ UniValue validateaddress(const JSONRPCRequest& request)
     ret.pushKV("isvalid", isValid);
     if (isValid)
     {
-        std::string currentAddress = EncodeDestination(dest);
-        ret.pushKV("address", currentAddress);
-
-        CScript scriptPubKey = GetScriptForDestination(dest);
-        ret.pushKV("scriptPubKey", HexStr(scriptPubKey.begin(), scriptPubKey.end()));
 
 #ifdef ENABLE_WALLET
-        isminetype mine = pwallet ? IsMine(*pwallet, dest) : ISMINE_NO;
-        ret.pushKV("ismine", bool(mine & ISMINE_SPENDABLE));
-        ret.pushKV("iswatchonly", bool(mine & ISMINE_WATCH_ONLY));
-        UniValue detail = DescribeWalletAddress(pwallet, dest);
-        ret.pushKVs(detail);
-        if (pwallet && pwallet->mapAddressBook.count(dest)) {
-            ret.pushKV("account", pwallet->mapAddressBook[dest].name);
+        if (!::vpwallets.empty() && IsDeprecatedRPCEnabled("validateaddress")) {
+            ret.pushKVs(getaddressinfo(request));
         }
-        if (pwallet) {
-            const CKeyMetadata* meta = nullptr;
-            CKeyID key_id = GetKeyForDestination(*pwallet, dest);
-            if (!key_id.IsNull()) {
-                auto it = pwallet->mapKeyMetadata.find(key_id);
-                if (it != pwallet->mapKeyMetadata.end()) {
-                    meta = &it->second;
-                }
-            }
-            if (!meta) {
-                auto it = pwallet->m_script_metadata.find(CScriptID(scriptPubKey));
-                if (it != pwallet->m_script_metadata.end()) {
-                    meta = &it->second;
-                }
-            }
-            if (meta) {
-                ret.pushKV("timestamp", meta->nCreateTime);
-                if (!meta->hdKeypath.empty()) {
-                    ret.pushKV("hdkeypath", meta->hdKeypath);
-                    ret.pushKV("hdmasterkeyid", meta->hdMasterKeyID.GetHex());
-                }
-            }
-        }
-#else
-        ret.pushKvs = DescribeAddress(dest);
 #endif
+        if (ret["address"].isNull()) {
+            std::string currentAddress = EncodeDestination(dest);
+            ret.pushKV("address", currentAddress);
+
+            CScript scriptPubKey = GetScriptForDestination(dest);
+            ret.pushKV("scriptPubKey", HexStr(scriptPubKey.begin(), scriptPubKey.end()));;
+
+            UniValue detail = DescribeAddress(dest);
+            ret.pushKVs(detail);
+        }
     }
     return ret;
 }
--- a/src/rpc/util.cpp
+++ b/src/rpc/util.cpp
@@ -66,3 +66,64 @@ CScript CreateMultisigRedeemscript(const int required, const std::vector<CPubKey
 
     return result;
 }
+
+class DescribeAddressVisitor : public boost::static_visitor<UniValue>
+{
+public:
+    explicit DescribeAddressVisitor() {}
+
+    UniValue operator()(const CNoDestination& dest) const
+    {
+        return UniValue(UniValue::VOBJ);
+    }
+
+    UniValue operator()(const CKeyID& keyID) const
+    {
+        UniValue obj(UniValue::VOBJ);
+        obj.pushKV("isscript", false);
+        obj.pushKV("iswitness", false);
+        return obj;
+    }
+
+    UniValue operator()(const CScriptID& scriptID) const
+    {
+        UniValue obj(UniValue::VOBJ);
+        obj.pushKV("isscript", true);
+        obj.pushKV("iswitness", false);
+        return obj;
+    }
+
+    UniValue operator()(const WitnessV0KeyHash& id) const
+    {
+        UniValue obj(UniValue::VOBJ);
+        obj.pushKV("isscript", false);
+        obj.pushKV("iswitness", true);
+        obj.pushKV("witness_version", 0);
+        obj.pushKV("witness_program", HexStr(id.begin(), id.end()));
+        return obj;
+    }
+
+    UniValue operator()(const WitnessV0ScriptHash& id) const
+    {
+        UniValue obj(UniValue::VOBJ);
+        obj.pushKV("isscript", true);
+        obj.pushKV("iswitness", true);
+        obj.pushKV("witness_version", 0);
+        obj.pushKV("witness_program", HexStr(id.begin(), id.end()));
+        return obj;
+    }
+
+    UniValue operator()(const WitnessUnknown& id) const
+    {
+        UniValue obj(UniValue::VOBJ);
+        obj.pushKV("iswitness", true);
+        obj.pushKV("witness_version", (int)id.version);
+        obj.pushKV("witness_program", HexStr(id.program, id.program + id.length));
+        return obj;
+    }
+};
+
+UniValue DescribeAddress(const CTxDestination& dest)
+{
+    return boost::apply_visitor(DescribeAddressVisitor(), dest);
+}
--- a/src/rpc/util.h
+++ b/src/rpc/util.h
@@ -5,6 +5,13 @@
 #ifndef BITCOIN_RPC_UTIL_H
 #define BITCOIN_RPC_UTIL_H
 
+#include <pubkey.h>
+#include <script/standard.h>
+#include <univalue.h>
+#include <utilstrencodings.h>
+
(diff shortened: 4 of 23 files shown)


laanwj:
Concept ACK. Makes a lot of sense to move the wallet-specific information to a separate RPC.

gmaxwell (commented):
Concept ACK.

sipa:
Currently this duplicates a singificant amount of code. Would it be possible to abstract out the 'info' part of `getaddressinfo`, and then call that from `validateaddress` as well (temporarily, until removed)?

achow101:
@sipa Apparently, no, this is not possible. My attempt at doing so has resulted in a linker failure. Apparently there is some makefile problems with that. https://github.com/achow101/bitcoin/tree/getaddressinfo-broken is the branch with the duplication removed, but there is a linker error with `IsMine` calls. The full build error is here: https://gist.github.com/achow101/0f0e2d9e75d9d77132ab83a7605f2377

achow101:
On second thought, I think I fixed the problem.

sipa:
I believe you can do:

```diff
 bitcoind_LDADD = \
   $(LIBBITCOIN_SERVER) \
+  $(LIBBITCOIN_WALLET) \
   $(LIBBITCOIN_COMMON) \
   $(LIBUNIVALUE) \
   $(LIBBITCOIN_UTIL) \
-  $(LIBBITCOIN_WALLET) \
   $(LIBBITCOIN_ZMQ) \
```
in `Makefile.am`.

Then the _ismine code can move from wallet to common (which I think belongs there, as signing code is already in common too).

achow101:
@sipa I did that in my latest commit, but it seems that travis doesn't like it.

laanwj:
> Then the _ismine code can move from wallet to common (which I think belongs there, as signing code is already in common too).

From what I remember it was already moved in the other direction at some point for the same reason.
The circular dependency between `_wallet` and `_server` remains a problem.
We'll need to solve #7965 to get rid of it.
Unfortunately, this is part of that...

jnewbery:
needs rebase

achow101:
rebased

jnewbery:
Concept ACK. This PR is currently difficult to review because there are broken intermediate commits, commits which reverse previous commits, random style changes in unconnected commits, etc. Are you able to tidy this up to make it easier to review? If not, I'm happy to take this PR and tidy it up.

My high-level feedback is that we should hide the deprecated fields in `validateaddress` behind a command-line argument (similar to #11031). The `ismine` move to libbitcoin_common seems fine to me.

achow101:
Rebased and tidied up

jnewbery:
Travis is failing on invalid scripted diff check.

jnewbery (commented):
A few nits inline. I think you could drop the third commit by making your scripted diff more focussed:

```
find ./test/functional -path '*py' -not -path ./test/functional/disablewallet.py -exec sed -i'' -e 's/validateaddress/getaddressinfo/g' {} \;
```

note that sed syntax differs between BSD and GNU. You'll need to use `sed -i ''` (space between `i` and empty string) on BSD.

(this also has the benefit of not breaking git bisect)

I still think it's a good idea to hide the deprecated functionality behind a command-line argument to make removing it in the next version easier.

src/rpc/misc.cpp
```diff
@@ -214,38 +166,24 @@ UniValue validateaddress(const JSONRPCRequest& request)
...
-            }
-            if (it != meta.end()) {
-                ret.push_back(Pair("timestamp", it->second.nCreateTime));
-                if (!it->second.hdKeypath.empty()) {
-                    ret.push_back(Pair("hdkeypath", it->second.hdKeypath));
-                    ret.push_back(Pair("hdmasterkeyid", it->second.hdMasterKeyID.GetHex()));
-                }
+        try {
+            ret.pushKVs(getaddressinfo(request));
+        } catch (UniValue e) {
```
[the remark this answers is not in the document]
achow101: That is swallowed in the event that `-disablewallet` is used; `getaddressinfo` will return `RPC_METHOD_NOT_FOUND` but `validateaddress` should still work and validate addresses.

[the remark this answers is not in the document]
jnewbery: In that case, I think it's preferable to test for wallet existence before calling `getaddressinfo`. The following should do it:

```
        if (!::vpwallets.empty()) {
            ret.pushKVs(getaddressinfo(request));
        }
```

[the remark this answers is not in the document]
achow101: Done

promag (commented):
New code (and probably moved code) should take into account developer notes.

src/wallet/rpcwallet.cpp
```diff
@@ -3169,6 +3170,114 @@ UniValue generate(const JSONRPCRequest& request)
...
+    if (request.fHelp || request.params.size() != 1)
+        throw std::runtime_error(
+            "getaddressinfo \"address\"\n"
+            "\nReturn information about the given bitcoin address. Some information requires the address\n"
+            "to be in the wallet.\n"
+            "\nArguments:\n"
+            "1. \"address\"     (string, required) The bitcoin address to get the information of.\n"
+            "\nResult:\n"
+            "{\n"
+            "  \"address\" : \"address\", (string) The bitcoin address validated\n"
```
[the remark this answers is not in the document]
achow101: Forgot to say that this was addressed.

src/wallet/rpcwallet.cpp
```diff
@@ -3169,6 +3170,114 @@ UniValue generate(const JSONRPCRequest& request)
     return generateBlocks(coinbase_script, num_generate, max_tries, true);
 }
 
+class DescribeAddressVisitor : public boost::static_visitor<UniValue>
+{
+public:
+    CWallet * const pwallet;
+
+    DescribeAddressVisitor(CWallet *_pwallet) : pwallet(_pwallet) {}
```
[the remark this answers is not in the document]
achow101: Done

src/wallet/rpcwallet.cpp
```diff
@@ -3169,6 +3170,114 @@ UniValue generate(const JSONRPCRequest& request)
...
+
+    UniValue operator()(const CNoDestination &dest) const { return UniValue(UniValue::VOBJ); }
```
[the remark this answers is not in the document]
achow101: I don't think that is necessary

achow101:
Rebased

src/rpc/misc.cpp
```diff
@@ -127,37 +78,21 @@ UniValue validateaddress(const JSONRPCRequest& request)
     ret.push_back(Pair("isvalid", isValid));
     if (isValid)
     {
+
+#ifdef ENABLE_WALLET
+        if (!::vpwallets.empty()) {
+            ret.pushKVs(getaddressinfo(request));
+        }
+#endif
+        if (ret["address"].isNull()) {
```
jnewbery: nit: indentation for this if block

src/wallet/rpcwallet.cpp
```diff
@@ -3180,6 +3181,123 @@ UniValue generate(const JSONRPCRequest& request)
...
+                obj.push_back(Pair("sigsrequired", nRequired));
+        }
+        return obj;
+    }
+};
+
+UniValue getaddressinfo(const JSONRPCRequest& request)
+{
+    CWallet * const pwallet = GetWalletForJSONRPCRequest(request);
+    if (!EnsureWalletIsAvailable(pwallet, request.fHelp))
```
jnewbery: nit: braces

src/wallet/rpcwallet.cpp
```diff
@@ -3180,6 +3181,123 @@ UniValue generate(const JSONRPCRequest& request)
...
+
+    UniValue ret(UniValue::VOBJ);
+    CTxDestination dest = DecodeDestination(request.params[0].get_str());
+    std::string currentAddress = EncodeDestination(dest);
+    ret.push_back(Pair("address", currentAddress));
+
+    CScript scriptPubKey = GetScriptForDestination(dest);
+    ret.push_back(Pair("scriptPubKey", HexStr(scriptPubKey.begin(), scriptPubKey.end())));;
+    
+    ret.push_back(Pair("isscript", scriptPubKey.IsPayToScriptHash()));
```
jnewbery: micronit: not in same ordering as help text (either for `getaddressinfo` or `validateaddress`. Suggest you move this line below `ret.push_back(Pair("iswatchonly", ...`

src/wallet/rpcwallet.cpp
```diff
@@ -3180,6 +3181,123 @@ UniValue generate(const JSONRPCRequest& request)
...
+    if (!EnsureWalletIsAvailable(pwallet, request.fHelp))
+        return NullUniValue;
+
+    if (request.fHelp || request.params.size() != 1) {
+        throw std::runtime_error(
+            "getaddressinfo \"address\"\n"
+            "\nReturn information about the given bitcoin address. Some information requires the address\n"
+            "to be in the wallet.\n"
+            "\nArguments:\n"
+            "1. \"address\"     (string, required) The bitcoin address to get the information of.\n"
```
jnewbery: nit: help text alignment

src/wallet/rpcwallet.cpp
```diff
@@ -3180,6 +3181,123 @@ UniValue generate(const JSONRPCRequest& request)
...
+            "\nResult:\n"
+            "{\n"
+            "  \"address\" : \"address\", (string) The bitcoin address validated\n"
```
jnewbery: nit: help text alignment

src/wallet/rpcwallet.cpp
```diff
@@ -3180,6 +3181,123 @@ UniValue generate(const JSONRPCRequest& request)
...
+            "  \"hex\" : \"hex\",                (string, optional) The redeemscript for the p2sh address\n"
+            "  \"addresses\"                   (string, optional) Array of addresses associated with the known redeemscript\n"
+            "    [\n"
+            "      \"address\"\n"
+            "      ,...\n"
+            "    ]\n"
+            "  \"sigsrequired\" : xxxxx        (numeric, optional) Number of signatures required to spend multisig output\n"
+            "  \"pubkey\" : \"publickeyhex\",    (string) The hex value of the raw public key\n"
+            "  \"iscompressed\" : true|false,  (boolean) If the address is compressed\n"
+            "  \"timestamp\" : timestamp,        (number, optional) The creation time of the key if available in seconds since epoch (Jan 1 1970 GMT)\n"
```
jnewbery: nit: help text alignment

src/rpc/misc.cpp
```diff
@@ -31,94 +31,45 @@
...
-
 UniValue validateaddress(const JSONRPCRequest& request)
 {
     if (request.fHelp || request.params.size() != 1)
         throw std::runtime_error(
             "validateaddress \"address\"\n"
             "\nReturn information about the given bitcoin address.\n"
+            "Parts of this command have been deprecated and moved to getaddressinfo\n"
             "\nArguments:\n"
             "1. \"address\"     (string, required) The bitcoin address to validate\n"
```
jnewbery: nit: help text alignment

src/rpc/misc.cpp
```diff
@@ -31,94 +31,45 @@
...
             "\nResult:\n"
             "{\n"
             "  \"isvalid\" : true|false,       (boolean) If the address is valid or not. If not, this is the only property returned.\n"
             "  \"address\" : \"address\", (string) The bitcoin address validated\n"
```
jnewbery: nit: help text alignment

src/rpc/misc.cpp
```diff
@@ -31,94 +31,45 @@
...
-            "  \"iscompressed\" : true|false,  (boolean) If the address is compressed\n"
+            "  \"sigsrequired\" : xxxxx        (numeric, optional) DEPRECATED. Number of signatures required to spend multisig output\n"
+            "  \"pubkey\" : \"publickeyhex\",    (string) DEPRECATED. The hex value of the raw public key\n"
+            "  \"iscompressed\" : true|false,  (boolean) DEPRECATED. If the address is compressed\n"
             "  \"account\" : \"account\"         (string) DEPRECATED. The account associated with the address, \"\" is the default account\n"
-            "  \"timestamp\" : timestamp,        (number, optional) The creation time of the key if available in seconds since epoch (Jan 1 1970 GMT)\n"
-            "  \"hdkeypath\" : \"keypath\"       (string, optional) The HD keypath if the key is HD and available\n"
-            "  \"hdmasterkeyid\" : \"<hash160>\" (string, optional) The Hash160 of the HD master pubkey\n"
+            "  \"timestamp\" : timestamp,        (number, optional) DEPRECATED. The creation time of the key if available in seconds since epoch (Jan 1 1970 GMT)\n"
```
jnewbery: nit: help text alignment

jnewbery (commented):
A bunch of nits inline. I'm currently running this over the extended test suite to verify no regressions.

My high-level feedback hasn't been addressed:

> we should hide the deprecated fields in validateaddress behind a command-line argument (similar to #11031).

We need to do something to phase out deprecated RPC methods over a couple of releases. The proposal in #11031 is my suggested way of doing that, but I'm open to other suggestions.

jnewbery:
extended test suite passes. Travis is failing the scripted diff check.

achow101:
Addressed most of @jnewbery's comments except for the deprecation thing

jnewbery:
Thanks for addressing the nits.

My preference is that this doesn't get merged without some kind of phased deprecation. I'm happy to provide a commit to cherry pick if that would help.

achow101:
> My preference is that this doesn't get merged without some kind of phased deprecation. I'm happy to provide a commit to cherry pick if that would help.

I'm planning on making a generic deprecation interface type thing that I will then rebase this and some other stuff on top of.

achow101:
Rebased onto #11031 and moved the deprecated parts behind `IsDeprecatedRPCEnabled`. I wasn't sure what to do with the help text so I just left it as is (except for info about using `-deprecatedrpc`).

src/rpc/misc.cpp
```diff
@@ -31,94 +31,46 @@
...
-};
-#endif
-
 UniValue validateaddress(const JSONRPCRequest& request)
 {
     if (request.fHelp || request.params.size() != 1)
         throw std::runtime_error(
             "validateaddress \"address\"\n"
             "\nReturn information about the given bitcoin address.\n"
+            "Parts of this command have been deprecated and moved to getaddressinfo. To have those deprecated\n"
```
jnewbery: Suggestion only: Perhaps start this with `DEPRECATION WARNING:` to draw attention. It might be easy to miss this warning.

jnewbery (commented):
Looks good. Tested ACK c288c958e96d9c2c856abcde901518f0de9fd3ce.

One suggestion inline, which you could take or leave.

A test for the difference between the return object when `-deprecatedrpc=validateaddress` is set and not would be a nice to have, but not essential.

src/rpc/misc.cpp
achow101: Done

achow101:
@jnewbery I added a super basic test to check that `-deprecatedrpc=validateaddress` changes what is returned.

src/rpc/misc.cpp
```diff
@@ -127,37 +79,21 @@ UniValue validateaddress(const JSONRPCRequest& request)
...
-                    ret.push_back(Pair("hdmasterkeyid", it->second.hdMasterKeyID.GetHex()));
-                }
-            }
+        if (!::vpwallets.empty() && IsDeprecatedRPCEnabled("validateaddress")) {
+            ret.pushKVs(getaddressinfo(request));
         }
 #endif
+        if (ret["address"].isNull()) {
+            std::string currentAddress = EncodeDestination(dest);
+            ret.push_back(Pair("address", currentAddress));
```
promag: Comply with #11386 here and below?

src/wallet/rpcwallet.h
```diff
@@ -10,6 +10,7 @@
 class CRPCTable;
 class CWallet;
 class JSONRPCRequest;
+class UniValue;
```
promag: Why?

src/wallet/rpcwallet.cpp
```diff
@@ -3180,6 +3181,125 @@ UniValue generate(const JSONRPCRequest& request)
...
+    UniValue ret(UniValue::VOBJ);
+    CTxDestination dest = DecodeDestination(request.params[0].get_str());
+    std::string currentAddress = EncodeDestination(dest);
+    ret.push_back(Pair("address", currentAddress));
+
+    CScript scriptPubKey = GetScriptForDestination(dest);
+    ret.push_back(Pair("scriptPubKey", HexStr(scriptPubKey.begin(), scriptPubKey.end())));;
+
+    isminetype mine = pwallet ? IsMine(*pwallet, dest) : ISMINE_NO;
+    ret.push_back(Pair("ismine", (mine & ISMINE_SPENDABLE) ? true : false));
```
promag: `bool(mine & ISMINE_SPENDABLE)`? Same below.

test/functional/deprecated_rpc.py
```diff
@@ -0,0 +1,29 @@
...
+    def run_test(self):
+        self.log.info("estimatefee: Shows deprecated message")
+        assert_raises_jsonrpc(-32, 'estimatefee is deprecated', self.nodes[0].estimatefee, 1)
+
+        self.log.info("Using -deprecatedrpc=estimatefee bypasses the error")
+        self.nodes[1].estimatefee(1)
+
+        # test validateaddress deprecation
+        dep_validate_address = self.nodes[0].validateaddress("mnvGjUy3NMj67yJ6gkK5o9e5RS33Z2Vqcu")
+        assert("ismine" not in dep_validate_address)
```
jnewbery: assert is a statement, not a function:

```python
assert "ismine" not in dep_validate_address
```

same for assert below.

test/functional/deprecated_rpc.py
```diff
@@ -0,0 +1,29 @@
...
+        self.extra_args = [[], ["-deprecatedrpc=estimatefee", "-deprecatedrpc=validateaddress"]]
+
+    def run_test(self):
+        self.log.info("estimatefee: Shows deprecated message")
+        assert_raises_jsonrpc(-32, 'estimatefee is deprecated', self.nodes[0].estimatefee, 1)
+
+        self.log.info("Using -deprecatedrpc=estimatefee bypasses the error")
+        self.nodes[1].estimatefee(1)
+
+        # test validateaddress deprecation
```
jnewbery: suggestion: change this to a `self.log.info()`

test/functional/deprecated_rpc.py
```diff
@@ -0,0 +1,29 @@
...
+        dep_validate_address = self.nodes[0].validateaddress("mnvGjUy3NMj67yJ6gkK5o9e5RS33Z2Vqcu")
```
jnewbery: Perhaps declare this address as a variable above (with comment) to avoid having magic strings in the RPC method call.

test/functional/deprecated_rpc.py
```diff
@@ -19,5 +19,11 @@ def run_test(self):
         self.log.info("Using -deprecatedrpc=estimatefee bypasses the error")
         self.nodes[1].estimatefee(1)
 
+        # test validateaddress deprecation
+        dep_validate_address = self.nodes[0].validateaddress("mnvGjUy3NMj67yJ6gkK5o9e5RS33Z2Vqcu")
+        assert("ismine" not in dep_validate_address)
```
promag: Is it enough to check `ismine` key? Maybe merge `nodes[0].validateaddress` with `nodes[0].getaddressinfo` and that should contain `nodes[1].validateaddress`?

jnewbery (commented):
The changed commit broke the scripted-diff checker. You'll need to add your `-deprecatedrpc=validateaddress` test as a separate commit.

promag (commented):
Only reviewed *[RPC] Split part of validateaddress into getaddressinfo* commit.

test/functional/deprecated_rpc.py
achow101: It's enough to determine that they are returning different things as one should have `ismine` and the other not.

src/wallet/rpcwallet.h
achow101: `UniValue` is needed below for `UniValue getaddressinfo...`

test/functional/deprecated_rpc.py
achow101: Done

test/functional/deprecated_rpc.py
achow101: Done

test/functional/deprecated_rpc.py
achow101: Done

src/rpc/misc.cpp
achow101: Done.

achow101:
> The changed commit broke the scripted-diff checker.

I noticed. I believe that should be fixed now as I changed the script to exclude `deprecatedrpc.py`.

achow101:
Rebased onto master

src/rpc/misc.cpp
```diff
@@ -31,94 +31,46 @@
...
-#endif
-
 UniValue validateaddress(const JSONRPCRequest& request)
 {
     if (request.fHelp || request.params.size() != 1)
         throw std::runtime_error(
             "validateaddress \"address\"\n"
             "\nReturn information about the given bitcoin address.\n"
+            "DEPRECATION WARNING: Parts of this command have been deprecated and moved to getaddressinfo. To have those deprecated\n"
+            "information shown, start bitcoind with -deprecatedrpc=validateaddress"
```
jnewbery: I think this warning should also say something like "Clients must transition to using getaddressinfo to access this information before upgrading to v0.17"

jnewbery (commented):
Looks generally good. One question. If I pass an invalid address to `getaddressinfo`, I get the following output:

```
→ bitcoin-cli getaddressinfo fAkEAdDrEsS
{
  "address": "",
  "scriptPubKey": "",
  "ismine": 0,
  "iswatchonly": 0,
  "isscript": 0
}
```

I think it would be more intuitive to either:
- have an `isvalid` field that returns False (like for `validateaddress`); or
- throw a `RPC_INVALID_ADDRESS_OR_KEY` error

What do you think?

jtimon:
Concept ACK

> throw a RPC_INVALID_ADDRESS_OR_KEY error

I like this option.

achow101:
> throw a RPC_INVALID_ADDRESS_OR_KEY error

Did that. I didn't realize that I hadn't added that error there.

src/rpc/misc.cpp
achow101: Done

jnewbery:
ACK ea088845fb8dd903f681c4c08b6bfe412cecc536

src/wallet/rpcwallet.cpp
```diff
@@ -3180,6 +3181,131 @@ UniValue generate(const JSONRPCRequest& request)
...
+    }
+
+    LOCK2(cs_main, pwallet->cs_wallet);
+
+    UniValue ret(UniValue::VOBJ);
+    CTxDestination dest = DecodeDestination(request.params[0].get_str());
+
+    // Make sure the destination is valid
+    if (!IsValidDestination(dest)) {
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid address");
```
promag: Missing test.

src/wallet/rpcwallet.cpp
```diff
@@ -3180,6 +3181,131 @@ UniValue generate(const JSONRPCRequest& request)
...
+    }
+
+    std::string currentAddress = EncodeDestination(dest);
+    ret.pushKV("address", currentAddress);
+
+    CScript scriptPubKey = GetScriptForDestination(dest);
+    ret.pushKV("scriptPubKey", HexStr(scriptPubKey.begin(), scriptPubKey.end()));
+
+    isminetype mine = pwallet ? IsMine(*pwallet, dest) : ISMINE_NO;
```
promag: `pwallet` is valid because of `EnsureWalletIsAvailable` above.

src/wallet/rpcwallet.cpp
```diff
@@ -3180,6 +3181,131 @@ UniValue generate(const JSONRPCRequest& request)
...
+        }
+        return obj;
+    }
+};
+
+UniValue getaddressinfo(const JSONRPCRequest& request)
+{
+    CWallet * const pwallet = GetWalletForJSONRPCRequest(request);
+    if (!EnsureWalletIsAvailable(pwallet, request.fHelp)) {
+            return NullUniValue;
```
promag: Nit, fix indentation.

src/wallet/rpcwallet.cpp
```diff
@@ -3180,6 +3181,131 @@ UniValue generate(const JSONRPCRequest& request)
...
+            int nRequired;
+            ExtractDestinations(subscript, whichType, addresses, nRequired);
+            obj.push_back(Pair("script", GetTxnOutputType(whichType)));
+            obj.push_back(Pair("hex", HexStr(subscript.begin(), subscript.end())));
+            UniValue a(UniValue::VARR);
+            for (const CTxDestination& addr : addresses) {
+                a.push_back(EncodeDestination(addr));
+            }
+            obj.push_back(Pair("addresses", a));
+            if (whichType == TX_MULTISIG)
```
promag: Nit, add `{` or single line.

src/rpc/misc.cpp
```diff
@@ -31,94 +31,47 @@
...
             + HelpExampleCli("validateaddress", "\"1PSSGeFHDnKNxiEyFrD1wcEaHr9hrQDDWc\"")
             + HelpExampleRpc("validateaddress", "\"1PSSGeFHDnKNxiEyFrD1wcEaHr9hrQDDWc\"")
         );
 
-#ifdef ENABLE_WALLET
-    CWallet * const pwallet = GetWalletForJSONRPCRequest(request);
-
-    LOCK2(cs_main, pwallet ? &pwallet->cs_wallet : nullptr);
-#else
     LOCK(cs_main);
```
TheBlueMatt: I do not believe you need cs_main here anymore.

src/wallet/rpcwallet.cpp
```diff
@@ -3180,6 +3181,131 @@ UniValue generate(const JSONRPCRequest& request)
...
+            "  \"hdkeypath\" : \"keypath\"       (string, optional) The HD keypath if the key is HD and available\n"
+            "  \"hdmasterkeyid\" : \"<hash160>\" (string, optional) The Hash160 of the HD master pubkey\n"
+            "}\n"
+            "\nExamples:\n"
+            + HelpExampleCli("getaddressinfo", "\"1PSSGeFHDnKNxiEyFrD1wcEaHr9hrQDDWc\"")
+            + HelpExampleRpc("getaddressinfo", "\"1PSSGeFHDnKNxiEyFrD1wcEaHr9hrQDDWc\"")
+        );
+    }
+
+    LOCK2(cs_main, pwallet->cs_wallet);
```
TheBlueMatt: I believe you only need cs_wallet here.

src/rpc/misc.cpp
```diff
@@ -31,94 +31,47 @@
...
-            "  \"address\" : \"address\", (string) The bitcoin address validated\n"
+            "  \"address\" : \"address\",        (string) The bitcoin address validated\n"
             "  \"scriptPubKey\" : \"hex\",       (string) The hex encoded scriptPubKey generated by the address\n"
-            "  \"ismine\" : true|false,        (boolean) If the address is yours or not\n"
-            "  \"iswatchonly\" : true|false,   (boolean) If the address is watchonly\n"
             "  \"isscript\" : true|false,      (boolean) If the key is a script\n"
-            "  \"script\" : \"type\"             (string, optional) The output script type. Possible types: nonstandard, pubkey, pubkeyhash, scripthash, multisig, nulldata, witness_v0_keyhash, witness_v0_scripthash\n"
-            "  \"hex\" : \"hex\",                (string, optional) The redeemscript for the p2sh address\n"
-            "  \"addresses\"                   (string, optional) Array of addresses associated with the known redeemscript\n"
+            "  \"ismine\" : true|false,        (boolean) DEPRECATED. If the address is yours or not\n"
```
TheBlueMatt: Why change the ordering here? UniValue returns things ordered in insertion-order (and hence bitcoin-cli and some JSON libraries will print it in that order) so might as well just leave it...having deprecation warnings interspersed is fine, they'll all go away soon enough :)

src/rpc/misc.cpp
```diff
@@ -31,94 +31,47 @@
...
             "      \"address\"\n"
             "      ,...\n"
             "    ]\n"
-            "  \"sigsrequired\" : xxxxx        (numeric, optional) Number of signatures required to spend multisig output\n"
-            "  \"pubkey\" : \"publickeyhex\",    (string) The hex value of the raw public key\n"
-            "  \"iscompressed\" : true|false,  (boolean) If the address is compressed\n"
+            "  \"sigsrequired\" : xxxxx        (numeric, optional) DEPRECATED. Number of signatures required to spend multisig output\n"
+            "  \"pubkey\" : \"publickeyhex\",    (string) DEPRECATED. The hex value of the raw public key\n"
+            "  \"iscompressed\" : true|false,  (boolean) DEPRECATED. If the address is compressed\n"
             "  \"account\" : \"account\"         (string) DEPRECATED. The account associated with the address, \"\" is the default account\n"
```
TheBlueMatt: You removed this from the output, but not the docs. For now I think we should leave the account display here, primarily just because we will want to keep it identical but just rename it label when we do accounts->label conversion.

TheBlueMatt (commented):
Thanks for tackling one of the ifdef WALLETs!

src/rpc/misc.cpp
achow101: The idea was to group the deprecated things together.

src/wallet/rpcwallet.cpp
achow101: Done

src/wallet/rpcwallet.cpp
achow101: Done

src/wallet/rpcwallet.cpp
achow101: Added a brief test for `getaddressinfo` in `wallet.py`

src/wallet/rpcwallet.cpp
achow101: Removed the `pwallet` check.

src/rpc/misc.cpp
```diff
@@ -31,94 +31,47 @@
...
-
 UniValue validateaddress(const JSONRPCRequest& request)
 {
     if (request.fHelp || request.params.size() != 1)
         throw std::runtime_error(
             "validateaddress \"address\"\n"
             "\nReturn information about the given bitcoin address.\n"
+            "DEPRECATION WARNING: Parts of this command have been deprecated and moved to getaddressinfo. Clients must\n"
+            "transition to using getaddressinfo to access this information before upgrading to v0.17. To have those deprecated\n"
+            "information shown, start bitcoind with -deprecatedrpc=validateaddress"
```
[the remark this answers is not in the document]
achow101: Done

src/rpc/misc.cpp
achow101: Added the accounts field back to the output.

src/wallet/rpcwallet.cpp
```diff
@@ -3180,6 +3181,131 @@ UniValue generate(const JSONRPCRequest& request)
...
+            "\nArguments:\n"
+            "1. \"address\"                    (string, required) The bitcoin address to get the information of.\n"
+            "\nResult:\n"
+            "{\n"
+            "  \"address\" : \"address\",        (string) The bitcoin address validated\n"
+            "  \"scriptPubKey\" : \"hex\",       (string) The hex encoded scriptPubKey generated by the address\n"
+            "  \"ismine\" : true|false,        (boolean) If the address is yours or not\n"
+            "  \"iswatchonly\" : true|false,   (boolean) If the address is watchonly\n"
+            "  \"isscript\" : true|false,      (boolean) If the key is a script\n"
+            "  \"script\" : \"type\"             (string, optional) The output script type. Possible types: nonstandard, pubkey, pubkeyhash, scripthash, multisig, nulldata, witness_v0_keyhash, witness_v0_scripthash\n"
```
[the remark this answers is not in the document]
achow101: I changed that field to `redeemscript`. I didn't make it a separate object.

src/wallet/rpcwallet.cpp
achow101: Done.

achow101:
Addressed @TheBlueMatt's and @promag's comments. I also changed two of the `pushKV`s back to `push_back(Pair(...))` because they were not properly showing booleans (they were displayed as 0 and 1).

MarcoFalke:
Needs rebase

test/functional/wallet.py
```diff
@@ -426,6 +426,15 @@ def run_test(self):
 
         # Verify nothing new in wallet
         assert_equal(total_txs, len(self.nodes[0].listtransactions("*",99999)))
+
```
MarcoFalke: No trailing whitespace, please.

achow101: Removed.

achow101:
Rebased onto master.

achow101:
n'th rebase

src/rpc/misc.cpp
```diff
@@ -32,176 +32,71 @@
...
             "  \"isscript\" : true|false,      (boolean) If the key is a script\n"
-            "  \"script\" : \"type\"             (string, optional) The output script type. Possible types: nonstandard, pubkey, pubkeyhash, scripthash, multisig, nulldata, witness_v0_keyhash, witness_v0_scripthash\n"
-            "  \"hex\" : \"hex\",                (string, optional) The redeemscript for the p2sh address\n"
-            "  \"addresses\"                   (string, optional) Array of addresses associated with the known redeemscript\n"
+            "  \"ismine\" : true|false,        (boolean) DEPRECATED. If the address is yours or not\n"
+            "  \"iswatchonly\" : true|false,   (boolean) DEPRECATED. If the address is watchonly\n"
+            "  \"iswitness\" : true|false,     (boolean) DEPRECATED. If the address is a witness address\n"
+            "  \"witness_version\" : version   (numeric, optional) DEPRECATED. The version number of the witness program\n"
+            "  \"witness_program\" : \"hex\"   (string, optional) DEPRECATED. The hex value of the witness program\n"
+            "  \"isscript\" : true|false,      (boolean) DEPRECATED. If the key is a script\n"
```
MeshCollider: Is this deprecated or not? `isscript` is on line 51 too which hasn't been deleted, and in the if-statement below `Pair("isscript", scriptPubKey.IsPayToScriptHash())` is pushed on line 98. I think the deprecation is correct, so both those lines should be removed?

src/wallet/rpcwallet.cpp
```diff
@@ -3287,6 +3288,179 @@ UniValue rescanblockchain(const JSONRPCRequest& request)
...
+            "  \"redeemscript\" : \"type\"       (string, optional) The type of the redeemscript. Possible types: nonstandard, pubkey, pubkeyhash, scripthash, multisig, nulldata, witness_v0_keyhash, witness_v0_scripthash\n"
+            "  \"hex\" : \"hex\",                (string, optional) The redeemscript for the p2sh address\n"
+            "  \"addresses\"                   (string, optional) Array of addresses associated with the known redeemscript\n"
+            "    [\n"
+            "      \"address\"\n"
+            "      ,...\n"
+            "    ]\n"
+            "  \"sigsrequired\" : xxxxx        (numeric, optional) Number of signatures required to spend multisig output\n"
+            "  \"pubkey\" : \"publickeyhex\",    (string) The hex value of the raw public key\n"
+            "  \"iscompressed\" : true|false,  (boolean) If the address is compressed\n"
```
MeshCollider: Should be `optional`. Same as `pubkey` on line above.
Then make same changes to the deprecated fields in `validateaddress` so they're consistent.

src/rpc/misc.cpp
```diff
@@ -32,176 +32,71 @@
...
+            "  \"redeemscript\" : \"type\"       (string, optional) DEPRECATED. The type of the redeemscript. Possible types: nonstandard, pubkey, pubkeyhash, scripthash, multisig, nulldata, witness_v0_keyhash, witness_v0_scripthash\n"
```
MeshCollider: If the deprecation switch is enabled, should `redeemscript` be named `script` for compatibility in the output of `validateaddress`? Otherwise its a breaking change anyway

meshcollider (commented):
utACK https://github.com/bitcoin/bitcoin/pull/10583/commits/aa32c679e75e18996a26f2d4659f91cfa5654964, generally looks good, modulo some feedback inline.

In the case of P2SH-P2WPKH and P2SH-P2WSH, IMO the output should include the `witness_version` and `witness_program` of the internal witness too. Although sipa mentioned he may have done this or something similar in 11403 I think

The deprecation warning mentions v0.17 specifically, perhaps this should have a 0.16.0 milestone, or that may need to be modified

Needs another rebase too sorry :)

src/rpc/misc.cpp
achow101: I think that `isscript` shouldn't be deprecated, so I fixed that

src/rpc/misc.cpp
achow101: Done

src/wallet/rpcwallet.cpp
achow101: Done.

achow101:
Rebased and addresses @MeshCollider's comments.

The P2SH nested witness info is being done in #11403.

jnewbery:
Sorry for the late feedback after several rebases, but looking at this again I don't really like DescribeAddressVisitor moving into bitcoin-wallet, since there's functionality in there that's useful even when running with wallet disabled.

Can you take a look at https://github.com/jnewbery/bitcoin/commits/pr10583.1 . It's almost the same as this, but I've moved the generic parts of DescribeAddressVisitor into rpc/util, so they can be accessed by bitcoin-server.

achow101:
@jnewbery I'll take a look.

achow101:
I cherry-picked @jnewbery's commit for splitting and moving `DescribeAddressVisitor`.

jnewbery:
I think this is better, but obviously needs more review from others.

Incidentally, I think that you've made this more difficult to review by squashing my commits down into a single commit (having a commit that splits a function into two *and* moves half of it to a new file is more difficult to review than one commit to split and a second commit that is move-only). You've also:

- re-added the `DEPRECATED` notice to `iswitness`, `witness_version` and `witness_program` when they're not actually now deprecated
- re-introduced the `script`/`redeemscript` error in the help text.

achow101:
I cherry-picked the split and move commits from @jnewbery and I think I fixed the regressions

They actually weren't squashed earlier. Rather I had only cherry picked one commit due to merge conflict stuff. That is resolved by taking the commits before mine.

achow101:
Rebased this but there's a linker error I can't quite figure out.

MeshCollider:
@achow101: I guess it might be because you're adding rpc/util.cpp to libbitcoin_server_a_SOURCES rather than libbitcoin_util_a_SOURCES in the Makefile.am
Also the scripted diff seems to have an issue

achow101:
Rebased and fixed the linker error.

achow101:
Rebased

src/rpc/util.cpp
```diff
@@ -66,3 +66,55 @@ CScript CreateMultisigRedeemscript(const int required, const std::vector<CPubKey
...
     return result;
 }
+
+UniValue DescribeAddressVisitor::operator()(const CNoDestination &dest) const {
+    return UniValue(UniValue::VOBJ);
+}
+
+UniValue DescribeAddressVisitor::operator()(const CKeyID &keyID) const {
+    UniValue obj(UniValue::VOBJ);
+    CPubKey vchPubKey;
```
jamesob: In https://github.com/bitcoin/bitcoin/pull/10583/commits/4dadd7c7ac0934447b8ca33942535cc8a22e78f5

Unused variable?

src/rpc/util.cpp
```diff
@@ -66,3 +66,55 @@ CScript CreateMultisigRedeemscript(const int required, const std::vector<CPubKey
...
+    obj.push_back(Pair("isscript", false));
+    obj.push_back(Pair("iswitness", false));
+    return obj;
+}
+
+UniValue DescribeAddressVisitor::operator()(const CScriptID &scriptID) const {
+    UniValue obj(UniValue::VOBJ);
+    CScript subscript;
```
jamesob: In https://github.com/bitcoin/bitcoin/pull/10583/commits/4dadd7c7ac0934447b8ca33942535cc8a22e78f5

Unused?

src/rpc/util.cpp
```diff
@@ -66,3 +66,55 @@ CScript CreateMultisigRedeemscript(const int required, const std::vector<CPubKey
...
+    obj.push_back(Pair("isscript", true));
+    obj.push_back(Pair("iswitness", false));
+    return obj;
+}
+
+UniValue DescribeAddressVisitor::operator()(const WitnessV0KeyHash& id) const
+{
+    UniValue obj(UniValue::VOBJ);
+    CPubKey pubkey;
```
jamesob: In https://github.com/bitcoin/bitcoin/pull/10583/commits/4dadd7c7ac0934447b8ca33942535cc8a22e78f5

Unused?

src/rpc/util.cpp
```diff
@@ -66,3 +66,55 @@ CScript CreateMultisigRedeemscript(const int required, const std::vector<CPubKey
...
+    obj.push_back(Pair("iswitness", true));
+    obj.push_back(Pair("witness_version", 0));
+    obj.push_back(Pair("witness_program", HexStr(id.begin(), id.end())));
+    return obj;
+}
+
+UniValue DescribeAddressVisitor::operator()(const WitnessV0ScriptHash& id) const
+{
+    UniValue obj(UniValue::VOBJ);
+    CScript subscript;
```
jamesob: In https://github.com/bitcoin/bitcoin/pull/10583/commits/4dadd7c7ac0934447b8ca33942535cc8a22e78f5

Unused?

src/rpc/util.cpp
```diff
@@ -66,3 +66,55 @@ CScript CreateMultisigRedeemscript(const int required, const std::vector<CPubKey
...
+    obj.push_back(Pair("iswitness", true));
+    obj.push_back(Pair("witness_version", 0));
+    obj.push_back(Pair("witness_program", HexStr(id.begin(), id.end())));
+    return obj;
+}
+
+UniValue DescribeAddressVisitor::operator()(const WitnessUnknown& id) const
+{
+    UniValue obj(UniValue::VOBJ);
+    CScript subscript;
```
jamesob: In https://github.com/bitcoin/bitcoin/pull/10583/commits/4dadd7c7ac0934447b8ca33942535cc8a22e78f5

Unused?

src/rpc/util.h
```diff
@@ -16,4 +23,17 @@ CPubKey HexToPubKey(const std::string& hex_in);
 CPubKey AddrToPubKey(CKeyStore* const keystore, const std::string& addr_in);
 CScript CreateMultisigRedeemscript(const int required, const std::vector<CPubKey>& pubkeys);
 
+class DescribeAddressVisitor : public boost::static_visitor<UniValue>
```
jamesob: In https://github.com/bitcoin/bitcoin/pull/10583/commits/4dadd7c7ac0934447b8ca33942535cc8a22e78f5

Curious: if we're looking to eventually get away from using Boost, should we reserve continued/new usage of it for cases when the feature at hand would be hard to replicate on our own? Seems like each use of
```
boost::apply_visitor(DescribeAddressVisitor(), foo);
```
could be replaced with `DescribeAddressVisitor()(foo)`. 

I guess usages of `static_visitor` are relatively easy to unwind if we ever want to, and it may be worthwhile to remain consistent with our usage of it in the wallet code.

Feel free to ignore; not proposing we hold this changeset up on the basis of Boost usage, just curious.

jamesob (commented):
Tested concept ACK https://github.com/bitcoin/bitcoin/pull/10583/commits/1de28c236c593fc3af770e901ae5d2977370be3f

A few small comments attached. I think the only blocking ones are about unused variables.

achow101:
Removed the unused variables.

jamesob:
ACK https://github.com/bitcoin/bitcoin/pull/10583/commits/8a82bf5dc6355486e1b5a86346812c33bc454a1d

Nice work (and a heroic amount of rebasing).

MeshCollider:
Hopefully this can go in soon now 0.16 is branched off, looks like a painful one to keep rebasing. Will re-review as soon as I can

achow101:
Rebased yet again.

instagibbs (approved):
utACK

unused variables don't break anything and quickly disappear, just noting them

sipa:
Looking very good, and the commits are easy to follow. A few comments:
* The unused variables @instagibbs pointed out are removed after your move commit, but not before.
* I think `addresses` and `pubkeys` also moved to `getaddressinfo`, but aren't listed in the help.
* Do we need to keep the full description of all deprecated fields in `validateaddress`? I think you could just say at the top "DEPRECATION WARNING: The following fields have moved to getaddressinfo and are only included here when running with -deprecatedrpc=validateaddress: ismine, iswatchonly, script, sigsrequired, pubkey, embedded, iscompressed, timestamp, hdkeypath, hdmasterkeyid".
* Can we remove `addresses` from `getaddressinfo` at the same time? It could remain in `validateaddress` when `-deprecatedrpc=validateaddress` is given, but this is the perfect opportunity to remove it in favor of `pubkeys` for new users.

achow101:
- Removed the extra unused variables
 - Shortened `validateaddress` help
 - Set `addresses` to only be shown if `-deprecatedrpc=validateaddress` is set.

src/rpc/misc.cpp
```diff
@@ -33,221 +33,56 @@
...
 UniValue validateaddress(const JSONRPCRequest& request)
 {
     if (request.fHelp || request.params.size() != 1)
         throw std::runtime_error(
             "validateaddress \"address\"\n"
             "\nReturn information about the given bitcoin address.\n"
+            "DEPRECATION WARNING: Parts of this command have been deprecated and moved to getaddressinfo. Clients must\n"
+            "transition to using getaddressinfo to access this information before upgrading to v0.18. The following deprecated\n"
+            "fields have moved to getaddressinfo and will only be shown here with -deprecatedrpc=validateaddress: ismine, iswatchonly,\n"
+            "script, hex, pubkeys, sigsrequired, pubkey, embedded, iscompressed, account, timestamp, hdkeypath, kdmasterkeyid.\n"
```
sipa: There probably still needs to be a mention of `addresses`, as there is now no information about it anywhere anymore.

src/wallet/rpcwallet.cpp
```diff
@@ -3515,6 +3516,200 @@ UniValue rescanblockchain(const JSONRPCRequest& request)
...
+                pubkeys.push_back(HexStr(key.begin(), key.end()));
+            }
+            obj.pushKV("pubkeys", std::move(pubkeys));
+        }
+
+        // The "addresses" field is confusing because it refers to public keys using their P2PKH address.
+        // For that reason, only add the 'addresses' field when needed for backward compatibility. New applications
+        // can use the 'embedded'->'address' field for P2SH or P2WSH wrapped addresses, and 'pubkeys' for
+        // inspecting multisig participants.
+        if (IsDeprecatedRPCEnabled("validateaddress")) obj.pushKV("addresses", std::move(a));
```
sipa: This will include the `addresses` field when `-deprecatedrpc=validateaddress` is set, even for witness addresses (currently it will only include it for legacy addresses).

src/rpc/util.h
sipa: I think you can keep the `DescribeAddressVisitor` class entirely local to `rpc/util.cpp`, and only expose a `UniValue DescribeAddress(const CTxDestination& dest)` function.

src/rpc/misc.cpp
achow101: Added mention of it back.

src/wallet/rpcwallet.cpp
achow101: I changed it back to the original one and used `IsDeprecatedRPC` in a different place to handle this issue.

src/rpc/util.h
achow101: Changed this and `DescribeWalletAddressVisitor` to helper functions which called the visitors.

src/rpc/util.cpp
```diff
@@ -66,3 +66,61 @@ CScript CreateMultisigRedeemscript(const int required, const std::vector<CPubKey
...
+        obj.push_back(Pair("witness_version", (int)id.version));
+        obj.push_back(Pair("witness_program", HexStr(id.program, id.program + id.length)));
+        return obj;
+    }
+};
+
+UniValue DescribeAddress(CTxDestination dest)
+{
+    return boost::apply_visitor(DescribeAddressVisitor(), dest);
+}
```
promag: Nit, add newline.

src/rpc/util.cpp
```diff
@@ -66,3 +66,61 @@ CScript CreateMultisigRedeemscript(const int required, const std::vector<CPubKey
...
+    }
+
+    UniValue operator()(const CKeyID &keyID) const {
+        UniValue obj(UniValue::VOBJ);
+        obj.push_back(Pair("isscript", false));
+        obj.push_back(Pair("iswitness", false));
+        return obj;
+    }
+
+    UniValue operator()(const CScriptID &scriptID) const {
```
promag: Nit, `const CScriptID& scriptID`. `{` on new line.

src/rpc/util.cpp
```diff
@@ -66,3 +66,61 @@ CScript CreateMultisigRedeemscript(const int required, const std::vector<CPubKey
 
     return result;
 }
+
+class DescribeAddressVisitor : public boost::static_visitor<UniValue>
+{
+public:
+    explicit DescribeAddressVisitor() {}
+
+    UniValue operator()(const CNoDestination &dest) const {
```
promag: Nit, `const CNoDestination& dest `. `{` on new line.

src/rpc/misc.cpp
```diff
@@ -33,221 +33,56 @@
...
-            "  \"iscompressed\" : true|false,  (boolean) If the address is compressed\n"
-            "  \"account\" : \"account\"         (string) DEPRECATED. The account associated with the address, \"\" is the default account\n"
-            "  \"timestamp\" : timestamp,      (number, optional) The creation time of the key if available in seconds since epoch (Jan 1 1970 GMT)\n"
-            "  \"hdkeypath\" : \"keypath\"       (string, optional) The HD keypath if the key is HD and available\n"
-            "  \"hdmasterkeyid\" : \"<hash160>\" (string, optional) The Hash160 of the HD master pubkey\n"
+            "  \"isscript\" : true|false,      (boolean) If the key is a script\n"
+            "  \"iswitness\" : true|false,     (boolean) If the address is a witness address\n"
+            "  \"witness_version\" : version   (numeric, optional) The version number of the witness program\n"
+            "  \"witness_program\" : \"hex\"   (string, optional) The hex value of the witness program\n"
```
promag: Fix alignment.

src/Makefile.am
```diff
@@ -335,6 +334,7 @@ libbitcoin_common_a_SOURCES = \
   scheduler.cpp \
   script/sign.cpp \
   script/standard.cpp \
+  script/ismine.cpp \
```
promag: Nit, sort.

src/Makefile.am
```diff
@@ -389,10 +389,10 @@ endif
 
 bitcoind_LDADD = \
   $(LIBBITCOIN_SERVER) \
+  $(LIBBITCOIN_WALLET) \
```
promag: Why this change? Does the order matter? 🙄

src/rpc/util.h
```diff
@@ -16,4 +23,6 @@ CPubKey HexToPubKey(const std::string& hex_in);
 CPubKey AddrToPubKey(CKeyStore* const keystore, const std::string& addr_in);
 CScript CreateMultisigRedeemscript(const int required, const std::vector<CPubKey>& pubkeys);
 
+UniValue DescribeAddress(CTxDestination dest);
```
promag: `const CTxDestination& dest`.

src/wallet/rpcwallet.cpp
```diff
@@ -27,6 +27,7 @@
 #include <wallet/feebumper.h>
 #include <wallet/wallet.h>
 #include <wallet/walletdb.h>
+#include <wallet/rpcwallet.h>
```
promag: Nit, sort.

src/wallet/rpcwallet.cpp
```diff
@@ -3515,6 +3516,207 @@ UniValue rescanblockchain(const JSONRPCRequest& request)
     return response;
 }
 
+class DescribeWalletAddressVisitor : public boost::static_visitor<UniValue>
+{
+public:
+    CWallet * const pwallet;
```
promag: Nit, m_wallet, why public?

promag (commented):
Needs rebase, some comments (mostly nits).

src/rpc/util.cpp
```diff
@@ -66,3 +66,61 @@ CScript CreateMultisigRedeemscript(const int required, const std::vector<CPubKey
...
+        return UniValue(UniValue::VOBJ);
+    }
+
+    UniValue operator()(const CKeyID &keyID) const {
```
promag: Nit, `const CKeyID& keyID`. `{` on new line.

src/Makefile.am
achow101: Yes, the order matters

src/wallet/rpcwallet.cpp
achow101: why not?

src/rpc/util.cpp
achow101: Done

src/rpc/util.cpp
achow101: Done

src/rpc/util.cpp
achow101: Done

src/Makefile.am
achow101: Done

src/rpc/util.h
achow101: Done

src/rpc/util.cpp
achow101: Done

src/wallet/rpcwallet.cpp
achow101: Done

src/rpc/misc.cpp
achow101: Done

src/wallet/rpcwallet.cpp
promag: See `CAffectedKeysVisitor`, `DestinationEncoder` and `CScriptVisitor` with private members. On the other hand `Witnessifier` members are public (could be "fixed") .

IMO all members, by default, should be private. When reading the code it's clear that a private member is not externally accessed/changed.

src/wallet/rpcwallet.cpp
```diff
@@ -3514,6 +3515,207 @@ UniValue rescanblockchain(const JSONRPCRequest& request)
...
+        }
+        return obj;
+    }
+
+    UniValue operator()(const WitnessUnknown& id) const { return UniValue(UniValue::VOBJ); }
+};
+
+UniValue DescribeWalletAddress(CWallet* pwallet, const CTxDestination& dest)
+{
+    UniValue ret(UniValue::VOBJ);
```
promag: Nit, could `AssertLockHeld(pwallet->cs_wallet)`.

src/wallet/rpcwallet.cpp
```diff
@@ -3514,6 +3515,207 @@ UniValue rescanblockchain(const JSONRPCRequest& request)
...
+            obj.pushKV("pubkey", HexStr(vchPubKey));
+            obj.pushKV("iscompressed", vchPubKey.IsCompressed());
+        }
+        return obj;
+    }
+
+    UniValue operator()(const CScriptID &scriptID) const {
+        UniValue obj(UniValue::VOBJ);
+        CScript subscript;
+        if (pwallet && pwallet->GetCScript(scriptID, subscript)) {
```
promag: In this context `pwallet` is valid so:
```cpp
if (pwallet->GetCScript(scriptID, subscript)) {
```

src/wallet/rpcwallet.cpp
```diff
@@ -3514,6 +3515,207 @@ UniValue rescanblockchain(const JSONRPCRequest& request)
...
+    }
+
+    explicit DescribeWalletAddressVisitor(CWallet *_pwallet) : pwallet(_pwallet) {}
+
+    UniValue operator()(const CNoDestination &dest) const { return UniValue(UniValue::VOBJ); }
+
+    UniValue operator()(const CKeyID &keyID) const {
+        UniValue obj(UniValue::VOBJ);
+        CPubKey vchPubKey;
+        if (pwallet && pwallet->GetPubKey(keyID, vchPubKey)) {
```
promag: In this context `pwallet` is valid so:
```cpp
if (pwallet->GetCScript(scriptID, subscript)) {
```

src/wallet/rpcwallet.cpp
```diff
@@ -3514,6 +3515,207 @@ UniValue rescanblockchain(const JSONRPCRequest& request)
...
+        return obj;
+    }
+
+    UniValue operator()(const WitnessV0ScriptHash& id) const
+    {
+        UniValue obj(UniValue::VOBJ);
+        CScript subscript;
+        CRIPEMD160 hasher;
+        uint160 hash;
+        hasher.Write(id.begin(), 32).Finalize(hash.begin());
```
promag: Unless you don't want to touch moved code, remove `hasher` and just do:
```cpp
CRIPEMD160().Write(id.begin(), 32).Finalize(hash.begin());
```

src/wallet/rpcwallet.cpp
```diff
@@ -3514,6 +3515,207 @@ UniValue rescanblockchain(const JSONRPCRequest& request)
...
+        if (pwallet && pwallet->GetCScript(CScriptID(hash), subscript)) {
```
promag: In this context `pwallet` is valid so:
```cpp
if (pwallet->GetCScript(CScriptID(hash), subscript)) {
```

src/rpc/misc.cpp
```diff
@@ -33,221 +33,56 @@
...
-                }
-            }
-            if (meta) {
-                ret.pushKV("timestamp", meta->nCreateTime);
-                if (!meta->hdKeypath.empty()) {
-                    ret.pushKV("hdkeypath", meta->hdKeypath);
-                    ret.pushKV("hdmasterkeyid", meta->hdMasterKeyID.GetHex());
-                }
-            }
+        if (!::vpwallets.empty() && IsDeprecatedRPCEnabled("validateaddress")) {
```
promag: Why `!::vpwallets.empty()`? There is alwasy one wallet right?

src/rpc/util.cpp
```diff
@@ -66,3 +66,64 @@ CScript CreateMultisigRedeemscript(const int required, const std::vector<CPubKey
...
+{
+public:
+    explicit DescribeAddressVisitor() {}
+
+    UniValue operator()(const CNoDestination &dest) const 
+    {
+        return UniValue(UniValue::VOBJ);
+    }
+
+    UniValue operator()(const CKeyID &keyID) const
```
promag: `&` before space.

src/rpc/util.cpp
```diff
@@ -66,3 +66,64 @@ CScript CreateMultisigRedeemscript(const int required, const std::vector<CPubKey
...
+    {
+        UniValue obj(UniValue::VOBJ);
+        obj.pushKV("isscript", false);
+        obj.pushKV("iswitness", false);
+        return obj;
+    }
+
+    UniValue operator()(const CScriptID &scriptID) const
```
promag: `&` before space.

src/wallet/rpcwallet.cpp
```diff
@@ -3514,6 +3515,207 @@ UniValue rescanblockchain(const JSONRPCRequest& request)
...
+            ProcessSubScript(subscript, obj, IsDeprecatedRPCEnabled("validateaddress"));
+        }
+        return obj;
+    }
+
+    UniValue operator()(const WitnessV0KeyHash& id) const
+    {
+        UniValue obj(UniValue::VOBJ);
+        CPubKey pubkey;
+        if (pwallet && pwallet->GetPubKey(CKeyID(id), pubkey)) {
```
promag: In this context `pwallet` is valid so:
```cpp
if (pwallet->GetPubKey(CKeyID(id), pubkey)) {
```

promag (commented):
Lightly tested ACK. Please consider the following comments.

jnewbery:
@promag - personally, I'd generally discourage nits after a PR has been open for many months and been rebased/re-reviewed many times already. Adding additional churn and workload to the contributor and reviewers at this point seems borderline sadistic :slightly_smiling_face: 

(just my opinion though)

src/rpc/util.cpp
```diff
@@ -66,3 +66,64 @@ CScript CreateMultisigRedeemscript(const int required, const std::vector<CPubKey
 
     return result;
 }
+
+class DescribeAddressVisitor : public boost::static_visitor<UniValue>
+{
+public:
+    explicit DescribeAddressVisitor() {}
+
+    UniValue operator()(const CNoDestination &dest) const
```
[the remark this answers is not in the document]
instagibbs: travis linter is complaining about trailing whitespace.... I don't know protocol if it matters but just reporting where the complaint is coming from

src/rpc/misc.cpp
achow101: No, you can run with `-disablewallet` which will disable wallets and thus `vpwallets` will be empty.

achow101:
Addressed some of @promag's nits. I don't really feel like fixing the other ones. It's also getting really tedious to do fix nits at this point, so I probably won't do so in the future for this PR.

jnewbery:
Tested ACK b22cce014852b082d80f1cc35f902b375cba0318. Travis failure looks unrelated.

Let's get this merged!

sipa:
utACK b22cce014852b082d80f1cc35f902b375cba0318

I have a few minor improvements, but I'll submit them as a separate PR after merge, nothing worth holding this up for.

promag:
@jnewbery @achow101 sorry! didn't want to cause such pain :) I did report tested ACK.

jonasschnelli:
Tested ACK b22cce014

PR #10584, opened 2017-06-13
Remove unused ResendWalletTransactions notification

ryanofsky:


sipa:
We're not rebroadcasting unconfirmed transactions anymore? :open_mouth:

gmaxwell:
wtf and why are tests passing?

instagibbs:
There's no test for automatic rebroadcasting, AFAICT, only manual?

jonasschnelli:
IMO we still do re-broadcast old transactions: https://github.com/bitcoin/bitcoin/blob/master/src/net_processing.cpp#L2907

This PR does only remove the connection from the main signals to the validation interface (the wallet) while the PR does not remove the signal (and the emit of the signal) itself.

Seems NACKish to me.

I think we do rebroadcast fine. But we have no tests for it (except of the intentional rebroadcast via RPC call).

laanwj:
NACK.

This code is *not* unused.
- `ResendWalletTransactions` is connected to `g_signals.Broadcast`
- `Broadcast` is invoked from here: https://github.com/bitcoin/bitcoin/blob/master/src/net_processing.cpp#L2907

Better tests would be obviously welcome.

ryanofsky:
Oops, dumb mistake. The PR just didn't delete the signal it was connecting to, which is why it compiled.

PR #10585, opened 2017-06-14
[WIP] PoW connection slots

kallewoof:
This is a design idea for the PoW connection slots ([BIP-154](https://github.com/bitcoin/bips/blob/master/bip-0154.mediawiki)).

I am mostly looking for feedback on the approach, and not so much on the quality of the code itself, at this point (but code feedback is of course also appreciated).

Does this seem like a reasonable approach?

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -126,6 +126,7 @@ BITCOIN_CORE_H = \
   pow/cuckoo_cycle/cuckoo.h \
   pow/cuckoo_cycle/siphash.h \
   pow/pow.h \
+  pow/sha256/sha256.h \
   pow.h \
   protocol.h \
   random.h \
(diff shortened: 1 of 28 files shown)


midnightmagic:
Consider making it possible for Bitcoin miners (that some of us have lying around) to help generate some form of these proofs, if it isn't already possible to do so.

kallewoof:
@midnightmagic Most Bitcoin miners are only able to do SHA256(SHA256) proofs -- someone could pose such a challenge if they wanted to and it could theoretically be used with e.g. a USB miner, but I don't know how frequent that will be. A key reason why Cuckoo Cycle was included is because it does not give users with miners an advantage (as this would give attackers an opportunity to DDoS with help of some HW miners).

midnightmagic:
HW miners are only owned by Bitcoin miners. Otherwise, botnets will simply be the empowered ones and normal people will not -- PoW demonstrates investment in Bitcoin mining, much like the PoW of the blockchain itself. :)

PR #10586, opened 2017-06-14
More economical fee estimates for opt-in-RBF transactions

morcos:
This PR takes advantage of the new fee estimation feature that will potentially give lower estimates if recent market conditions warrant it.  The logic used here is that any time a transaction signals opt-in-RBF and uses automatic fee estimation then it will use the non-conservative estimate.   Transactions which do not signal opt-in-RBF will still use the default conservative estimate.

In a nutshell conservative estimates require that your fee rate would meet the necessary confirmation threshold for double your target at longer time horizons as well.  This reduces the likelihood that you place a transaction just as the market is starting to get busy again that ends up being stuck for a very long time.

I'm working on a follow-on PR which will allow the specification of transaction confirmation target and whether the estimate should be conservative or not on a per-RPC call basis for sendtoaddress, sendmany, and fundrawtransaction.

I'd love it if someone else working on the QT ability to force estimates to be conservative or not (overriding the RBF implied default).

~~The first commit is #10582 which is a pre-existing bug fix~~

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -2722,7 +2722,10 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CWalletT
                 if (coinControl && coinControl->nConfirmTarget > 0)
                     currentConfirmationTarget = coinControl->nConfirmTarget;
 
-                CAmount nFeeNeeded = GetMinimumFee(nBytes, currentConfirmationTarget, ::mempool, ::feeEstimator, &feeCalc, false /* ignoreGlobalPayTxFee */);
+                // Allow to override the default fee estimate mode over the CoinControl instance
+                bool conservative_estimate = CalculateEstimateType(coinControl ? coinControl->m_fee_mode : FeeEstimateMode::UNSET);
+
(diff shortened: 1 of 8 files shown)


morcos:
Rebased due to conflict with #10422

laanwj:
Nice!

TheBlueMatt (commented):
I really have no ability to review the Qt changes here, but generally looks awesome.

src/wallet/wallet.cpp
```diff
@@ -4145,3 +4148,15 @@ bool CMerkleTx::AcceptToMemoryPool(const CAmount& nAbsurdFee, CValidationState&
 {
     return ::AcceptToMemoryPool(mempool, state, tx, true, NULL, NULL, false, nAbsurdFee);
 }
+
+bool CalculateEstimateType(FeeEstimateMode mode, bool opt_in_rbf) {
```
morcos: I think it makes more sense to let it be generic and ideally estimatesmartfee would take a FeeEstimateMode and this functions purpose would be to take any requested mode and other wallet parameters and use wallet logic to determine a final mode.  Rather than make all those changes now while not yet needed though, I think I'll just change the RPC interface to conservative estimates to take a string as per https://github.com/bitcoin/bitcoin/pull/10589#pullrequestreview-45713269

morcos:
Closing this since it's contained in #10589.

PR #10587, opened 2017-06-14
Net: Fix resource leak in ReadBinaryFile(...)

practicalswift:
Potential file descriptor leak introduced in commit 0b6f40d4cabb3bebf551a49a69ce36d4b0375b6a via PR #10408 (submitted 2017-05-16, merged 2017-05-18).

### DIFF
--- a/src/torcontrol.cpp
+++ b/src/torcontrol.cpp
@@ -376,8 +376,10 @@ static std::pair<bool,std::string> ReadBinaryFile(const fs::path &filename, size
     while ((n=fread(buffer, 1, sizeof(buffer), f)) > 0) {
         // Check for reading errors so we don't return any data if we couldn't
         // read the entire file (or up to maxsize)
-        if (ferror(f))
+        if (ferror(f)) {
+            fclose(f);
             return std::make_pair(false,"");
+        }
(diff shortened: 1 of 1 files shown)


paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/10587/commits/f2fb132cb09efb253df44db09035d4210cc61236

theuni:
utACK f2fb132cb09efb253df44db09035d4210cc61236, but RAII would be nice here.

sipa:
@theuni Agree, but since the existing code is using fopen/fread/fclose, let's keep using that in the bug fix. The rest can happen later.

utACK f2fb132cb09efb253df44db09035d4210cc61236

practicalswift:
Friendly ping @str4d – you might want this fix in zcash too :-)

str4d:
Thanks @practicalswift, nice catch!

laanwj:
> utACK f2fb132, but RAII would be nice here.

Yes it's pretty easy to make a RAII for `FILE*` using `unique_ptr` with a custom deleter, we could even make our `fsbridge::fopen` return this.
Agree with not doing it in this PR though.

PR #10588, opened 2017-06-14
doc: Note preexisting bug in display of fee calculation in coin control

morcos:
This is a display only bug

@laanwj please add to 0.14.2 if still possible


practicalswift:
utACK 60b967e

laanwj:
Having a "known bugs" as large as "notable changes" makes me feel kind of meh about the whole release.
Why did we want a 0.14.2 in the first place?

PR #10589, opened 2017-06-14
More economical fee estimates for RBF and RPC options to control

morcos:
This PR takes advantage of the new fee estimation feature that will potentially give lower estimates if recent market conditions warrant it. The logic used here is that any time a transaction signals opt-in-RBF and uses automatic fee estimation then it will use the non-conservative estimate. Transactions which do not signal opt-in-RBF will still use the default conservative estimate.

In a nutshell conservative estimates require that your fee rate would meet the necessary confirmation threshold for double your target at longer time horizons as well. This reduces the likelihood that you place a transaction just as the market is starting to get busy again that ends up being stuck for a very long time.

This PR also allows the specification of transaction confirmation target and whether the estimate should be conservative or not on a per-RPC call basis for sendtoaddress, sendmany, bumpfee, and fundrawtransaction using optional named arguments.

Left as an exercise to the reader is adding this to GUI functions to send transactions and bump fee.


RHavar:
Concept ACK, this is really great and something that's been needed for a long time. It'll simplify some of the locking and `settxfee` stuff I've needed to do. 

But it'd be nice if it used an `options` object or something.  Otherwise if I want to specify the `conservative_estimate` I'll need to fill in all the other fields, which I might not want to know?

But I guess that's something that can be done in a different commit, with `if (arguments.length == 3 && typeof arguments == 'object')` style

instagibbs:
@RHavar I had the same exact reaction fwiw. `fee_options` or something.

morcos:
I thought the whole point of named arguments is we didn't have to worry about that any more.  I think I made sure these work correctly with holes so we can use them with named arguments.

RHavar:
Oh yeah, you're right. Guess I should learn how to use named arguments.

sipa:
Needs rebase.

morcos:
Rebased due to conflict with #10422

src/wallet/rpcwallet.cpp
```diff
@@ -909,7 +927,10 @@ UniValue sendmany(const JSONRPCRequest& request)
             "      \"address\"          (string) Subtract fee from this address\n"
             "      ,...\n"
             "    ]\n"
-            "\nResult:\n"
+            "6. opt_in_rbf             (boolean, optional) Allow this transaction to be replaced by a transaction with higher fees\n"
+            "7. conf_target            (numeric, optional) Confirmation target (in blocks)\n"
+            "8. conservative_estimate  (boolean, optional) Use conservative (potentially higher) fee estimation\n"
```
laanwj: I'd prefer to use a string here (which maps to FeeEstimateMode) instead of a boolean. This would also allow adding additional fee estimation strategies in the future without changing the interface.

laanwj:
Concept ACK

morcos:
Switched to an estimate_mode string and squashed
5cbc1f6 ([rpcestimatechoice.ver2](https://github.com/morcos/bitcoin/commits/rpcestimatechoice.ver2)) -> 1bc6a4a ([rpcestimatechoices.ver2.squash](https://github.com/morcos/bitcoin/commits/rpcestimatechoice.ver2.squash))

morcos:
Changed argument name from opt_in_rbf to replaceable for conformance with bumpfee

instagibbs (commented):
utACK https://github.com/bitcoin/bitcoin/pull/10589/commits/ab07a1063ba495faa6029ebfc270e6c52d5a2cc9

src/wallet/wallet.cpp
```diff
@@ -4145,3 +4148,14 @@ bool CMerkleTx::AcceptToMemoryPool(const CAmount& nAbsurdFee, CValidationState&
 {
     return ::AcceptToMemoryPool(mempool, state, tx, true, NULL, NULL, false, nAbsurdFee);
 }
+
+bool CalculateEstimateType(FeeEstimateMode mode) {
```
morcos: https://github.com/bitcoin/bitcoin/pull/10586#discussion_r124370127

instagibbs: If you change the semantics of the return value all the other callers will have to change too. Why not just have the caller consume the returned mode? Anyways, not a blocker.

instagibbs: Ok we agree it should be a mode being returned ideally.

TheBlueMatt (commented):
I still dont really feel comfortable ACKing the Qt changes until I get a chance to test, but this is generally a very welcome improvement.

src/wallet/wallet.cpp
morcos: Please see #10706 
CalculateEstimateType is now only used one place, in GetMinimumFee.

instagibbs:
needs rebase, will review

morcos:
rebased for adjacent line change in #10698 and removed a couple extra \n's 

no other changes

I believe we've settled on removing the CalculateEstimateType function entirely at the end of #10706 since it just becomes a helper for GetMinimumFee at that point.

TheBlueMatt (commented):
Looks good. Definitely want #10706 to go in with this.

src/qt/sendcoinsdialog.cpp
```diff
@@ -166,6 +166,8 @@ void SendCoinsDialog::setModel(WalletModel *_model)
         connect(ui->checkBoxMinimumFee, SIGNAL(stateChanged(int)), this, SLOT(updateFeeSectionControls()));
         connect(ui->checkBoxMinimumFee, SIGNAL(stateChanged(int)), this, SLOT(updateGlobalFeeVariables()));
         connect(ui->checkBoxMinimumFee, SIGNAL(stateChanged(int)), this, SLOT(coinControlUpdateLabels()));
+        connect(ui->optInRBF, SIGNAL(stateChanged(int)), this, SLOT(updateSmartFeeLabel()));
+        connect(ui->optInRBF, SIGNAL(stateChanged(int)), this, SLOT(coinControlUpdateLabels()));
```
morcos: Not sure if I understand your question (or how QT works) but I'm pretty sure it's correct as is.  Calling `setCheckState` below will trigger the signal if the checkbox changed from its initial value.

src/wallet/rpcwallet.cpp
```diff
@@ -416,6 +416,12 @@ UniValue sendtoaddress(const JSONRPCRequest& request)
             "                             transaction, just kept in your wallet.\n"
             "5. subtractfeefromamount  (boolean, optional, default=false) The fee will be deducted from the amount being sent.\n"
             "                             The recipient will receive less bitcoins than you enter in the amount field.\n"
+            "6. replaceable            (boolean, optional) Allow this transaction to be replaced by a transaction with higher fees via BIP 125\n"
+            "7. conf_target            (numeric, optional) Confirmation target (in blocks)\n"
+            "8. \"estimate_mode\"      (string, optional, default=UNSET) The fee estimate mode, must be one of:\n"
```
promag: `fee_estimate_mode`, `fee_mode` or `estimate_mode`?

src/qt/sendcoinsdialog.cpp
promag: Either way, what about make an explicit call to `coinControlUpdateLabels()`?

src/wallet/coincontrol.h
```diff
@@ -26,6 +27,8 @@ class CCoinControl
     int nConfirmTarget;
     //! Signal BIP-125 replace by fee.
     bool signalRbf;
+    //! Fee estimation mode to control arguments to estimateSmartFee
+    FeeEstimateMode m_fee_mode;
```
promag: `m_fee_estimate_mode`?

src/wallet/rpcwallet.cpp
```diff
@@ -401,7 +401,7 @@ UniValue sendtoaddress(const JSONRPCRequest& request)
         return NullUniValue;
     }
 
-    if (request.fHelp || request.params.size() < 2 || request.params.size() > 5)
+    if (request.fHelp || request.params.size() < 2 || request.params.size() > 8)
         throw std::runtime_error(
             "sendtoaddress \"address\" amount ( \"comment\" \"comment_to\" subtractfeefromamount )\n"
```
promag: Missing new arguments.

src/wallet/rpcwallet.cpp
```diff
@@ -910,7 +935,13 @@ UniValue sendmany(const JSONRPCRequest& request)
             "      \"address\"          (string) Subtract fee from this address\n"
             "      ,...\n"
             "    ]\n"
-            "\nResult:\n"
+            "6. replaceable            (boolean, optional) Allow this transaction to be replaced by a transaction with higher fees via BIP 125\n"
```
promag: Add `options` instead of 3 more optional arguments?

src/wallet/rpcwallet.cpp
```diff
@@ -416,6 +416,12 @@ UniValue sendtoaddress(const JSONRPCRequest& request)
             "                             transaction, just kept in your wallet.\n"
             "5. subtractfeefromamount  (boolean, optional, default=false) The fee will be deducted from the amount being sent.\n"
             "                             The recipient will receive less bitcoins than you enter in the amount field.\n"
+            "6. replaceable            (boolean, optional) Allow this transaction to be replaced by a transaction with higher fees via BIP 125\n"
```
promag: Add `options` instead of 3 more optional arguments?

src/wallet/wallet.cpp
```diff
@@ -4154,3 +4157,15 @@ bool CMerkleTx::AcceptToMemoryPool(const CAmount& nAbsurdFee, CValidationState&
 {
     return ::AcceptToMemoryPool(mempool, state, tx, true, NULL, NULL, false, nAbsurdFee);
 }
+
+bool CalculateEstimateType(FeeEstimateMode mode, bool opt_in_rbf) {
```
promag: `IsFeeEstimateConservative`?

promag (commented):
IMO this would be a cleaner implementation
```cpp
bool FeeEstimateModeFromString(const std::string& string, FeeEstimateMode& fee_estimate_mode)
{
    // lookup string
    return false;
}

...


if (!FeeEstimateModeFromString(request.params[7].get_str(), coin_control.m_fee_mode)) {
    throw JSONRPCError(RPC_INVALID_PARAMETER, ...);
}
```

src/wallet/rpcwallet.cpp
promag: I saw the above comment about named arguments. IMO from the client side feels better as an option.

src/qt/sendcoinsdialog.cpp
morcos: I don't think this is necessary.  It's analogous to setting the default sliderSmartFee done in the same section of code.

src/wallet/rpcwallet.cpp
morcos: will do

src/wallet/rpcwallet.cpp
morcos: I believe named arguments are preferable

src/wallet/rpcwallet.cpp
morcos: Hmm..  Yes, I think it might make sense to change this to `fee_mode`, but leave it `estimate_mode` for estimateSmartFee (done in another PR).  Even though they take the same values now, I could imagine they would differ in the future.  Any other thoughts?

src/wallet/wallet.cpp
morcos: Function removed in #10706 (or will be shortly)

src/wallet/rpcwallet.cpp
promag: I just wanted to point out the inconsistencies between arguments, variable names and types.

src/wallet/wallet.cpp
promag: Still, name not clear don't you think?

morcos: yeah but who cares, its going away completely in a couple of commits

src/wallet/rpcwallet.cpp
morcos: I decided leaving it as estimate_mode in both RPC's is preferable. 
The inconsistency between variable and parameter name is not worth the churn imo.

morcos:
Took some of the nits
([rpcestimatechoice.ver3](https://github.com/morcos/bitcoin/commits/rpcestimatechoice.ver3)) -> f135923 ([rpcestimatechoices.ver3.squash](https://github.com/morcos/bitcoin/commits/rpcestimatechoice.ver3.squash))

ryanofsky (commented):
utACK f135923ee2cf1a1a9a436626dc5b9219f8ad97da

src/wallet/rpcwallet.cpp
promag: I understand, but for those that use indexed arguments, they have to pass the middle arguments.

laanwj:
utACK f135923

PR #10591, opened 2017-06-14
[tests] make pruning.py faster

jnewbery:
Headline: Brings time for pruning.py down to ~4 minutes on my VM when run without other tests in parallel.

A couple of modifications to make pruning.py much faster:

- pruning.py doesn't care what kind of transactions make up the big blocks that are pruned in the test. Instead of making blocks with several large, expensive to construct and validate transactions, instead make the large blocks contain a single coinbase transaction with a huge OP_RETURN txout.
- avoid stop-starting nodes where possible.

Builds on top of #10555 and #10556 

My VM has slow disk i/o. This would possibly be faster with faster disk i/o. Unfortunately, pruning.py still isn't able to run on Travis. It appears that disk access is very slow for in the Travis environment, and this test still times out/fails. Unfortunately, it's not possible to run this test without very heavy disk access - the whole point of pruning is to remove block files when disk usage goes over a certain threshold.

I'd be interested to know how this affects runtime for others.

@sdaftuar

### DIFF
--- a/test/functional/blockchain.py
+++ b/test/functional/blockchain.py
@@ -139,13 +139,13 @@ class BlockchainTest(BitcoinTestFramework):
         self.nodes[0].generate(6)
         assert_equal(self.nodes[0].getblockcount(), 206)
         self.log.debug('Node should not stop at this height')
-        assert_raises(subprocess.TimeoutExpired, lambda: self.bitcoind_processes[0].wait(timeout=3))
+        assert_raises(subprocess.TimeoutExpired, lambda: self.nodes[0].process.wait(timeout=3))
         try:
             self.nodes[0].generate(1)
         except (ConnectionError, http.client.BadStatusLine):
             pass  # The node already shut down before response
         self.log.debug('Node should stop at this height...')
-        self.bitcoind_processes[0].wait(timeout=BITCOIND_PROC_WAIT_TIMEOUT)
+        self.nodes[0].process.wait(timeout=BITCOIND_PROC_WAIT_TIMEOUT)
         self.nodes[0] = self.start_node(0, self.options.tmpdir)
         assert_equal(self.nodes[0].getblockcount(), 207)
 
--- a/test/functional/bumpfee.py
+++ b/test/functional/bumpfee.py
@@ -41,8 +41,7 @@ class BumpFeeTest(BitcoinTestFramework):
         self.nodes = self.start_nodes(self.num_nodes, self.options.tmpdir, extra_args)
 
         # Encrypt wallet for test_locked_wallet_fails test
-        self.nodes[1].encryptwallet(WALLET_PASSPHRASE)
-        self.bitcoind_processes[1].wait()
+        self.nodes[1].node_encrypt_wallet(WALLET_PASSPHRASE)
         self.nodes[1] = self.start_node(1, self.options.tmpdir, extra_args[1])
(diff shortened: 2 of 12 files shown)


MarcoFalke:
Concept ACK. Needs rebase

jnewbery:
I'll rebase if/when #10556 is merged, since I expect that may require further feedback and iteration. There's no rush for this PR since pruning.py isn't run by many people. Currently just looking for feedback on the concept and whether this helps reduce runtime for other people.

MarcoFalke:
Since this runs a lot faster, can we also enable it for the travis cron job?

jnewbery:
> Since this runs a lot faster, can we also enable it for the travis cron job?

I don't think so. See the original description:

> Unfortunately, pruning.py still isn't able to run on Travis. It appears that disk access is very slow for in the Travis environment, and this test still times out/fails. Unfortunately, it's not possible to run this test without very heavy disk access - the whole point of pruning is to remove block files when disk usage goes over a certain threshold.

If you can think of a cunning way to get this to run on Travis, that'd be great. I haven't been able to come up with anything.

MarcoFalke:
Needs rebase

jnewbery:
rebased

MarcoFalke:
Does this depend on the TestNode commit?

fanquake:
Testing this:
```
2017-07-06 01:08:57.788000 TestFramework (INFO): New best height: 1032
2017-07-06 01:08:57.788000 TestFramework (INFO): Generating new longer chain of 300 more blocks
2017-07-06 01:08:59.991000 TestFramework (INFO): Verify height on node 2: 1320
2017-07-06 01:08:59.992000 TestFramework (INFO): Usage possibly still high because of stale blocks in block files: 512
2017-07-06 01:08:59.992000 TestFramework (INFO): Mine 220 more large blocks so we have requisite history
2017-07-06 01:11:04.998000 TestFramework (INFO): Usage should be below target: 512
2017-07-06 01:11:04.999000 TestFramework (INFO): Test that we can rerequest a block we previously pruned if needed for a reorg
2017-07-06 01:11:06.476000 TestFramework (ERROR): Assertion failed
  File "/Users/x/Github/bitcoin/test/functional/test_framework/test_framework.py", line 143, in main
    self.run_test()
  File "test/functional/pruning.py", line 436, in run_test
    self.reorg_back()
  File "test/functional/pruning.py", line 211, in reorg_back
    assert_raises_jsonrpc(-1, "Block not available (pruned data)", self.nodes[2].getblock, self.forkhash)
  File "/Users/x/Github/bitcoin/test/functional/test_framework/util.py", line 87, in assert_raises_jsonrpc
AssertionError: No exception raised
2017-07-06 01:11:06.477000 TestFramework (INFO): Stopping nodes
```

MarcoFalke:
@jnewbery Needs rebase

jnewbery:
closing this for now. It seems to work most of the time for me, but intermittently fails for the same reason that @fanquake gives above. There's refactoring work going on in the test_framework right now, so this is likely to require more rebases and it doesn't seem worth the effort given that very few people run this test.

jtimon:
Fwiw, perhaps more people would run this test if it was faster. Personally I avoid it precisely for how slow it is.

adamjonas:
Can remove the up for grabs label with #15686 making this obsolete.

MarcoFalke:
Thanks @adamjonas

PR #10592, opened 2017-06-14
[trivial] fix indentation for ArgsManager class

jnewbery:
Whitespace change only. Cleanup following #9494 

@jtimon 

Hint for reviewers: append `?w=1` to ignore whitespace: https://github.com/bitcoin/bitcoin/pull/10592/files?w=1

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -187,62 +187,63 @@ public:
     void ParseParameters(int argc, const char*const argv[]);
     void ReadConfigFile(const std::string& confPath);
     std::vector<std::string> GetArgs(const std::string& strArg);
-/**
- * Return true if the given argument has been manually set
- *
- * @param strArg Argument to get (e.g. "-foo")
- * @return true if the argument has been set
- */
-bool IsArgSet(const std::string& strArg);
 
-/**
- * Return string argument or default value
- *
- * @param strArg Argument to get (e.g. "-foo")
- * @param default (e.g. "1")
- * @return command-line argument or default value
- */
-std::string GetArg(const std::string& strArg, const std::string& strDefault);
+    /**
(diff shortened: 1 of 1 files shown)


practicalswift:
ACK 78109933357fdc386b75f2629d9345c215f2978f

paveljanik:
trivial ACK 7810993

fanquake:
trivialACK 7810993

ryanofsky:
#10597, which is a similar whitespace-only change, was closed recently because of a developer guide admonition to "not submit patches solely to modify the style of existing code."

Is this policy worth keeping, or maybe modifying to only apply to global changes like #10597 and not local changes like this?

fanquake:
@ryanofsky I think changing it to mention global vs local would be ok. Global whitespace/indentation/etc changes only serve to break pull requests and disrupt other work. Small, contained changes like this, should be fine. (and merged quickly).

theuni:
@fanquake agreed, well put.

jnewbery:
@ryanofsky - I think the developer guide is referring to cases when a new style is adopted by the project or becomes the norm. In those cases, contributors shouldn't open PRs to make existing code conform to the new style (as this leads to constant rebase hell).

In this case, indenting class members has always been the accepted style, and this PR just fixes the indentation that was broken by a previous PR. In fact, the author of that PR said he intended to fix this himself after merge, but that he might forget: https://github.com/bitcoin/bitcoin/pull/9494#discussion_r115340509.

sipa:
Agree with @jnewbery; the indentation here is just broken and needs fixing. That's not the same as updating from one style to another.

jtimon:
fast review ACK. I didn't do it at the same time for #9494 to remain easier to review.

PR #10594, opened 2017-06-14
Bugfix: net: Apply whitelisting criteria to outgoing connections

luke-jr:
Whitelisting currently only works for inbound connections, but would be useful for outgoing just as well.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -404,7 +404,7 @@ std::string HelpMessage(HelpMessageMode mode)
 #endif
 #endif
     strUsage += HelpMessageOpt("-whitebind=<addr>", _("Bind to given address and whitelist peers connecting to it. Use [host]:port notation for IPv6"));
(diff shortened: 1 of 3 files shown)


practicalswift:
Concept ACK, but the [documentation for `-whitelist`](https://github.com/bitcoin/bitcoin/blob/228c319a944b0ba7c835b1909ee1c2056c652eb1/src/init.cpp#L418-L419) would need to be updated updated as well?

luke-jr:
(Updated docs)

practicalswift:
utACK a511b894f310bb88fab32b3004c7a6671e296cee

laanwj:
What is the difference with  #10051 by @rubensayshi?
Anyhow, at least @gmaxwell didn't think it was a good idea there, should probably keep the discussion in one place...

MarcoFalke:
Needs rebase if still relevant

laanwj:
Closing, this has been inactive for too long.

PR #10595, opened 2017-06-15
Bugfix: RPC/Mining: Use pre-segwit sigops and limits, when working with non-segwit GBT clients

luke-jr:


sipa:
utACK 4292a752c7eb59ef2cbee70b5cd186eedf16d0dc

luke-jr:
Rebased. At this point, it might make sense to just drop support for pre-segwit miners, but this fix is simple, so it makes sense to get in (and backportable) first.

test/functional/feature_segwit.py
```diff
@@ -225,6 +225,13 @@ def run_test(self):
         # Now try calling getblocktemplate() without segwit support.
         template = self.nodes[0].getblocktemplate()
 
+        # Check that we got a pre-Segwit template
+        assert(template['sizelimit'] == 1000000)
```
jnewbery: nit: prefer `assert_equal()` for asserting equality

test/functional/feature_segwit.py
```diff
@@ -225,6 +225,13 @@ def run_test(self):
...
+        assert('weightlimit' not in template)
```
jnewbery: nit: assert is a statement, not a function. Please remove parens.

jnewbery (commented):
I agree that we should just drop support for non-segwit miners at this point. Doesn't seem worth the trouble of making this change.

MarcoFalke:
This has been "fixed" by 
*    #14811 (Mining: Enforce that segwit option must be set in GBT by jnewbery)

PR #10596, opened 2017-06-15
Add vConnect to CConnman::Options

benma:
Split the "-connect" argument parsing out of CConnman and put it into
AppInitMain().

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -1651,7 +1651,14 @@ bool AppInitMain(boost::thread_group& threadGroup, CScheduler& scheduler)
     if (gArgs.IsArgSet("-seednode")) {
         connOptions.vSeedNodes = gArgs.GetArgs("-seednode");
     }
-
+    // Initiate outbound connections unless connect=0
+    connOptions.m_use_addrman_outgoing = !gArgs.IsArgSet("-connect");
+    if (!connOptions.m_use_addrman_outgoing) {
+        const auto connect = gArgs.GetArgs("-connect");
+        if (connect.size() != 1 || connect[0] != "0") {
+            connOptions.m_specified_outgoing = connect;
+        }
+    }
     if (!connman.Start(scheduler, connOptions)) {
         return false;
     }
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -1674,15 +1674,15 @@ void CConnman::ProcessOneShot()
     }
 }
 
-void CConnman::ThreadOpenConnections()
+void CConnman::ThreadOpenConnections(const std::vector<std::string> connect)
 {
     // Connect to specific addresses
-    if (gArgs.IsArgSet("-connect"))
+    if (!connect.empty())
     {
         for (int64_t nLoop = 0;; nLoop++)
         {
             ProcessOneShot();
-            for (const std::string& strAddr : gArgs.GetArgs("-connect"))
+            for (const std::string& strAddr : connect)
             {
                 CAddress addr(CService(), NODE_NONE);
                 OpenNetworkConnection(addr, false, NULL, strAddr.c_str());
@@ -2360,9 +2360,16 @@ bool CConnman::Start(CScheduler& scheduler, Options connOptions)
     // Initiate outbound connections from -addnode
     threadOpenAddedConnections = std::thread(&TraceThread<std::function<void()> >, "addcon", std::function<void()>(std::bind(&CConnman::ThreadOpenAddedConnections, this)));
 
-    // Initiate outbound connections unless connect=0
-    if (!gArgs.IsArgSet("-connect") || gArgs.GetArgs("-connect").size() != 1 || gArgs.GetArgs("-connect")[0] != "0")
-        threadOpenConnections = std::thread(&TraceThread<std::function<void()> >, "opencon", std::function<void()>(std::bind(&CConnman::ThreadOpenConnections, this)));
+    if (connOptions.m_use_addrman_outgoing && !connOptions.m_specified_outgoing.empty()) {
+        if (clientInterface) {
+            clientInterface->ThreadSafeMessageBox(
+                _("Cannot provide specific connections and have addrman find outgoing connections at the same."),
+                "", CClientUIInterface::MSG_ERROR);
+        }
+        return false;
+    }
+    if (connOptions.m_use_addrman_outgoing || !connOptions.m_specified_outgoing.empty())
+        threadOpenConnections = std::thread(&TraceThread<std::function<void()> >, "opencon", std::function<void()>(std::bind(&CConnman::ThreadOpenConnections, this, connOptions.m_specified_outgoing)));
 
     // Process messages
     threadMessageHandler = std::thread(&TraceThread<std::function<void()> >, "msghand", std::function<void()>(std::bind(&CConnman::ThreadMessageHandler, this)));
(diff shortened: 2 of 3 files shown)


benma:
@theuni please take a look if you can.

src/net.h
```diff
@@ -292,7 +294,7 @@ class CConnman
     void ThreadOpenAddedConnections();
     void AddOneShot(const std::string& strDest);
     void ProcessOneShot();
-    void ThreadOpenConnections();
+    void ThreadOpenConnections(std::vector<std::string> connect);
```
sipa: Pass by const reference?

theuni:
@benma Thanks, will review asap.

src/net.h
theuni: std::thread takes arguments by value/rvalue only. Taking a reference would be dangerous as the arg could go out of scope before it's used in the thread.

benma: I also read that the reference is turned into a copy, but I figured it's better to be explicit about it being a copy. I also didn't use a pointer because dealing with its lifetime is not worth it.

src/init.cpp
```diff
@@ -1664,6 +1664,13 @@ bool AppInitMain(boost::thread_group& threadGroup, CScheduler& scheduler)
     if (gArgs.IsArgSet("-seednode")) {
         connOptions.vSeedNodes = gArgs.GetArgs("-seednode");
     }
+    if (gArgs.IsArgSet("-connect")) {
+        // Initiate outbound connections unless connect=0
+        const auto connect = gArgs.GetArgs("-connect");
+        connOptions.bNoConnect = connect.size() == 1 && connect[0] == "0";
+        if (!connOptions.bNoConnect)
```
theuni: Please rename this so that they're more clear, since this is exposed outside of CConnman now.
connOptions.vConnect -> connOptions.m_specified_outgoing
connOptions.bNoConnect -> (!) connOptions.m_use_addrman_outgoing

src/net.cpp
```diff
@@ -2323,9 +2323,8 @@ bool CConnman::Start(CScheduler& scheduler, std::string& strNodeError, Options c
     // Initiate outbound connections from -addnode
     threadOpenAddedConnections = std::thread(&TraceThread<std::function<void()> >, "addcon", std::function<void()>(std::bind(&CConnman::ThreadOpenAddedConnections, this)));
 
-    // Initiate outbound connections unless connect=0
-    if (!gArgs.IsArgSet("-connect") || gArgs.GetArgs("-connect").size() != 1 || gArgs.GetArgs("-connect")[0] != "0")
-        threadOpenConnections = std::thread(&TraceThread<std::function<void()> >, "opencon", std::function<void()>(std::bind(&CConnman::ThreadOpenConnections, this)));
+    if (!connOptions.bNoConnect)
```
theuni: All outbound connections are moving to one thread soon, but for now, we should return an error if connOptions.m_use_addrman_outgoing && !connOptions.m_specified_outgoing.empty().

Also, If addrman isn't being used for outgoing and there are no specified connections (incoming connections only), we can just skip creating the thread completely.

theuni (commented):
Concept ACK. Looks good, just a few nits.

src/net.h
laanwj: I also prefer it this way, explicitly.

benma:
@theuni just to let you know that I'll address everything in about two weeks (will be pretty much offline until then). Cheers.

src/init.cpp
```diff
@@ -1664,6 +1664,13 @@ bool AppInitMain(boost::thread_group& threadGroup, CScheduler& scheduler)
     if (gArgs.IsArgSet("-seednode")) {
         connOptions.vSeedNodes = gArgs.GetArgs("-seednode");
     }
+    if (gArgs.IsArgSet("-connect")) {
+        // Initiate outbound connections unless connect=0
+        const auto connect = gArgs.GetArgs("-connect");
+        connOptions.bNoConnect = connect.size() == 1 && connect[0] == "0";
```
benma: Done (I originally thought it was better to keep the local style and fix them all in batches)

src/init.cpp

src/net.cpp
benma: Done, but should the error maybe be an assert? That code path cannot be triggered and is not meaningful to the user. If other things want to use Connman and mess it up, it's a coding error and they'll see the assert.

theuni: It will be possible (and reasonable) when vAddedNodes is added as an connman option like the others. It might be a good next target :)

src/init.cpp
```diff
@@ -1651,7 +1651,14 @@ bool AppInitMain(boost::thread_group& threadGroup, CScheduler& scheduler)
     if (gArgs.IsArgSet("-seednode")) {
         connOptions.vSeedNodes = gArgs.GetArgs("-seednode");
     }
-
+    if (gArgs.IsArgSet("-connect")) {
+        // Initiate outbound connections unless connect=0
+        const auto connect = gArgs.GetArgs("-connect");
+        connOptions.m_use_addrman_outgoing = connect.empty();
```
theuni: connOptions.m_use_addrman_outgoing should be false if ```!connect.size() == 1 && connect[0] == "0"```

benma: Right! Fixed.

src/net.h
```diff
@@ -145,6 +145,8 @@ class CConnman
         std::vector<std::string> vSeedNodes;
         std::vector<CSubNet> vWhitelistedRange;
         std::vector<CService> vBinds, vWhiteBinds;
+        bool m_use_addrman_outgoing = true;
```
theuni: I'd rather not do this, as everything else here is set to null, false, or uninitialized. A value of true is unexpected.

benma: Done. Note that now, the default options are to not make any outbound connections (I put it to `true` to have it make outbound connections by default, like bitcoind does).

src/init.cpp
```diff
@@ -1651,7 +1651,14 @@ bool AppInitMain(boost::thread_group& threadGroup, CScheduler& scheduler)
     if (gArgs.IsArgSet("-seednode")) {
         connOptions.vSeedNodes = gArgs.GetArgs("-seednode");
     }
-
+    // Initiate outbound connections unless connect=0
+    connOptions.m_use_addrman_outgoing = !gArgs.IsArgSet("-connect");
+    if (!connOptions.m_use_addrman_outgoing) {
+        const auto connect = gArgs.GetArgs("-connect");
+        if (connect.size() != 1 || connect[0] != "0") {
+            connOptions.m_specified_outgoing = connect;
```
theuni: this is broken again if ```connect.size() == 1 && connect[0] == "0"``` :)

benma: How? If `connect.size() == 1 && connect[0] == "0"`, then `m_use_addrman_outgoing = false` and `m_specified_outgoing` is empty, and the thread is not started.

theuni: Yes sorry, nevermind.

benma:
@theuni does the PR look good to you?

theuni:
utACK 352d582ba240b825cb834cdde041864bafca0e21. This will need to go in after the 0.15 split though, due to the new strings.

src/net.cpp
benma: What does it have to do with vAddedNodes? The specifc new error that is introduced doesn't depend on it.

benma:
@theuni thanks. Did you mean [this string](https://github.com/bitcoin/bitcoin/pull/10596/files#diff-9a82240fe7dfe86564178691cc57f2f1R2366)? If so, maybe we can just remove it (resp. replace it with an assertion)?

See https://github.com/bitcoin/bitcoin/pull/10596#discussion_r130467670 (sorry for the late response, somehow I missed it).

PR #10597, opened 2017-06-15
scripted-diff: Make use of the improved handling of two consecutive right angle brackets in C++11

practicalswift:
Improve code readability by removing the legacy handling of two consecutive right angle brackets that was required prior to C++11. Non-functional change.

After this commit:

```
$ git grep "> >" -- "*.cpp" "*.h" | grep -v '/leveldb/' | wc -l
0
```

### DIFF
--- a/src/base58.h
+++ b/src/base58.h
@@ -75,7 +75,7 @@ protected:
     std::vector<unsigned char> vchVersion;
 
     //! the actually encoded data
-    typedef std::vector<unsigned char, zero_after_free_allocator<unsigned char> > vector_uchar;
+    typedef std::vector<unsigned char, zero_after_free_allocator<unsigned char>> vector_uchar;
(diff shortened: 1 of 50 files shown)


ryanofsky:
Need to update [src/.clang-format](https://github.com/bitcoin/bitcoin/blob/fa5137c11d2d4accd9bfc68b3498e7e16829c3ab/src/.clang-format#L50) or it will keep introducing these:

```diff
-Standard:        Cpp03
+Standard:        Cpp11
```

MarcoFalke:
According to the developer notes, Do not submit patches solely to modify
the style of existing code.
https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md#developer-notes

practicalswift:
OK, closing! :-)

PR #10604, opened 2017-06-15
[wallet] [tests] Add listwallets RPC, include wallet name in `getwalletinfo` and add multiwallet test

jnewbery:
- fix comment for CWallet::Verify (cleanup after #8694)
- expose the wallet name in `getwalletinfo` rpc
- add `listwallets` rpc - returns array of wallet names
- add functional test for multiwallet using new rpc functionality

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -2491,6 +2491,39 @@ UniValue getwalletinfo(const JSONRPCRequest& request)
     return obj;
 }
 
+UniValue listwallets(const JSONRPCRequest& request)
+{
+    if (request.fHelp || request.params.size() != 0)
+        throw std::runtime_error(
+            "listwallets\n"
+            "Returns a list of currently loaded wallets.\n"
+            "For full information on the wallet, use \"getwalletinfo\"\n"
+            "\nResult:\n"
+            "[                         (json array of strings)\n"
+            "  \"walletname\"            (string) the wallet name\n"
+            "   ...\n"
+            "]\n"
+            "\nExamples:\n"
+            + HelpExampleCli("listwallets", "")
+            + HelpExampleRpc("listwallets", "")
+        );
+
+    UniValue obj(UniValue::VARR);
+
+    for (CWalletRef pwallet : vpwallets) {
+
+        if (!EnsureWalletIsAvailable(pwallet, request.fHelp)) {
+            return NullUniValue;
+        }
+
+        LOCK(pwallet->cs_wallet);
+
+        obj.push_back(pwallet->GetName());
+    }
+
+    return obj;
+}
+
 UniValue resendwallettransactions(const JSONRPCRequest& request)
 {
     CWallet * const pwallet = GetWalletForJSONRPCRequest(request);
@@ -3087,6 +3120,7 @@ static const CRPCCommand commands[] =
     { "wallet",             "listsinceblock",           &listsinceblock,           false,  {"blockhash","target_confirmations","include_watchonly"} },
     { "wallet",             "listtransactions",         &listtransactions,         false,  {"account","count","skip","include_watchonly"} },
     { "wallet",             "listunspent",              &listunspent,              false,  {"minconf","maxconf","addresses","include_unsafe","query_options"} },
+    { "wallet",             "listwallets",              &listwallets,              true,   {} },
     { "wallet",             "lockunspent",              &lockunspent,              true,   {"unlock","transactions"} },
     { "wallet",             "move",                     &movecmd,                  false,  {"fromaccount","toaccount","amount","minconf","comment"} },
     { "wallet",             "sendfrom",                 &sendfrom,                 false,  {"fromaccount","toaddress","amount","minconf","comment","comment_to"} },
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -2447,6 +2447,7 @@ UniValue getwalletinfo(const JSONRPCRequest& request)
             "Returns an object containing various wallet state info.\n"
             "\nResult:\n"
             "{\n"
+            "  \"walletname\": xxxxx,             (string) the wallet name\n"
             "  \"walletversion\": xxxxx,          (numeric) the wallet version\n"
             "  \"balance\": xxxxxxx,              (numeric) the total confirmed balance of the wallet in " + CURRENCY_UNIT + "\n"
             "  \"unconfirmed_balance\": xxx,      (numeric) the total unconfirmed balance of the wallet in " + CURRENCY_UNIT + "\n"
@@ -2469,6 +2470,7 @@ UniValue getwalletinfo(const JSONRPCRequest& request)
     UniValue obj(UniValue::VOBJ);
 
     size_t kpExternalSize = pwallet->KeypoolCountExternalKeys();
+    obj.push_back(Pair("walletname", pwallet->GetName()));
     obj.push_back(Pair("walletversion", pwallet->GetVersion()));
     obj.push_back(Pair("balance",       ValueFromAmount(pwallet->GetBalance())));
     obj.push_back(Pair("unconfirmed_balance", ValueFromAmount(pwallet->GetUnconfirmedBalance())));
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -1066,7 +1066,9 @@ public:
     //! Flush wallet (bitdb flush)
     void Flush(bool shutdown=false);
 
-    //! Verify the wallet database and perform salvage if required
+    //! Responsible for reading and validating the -wallet arguments and verifying the wallet database.
+    //  This function will perform salvage on the wallet if requested, as long as only one wallet is
+    //  being loaded (CWallet::ParameterInteraction forbids -salvagewallet, -zapwallettxes or -upgradewallet with multiwallet).
     static bool Verify();
     
     /** 
--- a/test/functional/multiwallet.py
+++ b/test/functional/multiwallet.py
@@ -2,9 +2,12 @@
 # Copyright (c) 2017 The Bitcoin Core developers
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
-"""Test multiwallet."""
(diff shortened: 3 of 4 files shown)


ryanofsky:
It's good to add the wallet name to `getwalletinfo` but I don't see a reason to break backwards compatibility by returning an array in place of the previous wallet info object.

I thought the approach we were going to take for multiwallet RPC support was to add some kind of `listwallets` API that would return wallet names, and some kind of RPC session variable that would control which wallet is returned by `GetWalletForJSONRPCRequest` and used by all current wallet apis (including `getwalletinfo`).

ryanofsky:
Jonas had a suggestion about how we could use different RPC endpoints to access individual wallets here: https://github.com/bitcoin/bitcoin/pull/8694#issuecomment-246095336

jnewbery:
> Jonas had a suggestion about how we could use different RPC endpoints to access individual wallets

Sounds good, although I don't think that's relevant for this PR, since it doesn't touch any RPCs that are specific to a single wallet.

> I thought the approach we were going to take for multiwallet RPC support was to add some kind of `listwallets` API

I missed that discussion (and can't find mention of `listwallets` in the PRs you've linked to). Can you point me at where this was discussed? It shouldn't be difficult to change this PR to implement a new `listwallets` RPC and use that in the test if that's what people want.

ryanofsky:
> Sounds good, although I don't think that's relevant for this PR, since it doesn't touch any RPCs that are specific to a single wallet.

Deriving the wallet from the RPC endpoint is relevant to this PR if you want getwalletinfo to be able to return information about all wallets without changing it to return an array.

> I missed that discussion (and can't find mention of listwallets in the PRs you've linked to). Can you point me at where this was discussed? It shouldn't be difficult to change this PR to implement a new listwallets RPC and use that in the test if that's what people want.

There's no discussion about listwallets, I was just suggesting that as an alternative to making getwalletinfo return an array.

laanwj:
> It's good to add the wallet name to getwalletinfo but I don't see a reason to break backwards compatibility by returning an array in place of the previous wallet info object.
> I thought the approach we were going to take for multiwallet RPC support was to add some kind of  listwallets API that would return wallet names, and some kind of RPC session variable that would control which wallet is returned by GetWalletForJSONRPCRequest and used by all current wallet apis (including getwalletinfo).

I agree - last meeting we decided on using RPC endpoints to distinguish wallets. This means that the current wallet RPCs, including `getwalletinfo`, will take derive the wallet to be use from this. Adding the wallet name to the current API makes sense, of course.

Also adding a new `listwallets` RPC on a global (not per-wallet) level makes sense. This will keep backward compatibility but make it possible to list the currently loaded wallets.

jnewbery:
@ryanofsky @laanwj - thanks for the useful feedback on the API here.

I've updated this PR to add a new `listwallets` RPC, which simply lists the names of the currently loaded wallets. The command intentionally shows no information about the wallet. If/when we add multi-user multiwallet functionality, the `listwallet` RPC should not reveal the contents of the individual wallets.

The PR still adds `walletname` to the output of the `getwalletinfo` RPC.

Also rebased on master.

instagibbs:
will review

jnewbery:
rebased

instagibbs:
API makes sense to me 

tACK https://github.com/bitcoin/bitcoin/pull/10604/commits/74882ca0ebc070ba80304e910af4562659e44a1f

jonasschnelli:
Looks good.
I'm not sure if we should call it `wallet_name` on the RPC layer. It's actually the `wallet_filename`. Maybe we never do, but there is a chance that we once allow to give wallets a name (like other multiwallet software) and not sure if we want the filename char limits there.

jnewbery:
> I'm not sure if we should call it wallet_name on the RPC layer

Yes, good point to bring up. I see there was some discussion of this here: https://github.com/bitcoin/bitcoin/pull/10650#issuecomment-310606962 . My personal opinion is that wallet name should always equal filename to avoid confusion, but I'm happy to discuss if other people have different ideas. Perhaps one to bring up in tomorrow's meeting?

gmaxwell (approved):
utACK

src/wallet/wallet.h
```diff
@@ -1053,7 +1053,9 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
     //! Flush wallet (bitdb flush)
     void Flush(bool shutdown=false);
 
-    //! Verify the wallet database and perform salvage if required
+    //! Responsible for reading and validating the -wallet arguments and verifying the wallet database.
+    //  This function will perform salvage on the wallet if requested, as long as only one wallet is
```
morcos: ultranit: extra white  space

src/wallet/rpcwallet.cpp
```diff
@@ -2432,6 +2434,41 @@ UniValue getwalletinfo(const JSONRPCRequest& request)
...
+            "[\n"
+            "  \"walletname\": xxxxx,             (string) the wallet name\n"
+            "   ...\n"
+            "]\n"
+            "\nExamples:\n"
+            + HelpExampleCli("listwallets", "")
+            + HelpExampleRpc("listwallets", "")
+        );
+
+    LOCK(cs_main);
```
ryanofsky: Probably should drop the cs_main lock. I don't think we decided what lock will be used to guard the vpwallets array, and no lock is used other places (it's not needed since the list doesn't change after initialization) so it would be inconsistent to add this here.

src/wallet/rpcwallet.cpp
```diff
@@ -2432,6 +2434,41 @@ UniValue getwalletinfo(const JSONRPCRequest& request)
...
+UniValue listwallets(const JSONRPCRequest& request)
+{
+    if (request.fHelp || request.params.size() != 0)
+        throw std::runtime_error(
+            "listwallets\n"
+            "Returns a list of currently loaded wallets.\n"
+            "For full information on the wallet, use \"getwalletinfo\"\n"
+            "\nResult:\n"
+            "[\n"
+            "  \"walletname\": xxxxx,             (string) the wallet name\n"
```
ryanofsky: Should drop the `:xxxx,` here since an array is just a list of items, not key/value pairs.

src/wallet/rpcwallet.cpp
```diff
@@ -2434,6 +2434,41 @@ UniValue getwalletinfo(const JSONRPCRequest& request)
...
```
morcos: it just prints the wallet name, not a pair

ryanofsky (commented):
utACK 74882ca0ebc070ba80304e910af4562659e44a1f

ryanofsky:
Also, should maybe clean up the PR description (remove all caps "EDITED") now that #10786 is merged.

src/wallet/wallet.h
jnewbery: Thanks. fixed

src/wallet/rpcwallet.cpp
jnewbery: quite right. Fixed

src/wallet/rpcwallet.cpp
jnewbery: yup. Fixed

src/wallet/rpcwallet.cpp
jnewbery: Done. Thanks!

jnewbery:
@ryanofsky's and @morcos's comments addressed

morcos:
ACK 011a860

will be awesome to show people all the wallets they can't access!

instagibbs:
re-ACK https://github.com/bitcoin/bitcoin/pull/10604/commits/011a860db94b179ca158dfc38acf7beefca2bed1

jnewbery:
This conflicts (slightly) with #10650. I have a branch of this rebased on top of that, so if both are going in to v0.15, #10650 can go in first and I'll push the rebased branch.

jnewbery:
Rebased on the latest #10650

instagibbs:
re-re-ACK https://github.com/bitcoin/bitcoin/pull/10604/commits/5baca3ad700e748d08b0b8f08ef014ea2c478d13

jnewbery:
rebased on master now that we have the endpoint PR merged.

@laanwj: Any chance of tagging this 0.15? It has 4 ACKs pre-rebase. I think it's useful multiwallet management functionality.

sipa:
utACK 3707fcd94e6251384235d16faafc975853d49e3d

instagibbs:
test failing:
    assert_raises_jsonrpc(-13, "Please enter the wallet passphrase with walletpassphrase first", self.nodes[0].dumpprivkey, address)
  File "/home/travis/build/bitcoin/bitcoin/build/bitcoin-i686-pc-linux-gnu/test/functional/test_framework/util.py", line 87, in assert_raises_jsonrpc

from walletencryption.py

jnewbery:
bah! Unrelated intermittent travis failure. Job restarted.

src/wallet/rpcwallet.cpp
```diff
@@ -2489,6 +2491,39 @@ UniValue getwalletinfo(const JSONRPCRequest& request)
...
+            "]\n"
+            "\nExamples:\n"
+            + HelpExampleCli("listwallets", "")
+            + HelpExampleRpc("listwallets", "")
+        );
+
+    UniValue obj(UniValue::VARR);
+
+    for (CWalletRef pwallet : vpwallets) {
+
```
jonasschnelli: nit newline

jonasschnelli:
utACK 3707fcd94e6251384235d16faafc975853d49e3d

src/wallet/rpcwallet.cpp
jnewbery: Yes, agree, but in the interests of getting this merged, I won't change this now that https://github.com/bitcoin/bitcoin/commit/3707fcd94e6251384235d16faafc975853d49e3d has two ACKs

instagibbs:
ACK https://github.com/bitcoin/bitcoin/pull/10604/commits/3707fcd94e6251384235d16faafc975853d49e3d

MarcoFalke:
Nit: No need to run multiwallet.py twice in the test runner ;)

jnewbery:
> Nit: No need to run multiwallet.py twice in the test runner ;)

Oops. Good catch!

PR #10608, opened 2017-06-16
Add a comment explaining the use of MAX_BLOCK_BASE_SIZE.

gmaxwell:


instagibbs:
so `network rule` here actually means "network consensus rule"? I misunderstood when I read this before.

src/consensus/consensus.h
```diff
@@ -12,7 +12,13 @@
 static const unsigned int MAX_BLOCK_SERIALIZED_SIZE = 4000000;
 /** The maximum allowed weight for a block, see BIP 141 (network rule) */
 static const unsigned int MAX_BLOCK_WEIGHT = 4000000;
-/** The maximum allowed size for a block excluding witness data, in bytes (network rule) */
+/**
+ * The maximum allowed size for a block excluding witness data, in bytes (network rule).
+ * This parameter is largely superfluous because it is directly implied by the above block
+ *  weight limit, even when BIP 141 is not active.  It continues to exist for use in
+ *  various early tests that run before the witness data has been checked.
```
practicalswift: Minor nit: Accidental double spaces after `*` on the two lines above.

practicalswift:
ACK 18873373bde38a026f154c5878fedac6c2a04446 modulo clarification of "network rule" vs "network consensus rule" as suggested by @instagibbs

src/consensus/consensus.h
gmaxwell: intentional indentation

gmaxwell:
@instagibbs if you decrease it below a million you will break consensus.  If you remove it entirely you will not, or set it to eleven million you will not.   It's a consensus rule but a redundant one which is subsumed by weight.

Do you see a way to make that more clear?

src/consensus/consensus.h
practicalswift: Just out of curiosity, why the indentation? :-)

benma: "tests" is overloaded and could be confused for unit/integration tests. "early checks" might be better.

Why are those checks not removed if they don't change the rules? Would be good to mention the reason for it also here.

practicalswift: After re-reading the diff I think I get it: the indentation is to clarify that the current line is part of the sentence of the line above.

gmaxwell: There are many early checks in validation which limit the resources used handling invalid things or avoid incorrect invalidation in the presence of malleability. Most of them explain themselves.

E.g. in this case you cannot check the weight limit until after you've verified the witness commitment, which takes a bunch of hashing-- otherwise you might invalidate a block you received with an tampered witness (take a valid block add another 2mb of junk to one of the witnesses) instead of just dropping it and the party that gave it to you.  But you can check the base size before doing that.

gmaxwell:
#10618 is an alternative that just removes the constant.

laanwj:
This can't hurt. utACK 1887337
#10618 is more risky, could be done later.

Issue #10601, opened 2017-06-15
QA: Document our preferred test writing style


sdaftuar:
When writing a new test, I often ask myself what the best existing test we have is, so that I can mimic the style and usage of the various test_framework components, especially as we re-work the test_framework and create new preferred idioms.

So I think it'd be great to update the docs in test/functional/README.md to have a sample short test, with lots of comments, that people can look to when writing a test from scratch.

jnewbery:
Addressed in #10612. Any feedback welcomed!

PR #10612, opened 2017-06-16
The young person's guide to the test_framework

jnewbery:
*Variations on a theme of Nakamoto*

- Adds a heavily commented example_test.py test script that can be used as a template for writing new functional tests.
- Expands test/README.md and test/functional/README.md:
    - test/README.md documents how to run tests locally, as well as giving information about logging and troubleshooting failing test cases
    - test/functional/README.md documents how to write tests, including a style guide and general tips on writing good test cases.

Addresses #10601

The style guide is of course completely up for grabs at the moment. This PR contains my first attempt, but I'm very happy for feedback including expanding/changing/reducing the guide.

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -15,84 +15,152 @@ The util tests are run as part of `make check` target. The functional
 tests are run by the travis continuous build process whenever a pull
 request is opened. Both sets of tests can also be run locally.
 
-Functional Test dependencies
-============================
+# Running tests locally
+
+Build for your system first. Be sure to enable wallet, utils and daemon when you configure. Tests will not run otherwise.
+
+### Functional tests
+
+#### Dependencies
+
 The ZMQ functional test requires a python ZMQ library. To install it:
 
 - on Unix, run `sudo apt-get install python3-zmq`
 - on mac OS, run `pip3 install pyzmq`
 
-Running tests locally
-=====================
+#### Running the tests
 
-Build for your system first. Be sure to enable wallet, utils and daemon when you configure. Tests will not run otherwise.
+Individual tests can be run by directly calling the test script, eg:
 
-Functional tests
-----------------
+```
+test/functional/replace-by-fee.py
+```
 
-You can run any single test by calling
+or can be run through the test_runner harness, eg:
 
-    test/functional/test_runner.py <testname>
+```
+test/functional/test_runner.py replace-by-fee.py
+```
 
-Or you can run any combination (incl. duplicates) of tests by calling
+You can run any combination (incl. duplicates) of tests by calling:
 
-    test/functional/test_runner.py <testname1> <testname2> <testname3> ...
+```
+test/functional/test_runner.py <testname1> <testname2> <testname3> ...
+```
 
-Run the regression test suite with
+Run the regression test suite with:
 
-    test/functional/test_runner.py
+```
+test/functional/test_runner.py
+```
 
 Run all possible tests with
 
-    test/functional/test_runner.py --extended
-
-By default, tests will be run in parallel. To specify how many jobs to run,
-append `--jobs=n` (default n=4).
-
-If you want to create a basic coverage report for the RPC test suite, append `--coverage`.
-
-Possible options, which apply to each individual test run:
-
 ```
-  -h, --help            show this help message and exit
-  --nocleanup           Leave bitcoinds and test.* datadir on exit or error
-  --noshutdown          Don't stop bitcoinds after the test execution
-  --srcdir=SRCDIR       Source directory containing bitcoind/bitcoin-cli
-                        (default: ../../src)
-  --tmpdir=TMPDIR       Root directory for datadirs
-  --tracerpc            Print out all RPC calls as they are made
-  --coveragedir=COVERAGEDIR
-                        Write tested RPC commands into this directory
+test/functional/test_runner.py --extended
 ```
 
-If you set the environment variable `PYTHON_DEBUG=1` you will get some debug
-output (example: `PYTHON_DEBUG=1 test/functional/test_runner.py wallet`).
+By default, up to 4 tests will be run in parallel by test_runner. To specify
+how many jobs to run, append `--jobs=n`
 
-A 200-block -regtest blockchain and wallets for four nodes
-is created the first time a regression test is run and
-is stored in the cache/ directory. Each node has 25 mature
-blocks (25*50=1250 BTC) in its wallet.
+The individual tests and the test_runner harness have many command-line
+options. Run `test_runner.py -h` to see them all.
 
-After the first run, the cache/ blockchain and wallets are
-copied into a temporary directory and used as the initial
-test state.
+#### Troubleshooting and debugging test failures
 
-If you get into a bad state, you should be able
-to recover with:
+##### Resource contention
+
+The P2P and RPC ports used by the bitcoind nodes-under-test are chosen to make
+conflicts with other processes unlikely. However, if there is another bitcoind
+process running on the system (perhaps from a previous test which hasn't successfully
+killed all its bitcoind nodes), then there may be a port conflict which will
+cause the test to fail. It is recommended that you run the tests on a system
+where no other bitcoind processes are running.
+
+On linux, the test_framework will warn if there is another
+bitcoind process running when the tests are started.
+
+If there are zombie bitcoind processes after test failure, you can kill them
+by running the following commands. **Note that these commands will kill all
+bitcoind processes running on the system, so should not be used if any non-test
+bitcoind processes are being run.**
+
+```bash
+killall bitcoind
+```
+
+or
+
+```bash
+pkill -9 bitcoind
+```
+
+
+##### Data directory cache
+
+A pre-mined blockchain with 200 blocks is generated the first time a
+functional test is run and is stored in test/cache. This speeds up
+test startup times since new blockchains don't need to be generated for
+each test. However, the cache may get into a bad state, in which case
+tests will fail. If this happens, remove the cache directory (and make
+sure bitcoind processes are stopped as above):
 
 ```bash
 rm -rf cache
 killall bitcoind
 ```
 
-Util tests
-----------
+##### Test logging
+
+The tests contain logging at different levels (debug, info, warning, etc). By
+default:
+
+- when run through the test_runner harness, *all* logs are written to
+  `test_framework.log` and no logs are output to the console.
+- when run directly, *all* logs are written to `test_framework.log` and INFO
+  level and above are output to the console.
+- when run on Travis, no logs are output to the console. However, if a test
+  fails, the `test_framework.log` and bitcoind `debug.log`s will all be dumped
+  to the console to help troubleshooting.
+
+To change the level of logs output to the console, use the `-l` command line
+argument.
+
+`test_framework.log` and bitcoind `debug.log`s can be combined into a single
+aggregate log by running the `combine_logs.py` script. The output can be plain
+text, colorized text or html. For example:
+
+```
+combine_logs.py -c <test data directory> | less -r
+```
+
+will pipe the colorized logs from the test into less.
+
+Use `--tracerpc` to trace out all the RPC calls and responses to the console. For
+some tests (eg any that use `submitblock` to submit a full block over RPC),
+this can result in a lot of screen output.
+
+By default, the test data directory will be deleted after a successful run.
+Use `--nocleanup` to leave the test data directory intact. The test data
+directory is never deleted after a failed test.
+
+##### Attaching a debugger
+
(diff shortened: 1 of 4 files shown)


jimmysong (approved):
ACK

I wish this documentation existed when I started writing tests! Only thing I would add is maybe some documentation on finding rpc commands that haven't really been covered yet (read: --coverage). Also, would be useful to point people to the actual rpc commands in C++ that can be used to find things to test.

test/README.md
```diff
@@ -15,84 +15,148 @@ The util tests are run as part of `make check` target. The functional
...
 
 - on Unix, run `sudo apt-get install python3-zmq`
 - on mac OS, run `pip3 install pyzmq`
 
-Running tests locally
-=====================
+#### Running the tests
 
-Build for your system first. Be sure to enable wallet, utils and daemon when you configure. Tests will not run otherwise.
+individual tests can be run by directly calling the test script, eg:
```
practicalswift: Nit: `s/^individual/Individual/` :-)

jnewbery: Thanks! Fixed

jnewbery:
Thanks @jimmysong - added sections on **Coverage** and **RPC and p2p definitions** to `test/functional/README.md`

laanwj:
Awesome! Thanks for adding a guide to make tests.

test/functional/example_test.py
```diff
@@ -0,0 +1,205 @@
...
+        self.log.info("Wait for node2 reach current tip. Test that it has propogated all the blocks to us")
+
+        for block in blocks:
+            getdata_request = msg_getdata()
+            getdata_request.inv.append(CInv(2, block))
+            node2.send_message(getdata_request)
+
+        # wait_until() will loop until a predicate condition is met. Use it to test properties of the
+        # ConnNodeCB objects.
+        assert wait_until(lambda: sorted(blocks) == sorted(list(node2.block_receive_map.keys())), timeout=5)
```
MarcoFalke: I'd prefer if the assert was moved into wait_until and the return statement removed from wait_until.

The places in the code where it is missing right now are likely oversights and should be fixed.

test/functional/example_test.py
```diff
@@ -0,0 +1,205 @@
...
+
+        # Notice above how we called an RPC by calling a method with the same
+        # name on the node object. Notice also how we used a keyword argument
+        # to specify a named RPC argument. Neither of those are defined on the
+        # node object. Instead there's some __getattr__() magic going on under
+        # the covers to dispatch unrecognised attribute calls to the RPC
+        # interface.
+
+        # Logs are nice. Do plenty of them. They can be used in place of comments for
+        # breaking the test into sub-sections.
```
MarcoFalke: 👍

test/README.md
```diff
@@ -15,84 +15,148 @@ The util tests are run as part of `make check` target. The functional
...
+##### Test logging
+
+The tests contain logging at different levels (debug, info, warning, etc). By
+default:
+
+- when run through the test_runner harness, *all* logs are written to
+  `test_framework.log` and no logs are output to the console.
+- when run directly, *all* logs are written to `test_framework.log` and INFO
+  level and above are output to the console.
+- when run on Travis, no logs are output to the console. However, if a test
```
MarcoFalke: Since travis supports color, could we run `combine_logs.py` on travis as well?

maflcko (commented):
Concept ACK! Just some meta comments.

Christewart:
Awesome! Concept ACK x2

test/functional/example_test.py
jnewbery: Indeed. I'd also like to see `wait_until()` assert itself. I've previously tried adding an assert into that function and it causes several tests to fail.

Fixing those cases where `wait_until()` isn't actually returning True is a job for another PR. When that happens, this example test script should also be updated.

test/README.md
jnewbery: Potentially, yes. `combine_logs.py` was written to be portable (and even usable on machines where test_framework is not installed).

Again, I think this is a job for another PR.

jnewbery:
@laanwj @Marcofalke @christewart - thanks for the feedback.

Any comments on the suggested style guide? I was expecting that to be the area where people would have opinions.

test/README.md
```diff
@@ -15,84 +15,148 @@ The util tests are run as part of `make check` target. The functional
...
+aggregate log by running the `combine_logs.py` script. The output can be plain
+text, colorized text or html. For example:
+
+```
+combine_logs.py -c <test data directory> | less -r
+```
+
+will pipe the colorized logs from the test into less.
+
+Use --tracerpc to trace out all the RPC calls and responses to the console. For
```
achow101: --tracerpc should be surrounded by backticks.

achow101 (approved):
ACK 5a0134a5deb09103e4ba70ebf519c26c1557985b

test/functional/README.md
```diff
@@ -1,108 +1,157 @@
...
+#### RPC and p2p definitions
+
+Test writers may find it helpful to refer to the definitions for the RPC and
+p2p messages. These can be found in the following source files:
 
-P2P test design notes
----------------------
+- `/src/rpc/*` for RPCs
+- `/src/wallet/rpc*` for wallet RPCs
+- `ProcessMessage()` in `/sfrc/net_processing.cpp` for parsing p2p messages
```
MarcoFalke: nit: sfrc/src

test/functional/test_runner.py
```diff
@@ -153,6 +153,7 @@
     # These are python files that live in the functional tests directory, but are not test scripts.
     "combine_logs.py",
     "create_cache.py",
+    "example_test.py",
```
MarcoFalke: Can we put this in the extended scripts, just to make sure it has correct syntax?

maflcko (commented):
utACK 5a0134a, one nit.

test/functional/example_test.py
```diff
@@ -0,0 +1,205 @@
...
+
+        self.log.info("Wait for node2 reach current tip. Test that it has propogated all the blocks to us")
+
+        for block in blocks:
+            getdata_request = msg_getdata()
+            getdata_request.inv.append(CInv(2, block))
+            node2.send_message(getdata_request)
+
+        # wait_until() will loop until a predicate condition is met. Use it to test properties of the
+        # ConnNodeCB objects.
```
sdaftuar: `NodeConnCB`, not `ConnNodeCB`

test/functional/example_test.py
sdaftuar: I think it'd be nice to mention the locking issues with reaching into a `NodeConnCB` from the main thread.  We could mention that `wait_until()` grabs the global lock before checking the predicate, or perhaps better still would be to add an example of acquiring the lock in this example before inspecting an object that is accessed from the other thread.

sdaftuar (approved):
Thanks for doing this!  Left some nits for you, but looks good to me.

test/README.md
```diff
@@ -15,84 +15,148 @@ The util tests are run as part of `make check` target. The functional
...
+cause the test to fail. It is recommended that you run the tests on a system
+where no other bitcoind processes are running.
+
+On linux, the test_framework will warn if there is another
+bitcoind process running when the tests are started.
+
+If there are zombie bitcoind processes after test failure, you can kill them by running:
+
+```bash
+killall bitcoind
```
laanwj: This could potentially be risky if any non-test bitcoinds are running as the same user. Not sure it warrants a warning or not.

test/README.md
```diff
@@ -15,84 +15,148 @@ The util tests are run as part of `make check` target. The functional
...
+By default, the test data directory will be deleted after a successful run.
+Use `--nocleanup` to leave the test data directory intact. The test data
+directory is never deleted after a failed test.
+
+##### Attaching a debugger
+
+A python debugger can be attached to tests at any point. Just add the line:
+
+```py
+import pdb; pdb.set_trace()
```
laanwj: Nice one to mention. After ~15 years of using Python I still forgot about this all the time when I need it.

MarcoFalke:
​Another nit I forgot to mention earlier: ​
​
​Since you remove the mention of `PYTHON_DEBUG`, you might as well remove
it in the code as well and replace it by `os.getenv('TRAVIS') == 'true'`​

test/functional/README.md
```diff
@@ -1,108 +1,157 @@
...
+  or not to use the cached data directories. The cached data directories
+  contain a 200-block pre-mined blockchain and wallets for four nodes. Each node
+  has 25 mature blocks (25x50=1250 BTC) in its wallet.
+- When calling RPCs with lots of arguments, consider using named keyword
+  arguments instead of positional arguments to make the intent of the call
+  clear to readers.
 
-### [test_framework/blocktools.py](test_framework/blocktools.py)
-Helper functions for creating blocks and transactions.
+#### RPC and p2p definitions
```
laanwj: also here: P2P uppercase please

test/README.md
jnewbery: thanks. Done

test/functional/README.md
jnewbery: fixed

test/functional/test_runner.py
jnewbery: Yes, good idea. `example_test` should be run in the extended test run to test the test.

test/functional/README.md
```diff
@@ -1,108 +1,157 @@
...
-* ```NodeConn``` is the class used to connect to a bitcoind.  If you implement
-a callback class that derives from ```NodeConnCB``` and pass that to the
-```NodeConn``` object, your code will receive the appropriate callbacks when
+- `NodeConn` is the class used to connect to a bitcoind.  If you implement
+a callback class that derives from `NodeConnCB` and pass that to the
+`NodeConn` object, your code will receive the appropriate callbacks when
 events of interest arrive.
 
-* You can pass the same handler to multiple ```NodeConn```'s if you like, or pass
+- You can pass the same handler to multiple `NodeConn`'s if you like, or pass
```
jnewbery: Yes, you're right. I've removed this line.

test/functional/example_test.py
```diff
@@ -0,0 +1,205 @@
...
+
+        If you do override this method, remember to start the nodes, assign
+        them to self.nodes, connect them and then sync."""
+
+        self.setup_nodes()
+
+        # In this test, we're not connecting node2 to node0 or node1. Calls to
+        # sync_all() should not include node2, since we're not expecting it to
+        # sync.
+        connect_nodes_bi(self.nodes, 0, 1)
```
jnewbery: Yes - removed from the example test.

test/functional/example_test.py
jnewbery: thanks! Fixed.

test/functional/example_test.py
jnewbery: Good idea. Added an example of accessing the NodeConnCB data using the mininode_lock

test/README.md
```diff
@@ -15,84 +15,148 @@ The util tests are run as part of `make check` target. The functional
...
-  -h, --help            show this help message and exit
-  --nocleanup           Leave bitcoinds and test.* datadir on exit or error
-  --noshutdown          Don't stop bitcoinds after the test execution
-  --srcdir=SRCDIR       Source directory containing bitcoind/bitcoin-cli
-                        (default: ../../src)
-  --tmpdir=TMPDIR       Root directory for datadirs
-  --tracerpc            Print out all RPC calls as they are made
-  --coveragedir=COVERAGEDIR
-                        Write tested RPC commands into this directory
+The p2p and RPC ports used by the bitcoind nodes-under-test are chosen to make
```
jnewbery: fixed

test/README.md
jnewbery: Added a warning

test/README.md
jnewbery: 🙂 hooray for `pdb`!

test/functional/README.md
jnewbery: fixed

jnewbery:
> Since you remove the mention of `PYTHON_DEBUG`, you might as well remove
it in the code as well and replace it by `os.getenv('TRAVIS') == 'true'`​

Absolutely agree, although this should be done in a separate PR.

jnewbery:
Nits fixed

laanwj:
ACK e7ba6c1

test/README.md
instagibbs: how is this such a secret!

Another pro-tip that might be worthwhile is if you need to introspect bitcoind itself during a test you set a pdb break, then `gdb path/to/bitcoind <pid>` to attach to it.

Sorry I never saw this PR while it was live :/

jnewbery: Good tip @instagibbs . I'll happily ACK a PR that adds that to `test/README.md`

PR #10621, opened 2017-06-17
[wallet] Use fixed pubkey during wallet init for default key

instagibbs:
vchDefaultKey is not used anymore except to mark a
wallet as initialized during LoadWallet. Instead of
taking a key from the keypool and adding to the wallet
simply use a fixed key that is valid, for backwards
compability.

This allows wallets to return `m/..../0` keys first with `getnewaddress`, being slightly less surprising to the user, as well as giving the meaning behind having a key loaded to the wallet.

Key was chosen via smashing keyboard for entropy.

### DIFF
--- a/test/functional/wallet-hd.py
+++ b/test/functional/wallet-hd.py
@@ -56,7 +56,7 @@ class WalletHDTest(BitcoinTestFramework):
         for i in range(num_hd_adds):
             hd_add = self.nodes[1].getnewaddress()
             hd_info = self.nodes[1].validateaddress(hd_add)
-            assert_equal(hd_info["hdkeypath"], "m/0'/0'/"+str(i+1)+"'")
+            assert_equal(hd_info["hdkeypath"], "m/0'/0'/"+str(i)+"'")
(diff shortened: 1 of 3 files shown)


test/functional/wallet-hd.py
```diff
@@ -83,7 +83,7 @@ def run_test (self):
         for _ in range(num_hd_adds):
             hd_add_2 = self.nodes[1].getnewaddress()
             hd_info_2 = self.nodes[1].validateaddress(hd_add_2)
-            assert_equal(hd_info_2["hdkeypath"], "m/0'/0'/"+str(_+1)+"'")
+            assert_equal(hd_info_2["hdkeypath"], "m/0'/0'/"+str(_)+"'")
```
MarcoFalke: nit: might want to rename the variable to something other than underscore, which is only used to denote something that is never used.

jonasschnelli:
I think https://github.com/bitcoin/bitcoin/pull/10952 is preferable.

instagibbs:
Closing in favor of #10952

PR #9380, opened 2016-12-19
Separate different uses of minimum fees

morcos:
I've been meaning to do this for a while.
There is no behavior change here, but there were unintended consequences of having these all of these different notions linked to minRelayTxFee.

- blockmintxfee should be set by miners to reflect their marginal cost of transmitting extra bytes.
- incrementalrelayfee was previously hard-coded to the default minRelayTxFee, but it should be configurable so it can be adjusted to reflect the cost of network wide relay of transactions.
- dustrelayfee should not be accidentally changed when someone wants to limit their mempool to higher fee transactions.

I plan a future change to make the wallet smarter about not generating change outputs near the dust limit at all.  I also think we should consider splitting dust into 2 dust limits if we ever raise it again and first raise the creation limit and second the standard limit a version later.  Since it is not clear we'll ever need to increase the dust limit, this change isn't introduced now.

PR #10626, opened 2017-06-18
doc: Remove outdated minrelaytxfee comment

maflcko:
This is no longer true after #9380


laanwj:
Ping @morcos

morcos:
utACK

But we should probably replace it with a more accurate comment.

MarcoFalke:
I doubt any user will download and read the source code to look for documentation of command line args. The documentation should be put in the manpages.

laanwj:
Adding a more accurate comment could be done later - let's remove the inaccurate one at least.

PR #10632, opened 2017-06-19
qa: Add stopatheight test

maflcko:
Motivated by  #10569

### DIFF
--- a/test/functional/blockchain.py
+++ b/test/functional/blockchain.py
@@ -18,13 +18,16 @@ Tests correspond to code in rpc/blockchain.cpp.
 """
 
 from decimal import Decimal
+import subprocess
 
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import (
     assert_equal,
+    assert_raises,
     assert_raises_jsonrpc,
     assert_is_hex_string,
     assert_is_hash_string,
+    bitcoind_processes,
 )
 
 
@@ -34,6 +37,7 @@ class BlockchainTest(BitcoinTestFramework):
         super().__init__()
         self.setup_clean_chain = False
         self.num_nodes = 1
+        self.extra_args = [['-stopatheight=207']]
 
     def run_test(self):
         self._test_getchaintxstats()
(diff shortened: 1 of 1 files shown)


jonasschnelli (approved):
ACK fa75f235f14c05c7d7d433396f4f49809860340d

test/functional/blockchain.py
```diff
@@ -34,14 +37,26 @@ def __init__(self):
...
         self._test_getchaintxstats()
         self._test_gettxoutsetinfo()
         self._test_getblockheader()
         self._test_getdifficulty()
         self._test_getnetworkhashps()
-        self.nodes[0].verifychain(4, 0)
+        self._test_stopatheight()
+        assert self.nodes[0].verifychain(4, 0)
+
+    def _test_stopatheight(self):
```
jnewbery: micronit: The other subtests are defined in the order that they're called. Perhaps place this definition after the other subtests?

jnewbery (commented):
Tested ACK fa75f235f14c05c7d7d433396f4f49809860340d. Looks great!

Not sure that this entirely fits in blockchain.py, which is currently just for testing rpcs in rpc/blockchain.py. However, I can't think of any better place for it right now and it's definitely an improvement to have this test!

achow101:
ACK 5555fa8b74bc3c8e68232acbf4c54b9a8fadb8f2

test/functional/blockchain.py
```diff
@@ -129,5 +134,18 @@ def _test_getnetworkhashps(self):
         # This should be 2 hashes every 10 minutes or 1/300
         assert abs(hashes_per_second * 300 - 1) < 0.0001
 
+    def _test_stopatheight(self):
+        assert_equal(self.nodes[0].getblockcount(), 200)
+        self.nodes[0].generate(6)
+        assert_equal(self.nodes[0].getblockcount(), 206)
+        self.log.debug('Node should not stop at this height')
+        assert_raises(subprocess.TimeoutExpired, lambda: bitcoind_processes[0].wait(timeout=3))
+        self.nodes[0].generate(1)
```
sdaftuar: I believe this call to generate might fail -- isn't it possible the node's shutdown will interrupt the RPC response?  That looks to me like what happened here:
https://travis-ci.org/bitcoin/bitcoin/jobs/245917226#L2275

I think we should wrap the generate(1) in a try...except and allow for the RPC to fail due to connection refused, etc.

PR #10633, opened 2017-06-19
doc: Fix various typos

dimitris-t:
"on a already existing" -> "on an already existing". Change made on two occurrences.

PR #10635, opened 2017-06-19
Fixed typo in rpcconsole.cpp

dimitris-t:
"without fully understanding the ramification of a command" -> "without fully understanding the ramifications of a command"


dimitris-t:
Combining into #10633 and closing.

PR #10636, opened 2017-06-20
[qa] util: Check return code after closing bitcoind proc

maflcko:
Just a minor cleanup.

This has no effect on the outcome of test cases, but
prints shorter and less confusing tracebacks on fails.

I.e. does not print an obvious "ConnectionRefusedError"
when shutting down the nodes due to an invalid return code.


laanwj:
Nice, that has been a frequent annoyance, surprised it was so imple to solve.
utACK 999923e

sdaftuar:
post-merge utACK

PR #10638, opened 2017-06-20
rpc: Clarify pblock assumption via an assertion

practicalswift:
`pblock->nNonce = 0` will result in a `nullptr` dereference if the added assertion does not hold true.

`pblock != nullptr` is not obvious from reading the code - hence the need for stating this assumption explicitly via an assertion. This has the added benefit getting rid of the following `clang-tidy` warning:

```
rpc/mining.cpp:560:20: note: Access to field 'nNonce' results in
  a dereference of a null pointer (loaded from variable 'pblock')
    pblock->nNonce = 0;
```


practicalswift:
Anyone willing to review? :-)

practicalswift:
Seems to be consensus NACK on this one. Closing :-)

PR #10642, opened 2017-06-21
Remove obsolete _MSC_VER check

practicalswift:
Prior to this commit the `_MSC_VER` macro was being used regardless of being defined due to a misplaced `#endif`.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -844,8 +844,6 @@ bool AppInitBasicSetup()
     // Turn off Microsoft heap dump noise
     _CrtSetReportMode(_CRT_WARN, _CRTDBG_MODE_FILE);
     _CrtSetReportFile(_CRT_WARN, CreateFileA("NUL", GENERIC_WRITE, 0, NULL, OPEN_EXISTING, 0, 0));
-#endif
-#if _MSC_VER >= 1400
     // Disable confusing "helpful" text message on abort, Ctrl-C
     _set_abort_behavior(0, _WRITE_ABORT_MSG | _CALL_REPORTFAULT);
 #endif


laanwj:
Thanks. However I investigated and we can remove this fallback completely:

-  `MSVC++ 8.0  _MSC_VER == 1400 (Visual Studio 2005)` (source  https://stackoverflow.com/questions/70013/how-to-detect-if-im-compiling-code-with-visual-studio-2008)
- C++11 is supported since MSVC 2010: https://msdn.microsoft.com/en-us/library/hh567368.aspx
- Compiling without C++11 support is no longer possible.

practicalswift:
@laanwj Thanks for the clarification! If I understand it correctly the new diff should be:

```diff
diff --git a/src/init.cpp b/src/init.cpp
index ed76953..e213958 100644
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -844,8 +844,6 @@ bool AppInitBasicSetup()
     // Turn off Microsoft heap dump noise
     _CrtSetReportMode(_CRT_WARN, _CRTDBG_MODE_FILE);
     _CrtSetReportFile(_CRT_WARN, CreateFileA("NUL", GENERIC_WRITE, 0, NULL, OPEN_EXISTING, 0, 0));
-#endif
-#if _MSC_VER >= 1400
     // Disable confusing "helpful" text message on abort, Ctrl-C
     _set_abort_behavior(0, _WRITE_ABORT_MSG | _CALL_REPORTFAULT);
 #endif
```

In other words keeping the `#ifdef _MSC_VER` but removing the `#if _MSC_VER >= 1400`. Looks correct? :-)

laanwj:
Yes, indeed. That retains the behavior to not do any of that stuff for non-MSVC compilers (e.g. mingw). 
(which I think is the intent? that code is ancient)

practicalswift:
@laanwj Updated accordingly :-)

practicalswift:
Travis fails at `p2p-segwit.py` in one of the seven builds. Seems totally unrelated. Non-determinism?

laanwj:
Not sure if it's worrying but that must be unrelated. I've restarted the travis build.

theuni:
Test failure caused by attempting to spend a negative value in p2p_segwit: https://travis-ci.org/bitcoin/bitcoin/jobs/245346398#L3109

> CTransaction(nVersion=1 vin=[CTxIn(prevout=COutPoint(hash=6d7956f3bdf3f5c646146311bcf255305fcfbc7ccc60e6d70be216491278a54c n=0) scriptSig= nSequence=0)] vout=[CTxOut(nValue=-1.99996597 scriptPubKey=0014b472a266d0bd89c13706a4132ccfb16f7c3b9fcb)] ...

I've seen this failure once or twice before from Travis. Looks like the test isn't verifying that the utxo it uses for splitting is large enough, though I'm not sure how it could end up negative.

@jl2012 I think you added this test, mind having a look?

PR #10649, opened 2017-06-22
Make sure we only mine via the first wallet

jonasschnelli:
If one used `generate` while running with multiple wallets `-wallet=file1.dat wallet=file2.dat` the wallet used for the coinbase script is pretty undefined (or lets say weak defined, it's probably always the last one).

This PR is a quick fix (should be okay for the internal miner) to ensure we always use the first wallet for `generate | setgenerate`.

Ideally, someone works on passing the endpoint (coming soon) down to the signal listener.

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -3476,6 +3476,15 @@ void CWallet::GetAllReserveKeys(std::set<CKeyID>& setAddress) const
 
 void CWallet::GetScriptForMining(std::shared_ptr<CReserveScript> &script)
 {
+    // make sure we only mine via the first wallet
+    // otherwise the used wallet is undefined (depends on the signal registration order)
+    // TODO: allow to pass the via generate/setgenerate used wallet to the signal
+    {
+        LOCK2(cs_main, cs_wallet);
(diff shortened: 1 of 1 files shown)


laanwj:
I proposed this before but not sure what the outcome of the discussion was - what about making `generate` a wallet method?
This means Instead of using the first wallet, could simply use `GetWalletForJSONRPCRequest(request)`.

ryanofsky (commented):
Tested ACK b43a646b0de0c147d935605bf7e57f7211f631c9. This is a good fix for current unpredictable behavior. It's hard to write python tests that depend on multiwallet without this.

> I proposed this before but not sure what the outcome of the discussion was - what about making generate a wallet method?

Seems like a good idea. Would you lose the ability to generate blocks when wallet is disabled (if we even have this ability)?

laanwj:
> Seems like a good idea. Would you lose the ability to generate blocks when wallet is disabled (if we even have this ability)?

No-
- `generatetoaddress` will work without wallet
- `generate` needs wallet

jnewbery:
I think #10683 is a better solution here.

laanwj:
Closing in favor of #10683

PR #10658, opened 2017-06-23
German translation of '%n year(s)'

ReneNyffenegger:


jonasschnelli:
Thanks for submitting.

We use transiflex for our translations:
https://www.transifex.com/bitcoin/bitcoin/

We directly download the files from there, editing must be done at transiflex.

ReneNyffenegger:
Thanks for the reply. When I went to transifex, I didn't find the string "%n year(s)". It even explicitely says "0 strings to translate". I am at a complete loss how I'd use transifex.

jonasschnelli:
The `"%n year(s)"` has just been merged... the new files for translation (for 0.15) will be uploaded to transiflex soon.
Editing here would just result in being overwritten once we download the transiflex files.

ReneNyffenegger:
OK. I understand.

PR #10662, opened 2017-06-23
Initialize randomness in benchmarks

achow101:
Call RandomInit() in bench_bitcoin to initialize the RNG so that it does not cause an assertion error.

### DIFF
--- a/src/bench/bench_bitcoin.cpp
+++ b/src/bench/bench_bitcoin.cpp
@@ -7,10 +7,12 @@
 #include "key.h"
 #include "validation.h"
 #include "util.h"
+#include "random.h"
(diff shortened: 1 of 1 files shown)


sipa:
utACK 5155d1101eb4fc9d4d797b583bb29f71807bd10b

theuni:
Caught one! :)

utACK 5155d1101eb4fc9d4d797b583bb29f71807bd10b

sipa:
> Caught one! :)

Yes, not using hardware randomness in our *benchmarks* would indeed be devastating for security.

... Ok, I'll concede having an assertion was a good idea.

PR #10669, opened 2017-06-25
Pow: Introduce MaybeGenerateProof

jtimon:
This simplifies generateBlocks and further encapsulates the use of nNonce and nBits inside pow.o by introducing a new function MaybeGenerateProof.

This should be helpful for any attempt to support a testnet with signed blocks (see #9177 )

### DIFF
--- a/src/pow.cpp
+++ b/src/pow.cpp
@@ -89,3 +89,16 @@ bool CheckProofOfWork(uint256 hash, unsigned int nBits, const Consensus::Params&
 
     return true;
 }
+
+bool MaybeGenerateProof(const Consensus::Params& params, CBlockHeader* pblock, uint64_t& max_tries)
+{
+    static const int nInnerLoopCount = 0x10000;
+    while (max_tries > 0 && pblock->nNonce < nInnerLoopCount) {
+        if (CheckProofOfWork(pblock->GetHash(), pblock->nBits, params)) {
+            return true;
+        }
+        ++pblock->nNonce;
+        --max_tries;
+    }
+    return false;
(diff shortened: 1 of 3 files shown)


MeshCollider:
utACK https://github.com/bitcoin/bitcoin/pull/10669/commits/f91c9df38bfb74c897896a0d0c59a77539e9e26a

src/pow.cpp
```diff
@@ -89,3 +89,15 @@ bool CheckProofOfWork(uint256 hash, unsigned int nBits, const Consensus::Params&
...
+bool MaybeGenerateProof(const Consensus::Params& params, CBlockHeader* pblock, uint64_t& max_tries)
+{
+    static const int nInnerLoopCount = 0x10000;
+    uint256 blockHash = pblock->GetHash();
+    while (max_tries > 0 && pblock->nNonce < nInnerLoopCount && !CheckProofOfWork(blockHash, pblock->nBits, params)) {
+        ++pblock->nNonce;
+        blockHash = pblock->GetHash();
+        --max_tries;
+    }
+    return CheckProofOfWork(blockHash, pblock->nBits, params);
```
sipa: It seems unnecessary to call `CheckProofOfWork` twice. You can move it from the while loop into the body, and make it return true immediately when it succeeds.

jtimon:
Fixed @sipa 's nit.

jtimon:
Rebased for no good reason.

MarcoFalke:
Closing for now due to inactivity. We can reopen, when there is more appetite.

PR #10673, opened 2017-06-26
[qt] Avoid potential null pointer dereference in TransactionView::exportClicked()

practicalswift:
Avoid potential null pointer dereference in `TransactionView::exportClicked()`.

Issue introduced in commit 8969828d069e4e55108618a493749535edc12ec7.

### DIFF
--- a/src/qt/transactionview.cpp
+++ b/src/qt/transactionview.cpp
@@ -336,6 +336,10 @@ void TransactionView::changedAmount(const QString &amount)
 
 void TransactionView::exportClicked()
 {
+    if (!model || !model->getOptionsModel()) {
+        return;
+    }
+
     // CSV is currently the only supported format
     QString filename = GUIUtil::getSaveFileName(this,
         tr("Export Transaction History"), QString(),


src/qt/transactionview.cpp
```diff
@@ -355,7 +355,11 @@ void TransactionView::exportClicked()
     writer.addColumn(tr("Type"), TransactionTableModel::Type, Qt::EditRole);
     writer.addColumn(tr("Label"), 0, TransactionTableModel::LabelRole);
     writer.addColumn(tr("Address"), 0, TransactionTableModel::AddressRole);
-    writer.addColumn(BitcoinUnits::getAmountColumnTitle(model->getOptionsModel()->getDisplayUnit()), 0, TransactionTableModel::FormattedAmountRole);
+    if (model && model->getOptionsModel()) {
```
laanwj: In case `model` or `model->getOptionsModel()` is not set, there's no reason to continue here - there's no data to export.
Might as well return from the function early if so.

practicalswift:
@laanwj Fixed! Looks good? Is an error message dialog necessary?

laanwj:
> Is an error message dialog necessary?

That's not necessary. It's fine for the export to do nothing in that case.

practicalswift:
A bit of context might be needed for this change – I assumed `model` can be `NULL` when entering `exportClicked()` due to [this check](https://github.com/practicalswift/bitcoin/blob/32325a3f58fd607c29030b0dd6f2388579b9c926/src/qt/transactionview.cpp#L352-L353):

```c++
    if (model && model->haveWatchOnly())
        writer.addColumn(tr("Watch-only"), TransactionTableModel::Watchonly);
```

Let me know if that assumption is incorrect :-)

laanwj:
In normal use that shouldn't happen. However the view class should be able to work (at least: not crash) with a NULL model, for UI-only testing.

practicalswift:
@laanwj OK, thanks for the clarification!

jonasschnelli:
utACK fd9599b1358a314b073a9ca0a68ca8037915d91d (runtime protection is always good).

Though I having a hard time finding a possibility how the user could press the "Export" button while not having a walletmodel attached to the `TransactionView`.

PR #10676, opened 2017-06-27
document script-based return fields for validateaddress

instagibbs:


laanwj:
utACK f9dc6e5

src/rpc/misc.cpp
```diff
@@ -174,6 +174,14 @@ UniValue validateaddress(const JSONRPCRequest& request)
             "  \"ismine\" : true|false,        (boolean) If the address is yours or not\n"
             "  \"iswatchonly\" : true|false,   (boolean) If the address is watchonly\n"
             "  \"isscript\" : true|false,      (boolean) If the key is a script\n"
+            "  \"script\" : \"type\"           (string, optional) The output script type. Possible types: nonstandard, pubkey, pubkeyhash, scripthash, multisig, nulldata, witness_v0_keyhash, witness_v0_scripthash\n"
```
jonasschnelli: I guess this and L178 needs two whitespace more for the correct indentation...

instagibbs: I don't see the mis-alignment here?

jonasschnelli: L176 has two escapes (2x `\"`) while this line has four escapes. I'd expect that it will be off by two whitespace against the other lines.

instagibbs: I'm matching lines 185 through 187 at least, should those be indented further as well?

jonasschnelli: This is how it looks. Mind the `\"`.
bildschirmfoto 2017-06-29 um 20 57 24

Sorry for the bike-shed :)

instagibbs:
@jonasschnelli nit addressed

MIGUELWAXX:
eill commit

PR #10681, opened 2017-06-27
add gdb attach process to test README

instagibbs:
Saved me many hours. h/t to @jnewbery for the new guide efforts

### DIFF
--- a/test/README.md
+++ b/test/README.md
@@ -155,6 +155,26 @@ import pdb; pdb.set_trace()
 anywhere in the test. You will then be able to inspect variables, as well as
 call methods that interact with the bitcoind nodes-under-test.
 
+If further introspection of the bitcoind instances themselves becomes
+necessary, this can be accomplished by first setting a pdb breakpoint
+at an appropriate location, running the test to that point, then using
+`gdb` to attach to the process and debug.
+
+For instance, to attach to `self.node[1]` during a run:
+
+```bash
+2017-06-27 14:13:56.686000 TestFramework (INFO): Initializing test directory /tmp/user/1000/testo9vsdjo3
+```
+
+use the directory path to get the pid from the pid file:
+
+```bash
+cat /tmp/user/1000/testo9vsdjo3/node1/regtest/bitcoind.pid
+gdb /home/example/bitcoind <pid>
+```
+
+Note: gdb attach step may require `sudo`
+
 ### Util tests
 
 Util tests can be run locally by running `test/util/bitcoin-util-test.py`. 


test/README.md
```diff
@@ -155,6 +155,26 @@ import pdb; pdb.set_trace()
 anywhere in the test. You will then be able to inspect variables, as well as
 call methods that interact with the bitcoind nodes-under-test.
 
+If further introspection of the bitcoind instances themselves becomes
+necessary, this can be accomplished by first setting a pdb breakpoint
+at an appropriate location, running the test to that point, then using
+`gdb` to attach to the process and debug:
+
+```bash
```
jnewbery: I think this can be more easily achieved by `cat`ing the `.pid` file in the data directory. Find the data directory:

```bash
2017-06-27 14:13:56.686000 TestFramework (INFO): Initializing test directory /tmp/user/1000/testo9vsdjo3
```

Then `cat` the `.pid` file:

```bash
cat /tmp/user/1000/testo9vsdjo3/node0/regtest/bitcoind.pid
gdb /home/example/bitcoind !$
```

(or if you really have to show off):

```bash
gdb /home/example/bitcoind $(cat /tmp/user/1000/testo9vsdjo3/node0/regtest/bitcoind.pid)
```

jnewbery (commented):
ACK

One suggested change for finding the wanted bitcoind pid more easily.

test/README.md
jnewbery: perhaps also mention that this may require superuser privileges, so may need `sudo gdb ...`

instagibbs: I assume most will get "permission denied" and sudo but maybe it's best to say

instagibbs:
updated with simpler steps

test/README.md
jnewbery: Yeah - if gdb can't attach to a specific PID, the error message is actually buried in the middle of the gdb text so it's not immediately obvious what the problem is. I think including the warning is helpful here.

Also, my above invocation was wrong:

```bash
cat /tmp/user/1000/testo9vsdjo3/node0/regtest/bitcoind.pid
gdb /home/example/bitcoind !$
```

it should be:

```
cat /tmp/user/1000/testo9vsdjo3/node0/regtest/bitcoind.pid
gdb /home/example/bitcoind <pid>
```

(sorry - trying to be too clever!)

test/README.md
```diff
@@ -155,6 +155,26 @@ import pdb; pdb.set_trace()
...
+```
+
+use the directory path to get the pid from the pid file:
+
+```bash
+cat /tmp/user/1000/testo9vsdjo3/node1/regtest/bitcoind.pid
+gdb /home/example/bitcoind <pid>
+```
+
+Note: gdb attach step may require `sudo`
```
laanwj: I usually prefer
```
echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope
```
to running gdb itself as root.

jnewbery:
reACK

instagibbs:
squashed last commit, same text

jonasschnelli:
ACK.

I personally use a different approach (I guess it's more complicated, though will not need to attach and works with IDE's)
* replace a self.node object in the test with a custom added AuthProxy (to localhost and a fix port with a given rpcusername/pw)
* Start (via IDE, gdb) a such localhost regtest peer (make sure it has an empty datadir)
* run test

fanquake:
utACK 7ec3343

jnewbery:
Can merge please? Simple docs change with 3 ACKs.

MarcoFalke:
utACK 7ec3343df21aae30bf782f77171f5e4d9a1d5e65

PR #10686, opened 2017-06-28
Avoid usage of uninitialized values in function call arguments

practicalswift:
Avoid usage of uninitialized values in function call arguments.

Rationale:
> Avoid used-before-set errors and their associated undefined behavior. Avoid problems with comprehension of complex initialization. Simplify refactoring.

For a more thorough discussion, see ["ES.20: Always initialize an object"](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es20-always-initialize-an-object) in the C++ Core Guidelines (Stroustrup & Sutter).

### DIFF
--- a/src/bench/base58.cpp
+++ b/src/bench/base58.cpp
@@ -13,28 +13,30 @@
 
 static void Base58Encode(benchmark::State& state)
 {
-    unsigned char buff[32] = {
-        17, 79, 8, 99, 150, 189, 208, 162, 22, 23, 203, 163, 36, 58, 147,
-        227, 139, 2, 215, 100, 91, 38, 11, 141, 253, 40, 117, 21, 16, 90,
-        200, 24
+    std::array<unsigned char, 32> buff = {
+        {
+            17, 79, 8, 99, 150, 189, 208, 162, 22, 23, 203, 163, 36, 58, 147,
+            227, 139, 2, 215, 100, 91, 38, 11, 141, 253, 40, 117, 21, 16, 90,
+            200, 24
+        }
     };
-    unsigned char* b = buff;
     while (state.KeepRunning()) {
-        EncodeBase58(b, b + 32);
+        EncodeBase58(buff.begin(), buff.end());
     }
 }
 
 
 static void Base58CheckEncode(benchmark::State& state)
 {
-    unsigned char buff[32] = {
-        17, 79, 8, 99, 150, 189, 208, 162, 22, 23, 203, 163, 36, 58, 147,
-        227, 139, 2, 215, 100, 91, 38, 11, 141, 253, 40, 117, 21, 16, 90,
-        200, 24
+    std::array<unsigned char, 32> buff = {
+        {
+            17, 79, 8, 99, 150, 189, 208, 162, 22, 23, 203, 163, 36, 58, 147,
+            227, 139, 2, 215, 100, 91, 38, 11, 141, 253, 40, 117, 21, 16, 90,
+            200, 24
+        }
     };
-    unsigned char* b = buff;
     std::vector<unsigned char> vch;
-    vch.assign(b, b + 32);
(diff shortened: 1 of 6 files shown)


src/bench/base58.cpp
```diff
@@ -13,7 +13,7 @@
 
 static void Base58Encode(benchmark::State& state)
 {
-    unsigned char buff[32] = {
```
sipa: Why?

practicalswift: @sipa To get rid of `clang-tidy`, Clang Static Analyzer and other static analyser warnings along the lines of:
```c++
bench/base58.cpp:23:9: warning: Function call argument is a pointer to uninitialized value [clang-analyzer-core.CallAndMessage]
```

sipa: That seems overly pedantic...

MeshCollider: Is that because b+32 is passed to EncodeBase58 instead of b+31? So the cleaner way to do it would be to change that function instead of expanding the array by an unneeded element?

MeshCollider: Is that because b+32 is passed to the EncodeBase58 function and clang thinks the function might use it?

sipa: It's perfectly legal to use a pointer that points one past the end of an object (and the function does). You can't dereference it, though.

MeshCollider: Yeah but is that the only reason clang is giving that warning? I agree that it would be pedantic to add an extra element to the array just to get rid of that warning, seems very hacky

theuni: If we're going to change this for pedantic correctness (which I'm not arguing for), I'd 100x rather make this a ```std::array<unsigned char, 32>``` and iterate properly.

gmaxwell:
> "ES.20: Always initialize an object" 

I disagree.  Adding superfluous initialization can conceal real bugs when it prevents valgrind and MSAN from seeing code accessing uninitialized memory because it's been filled with dummy data.  This isn't a hypothetical concern-- I've both seen errors caught this way, and real errors hidden by excess initialization.

If the initialization the appropriate initial state, then by all means... but if any use of the initialized value would be a program error then it probably should not be initialized.

src/bench/base58.cpp
practicalswift: Now using `std::array<unsigned char, 32>`. Both `clang-tidy` and Clang Static Analyzer are now happy :-)

practicalswift:
@gmaxwell You have a good point there w.r.t. discovering issues via valgrind and MSAN.

In this specific case with these six changes - are there such risks associated with any of the changes introduced here? Let me know if so and I'll skip any such change.

MeshCollider:
From what I can see, the specific changes here should be fine, I believe it was more of a general note (which I do agree with).

laanwj:
> In this specific case with these six changes - are there such risks associated with any of the changes introduced here? Let me know if so and I'll skip any such change.

There are always risks associated with changes. Especially in consensus-critical code we should thus not make unnecessary ones.

practicalswift:
@laanwj Is there any of these six changes that touches on consensus-critical code and should be considered unnecessary? Let me know and I'll exclude from this PR :-)

sipa:
I think none of the changes here on themselves are risky, but all of them hide potential future bugs.

practicalswift:
Thanks for reviewing and thanks for some insightful comments with regards to `valgrind` and MSAN:s ability to flag for code accessing uninitialized memory.

The consensus is a clear NACK with regards to early initialization, so I'll close this PR.

I'll create a separate PR with with the non-initialization change: switching from `char[32]` to `std::array<unsigned char, 32>` to avoid warnings from various static analyzers (`clang-tidy`, clang analyzer, etc.) regarding pointers to uninitialized values as function call arguments when we use `f(arr, arr + 32)`.

PR #10689, opened 2017-06-28
fixing a couple of typos

ghost:


README.md
```diff
@@ -49,13 +49,13 @@ lots of money.
...
 
-Developers are strongly encouraged to write [unit tests](src/test/README.md) for new code, and to
+Developers are strongly encouraged to write [unit tests](src/test/README.md) for new code and to
 submit new unit tests for old code. Unit tests can be compiled and run
 (assuming they weren't disabled in configure) with: `make check`. Further details on running
 and extending unit tests can be found in [/src/test/README.md](/src/test/README.md).
 
 There are also [regression and integration tests](/test), written
-in Python, that are run automatically on the build server.
+in Python, that is run automatically on the build server.
```
laudaa: Tests is plural. Why are you changing from 'are' to 'is'?

MeshCollider:
NACK, I think the commas were better before, and the only other 'correction' is wrong. Don't think this should be merged.

fanquake:
Agree that nothing here seems to be fixing anything.

PR #10690, opened 2017-06-28
[qa] Bugfix: allow overriding extra_args in ComparisonTestFramework

sdaftuar:
I noticed that bip65-cltv-p2p.py was setting extra_args, but the ComparisonTestFramework was silently ignoring it.  Though I think this framework is likely going to be decommissioned, this is easy enough to fix for now.

### DIFF
--- a/test/functional/test_framework/test_framework.py
+++ b/test/functional/test_framework/test_framework.py
@@ -380,8 +380,10 @@ class ComparisonTestFramework(BitcoinTestFramework):
                           help="bitcoind binary to use for reference nodes (if any)")
 
     def setup_network(self):
+        extra_args = [['-whitelist=127.0.0.1']]*self.num_nodes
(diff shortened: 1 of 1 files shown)


jnewbery:
Tested ACK 4ed36539bcdebb360c730b2492b6435d7cd2476c

laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/10690/commits/4ed36539bcdebb360c730b2492b6435d7cd2476c

(I'm surprised the test was not failing with the wrong command line)

sdaftuar:
> (I'm surprised the test was not failing with the wrong command line)

So was I! That test needs to be overhauled in general though, as it makes no sense now that BIP65 activates on a specific height.

MarcoFalke:
utACK 4ed3653

PR #10694, opened 2017-06-28
Remove redundant code in MutateTxSign(CMutableTransaction&, const std::string&)

practicalswift:
Remove redundant code in `MutateTxSign(CMutableTransaction&, const std::string&)`.

### DIFF
--- a/src/bitcoin-tx.cpp
+++ b/src/bitcoin-tx.cpp
@@ -551,7 +551,6 @@ static void MutateTxSign(CMutableTransaction& tx, const std::string& flagStr)
     // mergedTx will end up with all the signatures; it
     // starts as a clone of the raw tx:
     CMutableTransaction mergedTx(txVariants[0]);
-    bool fComplete = true;
     CCoinsView viewDummy;
     CCoinsViewCache view(&viewDummy);
 
@@ -639,7 +638,6 @@ static void MutateTxSign(CMutableTransaction& tx, const std::string& flagStr)
         CTxIn& txin = mergedTx.vin[i];
         const Coin& coin = view.AccessCoin(txin.prevout);
         if (coin.IsSpent()) {
-            fComplete = false;
             continue;
         }
         const CScript& prevPubKey = coin.out.scriptPubKey;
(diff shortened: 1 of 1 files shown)


laanwj:
So do we really not care if the resulting signed transaction is complete or not?
(if we do, this a bug instead of redundant code)

practicalswift:
@laanwj Yes, I agree it looks odd :-)

practicalswift:
What should we do about this one? Shall I close? :-)

MarcoFalke:
Needs rebase due to travis bug

practicalswift:
@MarcoFalke Thanks! Now rebased! :-)

practicalswift:
What shall we do about this one? Leaving dead code around feels icky :-)

MarcoFalke:
utACK b1149ee4c3e4b7301017f779a6dec4415bcb733d

laanwj:
> What shall we do about this one? Leaving dead code around feels icky :-)

No one really answered my question. Do we care about this incompleteness? Should it be in the output?
I never use bitcoin-tx myself so I can't answer it.

jonasschnelli:
@practicalswift: care to elaborate why this is dead code?
I haven't check all possible states, but for me it looks like that `VerifyScript()` after combining signatures and updated the transaction makes pretty sense... but maybe I miss something.

practicalswift:
Pushed an updated version which only removes trivial dead code. Now keeping the side-effects of `VerifyScript(…)`.

Thanks for reviewing @jonasschnelli 

Please re-review :-)

jonasschnelli:
This seems to make sense now.
utACK 347fd7ca1742183707541ef186f97218aab6e0a9

practicalswift:
@jonasschnelli Would you mind re-reviewing? Had to force push to get Travis to rebuild :-)  No changes since 347fd7c.

src/bitcoin-tx.cpp
```diff
@@ -654,14 +652,7 @@ static void MutateTxSign(CMutableTransaction& tx, const std::string& flagStr)
...
         UpdateTransaction(mergedTx, i, sigdata);
-
-        if (!VerifyScript(txin.scriptSig, prevPubKey, &txin.scriptWitness, STANDARD_SCRIPT_VERIFY_FLAGS, MutableTransactionSignatureChecker(&mergedTx, i, amount)))
-            fComplete = false;
-    }
-
-    if (fComplete) {
-        // do nothing... for now
-        // perhaps store this for later optional JSON output
+        VerifyScript(txin.scriptSig, prevPubKey, &txin.scriptWitness, STANDARD_SCRIPT_VERIFY_FLAGS, MutableTransactionSignatureChecker(&mergedTx, i, amount));
```
MarcoFalke: The return code is ignored and you are only passing const references. Mind to elaborate on the side effects this has?

practicalswift:
Updated version. Please re-review :-)

@jonasschnelli - did I misunderstand you regarding the call to `VerifyScript()`? As @MarcoFalke  commented there seems to be no side-effects and the return value is not used so no need to keep it around? What am I missing? :-)

MarcoFalke:
The current commit is now identical to the original commit (b1149ee). I'd prefer you force push b1149ee, so I don't have to re-ACK. ;)

practicalswift:
@MarcoFalke Fixed :-)

promag:
utACK b1149ee4.

Issue #10678, opened 2017-06-27
zapwallettxes.py test failure due to bitcoind return code


laanwj:
In #10677 there was what looks like a new kind of Travis false positive, or at least an intermittent error.
It is triggered by the assertion on return code added in #10636:
```
zapwallettxes.py failed, Duration: 11 s
stdout:
2017-06-27 03:50:43.814000 TestFramework (INFO): Initializing test directory /tmp/bitcoin_test_runner_20170627_034653/zapwallettxes_338
2017-06-27 03:50:46.305000 TestFramework (INFO): Mining blocks...
2017-06-27 03:50:54.465000 TestFramework (ERROR): Assertion failed
Traceback (most recent call last):
  File "/home/travis/build/bitcoin/bitcoin/build/bitcoin-x86_64-unknown-linux-gnu/test/functional/test_framework/test_framework.py", line 153, in main
    self.run_test()
  File "/home/travis/build/bitcoin/bitcoin/build/bitcoin-x86_64-unknown-linux-gnu/test/functional/zapwallettxes.py", line 67, in run_test
    self.stop_node(0)
  File "/home/travis/build/bitcoin/bitcoin/build/bitcoin-x86_64-unknown-linux-gnu/test/functional/test_framework/test_framework.py", line 217, in stop_node
    _stop_node(self.nodes[num_node], num_node)
  File "/home/travis/build/bitcoin/bitcoin/build/bitcoin-x86_64-unknown-linux-gnu/test/functional/test_framework/util.py", line 326, in _stop_node
    assert_equal(return_code, 0)
  File "/home/travis/build/bitcoin/bitcoin/build/bitcoin-x86_64-unknown-linux-gnu/test/functional/test_framework/util.py", line 440, in assert_equal
    raise AssertionError("not(%s)" % " == ".join(str(arg) for arg in (thing1, thing2) + args))
AssertionError: not(1 == 0)
2017-06-27 03:50:54.466000 TestFramework (INFO): Stopping nodes
stderr:
Traceback (most recent call last):
  File "/home/travis/build/bitcoin/bitcoin/build/bitcoin-x86_64-unknown-linux-gnu/test/functional/test_framework/authproxy.py", line 126, in _request
    return self._get_response()
  File "/home/travis/build/bitcoin/bitcoin/build/bitcoin-x86_64-unknown-linux-gnu/test/functional/test_framework/authproxy.py", line 169, in _get_response
    http_response = self.__conn.getresponse()
[...]

PR #10695, opened 2017-06-28
[qa] Rewrite BIP65/BIP66 functional tests

sdaftuar:
After 122786d0e0170c73536360b705af711e1338adbf, BIP65 and BIP66 activate at
particular fixed heights (without regard to version numbers of blocks
below those heights).  Rewrite the functional tests to take
this into account, and remove two tests that weren't really testing anything.

Moves the rewritten functional tests out of the extended test suite, so that they run in travis regularly.

Note: I discovered that the ComparisonTestFramework (which the original versions of these p2p tests were written is, has a bug that caused them to not catch obvious errors, eg if you just comment out setting the script flags for these softforks in ConnectBlock, the versions of these tests in master do not fail(!) -- will separately PR a fix for the comparison test framework).


TheBlueMatt (commented):
Looks good to me at 06e3f006e499e3238905b92cbca563c69e170410, my test framework/python chops dont make me want to ACK, but I did test a few cases that the test actually failed if tweaked.

test/functional/bipdersig-p2p.py
```diff
@@ -64,119 +56,74 @@ def create_transaction(self, node, coinbase, to_address, amount):
...
+        block_time += 1
+        block = create_block(tip, create_coinbase(DERSIG_HEIGHT), block_time)
         block.nVersion = 2
         block.rehash()
         block.solve()
-        self.last_block_time += 1
-        self.tip = block.sha256
-        height += 1
-        yield TestInstance([[block, True]])
+        node0.send_and_ping(msg_block(block))
```
jnewbery: The node-under-test sends a reject message at this point:

```
2017-06-29 08:56:00.945000 TestFramework.mininode (DEBUG): Received message from 127.0.0.1:11900: msg_reject: b'block' 17 b'bad-version(0x00000002)' [7bb0412e9813a1d13cd14ff2277a906e101d183f93b0136c848bdb94ffd1d6b7]
```

It'd be nice to add an assert that mininode received that reject.

I know that there's talk of removing reject messages from bitcoind, but while they're still there, they're useful for asserting that a block/tx was rejected for the correct reason. If there was some error in the code above (eg if we'd called `create_block()` with bad parameters or it'd created a bad block for some reason), the block would have been rejected, the test would pass, but we wouldn't actually be testing BIP66 activation.

test/functional/bipdersig-p2p.py
```diff
@@ -64,119 +56,74 @@ def create_transaction(self, node, coinbase, to_address, amount):
...
         block.hashMerkleRoot = block.calc_merkle_root()
         block.rehash()
         block.solve()
-        self.last_block_time += 1
-        yield TestInstance([[block, False]])
 
-        ''' Mine 1 old version block, should be invalid '''
-        block = create_block(self.tip, create_coinbase(height), self.last_block_time + 1)
-        block.nVersion = 2
+        node0.send_and_ping(msg_block(block))
```
jnewbery: Again, it'd be good to assert that we received a reject, this time with cause `'block-validation-failed'`

test/functional/bipdersig-p2p.py
```diff
@@ -4,27 +4,19 @@
...
-from test_framework.test_framework import ComparisonTestFramework
+from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import *
-from test_framework.mininode import CTransaction, NetworkThread
+from test_framework.mininode import *
 from test_framework.blocktools import create_coinbase, create_block
-from test_framework.comptool import TestInstance, TestManager
 from test_framework.script import CScript
 from io import BytesIO
 import time
```
jnewbery: nit: this import is unused and can be removed

test/functional/bipdersig-p2p.py
```diff
@@ -4,27 +4,19 @@
...
 
+DERSIG_HEIGHT = 1251
+
 # A canonical signature consists of:
```
jnewbery: nit: trailing whitespace

test/functional/bipdersig-p2p.py
```diff
@@ -40,18 +32,18 @@ def unDERify(tx):
         else:
             newscript.append(i)
     tx.vin[0].scriptSig = CScript(newscript)
+
+class BaseNode(NodeConnCB):
+    def __init__(self):
+        super().__init__()
```
jnewbery: nit: trailing whitespace

test/functional/bipdersig-p2p.py
```diff
@@ -64,119 +56,74 @@ def create_transaction(self, node, coinbase, to_address, amount):
...
-        self.last_block_time += 1
-        self.tip = block.sha256
-        height += 1
-        yield TestInstance([[block, True]])
+        node0.send_and_ping(msg_block(block))
+        assert_equal(int(self.nodes[0].getbestblockhash(), 16), tip)
 
-        ''' Mine 1 new version block '''
-        block = create_block(self.tip, create_coinbase(height), self.last_block_time + 1)
+        self.log.info("Test that transactions with non-DER signatures cannot appear in a block")
```
jnewbery: Are transactions with non-DER signatures added to the mempool or relayed after BIP66 activation? If not, is it possible to add a sub-test here to test that?

test/functional/bip65-cltv-p2p.py
```diff
@@ -4,27 +4,20 @@
...
+from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import *
-from test_framework.mininode import CTransaction, NetworkThread
+from test_framework.mininode import *
 from test_framework.blocktools import create_coinbase, create_block
-from test_framework.comptool import TestInstance, TestManager
-from test_framework.script import CScript, OP_1NEGATE, OP_CHECKLOCKTIMEVERIFY, OP_DROP
+from test_framework.script import CScript, OP_1NEGATE, OP_CHECKLOCKTIMEVERIFY, OP_DROP, CScriptNum
 from io import BytesIO
 import time
```
jnewbery: nit: import time not used

test/functional/bip65-cltv-p2p.py
```diff
@@ -33,19 +26,34 @@ def cltv_invalidate(tx):
...
+    Prepends <height> CLTV DROP in the scriptSig, and sets
+    the locktime to height'''
+    tx.vin[0].nSequence = 0
+    tx.nLockTime = height
+
+    # Need to re-sign, since nSequence and nLockTime changed
+    signed_result = node.signrawtransaction(ToHex(tx))
+    new_tx = CTransaction()
+    f = BytesIO(hex_str_to_bytes(signed_result['hex']))
+    new_tx.deserialize(f)
```
jnewbery: nit: perhaps combine this with the line above, rather than have a throwaway variable.

test/functional/bip65-cltv-p2p.py
```diff
@@ -33,19 +26,34 @@ def cltv_invalidate(tx):
...
 
-class BIP65Test(ComparisonTestFramework):
+    new_tx.vin[0].scriptSig = CScript([CScriptNum(height), OP_CHECKLOCKTIMEVERIFY, OP_DROP] +
+                                  list(CScript(new_tx.vin[0].scriptSig)))
+    return new_tx
 
+class BaseNode(NodeConnCB):
```
jnewbery: Not required. Just use the base `NodeConnCB` below.

test/functional/bip65-cltv-p2p.py
```diff
@@ -33,19 +26,34 @@ def cltv_invalidate(tx):
...
-        test.run()
+class BIP65Test(BitcoinTestFramework):
+
+    def __init__(self):
+        super().__init__()
+        self.num_nodes = 1
+        self.extra_args = [['-promiscuousmempoolflags=1', '-whitelist=127.0.0.1']]
+        self.setup_clean_chain = True
 
     def create_transaction(self, node, coinbase, to_address, amount):
```
jnewbery: This doesn't need to be a method in `BIP65Test`. Consider moving it above the class definition to be a module-level function (alongside `cltv_invalidate()` and `cltv_validate()`)

test/functional/bip65-cltv-p2p.py
```diff
@@ -58,119 +66,79 @@ def create_transaction(self, node, coinbase, to_address, amount):
         tx.deserialize(f)
         return tx
 
-    def get_tests(self):
+    def run_test(self):
+        node0 = BaseNode()
```
jnewbery: `self.nodes[0]` is used throughout this test. Consider adding an alias:

```py
node = self.nodes[0]
```
and then using `node` throughout the test.

I think that looks slightly neater for tests that only use one node, but that's just personal preference.

test/functional/bip65-cltv-p2p.py
```diff
@@ -58,119 +66,79 @@ def create_transaction(self, node, coinbase, to_address, amount):
...
 
-        block = create_block(self.tip, create_coinbase(height), self.last_block_time + 1)
-        block.nVersion = 4
+        tip = self.nodes[0].getbestblockhash()
+        block_time = self.nodes[0].getblockheader(tip)['mediantime'] + 1
+        block = create_block(int(tip, 16), create_coinbase(CLTV_HEIGHT - 1), block_time)
+        block.nVersion = 3
         block.vtx.append(spendtx)
         block.hashMerkleRoot = block.calc_merkle_root()
         block.rehash()
```
jnewbery: `block.solve()` calls `block.rehash()`, so there's no need for all of these explicit calls to `block.rehash()`. Up to you if you want to remove them all.

test/functional/bip65-cltv-p2p.py
```diff
@@ -58,119 +66,79 @@ def create_transaction(self, node, coinbase, to_address, amount):
...
+        block_time += 1
+        block = create_block(tip, create_coinbase(CLTV_HEIGHT), block_time)
         block.nVersion = 3
         block.rehash()
         block.solve()
-        self.last_block_time += 1
-        self.tip = block.sha256
-        height += 1
-        yield TestInstance([[block, True]])
+        node0.send_and_ping(msg_block(block))
```
jnewbery: I've added a comment to bipdersig-p2p.py about asserting on reject messages. Same comment applies here

jnewbery (commented):
Very nice cleanup. I love removing redundant tests!

A few nits below. Nothing super important, although I would like to see asserts on reject messages, to ensure that blocks are being rejected for the correct reason.

It'd be nice to rename these (and other tests) to remove the `-p2p` suffix, since that now doesn't mean much. That could be done in a future PR if people agree.

Consider using a linter like flake8 (together with the vim syntastic plugin) for catching obvious style nits like unused imports and trailing whitespace.

Even without the nit fixes, this is a definite improvement over the current tests. Tested ACK 06e3f006e499e3238905b92cbca563c69e170410

test/functional/bip65-cltv-p2p.py
```diff
@@ -33,19 +26,34 @@ def cltv_invalidate(tx):
     tx.vin[0].scriptSig = CScript([OP_1NEGATE, OP_CHECKLOCKTIMEVERIFY, OP_DROP] +
                                   list(CScript(tx.vin[0].scriptSig)))
 
+def cltv_validate(node, tx, height):
+    '''Modify the signature in vin 0 of the tx to pass CLTV
+    Prepends <height> CLTV DROP in the scriptSig, and sets
+    the locktime to height'''
+    tx.vin[0].nSequence = 0
+    tx.nLockTime = height
```
sdaftuar: Agree that this would be nice, but left this as a todo for now.

test/functional/bipdersig-p2p.py
sdaftuar: Default policy is to enforce DERSIG and CLTV on all transactions entering the mempool.  I suppose it might be nice to add a test that the policy is enforcing these rules, but in this test I set custom script flags so that I could show that a transaction is valid under script flags that don't include the soft fork, but invalid under the soft fork.

It might be nice to add a second node to these tests that uses default policy so that we can test that default policy is enforcing these script flags, but I would prefer to save that for a future PR.

sdaftuar:
@jnewbery Thanks for the review; I addressed all your nits except the one I responded to.

test/functional/bip65-cltv-p2p.py
sdaftuar: Also, I did add a test that shows that the transaction is accepted to the node's mempool, which is using -promiscuousmempoolflags so that it's not enforcing the soft-fork.  So hopefully that helps assuage the concern that the transaction might be invalid for non-soft-fork related reasons.

(Also, this helps sneak in a test that is useful post-#10192, by verifying that a block is invalid even if we may have cached validation success using a different set of flags.)

sdaftuar:
Bumped travis after spurious zapwallettxs error popped up (appeared to be an instance of #10678)

jnewbery:
Tested ACK https://github.com/bitcoin/bitcoin/pull/10695/commits/4ccc12a54a217892bd4ccfd94c46052a11cdb3fb. It'd be nice to add asserting on reject messages, but that can wait for a future PR.

MarcoFalke:
utACK 4ccc12a54a217892bd4ccfd94c46052a11cdb3fb

PR #9544, opened 2017-01-13
[trivial] Add end of namespace comments. Improve consistency.

practicalswift:

PR #10702, opened 2017-06-29
[Trivial] Improve end-of-namespace comment consistency

meshcollider:
Continuing on from #9544 this just makes the end-of-namespace comments consistent across all files

While purely style changes are not normally acceptable, this one should cause very few merge conflicts just because namespaces aren't usually touched often, and because of the nature of namespace indentation, I believe its quite a readability improvement to keep track of what the closing braces belong to.

### DIFF
--- a/src/memusage.h
+++ b/src/memusage.h
@@ -164,6 +164,6 @@ static inline size_t DynamicUsage(const std::unordered_map<X, Y, Z>& m)
     return MallocUsage(sizeof(unordered_node<std::pair<const X, Y> >)) * m.size() + MallocUsage(sizeof(void*) * m.bucket_count());
 }
 
-}
+}  // namespace memusage
 
 #endif // BITCOIN_MEMUSAGE_H
--- a/doc/developer-notes.md
+++ b/doc/developer-notes.md
@@ -450,11 +450,13 @@ Source code organization
   should be placed on the same line as the brace closing the namespace, e.g.
 
 ```c++
-namespace mynamespace {
+namespace mynamespace
+{
     ...
 } // namespace mynamespace
 
-namespace {
+namespace
+{
     ...
 } // namespace
 ```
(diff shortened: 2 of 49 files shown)


laanwj:
Meh, why all the end-of-namespace pulls lately? We don't have a clear rule for these, no need to have a PR every three days that updates them.
Also `src/leveldb` changes should go into the appropriate upstream repository.

MeshCollider:
Just saw #9544 hadn't finished the job, because that was merged then this just makes it all consistent.
Good point on the leveldb though

MeshCollider:
Removed changes to src/leveldb/

sipa:
The style guide says that braces must be on the next line for namespaces, classes, methods, and functions.

MeshCollider:
Almost all the existing namespaces have the brace on the same line, would you prefer if they were all changed in this PR too?

sipa:
I don't think there is a need for a massive change one way or another, but certainly not for a change in the wrong direction.

MeshCollider:
Refactored as per style guide, no extra effort to get it fixed really :)

promag (changes requested):
Missing documentation fix for https://github.com/bitcoin/bitcoin/commit/f3f1e2e7d38005fa3edecb01469a18a6f7500ea6.

MeshCollider:
@promag What do you mean?

src/memusage.h
```diff
@@ -17,8 +17,7 @@
 
 #include <boost/foreach.hpp>
 
-namespace memusage
-{
+namespace memusage {
```
promag: Remove change?

promag (changes requested):
At the end I would keep two commits like:
 - Fix namespace brace style (includes documentation fix)
 - Improve end-of-namespace comment consistency

Other than these, ACK.

MeshCollider:
Decided to squash them all into one commit because its all refactoring, and the previous commits were already squashed. Thanks for pointing out that missed brace @promag

fanquake:
Going to close this for now. You could rebase, extend the changes and reopen just after the next major branch off, as that is usually the best time to merge commits that touch a lot of files.

PR #10703, opened 2017-06-29
[tests] Allow tests to pass when stderr is non-empty

jnewbery:
Resurrect #10241 with nits addressed

Not sure how much people want this. Would be useful for functional tests which cause bitcoind to print to stderr.

### DIFF
--- a/test/functional/test_runner.py
+++ b/test/functional/test_runner.py
@@ -27,8 +27,7 @@ import tempfile
 import re
 import logging
 
-# Formatting. Default colors to empty strings.
-BOLD, BLUE, RED, GREY = ("", ""), ("", ""), ("", ""), ("", "")
+# Formatting.
 try:
     # Make sure python thinks it can write unicode to its stdout
     "\u2713".encode("utf_8").decode(sys.stdout.encoding)
@@ -40,16 +39,27 @@ except UnicodeDecodeError:
     CROSS = "x "
     CIRCLE = "o "
 
+# Default colors to empty strings.
+BOLD, BLUE, RED, GREY, MAGENTA = [("", "")] * 5
 if os.name == 'posix':
     # primitive formatting on supported
     # terminal via ANSI escape sequences:
     BOLD = ('\033[0m', '\033[1m')
-    BLUE = ('\033[0m', '\033[0;34m')
-    RED = ('\033[0m', '\033[0;31m')
     GREY = ('\033[0m', '\033[1;30m')
+    RED = ('\033[0m', '\033[0;31m')
+    BLUE = ('\033[0m', '\033[0;34m')
+    MAGENTA = ('\033[0m', '\033[0;35m')
 
 TEST_EXIT_PASSED = 0
 TEST_EXIT_SKIPPED = 77
+
+STATUS_PASSED = "Passed"
+STATUS_PASSED_WITH_WARNINGS = "Passed with warnings"
+STATUS_SKIPPED = "Skipped"
+STATUS_FAILED = "Failed"
+STATUSES = [STATUS_PASSED, STATUS_PASSED_WITH_WARNINGS, STATUS_SKIPPED, STATUS_FAILED]
+
+STATUS_MAX_LEN = max([len(st) for st in STATUSES])
 
 BASE_SCRIPTS= [
     # Scripts that are run by the travis build process.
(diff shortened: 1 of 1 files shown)


test/functional/test_runner.py
```diff
@@ -40,17 +39,28 @@
     CROSS = "x "
     CIRCLE = "o "
 
+# Default colors to empty strings.
+BOLD, BLUE, RED, GREY, MAGENTA = [("", "")] * 5
```
jimmysong: nit: alpha sort other than BOLD?

test/functional/test_runner.py
```diff
@@ -40,17 +39,28 @@
...
 if os.name == 'posix':
     # primitive formatting on supported
     # terminal via ANSI escape sequences:
     BOLD = ('\033[0m', '\033[1m')
-    BLUE = ('\033[0m', '\033[0;34m')
-    RED = ('\033[0m', '\033[0;31m')
     GREY = ('\033[0m', '\033[1;30m')
+    RED = ('\033[0m', '\033[0;31m')
+    BLUE = ('\033[0m', '\033[0;34m')
+    MAGENTA = ('\033[0m', '\033[0;35m')
```
jimmysong: nit: alpha sort?

test/functional/test_runner.py
```diff
@@ -40,17 +39,28 @@
...
 
 TEST_EXIT_PASSED = 0
 TEST_EXIT_SKIPPED = 77
 
+STATUS_PASSED = "Passed"
+STATUS_PASSED_WITH_WARNINGS = "Passed with warnings"
+STATUS_SKIPPED = "Skipped"
+STATUS_FAILED = "Failed"
+STATUSES = [STATUS_PASSED, STATUS_PASSED_WITH_WARNINGS, STATUS_SKIPPED, STATUS_FAILED]
```
jimmysong: This doesn't seem to be used anywhere else, maybe combine with line below?

jimmysong (commented):
utACK, a few nits.

test/functional/test_runner.py
MeshCollider: IMO its cleaner how it is, shouldn't be combined

test/functional/test_runner.py
```diff
@@ -422,17 +436,20 @@ def __init__(self, name, status, time):
...
+        if self.status == STATUS_PASSED:
             color = BLUE
             glyph = TICK
-        elif self.status == "Failed":
-            color = RED
-            glyph = CROSS
-        elif self.status == "Skipped":
+        if self.status == STATUS_PASSED_WITH_WARNINGS:
+            color = MAGENTA
+            glyph = TICK
```
MeshCollider: Is TICK the best glyph to use for passed with warnings? Or something more neutral?

MeshCollider:
utACK

sipa:
The code here looks like mostly output formatting changes, not actually dealing with the issue described by the title?

test/functional/test_runner.py
```diff
@@ -40,17 +39,28 @@
...
 TEST_EXIT_PASSED = 0
 TEST_EXIT_SKIPPED = 77
 
+STATUS_PASSED = "Passed"
+STATUS_PASSED_WITH_WARNINGS = "Passed with warnings"
+STATUS_SKIPPED = "Skipped"
+STATUS_FAILED = "Failed"
+STATUSES = [STATUS_PASSED, STATUS_PASSED_WITH_WARNINGS, STATUS_SKIPPED, STATUS_FAILED]
+
+STATUS_MAX_LEN = max([len(st) for st in STATUSES])
```
ryanofsky: Square brackets unneeded and probably should be dropped for slightly better readability & efficiency.

ryanofsky (commented):
utACK d64ac3f4aab13a1a4e8d320b52e7a2b934e011b2. Note that #10882 is currently depending on this so it needs 0.15 milestone.

> The code here looks like mostly output formatting changes, not actually dealing with the issue described by the title?

"Allow tests to pass" implies output formatting has to change at least somewhat to avoid giving impression  that tests failed when they actually succeeded, and these changes are a pretty straightforward way of updating the output.

MarcoFalke:
utACK d64ac3f4aab13a1a4e8d320b52e7a2b934e011b2

laanwj:
Reverted this in 60f9778 after discussion on IRC.
Sorry @MarcoFalke but there seems no agreement on doing this change.
It's unfortunate that #10882 requires this, but it seems like we'll have to look for an alternative solution there.

PR #10659, opened 2017-06-23
[qa] blockchain: Pass on closed connection during generate call

maflcko:
Identified by @sdaftuar https://github.com/bitcoin/bitcoin/pull/10632#discussion_r123657939

PR #10704, opened 2017-06-29
[tests] nits in dbcrash.py

jnewbery:
@sdaftuar - I never reviewed the dbcrash.py test in #10148 . This PR fixes up a few nits:

- tidy up imports (don't import `time` indirectly, no need for try-except importing `httplib` since that's a python 2/3 compatibility trick, and our tests always run in Py3)
- move function-level comments to docstrings
- combine module-level docstrings
- a few other small whitespace, etc nits

low-priority, should be fairly uncontroversial, but let me know if you disagree with any of these.

### DIFF
--- a/test/functional/dbcrash.py
+++ b/test/functional/dbcrash.py
@@ -2,21 +2,7 @@
 # Copyright (c) 2017 The Bitcoin Core developers
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
-"""Test recovery from a crash during chainstate writing."""
-
-from test_framework.test_framework import BitcoinTestFramework
-from test_framework.util import *
-from test_framework.script import *
-from test_framework.mininode import *
-import random
-try:
-    import http.client as httplib
-except ImportError:
-    import httplib
-import errno
-
-'''
-Test structure:
+"""Test recovery from a crash during chainstate writing.
 
 - 4 nodes
   * node0, node1, and node2 will have different dbcrash ratios, and different
@@ -37,11 +23,26 @@ Test structure:
        * submit block to node
        * if node crashed on/after submitting:
          - restart until recovery succeeds
-         - check that utxo matches node3 using gettxoutsetinfo
-'''
+         - check that utxo matches node3 using gettxoutsetinfo"""
+
+import errno
+import http.client
+import random
+import sys
+import time
+
+from test_framework.mininode import *
+from test_framework.script import *
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import *
+
+HTTP_DISCONNECT_ERRORS = [http.client.CannotSendRequest]
+try:
+    HTTP_DISCONNECT_ERRORS.append(http.client.RemoteDisconnected)
+except AttributeError:
+    pass
 
 class ChainstateWriteCrashTest(BitcoinTestFramework):
-
     def __init__(self):
         super().__init__()
         self.num_nodes = 4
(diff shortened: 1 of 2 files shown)


laanwj:
Looks like this test is acting up: 
https://travis-ci.org/bitcoin/bitcoin/jobs/248398016
https://travis-ci.org/bitcoin/bitcoin/jobs/248398018
Might need a similar change to #10659.

test/functional/dbcrash.py
```diff
@@ -37,45 +23,47 @@
...
+
+from test_framework.mininode import *
+from test_framework.script import *
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import *
 
+class ChainstateWriteCrashTest(BitcoinTestFramework):
     def __init__(self):
         super().__init__()
-        self.num_nodes = 4
```
sdaftuar: I don't feel strongly, but the reason I put this in was that I didn't think that changes to the default test framework should cause tests to break -- since this test relies on a particular configuration, I felt it was best to define that locally.

sdaftuar: Also, can you set a custom timewait for the RPC connections in this test to resolve the travis issues that have come up so far:
https://travis-ci.org/bitcoin/bitcoin/jobs/248398018#L2153

test/functional/dbcrash.py
```diff
@@ -99,14 +87,16 @@ def restart_node(self, node_index, expected_tip):
...
+        """Try submitting a block to the given node.
+
+        Catch any exceptions that indicate the node has crashed.
+        Returns true if the block was submitted successfully; false otherwise."""
+
         try:
             self.nodes[node_index].submitblock(block)
             return True
-        except (httplib.CannotSendRequest, httplib.RemoteDisconnected) as e:
+        except (http.client.CannotSendRequest, http.client.RemoteDisconnected) as e:
```
sdaftuar: I think this RemoteDisconnected exception may not exist pre-3.5, which would explain this travis failure:
https://travis-ci.org/bitcoin/bitcoin/jobs/248398016#L3321

I'm not sure what the best way to handle that is...

test/functional/dbcrash.py
```diff
@@ -175,9 +168,8 @@ def verify_utxo_hash(self):
                 nodei_utxo_hash = self.restart_node(i, self.nodes[3].getbestblockhash())
             assert_equal(nodei_utxo_hash, node3_utxo_hash)
 
-
     def generate_small_transactions(self, node, count, utxo_list):
-        FEE = 1000 # TODO: replace this with node relay fee based calculation
+        FEE = 1000  # TODO: replace this with node relay fee based calculation
```
sdaftuar: Feel free to address the TODO as well :)  I think you just do something like `relayfee = self.nodes[0].getnetworkinfo()['relayfee']`, and then convert to the right units.

sdaftuar (commented):
Looks good, thanks for cleaning this up.  Please take a look at the fixes needed to make travis happy as well.

test/functional/dbcrash.py
MarcoFalke: We might as well just get rid of the default value. There is nothing special about "4" that justifies to use it as default value.

jnewbery: ok, you've convinced me. I'll leave this in for now.

We can remove the default num_nodes in a future PR (and probably remove the default topology and premine while we're at it.

test/functional/dbcrash.py
jnewbery: I think I've fixed this in my fixup commit b394a00f6c34e20e5ff555c8a07c6558e319a34a. Just catch different errors depending on which version of Python is being run.

I think later we should move this logic to a separate method named `send_rpc_may_fail()` or similar in TestNode to keep the test script code clean.

test/functional/dbcrash.py
jnewbery: I haven't addressed this. What are the downsides to using a fixed fee?

jnewbery:
I think I've fixed the Travis issues, although it's difficult to tell for sure since Travis only runs extended tests in master.

For the first, I split the generate call in `create_confirmed_utxos()` into many smaller generate calls. For the second, I catch different errors depending on the Python version.

Both of these solutions should be moved into TestNode once that's merged (ie a `node_generate()` method which automatically breaks the generate call into smaller pieces, and a `send_rpc_may_fail()` method for RPCs that may or may not time out).

I haven't addressed the fixed fee todo. I don't think that's too much of a problem, but let me know if you think it could cause issues.

One question: I'm confused about the randomness in re-org'ing. As far as I can see, the way this works:

- pick a block at random between the start of the test and the current tip.
- pick random x in [0,1]. If x < 1 / (depth of random block + 4), then re-org back to random block
- build to current tip + 1
(repeat x 40)

That throws up events with very low probabilities. For example, we'll re-org back 40 blocks in 1/40 * 1/44 = 1/1760 tests. For me, such indeterminateness is troubling. If there was a problem that caused large re-orgs to fail, then they'd only be caught very intermittently by this test, and it'd be difficult to debug.

I prefer my tests to be more deterministic where possible. How would you feel about changing the re-orgs to a fixed schedule (ie make sure the test covers re-orgs of 1, 2, 4, 8, 16 and 32 blocks, or some similar schedule)?

MarcoFalke:
utACK b394a00f6c34e20e5ff555c8a07c6558e319a34a. Feel free to squash.

jnewbery:
squashed

MarcoFalke:
Going to merge this now, since it might fix the test failures we see on master right now. Further cleanup can always be completed in later commits.

sdaftuar:
> That throws up events with very low probabilities. For example, we'll re-org back 40 blocks in 1/40 * 1/44 = 1/1760 tests. For me, such indeterminateness is troubling. If there was a problem that caused large re-orgs to fail, then they'd only be caught very intermittently by this test, and it'd be difficult to debug.

> I prefer my tests to be more deterministic where possible. How would you feel about changing the re-orgs to a fixed schedule (ie make sure the test covers re-orgs of 1, 2, 4, 8, 16 and 32 blocks, or some similar schedule)?

@jnewbery I don't think the right way to think of this test is that we're testing reorgs.  Really, we're testing a more complex state (utxo cache + disk state) and trying to get as much coverage of various edge cases as we can.  Since it's hard for me to imagine all the possible states we could be in, I wanted to simulate as much normal looking bitcoin transaction volume as I reasonably could, and then randomly stress it with reorgs and crashes in the hopes that if an edge case had a bug, we'd eventually see it.

If I knew exactly how to enumerate all the edge cases, then I'd happily convert this to a deterministic set of tests, but lacking that I think the randomness here is an appropriate test strategy.

jnewbery:
> I don't think the right way to think of this test ... , but lacking that I think the randomness here is an appropriate test strategy.

ACK

PR #10710, opened 2017-06-30
REST/RPC example update

Mirobit:
* Added missing fields and fixed order in REST.doc
* Removed rpc parameter from getaddednodeinfo example


fanquake:
utACK b8bb425

laanwj:
utACK b8bb425

PR #10637, opened 2017-06-20
Coin Selection with Murch's algorithm

achow101:
This is an implementation of the [Branch and Bound coin selection algorithm written by Murch](http://murch.one/wp-content/uploads/2016/11/erhardt2016coinselection.pdf) (@xekyo). I have it set so this algorithm will run first and if it fails, it will fall back to the current coin selection algorithm. The coin selection algorithms and tests have been refactored to separate files instead of having them all in wallet.cpp.

I have added some tests for the new algorithm and a test for all of coin selection in general. However, more tests may be needed, but I will need help with coming up with more test cases.

This PR uses some code borrowed from #10360 to use effective values when selecting coins.

PR #10712, opened 2017-06-30
Add change output if necessary to reduce excess fee

morcos:
This is an alternative to #10333 

See commit messages.

The first commit is mostly code move, it just moves the change creation code out of the loop.

@instagibbs

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -2569,6 +2569,38 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CWalletT
             std::vector<COutput> vAvailableCoins;
             AvailableCoins(vAvailableCoins, true, coinControl);
 
+            // Create change script that will be used if we need change
+            // TODO: pass in scriptChange instead of reservekey so
+            // change transaction isn't always pay-to-bitcoin-address
+            CScript scriptChange;
+
+            // coin control: send change to custom address
+            if (coinControl && !boost::get<CNoDestination>(&coinControl->destChange))
+                scriptChange = GetScriptForDestination(coinControl->destChange);
+
+            // no coin control: send change to newly generated address
+            else
+            {
+                // Note: We use a new key here to keep it from being obvious which side is the change.
+                //  The drawback is that by not reusing a previous key, the change may be lost if a
+                //  backup is restored, if the backup doesn't have the new private key for the change.
+                //  If we reused the old key, it would be possible to add code to look for and
+                //  rediscover unknown transactions that were written with keys of ours to recover
+                //  post-backup change.
+
+                // Reserve a new key pair from key pool
+                CPubKey vchPubKey;
+                bool ret;
+                ret = reservekey.GetReservedKey(vchPubKey, true);
+                if (!ret)
+                {
+                    strFailReason = _("Keypool ran out, please call keypoolrefill first");
+                    return false;
+                }
+
+                scriptChange = GetScriptForDestination(vchPubKey.GetID());
+            }
+
             nFeeRet = 0;
             // Start with no fee and loop until there is enough fee
             while (true)
@@ -2627,37 +2659,6 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CWalletT
                 if (nChange > 0)
                 {
                     // Fill a vout to ourself
-                    // TODO: pass in scriptChange instead of reservekey so
-                    // change transaction isn't always pay-to-bitcoin-address
-                    CScript scriptChange;
-
-                    // coin control: send change to custom address
-                    if (coinControl && !boost::get<CNoDestination>(&coinControl->destChange))
-                        scriptChange = GetScriptForDestination(coinControl->destChange);
-
-                    // no coin control: send change to newly generated address
-                    else
-                    {
-                        // Note: We use a new key here to keep it from being obvious which side is the change.
-                        //  The drawback is that by not reusing a previous key, the change may be lost if a
-                        //  backup is restored, if the backup doesn't have the new private key for the change.
-                        //  If we reused the old key, it would be possible to add code to look for and
-                        //  rediscover unknown transactions that were written with keys of ours to recover
-                        //  post-backup change.
-
-                        // Reserve a new key pair from key pool
-                        CPubKey vchPubKey;
-                        bool ret;
-                        ret = reservekey.GetReservedKey(vchPubKey, true);
-                        if (!ret)
-                        {
-                            strFailReason = _("Keypool ran out, please call keypoolrefill first");
-                            return false;
-                        }
-
-                        scriptChange = GetScriptForDestination(vchPubKey.GetID());
-                    }
-
                     CTxOut newTxOut(nChange, scriptChange);
 
                     // Never create dust outputs; if we would, just
@@ -2666,7 +2667,6 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CWalletT
                     {
                         nChangePosInOut = -1;
                         nFeeRet += nChange;
-                        reservekey.ReturnKey();
                     }
                     else
                     {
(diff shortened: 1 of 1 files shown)


src/wallet/wallet.cpp
```diff
@@ -2744,8 +2751,18 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CWalletT
...
+                    CAmount max_excess_fee = GetMinimumFee(change_prototype_size, currentConfirmationTarget, ::mempool, ::feeEstimator, nullptr) +
+                        GetDustThreshold(change_prototype_txout, ::dustRelayFee);
+                    // If we have no change and a big enough excess fee, then
+                    // try to construct transaction again only without picking
+                    // new inputs. We now know we only need the smaller fee
+                    // (because of reduced tx size) and so we should add a
+                    // change input. Only try this once.
+                    if (nFeeRet > nFeeNeeded + max_excess_fee && nChangePosInOut == -1 && nSubtractFeeFromAmount == 0 && pick_new_inputs) {
+                        pick_new_inputs = false;
+                        nFeeRet = nFeeNeeded;
```
instagibbs: this needs to be `nFeeRet = nFeeNeeded + GetMinimumFee(change_prototype_size, currentConfirmationTarget, ::mempool, ::feeEstimator, nullptr) ;` at a minimum, right?

nFeeNeeded at this point will not pay for the new change output.

instagibbs (commented):
imo this is superior in its simplicity and targeting the exact situation better over #10333 .

A bit worried about the failure case triggering anytime fee estimations change. I know you don't want to in this PR, but pulling out the estimation and getting a feerate that is reapplied in the main loop seems worth it, is easy to review, and will have to be done anyways for effective value work.

src/wallet/wallet.cpp
morcos: oops, correct

src/wallet/wallet.cpp
```diff
@@ -2774,6 +2773,8 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CWalletT
             }
         }
 
+        if (nChangePosInOut == -1) reserveKey.ReturnKey(); // Return any reserved key if we don't have change
```
morcos: weird, i fixed that but must have not committed before push

src/wallet/wallet.cpp
```diff
@@ -2754,6 +2771,11 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CWalletT
                     }
                     break; // Done, enough fee included.
                 }
+                else if (!pick_new_inputs) {
+                    // This shouldn't happen
```
morcos: yes, i don't think that can happen (a change in fee estimation) but belts and suspenders.
In any case will be cleaned up post 0.15 with effective value logic and only calculating the needed fee rate once.

morcos:
Fixed bug and nits and squashed

@instagibbs I'm fine with doing the GetMinimumFeeRate refactor for 0.15, but I already did a lot of other work on GetMinimumFee in #10706 which I would really like to get in for 0.15, so I'm wary of piling on too many changes that touch the same lines of code and not being able to get them all in.

src/wallet/wallet.cpp
instagibbs: sorry to both you again on this but it appears to still be there

promag:
All return points should call `reservekey.ReturnKey()`?

promag:
Reference #10247.

promag (commented):
utACK.

IMO going for one last iteration and to have one more flag is a clear sign that a more thoughtful refactor should be done. Not that I'm against this change, but reading this code should be plain and easy.

instagibbs:
@promag this is a bugfix with limited scope. Switching to effective value calculation fixes this in a more natural fashion. See: https://github.com/bitcoin/bitcoin/pull/10360 and https://github.com/bitcoin/bitcoin/pull/10637

promag (commented):
Nit, should variables be in camel or snake case? Also could you fix brace styles for the moved code?

morcos:
@promag 
New variables are supposed to be snake_case, did I miss one?
If there are any other changes, I can fix the braces, but I didn't touch the moved code.

As for calling ReturnKey, I suppose that does make sense, but the existing code did not do that, and I haven't materially changed that behavior. I'm happy to add it though if I should?

promag:
@morcos another PR is fine. As @instagibbs said, this is a bugfix.

laanwj:
Replaced #10333 with this one in "High priority for review" (reviewing a closed PR makes little sense).

src/wallet/wallet.cpp
```diff
@@ -2736,16 +2743,30 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CWalletT
...
                     // to be addressed because it requires returning the fee to
                     // the payees and not the change output.
-                    // TODO: The case where there is no change output remains
-                    // to be addressed so we avoid creating too small an output.
+
+                    // If we have no change and a big enough excess fee, then
+                    // try to construct transaction again only without picking
+                    // new inputs. We now know we only need the smaller fee
+                    // (because of reduced tx size) and so we should add a
+                    // change input. Only try this once.
```
ryanofsky: change output

src/wallet/wallet.cpp
```diff
@@ -2744,8 +2751,18 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CWalletT
                     // TODO: The case where nSubtractFeeFromAmount > 0 remains
                     // to be addressed because it requires returning the fee to
                     // the payees and not the change output.
-                    // TODO: The case where there is no change output remains
-                    // to be addressed so we avoid creating too small an output.
+                    CAmount max_excess_fee = GetMinimumFee(change_prototype_size, currentConfirmationTarget, ::mempool, ::feeEstimator, nullptr) +
+                        GetDustThreshold(change_prototype_txout, ::dustRelayFee);
+                    // If we have no change and a big enough excess fee, then
```
ryanofsky: > please move first explanation a few lines before for the "increase change" to appropriate spot

I think order of the comments does makes sense. First comment describes the normal case, second comment describes unhandled special case, and third comment describes handled special case accompanied by implementation for that case.

I might move the code for the normal case above the code & comments for the two special cases, though this would increase the size of the diff. Just a thought, though.

ryanofsky (commented):
Wow, this code is ridiculously complicated. Pretty clean fix, though. Would be nice if we could mock `SelectCoins` and easily write unit tests for these cases.

utACK d64a4b94ea809179e8cc9a92be75036bc8bd4fbf

morcos:
Fixed commit history and fixed comment nit

```
git diff d64a4b9 533b3f0

-                    // change input. Only try this once.
+                    // change output. Only try this once.
```

TheBlueMatt (commented):
utACK 533b3f07f0525939d5fef2d22ec1ee699b7d437b

src/wallet/wallet.cpp
```diff
@@ -2754,6 +2775,12 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CWalletT
                     }
                     break; // Done, enough fee included.
                 }
+                else if (!pick_new_inputs) {
```
sipa: Nit: else on the same line as `}`

src/wallet/wallet.cpp
instagibbs: He moved it, the comment wasn't marked as resolved. Happy with current arrangement.

instagibbs:
utACK https://github.com/bitcoin/bitcoin/pull/10712/commits/533b3f07f0525939d5fef2d22ec1ee699b7d437b

ryanofsky (commented):
utACK 533b3f07f0525939d5fef2d22ec1ee699b7d437b. Changes since last review were what Alex mentioned.

morcos:
This was rebased to add 2 new arguments to GetMinimumFee line which were introduced in #10589 

Will change again as it conflicts with #10706 on that same point

ryanofsky (commented):
utACK 0f402b9263b0579b29aa0f841fc64ad58d3efba6. Only difference is rebase.

sipa:
Posthumous utACK 0f402b9

PR #10715, opened 2017-07-02
scripted-diff: Prefer x.empty() over x.size() == 0 and x.length() == 0

practicalswift:
Changes:
* Prefer `x.empty()` over `x.size() == 0` and `x.length() == 0`
* Prefer `!x.empty()` over `bool(x.size())`
* Prefer `!x.empty()` over `x.size() > 0`

Rationale:
* Improve readability.
* Avoid nongeneric code - see ["T.143: Don't write unintentionally nongeneric code"](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#t143-dont-write-unintentionally-nongeneric-code) in the C++ Core Guidelines (Stroustrup & Sutter): *"Emptiness works for more containers than size(), because some containers don't know their size or are conceptually of unbounded size."*

### DIFF
--- a/src/addrman.cpp
+++ b/src/addrman.cpp
@@ -448,9 +448,9 @@ int CAddrMan::Check_()
         }
     }
 
-    if (setTried.size())
+    if (!setTried.empty())
         return -13;
-    if (mapNew.size())
+    if (!mapNew.empty())
         return -15;
     if (nKey.IsNull())
         return -16;
--- a/src/addrman.h
+++ b/src/addrman.h
@@ -491,6 +491,11 @@ public:
         return vRandom.size();
     }
 
+    bool empty() const
+    {
+        return size() == 0;
+    }
+
     //! Consistency check
     void Check()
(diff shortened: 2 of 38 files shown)


promag:
Should these
```cpp
if (container.size()) { ... }
if (container.size() > 0) { ... }
```
be
```cpp
if (!container.empty()) { ... }
```
?

practicalswift:
@promag Updated! :-)

promag:
Rename PR to something along "Prefer empty over size or length"?

promag:
@practicalswift Why was this closed?

practicalswift:
@promag Two reasons - I had my doubts if this PR is worth doing (might be a bit too picky and quite heavy to review due to the number of change) and also that the PR failed to gain any concept ACKs from core developers :-)   But thanks for reviewing anyways!

promag:
NP.

PR #10729, opened 2017-07-03
Wrap EvalScript in a ScriptExecution class

luke-jr:
When we last discussed making scripts debuggable (sometime after #3901), the plan was to instead trace execution rather than single-step through it.

This is the first step toward that goal. The full implementation can be found on my `script_debugger` branch.

PR #10730, opened 2017-07-03
Move script flag to/from-string logic from tests to script/interpreter

luke-jr:
When we last discussed making scripts debuggable (sometime after #3901), the plan was to instead trace execution rather than single-step through it.

This is a dependency step toward that, and is eventually used for the RPC `verifyscript` to parse its `"flags"` option.

See also #10729


jonasschnelli:
Needs rebase

luke-jr:
Rebased

laanwj:
I'm okay with moving this out of the tests if it's necessary in another place, however the string stuff doesn't belong with the interpreter consensus code IMO. Better to keep it in a separate compilation unit, like with the CBitcoinAddress split done in #11117.

laanwj:
@luke-jr ping?

luke-jr:
Hmm, so where would you suggest? Bundle it in with RPC-specific stuff?

luke-jr:
Rebased, and moved it to script/script_flags.{h,cpp} for now.

luke-jr:
Closing due to apparent lack of interest/reviewers

PR #10733, opened 2017-07-03
getwalletinfo: return path for wallet file

instagibbs:
Makes handling multiwallet a bit easier in case user forgets which wallet is loaded.


jnewbery:
I believe this functionality is covered by #10604

instagibbs:
aw crap thanks

PR #10738, opened 2017-07-04
net: drop custom CNode refcounting in favor of smart pointers

theuni:
This is a more involved version of #10697, which instead completely gets rid of our nasty AddRef() and Release() in favor of automatic lifetime management.

Special care must be taken, though, to only delete CNodes from a single thread, and to control which thread that is. Eventually, this should move to the message processing thread, so that it's not possible to hit cs_main from the main network thread.

In order to do this safely, this PR introduces 2 new generic smart pointer types: strong_ptr and decay_ptr. They provide a functionality that I've wanted for a long time: the ability to safely decay a shared_ptr to a unique_ptr. That sounds somewhat nonsensical at first, but it's useful to be able to make copies of a pointer for a while, stop, wait until only one remains, then delete with guaranteed safety.

Please read shared_ptr.h and check out the tests before groaning. I think this is a very cool (and completely safe) pattern.

This functionality could potentially be accomplished with a shared_ptr and polling ptr.unique(), but that's inherently racy because a new copy could be created simultaneously. Even moving to a local instance and calling .unique() on that one is not safe, as a weak_ptr could be upgraded simultaneously.

Instead, a strong_ptr is created which acts like a unique_ptr but allows shared_ptrs to be "loaned" out. Once a strong_ptr is moved into a decay_ptr, the strong_ptr is reset and no new loans may be created. The decay_ptr tracks the lifetime of the loaned copies, and knows whether or not they have all expired. This can be queried, with no race concerns, with decay_ptr::decay().

Additionally, if the loaned shared_ptrs for some reason outlive the strong_ptr/decay_ptr, they are safely deleted once the last loaned shared_ptr expires. So there is no risk of leaks.

In order to make review easier, these changes were made in a few stages:
1. Where possible, make functions agnostic to the type of pointer being used
2. Switch to shared_ptr but keep existing refcounting on top
3. Switch to strong_ptr
4. Drop existing refcounting and now-unnecessary locking.

### DIFF
--- a/src/net.h
+++ b/src/net.h
@@ -166,8 +166,8 @@ public:
     {
         LOCK(cs_vNodes);
         for (auto&& node : vNodes) {
-            if (NodeFullyConnected(node))
-                func(node);
+            if (NodeFullyConnected(node.get()))
+                func(node.get());
         }
     };
 
@@ -176,8 +176,8 @@ public:
     {
         LOCK(cs_vNodes);
         for (auto&& node : vNodes) {
-            if (NodeFullyConnected(node))
-                func(node);
+            if (NodeFullyConnected(node.get()))
+                func(node.get());
         }
     };
 
@@ -186,8 +186,8 @@ public:
     {
         LOCK(cs_vNodes);
         for (auto&& node : vNodes) {
-            if (NodeFullyConnected(node))
-                pre(node);
+            if (NodeFullyConnected(node.get()))
+                pre(node.get());
         }
         post();
     };
@@ -197,8 +197,8 @@ public:
     {
         LOCK(cs_vNodes);
         for (auto&& node : vNodes) {
-            if (NodeFullyConnected(node))
-                pre(node);
+            if (NodeFullyConnected(node.get()))
+                pre(node.get());
         }
         post();
     };
@@ -303,10 +303,10 @@ private:
     uint64_t CalculateKeyedNetGroup(const CAddress& ad) const;
 
     bool AttemptToEvictConnection();
-    CNode* ConnectNode(CAddress addrConnect, const char *pszDest, bool fCountFailure);
+    std::shared_ptr<CNode> ConnectNode(CAddress addrConnect, const char *pszDest, bool fCountFailure);
     bool IsWhitelistedRange(const CNetAddr &addr);
 
-    void DeleteNode(CNode* pnode);
+    void DeleteNode(std::shared_ptr<CNode>&& pnode);
 
     NodeId GetNewNodeId();
 
@@ -329,11 +329,11 @@ private:
     static bool NodeFullyConnected(const CNode* pnode);
 
     template <typename Callable>
-    CNode* FindNode(Callable&& func)
+    std::shared_ptr<CNode> FindNode(Callable&& func)
     {
         LOCK(cs_vNodes);
         for (auto& node : vNodes) {
-            if(func(node)) {
+            if(func(node.get())) {
                 return node;
             }
         }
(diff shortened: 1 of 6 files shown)


theuni:
Note that https://github.com/bitcoin/bitcoin/pull/10697#issuecomment-311811754 applies here as well.

theuni:
Test failure looks unrelated and matches another one here: https://travis-ci.org/bitcoin/bitcoin/jobs/248773151. Kicking off a new build.

laanwj:
Needs rebase.

src/strong_ptr.h
```diff
@@ -0,0 +1,317 @@
...
+            // Some long-lived operation
+        }
+
+        int main()
+        {
+            strong_ptr<int> str(make_strong<int>(100));
+            std::thread(func, str.get_shared());
+            decay_ptr<int> dec(std::move(strong));
+            // The original strong_ptr is now reset
+            while (!dec.decayed()) {
```
ryanofsky: I don't understand how make_strong/strong_ptr/decay_ptr provide any benefit in this example. Why wouldn't you just use regular shared pointers, and write this loop as `while (str.use_count() > 1)`?

theuni: a weak_ptr can be created from a shared_ptr without bumping its refcount. That weak_ptr can ```lock()``` in a separate thread just after checking use_count() here. ```shared_ptr.unique()``` (```shared_ptr.use_count() == 1```) was deprecated in c++17 for that reason.

Once moved to a decay_ptr, ```decayed()``` is a trustworthy ```unique()```.

laanwj: I don't think this is a particularly good example, as it encourages busy waiting :-) But for the application in our net code it's a clever solution.

src/strong_ptr.h
```diff
@@ -0,0 +1,317 @@
...
+        m_shared.reset(nullptr, shared_deleter{m_data});
+        return *this;
+    }
+
+    void reset()
+    {
+        m_data.reset();
+        m_shared.reset();
+    }
+    void reset(std::nullptr_t)
```
laanwj: What is the advantage of having a specific overload for nullptr_t?

src/net.h
```diff
@@ -333,6 +328,19 @@ class CConnman
     // Whether the node should be passed out in ForEach* callbacks
     static bool NodeFullyConnected(const CNode* pnode);
 
+    template <typename Callable>
```
laanwj: Nice! I like how this cleans up the searching functions. Maybe add a comment (for doxygen) that this returns the first (arbitrary) node that matches a certain predicate, or `nullptr` otherwise.

src/net.h
```diff
@@ -329,6 +329,19 @@ class CConnman
     // Whether the node should be passed out in ForEach* callbacks
     static bool NodeFullyConnected(const CNode* pnode);
 
+    std::vector<std::shared_ptr<CNode>> GetNodesCopy() const
```
laanwj: Creating a copy - increasing all refcounts in the process, just to drop them again at the end of the function - feels like a lot of overhead, how many lock operations does that take internally? Can we somehow benchmark that this is more efficient than holding cs_vnodes for the entire time? (or at least, not much slower, not all the cases are performance critical and it does clean up the code)

laanwj (commented):
utACK.
I was able to review everything except some of the C++11 magic in strong.h (commit 2d720aaa626d3866ee3664f1ce9c3d654811e418), but the tests for that look convincing to me. In my opinion this is a big improvement from how things are currently handled.

src/net.h
theuni: Performance isn't the intention, this was done in order to avoid keeping cs_vNodes locked during the ForEachNode callbacks. Though I agree and also really dislike the overhead.

src/strong_ptr.h
theuni: Heh, I have no idea why I added that. Will remove.

jtimon:
The title sounds great because is not "reinventing the wheel" like our own ref_count data structure.
But strong_pointer feels like reinventing the wheel again...

PR #10740, opened 2017-07-04
[wallet] `loadwallet` RPC - load wallet at runtime

jnewbery:
Adds a `loadwallet` RPCs. This allows wallets to be loaded dynamically during runtime without having to stop-start the node with new `-wallet` params.

Includes functional tests and release notes.

Limitations:

- currently this functionality is only available through the RPC interface.
- wallets loaded in this way will not be displayed in the GUI.

### DIFF
--- a/doc/release-notes-pr10740.md
+++ b/doc/release-notes-pr10740.md
@@ -1,0 +1,8 @@
+Dynamic loading of wallets
+--------------------------
+
+Previously, wallets could only be loaded at startup, by specifying `-wallet` parameters on the command line or in the bitcoin.conf file. It is now possible to load wallets dynamically at runtime by calling the `loadwallet` RPC.
+
+The wallet can be specified as file/directory basename (which must be located in the `walletdir` directory), or as an absolute path to a file/directory.
+
+This feature is currently only available through the RPC interface. Wallets loaded in this way will not display in the bitcoin-qt GUI.
--- a/src/wallet/init.cpp
+++ b/src/wallet/init.cpp
@@ -6,6 +6,7 @@
 #include <chainparams.h>
 #include <init.h>
 #include <net.h>
+#include <scheduler.h>
 #include <util.h>
 #include <utilmoneystr.h>
 #include <validation.h>
@@ -264,8 +265,11 @@ bool WalletInit::Open() const
 void WalletInit::Start(CScheduler& scheduler) const
 {
     for (CWallet* pwallet : GetWallets()) {
-        pwallet->postInitProcess(scheduler);
+        pwallet->postInitProcess();
     }
+
+    // Run a thread to flush wallet periodically
+    scheduler.scheduleEvery(MaybeCompactWalletDB, 500);
 }
 
 void WalletInit::Flush() const
--- a/src/wallet/init.cpp
+++ b/src/wallet/init.cpp
@@ -207,7 +207,12 @@ bool WalletInit::Verify() const
             return InitError(strprintf(_("Error loading wallet %s. Duplicate -wallet filename specified."), wallet_file));
         }
 
-        if (!CWallet::Verify(wallet_file, salvage_wallet)) return false;
+        std::string error_string;
+        std::string warning_string;
+        bool verify_success = CWallet::Verify(wallet_file, salvage_wallet, error_string, warning_string);
+        if (!error_string.empty()) InitError(error_string);
+        if (!warning_string.empty()) InitWarning(warning_string);
+        if (!verify_success) return false;
     }
 
     return true;
--- a/src/wallet/init.cpp
+++ b/src/wallet/init.cpp
@@ -190,55 +190,24 @@ bool WalletInit::Verify() const
 
     uiInterface.InitMessage(_("Verifying wallet(s)..."));
 
+    std::vector<std::string> wallet_files = gArgs.GetArgs("-wallet");
+
+    // Parameter interaction code should have thrown an error if -salvagewallet
+    // was enabled with more than wallet file, so the wallet_files size check
+    // here should have no effect.
+    bool salvage_wallet = gArgs.GetBoolArg("-salvagewallet", false) && wallet_files.size() <= 1;
+
     // Keep track of each wallet absolute path to detect duplicates.
     std::set<fs::path> wallet_paths;
 
-    for (const std::string& walletFile : gArgs.GetArgs("-wallet")) {
-        // Do some checking on wallet path. It should be either a:
-        //
-        // 1. Path where a directory can be created.
-        // 2. Path to an existing directory.
-        // 3. Path to a symlink to a directory.
-        // 4. For backwards compatibility, the name of a data file in -walletdir.
-        fs::path wallet_path = fs::absolute(walletFile, GetWalletDir());
-        fs::file_type path_type = fs::symlink_status(wallet_path).type();
-        if (!(path_type == fs::file_not_found || path_type == fs::directory_file ||
-              (path_type == fs::symlink_file && fs::is_directory(wallet_path)) ||
-              (path_type == fs::regular_file && fs::path(walletFile).filename() == walletFile))) {
-            return InitError(strprintf(
-                _("Invalid -wallet path '%s'. -wallet path should point to a directory where wallet.dat and "
-                  "database/log.?????????? files can be stored, a location where such a directory could be created, "
-                  "or (for backwards compatibility) the name of an existing data file in -walletdir (%s)"),
-                walletFile, GetWalletDir()));
-        }
+    for (const auto wallet_file : wallet_files) {
+        fs::path wallet_path = fs::absolute(wallet_file, GetWalletDir());
 
         if (!wallet_paths.insert(wallet_path).second) {
-            return InitError(strprintf(_("Error loading wallet %s. Duplicate -wallet filename specified."), walletFile));
+            return InitError(strprintf(_("Error loading wallet %s. Duplicate -wallet filename specified."), wallet_file));
         }
 
-        std::string strError;
-        if (!WalletBatch::VerifyEnvironment(wallet_path, strError)) {
-            return InitError(strError);
-        }
-
-        if (gArgs.GetBoolArg("-salvagewallet", false)) {
-            // Recover readable keypairs:
-            CWallet dummyWallet("dummy", WalletDatabase::CreateDummy());
-            std::string backup_filename;
-            if (!WalletBatch::Recover(wallet_path, (void *)&dummyWallet, WalletBatch::RecoverKeysOnlyFilter, backup_filename)) {
-                return false;
-            }
-        }
-
-        std::string strWarning;
-        bool dbV = WalletBatch::VerifyDatabaseFile(wallet_path, strWarning, strError);
-        if (!strWarning.empty()) {
-            InitWarning(strWarning);
-        }
-        if (!dbV) {
-            InitError(strError);
-            return false;
-        }
+        if (!CWallet::Verify(wallet_file, salvage_wallet)) return false;
     }
 
     return true;
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -2994,6 +2994,53 @@ static UniValue listwallets(const JSONRPCRequest& request)
     return obj;
 }
 
+UniValue loadwallet(const JSONRPCRequest& request)
+{
+    if (request.fHelp || request.params.size() != 1)
+        throw std::runtime_error(
+            "loadwallet \"filename\"\n"
+            "\nLoads a wallet from a wallet file or directory."
+            "\nNote that all wallet command-line options used when starting bitcoind will be"
+            "\napplied to the new wallet (eg -zapwallettxes, upgradewallet, rescan, etc).\n"
+            "\nArguments:\n"
+            "1. \"filename\"    (string, required) The wallet directory or .dat file.\n"
+            "\nResult:\n"
+            "{\n"
+            "  \"name\" :    <wallet_name>,        (string) The wallet name if loaded successfully.\n"
+            "  \"warning\" : <warning>,            (string) Warning message if wallet was not loaded cleanly.\n"
+            "}\n"
+            "\nExamples:\n"
+            + HelpExampleCli("loadwallet", "\"test.dat\"")
+            + HelpExampleRpc("loadwallet", "\"test.dat\"")
+        );
+    std::string wallet_file = request.params[0].get_str();
+    std::string error;
+
+    fs::path wallet_path = fs::absolute(wallet_file, GetWalletDir());
+    if (fs::symlink_status(wallet_path).type() == fs::file_not_found) {
+        throw JSONRPCError(RPC_WALLET_NOT_FOUND, "Wallet " + wallet_file + " not found.");
+    }
+
+    std::string warning;
+    if (!CWallet::Verify(wallet_file, false, error, warning)) {
+        throw JSONRPCError(RPC_WALLET_ERROR, "Wallet file verification failed: " + error);
+    }
+
+    CWallet * const wallet = CWallet::CreateWalletFromFile(wallet_file, fs::absolute(wallet_file, GetWalletDir()));
+    if (!wallet) {
+        throw JSONRPCError(RPC_WALLET_ERROR, "Wallet loading failed.");
+    }
+    AddWallet(wallet);
+
+    wallet->postInitProcess();
+
+    UniValue obj(UniValue::VOBJ);
+    obj.pushKV("name", wallet->GetName());
+    obj.pushKV("warning", warning);
+
+    return obj;
+}
+
 static UniValue resendwallettransactions(const JSONRPCRequest& request)
 {
     CWallet * const pwallet = GetWalletForJSONRPCRequest(request);
@@ -4197,6 +4244,7 @@ static const CRPCCommand commands[] =
     { "wallet",             "listtransactions",                 &listtransactions,              {"account|dummy","count","skip","include_watchonly"} },
     { "wallet",             "listunspent",                      &listunspent,                   {"minconf","maxconf","addresses","include_unsafe","query_options"} },
     { "wallet",             "listwallets",                      &listwallets,                   {} },
+    { "wallet",             "loadwallet",                       &loadwallet,                    {"filename"} },
     { "wallet",             "lockunspent",                      &lockunspent,                   {"unlock","transactions"} },
     { "wallet",             "sendfrom",                         &sendfrom,                      {"fromaccount","toaddress","amount","minconf","comment","comment_to"} },
     { "wallet",             "sendmany",                         &sendmany,                      {"fromaccount|dummy","amounts","minconf","comment","subtractfeefrom","replaceable","conf_target","estimate_mode"} },
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -27,6 +27,7 @@
 #include <txmempool.h>
 #include <utilmoneystr.h>
 #include <wallet/fees.h>
+#include <wallet/walletutil.h>
 
 #include <algorithm>
 #include <assert.h>
@@ -3987,6 +3988,61 @@ void CWallet::MarkPreSplitKeys()
         set_pre_split_keypool.insert(index);
         it = setExternalKeyPool.erase(it);
     }
+}
+
+bool CWallet::Verify(std::string wallet_file, bool salvage_wallet)
+{
+    // Do some checking on wallet path. It should be either a:
+    //
+    // 1. Path where a directory can be created.
+    // 2. Path to an existing directory.
+    // 3. Path to a symlink to a directory.
+    // 4. For backwards compatibility, the name of a data file in -walletdir.
+    LOCK(cs_wallets);
+    fs::path wallet_path = fs::absolute(wallet_file, GetWalletDir());
+    fs::file_type path_type = fs::symlink_status(wallet_path).type();
+    if (!(path_type == fs::file_not_found || path_type == fs::directory_file ||
+          (path_type == fs::symlink_file && fs::is_directory(wallet_path)) ||
+          (path_type == fs::regular_file && fs::path(wallet_file).filename() == wallet_file))) {
+        return InitError(strprintf(
+            _("Invalid -wallet path '%s'. -wallet path should point to a directory where wallet.dat and "
+              "database/log.?????????? files can be stored, a location where such a directory could be created, "
+              "or (for backwards compatibility) the name of an existing data file in -walletdir (%s)"),
+            wallet_file, GetWalletDir()));
+    }
+
+    // Make sure that the wallet path doesn't clash with an existing wallet path
+    for (auto wallet : GetWallets()) {
+        if (fs::absolute(wallet->GetName(), GetWalletDir()) == wallet_path) {
+            return InitError(strprintf(_("Error loading wallet %s. Duplicate -wallet filename specified."), wallet_file));
+        }
+    }
+
+    std::string strError;
+    if (!WalletBatch::VerifyEnvironment(wallet_path, strError)) {
+        return InitError(strError);
+    }
+
+    if (salvage_wallet) {
+        // Recover readable keypairs:
+        CWallet dummyWallet("dummy", WalletDatabase::CreateDummy());
+        std::string backup_filename;
+        if (!WalletBatch::Recover(wallet_path, (void *)&dummyWallet, WalletBatch::RecoverKeysOnlyFilter, backup_filename)) {
+            return false;
+        }
+    }
+
+    std::string strWarning;
+    bool dbV = WalletBatch::VerifyDatabaseFile(wallet_path, strWarning, strError);
+    if (!strWarning.empty()) {
+        InitWarning(strWarning);
+    }
+    if (!dbV) {
+        InitError(strError);
+        return false;
+    }
+
+    return true;
 }
 
 CWallet* CWallet::CreateWalletFromFile(const std::string& name, const fs::path& path)
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -23,7 +23,6 @@
 #include <primitives/block.h>
 #include <primitives/transaction.h>
 #include <script/script.h>
-#include <scheduler.h>
 #include <timedata.h>
 #include <txmempool.h>
 #include <utilmoneystr.h>
@@ -4308,18 +4307,11 @@ CWallet* CWallet::CreateWalletFromFile(const std::string& name, const fs::path& 
     return walletInstance;
 }
 
-std::atomic<bool> CWallet::fFlushScheduled(false);
-
-void CWallet::postInitProcess(CScheduler& scheduler)
+void CWallet::postInitProcess()
 {
     // Add wallet transactions that aren't already in a block to mempool
     // Do this here as mempool requires genesis block to be loaded
     ReacceptWalletTransactions();
-
-    // Run a thread to flush wallet periodically
-    if (!CWallet::fFlushScheduled.exchange(true)) {
-        scheduler.scheduleEvery(MaybeCompactWalletDB, 500);
-    }
 }
 
 bool CWallet::BackupWallet(const std::string& strDest)
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -3990,7 +3990,7 @@ void CWallet::MarkPreSplitKeys()
     }
 }
 
-bool CWallet::Verify(std::string wallet_file, bool salvage_wallet)
+bool CWallet::Verify(std::string wallet_file, bool salvage_wallet, std::string& error_string, std::string& warning_string)
 {
     // Do some checking on wallet path. It should be either a:
     //
@@ -4004,23 +4004,24 @@ bool CWallet::Verify(std::string wallet_file, bool salvage_wallet)
     if (!(path_type == fs::file_not_found || path_type == fs::directory_file ||
           (path_type == fs::symlink_file && fs::is_directory(wallet_path)) ||
           (path_type == fs::regular_file && fs::path(wallet_file).filename() == wallet_file))) {
-        return InitError(strprintf(
-            _("Invalid -wallet path '%s'. -wallet path should point to a directory where wallet.dat and "
+        error_string = strprintf(
+              "Invalid -wallet path '%s'. -wallet path should point to a directory where wallet.dat and "
               "database/log.?????????? files can be stored, a location where such a directory could be created, "
-              "or (for backwards compatibility) the name of an existing data file in -walletdir (%s)"),
-            wallet_file, GetWalletDir()));
+              "or (for backwards compatibility) the name of an existing data file in -walletdir (%s)",
+              wallet_file, GetWalletDir());
+        return false;
     }
 
     // Make sure that the wallet path doesn't clash with an existing wallet path
     for (auto wallet : GetWallets()) {
         if (fs::absolute(wallet->GetName(), GetWalletDir()) == wallet_path) {
-            return InitError(strprintf(_("Error loading wallet %s. Duplicate -wallet filename specified."), wallet_file));
+            error_string = strprintf("Error loading wallet %s. Duplicate -wallet filename specified.", wallet_file);
+            return false;
         }
     }
 
-    std::string strError;
-    if (!WalletBatch::VerifyEnvironment(wallet_path, strError)) {
-        return InitError(strError);
+    if (!WalletBatch::VerifyEnvironment(wallet_path, error_string)) {
+        return false;
     }
 
     if (salvage_wallet) {
@@ -4032,17 +4033,7 @@ bool CWallet::Verify(std::string wallet_file, bool salvage_wallet)
         }
     }
 
-    std::string strWarning;
-    bool dbV = WalletBatch::VerifyDatabaseFile(wallet_path, strWarning, strError);
-    if (!strWarning.empty()) {
-        InitWarning(strWarning);
-    }
-    if (!dbV) {
-        InitError(strError);
-        return false;
-    }
-
-    return true;
+    return WalletBatch::VerifyDatabaseFile(wallet_path, warning_string, error_string);
 }
 
 CWallet* CWallet::CreateWalletFromFile(const std::string& name, const fs::path& path)
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -4012,7 +4012,10 @@ CWallet* CWallet::CreateWalletFromFile(const std::string& name, const fs::path& 
 
     int64_t nStart = GetTimeMillis();
     bool fFirstRun = true;
-    CWallet *walletInstance = new CWallet(name, WalletDatabase::Create(path));
+    // Make a temporary wallet unique pointer so memory doesn't get leaked if
+    // wallet creation fails.
+    auto temp_wallet = MakeUnique<CWallet>(name, WalletDatabase::Create(path));
+    CWallet* walletInstance = temp_wallet.get();
     DBErrors nLoadWalletRet = walletInstance->LoadWallet(fFirstRun);
     if (nLoadWalletRet != DBErrors::LOAD_OK)
     {
@@ -4224,7 +4227,6 @@ CWallet* CWallet::CreateWalletFromFile(const std::string& name, const fs::path& 
     }
 
     walletInstance->m_last_block_processed = chainActive.Tip();
-    RegisterValidationInterface(walletInstance);
 
     if (chainActive.Tip() && chainActive.Tip() != pindexRescan)
     {
@@ -4290,6 +4292,10 @@ CWallet* CWallet::CreateWalletFromFile(const std::string& name, const fs::path& 
             }
         }
     }
+
+    // Register with the validation interface. It's ok to do this after rescan since we're still holding cs_main.
+    RegisterValidationInterface(temp_wallet.release());
+
     walletInstance->SetBroadcastTransactions(gArgs.GetBoolArg("-walletbroadcast", DEFAULT_WALLETBROADCAST));
 
     {
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -1119,7 +1119,7 @@ public:
     bool MarkReplaced(const uint256& originalHash, const uint256& newHash);
 
     //! Verify wallet naming and perform salvage on the wallet if required
-    static bool Verify(std::string wallet_file, bool salvage_wallet);
+    static bool Verify(std::string wallet_file, bool salvage_wallet, std::string& error_string, std::string& warning_string);
 
     /* Initializes the wallet, returns a new CWallet instance or a null pointer in case of an error */
     static CWallet* CreateWalletFromFile(const std::string& name, const fs::path& path);
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -68,7 +68,6 @@ class CCoinControl;
 class COutput;
 class CReserveKey;
 class CScript;
-class CScheduler;
 class CTxMemPool;
 class CBlockPolicyEstimator;
 class CWalletTx;
@@ -675,7 +674,6 @@ class WalletRescanReserver; //forward declarations for ScanForWalletTransactions
 class CWallet final : public CCryptoKeyStore, public CValidationInterface
 {
 private:
-    static std::atomic<bool> fFlushScheduled;
     std::atomic<bool> fAbortRescan{false};
     std::atomic<bool> fScanningWallet{false}; // controlled by WalletRescanReserver
     std::mutex mutexScanning;
@@ -1127,7 +1125,7 @@ public:
      * Wallet post-init setup
      * Gives the wallet a chance to register repetitive tasks and complete post-init tasks
      */
-    void postInitProcess(CScheduler& scheduler);
+    void postInitProcess();
 
     bool BackupWallet(const std::string& strDest);
 
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -1118,6 +1118,9 @@ public:
     /** Mark a transaction as replaced by another transaction (e.g., BIP 125). */
     bool MarkReplaced(const uint256& originalHash, const uint256& newHash);
 
+    //! Verify wallet naming and perform salvage on the wallet if required
+    static bool Verify(std::string wallet_file, bool salvage_wallet);
+
     /* Initializes the wallet, returns a new CWallet instance or a null pointer in case of an error */
     static CWallet* CreateWalletFromFile(const std::string& name, const fs::path& path);
 
--- a/test/functional/wallet_multiwallet.py
+++ b/test/functional/wallet_multiwallet.py
@@ -170,5 +170,46 @@ class MultiWalletTest(BitcoinTestFramework):
         assert_equal(w1.getwalletinfo()['paytxfee'], 0)
         assert_equal(w2.getwalletinfo()['paytxfee'], 4.0)
 
+        self.log.info("Test dynamic wallet loading")
+
+        self.restart_node(0, ['-nowallet'])
+        assert_equal(node.listwallets(), [])
+        assert_raises_rpc_error(-32601, "Method not found", node.getwalletinfo)
+
+        self.log.info("Load first wallet")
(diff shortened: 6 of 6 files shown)


src/init.cpp
```diff
@@ -252,10 +245,7 @@ void Shutdown()
 #endif
     UnregisterAllValidationInterfaces();
 #ifdef ENABLE_WALLET
-    for (CWalletRef pwallet : vpwallets) {
-        delete pwallet;
-    }
-    vpwallets.clear();
+    DeleteWallets();
```
jonasschnelli: It'd probably call this `deallocWallets()`

src/wallet/db.cpp
```diff
@@ -227,40 +227,12 @@ bool CDB::Recover(const std::string& filename, void *callbackDataIn, bool (*reco
 
 bool CDB::VerifyEnvironment(const std::string& walletFile, const fs::path& dataDir, std::string& errorStr)
 {
-    LogPrintf("Using BerkeleyDB version %s\n", DbEnv::version(0, 0, 0));
```
jonasschnelli: What's the reasons for keeping this function (that now always returns `true`)?

src/wallet/wallet.cpp
```diff
@@ -35,6 +35,8 @@
 #include <boost/algorithm/string/replace.hpp>
 #include <boost/thread.hpp>
 
+/** Protects the vpwallets vector */
+CCriticalSection cs_wallets;
 std::vector<CWalletRef> vpwallets;
```
jonasschnelli: I guess this belongs to `walletinit.h` now?

src/init.cpp
jnewbery: Yes, `DeleteWallets()` is a bad name. `DeallocWallets()` is good, or perhaps `DetachWallets()`

src/wallet/db.cpp
jnewbery: You're right. This (and `CWalletDB::VerifyEnvironment()`) are now entirely vestigial and are not called. Both can be removed.

src/wallet/wallet.cpp
jnewbery: Why? `vpwallets` is accessed by functions in several of the wallet source files.

jnewbery:
This overlaps significantly with #10762. Not sure how I should ask for review. Perhaps find out which one people think is the priority? Or I could split the `walletinit.cpp` commits into their own PR.

src/wallet/rpcwallet.cpp
```diff
@@ -2432,6 +2434,133 @@ UniValue getwalletinfo(const JSONRPCRequest& request)
...
+
+        LOCK(pwallet->cs_wallet);
+
+        obj.push_back(pwallet->GetName());
+    }
+
+    return obj;
+}
+
+UniValue loadwallet(const JSONRPCRequest& request)
```
jnewbery: On further thought, I think I prefer the names `attachwallet` and `detachwallet`. That gives a bit more distinction from the dump/import RPCs and makes a stronger implication that the new `.dat` file is being loaded and attached as a separate wallet.

promag:
@jnewbery IMO this one should rebase on the other.

jnewbery:
> IMO this one should rebase on the other.

Makes logical sense. Downside is that the other is a more significant change, so may take longer to get merged.

promag:
If this really depends on that work then there's no other way than waiting. The other downside is that the other commits will show up here too.

jnewbery:
I've changed the PR sequence, so this will now depend on #10767. If I get positive concept feedbac on #10767 I'll rebase this on top of it.

For now, I'm still just looking for concept feedback.

src/wallet/wallet.cpp
jonasschnelli: I had the idea that `walletinit.h/.c` is kind of the multiwallet interface (wallet manager). The functions we have there (`DeleteWallets()`, `VerifyWallets()`, `ShutdownWallets();`, etc.) sounded for me after `vpwallets` belongs there... 

IMO, the multiwallet map should ideally not sit within the objects (`CWallet`) implementation.

jnewbery:
Rebased on #10767 . Reviewers, please review #10767 first.

jnewbery:
Rebased now that #10767 is merged. Still a work in progress. PR text updated.

jnewbery:
Bad automatic merge of the testcase. Re-merged manually.

jnewbery:
Manual rebase fixed the test bug. Travis now passes.

This is ready for initial review.

src/qt/bitcoin.cpp
```diff
@@ -484,16 +484,19 @@ void BitcoinApplication::initializeResult(bool success)
...
-            walletModel = new WalletModel(platformStyle, vpwallets[0], optionsModel);
-
-            window->addWallet(BitcoinGUI::DEFAULT_WALLET, walletModel);
-            window->setCurrentWallet(BitcoinGUI::DEFAULT_WALLET);
-
-            connect(walletModel, SIGNAL(coinsSent(CWallet*,SendCoinsRecipient,QByteArray)),
-                             paymentServer, SLOT(fetchPaymentACK(CWallet*,const SendCoinsRecipient&,QByteArray)));
+            // TODO: Expose secondary wallets
+            LOCK(cs_wallets);
+            if (!vpwallets.empty())
```
promag: `{`.

src/qt/bitcoin.cpp
```diff
@@ -484,16 +484,19 @@ void BitcoinApplication::initializeResult(bool success)
...
+            {
+                walletModel = new WalletModel(platformStyle, vpwallets[0], optionsModel);
+
+                window->addWallet(BitcoinGUI::DEFAULT_WALLET, walletModel);
+                window->setCurrentWallet(BitcoinGUI::DEFAULT_WALLET);
+
+                connect(walletModel, SIGNAL(coinsSent(CWallet*,SendCoinsRecipient,QByteArray)),
```
promag: Missing spaces after `,`?

src/wallet/init.cpp
```diff
@@ -185,51 +185,23 @@ bool VerifyWallets()
...
-    // Keep track of each wallet absolute path to detect duplicates.
-    std::set<fs::path> wallet_paths;
+    std::vector<std::string> wallet_files = gArgs.GetArgs("-wallet");
 
-    for (const std::string& walletFile : gArgs.GetArgs("-wallet")) {
-        if (boost::filesystem::path(walletFile).filename() != walletFile) {
-            return InitError(strprintf(_("Error loading wallet %s. -wallet parameter must only specify a filename (not a path)."), walletFile));
-        }
+    // -salvagewallet can only be used when loading a single wallet
+    bool salvage_wallet = gArgs.GetBoolArg("-salvagewallet", false) && wallet_files.size() <= 1;
```
promag: Throw an error if `-salvagewallet` an there are multiple wallets?

src/wallet/rpcwallet.cpp
```diff
@@ -37,6 +37,7 @@ static const std::string WALLET_ENDPOINT_BASE = "/wallet/";
 
 CWallet *GetWalletForJSONRPCRequest(const JSONRPCRequest& request)
 {
+    AssertLockHeld(cs_wallets);
```
promag: Why not just `LOCK(cs_wallets)` too?

promag (commented):
Quick review. All touched code could be updated as per dev notes.

Most importantly, With the `LOCK(cs_wallets)` it will be impossible to have concurrent RPC to different wallets, not to mention to the same wallet. Is this expected behavior? - I'm assuming there is a thread pool for RPC.

jnewbery:
Thanks for the review @promag!

Regarding the locking: Yes, I've implemented this as a mutex for now, which as you point out would limit wallet RPCs to be single-threaded. I'd like to update this to be something a bit better before v0.16. Suggestions are either a shared mutex or (suggested by @theuni) a version of the decay pointers in #10738. The problem with using a shared mutex is that C++ doesn't have a standard library shared mutex until C++14 (std::shared_timed_mutex) so I'd either have to use boost::shared_mutex or roll my own.

(apologies - I thought I'd already written a note in this PR explaining this, but I must have failed to hit send!)

For now, I think review of this PR with the mutex is helpful, and I'll change it around to use a shared_mutex later.

src/wallet/rpcwallet.cpp
jnewbery: Because I'd prefer to avoid recursive locking if possible.

src/wallet/init.cpp
jnewbery: Sounds like a sensible change in behaviour. I think it's beyond the scope of this PR.

luke-jr:
This seems to ignore the whole issue of long-term wallet references, and probably locks `cs_wallets` too long.

src/wallet/rpcwallet.cpp
promag: Just curious, why?

laanwj:
Adding 0.16.0 milestone, I think this should be merged after 0.15.1 to prevent extra rebasing in the wallet code when porting back to 0.15.1.

src/wallet/rpcwallet.cpp
```diff
@@ -2694,6 +2694,54 @@ UniValue openwallet(const JSONRPCRequest& request)
...
+    bool unloaded = false;
+
+    for (std::vector<CWalletRef>::iterator wallet_iter = vpwallets.begin(); wallet_iter != vpwallets.end(); wallet_iter++) {
+
+        if (!EnsureWalletIsAvailable(*wallet_iter, request.fHelp)) {
+            throw JSONRPCError(RPC_WALLET_ERROR, "Error accessing wallets");
+        }
+
+        if ((*wallet_iter)->GetName() == walletName) {
+
```
kallewoof: Unneeded newline.

src/wallet/rpcwallet.cpp
```diff
@@ -2694,6 +2694,54 @@ UniValue openwallet(const JSONRPCRequest& request)
...
+            + HelpExampleRpc("closewallet", "\"test.dat\"")
+        );
+
+    LOCK(cs_wallets);
+
+    std::string walletName = request.params[0].get_str();
+
+    bool unloaded = false;
+
+    for (std::vector<CWalletRef>::iterator wallet_iter = vpwallets.begin(); wallet_iter != vpwallets.end(); wallet_iter++) {
```
kallewoof: Any reason for not just doing `for (auto wallet : vpwallets) {` here?

src/wallet/rpcwallet.cpp
```diff
@@ -2658,35 +2658,53 @@ UniValue listwallets(const JSONRPCRequest& request)
 
 UniValue openwallet(const JSONRPCRequest& request)
 {
-    if (request.fHelp || request.params.size() != 1)
+    if (request.fHelp || request.params.size() > 2)
         throw std::runtime_error(
-            "openwallet \"filename\"\n"
+            "openwallet \"filename\" (\"passphrase\")\n"
```
kallewoof: Ugh, passwords on prompts leaves the password as plaintext in shell history etc. Can't really be fixed for HTTP-RPC I guess, but CLI could *probably* take a bool to mean "prompt me for password".

src/wallet/rpcwallet.cpp
```diff
@@ -2658,35 +2658,53 @@ UniValue listwallets(const JSONRPCRequest& request)
...
             "\nOpens a wallet from a wallet.dat file.\n"
+            "\nUse the passphrase argument to unlock an encrypted wallet while opening.\n"
             "\nArguments:\n"
             "1. \"filename\"    (string, required) The wallet.dat file\n"
+            "2. \"passphrase\"    (string, optional) The wallet.dat file\n"
```
kallewoof: Description is wrong.

kallewoof (commented):
utACK 6c5b84119d1ffa547bf094b28ca602bdaf07c21a

promag:
Needs rebase.

The `openwallet` should rescan?

Sjors:
Concept ACK.

I was able to use `openwallet` and `closewallet`.

I wonder how this interacts with #11383. E.g. I'm guessing the UI wouldn't pick up on this (not a big issue). I can try once both are rebased.

It would be nice if `bitcoin-cli -rpcwallet=X` would open `X` if it's not open yet, maybe close it when done (future PR). 

> Should there be a separate RPC for creating a new wallet or should that be done using openwallet? (my initial thoughts are that there should be a separate RPC)

I would prefer it if `openwallet` fails for wallets that don't exist, rather than creating a new wallet. I understand it makes sense for `bitcoind -wallet` to create a new one, because that's what a new user would expect, but I don't think that's needed for `openwallet`.

promag:
@Sjors this PR must be rebased with #11383 (probably when that gets merged).

src/wallet/rpcwallet.cpp
TheBlueMatt: Yes, definitely. cs_wallets is locked *before* cs_main in global lock order, so putting it too many places is just asking for trouble, IMO.

src/wallet/wallet.cpp
TheBlueMatt: Tend to agree with @jonasschnelli here, though I agree the barrier isn't exactly clear (yet). Ideally we'd move more towards clarifying the interface here - making CWallet less and less aware of all the things around it (including other wallets).

ryanofsky (commented):
Maybe mark work in progress due to unresolved locking issues.

It's not clear to me that exclusively locking `cs_wallets` is really that much worse than current behavior, since most current RPCs seem to acquire locks for a long time and effectively be single threaded. But it would be good to avoid this.

It might be straightforward to fix by making vpwallets hold shared pointers, and by releasing the cs_wallets lock except when accessing vpwallets. I don't think there is any need for strong_ptrs because there should be no reason to care what threads wallets are closed on.

Using a shared lock as a substitute for shared pointers seems a little kludgy to me since it would just be speeding up overbroad cs_wallets locks instead of removing them, and would it would still block operations on unrelated wallets when any wallet was being opened or closed.

Verify and Flush refactoring commits in this PR seem like they might be useful changes on their own, and could maybe be pulled out into separate PRs to simplify this one.

wtogami:
Needs rebase

> Should there be a separate RPC for creating a new wallet or should that be done using openwallet? (my initial thoughts are that there should be a separate RPC)

I suggest there should be a new RPC command to create new wallets. Aside from least surprise, there is more than one type of wallet that the user could want created.

jonasschnelli:
Needs rebase and issues (@ryanofsky) addressed.

achow101:
Closing the wallet that is in use by qt (the one at `vpwallets[0]` will result in a crash.

jnewbery:
I plan to pick this up again in the next few days. it requires:

- rebasing on top of all the recent wallet refactors
- addressing the locking issues (potentially pulling in #11402)
- considering how this interacts with multiwallet in the GUI #12610
- potentially pulling out the Verify and Flush refactoring commits into their own PR (https://github.com/bitcoin/bitcoin/pull/10740#pullrequestreview-84849103)

Sjors:
> considering how this interacts with multiwallet in the GUI

A good enough first step would be if the dropdowns update when wallets are loaded / unloaded through the console. Adding UI to load and unload wallets can be a followup PR.

jnewbery:
This was discussed in the IRC meeting today: https://botbot.me/freenode/bitcoin-core-dev/2018-04-12/?msg=98919702&page=5

```
15:26 < wumpus> #topic dynamic wallet load/unload (promag)
15:26 < Randolf> Ack.
15:26 < promag> not sure what you guys think
15:26 < instagibbs_> what's the controversy in this topic :)
15:27 < jonasschnelli> #10740
15:27 < sipa> it should happen, duh
15:27 < gribble> https://github.com/bitcoin/bitcoin/issues/10740 | [WIP] [wallet] dynamic loading/unloading of wallets by jnewbery · Pull Request #10740 · bitcoin/bitcoin · GitHub
15:27 < wumpus> seems like something we want to have at some point...
15:27 < promag> but I think luke agrees that wallet management should be with shared pointers
15:27 < sipa> how and when is another :)
15:27 < luke-jr> indeed, using names just asks for chaos with runtime-changing wallets
15:28 < promag> please read #11402 for some reasons to switch
15:28 < gribble> https://github.com/bitcoin/bitcoin/issues/11402 | Use shared pointer for wallet instances by promag · Pull Request #11402 · bitcoin/bitcoin · GitHub
15:28 < jonasschnelli> IMO 10740 can't create wallets, IMO first step would be to add a createwallet RPC call
15:29 < jonasschnelli> the whole creation/configuration setup if flawed since multiwallet
15:29 < jonasschnelli> stuff like -keypool should be per wallet
15:29 < jnewbery> jonasschnelli: you think createwallet should go in *before* load/unload?
15:29 < jonasschnelli> and persisted in the wallet file (as configuration section)
15:29 < jonasschnelli> jnewbery: not sure,.. just thinking
15:30 < jnewbery> seems reasonable to me
15:30 < jonasschnelli> createwallet could also *not* load the wallet in the first step (not ideal, but maybe reduces complexity)
15:30 < sipa> that seems strange... you could create a new wallet at run time but not use it?
15:30 < jonasschnelli> sipa: createwallet could also directly load/use the wallet
15:30 < jnewbery> I think createwallet would also load the new wallet, no?
15:30 < promag> create implies loading
15:30 < luke-jr> sipa: iow, 0 to 1 only.
15:31 < sipa> jonasschnelli: well then you need to have loading functionality first!
15:31 < sipa> and if you have it, why not expose it
15:31 < jonasschnelli> sipa: yes. That's a point.
15:31 < jnewbery> createwallet could also be done by bitcoin-wallet-tool
15:32 < jnewbery> (#8745)
15:32 < gribble> https://github.com/bitcoin/bitcoin/issues/8745 | [PoC] Add wallet inspection and modification tool "bitcoin-wallet-tool" by jonasschnelli · Pull Request #8745 · bitcoin/bitcoin · GitHub
15:32 < jonasschnelli> Yes. Would be possible...
15:32 < jonasschnelli> I just think the create-during-startup approach is not good
15:33 < promag> also related #10973
15:33 < jnewbery> jonasschnelli: I agree
15:33 < gribble> https://github.com/bitcoin/bitcoin/issues/10973 | Refactor: separate wallet from node by ryanofsky · Pull Request #10973 · bitcoin/bitcoin · GitHub
15:33 < sipa> jonasschnelli: agree
15:33 < jonasschnelli> And as a first step I though createwallet would make sense.. but not loading it seems after a strange use-case
15:33 < luke-jr> load -> create -> unload
15:33 < jonasschnelli> but a nice first code/impl. step
15:33 < luke-jr> unload is the complex part tbh
15:33 < jnewbery> luke-jr: agree
15:34 < jonasschnelli> Agree with luke-jr. Maybe split unload away from the existing PR jnewbery ?
15:34 < jnewbery> yes
15:34 < jnewbery> I intend to pick up 10740 again soon, rebase and rework it
15:34 < promag> consider the use case: 1) rpc rescan wallet 2) in parallel unload wallet - should 2) wait for 1) ?
15:34 < luke-jr> probably
15:35 < jonasschnelli> Great. Dynamic loading/creating is a nice feature that we probably want for 0.17!
15:35 < promag> luke-jr: and if the unload is from the UI?
15:35 < jnewbery> promag: do you consider 11402 a prereq for load/unload? What about just load?
15:36 < jonasschnelli> the wallet-tool is IMO orthogonal to wallet creation
15:36 < jonasschnelli> *via RPC
15:36 < luke-jr> promag: probably the same
15:36 < promag> jnewbery: IMHO for both
15:36 < jonasschnelli> RPC seems to be a must, wallet-tool can be a better place to create some sorts of wallets (or inspect it), .. like encrypted wallets
15:36 < luke-jr> promag: at least initially
15:37 < jnewbery> promag: want to rebase and put on high priority for review then, if you consider it a blocker?
15:37 < promag> luke-jr: my point is that it should not block, you request the unload and go on, when the wallet is not used anymore it gets unloaded
15:38 < jonasschnelli> #11402
15:38 < gribble> https://github.com/bitcoin/bitcoin/issues/11402 | Use shared pointer for wallet instances by promag · Pull Request #11402 · bitcoin/bitcoin · GitHub
15:38 < luke-jr> promag: you mean leave the wallet loaded, but invisible? that seems worst outcome IMO
15:38 < luke-jr> user may unload and just shut off the PC
15:38 < wumpus> the unload should probably be in two stages: after requesting it, RPC and the GUI lose access to it. Then it waits for current operations tofinish. Then the thing really gets delted.
15:38 < luke-jr> yes
15:38 < promag> luke-jr: then the application will wait for wallets to unload
15:38 < jnewbery> I don't think we need to worry about unload at this stage. First step is add load functionality, then createwallet functionality
15:38 < luke-jr> and make it visible to the user in the meantime
15:39 < luke-jr> jnewbery: +1
15:39 < promag> wumpus: right, hence shared pointers
```

I should reduce the scope of this PR to just a `loadwallet` RPC. A `createwallet` RPC should come next, followed by `unloadwallet`. (`unloadwallet` is where most of the difficulties are).

jnewbery:
Updated PR text since this PR is now only for loading wallets (not creating or unloading).

Original text also had the following motivation:

> Main motivation for this was the fact that several wallet parameters are actions for individual wallet load/creation, rather than properties of the wallet component. Examples are `-salvagewallet`, `-rescan`, `-usehd` and `-upgradewallet`. Continuing with that config/loading model is difficult in a multi-wallet world - how can users run those actions on individual wallets when loading multiple wallets? This PR offers a solution: individual wallets can be loaded at run-time, and RPC parameters can be passed in to carry out the various wallet-loading actions. Note that none of those load actions are yet implemented in this PR, but can easily be added.

jnewbery:
Overhauled this PR to just contain the `loadwallet` functionality.

It is built on @promag's PR #13017. Please review that first.

This also needs the `cs_wallets` locking functionality from #11402 to be thread-safe.

I'll rebase once those PRs are merged. I've pushed the branch now in case anyone wants to do early review.

promag:
@jnewbery Nice. BTW I was thinking creating a different PR to just make #13017 thread safe.

jnewbery:
> I was thinking creating a different PR to just make #13017 thread safe.

Great. I'll keep an eye out and rebase on top of that when it's ready.

src/wallet/rpcwallet.cpp
```diff
@@ -2874,6 +2874,49 @@ UniValue listwallets(const JSONRPCRequest& request)
...
+        throw JSONRPCError(RPC_WALLET_ERROR, "Wallet file verification failed: " + error);
+    }
+
+    CWallet * const wallet = CWallet::CreateWalletFromFile(wallet_file, fs::absolute(wallet_file, GetWalletDir()));
+    if (!wallet) {
+        throw JSONRPCError(RPC_WALLET_ERROR, "Wallet creation failed.");
+    }
+    AddWallet(wallet);
+
+    wallet->ReacceptWalletTransactions();
```
promag: Should call `CWallet::postInitProcess` instead? Otherwise `MaybeCompactWalletDB` is not called if the daemon is launched without wallets.

jnewbery:
Thanks for the review @promag - I've done as you suggested and called `CWallet::postInitProcess()`. Because rpc doesn't have the CScheduler reference when loading the wallet, I had to change the `WalletInit::Start()` function to always schedule background wallet flushes, even when started with no wallets (ie `-nowallet`). Note that when started with `-disablewallet`, the entire wallet module is disabled and it's not possible to load wallets later.

I've also rebased on [#13028 rebased on master].

Removed the **[WIP]** tag since I think this implementation is now complete and ready for review.

src/wallet/init.cpp
```diff
@@ -324,7 +325,12 @@ bool WalletInit::Open() const
 void WalletInit::Start(CScheduler& scheduler) const
 {
     for (CWallet* pwallet : GetWallets()) {
-        pwallet->postInitProcess(scheduler);
+        pwallet->postInitProcess();
+    }
+
+    // Run a thread to flush wallet periodically
+    if (!CWallet::fFlushScheduled.exchange(true)) {
```
promag: `WalletInitInterface::Start()` is called only once, remove this check (and fFlushScheduled)?

src/wallet/wallet.h
```diff
@@ -1103,7 +1102,7 @@ class CWallet final : public CCryptoKeyStore, public CValidationInterface
      * Wallet post-init setup
      * Gives the wallet a chance to register repetitive tasks and complete post-init tasks
      */
-    void postInitProcess(CScheduler& scheduler);
+    void postInitProcess();
```
promag: Post-init doesn't matter now that wallets are dynamically loaded. Remove?

promag:
> I had to change the WalletInit::Start() function to always schedule background wallet flushes

Much better out of CWallet IMO.

jonasschnelli:
Needs rebase again.

jnewbery:
Rebased on #13028

src/wallet/wallet.cpp
```diff
@@ -3968,6 +3969,64 @@ std::vector<std::string> CWallet::GetDestValues(const std::string& prefix) const
     return values;
 }
 
+bool CWallet::Verify(std::string wallet_file, bool salvage_wallet)
```
promag: Lock `cs_wallets`?

promag: nit, `const std::string& wallet_file`.

src/wallet/rpcwallet.cpp
```diff
@@ -2874,6 +2874,49 @@ UniValue listwallets(const JSONRPCRequest& request)
...
+
+    CWallet * const wallet = CWallet::CreateWalletFromFile(wallet_file, fs::absolute(wallet_file, GetWalletDir()));
+    if (!wallet) {
+        throw JSONRPCError(RPC_WALLET_ERROR, "Wallet creation failed.");
+    }
+    AddWallet(wallet);
+
+    wallet->postInitProcess();
+
+    UniValue obj(UniValue::VSTR);
```
promag: Not `obj`ect :trollface: 

Anyway, could return object instead?
```json
{
    "name": ...
}
```

src/wallet/rpcwallet.cpp
```diff
@@ -2874,6 +2874,49 @@ UniValue listwallets(const JSONRPCRequest& request)
...
+    obj = wallet_file;
+
+    return obj;
+
```
promag: nit, empty line.

test/functional/wallet_multiwallet.py
```diff
@@ -150,5 +150,24 @@ def run_test(self):
...
+        for wallet_name in wallet_names:
+            self.nodes[0].loadwallet(wallet_name)
+
+        assert_equal(set(self.nodes[0].listwallets()), set(wallet_names))
+
+        # Fail to load if wallet doesn't exist
+        assert_raises_rpc_error(-18, 'Wallet wallets not found.', self.nodes[0].loadwallet, 'wallets')
+
+        # Fail to load duplicate wallets
+        assert_raises_rpc_error(-4, 'Wallet file verification failed: Error loading wallet w1. Duplicate -wallet filename specified.', self.nodes[0].loadwallet, 'w1')
```
promag: Use `wallet_names[0]` instead of `'w1'`?

promag (commented):
Some comments after testing.

src/wallet/init.cpp
jnewbery: Updated comment as suggested by @ryanofsky .

src/wallet/wallet.cpp
jnewbery: > Lock cs_wallets?

Done. How does this interact with your WalletManager work? Was the idea that cs_wallets was only held by WalletManager?

jnewbery: > nit, const std::string& wallet_file.

Thanks - fixed

src/wallet/rpcwallet.cpp
jnewbery: yep - changed to return an object

test/functional/wallet_multiwallet.py
jnewbery: yes, better. Changed

jnewbery:
Rebased on master and addressed @promag's comments.

test/functional/wallet_multiwallet.py
```diff
@@ -170,5 +170,24 @@ def run_test(self):
         assert_equal(w1.getwalletinfo()['paytxfee'], 0)
         assert_equal(w2.getwalletinfo()['paytxfee'], 4.0)
 
+        # dynamic wallet loading
+        self.restart_node(0, ['-nowallet'])
+        for wallet_name in wallet_names:
+            self.nodes[0].loadwallet(wallet_name)
```
conscott: If I add a line here to inspect the wallet after the load:
```
self.nodes[0].getwalletinfo()
``` 
OR
```
self.nodes[0].getbalance()
```
I get the [json exception throw here](https://github.com/bitcoin/bitcoin/blob/master/src/wallet/rpcwallet.cpp#L78-L79) after the second load operation. Is this expected behavior?

jnewbery: Yes - if there are multiple wallets loaded, you need to send the RPC request to the correct wallet endpoint. Look at `TestNode`'s `get_wallet_rpc()` method.

promag:
@conscott thanks for the review, mind review #13028 too?

test/functional/wallet_multiwallet.py
jnewbery: Yes, good idea. I'll update the test

jnewbery:
Updated test following @conscott's suggestion and rebased on master to get #13028

doc/release-notes-pr10740.md
```diff
@@ -0,0 +1,6 @@
+Dynamic loading of wallets
+--------------------------
+
+Previously, wallets could only be loaded at startup, by specifying `-wallet` parameters on the command line or in the bitcoin.conf file. It is now possible to load wallets dynamically at runtime by calling the `loadwallet` RPC. The wallet file/directory must be located in the `walletdir` directory in order to be loaded.
```
promag: > The wallet file/directory must be located in the `walletdir` directory in order to be loaded.

It works with external wallets (absolute path not related with `-walletdir`).

promag:
Consider:
```
mkdir /tmp/ws
bitcoind -regtest -walletdir=/tmp/ws -wallet=w1 -wallet=w2
bitcoin-cli -regtest listwallets
[
  "w1",
  "w2"
]
```
Then:
```
bitcoind -regtest -walletdir=/tmp/ws -wallet=w1
bitcoin-cli -regtest loadwallet /tmp/ws/w2
bitcoin-cli -regtest listwallets
[
  "w1",
  "/tmp/ws/w2"
]
```
Also
```
bitcoind -regtest -walletdir=/tmp/ws -wallet=w1 -wallet=/tmp/ws/w2
bitcoin-cli -regtest listwallets
[
  "w1",
  "/tmp/ws/w2"
]
```
Is this fine or should the outputs be the same?
Should it be possible to specify an absolute path if it's in the `walletdir`?

jnewbery:
> Is this fine or should the outputs be the same?
> Should it be possible to specify an absolute path if it's in the walletdir?

I think that question is orthogonal to this PR. As you've noted, this is pre-existing behaviour and is exhibited when specifying wallet names on the command line or .conf file. This behaviour was specified in the PR that introduced external wallet files (here: https://github.com/bitcoin/bitcoin/pull/11687#discussion_r158150913).

Can we save discussion on whether that's the best behaviour for outside this PR?

doc/release-notes-pr10740.md
jnewbery: Thanks promag. Release notes update. Please let me know what you think.

conscott:
Tested Re-ACK a395059d4f576cc1b3b61ef78f2e01dec7758d3f

jnewbery:
@promag , @conscott - thanks for the review!

This could really do with some review from people with longer term wallet experience - nagging @TheBlueMatt @jonasschnelli @luke-jr @laanwj :slightly_smiling_face:

jonasschnelli:
utACK a395059d4f576cc1b3b61ef78f2e01dec7758d3f

promag (commented):
@jnewbery please check these missing changes.

jnewbery:
Oops. Thanks @promag . I accidentally dropped those changes in one of my force-pushes. They should be fixed now.

Thanks for keeping me honest!

jnewbery:
rebased

src/wallet/rpcwallet.cpp
```diff
@@ -2994,6 +2994,49 @@ static UniValue listwallets(const JSONRPCRequest& request)
...
+    CWallet * const wallet = CWallet::CreateWalletFromFile(wallet_file, fs::absolute(wallet_file, GetWalletDir()));
+    if (!wallet) {
+        throw JSONRPCError(RPC_WALLET_ERROR, "Wallet creation failed.");
+    }
+    AddWallet(wallet);
+
+    wallet->postInitProcess();
+
+    UniValue obj(UniValue::VOBJ);
+    obj.pushKV("name", wallet_file);
```
promag: nit, `wallet->GetName()` instead.

promag (commented):
Tested ACK 3d3e785.

src/wallet/rpcwallet.cpp
jnewbery: Yes, good idea. Updated.

src/wallet/rpcwallet.cpp
```diff
@@ -2994,6 +2994,49 @@ static UniValue listwallets(const JSONRPCRequest& request)
...
+    std::string wallet_file = request.params[0].get_str();
+    std::string error;
+
+    fs::path wallet_path = fs::absolute(wallet_file, GetWalletDir());
+    if (fs::symlink_status(wallet_path).type() == fs::file_not_found) {
+        throw JSONRPCError(RPC_WALLET_NOT_FOUND, "Wallet " + wallet_file + " not found.");
+    }
+
+    std::string dummy_warning;
+    if (!CWallet::Verify(wallet_file, false, error, dummy_warning)) {
```
TheBlueMatt: This needs some kind of lock to avoid double-loading a wallet in two rpc threads at once.

src/wallet/rpcwallet.cpp
```diff
@@ -2994,6 +2994,49 @@ static UniValue listwallets(const JSONRPCRequest& request)
     return obj;
 }
 
+UniValue loadwallet(const JSONRPCRequest& request)
+{
+    if (request.fHelp || request.params.size() != 1)
+        throw std::runtime_error(
+            "loadwallet \"filename\"\n"
+            "\nLoads a wallet from a wallet file or directory.\n"
```
TheBlueMatt: Maybe some help messages noting that certain config options still apply here - eg if you started with something like zapwallettxes, upgradewallet, rescan, etc it will take the same action for newly-loaded wallets here.

TheBlueMatt (commented):
Should probably fix the various memory leaks in CreateWalletFromFile first - we didnt really use to care cause it would just persist through operation, now we do.

promag:
> Should probably fix the various memory leaks in CreateWalletFromFile first - we didnt really use to care cause it would just persist through operation, now we do.

@TheBlueMatt #13063 fixes that.

src/wallet/wallet.h
```diff
@@ -1080,6 +1080,9 @@ class CWallet final : public CCryptoKeyStore, public CValidationInterface
     /** Mark a transaction as replaced by another transaction (e.g., BIP 125). */
     bool MarkReplaced(const uint256& originalHash, const uint256& newHash);
 
+    //! Verify wallet naming and perform salvage on the wallet if required
+    static bool VerifyWallet(std::string walletFile, bool salvage_wallet);
```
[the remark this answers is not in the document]
promag: @TheBlueMatt btw my suggestion is to move (some) static code to a `WalletManager` (#13034).

TheBlueMatt:
#13063 seems overkill for this issue, we could just use a unique_ptr within CreateWalletFromFile.

src/wallet/rpcwallet.cpp
promag: Hmm @jnewbery this is missing https://github.com/bitcoin/bitcoin/pull/10740#discussion_r183742636.

promag:
@TheBlueMatt Then you mean #12647. Either way it's something that can be fixed in another PR.

TheBlueMatt:
@promag it can be fixed in another PR, but that would have to be merged before this one. Its really no big deal on master, but with an RPC that allows you to create arbitrary memory leaks, that's not great...

src/wallet/wallet.h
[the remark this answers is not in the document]
jnewbery: I agree with @promag that a WalletManager class is a good place for this in the long run.

Since this is a nit only, can we save it for a future PR.

src/wallet/rpcwallet.cpp
jnewbery: Re-added. Thanks @promag

jnewbery:
> Should probably fix the various memory leaks in CreateWalletFromFile first - we didnt really use to care cause it would just persist through operation, now we do.

I've added the minimal fix from here: https://github.com/bitcoin/bitcoin/pull/12647#discussion_r174875670 to address the memory lead. @promag - that conflicts with your shared pointer PR, but it should be trivial for you to rebase that.

src/wallet/rpcwallet.cpp
```diff
@@ -2994,6 +2994,51 @@ static UniValue listwallets(const JSONRPCRequest& request)
...
+    }
+
+    std::string dummy_warning;
+    if (!CWallet::Verify(wallet_file, false, error, dummy_warning)) {
+        throw JSONRPCError(RPC_WALLET_ERROR, "Wallet file verification failed: " + error);
+    }
+
+    CWallet * const wallet = CWallet::CreateWalletFromFile(wallet_file, fs::absolute(wallet_file, GetWalletDir()));
+    if (!wallet) {
+        throw JSONRPCError(RPC_WALLET_ERROR, "Wallet creation failed.");
```
laanwj: "creation failed" is a confusing error message to give to the user here, "Wallet loading failed" would be more straightforward

src/wallet/wallet.cpp
```diff
@@ -3951,6 +3951,56 @@ std::vector<std::string> CWallet::GetDestValues(const std::string& prefix) const
...
+    // 2. Path to an existing directory.
+    // 3. Path to a symlink to a directory.
+    // 4. For backwards compatibility, the name of a data file in -walletdir.
+    LOCK(cs_wallets);
+    fs::path wallet_path = fs::absolute(wallet_file, GetWalletDir());
+    fs::file_type path_type = fs::symlink_status(wallet_path).type();
+    if (!(path_type == fs::file_not_found || path_type == fs::directory_file ||
+          (path_type == fs::symlink_file && fs::is_directory(wallet_path)) ||
+          (path_type == fs::regular_file && fs::path(wallet_file).filename() == wallet_file))) {
+        error_string = strprintf(
```
laanwj: As these errors are reported over RPC, we should stop translating them; RPC errors should not depend on the locale.

src/wallet/wallet.cpp
```diff
@@ -3973,7 +4023,8 @@ CWallet* CWallet::CreateWalletFromFile(const std::string& name, const fs::path&
 
     int64_t nStart = GetTimeMillis();
     bool fFirstRun = true;
-    CWallet *walletInstance = new CWallet(name, WalletDatabase::Create(path));
+    auto wallet_deleter = MakeUnique<CWallet>(name, WalletDatabase::Create(path));
```
laanwj: The name `wallet_deleter` scares me.

laanwj (commented):
Looks overall good to me. Some small comments.

src/wallet/rpcwallet.cpp
jnewbery: Agree. Changed

src/wallet/wallet.cpp
jnewbery: Yes - removed translation.

src/wallet/wallet.cpp
jnewbery: is `temp_wallet` (with a comment above) less scary?

laanwj: Yes, certainly. What I'd prefer, ideally, is to just call this wallet_instance and use this smart pointer everywhere instead of `CWallet* walletInstance = temp_wallet.get()`.
But this is fine, it's no longer scary.

laanwj:
utACK 5ef858bfed2438746db0ba096ff3c98755603cb8

src/wallet/wallet.cpp
jnewbery: ok, I'll change it if I need to update the branch for another reason.

src/wallet/wallet.cpp
```diff
@@ -3950,6 +3951,65 @@ std::vector<std::string> CWallet::GetDestValues(const std::string& prefix) const
...
+          (path_type == fs::regular_file && fs::path(wallet_file).filename() == wallet_file))) {
+        return InitError(strprintf(
+            _("Invalid -wallet path '%s'. -wallet path should point to a directory where wallet.dat and "
+              "database/log.?????????? files can be stored, a location where such a directory could be created, "
+              "or (for backwards compatibility) the name of an existing data file in -walletdir (%s)"),
+            wallet_file, GetWalletDir()));
+    }
+
+    // Make sure that the wallet path doesn't clash with an existing wallet path
+    std::set<fs::path> wallet_paths;
```
jimpo: commit: [wallet] Add CWallet::Verify function

Using a set here is unnecessary, can just do a direct equality check.

src/wallet/rpcwallet.cpp
```diff
@@ -2994,6 +2994,51 @@ static UniValue listwallets(const JSONRPCRequest& request)
...
+        );
+    std::string wallet_file = request.params[0].get_str();
+    std::string error;
+
+    fs::path wallet_path = fs::absolute(wallet_file, GetWalletDir());
+    if (fs::symlink_status(wallet_path).type() == fs::file_not_found) {
+        throw JSONRPCError(RPC_WALLET_NOT_FOUND, "Wallet " + wallet_file + " not found.");
+    }
+
+    std::string dummy_warning;
```
jimpo: commit: [wallet] [rpc] Add loadwallet RPC

Feels like this warning should get logged or maybe even returned in a "warning" field in the JSON response.

src/wallet/wallet.cpp
```diff
@@ -4141,7 +4144,7 @@ CWallet* CWallet::CreateWalletFromFile(const std::string& name, const fs::path&
     }
 
     walletInstance->m_last_block_processed = chainActive.Tip();
-    RegisterValidationInterface(walletInstance);
+    RegisterValidationInterface(temp_wallet.release());
```
jimpo: commit: [wallet] Fix potential memory leak in CreateWalletFromFile

This isn't a complete fix because there are still early returns with errors in the rescan logic below. I believe it is safe to move the `RegisterValidationInterface` call to the bottom of the function as I argued in #12647 because `cs_main` gets locked on line 4135.

src/wallet/init.cpp
jimpo: Appears this comment update was lost somewhere?

laanwj:
This seems very near ready for merge except for @jimpo's nits.

src/wallet/wallet.cpp
jnewbery: Yes, much better idea.

src/wallet/rpcwallet.cpp
jnewbery: ok, added to the JSON response.

src/wallet/wallet.cpp
jnewbery: Done

src/wallet/wallet.cpp
```diff
@@ -3989,6 +3990,66 @@ void CWallet::MarkPreSplitKeys()
...
+
+    // Make sure that the wallet path doesn't clash with an existing wallet path
+    std::set<fs::path> wallet_paths;
+    for (auto wallet : GetWallets()) {
+        if (fs::absolute(wallet->GetName(), GetWalletDir()) == wallet_path) {
+            return InitError(strprintf(_("Error loading wallet %s. Duplicate -wallet filename specified."), wallet_file));
+        }
+    }
+
+    if (!wallet_paths.insert(wallet_path).second) {
```
jimpo: [wallet] Add CWallet::Verify function

Can drop this and the declaration of `wallet_paths` above.

jimpo (commented):
Sorry, one more thing then I'll ACK.

src/wallet/wallet.cpp
promag: True, no longer necessary.

jnewbery: Thanks @jimpo. Fixed

PR #10743, opened 2017-07-04
[test] don't run dbcrash.py on Travis

jnewbery:
dbcrash.db is disk intensive (it's stop-starting nodes and flushing to disk many times). Travis performs badly with heavy disk load and even with the fixes in #10704 the test still fails with timeouts in individual calls.

Even if none of the individual calls timeout, the test takes far too long on Travis. Here's an example of the test timing out the 20 minute max for an individual functional test: https://travis-ci.org/bitcoin/bitcoin/jobs/250016845. Note that the test was only about half way through when it was killed, so without the timeout in test_runner this would have taken ~40 minutes and travis would have killed the job.

For now I think we should just exclude dbcrash.py from Travis in the same way as pruning.py. It doesn't make sense to run disk-heavy tests on Travis.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -68,7 +68,7 @@ script:
     - make $MAKEJOBS $GOAL || ( echo "Build failure. Verbose build follows." && make $GOAL V=1 ; false )
     - export LD_LIBRARY_PATH=$TRAVIS_BUILD_DIR/depends/$HOST/lib
     - if [ "$RUN_TESTS" = "true" ]; then travis_wait 30 make $MAKEJOBS check VERBOSE=1; fi
-    - if [ "$TRAVIS_EVENT_TYPE" = "cron" ]; then extended="--extended --exclude pruning"; fi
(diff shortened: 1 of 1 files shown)


fanquake:
utACK bd00fa5

practicalswift:
utACK bd00fa52a1818f125c32d9759c3ddf506310271a

gmaxwell:
:-/   No disagreement, but we need a way to run these tests.

jnewbery:
> we need a way to run these tests.

for now, it's locally or not at all. Note that this is an extended test that was being run in the travis daily cron job. We weren't even running the extended tests on Travis at all until a couple of months ago, so we're in a lot better place than we were then.

laanwj:
> :-/ No disagreement, but we need a way to run these tests.

The way to run these tests, as well as all others, is simply:
```
test/functional/test_runner.py --extended
```
Running tests on Travis is not meant as replacement for running them locally.

PR #10744, opened 2017-07-04
Use method name via __func__ macro

darksh1ne:
Use __func__ macro in std::runtime_exception to:
1. fix method name in CWalletTx::GetAvailableWatchOnlyCredit()
2. refactor CWalletTx::GetAvailableCredit()

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -1718,7 +1718,7 @@ CAmount CWalletTx::GetAvailableCredit(bool fUseCache) const
             const CTxOut &txout = tx->vout[i];
             nCredit += pwallet->GetCredit(txout, ISMINE_SPENDABLE);
             if (!MoneyRange(nCredit))
-                throw std::runtime_error("CWalletTx::GetAvailableCredit() : value out of range");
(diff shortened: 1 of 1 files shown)


gmaxwell:
Other places do this, but it frustrates grepping the source code.

jonasschnelli:
If you do so, maybe also tackle the LogPrintf's and get rid of static "CWalletXX::" there?

jnewbery:
> it frustrates grepping the source code

At first glance your comment seems reasonable, but is it really a problem? Why not just grep for the function name? Blindly grepping for the complete log string is already going to fail for those log messages that contain C string format specifiers.

laanwj:
Using `__func__` is preferred because it means one less thing that can be forgotten to be updated when moving the code around to another function.
This guarantees that the message will actually have been emitted in the function that is mentioned in the message.
As for grepping the message, you could search either the function itself, then locate it manually, or for part of the message that follows the function name. This is already the case as `__func__` is used almost everywhere, it's not a discussion worth having IMO.

MarcoFalke:
utACK 9bbf60047ea3300ebaffbe8cdc7c71e984418c20

PR #10748, opened 2017-07-05
[config] Help text cleanup

jnewbery:
This is very much a matter of personal taste, but I think this changeset makes command line arguments help text code clearer.

A bunch of changes:

- order groupings alphabetically
- order arguments alphabetically within groupings
- make `HelpMessageOpt()` responsible for whether to print debug arguments. This is done by passing a `HELP_MESSAGE_FILTER`, which will result in the help text being printed only if `-debug-help` is set
- move server-specific option into bitcoind.cpp, and remove the `mode` argument from `HelpMessage()`
- move the calls to `GetWalletHelpString()` into bitcoind.cpp and qt (removes one dependency on bitcoin_wallet from bitcoin_server).

the help message filter is implemented as a bitfield, so could be updated to include other filters (eg `HELP_MESSAGE_FILTER_NOT_WIN32`) and remove the preprocessor conditionals from `HelpMessage()`

### DIFF
--- a/src/bitcoin-cli.cpp
+++ b/src/bitcoin-cli.cpp
@@ -37,15 +37,15 @@ std::string HelpMessageCli()
     strUsage += HelpMessageOpt("-?", _("This help message"));
     strUsage += HelpMessageOpt("-conf=<file>", strprintf(_("Specify configuration file (default: %s)"), BITCOIN_CONF_FILENAME));
     strUsage += HelpMessageOpt("-datadir=<dir>", _("Specify data directory"));
-    AppendParamsHelpMessages(strUsage);
     strUsage += HelpMessageOpt("-named", strprintf(_("Pass named instead of positional arguments (default: %s)"), DEFAULT_NAMED));
-    strUsage += HelpMessageOpt("-rpcconnect=<ip>", strprintf(_("Send commands to node running on <ip> (default: %s)"), DEFAULT_RPCCONNECT));
-    strUsage += HelpMessageOpt("-rpcport=<port>", strprintf(_("Connect to JSON-RPC on <port> (default: %u or testnet: %u)"), defaultBaseParams->RPCPort(), testnetBaseParams->RPCPort()));
-    strUsage += HelpMessageOpt("-rpcwait", _("Wait for RPC server to start"));
-    strUsage += HelpMessageOpt("-rpcuser=<user>", _("Username for JSON-RPC connections"));
-    strUsage += HelpMessageOpt("-rpcpassword=<pw>", _("Password for JSON-RPC connections"));
     strUsage += HelpMessageOpt("-rpcclienttimeout=<n>", strprintf(_("Timeout in seconds during HTTP requests, or 0 for no timeout. (default: %d)"), DEFAULT_HTTP_CLIENT_TIMEOUT));
+    strUsage += HelpMessageOpt("-rpcconnect=<ip>", strprintf(_("Send commands to node running on <ip> (default: %s)"), DEFAULT_RPCCONNECT));
+    strUsage += HelpMessageOpt("-rpcpassword=<pw>", _("Password for JSON-RPC connections"));
+    strUsage += HelpMessageOpt("-rpcport=<port>", strprintf(_("Connect to JSON-RPC on <port> (default: %u or testnet: %u)"), defaultBaseParams->RPCPort(), testnetBaseParams->RPCPort()));
+    strUsage += HelpMessageOpt("-rpcuser=<user>", _("Username for JSON-RPC connections"));
+    strUsage += HelpMessageOpt("-rpcwait", _("Wait for RPC server to start"));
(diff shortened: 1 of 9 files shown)


practicalswift:
Concept ACK

src/bitcoind.cpp
```diff
@@ -19,6 +19,9 @@
 #include "httpserver.h"
 #include "httprpc.h"
 #include "utilstrencodings.h"
+#ifdef ENABLE_WALLET
```
laanwj: Right, we need to move the wallet dependency up here if we want to remove the wallet dependency from `libbitcoin_server`.

src/util.h
```diff
@@ -44,6 +44,9 @@ class CTranslationInterface
     boost::signals2::signal<std::string (const char* psz)> Translate;
 };
 
+extern const int HELP_MESSAGE_FILTER_NONE;
```
laanwj: Why not define the flags in the header? That's the usual approach used for constants throughout the code, and has the advantage of making the values available to the compiler so the linker doesn't have to reserve memory for them.

src/qt/utilitydialog.cpp
```diff
@@ -20,6 +20,9 @@
 #include "clientversion.h"
 #include "init.h"
 #include "util.h"
+#ifdef ENABLE_WALLET
```
laanwj: Please don't introduce a wallet dependency here.

jnewbery: I'm not very familiar with the qt code. I see that there are already plenty of `#ifdef ENABLE_WALLET` in the qt code already. Is there something special about utilitydialog.cpp ?

I'm very happy to take an alternative approach if you have suggestions.

src/util.h
jnewbery: I've moved the definition to the header file

jnewbery:
Addressed @laanwj's comment, squashed fixup commit and rebased.

src/qt/utilitydialog.cpp
laanwj: > Is there something special about utilitydialog.cpp ?

No, but adding `ENABLE_WALLET` should be avoided everywhere if possible. I think it may be unavoidable here, though.

jonasschnelli: I think a possible solution would be to have a signal for collecting help strings. The signal emitter then could take care of sorting the groups. But probably out of scope.

jnewbery: > possible solution would be to have a signal for collecting help strings

Certainly possible to have a signal for wallet to pass its help strings to server, but qt also has its own help strings (bitcoind should also have its own help string for `-server`, but that's actually implemented in init.cpp for now). Difficult to see where/how those would all be sorted.

Really I don't think it matters too much, and I'm happy with other approaches, or to drop this PR entirely if people don't think this is an improvement.

jnewbery:
This is a heavily-conflicting PR that will require frequent rebase. I'm not keen on doing that unless people think this is an improvement and should get merged.

@laanwj @jonasschnelli you've both taken a look at this and left comments. Can I assume concept ACKs from you? Do you think this is an improvement?

src/bitcoin-cli.cpp
```diff
@@ -37,15 +37,15 @@ std::string HelpMessageCli()
     strUsage += HelpMessageOpt("-?", _("This help message"));
     strUsage += HelpMessageOpt("-conf=<file>", strprintf(_("Specify configuration file (default: %s)"), BITCOIN_CONF_FILENAME));
     strUsage += HelpMessageOpt("-datadir=<dir>", _("Specify data directory"));
-    AppendParamsHelpMessages(strUsage);
     strUsage += HelpMessageOpt("-named", strprintf(_("Pass named instead of positional arguments (default: %s)"), DEFAULT_NAMED));
+    strUsage += HelpMessageOpt("-rpcclienttimeout=<n>", strprintf(_("Timeout in seconds during HTTP requests, or 0 for no timeout. (default: %d)"), DEFAULT_HTTP_CLIENT_TIMEOUT));
     strUsage += HelpMessageOpt("-rpcconnect=<ip>", strprintf(_("Send commands to node running on <ip> (default: %s)"), DEFAULT_RPCCONNECT));
+    strUsage += HelpMessageOpt("-rpcpassword=<pw>", _("Password for JSON-RPC connections"));
```
MarcoFalke: Imo rpcpassword belongs to rpcuser, no need to sort alphabetically.

MarcoFalke:
There seems to be no conceptual acknowledgment to do this. Closing for now.

jnewbery:
Please re-open. There hasn't been a rejection of this idea.

MarcoFalke:
Ok, leaving this open for another month. Though, if no one wants to take a look I don't see a reason to keep it open forever.

jnewbery:
I think you're right - there's no appetite for this PR. I may break it up and PR some parts of it, since I think they're useful changes.

PR #10750, opened 2017-07-05
[wallet] Change bumpfee totalFee option to feeRate option

instagibbs:
Using a total fee argument is of dubious utility and also makes further improvements where you are changing inputs/outputs difficult. To use the previous argument successfully the user must do additional math, rather than taking any fee estimate desired and plugging it in

Did satoshis/byte since `totalFee` was satoshis.

### DIFF
--- a/src/qt/walletmodel.cpp
+++ b/src/qt/walletmodel.cpp
@@ -667,7 +667,7 @@ bool WalletModel::bumpFee(uint256 hash)
     std::unique_ptr<CFeeBumper> feeBump;
     {
         LOCK2(cs_main, wallet->cs_wallet);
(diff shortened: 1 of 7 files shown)


morcos:
@instagibbs I think this is intentionally a total fee.  There are BIP 125 requirements both for the fee rate and the total fee of the replacement transaction, but the one that is more likely to cause an issue is the total fee because it requires looking at all the descendant transactions of the original.  Thus if manually specifying the replacement transaction, you have clearer control by specifying the total fee.

instagibbs:
After much discussion, it seems that this can not be gotten rid of(yet), so I will close for now. Subsequent improvements may just have to route around this feature.

PR #10753, opened 2017-07-06
test: Check RPC argument mapping

laanwj:
Parse the dispatch tables from the server implementation files, and the conversion table from the client (see #10751).

Perform the following consistency checks:

- Arguments defined in conversion table, must be present in dispatch table. If not, it was probably forgotten to add them to the dispatch table, and they will not work.

- Arguments defined in conversion table must have the same names as in the dispatch table. If not, they will not work.

- All aliases for an argument must either be present in the conversion table, or not. Anything in between means an oversight and some aliases won't work.

Any of these results in an error.

It also performs a consistency check to see if the same named argument is sometimes converted, and sometimes not. E.g. one RPC call might have a 'verbose' argument that is converted,
another RPC call might have one that is not converted. This is not necessarily wrong, but points at a possible error (as well as makes the API harder to memorize) - so it is emitted as a warning (could upgrade this to error).

This test is added to travis and run when `CHECK_DOC`. Currently fails with the following output:
```
* Checking consistency between dispatch tables and vRPCConvertParams
ERROR: createrawtransaction argument 3 (named optintorbf in vRPCConvertParams) is not defined in dispatch table
ERROR: getblock argument ['verbosity', 'verbose'] has conflicts in vRPCConvertParams conversion specifier [True, False]
WARNING: conversion mismatch for argument named verbose ([('getblock', False), ('getblockheader', True), ('getmempoolancestors', True), ('getmempooldescendants', True), ('getrawmempool', True), ('getrawtransaction', True)])
```
- ~#10698 fixes the first ERROR~
- #10747 fixes the second ERROR, as well as the WARNING

Update: #10698 was merged, leaving:
```
* Checking consistency between dispatch tables and vRPCConvertParams
ERROR: getblock argument ['verbosity', 'verbose'] has conflicts in vRPCConvertParams conversion specifier [True, False]
WARNING: conversion mismatch for argument named verbose ([('getblock', False), ('getblockheader', True), ('getmempoolancestors', True), ('getmempooldescendants', True), ('getrawmempool', True), ('getrawtransaction', True)])
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -45,6 +45,7 @@ install:
 before_script:
     - if [ "$CHECK_DOC" = 1 -a "$TRAVIS_EVENT_TYPE" = "pull_request" ]; then contrib/devtools/commit-script-check.sh $TRAVIS_COMMIT_RANGE; fi
     - if [ "$CHECK_DOC" = 1 ]; then contrib/devtools/check-doc.py; fi
+    - if [ "$CHECK_DOC" = 1 ]; then contrib/devtools/check-rpc-mappings.py .; fi
     - unset CC; unset CXX
     - mkdir -p depends/SDKs depends/sdk-sources
     - if [ -n "$OSX_SDK" -a ! -f depends/sdk-sources/MacOSX${OSX_SDK}.sdk.tar.gz ]; then curl --location --fail $SDK_URL/MacOSX${OSX_SDK}.sdk.tar.gz -o depends/sdk-sources/MacOSX${OSX_SDK}.sdk.tar.gz; fi
--- a/contrib/devtools/check-rpc-mappings.py
+++ b/contrib/devtools/check-rpc-mappings.py
@@ -1,0 +1,158 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Check RPC argument consistency."""
+
+from collections import defaultdict
+import os
+import re
+import sys
+
+# Source files (relative to root) to scan for dispatch tables
+SOURCES = [
+    "src/rpc/server.cpp",
+    "src/rpc/blockchain.cpp",
+    "src/rpc/mining.cpp",
+    "src/rpc/misc.cpp",
+    "src/rpc/net.cpp",
+    "src/rpc/rawtransaction.cpp",
+    "src/wallet/rpcwallet.cpp",
+]
+# Source file (relative to root) containing conversion mapping
+SOURCE_CLIENT = 'src/rpc/client.cpp'
+# Argument names that should be ignored in consistency checks
+IGNORE_DUMMY_ARGS = {'dummy', 'arg0', 'arg1', 'arg2', 'arg3', 'arg4', 'arg5', 'arg6', 'arg7', 'arg8', 'arg9'}
+
+class RPCCommand:
+    def __init__(self, name, args):
+        self.name = name
+        self.args = args
+
+class RPCArgument:
+    def __init__(self, names, idx):
+        self.names = names
+        self.idx = idx
+        self.convert = False
+
+def parse_string(s):
+    assert s[0] == '"'
+    assert s[-1] == '"'
+    return s[1:-1]
+
+def process_commands(fname):
+    """Find and parse dispatch table in implementation file `fname`."""
+    cmds = []
+    in_rpcs = False
+    with open(fname, "r") as f:
+        for line in f:
+            line = line.rstrip()
+            if not in_rpcs:
+                if re.match("static const CRPCCommand .*\[\] =", line):
+                    in_rpcs = True
+            else:
+                if line.startswith('};'):
+                    in_rpcs = False
+                elif '{' in line and '"' in line:
+                    m = re.search('{ *("[^"]*"), *("[^"]*"), *&([^,]*), *{([^}]*)} *},', line)
+                    assert m, 'No match to table expression: %s' % line
+                    name = parse_string(m.group(2))
+                    args_str = m.group(4).strip()
+                    if args_str:
+                        args = [RPCArgument(parse_string(x.strip()).split('|'), idx) for idx, x in enumerate(args_str.split(','))]
+                    else:
+                        args = []
+                    cmds.append(RPCCommand(name, args))
+    assert not in_rpcs, "Something went wrong with parsing the C++ file: update the regexps"
+    return cmds
+
+def process_mapping(fname):
+    """Find and parse conversion table in implementation file `fname`."""
+    cmds = []
+    in_rpcs = False
+    with open(fname, "r") as f:
+        for line in f:
+            line = line.rstrip()
+            if not in_rpcs:
+                if line == 'static const CRPCConvertParam vRPCConvertParams[] =':
+                    in_rpcs = True
+            else:
+                if line.startswith('};'):
+                    in_rpcs = False
+                elif '{' in line and '"' in line:
+                    m = re.search('{ *("[^"]*"), *([0-9]+) *, *("[^"]*") *},', line)
+                    assert m, 'No match to table expression: %s' % line
+                    name = parse_string(m.group(1))
+                    idx = int(m.group(2))
+                    argname = parse_string(m.group(3))
+                    cmds.append((name, idx, argname))
+    assert not in_rpcs
+    return cmds
+
+def main():
+    root = sys.argv[1]
+
+    # Get all commands from dispatch tables
+    cmds = []
+    for fname in SOURCES:
+        cmds += process_commands(os.path.join(root, fname))
+
+    cmds_by_name = {}
+    for cmd in cmds:
+        cmds_by_name[cmd.name] = cmd
+
+    # Get current convert mapping for client
+    client = SOURCE_CLIENT
+    mapping = set(process_mapping(os.path.join(root, client)))
+
+    print('* Checking consistency between dispatch tables and vRPCConvertParams')
+
+    # Check mapping consistency
+    errors = 0
+    for (cmdname, argidx, argname) in mapping:
+        try:
+            rargnames = cmds_by_name[cmdname].args[argidx].names
+        except IndexError:
+            print('ERROR: %s argument %i (named %s in vRPCConvertParams) is not defined in dispatch table' % (cmdname, argidx, argname))
+            errors += 1
+            continue
+        if argname not in rargnames:
+            print('ERROR: %s argument %i is named %s in vRPCConvertParams but %s in dispatch table' % (cmdname, argidx, argname, rargnames), file=sys.stderr)
+            errors += 1
+
+    # Check for conflicts in vRPCConvertParams conversion
+    # All aliases for an argument must either be present in the
+    # conversion table, or not. Anything in between means an oversight
+    # and some aliases won't work.
+    for cmd in cmds:
+        for arg in cmd.args:
+            convert = [((cmd.name, arg.idx, argname) in mapping) for argname in arg.names]
+            if any(convert) != all(convert):
+                print('ERROR: %s argument %s has conflicts in vRPCConvertParams conversion specifier %s' % (cmd.name, arg.names, convert))
+                errors += 1
+            arg.convert = all(convert)
+
+    # Check for conversion difference by argument name.
+    # It is preferable for API consistency that arguments with the same name
+    # have the same conversion, so bin by argument name.
+    all_methods_by_argname = defaultdict(list)
+    converts_by_argname = defaultdict(list)
+    for cmd in cmds:
+        for arg in cmd.args:
+            for argname in arg.names:
+                all_methods_by_argname[argname].append(cmd.name)
+                converts_by_argname[argname].append(arg.convert)
+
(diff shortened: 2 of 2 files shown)


jnewbery:
Concept ACK. Will review later.

contrib/devtools/check-rpc-mappings.py
```diff
@@ -0,0 +1,165 @@
...
+                if re.match("static const CRPCCommand .*\[\] =", line):
+                    in_rpcs = True
+            else:
+                if line.startswith('};'):
+                    in_rpcs = False
+                elif '{' in line and '"' in line:
+                    m = re.search('{ *("[^"]*"), *("[^"]*"), *&([^,]*), *(true|false), *{([^}]*)} *},', line)
+                    if not m:
+                        print('No match: %s' % line)
+                        exit(1)
```
practicalswift: Nit: Use `sys.exit(1)`

laanwj: Why?

practicalswift: According to the Python documentation `exit(…)` should be used in the interactive interpreter shell but not in programs:
> The site module (which is imported automatically during startup, except if the -S command-line option is given) adds several constants to the built-in namespace. They are useful for the interactive interpreter shell and should not be used in programs.

Links:
* [`sys.exit(…)`](https://docs.python.org/3/library/sys.html#sys.exit): _"Exit from Python."_
* [`exit(…)`](https://docs.python.org/3/library/constants.html#exit): _"raise SystemExit with the specified exit code. […] should not be used in programs."_

laanwj: Agree then! I never knew that.

promag:
IMO we could somehow share the same data between the server table and the client table so that none of this would be necessary. Don't take me wrong, but parsing code feels bad practice. Beside that, keeping these tables in sync could very well be avoided. I could work in an alternative, unless I'm missing something.

laanwj:
@promag That discussion is in https://github.com/bitcoin/bitcoin/issues/10751#issuecomment-313351532

And yes, in the long run there are absolutely better solutions, but in the short term I think adding this check is good.

promag:
Thanks for pointing that out.

@laanwj could this be executed in pre-commit hook?

laanwj:
> @laanwj could this be executed in pre-commit hook?

Sure, would be a matter of adding the line to execute it to the `.git/hooks/pre-commit` or `.git/hooks/pre-push` script.

promag:
@laanwj another PR or care to add it here?

laanwj:
You have to do it locally. Hooks are not part of the repository (for
security reasons, among others.) Shipping some examples of hooks would work
but that's definitely out of scope here.

ryanofsky:
I think this might be simpler and more robust if it were written as a small c++ tool that just accesses the global variables directly, instead of a python tool that has to parse c++ with regexes.

promag:
@laanwj reason?

jnewbery:
@laanwj did you close this because you're planning to propose an alternative?

laanwj:
@jnewbery No, I don't have the time nor motivation to work on this, but I'm looking forward to a better alternative as no one seems to like this solution.

jnewbery:
I like it! At the very least I like it better than no tests at all.

Running code beats suggestions of better ways to do it without any code.

If you change your mind, I'm happy to review.

contrib/devtools/check-rpc-mappings.py
```diff
@@ -0,0 +1,165 @@
...
+    def __init__(self, name, args):
+        self.name = name
+        self.args = args
+
+class RPCArgument:
+    def __init__(self, names, idx):
+        self.names = names
+        self.idx = idx
+
+def process_commands(fname):
```
jnewbery: I think it'd simplify implementation in `main()` if `process_commands()` and `process_mappings()` both returned a set of tuples `(command, index, argname)`.

jnewbery (commented):
A few nits inline.

I think it'd simplify the implementation if you got rid of the `RPCCommand` and `RPCArgument` classes and returned tuples from the `process_commands()` and `process_mappings()` functions. The different format of the return values makes some of the logic in `main()` a little fiddly.

However, this does the job, and I don't think it's worth the time optimizing this, so ACK 603159515f1fa2b9aa0c4f37c6726803b7c39ded (once #10747 is merged).

I also recommend running a linter such as flake8 over python modules before committing, since removing all warnings can sometimes uncover bugs. Rather than fill this review with nits, I've pushed a branch here that fixes them: https://github.com/jnewbery/bitcoin/tree/pr10753 . Feel free to take/leave/squash as you see fit.

If @ryanofsky or @promag want to implement a C++ unit test, I'm happy to review.

jnewbery:
@laanwj - do you still plan to work on this? If not, do you mind if I take over?

laanwj:
I'll get to this after 0.15. There's not much to be done but I'm completely distracted at the moment.

laanwj:
Rebased and updated for new table syntax without safemode flag.

theuni:
Concept ACK.

sipa:
Concept ACK

contrib/devtools/check-rpc-mappings.py
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/python3
+'''
```
jnewbery: nit: One-line docstrings should be on one line.

contrib/devtools/check-rpc-mappings.py
```diff
@@ -0,0 +1,165 @@
...
+Check RPC argument consistency.
+'''
+import sys, os, re
```
jnewbery: nit: one import per line, sorted by name of module

contrib/devtools/check-rpc-mappings.py
```diff
@@ -0,0 +1,165 @@
...
+from collections import defaultdict
+
+# Source files (relative to root) to scan for dispatch tables
+SOURCES = [
+"src/rpc/server.cpp",
```
jnewbery: nit: indentation for continuation lines.

contrib/devtools/check-rpc-mappings.py
```diff
@@ -0,0 +1,165 @@
...
+SOURCE_CLIENT = 'src/rpc/client.cpp'
+# Argument names that should be ignored in consistency checks
+IGNORE_DUMMY_ARGS = {'dummy','arg0','arg1','arg2','arg3','arg4','arg5','arg6','arg7','arg8','arg9'}
+
+def parse_string(s):
+    assert(s[0] == '"')
+    assert(s[-1] == '"')
+    return s[1:-1]
+
+def make_string(s):
```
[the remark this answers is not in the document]
jnewbery: still unused, please remove

contrib/devtools/check-rpc-mappings.py
```diff
@@ -0,0 +1,165 @@
...
+
+class RPCCommand:
+    def __init__(self, name, args):
+        self.name = name
+        self.args = args
+
+class RPCArgument:
+    def __init__(self, names, idx):
+        self.names = names
+        self.idx = idx
```
jnewbery: suggestion: add `self.convert = False` for self-documentation

contrib/devtools/check-rpc-mappings.py
```diff
@@ -0,0 +1,165 @@
...
+            line = line.rstrip()
+            if not in_rpcs:
+                if re.match("static const CRPCCommand .*\[\] =", line):
+                    in_rpcs = True
+            else:
+                if line.startswith('};'):
+                    in_rpcs = False
+                elif '{' in line and '"' in line:
+                    m = re.search('{ *("[^"]*"), *("[^"]*"), *&([^,]*), *{([^}]*)} *},', line)
+                    if not m:
```
jnewbery: Perhaps replace with an assert:

`assert m, "No match: %s" % line`

contrib/devtools/check-rpc-mappings.py
```diff
@@ -0,0 +1,165 @@
...
+                        print('No match: %s' % line)
+                        exit(1)
+                    name = parse_string(m.group(2))
+                    args_str = m.group(5).strip()
+                    if args_str:
+                        args = [RPCArgument(parse_string(x.strip()).split('|'), idx) for idx,x in enumerate(args_str.split(','))]
+                    else:
+                        args = []
+                    cmds.append(RPCCommand(name, args))
+    assert(not in_rpcs) # If this assertion is hit, something went wrong with parsing the C++ file: update the regexps
```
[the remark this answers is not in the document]
jnewbery: please remove parens from assert

contrib/devtools/check-rpc-mappings.py
```diff
@@ -0,0 +1,165 @@
...
+            line = line.rstrip()
+            if not in_rpcs:
+                if line == 'static const CRPCConvertParam vRPCConvertParams[] =':
+                    in_rpcs = True
+            else:
+                if line.startswith('};'):
+                    in_rpcs = False
+                elif '{' in line and '"' in line:
+                    m = re.search('{ *("[^"]*"), *([0-9]+) *, *("[^"]*") *},', line)
+                    if not m:
```
jnewbery: perhaps replace with an assert?

contrib/devtools/check-rpc-mappings.py
```diff
@@ -0,0 +1,165 @@
...
+        try:
+            rargnames = cmds_by_name[cmdname].args[argidx].names
+        except IndexError:
+            print('ERROR: %s argument %i (named %s in vRPCConvertParams) is not defined in dispatch table' % (cmdname, argidx, argname))
+            errors += 1
+            continue
+        if argname not in rargnames:
+            print('ERROR: %s argument %i is named %s in vRPCConvertParams but %s in dispatch table' % (cmdname, argidx, argname, rargnames), file=sys.stderr)
+            errors += 1
+
```
jnewbery: Suggestion: add prints for each check (like you have for the first one):

`print('* Checking for conflicts in vRPCConvertParams conversion')`

etc

contrib/devtools/check-rpc-mappings.py
```diff
@@ -0,0 +1,165 @@
...
+                converts_by_argname[argname].append(arg.convert)
+
+    for argname,convert in converts_by_argname.items():
+        if all(convert) != any(convert):
+            if argname in IGNORE_DUMMY_ARGS: # these are testing or dummy, don't warn for them
+                continue
+            print('WARNING: conversion mismatch for argument named %s (%s)' %
+                    (argname, list(zip(all_methods_by_argname[argname], converts_by_argname[argname]))))
+
+    exit(errors > 0)
```
jnewbery: use `sys.exit()`

contrib/devtools/check-rpc-mappings.py
```diff
@@ -0,0 +1,165 @@
...
+
+if __name__ == '__main__':
+    main()
+
```
jnewbery: Remove trailing blank line

contrib/devtools/check-rpc-mappings.py
```diff
@@ -0,0 +1,165 @@
+#!/usr/bin/python3
```
jnewbery: For portability, use:

```python
#!/usr/bin/env python3
```

Also add copyright notice

jnewbery (commented):
A bunch of nits inline. I've implemented them here: https://github.com/jnewbery/bitcoin/commit/b73ce0a785aa8d6ea69078ae8b7222534f6d01c7. Feel free to take and squash

jtimon:
Concept ACK. If somebody writes a unittest alternative later, it is easy to replace one for the other, not a good reason to stop this.

contrib/devtools/check-rpc-mappings.py
laanwj: Meh, I personally prefer them on multiple lines, makes it easier to extend them later.

laanwj:
Addressed @jnewbery's nits and updates for .travis.yml conflict.

.travis.yml
```diff
@@ -45,7 +45,7 @@ install:
 before_script:
     - if [ "$CHECK_DOC" = 1 -a "$TRAVIS_EVENT_TYPE" = "pull_request" ]; then contrib/devtools/commit-script-check.sh $TRAVIS_COMMIT_RANGE; fi
     - if [ "$CHECK_DOC" = 1 ]; then contrib/devtools/check-doc.py; fi
-    - unset CC; unset CXX
```
jnewbery: Is this removal intended? How is it related to this PR?

contrib/devtools/check-rpc-mappings.py
jnewbery: Thanks for changing it anyway! (for context, I picked up the preference for one-line docstrings from https://www.python.org/dev/peps/pep-0257/#id16)

.travis.yml
laanwj: Sigh, rebase fail

jnewbery:
Tested ACK 77aa9e59ea0e6a000a0aea5dec4ef9585356147d

It's definitely working:

```
ERROR: estimaterawfee argument 0 is named nblocks in vRPCConvertParams but ['conf_target'] in dispatch table
WARNING: conversion mismatch for argument named conf_target ([('estimaterawfee', False), ('sendmany', True), ('sendtoaddress', True)])
```

MarcoFalke:
utACK 77aa9e59ea0e6a000a0aea5dec4ef9585356147d

PR #10757, opened 2017-07-06
RPC: Introduce getblockstats to plot things

jtimon:
It returns per block statistics about several things. It should be easy to add more if people think of other things to add or remove some if I went too far (but once written, why not keep it? EDIT: answer: not to test or maintain them).

The currently available options are: minfee,maxfee,totalfee,minfeerate,maxfeerate,avgfee,avgfeerate,txs,ins,outs (EDIT: see updated list in the rpc call documentation)

For the x axis, one can use height or block.nTime (I guess I could add mediantime if there's interest [EDIT: nobody showed interest but I implemented mediantime nonetheless, in fact there's no distinction between x or y axis anymore, that's for the caller to judge]).

To calculate fees, -txindex is required.


clarkmoody:
Some ideas for additions:
- Non-fee total output amount
- Coinbase reward
- Money supply including this block
- Transaction weight `txweight` (it can be derived from existing fields, however)

I would prefer to see both `time` and `mediantime` returned, since they are available.

Should we return non-independent fields, such as `avgfee` when also including `totalfee` and `txs`?

I find that for bitcoin-related data, the median is often more useful than the average of a distribution. Including `medianweight`, `medianfee`, `medianfeerate`, `medianoutput` etc would expose these useful quantities to the user.

jnewbery:
> but once written, why not keep it?

Because more code => more bugs and more maintenance effort. I prefer:

If it's not really needed, why add it?

This is perhaps a nice-to-have, but since #8704, `getblock` can return all transactions in a block (without requiring `txindex`). Those can then be parsed and analysed offline.

Is there a compelling use-case I'm missing here? This seems like a feature only a small subset of users would be interested in, in which case an offline tools seems more appropriate.

Sorry - not meaning to be negative, but my default reaction to new RPCs/arguments tends towards NACK unless I can see a compelling and widespread use-case.

clarkmoody:
> This is perhaps a nice-to-have, but since #8704, getblock can return all transactions in a block (without requiring txindex). Those can then be parsed and analysed offline.

This code pulls each transaction input's previous outpoint in order to compute transaction fees. Replicating that in RPC would require thousands of calls for most blocks.

jnewbery:
> This code pulls each transaction input's previous outpoint

Ah yes, of course. Concept ACK in that case. Doing this with `getblock` / `getrawtransaction` is infeasible.

jtimon:
> Because more code => more bugs and more maintenance effort. I prefer:

Sure, but I mean, removing for example the avgfee or avgfeerate won't  safe much code or testing code, just a few lines. Forget I said this, if there's specific functions to remove because nobody will want them, let's remove those and focus on the ones people want. Adding specific things only a a few people want can also happen in their own branches, so it's no big deal.

The only use case is gather statistics, presumably to plot things, create charts. That is, at least, compelling to me, but I don't think that will have widespread usage. I also don't think all rpc calls have it. Is getchaintxstats, for example, a widespread use case?

If that's enough reason not to merge this, it's fine, I can maintain it as a separate branch that I periodically rebase, it is simple enough, so that won't be a big deal. On the other hand, if I can get it reviewed and merged it'll be less work for me in the long run and I also get the review.

> Non-fee total output amount
> Coinbase reward

Sounds good.

> Money supply including this block

Mhmm, it would be simpler to calculate here from start to end here than from genesis. But it's pretty trivial to write a function in any language that returns the total supply for a given height without access to any historic data. Unless you are talking about discounting op_return outputs or something like that. I don't think this is very interesting here. Perhaps that can be done in getchaintxstats ?

> Transaction weight txweight (it can be derived from existing fields, however)

In fact I'm using weight for everything. I should s/size/weight/ and probably also show size separately. 
Maybe separate feerates in by weight and serialize size? I don't know...

> I would prefer to see both time and mediantime returned, since they are available.

Yeah, the mediantime takes a little bit longer to be calculated but not much and one can always disable anything. In fact, the height and time shouldn't be treated in any special way for being "the x axis" and should be allowed to be disabled like the rest.

> Should we return non-independent fields, such as avgfee when also including totalfee and txs?

This is a good question. This is mostly what I meant by "why not if it's this easy?".
But yeah, I guess non-independent are good candidates to be removed.

re median: yeah, that sounds interesting too, good idea!

clarkmoody:
> Mhmm, it would be simpler to calculate here from start to end here than from genesis. But it's pretty trivial to write a function in any language that returns the total supply for a given height without access to any historic data. Unless you are talking about discounting op_return outputs or something like that. I don't think this is very interesting here. Perhaps that can be done in getchaintxstats ?

I was thinking of the more trivial version, rather than the `supply - provably_unspendable` version, so keeping that as external code makes more sense. Maintaining the sum of spendable outputs against block height is a much more ambitious idea, and it may make sense in the future. However, it is probably out of scope of this PR.

src/rpc/blockchain.cpp
```diff
@@ -1532,6 +1539,178 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+    }
+    if (end < 0 || end > chainActive.Height()) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "End block height out of range");
+    }
+
+    std::string str_plot_values = str_allowed_plot_values;
+    if (request.params.size() > 2) {
+        str_plot_values = request.params[2].get_str();
+    }
+    std::vector<std::string> plot_values;
```
promag:
```cpp
std::set<std::string> plot_values;
if (request.params.size() > 2) {
  boost::split(plot_values, request.params[2].get_str(), boost::is_any_of(","));

  // only validate in this case
  // ... 
} else {
  plot_values = allowed_plot_values;
}
```

src/rpc/blockchain.cpp
```diff
@@ -1532,6 +1539,178 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+            map_stats[plot_value].push_back((int64_t)block.vtx.size());
+        } else if (plot_value == "ins") {
+            map_stats[plot_value].push_back(inputs);
+        } else if (plot_value == "outs") {
+            map_stats[plot_value].push_back(outputs);
+        }
+    }
+}
+
+static bool IsAllowedPlotValue(const std::string& plot_value, std::vector<std::string>& allowed_plot_values)
```
promag: Remove.

src/rpc/blockchain.cpp
```diff
@@ -1532,6 +1539,178 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+    }
+
+    std::string str_plot_values = str_allowed_plot_values;
+    if (request.params.size() > 2) {
+        str_plot_values = request.params[2].get_str();
+    }
+    std::vector<std::string> plot_values;
+    boost::split(plot_values, str_plot_values, boost::is_any_of(","));
+    for (const std::string plot_value : plot_values) {
+        if (!IsAllowedPlotValue(plot_value, allowed_plot_values)) {
```
promag:
```cpp
if (allowed_plot_values.count(plot_value) == 0) {
```

promag (commented):
Partial review, suggestion to use `std::set`.

Nit, rename `allowed_plot_values` to `valid_plot_values`.
Nit, rename `getperblockstats` to just `getblockstats`?

src/rpc/blockchain.cpp
```diff
@@ -1532,6 +1539,178 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+            "\nArguments:\n"
+            "1. \"start\"      (numeric, required) The height of the block that starts the window.\n"
+            "2. \"end\"        (numeric, optional) The height of the block that ends the window (default: current tip).\n"
+            "3. \"plotvalues\"  (string,  optional) Values to plot (comma separated), default(all): " + str_allowed_plot_values +
+            "\nResult:\n"
+            "{\n"
+            "}\n"
+            "\nExamples:\n"
+            + HelpExampleCli("getperblockstats", "1000 1000 \"minfeerate,avgfeerate\"")
+            + HelpExampleRpc("getperblockstats", "1000 1000 \"maxfeerate,avgfeerate\"")
```
jtimon: One is HelpExampleCli and the other is HelpExampleRpc

promag: Ops sorry, overlooked it.

src/rpc/blockchain.cpp
jtimon: duh, I was so much over-complicating things so much for no good reason...thank you!

jtimon:
Thanks again for the great feedback!

@promag I think I solved all your nits except for https://github.com/bitcoin/bitcoin/pull/10757#discussion_r126261101

@clarkmoody I think I added most of your suggestions, explicitly excluding anything that involved accumulations neither from height=1 nor from height=start.
The former potentially implies a world of complexity and the latter can be trivially calculated on the visual side: I would completely discard any accumulator redundancy in this rpc beforehand.

And for the rest of the redundancies, @jnewbery and @clarkmoody - thanks again for pointing it out -, it's never too late to remove them before merging like a trivial squash and it's never too soon to start saying which ones you would bikesay* out first. Also bikesay the names for the curves and even the order in the list (duplicated for c++ and python).

In the meantime, I embraced redundancy since, as said, it will be trivial for me to remove later. And also the pertinent optimizations to skip calculations when plot_values.count("minfee") == 0 or actually only when the extra calculation is more expensive than the searching in plot_values which is a set of strings.

For example, we have blockfees, reward, subsidy, complying with consensus rule ```reward == blockfees + subsidy```. Only 2 of the 3 are necessary, at least one is redundant. My personal preference is removing either subsidy or reward or subsidy, but not blockfees. But at said once written there's no problem with me in just making sure their tests don't surprise me until we decide which ones didn't deserve it. 

Which one seems bikesaying in principle. But not in this case. 
blockfees/total_fees serves for other calculations like avgfeerate. Let's not remove that one, just rename it.

But it is more interesting to propose new ones than to rename or vote for removal IMO. I believe the most interesting addition to this point was utxo_size_inc, which would welcomed some review from people who measures sizes more carefully like @sipa , since this doesn't use GetSerializeSize for Coin intentionally, independently of the optimization to read Coin if available in the utxo before calling RpcGetTx. I'm still not sure what to do with pre/post segwit feerates, does anybody care about the pre ones? which one needs the scale factor? none?

REM CalculateTruncatedMedian doesn't need to be a template at this point, but there's no harm being static IMO

promag:
@jtimon no problem. There are some nits to fix but I'll review more in depth later.

src/rpc/blockchain.cpp
```diff
@@ -1532,6 +1539,262 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+    } else if (size % 2 == 0) {
+        median = (scores[size / 2 - 1] + scores[size / 2]) / 2;
+    } else {
+        median = scores[size / 2];
+    }
+    return median;
+}
+
+// outpoint (needed for the utxo index) + nHeight + fCoinBase
+static const size_t PER_UTXO_OVERHEAD = sizeof(COutPoint) + sizeof(uint32_t) + sizeof(bool);
```
promag: constexpr.

src/rpc/blockchain.cpp
```diff
@@ -1532,6 +1539,262 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+    CAmount maxfeerate_old = 0;
+    std::vector<CAmount> fee_array;
+    std::vector<CAmount> feerate_old_array;
+    std::vector<CAmount> feerate_array;
+
+    CBlock block;
+    ReadBlockCheckPruned(pindex, block);
+
+    for (const auto& tx : block.vtx) {
+
```
promag: Remove empty line.

src/rpc/blockchain.cpp
```diff
@@ -1532,6 +1539,262 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+    } else {
+        plot_values = valid_plot_values;
+    }
+
+    std::map<std::string, UniValue> map_stats;
+    for (const std::string& plot_value : plot_values) {
+        map_stats[plot_value] = UniValue(UniValue::VARR);
+    }
+
+    CBlockIndex* pindex = chainActive[end];
```
promag: Remove.

src/rpc/blockchain.cpp
```diff
@@ -1532,6 +1539,262 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+    for (int i = end; i >= start; i--) {
+        UpdateBlockStats(pindex, plot_values, map_stats);
```
promag:
```cpp
UpdateBlockStats(chainActive[i], ...);
```

src/rpc/blockchain.cpp
```diff
@@ -1532,6 +1539,262 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+        "avgfee",
+    };
+    if (request.fHelp || request.params.size() < 1 || request.params.size() > 4)
+        throw std::runtime_error(
+            "getblockstats ( nStart nEnd plotValues )\n"
+            "\nCompute per block statistics for a given window.\n"
+            "\nArguments:\n"
+            "1. \"start\"      (numeric, required) The height of the block that starts the window.\n"
+            "2. \"end\"        (numeric, optional) The height of the block that ends the window (default: current tip).\n"
+            "3. \"plotvalues\"  (string,  optional) Values to plot (comma separated), default(all): " + boost::join(valid_plot_values, ",") +
```
promag: Replace `plotvalues` with `stats`? Also, 3rd argument could be object `options`?

src/rpc/blockchain.cpp
```diff
@@ -1532,6 +1539,262 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+
+    int start = request.params[0].get_int();
+    if (start < 1 || start > chainActive.Height()) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Start block height out of range");
+    }
+
+    int end;
+    if (request.params.size() > 1) {
+        end = request.params[1].get_int();
+        if (end < 0 || end > chainActive.Height()) {
```
promag: Nit, negative block could mean `end = height - end` to avoid early blocks (not new concept here I believe)?

src/rpc/blockchain.cpp
```diff
@@ -685,6 +686,22 @@ UniValue getblockheader(const JSONRPCRequest& request)
     return blockheaderToJSON(pblockindex);
 }
 
+static void ReadBlockCheckPruned(const CBlockIndex* pblockindex, CBlock& block)
```
jtimon: Not that I know of, but your proposed change sounds good to me.

src/rpc/blockchain.cpp
jtimon: what's the gain? https://stackoverflow.com/a/41132221/935325 says it's the same...

src/rpc/blockchain.cpp
jtimon: This will be slightly less efficient, no?

src/rpc/blockchain.cpp
jtimon: Mhmm, interesting. To be clear you mean start=-10 end=200 would be equivalent to start=190 end=200, right?

src/rpc/blockchain.cpp
jtimon: Isn't the string simpler?

src/policy/feerate.cpp
```diff
@@ -20,20 +20,22 @@ CFeeRate::CFeeRate(const CAmount& nFeePaid, size_t nBytes_)
         nSatoshisPerK = 0;
 }
 
-CAmount CFeeRate::GetFee(size_t nBytes_) const
+CAmount CFeeRate::GetTruncatedFee(size_t nBytes_) const
```
jtimon: Well, yes, conserving the old name only saves 1 line of extra disruption. But I guess if we're touching the variable name we should use the new style. just bytes?

src/rpc/blockchain.cpp
promag: No, I meant negative values are relative to the tip. To get the stats for the last 10 blocks you would pass start = -10 without querying the current block height.

src/rpc/blockchain.cpp
promag: Ignore `options` suggestion.

src/rpc/blockchain.cpp
promag: I guess it takes few more cycles but non critical code should be cleaner?

jtimon:
Here are some images generated using this branch in combination with (WIP): https://github.com/jtimon/rpc-explorer

screenshot_plotter
screenshot_plotter_zoom
plotter1
plotter2
plotterfees
plotter_utxo

src/rpc/blockchain.cpp
jtimon: Maybe. I'll wait for someone else to weight in since it doesn't seem a big deal either way.

jtimon:
Without the documentation for the result it was impossible to distinguish a weird choice to spring discussion from an implementation mistake. Removed the other TODO comments.
Coded more pending suggestions by @promag (hopefully all pending ones? if not, please insist) with some extra bikeshedding derived from s/plotvalues/stats/ and adapt tests to start and end being allowed to be negative.

More cleanups can be done, specially in the tests if we go further with https://github.com/bitcoin/bitcoin/pull/10757#discussion_r126462704 and not calculate in inverse order (there's no point if we don't get the slight optimization).

jtimon:
Needed rebase.
If somebody made a web for it, it may be interesting to show number of segwit txs too http://segwit.5gbfree.com/countsegwit

forklol:
Just wanted to say that this would be massively helpful to track statistics. I hope this finds it's way into a release soon.

jtimon:
Reversed the order of the values to the natural one, since as discussed the optimization of doing fetching the blocks in reverse order is not worth the loss in clarity of the code.
Added segwit tx counter stat, and also the total size and weight for those txs (txs that at least have one sw input, txs sending to sw outputs don't count).

trippysalmon:
Perhaps a better name for \<stat\>_old is \<stat\>_virtual, _virt or _v. Or perhaps prepend it with "v"  just like the tx size in the output of `getrawtransaction` (vsize).

For example:

`avgfeerate_old` becomes `vavgfeerate`
`maxfeerate_old` becomes `vmaxfeerate`
`medianfeerate_old` becomes `vmedianfeerate`

etc.

trippysalmon:
I just finished calling `getblockstats` on every block in the chain and saving it into a database. I didn't encounter any issues and the performance is quite good (100-1000ms per "full" block on an i7 7900k /w 32gb ram + nvme ssd).

Btw, if anyone is interested in the dataset I can share it. Just convo me at freenode irc (nick: "trippysalmon"). It includes some other stats as well, like rolling average hashrates.

src/rpc/blockchain.cpp
```diff
@@ -1531,6 +1538,313 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+            utxo_size_inc -= GetSerializeSize(prevoutput, SER_NETWORK, PROTOCOL_VERSION) + PER_UTXO_OVERHEAD;
+        }
+        CAmount txfee = tx_total_in - tx_total_out;
+        assert(MoneyRange(txfee));
+        fee_array.push_back(txfee);
+        totalfee += txfee;
+        minfee = std::min(minfee, txfee);
+        maxfee = std::max(maxfee, txfee);
+
+        CAmount feerate_old = CFeeRate(txfee, tx_size).GetTruncatedFee(1);
```
sipa: For feerate you should use the virtual size, not total size.

jlopp:
Just noticed this PR; I'll definitely be incorporating it into Statoshi once it's merged! :+1:

src/rpc/blockchain.cpp
jtimon: feerate is using vsize, feerate_old is using old size. Perhaps we just want to remove the whole *_old family.

jtimon:
@trippysalmon I think you mean replacing s/<stat>feerate/v<stat>feerate/ and s/<stat>feerate_old/<stat>feerate/.
As discussed with @sipa the *feerate_old are using old size and the *feerate ones are using vsize.

How much interest there is in maintaining the old ones?
Pre-segwit both are identical and post segwit the old ones mean the how high the feerate would have been for mempool and mining purposes if you weren't using segwit.
For example, the same tx can have feerate 4 sat/vbyte (new) and 2 sat/byte (old), meaning for the same fee, your tx gets propagated/mined as if you had paid twice as much thanks to segwit's discount.

If there's not much interest, perhaps it's just better to just remove all the old ones.
By the way, I said before that we hold on removing redundant or uninteresting stats.
I would start with reward, which the caller can calculate by simply adding subsidy and totalfees.

Needs rebase.

jtimon:
Rebased

trippysalmon:
@jtimon ah yes, I got it now. I misinterpreted the meaning of the `_old` statistics.

In that case the `_old` statistics are indeed not that interesting. I would however like to see a pre segwit `total_size` statistic. That one is currently missing and it's kind of a hassle to calculate it through RPC calls.

I will update my pre segwit total_size [PR](https://github.com/jtimon/bitcoin/pull/9) soon.

jtimon:
@trippysalmon not sure what you mean by "pre segwit total size". Total size is included, but size it's size post and pre segwit. You mean you want a stat for total vsize ? Not sure that's very interesting...

trippysalmon:
@jtimon yes I'm talking about total_vsize. Perhaps it depends on my particular use-case (comparing and graphing total_size vs total_vsize). My reasoning is that it is easy to add and compute inside core but not so much through RPC calls.

If you think total_vsize is not useful/interesting to others I don't mind maintaining it in a custom patch. I'm actually already using it in a project atm.

jtimon:
Removed reward and *feerate_old as discussed, but didn't squashed just in case.
Discussing with @trippysalmon we thought could show vsize instead of weight, but the former is just the latter / 4 (ie WITNESS_SCALE_FACTOR) and presenting the weight we can completely forget about rounding concerns (plus that's what getblock presents too, perhaps that's a reason to exclude it here).

jtimon:
Using this branch more, at first a range of heights was convenient but I was ready to wait even if I had to call this rpc block by block.
Right now, that's what I'm doing since once you start caching, calling this function is never a success but a cache failure. It is completely understandable that one doesn't want to manage failure ranges when failures can be discovered individually and thus processed in parallel.
That's probably opinionated, but perhaps others trying to use this got many errors forgetting "[0]" in some places.

Supporting ranges may be an optimization for users, but since I have been lucky enough to attract some potential users of this rpc call, I can ask: will anybody miss the height ranges?

Since I'm indexing my cache by block height and that's inherently "reorg unfriendly" I mostly see 2 options going forward:

1) Subscribe to the zmq interface, detect reorgs and remove block stats above the reorg height (while at it, one can put new blocks in the cache preemtively even if nobody asked for them yet)

2) Stop indexing the block stat cache by  height, do it by block hash. This is compatible with reorgs (more space but less disk writing and it's also required for advanced features like plotting abandoned/orphan chains).

I am inclined to trying both as a user since I want to subscribe to blocks but I also don't want to erase blocks I have seen and don't want to lose the option to chart reorged chains.

A second question is simply, assuming the height ranges are removed, would anyone else apart from me use the option to search by single block hash instead of single height?

Anyway, that would be my preference: replace height ranges with single height or single block height (both options).
Please let me know what you think.

robvanbentem:
I use this RPC call extensively (and persist the results in a DB) and never used the block height range, only single blocks.

> Anyway, that would be my preference: replace height ranges with single height or single block height (both options).

I think you meant block hash as the second option? For my use-case it doesn't really matter if I have to use the block height or block hash. Either option works for me.

--

Also, something that I needed today was the median transaction size. I don't think it can be calculated using the existing statistics or am I wrong?

It's only a [minor change](https://github.com/trippysalmon/bitcoin/commit/4a1d4d06e4c20356fe0e4359a3af2acfda06b1fc) and could be interesting to track over time to see it change when more complex scripts are made available.

trippysalmon:
I use this RPC call extensively (and persist the results in a DB) and never used the block height range, only single blocks.

> Anyway, that would be my preference: replace height ranges with single height or single block height (both options).

I think you meant block hash as the second option? For my use-case it doesn't really matter if I have to use the block height or block hash. Either option works for me.

--

Also, something that I needed today was the median transaction size. I don't think it can be calculated using the existing statistics or am I wrong?

It's only a [minor change](https://github.com/trippysalmon/bitcoin/commit/4a1d4d06e4c20356fe0e4359a3af2acfda06b1fc) and could be interesting to track over time to see it change when more complex scripts are made available.

jtimon:
Rebased, squashed, added "mintxsize", "maxtxsize", "mediantxsize" and "avgtxsize" stats.

Regarding the ranges, thanks for the feedback, I'll wait to hear from other people.

src/rpc/blockchain.cpp
```diff
@@ -1564,6 +1571,309 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
     return ret;
 }
 
+static void RpcGetTx(const uint256& hash, CTransactionRef& tx_out)
+{
+    uint256 hashBlock;
+    if (!GetTransaction(hash, tx_out, Params().GetConsensus(), hashBlock, true)) {
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string(fTxIndex ? "No such mempool or blockchain transaction"
```
ajtowns: Shouldn't this be GetTransaction(..., false) ?

src/rpc/blockchain.cpp
```diff
@@ -1570,6 +1571,309 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+    for (const std::string& stat : stats) {
+        // Update map_stats
+        if (stat == "height") {
+            map_stats[stat].push_back((int64_t)pindex->nHeight);
+        } else if (stat == "time") {
+            map_stats[stat].push_back(pindex->GetBlockTime());
+        } else if (stat == "mediantime") {
+            map_stats[stat].push_back(pindex->GetMedianTimePast());
+        } else if (stat == "subsidy") {
+            map_stats[stat].push_back(GetBlockSubsidy(pindex->nHeight, Params().GetConsensus()));
```
ajtowns: I think the values, fees and fee rates should be wrapped with ValueFromAmount, so outputs are decimal bitcoins rather than integer satoshis, for consistency with other RPC calls.

test/functional/getblockstats.py
```diff
@@ -0,0 +1,187 @@
...
+        ]
+        assert_contains(stats, all_values)
+        # Make sure all valid statistics are included
+        assert_contains(all_values, stats.keys())
+
+        assert_equal(stats['height'][0], start_height)
+        assert_equal(stats['height'][max_stat_pos], start_height + max_stat_pos)
+
+        assert_equal(stats['txs'][0], 1)
+        assert_equal(stats['swtxs'][0], 0)
```
ajtowns: None of the transactions have witness data, so this isn't actually checking whether swtxs, swtotal_size, swtotal_weight ever output anything other than 0, or whether weight is ever anything other than size*4.

ajtowns (commented):
Suggestions at https://github.com/jtimon/bitcoin/pull/10

I think requesting by height/blockhash makes more sense than a height range. (I was expecting a range to give me aggregate statistics over the entire range before looking into it)

src/rpc/blockchain.cpp
jtimon: No, we want to allow the slow fetch too. If the data is calculable we want to provide it no matter if more slowly. This rpc call shouldn't be expected to be fast anyway.

test/functional/getblockstats.py
jtimon: This is true. When I wrote this segwit wasn't activated, but now it would be nice to add segwit txs to the test because otherwise this functionality isn't really being tested. Good call.

src/rpc/blockchain.cpp
ajtowns: I'm presuming this reply should be to the "GetTransaction(..., false)" comment. Getting fee info without txindex seems like a weird thing to do to me, but if it's supported then presumably the getblockstats RPC help text doesn't need to still say "It won't work .. without -txindex." ?

jtimon: Yeah, sorry. Was meant for the other comment. Actually looking at the code again the slow search will only find txs in the utxo, so using false should be fine. I'll test it with false.

Regarding using BTC instead of satoshis, I think we want to move everything to satoshis but we don't do it because it would be too disruptive, so we only do it with new calls.
The caller can trivially divide by 100000000 to get BTC if they want t show that.

src/rpc/blockchain.cpp
```diff
@@ -1570,6 +1571,286 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+            + HelpExampleCli("getblockstats", "1000 \"minfeerate,avgfeerate\"")
+            + HelpExampleRpc("getblockstats", "1000 \"maxfeerate,avgfeerate\"")
+        );
+
+    LOCK(cs_main);
+
+    int height = request.params[0].get_int();
+    int current_tip = chainActive.Height();
+    if (height < 0) {
+        height = current_tip + height;
```
ryanofsky: In commit "RPC: Introduce getblockstats":

Bug: This should be `current_tip + height + 1` so height -1 will return tip.

test/functional/getblockstats.py
```diff
@@ -0,0 +1,186 @@
...
+#
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import (
+    assert_equal,
+    assert_raises_rpc_error,
+)
+
+def assert_contains(data, values, check_cointains=True):
+    for val in values:
+        if (check_cointains):
```
ryanofsky: In commit "QA: Test new getblockstats RPC"

Unneeded parentheses

test/functional/getblockstats.py
```diff
@@ -0,0 +1,186 @@
...
+            assert(val in data)
```
ryanofsky: In commit "QA: Test new getblockstats RPC"

Unneeded parentheses (assert is not a function)

test/functional/getblockstats.py
```diff
@@ -0,0 +1,186 @@
...
+            "medianfeerate",
+            "avgfeerate",
+            "mintxsize",
+            "maxtxsize",
+            "mediantxsize",
+            "avgtxsize",
+        ]
+        assert_contains(stats[0], all_values)
+        # Make sure all valid statistics are included
+        assert_contains(all_values, stats[0].keys())
```
ryanofsky: In commit "QA: Test new getblockstats RPC"

Seems like it would be simpler to combine the two asserts

```python
assert_equal(set(all_values), set(stats[0]))
```

test/functional/getblockstats.py
```diff
@@ -0,0 +1,186 @@
...
+
+        node.sendtoaddress(address=node.getnewaddress(), amount=10, subtractfeefromamount=True)
+        node.sendtoaddress(address=node.getnewaddress(), amount=10, subtractfeefromamount=True)
+        self.nodes[1].sendtoaddress(address=node.getnewaddress(), amount=1, subtractfeefromamount=True)
+        self.sync_all()
+        node.generate(1)
+
+        start_height = 101
+        max_stat_pos = 2
+        stats = []
```
ryanofsky: In commit "QA: Test new getblockstats RPC"

Could simplify with list comprehension:

```python
stats = [node.getblockstats(height=start_height + i) for i in range(max_stat_pos+1)]
```

test/functional/getblockstats.py
```diff
@@ -0,0 +1,186 @@
...
+            'minfee,%s,maxfee' % inv_sel_stat,
+        ]
+        for inv_stat in inv_stats:
+            assert_raises_rpc_error(-8, 'Invalid selected statistic %s' % inv_sel_stat, node.getblockstats, height=1, stats=inv_stat)
+        # Make sure we aren't always returning inv_sel_stat as the culprit stat
+        assert_raises_rpc_error(-8, 'Invalid selected statistic aaa%s' % inv_sel_stat, node.getblockstats, height=1, stats='minfee,aaa%s' % inv_sel_stat)
+
+        # Make sure only the selected statistics are included
+        stats = node.getblockstats(height=1, stats='minfee,maxfee')
+        some_values = [
```
ryanofsky: In commit "QA: Test new getblockstats RPC"

Again, seems like it would be simpler to just check for equality:

```python
stats = node.getblockstats(height=1, stats='minfee,maxfee')
assert_equal(set(stats), {"minfee", "maxfee"})
```

ryanofsky (commented):
utACK 44dcea038b44890e5f49c71ff26cfcb4ecf0da15. Left some comments but all are minor (feel free to ignore them) except the height + 1 bug.

src/rpc/blockchain.cpp
```diff
@@ -1570,6 +1571,286 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+    };
+
+    if (request.fHelp || request.params.size() < 1 || request.params.size() > 4)
+        throw std::runtime_error(
+            "getblockstats ( nStart nEnd stats )\n"
+            "\nCompute per block statistics for a given window. All amounts are in satoshis.\n"
+            "\nIt won't work in some cases with pruning or without -txindex.\n"
+            "\nArguments:\n"
+            "1. \"height\"     (numeric, required) The height of the target block.Negative values count back from the current tip.\n"
+            "2. \"stats\"      (string,  optional) Values to plot (comma separated), default(all): " + boost::join(valid_stats, ",") +
```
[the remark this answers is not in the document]
jtimon: I don't know. I don't mind moving to a list of strings if that's preferred. Makes sense.

src/rpc/blockchain.cpp
```diff
@@ -1570,6 +1571,286 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+            map_stats[stat] = CalculateTruncatedMedian(txsize_array);
+        } else if (stat == "avgtxsize") {
+            map_stats[stat] = (block.vtx.size() > 1) ? total_size / (block.vtx.size() - 1) : 0;
+        }
+    }
+}
+
+UniValue getblockstats(const JSONRPCRequest& request)
+{
+    std::set<std::string> valid_stats = {
```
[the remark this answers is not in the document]
jtimon: Some statistics are already more expensive to compute than others, it seems like a good feature to allow the user to optionally skip some from the beginning.

src/rpc/blockchain.cpp
```diff
@@ -1570,6 +1571,286 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+        return (scores[size / 2 - 1] + scores[size / 2]) / 2;
+    } else {
+        return scores[size / 2];
+    }
+}
+
+// outpoint (needed for the utxo index) + nHeight + fCoinBase
+static const size_t PER_UTXO_OVERHEAD = sizeof(COutPoint) + sizeof(uint32_t) + sizeof(bool);
+
+static void UpdateBlockStats(const CBlockIndex* pindex, std::set<std::string>& stats, std::map<std::string, UniValue>& map_stats)
```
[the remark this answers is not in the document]
jtimon: Yeah, it made sense to be update when the call did a loop from start height to end height, but not now that it calculates stats for a single block.

src/rpc/blockchain.cpp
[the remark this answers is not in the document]
jtimon: Oh, but it is currently calculating everything regardless of the stats requested...it wasn't like that before, I should re-introduce the optimzations for the option to really make sense.
Even if not all of them are expensive, perhaps one caller is only interested in one or a few stats and if the caller is going to make many calls for many heights, the performance hit of any undesired stat adds up.

jtimon:
Fixed all nits by @ryanofsky except for https://github.com/bitcoin/bitcoin/pull/10757#discussion_r148039141 . Instead of solving it, I added a few new commits for people to discuss on.

The tests still can't test size or feerates while passing/failing in a deterministic fashion, but just by rebasing now "swtxs" seems to be actually tested without me doing anything (probably by https://github.com/bitcoin/bitcoin/pull/11403 being merged), as requested in  https://github.com/bitcoin/bitcoin/pull/10757#discussion_r145659877

I thought of a simpler way to deterministacally get sizes and feerates: simply force the desired/expected size for each of the few txs this test creates. Using something like:

```
while not desired_size(tx, desired_size)
     tx = try_creating_tx_again()
```

or something of the sort. 

I'll try that next, but please comment on the latest commits.

jtimon:
This needs rebase, but some feedback on the latest things and potential squashes would be nice before doing so.

ajtowns:
What do you think about using a static blockchain so that testing is deterministic? I added some code to save the generated chain, and to reuse it if it's present and it doesn't seem too bad -- just using getblock(__,0) to get hex encoded blocks, and dumping them gives a 56kB file (or 10kB gzipped). That seems within the ballpark of something that could easily just be added as a static data file to the test suite?

I don't think `while tx_isnt_what_i_want: try_again()` is a good approach -- it seems a bit too easy to accidentally become `while i_trigger_a_bug: try_again()` which kind-of defeats the point of testing.

I'm not a huge fan of the `for(stat:stats) { if (stat == "x") m[stat] = x(); else if ...; }` approach. What about something along the lines of:

```
    auto set_stat = [&](const char* stat, const UniValue val)
                    { if (stats.count(stat) != 0) map_stats[stat] = val; };

    set_stat("height", (int64_t)pindex->nHeight);
    set_stat("time", pindex->GetBlockTime());
    set_stat("mediantime", pindex->GetMedianTimePast());
    ...
```
? I don't think any of the calculations at that point are particularly heavy, so it shouldn't make things much slower (and any that are could probably just be put in an `if (is_loop_outputs_required) { .. }` block or similar).

I think the optimisation commits make sense. Maybe add bools for `medianfeerate`, `medianfee`, `mediantxsize` so you're not calling `.count()` inside the loops. The naming of the bools is annoyingly ungrammatical ("is_loop_inputs_required" should be "are_loop_inputs_required" in english), but I don't have a better suggestion.

Might be good to explicitly say which stats want -txindex enabled and won't work with pruning in the help (fee related and utxo size I think?) Could make the default be "whatever stats we can output efficiently" rather than "everything", though not sure that's a good idea.

jtimon:
> What do you think about using a static blockchain so that testing is deterministic? I added some code to save the generated chain, and to reuse it if it's present and it doesn't seem too bad -- just using getblock(__,0) to get hex encoded blocks, and dumping them gives a 56kB file (or 10kB gzipped). That seems within the ballpark of something that could easily just be added as a static data file to the test suite?

I didn't found the time to try that but if you can share that code that would be great. What you're saying seems reasonable to me, I don't know if other people would have a problem with that. And I cannot think of a simpler way to make this test deterministic.

I will look into the other suggestions, thanks.

jtimon:
@MarcoFalke what is the status of this in terms of testing?

I think @ajtowns is right that *size, and *feerate fields can only be deterministically tested, but I don't have the time to write that at this point. I would be happy to review and test if anybody wants to take over though.

In the meantime, I would be happy to include the new call even if it's without any stat that depends on signature size or coin selection for deterministic testing and those fields and their tests are already written.

jtimon:
Needed rebase. And still needs squashing pending on requested feedback.

Also, since testing some of the stats deterministically is not trivial, I think the new rpc call should be introduced only with the stats that can be deterministacally tested (even though we will miss some of the most interesting ones like sizes, fees and feerates) and then make the tests deterministic and add the missing stats.
Thoughts?

ajtowns:
Proposed patch that makes testing of the stats kind of deterministic at https://github.com/jtimon/bitcoin/pull/11/commits

Idea is you do `./feature_rpc_getblockstats --gen-test-data` to  generate a test blockchain and the expected statistics for that blockchain, and will write it to data/rpc_getblockstats.json file (about 50kB uncompressed text). You then check the stats look reasonable, and commit that file to git, and then when you/travis runs the test case without --gen-test-data it will load the chain from the file and check that the calculated stats still match what was recorded as being expected.

jtimon:
Thanks a lot, incorporated those changes. Fixed some older nits too.
Please re-review.

src/rpc/blockchain.cpp
```diff
@@ -1612,6 +1613,304 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
     return ret;
 }
 
+static CTransactionRef GetTransactionChecked(const uint256& hash)
+{
+    CTransactionRef tx_out;
+    uint256 hashBlock;
+    if (!GetTransaction(hash, tx_out, Params().GetConsensus(), hashBlock, false)) {
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string(fTxIndex ? "No such mempool or blockchain transaction"
```
jimpo: Surfacing this error seems like the the wrong error handling strategy. In the context this is used, the transaction should never be in the mempool and it should *always* be found if the txindex is present.

I think it would be better for the RPC to raise an RPC error that the txindex is required if it is not present, and an RPC error for an unexpected internal error (txindex is corrupt) otherwise.

test/functional/rpc_getblockstats.py
```diff
@@ -0,0 +1,180 @@
...
+#
+# Test getblockstats rpc call
+#
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import (
+    assert_equal,
+    assert_raises_rpc_error,
+)
+
+def assert_contains(data, values, check_cointains=True):
```
jimpo: commit: Tests: Test new getblockstats RPC

typo: check_co**i**ntains

test/functional/rpc_getblockstats.py
```diff
@@ -0,0 +1,180 @@
...
+        "avgfeerate",
+        "mintxsize",
+        "maxtxsize",
+        "mediantxsize",
+        "avgtxsize",
+    ]
+
+    def set_test_params(self):
+        self.num_nodes = 2
+        self.extra_args = [['-txindex'], ['-paytxfee=0.003']]
```
jimpo: commit: Tests: Test new getblockstats RPC

I think it's worth testing the case of no txindex.

test/functional/rpc_getblockstats.py
jimpo: commit: Tests: Test new getblockstats RPC

The `assert_contains` function appears unused.

test/functional/rpc_getblockstats.py
```diff
@@ -70,90 +86,63 @@ def run_test(self):
...
+        else:
+            self.load_test_data(test_data)
+
+        node = self.nodes[0]
+        stats = self.get_stats()
 
         # Make sure all valid statistics are included but nothing else is
         assert_equal(set(stats[0].keys()), set(self.EXPECTED_STATS))
 
-        print(stats)
```
jimpo: commit: Tests: Save and load block and corresponding expected statistics

If all these test cases are getting removed anyway, I'd just drop them from the previous commit.

test/functional/rpc_getblockstats.py
```diff
@@ -70,90 +86,63 @@ def run_test(self):
...
+        tip = node.getbestblockhash()
+        blkhash = None
+        height = 0
+        while tip != blkhash:
+            blkhash = node.getblockhash(height)
+            blocks.append(node.getblock(blkhash, 0))
+            height += 1
+
+        with open(filename, "w") as f:
+            f.write("""{\n  "stats": """)
```
jimpo: commit: Tests: Save and load block and corresponding expected statistics

Instead of hand-crafting the JSON, I think it's better to just create a dict with `stats` and `blocks` keys and json.dump that. It seems you might be doing this for better control over the indentation, but the data file added in the next commit has each stat on its own line.

ajtowns: It puts the stats on separate lines, but all the blocks on a single line. Seemed like a good idea at the time (multiline stats hand editable, block data as a minimal blob), but doesn't make much difference.

src/rpc/blockchain.cpp
```diff
@@ -1612,6 +1613,303 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+    const bool loop_outputs = stats.count("total_out") != 0 || stats.count("utxo_size_inc") != 0 ||
+        stats.count("totalfee") != 0 || stats.count("avgfee") != 0 || stats.count("avgfeerate") != 0 ||
+        do_medianfee || stats.count("minfee") != 0 || stats.count("maxfee") != 0 ||
+        do_medianfeerate || stats.count("minfeerate") != 0 || stats.count("maxfeerate") != 0;
+
+    const bool do_calculate_size = do_mediantxsize || stats.count("total_size") != 0 || stats.count("avgtxsize") != 0 ||
+        stats.count("mintxsize") != 0 || stats.count("maxtxsize") != 0 || stats.count("swtotal_size") != 0;
+
+    const bool do_calculate_weight = stats.count("total_weight") != 0 || stats.count("avgfeerate") != 0 ||
+        stats.count("swtotal_weight") != 0 || stats.count("medianfeerate_weight") != 0;
```
ajtowns: Should be `medianfeerate`, not `_weight`.

src/rpc/blockchain.cpp
```diff
@@ -1612,6 +1613,303 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+
+    const bool do_calculate_sw = stats.count("swtxs") != 0 || stats.count("swtotal_size") != 0 ||
+        stats.count("swtotal_weight") != 0;
+
+    const bool loop_inputs = stats.count("totalfee") != 0 || stats.count("avgfee") != 0 || stats.count("avgfeerate") != 0 ||
+        do_medianfee || stats.count("minfee") != 0 || stats.count("maxfee") != 0 ||
+        do_medianfeerate || stats.count("minfeerate") != 0 || stats.count("maxfeerate") != 0;
```
ajtowns: `loop_inputs` is also needed when `utxo_size_inc` is requested. Setting `loop_outputs = loop_inputs || stats.count("total_out")` might simplify things a little.

src/rpc/blockchain.cpp
```diff
@@ -1612,6 +1613,303 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+    }
+
+    const CBlockIndex* pblockindex = chainActive[height];
+    const CBlock block = GetBlockChecked(pblockindex);
+    std::map<std::string, UniValue> map_stats = GetBlockStatsMap(block, pblockindex, stats);
+
+    UniValue ret(UniValue::VOBJ);
+    for (const std::string stat : stats) {
+        ret.pushKV(stat, map_stats[stat]);
+    }
```
ajtowns: Could move the "Invalid requested statistic" check to here, ie:

```
    for (auto stat : stats) {
        if (map_stats.count(stat) == 0) {
            throw JSONRPCError(...);
        }
        ret.pushKV(stat, map_stats[stat]);
   }
```

Then you wouldn't need `valid_stats`. Downside is the error response would only come after the valid stats were corrected.

src/rpc/blockchain.cpp
```diff
@@ -1612,6 +1613,303 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+
+    const bool do_mediantxsize = stats.count("mediantxsize") != 0;
+    const bool do_medianfee = stats.count("medianfee") != 0;
+    const bool do_medianfeerate = stats.count("medianfeerate") != 0;
+
+    const bool loop_outputs = stats.count("total_out") != 0 || stats.count("utxo_size_inc") != 0 ||
+        stats.count("totalfee") != 0 || stats.count("avgfee") != 0 || stats.count("avgfeerate") != 0 ||
+        do_medianfee || stats.count("minfee") != 0 || stats.count("maxfee") != 0 ||
+        do_medianfeerate || stats.count("minfeerate") != 0 || stats.count("maxfeerate") != 0;
+
```
ajtowns: These initialisations are a bit hard to follow. What do you think of:

```
template<typename T> static inline bool SetHasKeys(const std::set<T>& set) {return false;}
template<typename T, typename Tk, typename... Args> static inline bool SetHasKeys(const std::set<T>& set, const Tk& key, const Args&... args)
{
    return (set.count(key) != 0) || SetHasKeys(set, args...);
}

    const bool loop_inputs = do_medianfee || do_medianfeerate ||
       SetHasKeys(stats, "totalfee", "avgfee", "avgfeerate", "minfee", "maxfee", "minfeerate", "maxfeerate", "utxo_size_inc");
```

instead? Seems easier to understand to me, should be equally efficient at runtime, and the recursive template thing is in use elsewhere already.
```

File: test/functional/rpc_getblockstats.py [hunk 35]
```diff
@@ -0,0 +1,180 @@
...
+            assert_raises_rpc_error(-8, 'Invalid requested statistic %s' % inv_sel_stat, node.getblockstats, height=1, stats=inv_stat)
+        # Make sure we aren't always returning inv_sel_stat as the culprit stat
+        assert_raises_rpc_error(-8, 'Invalid requested statistic aaa%s' % inv_sel_stat, node.getblockstats, height=1, stats=['minfee' , 'aaa%s' % inv_sel_stat])
+
+        # Make sure only the selected statistics are included
+        some_stats = {'minfee', 'maxfee'}
+        stats = node.getblockstats(height=1, stats=list(some_stats))
+        # Make sure only valid stats that have been requested appear
+        assert_equal(set(stats.keys()), some_stats)
+
```
ajtowns: Suggest adding:

```
       # Make sure each stat can be queried on its own
       for stat in self.EXPECTED_STATS:
           for i in range(self.max_stat_pos+1):
               result = node.getblockstats(height=self.start_height + i, stats=[stat])
               assert_equal(list(result.keys()), [stat])
               if result[stat] != self.expected_stats[i][stat]:
                   self.log.info("result[%s] (%d) failed, %r != %r" % (stat, i, result[stat], self.expected_stats[i][stat]))
               assert_equal(result[stat], self.expected_stats[i][stat])
```

to ensure that each statistic can be queried on its own (in particular ensuring the GetBlockStatsMap() doesn't optimise out collecting the data needed for each stat). The above should automatically catch problems if new stats are introduced since it's just re-using the results from the all-the-stats check.

ajtowns (commented):
I'm a bit confused by the bit about the x-axis in the PR description -- isn't plotting stuff in a different PR? 

Seems a bit strange to exclude the coinbase from total_out; there's no way to tell from the output you've got if the coinbase failed to claim all its reward.

src/rpc/blockchain.cpp
jtimon: Yeah, it could be a nice code simplification, but we would still need valid_stats for the default (or we would need to hardcode the default). Besides, I don't like the downside.
I'm not sure about this, but it's not a blocker, right?

src/rpc/blockchain.cpp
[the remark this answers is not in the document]
jtimon: Oops, gihub hid this reminder about the comment from me, fixing.

src/rpc/blockchain.cpp
jimpo: Do you know if fancy templating stuff like this will slow down compilation times noticeably?

src/rpc/blockchain.cpp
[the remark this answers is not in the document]
jtimon: I ended up taking the code suggestion, even though I'm not eliminating the "stats" argument.
It is still much closer to what you wanted now.

jtimon:
Fixed nits. The latest commit I did not squash because I'm not convinced it is an improvement even though is +32-48 (see https://github.com/bitcoin/bitcoin/pull/10757#discussion_r182942644 )

src/rpc/blockchain.cpp
```diff
@@ -1673,16 +1673,27 @@ static std::map<std::string, UniValue> GetBlockStatsMap(const CBlock& block, con
...
+    bool do_mediantxsize = true;
+    bool do_medianfee = true;
+    bool do_medianfeerate = true;
+    bool loop_inputs = true;
+    bool loop_outputs = true;
+    bool do_calculate_size = true;
+    bool do_calculate_weight = true;
+    bool do_calculate_sw = true;
+    // Calculate everything if nothing selected (default)
+    if (stats.size() > 0) {
```
ajtowns: Could do:

```
    const bool do_all = stats.size() > 0;
    const bool do_mediantxsize = do_all || SetHasKeys(stats, "mediantxsize);
    const bool do_medianfee = do_all || ...;
    ...
```

src/rpc/blockchain.cpp
```diff
@@ -1603,6 +1613,287 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+
+    LOCK(cs_main);
+
+    int height = request.params[0].get_int();
+    int current_tip = chainActive.Height();
+    if (height < 0) {
+        height = current_tip + height + 1;
+    }
+    if (height < 0) {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Target block height %d is negative", height));
```
ajtowns: There's two tests for `(height < 0)` here, the first of which should be setting height to a positive value. Seems like a rebasing error?

ajtowns:
I don't really have a strong opinion on `total_out`, I guess I just found it a little weird it doesn't include the coinbase outputs.

As far as height vs blockhash is concerned, I think specifying by height is probably easiest for people using the api, but maybe it would be good to include the blockhash in the output (so if you get different results for a given height due to a reorg, it's easy to tell why), and/or allow either height/hash as the first param? I don't feel strongly about this either way, either.

src/rpc/blockchain.cpp
jtimon: I think it is a rest from when moving from a range to a single height, but yeah, thanks.

jtimon:
Yeah, my concern is precisely that, asking for a given height and getting the results for a block it's not what you were expecting, and yeah, both solutions (optionally accepting hash instead of height or always returning the block hash work) solve that problem.
I guess I like more the first option because it's less data sent, but I guess the "hash_or_height" param name is ugly. I will do that and ask for "squash or remove"?

By the way, thoughts on the "squash or remove" for "?? f'RPC: Introduce getblockstats'"?
Is that what you were expecting? Oh, wait, you have a nit on that commit. I'll solve them and then ask again (unless you say "yeah it's not worth it" now and I skip solving that nit).

jtimon:
Since all the stats are optional anyway, I ended up doing both adding an optional blockhash in the result and my preference, which was moving from height to hash_or_height.
Also did some squashing and did further simplifications on both the rpc and the tests, including inling the function, since it didn't sense anymore after moving away from height ranges and removing the hardcoded list of expected stats from the tests (since they're already hardcoded in the data file).

Since the diff from last review was kind of big, I took the opportunity to rebase too.

jtimon:
Fixed the remaining nits: test when there's no txindex and give a specific error when txindex is required but not set as suggested by @jimpo.

jtimon:
Fixed the tests with the node 1 in travis (in my computer the sync_all wasn't needed, oops) and did some bikeshedding.

promag (commented):
Code looks good, some nits though. I'll try it out so I can give a tested ACK.

src/rpc/blockchain.cpp
```diff
@@ -1612,6 +1613,274 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+            "  \"total_out\": xxxxx,       (numeric) Total amount in all outputs (excluding coinbase and thus reward [ie subsidy + totalfee])\n"
+            "  \"total_size\": xxxxx,      (numeric) Total size of all non-coinbase transactions\n"
+            "  \"total_weight\": xxxxx,    (numeric) Total weight of all non-coinbase transactions divided by segwit scale factor (4)\n"
+            "  \"totalfee\": xxxxx,        (numeric) The fee total\n"
+            "  \"txs\": xxxxx,             (numeric) The number of transactions (excluding coinbase)\n"
+            "  \"utxo_increase\": xxxxx,   (numeric) The increase/decrease in the number of unspent outputs\n"
+            "  \"utxo_size_inc\": xxxxx,   (numeric) The increase/decrease in size for the utxo index (not discounting op_return and similar)\n"
+            "}\n"
+            "\nExamples:\n"
+            + HelpExampleCli("getblockstats", "1000 \"minfeerate,avgfeerate\"")
```
jimpo: CLI help doc is incorrect. It should be `"1000 '[\"minfeerate\",\"avgfeerate\"]'"` (once stats is added in client.cpp).

jimpo (commented):
Looks good aside from the CLI usage. I'd ACK with that change.

src/rpc/client.cpp
```diff
@@ -122,6 +122,7 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "importmulti", 1, "options" },
     { "verifychain", 0, "checklevel" },
     { "verifychain", 1, "nblocks" },
+    { "getblockstats", 0, "hash_or_height" },
```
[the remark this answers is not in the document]
jtimon: Mhmm, yeah. Why is this working without it?

jtimon:
Hopefully fixed the newest nits, thanks again.

jimpo:
ACK ddfe091

ajtowns:
Weird. I'm getting consistent failures with 2dffcdc2afd7ee98f170af933b94186c455425bc on the `sync_all()` after `load_test_data()` -- it just hangs, because node 1 is never actually seeing the blocks. If I add a `node.generate(1)` beforehand, things work. The problem seems to be that the node doing the `submitblock` calls is never leaving InitialBlockDownload, and while it's in IBD it's not relaying blocks to its peer. Ah, it looks like this only shows up when it's been more than 24 hours since `--gen-test-data` was run -- so rerunning travis or running it locally should fail reliably now, I think.

Adding a call to `node.generate(1)` at the start of `load_test_data` should fix this:

```
     def load_test_data(self, filename):
         node = self.nodes[0]
+        node.generate(1) # finish IBD; will get reorged out
```

jtimon:
@ajtowns I think I solved it with mocktime as discussed, but since the time is in the file generated again, I advice not to merge it after we test it again after 24 of the data file being created (last time it only started failing after that),

src/rpc/blockchain.cpp
```diff
@@ -1614,6 +1625,275 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+            assert(MoneyRange(txfee));
+            if (do_medianfee) {
+                fee_array.push_back(txfee);
+            }
+            maxfee = std::max(maxfee, txfee);
+            minfee = std::min(minfee, txfee);
+            totalfee += txfee;
+
+            // New feerate uses satoshis per virtual byte instead of per serialized byte
+            CAmount feerate = CFeeRate(txfee, weight).GetTruncatedFee(WITNESS_SCALE_FACTOR);
```
TheBlueMatt: I find this incredibly hard to read. Why not just use floating point and calculate it all manually? Then you wouldn't have to work around rounding issues (and wouldnt have the rounding issues you see here where things are rounded to the nearest 1).

test/functional/rpc_getblockstats.py
```diff
@@ -0,0 +1,175 @@
...
+            blocks = d['blocks']
+            mocktime = d['mocktime']
+            self.expected_stats = d['stats']
+
+        for b in blocks:
+            self.nodes[0].submitblock(b)
+
+        # Set the timestamps from the file so that the nodes can get out of Initial Block Download
+        self.nodes[0].setmocktime(mocktime)
+        self.nodes[1].setmocktime(mocktime)
```
ajtowns: Think mocktime needs to be set prior to `submitblock`

src/rpc/blockchain.cpp
```diff
@@ -1614,6 +1625,275 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+
+        pindex = chainActive[height];
+    } else {
+        const std::string strHash = request.params[0].get_str();
+        const uint256 hash(uint256S(strHash));
+        pindex = LookupBlockIndex(hash);
+    }
+
+    std::set<std::string> stats;
+    if (request.params.size() > 1 && !request.params[1].isNull()) {
```
TheBlueMatt: You dont need the size() check.

src/rpc/blockchain.cpp
```diff
@@ -1614,6 +1625,275 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+        }
+        if (height > current_tip) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, strprintf("Target block height %d after current tip %d", height, current_tip));
+        }
+
+        pindex = chainActive[height];
+    } else {
+        const std::string strHash = request.params[0].get_str();
+        const uint256 hash(uint256S(strHash));
+        pindex = LookupBlockIndex(hash);
```
TheBlueMatt: I believe if the hash is invalid GetBlockChecked will crash trying to deref the blockindex nullptr.

TheBlueMatt (commented):
Matt was here.

jtimon:
Fixed last 2 nits. Added a test for when blocks aren't found.
Also needed rebase.

test/functional/rpc_getblockstats.py
```diff
@@ -0,0 +1,181 @@
...
+                                self.nodes[0].getblockstats, hash_or_height=1, stats=['minfee' , 'aaa%s' % inv_sel_stat])
+
+        assert_raises_rpc_error(-8, 'One or more of the selected stats requires -txindex enabled',
+                                self.nodes[1].getblockstats, hash_or_height=self.start_height + self.max_stat_pos)
+
+        # Mainchain's genesis block shouldn't be found on regtest
+        assert_raises_rpc_error(-5, 'Block not found', self.nodes[0].getblockstats,
+                                hash_or_height='000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f')
+
+
```
promag: Linter error:
```
W293 blank line contains whitespace.
```

jtimon: Fixed

src/rpc/blockchain.cpp
```diff
@@ -1614,6 +1625,283 @@ static UniValue getchaintxstats(const JSONRPCRequest& request)
...
+        if (!pindex) {
+            throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Block not found");
```
kallewoof: Just a note here. In https://github.com/bitcoin/bitcoin/blob/4cfe17c3382ba750131cdc8703b2978132822070/src/rpc/blockchain.cpp#L809-L818 `RPC_MISC_ERROR` is used, while `RPC_INVALID_ADDRESS_OR_KEY` is used here. Not sure which is better, but I think they should both use the same error type. (This PR touches the other one so could easily tweak, I think.)

src/rpc/blockchain.cpp
```diff
@@ -1614,6 +1625,283 @@ static UniValue getchaintxstats(const JSONRPCRequest& request)
...
+            assert(MoneyRange(txfee));
+            if (do_medianfee) {
+                fee_array.push_back(txfee);
+            }
+            maxfee = std::max(maxfee, txfee);
+            minfee = std::min(minfee, txfee);
+            totalfee += txfee;
+
+            // New feerate uses satoshis per virtual byte instead of per serialized byte
+            CAmount feerate = weight ? (txfee * WITNESS_SCALE_FACTOR) / weight : 0;
```
kallewoof: May be ok, but I think `(txfee * WITNESS_SCALE_FACTOR) / (weight + WITNESS_SCALE_FACTOR - 1)` is precisely accurate while the above will have truncation errors sometimes.

kallewoof (approved):
utACK 563eee92c9ddc5b537fca1ce08bce811d97aed97 w/ minor nits

src/rpc/blockchain.cpp
jimpo: In #13144, I introduce the `RPC_DATA_UNAVAILABLE` error type, which is meant for cases like this.

src/rpc/blockchain.cpp
ajtowns: The fee rate for a tx with weight 3 and fee 100 is the same as if there were 100 such tx's. That would have weight 300 or vsize 75, and fee 10,000, for a fee rate of 10,000/75 or 133.3, not 1, 100, or 66. :) Formula looks fine to me, though I guess estimatefee uses fee/vkbyte rather than fee/byte, which would be `(txfee * WITNESS_SCALE_FACTOR * 1000) / weight`.

ajtowns (approved):
ACK 563eee92c9ddc5b537fca1ce08bce811d97aed97 modulo nits

src/rpc/blockchain.cpp
jtimon: Well, we can move from sat/vbyte to sat/vkbyte, sure. I t should be a simple change if people prefer that.

jtimon:
Fixed help nits (also made other improvements, some other things in the help were ugly).

PR #10769, opened 2017-07-07
[Qt] replace fee slider with a Dropdown, extend conf. targets

jonasschnelli:
Addresses #10590.

This PR replaces the fee slider with a Dropbox box. The Dropdown contains the target in ~blocks and ~estimated time to confirm.

The current supported confirmation targets are: 2, 4, 6, 12, 48, 144, 504, 1008

bildschirmfoto 2017-07-07 um 21 56 19

### DIFF
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -31,6 +31,25 @@
 #include <QTextDocument>
 #include <QTimer>
 
+static const std::array<int, 9> confTargets = { {2, 4, 6, 12, 24, 48, 144, 504, 1008} };
+int getConfTargetForIndex(int index) {
+    if (index+1 > static_cast<int>(confTargets.size())) {
+        return confTargets.back();
+    }
+    if (index < 0) {
+        return confTargets[0];
+    }
+    return confTargets[index];
+}
+int getIndexForConfTarget(int target) {
+    for (unsigned int i = 0; i < confTargets.size(); i++) {
+        if (confTargets[i] >= target) {
+            return i;
+        }
+    }
+    return confTargets.size() - 1;
+}
+
 SendCoinsDialog::SendCoinsDialog(const PlatformStyle *_platformStyle, QWidget *parent) :
     QDialog(parent),
     ui(new Ui::SendCoinsDialog),
@@ -152,9 +171,12 @@ void SendCoinsDialog::setModel(WalletModel *_model)
         coinControlUpdateLabels();
 
         // fee section
-        connect(ui->sliderSmartFee, SIGNAL(valueChanged(int)), this, SLOT(updateSmartFeeLabel()));
-        connect(ui->sliderSmartFee, SIGNAL(valueChanged(int)), this, SLOT(updateGlobalFeeVariables()));
-        connect(ui->sliderSmartFee, SIGNAL(valueChanged(int)), this, SLOT(coinControlUpdateLabels()));
+        for (const int &n : confTargets) {
+            ui->confTargetSelector->addItem(tr("%1 (%2 blocks)").arg(GUIUtil::formatNiceTimeOffset(n*Params().GetConsensus().nPowTargetSpacing)).arg(n));
+        }
+        connect(ui->confTargetSelector, SIGNAL(currentIndexChanged(int)), this, SLOT(updateSmartFeeLabel()));
+        connect(ui->confTargetSelector, SIGNAL(currentIndexChanged(int)), this, SLOT(updateGlobalFeeVariables()));
+        connect(ui->confTargetSelector, SIGNAL(currentIndexChanged(int)), this, SLOT(coinControlUpdateLabels()));
         connect(ui->groupFee, SIGNAL(buttonClicked(int)), this, SLOT(updateFeeSectionControls()));
         connect(ui->groupFee, SIGNAL(buttonClicked(int)), this, SLOT(updateGlobalFeeVariables()));
         connect(ui->groupFee, SIGNAL(buttonClicked(int)), this, SLOT(coinControlUpdateLabels()));
@@ -177,10 +199,10 @@ void SendCoinsDialog::setModel(WalletModel *_model)
 
         // set the smartfee-sliders default value (wallets default conf.target or last stored value)
         QSettings settings;
-        if (settings.value("nSmartFeeSliderPosition").toInt() == 0)
-            ui->sliderSmartFee->setValue(ui->sliderSmartFee->maximum() - model->getDefaultConfirmTarget() + 2);
+        if (settings.value("nConfTarget").toInt() == 0)
+            ui->confTargetSelector->setCurrentIndex(getIndexForConfTarget(model->getDefaultConfirmTarget()));
         else
-            ui->sliderSmartFee->setValue(settings.value("nSmartFeeSliderPosition").toInt());
+            ui->confTargetSelector->setCurrentIndex(getIndexForConfTarget(settings.value("nConfTarget").toInt()));
     }
 }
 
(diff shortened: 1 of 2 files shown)


jonasschnelli:
I probably need to mention why a dropbox and not a slider.

* Sliders tend to be linear (otherwise nobody would understand them and I bet it would require a custom, non standard UI object) which would make it useless if we would support conf. targets up to 1008.
* A dropbox allows one to give a better overview (slider: you only see the underlaying value if you change it).
* Dropbox would allow to show the feerate and or a text ("fast", "economy", etc,), ... and eventual the absolute fee by running the coin-selections if the parameters are provided.

src/qt/sendcoinsdialog.cpp
```diff
@@ -31,6 +31,25 @@
 #include <QTextDocument>
 #include <QTimer>
 
+static const std::array<int, 8> confTargets = { {2, 4, 6, 12, 48, 144, 504, 1008} };
```
promag: Rename `conf_targets`.

src/qt/sendcoinsdialog.cpp
```diff
@@ -31,6 +31,25 @@
...
+int getConfTargetForIndex(int index) {
+    if (index+1 > static_cast<int>(confTargets.size())) {
```
promag:
```cpp
if (index >= ...
```

promag: Or just:
```cpp
  return conf_targets[max(0, min(conf_targets.size() - 1, index)]
```

flack:
Looking at the screenshot, I would drop the tilde before the number of blocks, i.e. `2 Blocks, ~20 minutes` instead of `~2 Blocks, ~20 minutes`. It already says "confirmation target", so it should be obvious that it's not exactly x blocks. In other words, make it say "target x blocks (approx. y minutes)" instead of "target approx. x blocks (approx. y minutes)".

Also, shouldn't blocks have a lowercase b?

morcos:
Will review shortly.  Thanks for doing!

Also please add a target for 24 for compatibility with the old estimates (where 25 was the highest possible target).  It may be a target that people are used to using.

luke-jr:
Confirmation is when it's 6 blocks deep, so +5 to all the numbers here? Or change it to "begin confirmation" like we currently have...

gmaxwell:
I like how it looks, @morcos any concerns about missing 3,5 making users fee choices more clumpy?

jonasschnelli:
Re-added targets 3, 5, 24 and removed the tilde in front of the target-in-blocks/-time.

src/qt/sendcoinsdialog.cpp
```diff
@@ -177,10 +199,10 @@ void SendCoinsDialog::setModel(WalletModel *_model)
 
         // set the smartfee-sliders default value (wallets default conf.target or last stored value)
         QSettings settings;
-        if (settings.value("nSmartFeeSliderPosition").toInt() == 0)
-            ui->sliderSmartFee->setValue(ui->sliderSmartFee->maximum() - model->getDefaultConfirmTarget() + 2);
```
promag: Should call `settings.remove("nSmartFeeSliderPosition")`?

src/qt/sendcoinsdialog.cpp
```diff
@@ -152,9 +171,12 @@ void SendCoinsDialog::setModel(WalletModel *_model)
         coinControlUpdateLabels();
 
         // fee section
-        connect(ui->sliderSmartFee, SIGNAL(valueChanged(int)), this, SLOT(updateSmartFeeLabel()));
-        connect(ui->sliderSmartFee, SIGNAL(valueChanged(int)), this, SLOT(updateGlobalFeeVariables()));
-        connect(ui->sliderSmartFee, SIGNAL(valueChanged(int)), this, SLOT(coinControlUpdateLabels()));
+        for (const int &n : confTargets) {
+            ui->confTargetSelector->addItem(tr("%1 blocks, %2").arg(n).arg(GUIUtil::formatNiceTimeOffset(n*Params().GetConsensus().nPowTargetSpacing)));
```
promag: Alternative format: `"%2 (%1 blocks)"` since the label is `Confirmation time`.

src/qt/sendcoinsdialog.cpp
```diff
@@ -31,6 +31,25 @@
 #include <QTextDocument>
 #include <QTimer>
 
+static const std::array<int, 11> confTargets = { {2, 3, 4, 5, 6, 12, 24, 48, 144, 504, 1008} };
+int getConfTargetForIndex(int index) {
```
promag: Update block style. Same for `getIndexForConfTarget`.

src/qt/sendcoinsdialog.cpp
```diff
@@ -31,6 +31,25 @@
 #include <QTextDocument>
 #include <QTimer>
 
+static const std::array<int, 11> confTargets = { {2, 3, 4, 5, 6, 12, 24, 48, 144, 504, 1008} };
```
promag: Snake case.

morcos:
ACK 19eae9c

@gmaxwell Actually my preference would be to remove the 3 and 5 targets.  I think it gives a false sense of accuracy with the estimates.  I don't imagine much clumping concern as those estimates often aren't that different from each other and there is a much larger range of estimates to choose from as well as time variability.  I also think its a bit cleaner not to be faced with too many choices in a dropdown.

src/qt/sendcoinsdialog.cpp
TheBlueMatt: nit: not sure its worth having 5 here, maybe not 3 either.

src/qt/sendcoinsdialog.cpp
```diff
@@ -31,6 +31,25 @@
...
+    }
+    return confTargets[index];
+}
+int getIndexForConfTarget(int target) {
+    for (unsigned int i = 0; i < confTargets.size(); i++) {
+        if (confTargets[i] >= target) {
+            return i;
+        }
+    }
+    return confTargets.back();
```
TheBlueMatt: I think you meant .size() - 1.

TheBlueMatt (commented):
Really think we need this for 15.

sipa:
GUI-screenshot-looks-nice-ACK. Didn't look at the code.

src/qt/sendcoinsdialog.cpp
TheBlueMatt: Also, why the extra set of {}?

src/qt/sendcoinsdialog.cpp
jonasschnelli: Oh. Yes. Will fix.

src/qt/sendcoinsdialog.cpp
jonasschnelli: > Also, why the extra set of {}?

Compiler warning.

achow101:
utACK 19eae9c3602c784d55f8edc94300da1ece686a6c modulo the thing @TheBlueMatt pointed out.

I don't think we need the 3, 4, or 5 block target estimates.

It may also be useful to include a tooltip that explains why you see the line "Estimated to begin confirmation within N blocks" where N is less than the confirmation target that you chose as seen in the screenshot below. This discrepancy may confuse users.

image

morcos:
>  tooltip that explains why you see the line "Estimated to begin confirmation within N blocks" where N is less than the confirmation target that you chose

Yes would be a nice improvement, could be left for another PR as it's a preexisting issue.  Would be great to get than in for 0.15 as well, but wouldn't want it to hold up this.

The other feature that I'd ideally like is the ability to force estimates conservative or economical regardless of the Replace-By-Fee setting.  Again just a wish list item.

src/qt/sendcoinsdialog.cpp
jonasschnelli: Yes. Your right.. will fix.

src/qt/sendcoinsdialog.cpp
jonasschnelli: Hmm... I think an addition is required: we probably need to migrate the old slider value to the new dropdown?

morcos: That doesn't seem particularly important to me.

jonasschnelli:
Removed again target 3 and 5 and fixed points found by @TheBlueMatt and @promag.
Added another commit that add a tiny migration logic.

Thanks for a retest/re-ack.

morcos:
Tested ACK 2aef1f1

TheBlueMatt:
utACK 2aef1f18296fcd3aa3c91afdf152add8a8e80bd4

PR #10771, opened 2017-07-07
Remove unused variables in shell scripts

practicalswift:
Remove unused variables in shell scripts. Use `_` where we don't care about the result.


laanwj:
LGTM, utACK ab8e8b9

PR #10772, opened 2017-07-08
Implementation of BIP8

shaolinfry:
Implementation of [BIP8](https://github.com/bitcoin/bips/blob/master/bip-0008.mediawiki) workflow for soft fork deployment. 

states


luke-jr:
NACK.

instagibbs:
@luke-jr from https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md "NACKs without accompanying reasoning may be disregarded;"

luke-jr:
BIP8 is not currently used, nor may ever be used at least in its current form, since it fails to address some outstanding issues with BIP9. As such, it doesn't make sense to merge code for it.

laanwj:
Closing for now. Doesn't get much attention at the moment. This will be relevant when a softfork is to be done again.

Issue #8639, opened 2016-09-01
Docs: Minimum required dependencies and current CVEs


fanquake:
Requested by @laanwj in #8423 (which addresses some of these by upgrading packages) + #8923.

This should probably be a new file in `doc/`, as it is not specific to any one platform.
Also checks in `configure.ac` would need to be added to check if these versions are present.

| Dependancy | Version used | Minimum Required | Latest available | CVEs? | Shared | [Bundled Qt Library](https://doc.qt.io/qt-5/configure-options.html) |
| --- | --- | --- | --- | --- | --- | --- |
| libevent | 2.1.11-stable | 2.0.22 | [2.1.12-stable](https://github.com/libevent/libevent/releases) | No |  |  |
| Qt | 5.9.8 | [5.5.1](https://github.com/bitcoin/bitcoin/issues/13478) | [6.0.0](https://download.qt.io/official_releases/qt/) | No |  |  |
| Freetype | 2.7.1 |  | [2.10.4](https://download.savannah.gnu.org/releases/freetype) | No |  |  |
| Boost | 1.70.0 | [1.58.0](https://github.com/bitcoin/bitcoin/pull/19667) | [1.75.0](http://www.boost.org/users/download/) | No |  |  |
| Zeromq (optional) | 4.3.1 | 4.0  | [4.3.3](https://github.com/zeromq/libzmq/releases) | Yes |  |  |
| miniupnpc (optional) | 2.0.20180203 |  | [2.2.1](http://miniupnp.free.fr/files) | No |  |  |
| qrencode (optional) | 3.4.4 |  | [4.1.1](https://fukuchi.org/works/qrencode) | No |  |  |
| berkeley-db | 4.8.30 | 4.8.x | [18.1.40](https://www.oracle.com/technetwork/database/database-technologies/berkeleydb/downloads/index.html) | No |  |  |
| expat | 2.2.7 |  | [2.2.10](https://libexpat.github.io/) | Yes | Yes |  |
| fontconfig | 2.12.1 |  | [2.13.1](https://www.freedesktop.org/software/fontconfig/release/) | No | Yes |  |
| freetype |  |  |  |  |  | [no](https://github.com/bitcoin/bitcoin/blob/master/depends/packages/qt.mk#L34) (linux uses system) |
| zlib | 1.2.11 |  | [1.2.11](https://zlib.net/) |  |  | no |
[...]

PR #10779, opened 2017-07-09
Create dependencies.md

flack:
As @fanquake mentioned in #8639, this should probably be a file in `doc/`, so I went ahead and pulled the issue text via the github API and dumped it into a file. No modifications made, except one spelling fix. This makes the info easier to find, and it will get a proper version history, too.

### DIFF
--- a/doc/dependencies.md
+++ b/doc/dependencies.md
@@ -1,0 +1,32 @@
+Dependencies
+============
+
+These are the dependencies currently used by Bitcoin Core. You can find instructions for installing them in the `build-*.md` file for your platform.
+
+| Dependency | Version used | Minimum Required | CVEs? | Shared | [Bundled Qt Library](https://doc.qt.io/qt-5/configure-options.html) |
+| --- | --- | --- | --- | --- | --- | --- |
+| openssl | [1.0.1k]](https://www.openssl.org/source) |  | Yes |  |  |
+| ccache | [3.3.4](https://ccache.samba.org/download.html) |  | No |  |  |
+| libevent | [2.1.8-stable](https://github.com/libevent/libevent/releases) | 2.0.22 | No |  |  |
+| Qt | [5.7.1](https://download.qt.io/official_releases/qt/) | 4.7+ | No |  |  |
+| Freetype | [2.7.1](http://download.savannah.gnu.org/releases/freetype) |  | No |  |  |
+| Boost | [1.64.0](http://www.boost.org/users/download/) | [1.47.0](https://github.com/bitcoin/bitcoin/pull/8920) | No |  |  |
+| Protobuf | [2.6.3](https://github.com/google/protobuf/releases) |  | No |  |  |
+| Zeromq | [4.1.5](https://github.com/zeromq/libzmq/releases) |  | No |  |  |
+| miniupnpc | [2.0.20170509](http://miniupnp.free.fr/files) |  | No |  |  |
+| qrencode | [3.4.4](https://fukuchi.org/works/qrencode) |  | No |  |  |
+| berkeley-db | [4.8.30](http://www.oracle.com/technetwork/database/database-technologies/berkeleydb/downloads/index.html) | 4.8.x | No |  |  |
+| dbus | [1.10.18](https://cgit.freedesktop.org/dbus/dbus/tree/NEWS?h=dbus-1.10) |  | No | yes |  |
+| expat | [2.2.1](https://libexpat.github.io/) |  | No | yes |  |
+| fontconfig | [2.12.1](https://www.freedesktop.org/software/fontconfig/release/) |  | No | yes |  |
+| freetype |  |  |  |  | [no](https://github.com/bitcoin/bitcoin/blob/master/depends/packages/qt.mk#L38) (linux uses system) |
+| zlib | [1.2.11](http://zlib.net/) |  |  |  | no |
+| libjpeg |  |  |  |  | [yes](https://github.com/bitcoin/bitcoin/blob/master/depends/packages/qt.mk#L75) |
+| libpng |  |  |  |  | [yes](https://github.com/bitcoin/bitcoin/blob/master/depends/packages/qt.mk#L74) |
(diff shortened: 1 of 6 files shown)


doc/dependencies.md
```diff
@@ -0,0 +1,36 @@
+Dependencies
+============
+
+| Dependency | Version used | Minimum Required | Latest available | CVEs? | Shared | [Bundled Qt Library](https://doc.qt.io/qt-5/configure-options.html) |
```
laanwj: Would be useful to have somewhat of an introduction here - what is this document, what should it be used for, etc.
Also that instructions for *installing* dependencies are in the appropriate `build-*.md`.

laanwj:
Concept ACK
Needs an entry in `doc/README.md`

flack:
@laanwj where would you put it in `README`, under "Building" or under "Development"?

An introduction seems like a good idea, and I could write some small blurb, but it would probably better if this is done by someone who knows what he's talking about :-)

Also, do you think the link in the last line of this paragraph https://github.com/bitcoin/bitcoin/blob/master/doc/build-unix.md#dependencies (and corresponding lines in the oder `build` docs) should be changed to point to the new file?

laanwj:
> where would you put it in README, under "Building" or under "Development"?

Building, I guess. People look there sooner I think, and it's relevant to both.

> Also, do you think the link in the last line of this paragraph https://github.com/bitcoin/bitcoin/blob/master/doc/build-unix.md#dependencies (and corresponding lines in the oder build docs) should be changed to point to the new file?

Yes, such a overall overview can better go into `dependencies.md`. Do keep the instructions for installing the dependencies in the appropriate build instructions though.

flack:
I've added links from `README` and the `build-*` files & a small intro sentence as requested. Let me know if there's anything else I should change

doc/dependencies.md
```diff
@@ -0,0 +1,38 @@
+Dependencies
+============
+
+These are the dependencies currently used by Bitcoin Core. You can find instructions for installing them in the `build-*.md` file for your platform.
+
+| Dependency | Version used | Minimum Required | Latest available | CVEs? | Shared | [Bundled Qt Library](https://doc.qt.io/qt-5/configure-options.html) |
```
luke-jr: Let's not try to track latest available here...

For CVEs, rather than yes/no, it'd be more useful to link where to check for exploits.

doc/dependencies.md
```diff
@@ -0,0 +1,38 @@
...
+| --- | --- | --- | --- | --- | --- | --- |
+| openssl | 1.0.1k |  | [1.0.1u](https://www.openssl.org/source) | Yes |  |  |
+| ccache | 3.3.4 |  | [3.3.4](https://ccache.samba.org/download.html) | No |  |  |
```
luke-jr: No indication on which dependencies are optional?

doc/dependencies.md
```diff
@@ -0,0 +1,38 @@
...
+| libevent | 2.1.8-stable | 2.0.22 | [2.1.8-stable](https://github.com/libevent/libevent/releases) | No |  |  |
+| Qt | 5.7.1 | 4.7+ | [5.9.1](https://download.qt.io/official_releases/qt/) | No |  |  |
```
luke-jr: Qt is typically considered as individual components. We don't need all of it.

doc/dependencies.md
```diff
@@ -0,0 +1,38 @@
...
...=CVE-2016-2107), [CVE-2016-2105](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-2105), [CVE-2016-2106](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-2106), [CVE-2016-2108](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-2108), [CVE-2016-2109](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-2109), [CVE-2016-2176](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-2176), [CVE-2016-2177](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-2177), [CVE-2016-2178](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-2178), [CVE-2016-2179](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-2179), [CVE-2016-2180](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-2180), [CVE-2016-2181](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-2181), [CVE-2016-2182](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-2182), [CVE-2016-6302](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-6302), [CVE-2016-6303](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-6303), [CVE-2016-6304](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-6304), [CVE-2016-6306](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2016-6306)
```
luke-jr: This doesn't seem to belong here.

doc/dependencies.md
flack: wouldn't that info have to be in the platform-specific files?

doc/dependencies.md
flack: like I've written in the original PR description, for now I just took the text verbatim from #8639, but I can ofc adapt it

flack:
I've removed the "latest available" column now as requested by @luke-jr . I can also remove the listed CVEs, since they will probably get outdated pretty quickly, too. I also like the idea of putting in links to check for vulnerabilities, but I don't really know where they should link to.

fanquake:
@flack Can you update this to remove the CVEs list, and pull in the latest data from #8639 ?

flack:
@fanquake I've removed the CVEs, and compared the data with #8639, but everything still seems to be in sync

laanwj:
ACK after squashing into one commit

flack:
squashed

PR #10781, opened 2017-07-09
Python cleanups

practicalswift:
Python cleanups:
* Avoid reference to undefined name: `stderr` does not exist, `sys.stderr` does
* Use `print(...)` instead of undefined `printf(...)`
* Avoid redefinition of variable (`tx`) in list comprehension
* Remove unused variables and/or function calls
* Use `sys.exit(...)` instead of `exit(...)`: [`exit(...)` should not be used in programs](https://github.com/bitcoin/bitcoin/pull/10753#discussion_r125935027)


jnewbery (commented):
Looks good. utACK. One question inline and another general question:

In commit *Use the variable name _ to show that we are intentionally ignoring a return value*, why not just ignore the return value?

test/functional/bumpfee.py
```diff
@@ -168,7 +168,7 @@ def test_bumpfee_with_descendant_fails(rbf_node, rbf_node_address, dest_address)
     parent_id = spend_one_input(rbf_node, rbf_node_address)
     tx = rbf_node.createrawtransaction([{"txid": parent_id, "vout": 0}], {dest_address: 0.00020000})
     tx = rbf_node.signrawtransaction(tx)
-    txid = rbf_node.sendrawtransaction(tx["hex"])
+    _ = rbf_node.sendrawtransaction(tx["hex"])
```
jtimon: what about something more clear like throwaway_txid or something of the sort?

jtimon:
fast review ACK

test/functional/p2p-compactblocks.py
```diff
@@ -293,7 +293,7 @@ def test_compactblock_construction(self, node, test_node, version, use_witness_a
 
         # Store the raw block in our internal format.
         block = FromHex(CBlock(), node.getblock("%02x" % block_hash, False))
-        [tx.calc_sha256() for tx in block.vtx]
```
practicalswift: Yes it does :-)

```shell
$ flake8 test/functional/p2p-compactblocks.py | grep redefine
test/functional/p2p-compactblocks.py:296:31: F812 list comprehension redefines 'tx' from line 276
```

Should I remove the commit 263686d03fbfa2191c160d4c50951852fa1a4900 from this PR?

practicalswift:
@jnewbery Good point! I've now changed from `_ = f()` to `f()` for calls that appear to have side effects. For calls that appear to have no side effects I've simply removed the call.

@jnewbery @jtimon Could you take a look at the updated version and see if it looks correct? :-)

practicalswift:
Added two commits :-)

test/functional/test_framework/socks5.py
```diff
@@ -91,7 +91,7 @@ def handle(self):
                 self.conn.sendall(bytearray([0x01, 0x00]))
 
             # Read connect request
-            (ver,cmd,rsv,atyp) = recvall(self.conn, 4)
+            (ver, cmd, _, atyp) = recvall(self.conn, 4)
```
jnewbery: no need for this to be a tuple. Just:

```
ver, cmd, _, atyp = recvall(self.conn, 4)
```

test/functional/preciousblock.py
```diff
@@ -47,16 +47,16 @@ def run_test(self):
         self.log.info("Ensure submitblock can in principle reorg to a competing chain")
         self.nodes[0].generate(1)
         assert_equal(self.nodes[0].getblockcount(), 1)
-        (hashY, hashZ) = self.nodes[1].generate(2)
+        (_, hashZ) = self.nodes[1].generate(2)
```
jnewbery: prefer:

```
hashZ = self.nodes[1].generate(2)[-1]
```

since it conveys the meaning better: "give me the hash of the last block generated".

same for the two examples below

test/functional/test_framework/netutil.py
```diff
@@ -15,17 +15,7 @@
 import os
 from binascii import unhexlify, hexlify
 
-STATE_ESTABLISHED = '01'
```
jnewbery: Perhaps change these to comments rather than erase them (for context).

test/functional/test_framework/mininode.py
```diff
@@ -48,8 +48,6 @@
 COIN = 100000000 # 1 btc in satoshis
 
 NODE_NETWORK = (1 << 0)
-NODE_GETUTXO = (1 << 1)
```
jnewbery: No harm in leaving these as comments.

test/functional/test_framework/key.py
```diff
@@ -112,40 +112,16 @@ def set_secretbytes(self, secret):
         ssl.BN_CTX_free(ctx)
         return self.k
 
-    def set_privkey(self, key):
```
jnewbery: I can imagine these methods potentially being useful at some point. I don't think there's any need to remove them.

test/functional/test_framework/authproxy.py
```diff
@@ -158,11 +158,6 @@ def __call__(self, *args, **argsn):
         else:
             return response['result']
 
-    def _batch(self, rpc_call_list):
```
jnewbery: Please leave this for now.

jnewbery:
Mostly ACK, but a few comments.

test/functional/test_framework/mininode.py
practicalswift: Good point! Updated but GitHub doesn't show the `# NODE_GETUTXO = (1 << 1)` below, so it appears as if it is not fixed :-)

test/functional/test_framework/netutil.py
practicalswift: Good point! Fixed!

practicalswift:
@jnewbery Thanks for reviewing! Good feedback - I've addressed all points raised. Would you mind re-reviewing? :-)

jnewbery:
Looks good. Just https://github.com/bitcoin/bitcoin/pull/10781#discussion_r128618950 unaddressed (using a for loop instead of list comprehension for functions with side-effects)

practicalswift:
@jnewbery Thanks! List comprehension change now fixed as well. Looks good? :-)

jnewbery:
looks good. utACK db442ada2b7e162c3748ddcc130c414fd86e2bab

practicalswift:
Rebased and and added another commit ("Remove import of unused constants").

@jnewbery, would you mind re-reviewing? :-)

jnewbery:
Looks good to me, but can you remove the final commit? Marco's #11067 will already remove those unused constants.

@MarcoFalke - if you have time, can you check whether this is suitable for merge? It's likely to require continual rebase if left open.

I've verified that @practicalswift hasn't added any backdoors in the last rebase :)

practicalswift:
@jnewbery Removed last commit :-)  Thanks fo reviewing :-)

jtimon:
repeatead fast review ack

MeshCollider:
Fast review utACK

PR #10785, opened 2017-07-10
Serialization improvements

sipa:
This PR improves correctness (removing potentially unsafe `const_cast`s) and flexibility of the serialization code.

The main issue is that use of the current `ADD_SERIALIZE_METHODS` macro (which is the only way to not duplicate serialization and deserialization code) only expands to a single class method, and thus can only be qualified as either const or non-const - not both. In many cases, serialization needs to work on const objects however, and preferably that is done without casts that could hide const-correctness bugs.

To deal with that, this PR introduces a new approach that includes a `SERIALIZE_METHODS(obj)` macro, where `obj` is a variable name. It expands to some boilerplate and a static method to which the object itself is an argument. The advantage is that its type can be templated, and be `const` when serializing.

Another issue is the various serialization-wrapping macros (`VARINT`, `COMPACTSIZE`, `FLATDATA` and `LIMITED_STRING`). They all `const_cast` their argument in order to construct a wrapper object, which supports both serialization and deserialization. This PR makes them templated in the underlying data type (for example, `CompactSizeWrapper<uint64_t>`). This has the advantage that we can make the template type `const` when invoked on a `const` variable (so it would be `CompactSizeWrapper<const uint64_t>` in that case).

A last issue is the persistent use of the `REF` macro to deal with temporary expressions being passed in. Since C++11, this is not needed anymore as temporaries are explicitly represented as rvalue references. Thus we can remove `REF` invocations and instead just make the various classes and helper functions deal correctly with references.

The above changes permit a fully const-correct version of all serialization code. However, it is cumbersome. Any existing `ADD_SERIALIZE_METHODS` instances in the code that do more than just (conditionally) serializing/deserializing some fields (in particular, it contains branches that assign to some of the variables) need to be split up into an explicit `Serialize` and `Unserialize` method instead. In some cases this is inevitable (wallet serializers do some crazy transformations while serializing!), but in many cases it is just annoying duplication.

To improve upon this, a few more primitives that are currently inlined are turned into serialization wrappers:
* `BigEndianWrapper`: Serializes/deserializes an integer as big endian rather than little endian (only for 16-bit). This permits the CService serialization to become a oneliner.
* `Uint48Wrapper`: Serializes/deserializes only the lower 48 bits of an integer (used in BIP152 code).
* `VectorApplyWrapper`: Serializes/deserializes a vector while using a custom serializer for its elements. This simplifies the undo and blockencoding serializers a lot.

Best of all, it removes 147 lines of while code adding a bunch of comments (though the increased use of vararg `READWRITE` is probably cheating a bit).

The commits are ordered into 3 sections:
* First, introduce new classes that permit const-correct serialization.
* Then one by one transform the various files to use the new serializers.
* Finally, remove the old serializers.

This may be too much to go in at once. I'm happy to split things up as needed.

### DIFF
--- a/src/blockencodings.h
+++ b/src/blockencodings.h
@@ -11,18 +11,81 @@
 class CTxMemPool;
 
 // Dumb helper to handle CTransaction compression at serialize-time
-struct TransactionCompressor {
-private:
-    CTransactionRef& tx;
-public:
-    explicit TransactionCompressor(CTransactionRef& txIn) : tx(txIn) {}
+struct TransactionCompression
+{
+    template<typename T>
+    struct Wrapper {
+    private:
+        T& tx;
+    public:
+        explicit Wrapper(T& txIn) : tx(txIn) {}
+        SERIALIZE_METHODS(Wrapper, obj) { READWRITE(obj.tx); }
+    };
+};
 
-    ADD_SERIALIZE_METHODS;
+struct Uint48
+{
+    template<typename I>
+    struct Wrapper {
+    private:
+        I& m_int;
+    public:
+        explicit Wrapper(I& i) : m_int(i)
+        {
+            static_assert(std::is_unsigned<I>::value, "Uint48 needs an unsigned integer");
+            static_assert(sizeof(I) >= 6, "Uint48 needs a 48+ bit type");
+        }
 
-    template <typename Stream, typename Operation>
-    inline void SerializationOp(Stream& s, Operation ser_action) {
-        READWRITE(tx); //TODO: Compress tx encoding
-    }
+        template <typename Stream> void Serialize(Stream& s) const
+        {
+            uint32_t lsb = m_int & 0xffffffff;
+            uint16_t msb = (m_int >> 32) & 0xffff;
+            s << lsb << msb;
+        }
+
+        template <typename Stream> void Unserialize(Stream& s)
+        {
+            uint32_t lsb;
+            uint16_t msb;
+            s >> lsb >> msb;
+            m_int = (uint64_t(msb) << 32) | uint64_t(lsb);
+        }
+    };
+};
+
+/** Vector-wrapper (compatible with VectorApply) to differentially encode values. */
+struct Differential
+{
+    template<typename V>
+    struct Wrapper {
+    private:
+        V& m_v;
+
+    public:
+        typedef typename V::value_type value_type;
+
+        explicit Wrapper(V& v) : m_v(v) {}
+        size_t size() const { return m_v.size(); }
+        void clear() { m_v.clear(); }
+        void reserve(size_t size) { m_v.reserve(size); }
+
+        value_type operator[](size_t pos) const
+        {
+            if (pos == 0) return m_v[0];
+            return m_v[pos] - (m_v[pos - 1] + 1);
+        }
+
+        void push_back(value_type val)
+        {
+            if (m_v.size() == 0) {
+                m_v.push_back(val);
+            } else {
+                value_type add = val + (m_v.back() + 1);
+                if (add <= val) throw std::ios_base::failure("differential value overflow");
+                m_v.push_back(add);
+            }
+        }
+    };
 };
 
 class BlockTransactionsRequest {
@@ -31,39 +94,9 @@ public:
     uint256 blockhash;
     std::vector<uint16_t> indexes;
 
-    ADD_SERIALIZE_METHODS;
-
-    template <typename Stream, typename Operation>
-    inline void SerializationOp(Stream& s, Operation ser_action) {
-        READWRITE(blockhash);
-        uint64_t indexes_size = (uint64_t)indexes.size();
-        READWRITE(COMPACTSIZE(indexes_size));
-        if (ser_action.ForRead()) {
-            size_t i = 0;
-            while (indexes.size() < indexes_size) {
-                indexes.resize(std::min((uint64_t)(1000 + indexes.size()), indexes_size));
-                for (; i < indexes.size(); i++) {
-                    uint64_t index = 0;
-                    READWRITE(COMPACTSIZE(index));
-                    if (index > std::numeric_limits<uint16_t>::max())
-                        throw std::ios_base::failure("index overflowed 16 bits");
-                    indexes[i] = index;
-                }
-            }
-
-            int32_t offset = 0;
-            for (size_t j = 0; j < indexes.size(); j++) {
-                if (int32_t(indexes[j]) + offset > std::numeric_limits<uint16_t>::max())
-                    throw std::ios_base::failure("indexes overflowed 16 bits");
-                indexes[j] = indexes[j] + offset;
-                offset = int32_t(indexes[j]) + 1;
-            }
-        } else {
-            for (size_t i = 0; i < indexes.size(); i++) {
-                uint64_t index = indexes[i] - (i == 0 ? 0 : (indexes[i - 1] + 1));
-                READWRITE(COMPACTSIZE(index));
-            }
-        }
+    SERIALIZE_METHODS(BlockTransactionsRequest, obj)
+    {
+        READWRITE(obj.blockhash, Wrap<VectorApply<CompactSize>>(Wrap<Differential>(obj.indexes)));
     }
 };
 
@@ -77,24 +110,9 @@ public:
     explicit BlockTransactions(const BlockTransactionsRequest& req) :
         blockhash(req.blockhash), txn(req.indexes.size()) {}
 
-    ADD_SERIALIZE_METHODS;
-
-    template <typename Stream, typename Operation>
-    inline void SerializationOp(Stream& s, Operation ser_action) {
-        READWRITE(blockhash);
-        uint64_t txn_size = (uint64_t)txn.size();
-        READWRITE(COMPACTSIZE(txn_size));
-        if (ser_action.ForRead()) {
-            size_t i = 0;
-            while (txn.size() < txn_size) {
-                txn.resize(std::min((uint64_t)(1000 + txn.size()), txn_size));
-                for (; i < txn.size(); i++)
-                    READWRITE(TransactionCompressor(txn[i]));
-            }
-        } else {
-            for (size_t i = 0; i < txn.size(); i++)
-                READWRITE(TransactionCompressor(txn[i]));
-        }
+    SERIALIZE_METHODS(BlockTransactions, obj)
+    {
+        READWRITE(obj.blockhash, Wrap<VectorApply<TransactionCompression>>(obj.txn));
     }
 };
 
@@ -105,16 +123,21 @@ struct PrefilledTransaction {
     uint16_t index;
     CTransactionRef tx;
 
-    ADD_SERIALIZE_METHODS;
+    template<typename Stream>
+    void Serialize(Stream& s) const
+    {
+        s << COMPACTSIZE(index) << Wrap<TransactionCompression>(tx);
+    }
 
-    template <typename Stream, typename Operation>
-    inline void SerializationOp(Stream& s, Operation ser_action) {
-        uint64_t idx = index;
-        READWRITE(COMPACTSIZE(idx));
+    template<typename Stream>
+    void Unserialize(Stream& s)
+    {
+        uint64_t idx;
+        s >> COMPACTSIZE(idx);
         if (idx > std::numeric_limits<uint16_t>::max())
             throw std::ios_base::failure("index overflowed 16-bits");
         index = idx;
-        READWRITE(TransactionCompressor(tx));
+        s >> Wrap<TransactionCompression>(tx);
     }
 };
 
@@ -153,43 +176,21 @@ public:
 
     size_t BlockTxCount() const { return shorttxids.size() + prefilledtxn.size(); }
 
-    ADD_SERIALIZE_METHODS;
+    template <typename Stream>
+    void Serialize(Stream& s) const
+    {
+        s << header << nonce << Wrap<VectorApply<Uint48>>(shorttxids) << prefilledtxn;
+    }
 
-    template <typename Stream, typename Operation>
-    inline void SerializationOp(Stream& s, Operation ser_action) {
-        READWRITE(header);
-        READWRITE(nonce);
-
-        uint64_t shorttxids_size = (uint64_t)shorttxids.size();
-        READWRITE(COMPACTSIZE(shorttxids_size));
-        if (ser_action.ForRead()) {
-            size_t i = 0;
-            while (shorttxids.size() < shorttxids_size) {
-                shorttxids.resize(std::min((uint64_t)(1000 + shorttxids.size()), shorttxids_size));
-                for (; i < shorttxids.size(); i++) {
-                    uint32_t lsb = 0; uint16_t msb = 0;
-                    READWRITE(lsb);
-                    READWRITE(msb);
-                    shorttxids[i] = (uint64_t(msb) << 32) | uint64_t(lsb);
-                    static_assert(SHORTTXIDS_LENGTH == 6, "shorttxids serialization assumes 6-byte shorttxids");
-                }
-            }
-        } else {
-            for (size_t i = 0; i < shorttxids.size(); i++) {
-                uint32_t lsb = shorttxids[i] & 0xffffffff;
-                uint16_t msb = (shorttxids[i] >> 32) & 0xffff;
-                READWRITE(lsb);
-                READWRITE(msb);
-            }
-        }
-
-        READWRITE(prefilledtxn);
-
-        if (BlockTxCount() > std::numeric_limits<uint16_t>::max())
+    template <typename Stream>
+    inline void Unserialize(Stream& s)
+    {
+        static_assert(SHORTTXIDS_LENGTH == 6, "shorttxids serialization assumes 6-byte shorttxids");
+        s >> header >> nonce >> Wrap<VectorApply<Uint48>>(shorttxids) >> prefilledtxn;
+        if (BlockTxCount() > std::numeric_limits<uint16_t>::max()) {
             throw std::ios_base::failure("indexes overflowed 16 bits");
-
-        if (ser_action.ForRead())
-            FillShortTxIDSelector();
+        }
+        FillShortTxIDSelector();
     }
 };
 
--- a/src/blockencodings.h
+++ b/src/blockencodings.h
@@ -13,79 +13,61 @@ class CTxMemPool;
 // Dumb helper to handle CTransaction compression at serialize-time
 struct TransactionCompression
 {
-    template<typename T>
-    struct Wrapper {
-    private:
-        T& tx;
-    public:
-        explicit Wrapper(T& txIn) : tx(txIn) {}
-        SERIALIZE_METHODS(Wrapper, obj) { READWRITE(obj.tx); }
-    };
+    template<typename Stream> void Ser(Stream& s, const CTransactionRef& tx) { s << tx; }
+    template<typename Stream> void Unser(Stream& s, CTransactionRef& tx) { s >> tx; }
 };
 
 struct Uint48
 {
-    template<typename I>
-    struct Wrapper {
-    private:
-        I& m_int;
-    public:
-        explicit Wrapper(I& i) : m_int(i)
-        {
-            static_assert(std::is_unsigned<I>::value, "Uint48 needs an unsigned integer");
-            static_assert(sizeof(I) >= 6, "Uint48 needs a 48+ bit type");
-        }
+    template <typename Stream, typename I> void Ser(Stream& s, I v)
+    {
+        static_assert(std::is_unsigned<I>::value, "Uint48 needs an unsigned integer");
+        static_assert(sizeof(I) >= 6, "Uint48 needs a 48+ bit type");
 
-        template <typename Stream> void Serialize(Stream& s) const
-        {
-            uint32_t lsb = m_int & 0xffffffff;
-            uint16_t msb = (m_int >> 32) & 0xffff;
-            s << lsb << msb;
-        }
+        uint32_t lsb = v & 0xffffffff;
+        uint16_t msb = (v >> 32) & 0xffff;
+        s << lsb << msb;
+    }
 
-        template <typename Stream> void Unserialize(Stream& s)
-        {
-            uint32_t lsb;
-            uint16_t msb;
-            s >> lsb >> msb;
-            m_int = (uint64_t(msb) << 32) | uint64_t(lsb);
-        }
-    };
+    template <typename Stream, typename I> void Unser(Stream& s, I& v)
+    {
+        static_assert(std::is_unsigned<I>::value, "Uint48 needs an unsigned integer");
+        static_assert(sizeof(I) >= 6, "Uint48 needs a 48+ bit type");
+
+        uint32_t lsb;
+        uint16_t msb;
+        s >> lsb >> msb;
+        v = (uint64_t(msb) << 32) | uint64_t(lsb);
+    }
 };
 
-/** Vector-wrapper (compatible with VectorApply) to differentially encode values. */
-struct Differential
+template<typename I>
+struct DifferentialSerTransform
 {
-    template<typename V>
-    struct Wrapper {
-    private:
-        V& m_v;
+    I m_subtract = 0;
 
-    public:
-        typedef typename V::value_type value_type;
+    I operator()(I val)
+    {
+        I tmp = val - m_subtract;
+        m_subtract = I(val + 1);
+        return tmp;
+    }
+};
 
-        explicit Wrapper(V& v) : m_v(v) {}
-        size_t size() const { return m_v.size(); }
-        void clear() { m_v.clear(); }
-        void reserve(size_t size) { m_v.reserve(size); }
+template<typename I>
+struct DifferentialUnserTransform
+{
+    bool m_first = true;
+    I m_add = 0;
 
-        value_type operator[](size_t pos) const
-        {
-            if (pos == 0) return m_v[0];
-            return m_v[pos] - (m_v[pos - 1] + 1);
-        }
-
-        void push_back(value_type val)
-        {
-            if (m_v.size() == 0) {
-                m_v.push_back(val);
-            } else {
-                value_type add = val + (m_v.back() + 1);
-                if (add <= val) throw std::ios_base::failure("differential value overflow");
-                m_v.push_back(add);
-            }
-        }
-    };
+    I operator()(I val)
+    {
+        I tmp = I(val + m_add);
+        if (!m_first && tmp <= val) throw std::ios_base::failure("differential value overflow");
+        m_first = false;
+        m_add = I(tmp + 1);
+        return tmp;
+    }
 };
 
 class BlockTransactionsRequest {
@@ -96,7 +78,7 @@ public:
 
     SERIALIZE_METHODS(BlockTransactionsRequest, obj)
     {
-        READWRITE(obj.blockhash, Wrap<VectorApply<CompactSize>>(Wrap<Differential>(obj.indexes)));
+        READWRITE(obj.blockhash, Using<VectorUsing<CompactSizeFormatter, DifferentialSerTransform<uint16_t>, DifferentialUnserTransform<uint16_t>>>(obj.indexes));
     }
 };
 
@@ -112,7 +94,7 @@ public:
 
     SERIALIZE_METHODS(BlockTransactions, obj)
     {
-        READWRITE(obj.blockhash, Wrap<VectorApply<TransactionCompression>>(obj.txn));
+        READWRITE(obj.blockhash, Using<VectorUsing<TransactionCompression>>(obj.txn));
     }
 };
 
@@ -126,7 +108,7 @@ struct PrefilledTransaction {
     template<typename Stream>
     void Serialize(Stream& s) const
     {
-        s << COMPACTSIZE(index) << Wrap<TransactionCompression>(tx);
+        s << COMPACTSIZE(index) << Using<TransactionCompression>(tx);
     }
 
     template<typename Stream>
@@ -137,7 +119,7 @@ struct PrefilledTransaction {
         if (idx > std::numeric_limits<uint16_t>::max())
             throw std::ios_base::failure("index overflowed 16-bits");
         index = idx;
-        s >> Wrap<TransactionCompression>(tx);
+        s >> Using<TransactionCompression>(tx);
     }
 };
 
@@ -179,14 +161,14 @@ public:
     template <typename Stream>
     void Serialize(Stream& s) const
     {
-        s << header << nonce << Wrap<VectorApply<Uint48>>(shorttxids) << prefilledtxn;
+        s << header << nonce << Using<VectorUsing<Uint48>>(shorttxids) << prefilledtxn;
     }
 
     template <typename Stream>
     inline void Unserialize(Stream& s)
     {
         static_assert(SHORTTXIDS_LENGTH == 6, "shorttxids serialization assumes 6-byte shorttxids");
-        s >> header >> nonce >> Wrap<VectorApply<Uint48>>(shorttxids) >> prefilledtxn;
+        s >> header >> nonce >> Using<VectorUsing<Uint48>>(shorttxids) >> prefilledtxn;
         if (BlockTxCount() > std::numeric_limits<uint16_t>::max()) {
             throw std::ios_base::failure("indexes overflowed 16 bits");
         }
--- a/src/compressor.h
+++ b/src/compressor.h
@@ -11,10 +11,6 @@
 #include <serialize.h>
 #include <span.h>
 
-class CKeyID;
-class CPubKey;
-class CScriptID;
-
 bool CompressScript(const CScript& script, std::vector<unsigned char> &out);
 unsigned int GetSpecialScriptSize(unsigned int nSize);
 bool DecompressScript(CScript& script, unsigned int nSize, const std::vector<unsigned char> &out);
@@ -33,79 +29,94 @@ uint64_t DecompressAmount(uint64_t nAmount);
  *  Other scripts up to 121 bytes require 1 byte + script length. Above
  *  that, scripts up to 16505 bytes require 2 bytes + script length.
  */
-class CScriptCompressor
+struct ScriptCompression
 {
-private:
-    /**
-     * make this static for now (there are only 6 special scripts defined)
-     * this can potentially be extended together with a new nVersion for
-     * transactions, in which case this value becomes dependent on nVersion
-     * and nHeight of the enclosing transaction.
-     */
-    static const unsigned int nSpecialScripts = 6;
+    template<typename S>
+    class Wrapper
+    {
+    private:
+        /**
+         * make this static for now (there are only 6 special scripts defined)
+         * this can potentially be extended together with a new nVersion for
+         * transactions, in which case this value becomes dependent on nVersion
+         * and nHeight of the enclosing transaction.
+         */
+        static const unsigned int nSpecialScripts = 6;
 
-    CScript &script;
-public:
-    explicit CScriptCompressor(CScript &scriptIn) : script(scriptIn) { }
+        S &script;
+    public:
+        explicit Wrapper(S &scriptIn) : script(scriptIn) { }
 
-    template<typename Stream>
-    void Serialize(Stream &s) const {
-        std::vector<unsigned char> compr;
-        if (CompressScript(script, compr)) {
-            s << MakeSpan(compr);
-            return;
+        template<typename Stream>
+        void Serialize(Stream &s) const {
+            std::vector<unsigned char> compr;
+            if (CompressScript(script, compr)) {
+                s << MakeSpan(compr);
+                return;
+            }
+            unsigned int nSize = script.size() + nSpecialScripts;
+            s << VARINT(nSize);
+            s << MakeSpan(script);
         }
-        unsigned int nSize = script.size() + nSpecialScripts;
-        s << VARINT(nSize);
-        s << MakeSpan(script);
-    }
 
-    template<typename Stream>
-    void Unserialize(Stream &s) {
-        unsigned int nSize = 0;
-        s >> VARINT(nSize);
-        if (nSize < nSpecialScripts) {
-            std::vector<unsigned char> vch(GetSpecialScriptSize(nSize), 0x00);
-            s >> MakeSpan(vch);
-            DecompressScript(script, nSize, vch);
-            return;
+        template<typename Stream>
+        void Unserialize(Stream &s) {
+            unsigned int nSize = 0;
+            s >> VARINT(nSize);
+            if (nSize < nSpecialScripts) {
+                std::vector<unsigned char> vch(GetSpecialScriptSize(nSize), 0x00);
+                s >> MakeSpan(vch);
+                DecompressScript(script, nSize, vch);
+                return;
+            }
+            nSize -= nSpecialScripts;
+            if (nSize > MAX_SCRIPT_SIZE) {
+                // Overly long script, replace with a short invalid one
+                script << OP_RETURN;
+                s.ignore(nSize);
+            } else {
+                script.resize(nSize);
+                s >> MakeSpan(script);
+            }
         }
-        nSize -= nSpecialScripts;
-        if (nSize > MAX_SCRIPT_SIZE) {
-            // Overly long script, replace with a short invalid one
-            script << OP_RETURN;
-            s.ignore(nSize);
-        } else {
-            script.resize(nSize);
-            s >> MakeSpan(script);
+    };
+};
+
+struct AmountCompression
+{
+    template<typename I>
+    class Wrapper
+    {
+        I& m_val;
+    public:
+        explicit Wrapper(I& val) : m_val(val) {}
+        template<typename Stream> void Serialize(Stream& s) const
+        {
+            s << VARINT(CompressAmount(m_val));
         }
-    }
+        template<typename Stream> void Unserialize(Stream& s)
+        {
+            uint64_t v;
+            s >> VARINT(v);
+            m_val = DecompressAmount(v);
+        }
+    };
 };
 
 /** wrapper for CTxOut that provides a more compact serialization */
-class CTxOutCompressor
+struct TxOutCompression
 {
-private:
-    CTxOut &txout;
-
-public:
-    explicit CTxOutCompressor(CTxOut &txoutIn) : txout(txoutIn) { }
-
-    ADD_SERIALIZE_METHODS;
-
-    template <typename Stream, typename Operation>
-    inline void SerializationOp(Stream& s, Operation ser_action) {
-        if (!ser_action.ForRead()) {
-            uint64_t nVal = CompressAmount(txout.nValue);
-            READWRITE(VARINT(nVal));
-        } else {
-            uint64_t nVal = 0;
-            READWRITE(VARINT(nVal));
-            txout.nValue = DecompressAmount(nVal);
+    template<typename O>
+    class Wrapper
+    {
+    private:
+        O &txout;
+    public:
+        Wrapper(O &txoutIn) : txout(txoutIn) { }
+        SERIALIZE_METHODS(Wrapper, obj) {
+            READWRITE(Wrap<AmountCompression>(obj.txout.nValue), Wrap<ScriptCompression>(obj.txout.scriptPubKey));
         }
-        CScriptCompressor cscript(REF(txout.scriptPubKey));
-        READWRITE(cscript);
-    }
+    };
 };
 
 #endif // BITCOIN_COMPRESSOR_H
--- a/src/compressor.h
+++ b/src/compressor.h
@@ -31,92 +31,74 @@ uint64_t DecompressAmount(uint64_t nAmount);
  */
 struct ScriptCompression
 {
-    template<typename S>
-    class Wrapper
-    {
-    private:
-        /**
-         * make this static for now (there are only 6 special scripts defined)
-         * this can potentially be extended together with a new nVersion for
-         * transactions, in which case this value becomes dependent on nVersion
-         * and nHeight of the enclosing transaction.
-         */
-        static const unsigned int nSpecialScripts = 6;
+    /**
+     * make this static for now (there are only 6 special scripts defined)
+     * this can potentially be extended together with a new nVersion for
+     * transactions, in which case this value becomes dependent on nVersion
+     * and nHeight of the enclosing transaction.
+     */
+    static const unsigned int nSpecialScripts = 6;
 
-        S &script;
-    public:
-        explicit Wrapper(S &scriptIn) : script(scriptIn) { }
-
-        template<typename Stream>
-        void Serialize(Stream &s) const {
-            std::vector<unsigned char> compr;
-            if (CompressScript(script, compr)) {
-                s << MakeSpan(compr);
-                return;
-            }
-            unsigned int nSize = script.size() + nSpecialScripts;
-            s << VARINT(nSize);
-            s << MakeSpan(script);
+    template<typename Stream>
+    void Ser(Stream &s, const CScript& script) {
+        std::vector<unsigned char> compr;
+        if (CompressScript(script, compr)) {
+            s << MakeSpan(compr);
+            return;
         }
+        unsigned int nSize = script.size() + nSpecialScripts;
+        s << VARINT(nSize);
+        s << MakeSpan(script);
+    }
 
-        template<typename Stream>
-        void Unserialize(Stream &s) {
-            unsigned int nSize = 0;
-            s >> VARINT(nSize);
-            if (nSize < nSpecialScripts) {
-                std::vector<unsigned char> vch(GetSpecialScriptSize(nSize), 0x00);
-                s >> MakeSpan(vch);
-                DecompressScript(script, nSize, vch);
-                return;
-            }
-            nSize -= nSpecialScripts;
-            if (nSize > MAX_SCRIPT_SIZE) {
-                // Overly long script, replace with a short invalid one
-                script << OP_RETURN;
-                s.ignore(nSize);
-            } else {
-                script.resize(nSize);
-                s >> MakeSpan(script);
-            }
+    template<typename Stream>
+    void Unser(Stream &s, CScript& script) {
+        unsigned int nSize = 0;
+        s >> VARINT(nSize);
+        if (nSize < nSpecialScripts) {
+            std::vector<unsigned char> vch(GetSpecialScriptSize(nSize), 0x00);
+            s >> MakeSpan(vch);
+            DecompressScript(script, nSize, vch);
+            return;
         }
-    };
+        nSize -= nSpecialScripts;
+        if (nSize > MAX_SCRIPT_SIZE) {
+            // Overly long script, replace with a short invalid one
+            script << OP_RETURN;
+            s.ignore(nSize);
+        } else {
+            script.resize(nSize);
+            s >> MakeSpan(script);
+        }
+    }
 };
 
 struct AmountCompression
 {
-    template<typename I>
-    class Wrapper
+    template<typename Stream, typename I> void Ser(Stream& s, I val)
     {
-        I& m_val;
-    public:
-        explicit Wrapper(I& val) : m_val(val) {}
-        template<typename Stream> void Serialize(Stream& s) const
-        {
-            s << VARINT(CompressAmount(m_val));
-        }
-        template<typename Stream> void Unserialize(Stream& s)
-        {
-            uint64_t v;
-            s >> VARINT(v);
-            m_val = DecompressAmount(v);
-        }
-    };
+        s << VARINT(CompressAmount(val));
+    }
+    template<typename Stream, typename I> void Unser(Stream& s, I& val)
+    {
+        uint64_t v;
+        s >> VARINT(v);
+        val = DecompressAmount(v);
+    }
 };
 
 /** wrapper for CTxOut that provides a more compact serialization */
 struct TxOutCompression
 {
-    template<typename O>
-    class Wrapper
+    template<typename Stream> void Ser(Stream& s, const CTxOut& txout)
     {
-    private:
-        O &txout;
-    public:
-        Wrapper(O &txoutIn) : txout(txoutIn) { }
-        SERIALIZE_METHODS(Wrapper, obj) {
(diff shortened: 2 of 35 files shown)


gmaxwell:
Should probably be tested on big endian. :)

src/blockencodings.h
```diff
@@ -12,58 +12,70 @@
...
+            const uint64_t index = indexes[i] - (i == 0 ? 0 : (indexes[i - 1] + 1));
+            s << CompactSize(index);
+        }
+    }
+
+    template <typename Stream>
+    void Unserialize(Stream& s)
+    {
+        std::vector<uint64_t> tmp;
+        s >> blockhash >> VectorApply<CompactSizeWrapper>(tmp);
```
sipa: For ease of implementation, deserialization first happens into a uint64_t, and is then converted. This means a temporary is created and allocated, which is an overhead that the old implementation didn't have.

If this is considered too much, an alternative is to lose the overflow checking, and instead use `s >> blockhash >> VectorApply<CompactSizeWrapper>(indexes)` directly, which will work, but compute results modulo 2^16.

src/test/serialize_tests.cpp
```diff
@@ -49,11 +48,10 @@ class CSerializeMethodsTestMany : public CSerializeMethodsTestSingle
...
 public:
     using CSerializeMethodsTestSingle::CSerializeMethodsTestSingle;
-    ADD_SERIALIZE_METHODS;
 
-    template <typename Stream, typename Operation>
-    inline void SerializationOp(Stream& s, Operation ser_action) {
-        READWRITEMANY(intval, boolval, stringval, FLATDATA(charstrval), txval);
+    SERIALIZE_METHODS(obj)
+    {
+        READWRITE(obj.intval, obj.boolval, obj.stringval, FlatData(obj.charstrval), obj.txval);
```
sipa: This whole test is somewhat less valuable now, as both cases use `READWRITE`.

src/wallet/wallet.h
```diff
@@ -378,42 +376,36 @@ class CWalletTx : public CMerkleTx
         nOrderPos = -1;
     }
 
-    ADD_SERIALIZE_METHODS;
-
-    template <typename Stream, typename Operation>
-    inline void SerializationOp(Stream& s, Operation ser_action) {
-        if (ser_action.ForRead())
-            Init(NULL);
+    template<typename Stream>
```
sipa: This is one of the more involved changes, as it's both splitting the serializer into two versions, and the `Serialize` code no longer modifies `mapValue` in-place (wtf?).

src/wallet/wallet.h
```diff
@@ -378,42 +376,36 @@ class CWalletTx : public CMerkleTx
...
-        READWRITE(fSpent);
-
-        if (ser_action.ForRead())
-        {
-            strFromAccount = mapValue["fromaccount"];
+        s << vUnused << mapValueCopy << vOrderForm << fTimeReceivedIsTxTime << nTimeReceived << fFromMe << fSpent;
+    }
 
-            ReadOrderPos(nOrderPos, mapValue);
+    template<typename Stream>
```
sipa: Here is another big change, that avoids modifying `mapValue` and `strAccount` (and then later fixing it up before returning!).

src/serialize.h
```diff
@@ -456,14 +458,104 @@ class LimitedString
...
+ *     std::vector<uint64_t> v;
+ *     SERIALIZE_METHODS(obj) { READWRITE(VectorApply<VarIntWrapper>(obj.v)); }
+ *   };
+ * will define a struct that contains a vector of uint64_t, which is serialized
+ * as a vector of VarInt-encoded integers.
+ *
+ * V is not required to be an std::vector type. It works for any class that
+ * exposes a value_type, iteration, and resize method that behave like vectors.
+ */
+template<template <typename> class W, typename V> class VectorApplyWrapper
```
sipa: Notice the unusual construction of a template that takes a template as parameter here. See "Template template parameter" here: http://en.cppreference.com/w/cpp/language/template_parameters

src/serialize.h
```diff
@@ -898,16 +978,16 @@ void SerializeMany(Stream& s)
 }
 
 template<typename Stream, typename Arg>
-void SerializeMany(Stream& s, Arg&& arg)
+void SerializeMany(Stream& s, const Arg& arg)
 {
-    ::Serialize(s, std::forward<Arg>(arg));
+    ::Serialize(s, arg);
```
sipa: The reason for removing the `std::forward` calls here is explained in the commit message (there is no benefit in passing down the rvalue-ness).

sipa (commented):
Here is a self-review in which I point out some of the things reviewers may want to be aware of.

jonasschnelli:
Concept ACK.
Binaries: https://bitcoin.jonasschnelli.ch/build/210 (Currently running on a fresh node)
Agree with @gmaxwell that some BE testing would be good.

Will code-review soon.

sipa:
Made some changes to reduce the size of the overall diff.

src/serialize.h
```diff
@@ -495,11 +494,14 @@ class LimitedString
...
     {
-        WriteCompactSize(s, string.size());
-        if (!string.empty())
-            s.write((char*)string.data(), string.size());
+        s << string;
     }
 };
+//! Add a LimitedString wrapper around a const string (identity)
+template<size_t I> static inline const std::string& LimitedString(const std::string& str) { return str; }
+//! Add a LimitedString wrapper around a non-const string
```
promag: Missing periods in comments.

promag (commented):
ACK b0652ac. Everything works as expected.

Indeed a lot in one shot :).

sipa:
Rebased.

sipa:
Rebased.

src/serialize.h
```diff
@@ -446,6 +428,16 @@ class CFlatData
...
+//! Construct a FlatRange wrapper around a const vector.
+template<typename T> static inline const FlatRangeWrapper<const char> FlatVector(const T& t) { return FlatRangeWrapper<const char>((const char*)t.data(), (const char*)(t.data() + t.size())); }
+//! Construct a FlatRange wrapper around a non-const vector.
+template<typename T> static inline FlatRangeWrapper<char> FlatVector(T& t) { return FlatRangeWrapper<char>((char*)t.data(), (char*)(t.data() + t.size())); }
+//! Construct a FlatRange wrapper around a const POD and array types.
+template<typename T> static inline const FlatRangeWrapper<const char> FlatDataInner(const T* t, size_t len) { return FlatRangeWrapper<const char>((const char*)t, ((const char*)t) + len); }
+//! Construct a FlatRange wrapper around a non-const POD and array types.
+template<typename T> static inline FlatRangeWrapper<char> FlatDataInner(T* t, size_t len) { return FlatRangeWrapper<char>((char*)t, ((char*)t) + len); }
+//! Helper macro to easily serialize POD types.
+#define FLATDATA(x) FlatDataInner(&(x), sizeof(x))
```
ryanofsky: In commit "Generalize FlatData wrapper"

Though these changes don't make FLATDATA more dangerous than it was previously, the lack of type safety relying on C casts and sizeof here is a little scary. I experimented a little, and it seems this could be cleaned up with some simple changes I posted here: 6ef78bcd83dd6f88362dec29736811b762ad75eb. Could you take a look, and maybe incorporate these into the PR?

ryanofsky (commented):
Reviewed first several commits. Will continue reviewing.

src/serialize.h
```diff
@@ -468,23 +467,27 @@ class CVarInt
...
+    I &n;
 public:
-    explicit CCompactSize(uint64_t& nIn) : n(nIn) { }
+    explicit CompactSizeWrapper(I& nIn) : n(nIn) { }
 
     template<typename Stream>
-    void Serialize(Stream &s) const {
-        WriteCompactSize<Stream>(s, n);
+    void Unserialize(Stream& s) {
+        n = ReadCompactSize<Stream>(s);
```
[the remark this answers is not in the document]
sipa: Fixed.

src/serialize.h
```diff
@@ -468,23 +467,27 @@ class CVarInt
...
     }
 
     template<typename Stream>
-    void Unserialize(Stream& s) {
-        n = ReadCompactSize<Stream>(s);
+    void Serialize(Stream& s) const {
+        WriteCompactSize<Stream>(s, n);
```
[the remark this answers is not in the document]
sipa: Fixed.

src/serialize.h
```diff
@@ -493,9 +515,44 @@ class LimitedString
     }
 };
 
+/** Serialization wrapper class for big-endian integers.
+ *
+ * Use this wrapper around integer types that are stored in memory in native
```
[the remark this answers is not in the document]
sipa: Fixed.

src/serialize.h
sipa: I've incorporated part of your changes, but gone further and just added native support for serializing char arrays (without any wrapper). I think this is much cleaner now.

src/serialize.h
```diff
@@ -493,9 +515,44 @@ class LimitedString
...
+ * byte order, but serialized in big endian notation.
+ *
+ * Onlyy 16-bit types are supported for now.
```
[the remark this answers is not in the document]
sipa: Fixed.

src/serialize.h
```diff
@@ -493,9 +515,44 @@ class LimitedString
...
+ */
+template<typename I, size_t S> class BigEndianWrapper
+{
+protected:
+    I& m_val;
+public:
+    explicit BigEndianWrapper(I& val) : m_val(val)
+    {
+        static_assert(S == 2 || S == 4, "Unsupported BigEndian size");
```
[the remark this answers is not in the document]
sipa: Fixed.

src/netaddress.h
```diff
@@ -132,7 +132,7 @@ class CSubNet
         inline void SerializationOp(Stream& s, Operation ser_action) {
             READWRITE(network);
             READWRITE(FLATDATA(netmask));
-            READWRITE(FLATDATA(valid));
+            READWRITE(valid);
```
ryanofsky: In commit "Overhaul FLATDATA"

Might be worth splitting this change out into separate commit, or noting in commit message here that that this change is not backwards compatible on platforms where sizeof(bool) is not 1.

src/serialize.h
```diff
@@ -406,48 +415,36 @@ I ReadVarInt(Stream& is)
...
 }
 
-#define FLATDATA(obj) REF(CFlatData((char*)&(obj), (char*)&(obj) + sizeof(obj)))
 #define LIMITED_STRING(obj,n) REF(LimitedString< n >(REF(obj)))
 
-/** 
- * Wrapper for serializing arrays and POD.
- */
-class CFlatData
+/** Wrapper for serializing arrays and POD. */
```
ryanofsky: In commit "Overhaul FLATDATA"

Probably more accurate to say "wrapper for serializing char arrays". (Though in principle this could work with stream classes with read/write methods not taking char pointers.)

src/serialize.h
```diff
@@ -406,48 +415,36 @@ I ReadVarInt(Stream& is)
...
 
     template<typename Stream>
     void Unserialize(Stream& s)
     {
-        s.read(pbegin, pend - pbegin);
+        s.read(m_begin, m_end - m_begin);
     }
 };
+//! Construct a FlatRange wrapper around a const char vector.
+template<typename T> static inline const FlatRangeWrapper<const char> FlatVector(const T& t) { return FlatRangeWrapper<const char>(CharCast(t.data()), CharCast(t.data() + t.size())); }
```
ryanofsky: In commit "Overhaul FLATDATA"

Maybe call it CharVector or CharArray instead of FlatVector. FlatVector is kind of redundant because any vectors should be flat. But also the vector part is limiting because these functions can work for other types (std::array, std::string, SecureString, std::basic_string_view, etc)

src/serialize.h
sipa: Fixed.

src/serialize.h
```diff
@@ -470,23 +469,36 @@ class CVarInt
...
+        }
+        m_n = n;
     }
 
     template<typename Stream>
-    void Unserialize(Stream& s) {
-        n = ReadCompactSize<Stream>(s);
+    void Serialize(Stream& s) const
+    {
+        WriteCompactSize<Stream>(s, m_n);
```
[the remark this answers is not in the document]
sipa: Done.

src/serialize.h
```diff
@@ -495,9 +517,48 @@ class LimitedString
...
+    void Serialize(Stream& s) const
+    {
+        if (S == 2) ser_writedata16be(s, m_val);
+        if (S == 4) ser_writedata32be(s, m_val);
+    }
+
+    template<typename Stream>
+    void Unserialize(Stream& s)
+    {
+        if (S == 2) m_val = ser_readdata16be(s);
```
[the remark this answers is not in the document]
sipa: Done.

src/serialize.h
sipa: Done.

src/netaddress.h
sipa: Done.

src/serialize.h
```diff
@@ -573,6 +573,64 @@ template<typename I> class BigEndianWrapper
...
+        unsigned int nSize = ReadCompactSize(s);
+        unsigned int nMid = 0;
+        while (nMid < nSize) {
+            nMid += 5000000 / sizeof(value_type);
+            if (nMid > nSize) {
+                nMid = nSize;
+            }
+            m_vector.resize(nMid);
+            for (value_type& x : m_vector) {
+                s >> W<value_type>(x);
```
ryanofsky: In commit "Add custom vector-element serialization wrapper"

Doesn't this overwrite elements in the front of the vector each time through the while loop?

src/blockencodings.h
```diff
@@ -12,58 +12,70 @@
...
+    template <typename Stream> void Unserialize(Stream& s)
+    {
+        uint32_t lsb;
+        uint16_t msb;
+        s >> lsb >> msb;
+        m_int = (uint64_t(msb) << 32) | uint64_t(lsb);
     }
 };
 
+template<typename T>
```
ryanofsky: In commit "Convert blockencodings to new serialization"

Maybe move this up closer to TransactionCompressWrapper class

src/blockencodings.h
```diff
@@ -154,40 +156,18 @@ class CBlockHeaderAndShortTxIDs {
...
+    }
 
-        if (ser_action.ForRead())
-            FillShortTxIDSelector();
+    template <typename Stream>
+    inline void Unserialize(Stream& s)
+    {
+        static_assert(SHORTTXIDS_LENGTH == 6, "shorttxids serialization assumes 6-byte shorttxids");
+        s >> header >> nonce >> VectorApply<Uint48Wrapper>(shorttxids) >> prefilledtxn;
+        FillShortTxIDSelector();
```
ryanofsky: In commit "Convert blockencodings to new serialization"

Would you be opposed to adding mutable object access to SERIALIZE_METHODS so Serialize and Unserialize methods don't need to be split up? I could think of a number of ways to do this. Maybe easiest would be to stick a mutable pointer inside ser_action:

```
SERIALIZE_METHODS(obj) {
  READWRITE(obj.header, obj.nonce, VectorApply<Uint48Wrapper>(obj.shorttxids), obj.prefilledtxn);
  if (ser_action.ForRead()) ser_action.MutableObj()->FillShortTxId();
}
```

The pointer would be null when serializing.

ryanofsky (commented):
More review comments. I made it up to the "Convert blockencodings to new serialization" commit this time.

src/serialize.h
```diff
@@ -148,9 +148,21 @@ enum
     SER_GETHASH         = (1 << 2),
 };
 
+// Convert the reference base type to X, without changing constness or reference type.
```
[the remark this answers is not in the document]
sipa: Done.

src/serialize.h
```diff
@@ -167,6 +167,28 @@ enum
...
     }
 
+/**
+ * Implement the Serialize and Unserialize methods by delegating to a single templated
+ * static method that takes the to-be-(de)serialized object as a parameter. This approach
+ * has the advantage that the constness of the object becomes a template parameter, and
+ * thus allows a single implementation that sees the object as const for serializing
+ * and non-const for deserializing, without casts.
+ */
+#define SERIALIZE_METHODS(obj)                                                      \
```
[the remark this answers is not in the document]
sipa: Done!

src/serialize.h
```diff
@@ -898,16 +898,16 @@ void SerializeMany(Stream& s)
 }
 
 template<typename Stream, typename Arg>
-void SerializeMany(Stream& s, Arg&& arg)
+void SerializeMany(Stream& s, const Arg& arg)
```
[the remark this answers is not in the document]
sipa: Done.

src/serialize.h
```diff
@@ -519,19 +518,23 @@ class LimitedString
         if (size > Limit) {
             throw std::ios_base::failure("String length limit exceeded");
         }
-        string.resize(size);
-        if (size != 0)
-            s.read((char*)string.data(), size);
+        m_string.resize(size);
+        if (size != 0) {
+            s.read((char*)m_string.data(), size);
```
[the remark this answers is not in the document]
sipa: Done.

src/serialize.h
```diff
@@ -450,24 +449,29 @@ class CFlatData
...
     template<typename Stream>
-    void Unserialize(Stream& s) {
-        n = ReadVarInt<Stream,I>(s);
+    void Unserialize(Stream& s)
+    {
+        m_n = ReadVarInt<Stream,I>(s);
     }
 };
+//! Automatically construct a VarInt wrapper around the argument.
+template<typename I> static inline VarIntWrapper<typename std::remove_reference<I>::type> VARINT(I&& i) { return VarIntWrapper<typename std::remove_reference<I>::type>(i); }
```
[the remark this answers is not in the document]
sipa: The problem is that VARINT is called with temporaries as arguments, which is not true for the other ones. Either it's written as 4 cases, or using `std::remove_refence`.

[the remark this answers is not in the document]
sipa: Oh, you're right. An lvalue reference parameter binds to rvalue reference argument, so all good.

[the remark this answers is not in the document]
sipa: Done.

src/serialize.h
```diff
@@ -189,22 +170,6 @@ template<typename X> const X&& AsBaseType(const X&& x) { return std::move(x); }
  */
 #define READWRITEAS(obj, typ) (::SerReadWriteMany(s, ser_action, AsBaseType<typ>(obj)))
 
-/** 
- * Implement three methods for serializable objects. These are actually wrappers over
- * "SerializationOp" template, which implements the body of each class' serialization
- * code. Adding "ADD_SERIALIZE_METHODS" in the body of the class causes these wrappers to be
- * added as members. 
- */
-#define ADD_SERIALIZE_METHODS                                         \
```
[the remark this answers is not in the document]
sipa: Done.

src/serialize.h
```diff
@@ -573,6 +573,64 @@ template<typename I> class BigEndianWrapper
...
+    }
+
+    template<typename Stream>
+    void Unserialize(Stream& s)
+    {
+        m_vector.clear();
+        unsigned int nSize = ReadCompactSize(s);
+        unsigned int nMid = 0;
+        while (nMid < nSize) {
+            nMid += 5000000 / sizeof(value_type);
```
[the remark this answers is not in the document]
sipa: Done. The code was also totally broken, so I've rewritten it.

src/serialize.h
```diff
@@ -470,23 +469,37 @@ class CVarInt
...
     template<typename Stream>
-    void Unserialize(Stream& s) {
-        n = ReadCompactSize<Stream>(s);
+    void Serialize(Stream& s) const
+    {
+        WriteCompactSize<Stream>(s, m_n);
     }
 };
+//! Automatically construct a CompactSize wrapper around the argument.
+template<typename I> static inline CompactSizeWrapper<I> COMPACTSIZE(I& i) { return CompactSizeWrapper<I>(i); }
```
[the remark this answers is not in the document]
sipa: Done (and for other wrappers).

src/serialize.h
sipa: Yes, this was bogus. Thanks for pointing that out; fixed by rewriting,

src/serialize.h
```diff
@@ -406,48 +415,38 @@ I ReadVarInt(Stream& is)
...
-        pbegin = (char*)v.data();
-        pend = (char*)(v.data() + v.size());
-    }
-    template <unsigned int N, typename T, typename S, typename D>
-    explicit CFlatData(prevector<N, T, S, D> &v)
+    CharArrayWrapper(C* begin, size_t size) : m_begin(begin), m_size(size)
     {
-        pbegin = (char*)v.data();
-        pend = (char*)(v.data() + v.size());
+        static_assert(sizeof(C) == 1, "CharArrayWrapper only works for char types");
```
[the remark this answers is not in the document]
sipa: I'm not sure that wchar_t has a well-defined in-memory representation.

src/blockencodings.h
sipa: Done.

src/blockencodings.h
sipa: That's a neat trick. It seems a bit ugly to need to fake returning a mutable object nullptr. I have an alternative, but I'm not sure it's any better.:

```c++
template <typename T, typename F>
void IfUnserializer(T& obj, CSerActionUnserialize ser_action, const F& fn) { fn(obj); }
template <typename T, typename F>
void IfUnserialize(const T& obj, CSerActionSerialize ser_action, const F& fn) {}
#define IF_UNSERIALIZE(typ, obj, code) (::IfUnserialize<typ>(obj, ser_action, [&](typ& obj)code))
```

Which you'd then invoke using

```c++
     SERIALIZE_METHODS(SomeType, obj)
     {
          READWRITE(obj.member);
          IF_UNSERIALIZE(SomeType, obj, {obj.Cleanup();});
     }
```

src/wallet/wallet.h
```diff
@@ -381,42 +379,36 @@ class CWalletTx : public CMerkleTx
...
+        Init(nullptr);
+        char fSpent;
 
-            nTimeSmart = mapValue.count("timesmart") ? (unsigned int)atoi64(mapValue["timesmart"]) : 0;
-        }
+        s >> AsBaseType<CMerkleTx>(*this);
+        std::vector<CMerkleTx> vUnused; //!< Used to be vtxPrev
+        s >> vUnused >> mapValue >> vOrderForm >> fTimeReceivedIsTxTime >> nTimeReceived >> fFromMe >> fSpent;
+
+        strFromAccount = mapValue["fromaccount"];
```
ryanofsky: Could use std::move here.

ryanofsky (commented):
utACK b49e84ea6dd7e2c0aacb8546c8d8c2d99d1d8214. Finally reviewed the full change.

src/blockencodings.h
sipa: @ryanofsky I don't see how to get rid of passing in the type to IF_UNSERIALIZE.

src/serialize.h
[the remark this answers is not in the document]
sipa: I think I'm talking about something else.

If you serialize a 16-bit wchar_t on a big endian system using this class, the characters will be serialized using 2 big endian bytes. If you then deserialize it on a little endian system, you won't get the same wchar_t values back.

src/serialize.h
```diff
@@ -577,6 +577,66 @@ template<typename I> class BigEndianWrapper
...
+    {
+        m_vector.clear();
+        size_t size = ReadCompactSize(s);
+        size_t allocated = 0;
+        size_t deserialized = 0;
+        while (allocated < size) {
+            // For DoS prevention, do not blindly allocate as much as the stream claims to contain.
+            // Instead, allocate in 5MiB batches, so that an attacker actually needs to provide
+            // X MiB of data to make us allocate X+5 Mib.
+            allocated = std::min(size, allocated + 5000000 / sizeof(value_type));
```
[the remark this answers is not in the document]
sipa: Done.

src/blockencodings.h
```diff
@@ -12,17 +12,35 @@
...
+struct Uint48Wrapper {
+private:
+    I& m_int;
+public:
+    Uint48Wrapper(I& i) : m_int(i) {}
+    template <typename Stream> void Serialize(Stream& s) const
+    {
+        uint32_t lsb = m_int & 0xffffffff;
+        uint16_t msb = (m_int >> 32) & 0xffff;
+        s << lsb << msb;
```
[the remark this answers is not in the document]
sipa: Added static assert; I'd like to avoid runtime impact.

src/blockencodings.h
```diff
@@ -12,17 +12,35 @@
...
+    }
+    template <typename Stream> void Unserialize(Stream& s)
+    {
+        uint32_t lsb;
+        uint16_t msb;
+        s >> lsb >> msb;
+        m_int = (uint64_t(msb) << 32) | uint64_t(lsb);
```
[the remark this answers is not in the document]
sipa: Added static assert.

src/blockencodings.h
```diff
@@ -32,38 +50,31 @@ class BlockTransactionsRequest {
...
-                offset = indexes[j] + 1;
-            }
-        } else {
-            for (size_t i = 0; i < indexes.size(); i++) {
-                uint64_t index = indexes[i] - (i == 0 ? 0 : (indexes[i - 1] + 1));
-                READWRITE(COMPACTSIZE(index));
-            }
+        s << COMPACTSIZE(indexes_size);
+        for (size_t i = 0; i < indexes.size(); i++) {
+            const uint64_t index = indexes[i] - (i == 0 ? 0 : (indexes[i - 1] + 1));
```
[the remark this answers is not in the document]
sipa: I'd rather avoid runtime overhead.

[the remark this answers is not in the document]
ryanofsky: In that case, could add a comment like "this code will produce an invalid serialization if indices are not increasing."

src/blockencodings.h
ryanofsky: @ryanofsky I don't see how to get rid of passing in the type to IF_UNSERIALIZE.

You could pass it as a template parameter to CSerActionSerialize / CSerActionUnserialize and access it through ser_action.

ryanofsky (commented):
utACK a9770fbe6ef6d7d9bd5b61e4c51531af7cc2a1fc. Still looks good. Only minor changes since previous review, described in comments above.

laanwj:
Big concept ACK, happy to get rid of FLATDATA and similar ugly macros. This is a lot to review/test though, and reasonably high-risk.

src/undo.h
```diff
@@ -18,81 +18,49 @@
...
-            ::Serialize(s, REF(TxInUndoSerializer(&prevout)));
-        }
-    }
-
-    template <typename Stream>
-    void Unserialize(Stream& s) {
-        // TODO: avoid reimplementing vector deserializer
-        uint64_t count = 0;
-        ::Unserialize(s, COMPACTSIZE(count));
-        if (count > MAX_INPUTS_PER_BLOCK) {
```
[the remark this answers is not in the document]
sipa: It could use slightly more memory when deserializing an otherwise invalid object, but that shouldn't change behaviour otherwise - if the number of transaction undo objects doesn't match the number of transaction in the block, it's invalid anyway.

src/qt/recentrequeststablemodel.h
```diff
@@ -24,19 +24,19 @@ class RecentRequestEntry
...
-        READWRITE(recipient);
-
-        if (ser_action.ForRead())
-            date = QDateTime::fromTime_t(nDate);
+    template<typename Stream>
+    void Unserialize(Stream& s)
+    {
+        unsigned int nDate;
+        s >> this->nVersion >> id >> nDate >> recipient;
+        date = QDateTime::fromTime_t(nDate);
```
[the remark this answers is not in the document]
sipa: Introduced a wrapper for QDateTime instead.

src/wallet/wallet.h
sipa: Done.

src/qt/walletmodel.h
```diff
@@ -62,35 +62,32 @@ class SendCoinsRecipient
...
+    {
+        std::string sAddress, sLabel, sMessage, sPaymentRequest, sAuthenticatedMerchant;
+        s >> this->nVersion >> sAddress >> sLabel >> amount >> sMessage >> sPaymentRequest >> sAuthenticatedMerchant;
+        address = QString::fromStdString(sAddress);
+        label = QString::fromStdString(sLabel);
+        message = QString::fromStdString(sMessage);
+        if (!sPaymentRequest.empty()) {
+            paymentRequest.parse(QByteArray::fromRawData(sPaymentRequest.data(), sPaymentRequest.size()));
         }
+        authenticatedMerchant = QString::fromStdString(sAuthenticatedMerchant);
```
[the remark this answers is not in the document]
sipa: Done using wrappers for QString and proto.

sipa:
Rebased.

ryanofsky (commented):
utACK c3e505c8551a29d12edcebbed8a69f1e798823ef. Changes since previous review were rebase and addressing a few comments. Main one being new guiutil.h wrappers for qt strings, dates, and protobufs.

MarcoFalke:
Travis can not run the qt tests. Are they passing for you locally?

ryanofsky:
They passed for me locally. Current travis failure appears to be "/home/travis/.travis/job_stages: fork: Cannot allocate memory"

MarcoFalke:
I guess they time out and you could try setting the `QTEST_FUNCTION_TIMEOUT` env var to 600,000. Though, it might be worthwhile to check why they take longer than before.

ryanofsky:
Actually they are failing for me locally. Stack trace shows an infinite recursion and looks like:

```
(gdb) bt -50
#2355058 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355059 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355060 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355061 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355062 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355063 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355064 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355065 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355066 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355067 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355068 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355069 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355070 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355071 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355072 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355073 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355074 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355075 0x00000001000dbe89 in GUIUtil::AsStdString<QString const>(QString const&) (qstring=...) at qt/guiutil.h:285
#2355076 0x00000001000da01d in RecentRequestsTableModel::addNewRequest(SendCoinsRecipient const&) (qstring=...) at qt/guiutil.h:285
#2355077 0x00000001000da01d in RecentRequestsTableModel::addNewRequest(SendCoinsRecipient const&) (qstring=...) at qt/guiutil.h:285
#2355078 0x00000001000da01d in RecentRequestsTableModel::addNewRequest(SendCoinsRecipient const&) (ser_action=..., s=..., obj=...) at qt/walletmodel.h:69
#2355079 0x00000001000da01d in RecentRequestsTableModel::addNewRequest(SendCoinsRecipient const&) (s=..., this=0x7fffffffb728) at qt/walletmodel.h:66
#2355080 0x00000001000da01d in RecentRequestsTableModel::addNewRequest(SendCoinsRecipient const&) (a=..., os=...) at ./serialize.h:677
#2355081 0x00000001000da01d in RecentRequestsTableModel::addNewRequest(SendCoinsRecipient const&) (arg=..., s=...) at ./serialize.h:1030
#2355082 0x00000001000da01d in RecentRequestsTableModel::addNewRequest(SendCoinsRecipient const&) (arg=..., s=...) at ./serialize.h:1031
#2355083 0x00000001000da01d in RecentRequestsTableModel::addNewRequest(SendCoinsRecipient const&) (arg=@0x7fffffffb718: 1, s=...) at ./serialize.h:1031
#2355084 0x00000001000da01d in RecentRequestsTableModel::addNewRequest(SendCoinsRecipient const&) (arg=@0x7fffffffb710: 1, s=...) at ./serialize.h:1031
#2355085 0x00000001000da01d in RecentRequestsTableModel::addNewRequest(SendCoinsRecipient const&) (ser_action=..., s=...) at ./serialize.h:1049
#2355086 0x00000001000da01d in RecentRequestsTableModel::addNewRequest(SendCoinsRecipient const&) (s=..., obj=..., ser_action=...) at qt/recentrequeststablemodel.h:30
#2355087 0x00000001000da01d in RecentRequestsTableModel::addNewRequest(SendCoinsRecipient const&) (s=..., this=0x7fffffffb710) at qt/recentrequeststablemodel.h:28
#2355088 0x00000001000da01d in RecentRequestsTableModel::addNewRequest(SendCoinsRecipient const&) (a=..., os=...) at ./serialize.h:677
#2355089 0x00000001000da01d in RecentRequestsTableModel::addNewRequest(SendCoinsRecipient const&) (obj=..., this=0x7fffffffb600) at ./streams.h:399
#2355090 0x00000001000da01d in RecentRequestsTableModel::addNewRequest(SendCoinsRecipient const&) (this=0x1014511d0, recipient=...) at qt/recentrequeststablemodel.cpp:173
#2355091 0x00000001000d2a92 in ReceiveCoinsDialog::on_receiveButton_clicked() (this=0x7fffffffbf50) at qt/receivecoinsdialog.cpp:167
#2355092 0x000000010010dd95 in ReceiveCoinsDialog::qt_static_metacall(QObject*, QMetaObject::Call, int, void**) (_o=<optimized out>, _c=<optimized out>, _id=<optimized out>, _a=<optimized out>)
    at qt/moc_receivecoinsdialog.cpp:123
#2355093 0x000000010010e0e5 in ReceiveCoinsDialog::qt_metacall(QMetaObject::Call, int, void**) (this=0x7fffffffbf50, _c=QMetaObject::InvokeMetaMethod, _id=3, _a=0x7fffffffbbb0) at qt/moc_receivecoinsdialog.cpp:177
#2355094 0x00007ffff62b2ee0 in QMetaObject::activate(QObject*, int, int, void**) (sender=sender@entry=0x10163e380, signalOffset=<optimized out>, local_signal_index=local_signal_index@entry=2, argv=argv@entry=0x7fffffffbbb0) at kernel/qobject.cpp:3728
#2355095 0x00007ffff62b3537 in QMetaObject::activate(QObject*, QMetaObject const*, int, void**) (sender=sender@entry=0x10163e380, m=m@entry=0x7ffff7098a40 <QAbstractButton::staticMetaObject>, local_signal_index=local_signal_index@entry=2, argv=argv@entry=0x7fffffffbbb0) at kernel/qobject.cpp:3578
#2355096 0x00007ffff6f112b2 in QAbstractButton::clicked(bool) (this=this@entry=0x10163e380, _t1=false) at .moc/moc_qabstractbutton.cpp:303
#2355097 0x00007ffff6c73f44 in QAbstractButtonPrivate::emitClicked() (this=0x1016cc1d0) at widgets/qabstractbutton.cpp:534
#2355098 0x00007ffff6c745e1 in QAbstractButton::click() (this=0x10163e380) at widgets/qabstractbutton.cpp:992
#2355099 0x0000000100086686 in (anonymous namespace)::TestGUI() () at qt/test/wallettests.cpp:222
#2355100 0x00007ffff628f84a in QMetaMethod::invoke(QObject*, Qt::ConnectionType, QGenericReturnArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument) const (this=this@entry=0x7fffffffce98, object=object@entry=0x7fffffffd830, connectionType=Qt::DirectConnection, 
    connectionType@entry=4294954784, returnValue=..., val0=..., val1=..., val2=..., val3=..., val4=..., val5=..., val6=..., val7=..., val8=..., val9=...) at kernel/qmetaobject.cpp:2212
#2355101 0x00007ffff6294f0d in QMetaObject::invokeMethod(QObject*, char const*, Qt::ConnectionType, QGenericReturnArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument, QGenericArgument) (obj=0x7fffffffd830, member=member@entry=0x100ca1990 "walletTests", type=4294954784, 
    type@entry=Qt::DirectConnection, ret=..., val0=..., val1=..., val2=..., val3=..., val4=..., val5=..., val6=..., val7=..., val8=..., val9=...) at kernel/qmetaobject.cpp:1479
#2355102 0x00007ffff7fa408c in QTest::qInvokeTestMethod(char const*, char const*) (val9=..., val8=..., val7=..., val6=..., val5=..., val4=..., val3=..., val2=..., val1=..., val0=..., type=Qt::DirectConnection, member=0x100ca1990 "walletTests", obj=<optimized out>) at ../../include/QtCore/../../src/corelib/kernel/qobjectdefs.h:409
#2355103 0x00007ffff7fa408c in QTest::qInvokeTestMethod(char const*, char const*) (slot=0x100ca1990 "walletTests") at qtestcase.cpp:1953
#2355104 0x00007ffff7fa408c in QTest::qInvokeTestMethod(char const*, char const*) (slotName=0x100948698 "walletTests()", data=data@entry=0x0) at qtestcase.cpp:2082
#2355105 0x00007ffff7fa4bb8 in QTest::qExec(QObject*, int, char**) (testObject=0x7fffffffd830) at qtestcase.cpp:2390
#2355106 0x00007ffff7fa4bb8 in QTest::qExec(QObject*, int, char**) (testObject=0x7fffffffd830, argc=<optimized out>, argv=<optimized out>) at qtestcase.cpp:2652
#2355107 0x000000010005d9df in main(int, char**) (argc=1, argv=0x7fffffffdd78) at qt/test/test_main.cpp:99
```

src/qt/guiutil.h
```diff
@@ -239,6 +240,82 @@ namespace GUIUtil
...
+
+    template<typename Stream>
+    void Unserialize(Stream& s)
+    {
+        std::string str;
+        s >> str;
+        m_qstring = QString::fromStdString(std::move(str));
+    }
+};
+template<typename Q> AsStdStringWrapper<Q> AsStdString(Q& qstring) { return AsStdString<Q>(qstring); }
```
ryanofsky: This should say `return AsStdStringWrapper<Q>(qstring)` or just `return {qstring}`

sipa: Fixed! Thanks for figuring this out!

sipa:
Rebased.

sipa:
Removed the need for creating a vector with absolute values for serializing/deserializing BlockTransactionsRequest.

sipa:
Rebased.

sipa:
Rebased.

Christewart:
I ran this with #8469 -- which tests serialization code pretty extensively -- and it everything passed. 

If anyone wants to re-run they can do with this branch: https://github.com/Christewart/bitcoin/commits/pbt_noncastserial

sipa:
Rebased.

sipa:
I've reordered the commits again a bit, and made a small (but invasive) change to the design:
* Removed macro READWRITEAS (using `AsType<type>(obj)` instead)
* Added a `Wrap<wrapper>(obj)` function to later avoid the need for `COMPACTSIZE`, `VARINT`, and `LIMITED_STRING`, and removing the need for helper functions for each wrapper class.

Feel free to ignore the changes here; I'll keep taking small bits from it and submitting them as separate PRs.

laanwj:
Concept ACK

sipa:
@laanwj I'm going to keep extracting smaller changes from this PR into separate ones (like #12916, #12886, #12752, #12740, #12712, #12683, #12658). No need to review the whole thing.

MarcoFalke:
Needs rebase due to merge of #12924

MarcoFalke:
Travis output:

```
A wild circular dependency in the form of "qt/guiutil -> qt/walletmodel -> qt/guiutil" appears...

**Comment by jb55** [comment]
Chris Stewart <notifications@github.com> writes:
> I ran this with #8469 -- which tests serialization code pretty extensively -- and it everything passed. 

I gave this a spin with the property tests as well. Seems to still pass.
There's a recent change to bench/prevector that needs to be updated but
other than that it appears to work.

**Comment by sipa** [comment]
Rebased.

File: src/compressor.h [hunk 40]
```diff
@@ -33,79 +29,94 @@ uint64_t DecompressAmount(uint64_t nAmount);
...
-            script << OP_RETURN;
-            s.ignore(nSize);
-        } else {
-            script.resize(nSize);
-            s >> MakeSpan(script);
+        template<typename Stream> void Unserialize(Stream& s)
+        {
+            uint64_t v;
+            s >> VARINT(v);
+            m_val = DecompressAmount(v);
```
practicalswift: Make this conversion explicit since it changes signedness?

src/compressor.h
```diff
@@ -33,79 +29,94 @@ uint64_t DecompressAmount(uint64_t nAmount);
...
+{
+    template<typename I>
+    class Wrapper
+    {
+        I& m_val;
+    public:
+        explicit Wrapper(I& val) : m_val(val) {}
+        template<typename Stream> void Serialize(Stream& s) const
+        {
+            s << VARINT(CompressAmount(m_val));
```
practicalswift: Make this conversion explicit since it changes signedness?

src/compressor.h
```diff
@@ -33,79 +29,94 @@ uint64_t DecompressAmount(uint64_t nAmount);
...
-            uint64_t nVal = 0;
-            READWRITE(VARINT(nVal));
-            txout.nValue = DecompressAmount(nVal);
+    template<typename O>
+    class Wrapper
+    {
+    private:
+        O &txout;
+    public:
+        Wrapper(O &txoutIn) : txout(txoutIn) { }
```
practicalswift: Make this `explicit`? :-)

src/wallet/wallet.h
```diff
@@ -583,17 +579,13 @@ class CWalletKey
 
     explicit CWalletKey(int64_t nExpires=0);
 
-    ADD_SERIALIZE_METHODS;
-
-    template <typename Stream, typename Operation>
-    inline void SerializationOp(Stream& s, Operation ser_action) {
+    SERIALIZE_METHODS(CWalletKey, obj)
+    {
         int nVersion = s.GetVersion();
```
practicalswift: The scope of `nVersion` can be reduced?

src/qt/walletmodel.h
```diff
@@ -51,6 +51,65 @@ QT_BEGIN_NAMESPACE
...
+//! Wrapper class to serialize QString objects as std::strings.
+struct AsStdString
+{
+    template<typename Q>
+    class Wrapper
+    {
+    private:
+        Q& m_qstring;
+    public:
+        Wrapper(Q& qstring) : m_qstring(qstring) {}
```
practicalswift: `explicit`?

src/test/dbwrapper_tests.cpp
```diff
@@ -251,24 +251,26 @@ struct StringContentsSerializer {
     }
     StringContentsSerializer& operator+=(const StringContentsSerializer& s) { return *this += s.str; }
 
-    ADD_SERIALIZE_METHODS;
-
-    template <typename Stream, typename Operation>
-    inline void SerializationOp(Stream& s, Operation ser_action) {
-        if (ser_action.ForRead()) {
```
practicalswift: `ForRead` is no longer needed and can be removed from `serialize.h`?

src/compressor.h
sipa: The best I can do in this place is cast it to `I`, which isn't making anything interesting explicit, and will just hide the warning.

The issue is that CTxOut has a signed `int64_t`, but for UTXOs only ever contains unsigned values. I think the warning, if any, is appropriate, and should remain until the actual "issue" is resolved.

practicalswift: Makes sense! Thanks for the clarification.

practicalswift:
Concept ACK

@sipa has this been tested on a BE system?

laanwj:
This PR is really old by now, even though there was no disagreement. What is holding this up?

IMO. we should try to merge this after the 0.19 split-off in a few days.

practicalswift:
Has this been properly fuzz tested?

MarcoFalke:
@practicalswift Pretty much all of our fuzzers are  deserialization fuzzers

practicalswift:
@MarcoFalke Yes I know that :)  I was curious if these suggested serialization improvements had been subject to fuzz testing by the PR author - that is something different and hence my question :)

ryanofsky:
> @MarcoFalke Yes I know that :) I was curious if these suggested serialization improvements had been subject to fuzz testing by the PR author - that is something different and hence my question :)

I haven't paid much attention to fuzz testing, but I did spend a good amount of time reviewing this, and wonder if you could give more detail on the extra testing that would be useful here that goes beyond the normal testing.

It is probably safe to assume that if the author here went out of their way to do special testing, this would be mentioned in the description.

practicalswift:
> I haven't paid much attention to fuzz testing, but I did spend a good amount of time reviewing this, and wonder if you could give more detail on the extra testing that would be useful here that goes beyond the normal testing.

It would be useful to take this new code for a spin with our fuzzing test harnesses to make sure no new obvious issues are introduced.

This can be done using:

```
$ CC=clang CXX=clang++ ./configure --enable-fuzz --with-sanitizers=address,fuzzer,undefined
$ make
$ find src/test/fuzz/ -type f -executable
[…] then run the relevant fuzzers and observe any anomalies […]
```

> It is probably safe to assume that if the author here went out of their way to do special testing, this would be mentioned in the description.

TBH I'm not fuzz testing qualifies as special testing these days - in security critical project like ours it really should be standard procedure for this type of code :)

I've simply assumed that PR authors do fuzz testing routinely. I guess I'm mistaken :)

jb55:
> @practicalswift Has this been properly fuzz tested?

both me and @Christewart ran the serialization property-based tests against this at one point. Would be good to revisit.

practicalswift:
@jb55 Great to hear! I'll try to break it by fuzzing once rebased :)

sipa:
I'll gladly rebase again next week, or try to cherry-pick pieces of it.

sipa:
Rebased.

sipa:
Rebased.

sipa:
Rebased.

sipa:
Rebased.

MarcoFalke:
Is this going to be split into smaller bites?

sipa:
I certainly can, though I think all the standalone improvements have already been split off and merged independently.

What remains is introducing new serialization macros, converting all serializers to use them, and then deleting the old ones. I'm happy to split those up into separate PRs if you think that's useful.

MarcoFalke:
Fair enough. I was asking for the general public

sipa:
Rebased on top of (a rebased version of) #17591, to test on a BE platform.

src/qt/sendcoinsrecipient.h
```diff
@@ -33,8 +57,8 @@ class SendCoinsRecipient
     CAmount amount;
     // If from a payment request, this is used for storing the memo
     QString message;
-    // Keep the payment request around as a serialized string to ensure
-    // load/store is lossless.
+    // If building with BIP70 is disabled, keep the payment request around as
```
laanwj: This is no longer an "if" :smiley:

sipa: Oops, accident revert added in a rebase. Fixed.

sipa:
Now also rebased on top of #17599 to get functional tests run on BE.

practicalswift:
Could bring in the new improved deserialization fuzz harness from #17225 to fuzz test this? The new fuzz harness does roundtrip testing where possible and covers more types. Would be nice to see this new code fuzz tested :)

---

Fuzzing in three easy steps:

```
$ CC=clang CXX=clang++ ./configure --enable-fuzz --with-sanitizers=address,fuzzer,undefined
$ make
$ find src/test/fuzz/ -type f -executable
[…] then run the relevant fuzzers and observe any anomalies […]
```

sipa:
Rebased, and fixes a bug in the deserialization fuzzing code.

practicalswift:
This needs to be addressed to make the sanitizers happy :)

```
blockencodings.h:83:34: runtime error: implicit conversion from type 'int' of 
    value 65536 (32-bit, signed) to type 'Differential::Wrapper<std::vector<unsigned short, std::allocator<unsigned short> > >::value_type'
    (aka 'unsigned short') changed the value to 0 (16-bit, unsigned)
```

laanwj:
code review ACK 9250a087d2f450f37342010aea0a5d583eea508b

sipa:
I've added a new commit with a few style improvements, and a new approach for writing wrapper classes, as well a different way of doing the differential encoding of compact block indexes.

Unless people object, I'll use these modified approaches in follow-up PRs that carve out pieces out of this PR.

ryanofsky:
Just reviewed commit "Switch to different wrapping model + different differential vector" (7da03b3ce5f4be94555b15bf4668bc251a8faac4) and I think it is a nice improvement. Getting rid of the various `Format::Wrapper` inner classes and pulling them into a standalone `Wrapper` class eliminates redundant code. And another really nice part of the change is replacing the `Differential` struct that had lots of operator overloads with simpler value function transform objects.

I did have some ideas and suggestions from looking at this, but feel free to take or leave them:

1. "Wrapper" terminology doesn't seem helpful when it isn't specifically referring to the actual wrapper class. Would maybe refer to the various classes that have `Ser` and `Unser` methods as formatters or helpers instead of wrappers since they don't really wrap anything. Similarly might rename `Wrap` function to `SerializeAs` or `FormatAs` and rename `W` template arguments to `F` or `H`.
2. I don't think there's a reason to require `Ser` and `Unser` methods to be static. It should be no problem to just call `W()::Ser()` instead of `W::Ser()` inside the new `Wrapper` class, and not requiring the methods to be static could be beneficial if the formatter/helper objects are used other contexts in the future. For example if a compression formatter was used to serialize many elements in a data structure, maybe it could save some state to avoid repeating work.

sipa:
@ryanofsky Thanks, I'll see how hard it is to address the non-static methods suggestion. What so you think of the name "Using" instead of "Wrap" ?

sipa:
Changed to terminology "formatters", with function "Using<Formatter>", and made the Ser/Unser functions (permitted to be) non-static.

ryanofsky:
> @ryanofsky Thanks, I'll see how hard it is to address the non-static methods suggestion. What so you think of the name "Using" instead of "Wrap" ?

I think it'd be good, assuming it's not too much trouble to rename. I also realized the names I suggested yesterday `SerializeAs` or `FormatAs` don't really make sense given how the template argument is used. More accurate would be `SerializeWith`, `FormatWith`, or `SerializeUsing`. No strong opinions about any of this, though.

laanwj:
Travis error:
```
blockencodings.h:68:17: runtime error: implicit conversion from type 'int' of value 65536 (32-bit, signed) to type 'unsigned short' changed the value to 0 (16-bit, unsigned)

    #0 0x561bb3939927 in DifferentialUnserTransform<unsigned short>::operator()(unsigned short) /home/travis/build/bitcoin/bitcoin/build/bitcoin-x86_64-pc-linux-gnu/src/./blockencodings.h:68:17
    #1 0x561bb3938b15 in void VectorUsing<CompactSizeFormatter, DifferentialSerTransform<unsigned short>, DifferentialUnserTransform<unsigned short> >::Unser<CDataStream, std::vector<unsigned short, std::allocator<unsigned short> > >(CDataStream&, std::vector<unsigned short, std::allocator<unsigned short> >&) /home/travis/build/bitcoin/bitcoin/build/bitcoin-x86_64-pc-linux-gnu/src/./serialize.h:625:29
    #2 0x561bb39387c3 in void Wrapper<VectorUsing<CompactSizeFormatter, DifferentialSerTransform<unsigned short>, DifferentialUnserTransform<unsigned short> >, std::vector<unsigned short, std::allocator<unsigned short> > >::Unserialize<CDataStream>(CDataStream&) /home/travis/build/bitcoin/bitcoin/build/bitcoin-x86_64-pc-linux-gnu/src/./serialize.h:444:73
    #3 0x561bb39385bb in void Unserialize<CDataStream, Wrapper<VectorUsing<CompactSizeFormatter, DifferentialSerTransform<unsigned short>, DifferentialUnserTransform<unsigned short> >, std::vector<unsigned short, std::allocator<unsigned short> > >&>(CDataStream&, Wrapper<VectorUsing<CompactSizeFormatter, DifferentialSerTransform<unsigned short>, DifferentialUnserTransform<unsigned short> >, std::vector<unsigned short, std::allocator<unsigned short> > >&) /home/travis/build/bitcoin/bitcoin/build/bitcoin-x86_64-pc-linux-gnu/src/./serialize.h:708:7
```

sipa:
Too many changes have been introduced compared to this PR in #17850, #17896, #17957, #18021, and #18112,  and I'm not going to rebase this on top of them. I'll keep picking changes from the branch, though, but no need to keep this open.

PR #10786, opened 2017-07-10
Add PR description to merge commit in github-merge.py

sipa:
There is often some context given in PR descriptions that is missing from commits, and it may be worthwhile to retain that information in our history in git. This PR adds that information to the merge commit when created through `github-merge.py`.

We should also encourage people to provide as much information as possible in the PR commits themselves, but I believe that is an orthogonal issue. Individual commits don't need to have a description of the overall goal of a PR.

### DIFF
--- a/contrib/devtools/github-merge.py
+++ b/contrib/devtools/github-merge.py
@@ -175,6 +175,7 @@ def main():
     if info is None:
         exit(1)
     title = info['title'].strip()
+    body = info['body'].strip()
     # precedence order for destination branch argument:
     #   - command line argument
     #   - githubmerge.branch setting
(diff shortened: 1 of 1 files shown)


MarcoFalke:
ACK

contrib/devtools/github-merge.py
```diff
@@ -175,6 +175,7 @@ def main():
     if info is None:
         exit(1)
     title = info['title'].strip()
+    body = '  ' + info['body'].replace('\n', '\n  ')
```
promag: I think here should be `body = info['body'].strip()` like the above. And the body indentation should be below.

sipa: Agree, fixed.

promag:
ACK 475c08c.

For the record, after this, the PR description must be part of the review process. Don't forget that there can be typos, markdown, links for temporary resources, etc that will be written in stone. Maybe for future improvement is to print the body and prompt for confirmation.

jonasschnelli:
Great idea.
Tested ACK 475c08cb446371dd75fb37e067613effc0572442 (https://github.com/libbtc/libbtc/commit/7c6168c4c83403be1f64ff6a8baa94619fb57ee0)

laanwj:
Concept ACK

jonasschnelli:
I guess we are fine with HTML code in the PR description (=> goes then into git commit message)?
PRs like #10769 may include images (via HTML ``code).

PR #10788, opened 2017-07-11
[RPC] Fix addwitnessaddress by replacing ismine with producesignature

achow101:
Instead of using ismine to check whether an address can be spent by us, make the witness version of the script or address first and then use ProduceSignature with the DummySignatureCreator to check if we can
solve for the script.

This is to fix cases where we don't have all of the private keys (for something like a multisig address) but have the redeemscript so we can witnessify it.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -1067,11 +1067,15 @@ public:
     bool operator()(const CKeyID &keyID) {
         if (pwallet) {
             CScript basescript = GetScriptForDestination(keyID);
-            isminetype typ;
-            typ = IsMine(*pwallet, basescript, SIGVERSION_WITNESS_V0);
-            if (typ != ISMINE_SPENDABLE && typ != ISMINE_WATCH_SOLVABLE)
-                return false;
             CScript witscript = GetScriptForWitness(basescript);
+            SignatureData sigs;
+            // This check is to make sure that the script we created can actually be solved for and signed by us
+            // if we were to have the private keys. This is just to make sure that the script is valid and that,
+            // if found in a transaction, we would still accept and relay that transcation.
+            if (!ProduceSignature(DummySignatureCreator(pwallet), witscript, sigs) ||
+                !VerifyScript(sigs.scriptSig, witscript, &sigs.scriptWitness, MANDATORY_SCRIPT_VERIFY_FLAGS | SCRIPT_VERIFY_WITNESS_PUBKEYTYPE, DummySignatureCreator(pwallet).Checker())) {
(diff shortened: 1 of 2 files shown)


sipa:
Concept ACK, but you'll need to adapt the tests (there is a test in segwit.py which specifically tests for this behaviour).

achow101:
@sipa I have adapted segwit.py to reflect the new behavior.

jnewbery:
Travis failed on an unrelated call to `encryptwallet()` in wallet-dump.py: https://travis-ci.org/bitcoin/bitcoin/jobs/252653271 . I've restarted the job.

(`encryptwallet()` causes the node to shutdown, so I guess there's a race between the node shutting down and responding to the RPC. That's unrelated to this PR and is something we should think about fixing independently)

gmaxwell (approved):
ACK.

sipa:
@sdaftuar Would you mind reviewing this?

sipa:
utACK db5cd10bbe1688a5e05e558d1323f8cab923567c, but needs squash before merge.

sdaftuar (commented):
Seems like this is the right general direction to me (seems vastly simpler than what happens in IsMine()), though I have two concerns that I need to think more about before I can ACK:
- The protection against accidentally allowing an uncompressed pubkey in a segwit script via addwitnessaddress has now moved from the logic in IsMine() to the VerifyScript() call at the end of ProduceSignature.   First, I think that's generally surprising and I could imagine someone not understanding that it's important that we call VerifyScript at the end of ProduceSignature.  But second, I think it's unexpected that the script flags passed to VerifyScript in ProduceSignature (which are just our regular policy script flags) are what we're relying on for this wallet behavior.  The policy script flags are not something that generally affects our wallet, so I think it'd be unexpected if someone tweaked theirs and got different wallet behavior as a result.

- Along the lines of the above, it seems like we could use some tests to ensure that IsMine() interacts with ProduceSignature() as expected.  In the above example of someone tweaking their policy script flags (eg to turn off SCRIPT_VERIFY_WITNESS_PUBKEYTYPE) then I think you could get into a situation where you could add an address to your wallet that fails IsMine().  This is pretty edge-case, but I want to spend more time thinking about whether other kinds of inconsistencies could exist between IsMine() and this ProduceSignature(...) method.

sipa:
@sdaftuar Would an explicit VerifyScript call (with consensus + compressed pubkey flags) after ProduceSignature help alleviate your concerns?

achow101:
I have updated this with @sipa's suggestion and squashed.

laanwj:
utACK e222dc2

PR #10790, opened 2017-07-11
Use vector::data() instead of &vch[0] in base58.cpp

meshcollider:
`&vch[0]` has undefined behavior for empty vectors, c++11 introduced `data()` as a safer alternative

`EncodeBase58(const std::vector<unsigned char>& vch)` didn't check for empty vectors before using `&vch[0]` so suggest replacing with data() in this file, also makes it consistent with base58_tests.cpp


laanwj:
Have you checked that this doesn't overlap with #9804? If it does, please review that one instead.

MeshCollider:
Ah it does, missed that sorry. Closing this then.

PR #10793, opened 2017-07-11
Changing &var[0] to var.data()

meshcollider:
This just continues the work of https://github.com/bitcoin/bitcoin/pull/9804

Modifies a lot of `&vector[]`'s to `vector.data()`'s across all the files including tests, just the stuff that 9804 missed

### DIFF
--- a/src/qt/test/paymentservertests.cpp
+++ b/src/qt/test/paymentservertests.cpp
@@ -24,7 +24,7 @@ X509 *parse_b64der_cert(const char* cert_data)
 {
     std::vector<unsigned char> data = DecodeBase64(cert_data);
     assert(data.size() > 0);
-    const unsigned char* dptr = &data[0];
+    const unsigned char* dptr = data.data();
     X509 *cert = d2i_X509(nullptr, &dptr, data.size());
     assert(cert);
     return cert;
@@ -43,7 +43,7 @@ static SendCoinsRecipient handleRequest(PaymentServer* server, std::vector<unsig
     // Write data to a temp file:
     QTemporaryFile f;
     f.open();
-    f.write((const char*)&data[0], data.size());
+    f.write((const char*)data.data(), data.size());
     f.close();
 
     // Create a QObject, install event filter from PaymentServer
@@ -139,7 +139,7 @@ void PaymentServerTests::paymentServerTests()
 
     // Contains a testnet paytoaddress, so payment request network doesn't match client network:
     data = DecodeBase64(paymentrequest1_cert2_BASE64);
-    byteArray = QByteArray((const char*)&data[0], data.size());
+    byteArray = QByteArray((const char*)data.data(), data.size());
     r.paymentRequest.parse(byteArray);
     // Ensure the request is initialized, because network "main" is default, even for
     // uninitialized payment requests and that will fail our test here.
@@ -148,7 +148,7 @@ void PaymentServerTests::paymentServerTests()
 
     // Expired payment request (expires is set to 1 = 1970-01-01 00:00:01):
     data = DecodeBase64(paymentrequest2_cert2_BASE64);
-    byteArray = QByteArray((const char*)&data[0], data.size());
+    byteArray = QByteArray((const char*)data.data(), data.size());
     r.paymentRequest.parse(byteArray);
     // Ensure the request is initialized
     QVERIFY(r.paymentRequest.IsInitialized());
@@ -159,7 +159,7 @@ void PaymentServerTests::paymentServerTests()
     // 9223372036854775807 (uint64), 9223372036854775807 (int64_t) and -1 (int32_t)
     // -1 is 1969-12-31 23:59:59 (for a 32 bit time values)
     data = DecodeBase64(paymentrequest3_cert2_BASE64);
-    byteArray = QByteArray((const char*)&data[0], data.size());
+    byteArray = QByteArray((const char*)data.data(), data.size());
     r.paymentRequest.parse(byteArray);
     // Ensure the request is initialized
     QVERIFY(r.paymentRequest.IsInitialized());
@@ -170,7 +170,7 @@ void PaymentServerTests::paymentServerTests()
     // 9223372036854775808 (uint64), -9223372036854775808 (int64_t) and 0 (int32_t)
     // 0 is 1970-01-01 00:00:00 (for a 32 bit time values)
     data = DecodeBase64(paymentrequest4_cert2_BASE64);
-    byteArray = QByteArray((const char*)&data[0], data.size());
+    byteArray = QByteArray((const char*)data.data(), data.size());
     r.paymentRequest.parse(byteArray);
     // Ensure the request is initialized
     QVERIFY(r.paymentRequest.IsInitialized());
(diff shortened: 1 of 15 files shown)


promag:
#9804 should cherry pick this one right?

MeshCollider:
#9804 has just been merged

src/rpc/server.cpp
```diff
@@ -444,7 +444,7 @@ static inline JSONRPCRequest transformNamedArguments(const JSONRPCRequest& in, c
     const std::vector<UniValue>& values = in.params.getValues();
     std::unordered_map<std::string, const UniValue*> argsIn;
     for (size_t i=0; i<keys.size(); ++i) {
-        argsIn[keys[i]] = &values[i];
+        argsIn[keys[i]] = values.data() + i;
```
jtimon: Is this really clearer?

src/test/bloom_tests.cpp
```diff
@@ -154,8 +154,8 @@ BOOST_AUTO_TEST_CASE(bloom_match)
     COutPoint prevOutPoint(uint256S("0x90c122d70786e899529d71dbeba91ba216982fb6ba58f3bdaab65e73b7e9260b"), 0);
     {
         std::vector<unsigned char> data(32 + sizeof(unsigned int));
-        memcpy(&data[0], prevOutPoint.hash.begin(), 32);
-        memcpy(&data[32], &prevOutPoint.n, sizeof(unsigned int));
+        memcpy(data.data(), prevOutPoint.hash.begin(), 32);
+        memcpy(data.data()+32, &prevOutPoint.n, sizeof(unsigned int));
```
jtimon: same here

src/test/skiplist_tests.cpp
```diff
@@ -20,13 +20,13 @@ BOOST_AUTO_TEST_CASE(skiplist_test)
 
     for (int i=0; i<SKIPLIST_LENGTH; i++) {
         vIndex[i].nHeight = i;
-        vIndex[i].pprev = (i == 0) ? NULL : &vIndex[i - 1];
+        vIndex[i].pprev = (i == 0) ? NULL : (vIndex.data() + i - 1);
```
jtimon: same here

src/test/skiplist_tests.cpp
```diff
@@ -20,13 +20,13 @@ BOOST_AUTO_TEST_CASE(skiplist_test)
...
         vIndex[i].BuildSkip();
     }
 
     for (int i=0; i<SKIPLIST_LENGTH; i++) {
         if (i > 0) {
-            BOOST_CHECK(vIndex[i].pskip == &vIndex[vIndex[i].pskip->nHeight]);
+            BOOST_CHECK(vIndex[i].pskip == vIndex.data() + vIndex[i].pskip->nHeight);
```
jtimon: here you aren't even consistent

src/test/skiplist_tests.cpp
```diff
@@ -37,9 +37,9 @@ BOOST_AUTO_TEST_CASE(skiplist_test)
         int from = InsecureRandRange(SKIPLIST_LENGTH - 1);
         int to = InsecureRandRange(from + 1);
 
-        BOOST_CHECK(vIndex[SKIPLIST_LENGTH - 1].GetAncestor(from) == &vIndex[from]);
-        BOOST_CHECK(vIndex[from].GetAncestor(to) == &vIndex[to]);
-        BOOST_CHECK(vIndex[from].GetAncestor(0) == &vIndex[0]);
+        BOOST_CHECK(vIndex[SKIPLIST_LENGTH - 1].GetAncestor(from) == vIndex.data() + from);
```
jtimon: well, you get the idea, if the index was other than 0, it seems to me that the change makes things worse.

jtimon:
Fast review ACK modulo nits.

src/test/skiplist_tests.cpp
dcousens: what lacks consistency?

jtimon: One time vIndex is accessed by ```vIndex.data() + n``` and another with ```vIndex[m]```

MeshCollider: That's because one is accessing the element not a pointer...

MeshCollider:
@jtimon this is consistent with how #9804 did it

src/test/skiplist_tests.cpp
dcousens: The `.data()` usage here however,  is not really inline with #9804.
#9804 replaced usage of `&v[0]` and `&v[N]` for specifying a slice/range.
In this change above,  you are specifically referring to the memory location of 1 element,  which isn't an rvalue... probably unnecessary to use `.data()`.

src/test/bloom_tests.cpp
dcousens: IMHO this is clearer

src/test/skiplist_tests.cpp
MeshCollider: Isn't it still preferable to access the memory location of an element this way?

dcousens: @MeshCollider the advantage of `.data()` is that it is defined to return `nullptr` if `.empty()`,  compared to `&v[0]` potentially de-referencing unitialized memory.
In this event,  we assume `vIndex` is not empty, as you are de-referencing `vIndex[i]`,  therefore, it is probably not as necessary.

MeshCollider:
Reverted the changes where specific elements were accessed, thanks @jtimon and @dcousens

MeshCollider:
Rebased

MeshCollider:
Rebased

src/qt/test/paymentservertests.cpp
```diff
@@ -43,7 +43,7 @@ static SendCoinsRecipient handleRequest(PaymentServer* server, std::vector<unsig
     // Write data to a temp file:
     QTemporaryFile f;
     f.open();
-    f.write((const char*)&data[0], data.size());
+    f.write((const char*) data.data(), data.size());
```
laanwj: nit: Please remove space after the case here

laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/10793/commits/db0e991fb7975c071818b058376644be08ced343

src/qt/test/paymentservertests.cpp
MarcoFalke: @MeshCollider You can pipe the diff through clang-format to prevent such nits in the future. See dev notes.

MeshCollider: Fixed, thanks :)

src/test/bloom_tests.cpp
laanwj: If data is empty, I'd say copying to data()+32 is U/B no matter what.

src/rpc/server.cpp
laanwj: Agree with @jtimon here. This PR is titled `Changing &var[0] to var.data()`, this seems to overshoot that scope. Addressing out of an vector is UB no matter what, so this doesn't fix any potential issues, and isn't more readable IMO either.

laanwj:
The commit message is weird now :)

MeshCollider:
Argh fixed, thanks :)

PR #9171, opened 2016-11-16
Introduce auxiliary block requests

jonasschnelli:
This PR will allow auxiliary block requests. This will be required to run SPV wallets as well as this solves some other interesting use cases.

### CAuxiliaryBlockRequest
* A new "lock-free" class that handles the auxiliary blocks-download/processing
* Contains a single global CAuxiliaryBlockRequest pointer (switching to a queue would be trivial).

### SyncTransaction signal
* A new parameter will be added that defines if the transactions are coming from a validated or non-validated block

### New RPC call "requestblocks"
Allows to request blocks, if available on disk, they will be optionally passed tough the SyncTransaction signal (Wallet/ZMQ). If some or all of the blocks are not available, they will be downloaded (prioritized over "normal" IBD downloads).
The blocks will be downloaded in parallel using the current block download mechanisms.
`CAuxiliaryBlockRequest` will ensure that the blocks get processes in the correct order (as requested).

If someone has a good testplan for this, please stand up.

PR #10794, opened 2017-07-11
Add simple light-client mode (RPC only)

jonasschnelli:
This adds a simple light client mode (RPC only, no wallet support).

With this PR, It is possible to disable auto-request-blocks by passing in `-autorequestblocks=0`.
In that mode, one can request out-of-band blocks by calling `requestblock add ["<blockhash>", ...]`.
Those blocks will then be requested/downloaded and can be loaded via `getblock` (and they will also be passed through ZMQ).

This allows a very simple light-client mode ideally when you already have a validated peer in your trusted network.

This is also a reviewable step towards light client mode for the wallet (which will ultimately allow process separation o the wallet).

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -114,6 +114,7 @@ namespace {
         const CBlockIndex* pindex;                               //!< Optional.
         bool fValidatedHeaders;                                  //!< Whether this block has validated headers at the time of request.
         std::unique_ptr<PartiallyDownloadedBlock> partialBlock;  //!< Optional, used for CMPCTBLOCK downloads
+        bool priorityRequest;                                    //!< Whether its a priority download
     };
     std::map<uint256, std::pair<NodeId, std::list<QueuedBlock>::iterator> > mapBlocksInFlight;
 
@@ -137,6 +138,14 @@ namespace {
     MapRelay mapRelay;
     /** Expiration-time ordered list of (expire time, relay map entry) pairs, protected by cs_main). */
     std::deque<std::pair<int64_t, MapRelay::iterator>> vRelayExpiration;
+
+    struct PriorityBlockRequest
+    {
+        const CBlockIndex* pindex;
+        bool downloaded;
+    };
+
+    std::vector<PriorityBlockRequest> blocksToDownloadFirst;
 } // namespace
 
 namespace {
@@ -309,9 +318,14 @@ void PushNodeVersion(CNode *pnode, CConnman* connman, int64_t nTime)
 }
 
 // Requires cs_main.
-// Returns a bool indicating whether we requested this block.
+// Returns a MarkBlockAsReceivedResult struct to indicating whether we requested this block and if it was via the priority request queue
 // Also used if a block was /not/ received and timed out or started with another peer
-bool MarkBlockAsReceived(const uint256& hash) {
+struct MarkBlockAsReceivedResult
+{
+    bool fRequested;
+    bool fPriorityRequest;
+ };
+MarkBlockAsReceivedResult MarkBlockAsReceived(const uint256& hash) {
     std::map<uint256, std::pair<NodeId, std::list<QueuedBlock>::iterator> >::iterator itInFlight = mapBlocksInFlight.find(hash);
     if (itInFlight != mapBlocksInFlight.end()) {
         CNodeState *state = State(itInFlight->second.first);
@@ -325,19 +339,28 @@ bool MarkBlockAsReceived(const uint256& hash) {
             // First block on the queue was received, update the start download time for the next one
             state->nDownloadingSince = std::max(state->nDownloadingSince, GetTimeMicros());
         }
-        state->vBlocksInFlight.erase(itInFlight->second.second);
+        bool priorityRequest = itInFlight->second.second->priorityRequest;
         state->nBlocksInFlight--;
         state->nStallingSince = 0;
+        if (priorityRequest) {
+            // mark as downloaded
+            auto it = std::find_if(blocksToDownloadFirst.begin(), blocksToDownloadFirst.end(),  [&itInFlight](const PriorityBlockRequest &r) { return r.pindex == itInFlight->second.second->pindex; });
+            if (it != blocksToDownloadFirst.end()) {
+                (*it).downloaded = true;
+            }
+        }
+        state->vBlocksInFlight.erase(itInFlight->second.second);
         mapBlocksInFlight.erase(itInFlight);
-        return true;
+
+        return {true, priorityRequest};
     }
-    return false;
+    return {false, false};
 }
 
 // Requires cs_main.
 // returns false, still setting pit, if the block was already in flight from the same peer
 // pit will only be valid as long as the same cs_main lock is being held
-bool MarkBlockAsInFlight(NodeId nodeid, const uint256& hash, const CBlockIndex* pindex = nullptr, std::list<QueuedBlock>::iterator** pit = nullptr) {
+bool MarkBlockAsInFlight(NodeId nodeid, const uint256& hash, const CBlockIndex* pindex = nullptr, std::list<QueuedBlock>::iterator** pit = nullptr, bool priorityRequest = false) {
     CNodeState *state = State(nodeid);
     assert(state != nullptr);
 
@@ -354,7 +377,7 @@ bool MarkBlockAsInFlight(NodeId nodeid, const uint256& hash, const CBlockIndex* 
     MarkBlockAsReceived(hash);
 
     std::list<QueuedBlock>::iterator it = state->vBlocksInFlight.insert(state->vBlocksInFlight.end(),
-            {hash, pindex, pindex != nullptr, std::unique_ptr<PartiallyDownloadedBlock>(pit ? new PartiallyDownloadedBlock(&mempool) : nullptr)});
+            {hash, pindex, pindex != nullptr, std::unique_ptr<PartiallyDownloadedBlock>(pit ? new PartiallyDownloadedBlock(&mempool) : nullptr), priorityRequest});
     state->nBlocksInFlight++;
     state->nBlocksInFlightValidHeaders += it->fValidatedHeaders;
     if (state->nBlocksInFlight == 1) {
@@ -470,10 +493,12 @@ bool PeerHasHeader(CNodeState *state, const CBlockIndex *pindex)
 }
 
 /** Update pindexLastCommonBlock and add not-in-flight missing successors to vBlocks, until it has
- *  at most count entries. */
-void FindNextBlocksToDownload(NodeId nodeid, unsigned int count, std::vector<const CBlockIndex*>& vBlocks, NodeId& nodeStaller, const Consensus::Params& consensusParams) {
+ *  at most count entries.
+ *  returns true if priority downloads where used
+ */
+bool FindNextBlocksToDownload(NodeId nodeid, unsigned int count, std::vector<const CBlockIndex*>& vBlocks, NodeId& nodeStaller, const Consensus::Params& consensusParams) {
     if (count == 0)
-        return;
+        return false;
 
     vBlocks.reserve(vBlocks.size() + count);
     CNodeState *state = State(nodeid);
@@ -482,9 +507,22 @@ void FindNextBlocksToDownload(NodeId nodeid, unsigned int count, std::vector<con
     // Make sure pindexBestKnownBlock is up to date, we'll need it.
     ProcessBlockAvailability(nodeid);
 
+    if (!blocksToDownloadFirst.empty()) {
+        for (const PriorityBlockRequest &r: blocksToDownloadFirst) {
+            if (r.downloaded) continue;
+            if (r.pindex && state->pindexBestKnownBlock != nullptr && state->pindexBestKnownBlock->nHeight >= r.pindex->nHeight && !mapBlocksInFlight.count(r.pindex->GetBlockHash())) {
+                vBlocks.push_back(r.pindex);
+                if (vBlocks.size() == count) {
+                    break;
+                }
+            }
+        }
+        return true;
+    }
+
     if (state->pindexBestKnownBlock == nullptr || state->pindexBestKnownBlock->nChainWork < chainActive.Tip()->nChainWork || state->pindexBestKnownBlock->nChainWork < nMinimumChainWork) {
         // This peer has nothing interesting.
-        return;
+        return false;
     }
 
     if (state->pindexLastCommonBlock == nullptr) {
@@ -497,7 +535,7 @@ void FindNextBlocksToDownload(NodeId nodeid, unsigned int count, std::vector<con
     // of its current tip anymore. Go back enough to fix that.
     state->pindexLastCommonBlock = LastCommonAncestor(state->pindexLastCommonBlock, state->pindexBestKnownBlock);
     if (state->pindexLastCommonBlock == state->pindexBestKnownBlock)
-        return;
+        return false;
 
     std::vector<const CBlockIndex*> vToFetch;
     const CBlockIndex *pindexWalk = state->pindexLastCommonBlock;
@@ -526,11 +564,11 @@ void FindNextBlocksToDownload(NodeId nodeid, unsigned int count, std::vector<con
         for (const CBlockIndex* pindex : vToFetch) {
             if (!pindex->IsValid(BLOCK_VALID_TREE)) {
                 // We consider the chain that this peer is on invalid.
-                return;
+                return false;
             }
             if (!State(nodeid)->fHaveWitness && IsWitnessEnabled(pindex->pprev, consensusParams)) {
                 // We wouldn't download this block or its descendants from this peer.
-                return;
+                return false;
             }
             if (pindex->nStatus & BLOCK_HAVE_DATA || chainActive.Contains(pindex)) {
                 if (pindex->nChainTx)
@@ -543,11 +581,11 @@ void FindNextBlocksToDownload(NodeId nodeid, unsigned int count, std::vector<con
                         // We aren't able to fetch anything, but we would be if the download window was one larger.
                         nodeStaller = waitingfor;
                     }
-                    return;
+                    return false;
                 }
                 vBlocks.push_back(pindex);
                 if (vBlocks.size() == count) {
-                    return;
+                    return false;
                 }
             } else if (waitingfor == -1) {
                 // This is the first already-in-flight block.
@@ -555,6 +593,7 @@ void FindNextBlocksToDownload(NodeId nodeid, unsigned int count, std::vector<con
             }
         }
     }
+    return false;
 }
 
 } // namespace
@@ -2667,7 +2706,8 @@ bool static ProcessMessage(CNode* pfrom, const std::string& strCommand, CDataStr
             LOCK(cs_main);
             // Also always process if we requested the block explicitly, as we may
             // need it even though it is not a candidate for a new best tip.
-            forceProcessing |= MarkBlockAsReceived(hash);
+            MarkBlockAsReceivedResult result = MarkBlockAsReceived(hash);
+            forceProcessing |= result.fRequested;
             // mapBlockSource is only used for sending reject messages and DoS scores,
             // so the race between here and cs_main in ProcessNewBlock is fine.
             mapBlockSource.emplace(hash, std::make_pair(pfrom->GetId(), true));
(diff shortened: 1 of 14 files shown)


jonasschnelli:
Contains overhauled parts from #9483

MIGUELWAXX:
cheers

ryanofsky:
I'd like to help out here, and I've spent literally days reviewing previous iterations of this change (#9076 #9483 #9171), but I can't figure out if this is going anywhere and if providing more review now is a good use of time.

As I've mentioned previously, I don't think the auxiliary block download class design is great, because it duplicates functionality from the networking layer in a wholly different part of the code, and gives the wallet too much control & responsibility over low level p2p details in the case where it's used to let the wallet code prioritize which blocks to download. I've tried to describe what I think would be better alternatives in my previous review comments, like https://github.com/bitcoin/bitcoin/pull/9483#issuecomment-278799990.

@jonasschnelli, assuming you don't want to take my previous suggestions, and do want to stick with the current design, I think it would be helpful if you could reach out to some other reviewers and get some concept acks for your current approach. It might help to put together some kind of design document, or at least a detailed comment to the top of `auxiliaryblockrequest.h` that lays out how the class works and what your future plans for it are in as much detail as possible, so someone can understand how it's intended to be used without having to wade through all the PRs.

MIGUELWAXX:
cheers

jonasschnelli:
@ryanofsky:
Thank you very much for the reviews in #9483. I implemented almost all of your suggesting and some of them where really great.
However, I think the current design of having a dedicated class (`CauxiliaryBlockRequest`) makes sense, because...

* An out-of-band (auxiliary) block request in an object, you could have multiple in parallel (assume multiwallet, etc.), in future, we may want to serialise them to disk, etc. Therefore I think it should be capsulated in its own class.

* I think the wallet needs control. For a light-client mode, the wallet is the one that tells the node what it needs. More or less, wallet tells node: "I need block A to E, give them to me as fast as possible", then node may call back "I have block A already ready (on disk), rest will follow soon", etc. Wallet then may say: "Okay, thats enough, you can stop at B already", etc.

* Having the implementation in a designated file makes code overview simpler. It can result in having slightly more lines of code but it should worth it. Having another `main.cpp` for everything that is network related should be avoided now, early enough. I don't think we have a lot of duplicated code, or can you point out what would be more compact if we would implement it directly in `net_processing.cpp`?

* Last but not least, it allows to have a patch-set the requires less rebasing. And this is an important one to me. Eventually, this will be something that "runs along" with the master branch until there has been reasonable amount of testing before it gets eventually merged.

jonasschnelli:
Followed yesterdays discussion we had in #bitcoin-core-dev. Removed the `CAuxiliaryBlockRequests` and added a `std::map blocksToDownloadFirst`.
Such manually added, priority block downloads will not trigger `ActivateBestChain`.

This PR now also adds a new signal `BlockProcessed()`.

The scope of this PR is not to make the block download interface flexible for multiple "users" (like the validation and eventually the light-client wallet).

jonasschnelli:
fixed @ryanofsky points.
This does pass travis now. Thanks in advance for reviews...

ryanofsky (commented):
Change seems pretty straightforward. Main thing missing is a test for the new RPC call.

It would be useful to get concept ACKs from @gmaxwell and @sipa to make sure concerns from the earlier design discussion (https://botbot.me/freenode/bitcoin-core-dev/msg/88437543/) are addressed now.

I don't have a very clear idea of how this functionality is going to be used, and what future plans for it are, so I personally wouldn't mind seeing a list of what the next steps & future prs might look like, or knowing some applications that could use the new option & RPC. But probably these things are clearer to others.

jonasschnelli:
Thanks @ryanofsky. Agree about required conceptual ACKs from other devs.
The long term goal was sketched in #9483, basically a light client mode for Bitcoin-Core that would allow node/wallet separation.

ryanofsky (commented):
utACK 1794a11037150b0d2c6ba2eb9f06e310f51f7c1c. Change looks great: code & test are straightforward and commit history is well thought out.

I still can't figure out if new RPC is actually supposed to be useful for something other than debugging/testing. Seems fine if it isn't, though, because it's a simple wrapper around the functions for downloading blocks of order, which obviously are useful for SPV.

jonasschnelli:
Overhauled and fixed @ryanofsky points (mostly nits), also removed the new (unused) signal.

src/rpc/blockchain.cpp
```diff
@@ -1532,6 +1537,63 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+    else if (request.params[0].get_str() == "status") {
+        UniValue ret(UniValue::VARR);
+        std::map<uint256, const CBlockIndex*> prioBlocks = GetPriorityDownloads();
+        for(auto& kv: prioBlocks) {
+            ret.push_back(kv.first.ToString());
+        }
+        return ret;
+    }
+    else if (request.params[0].get_str() == "add") {
+        if (request.params.size() < 2) {
```
ryanofsky: In commit "[RPC] Add requestblocks - a simple way to priorize block downloads"

Don't really need this check, because the get_array call below will trigger it's own error about the value not being an array. But if you prefer this error, you should change the check to request.params[1].isNull() to avoid making a distinction between a null and a missing value (#10281, #10783).

ryanofsky (commented):
utACK 9723672ced8b48eeee4847e6c0e0ed39833c7eaf. Changes since last review were dropping a commit that added a `BlockProcessed` signal, changing order of other commits, changing brace styles, and making various suggested tweaks.

jonasschnelli:
Overhauled first commit to make sure we request blocks in order:
https://github.com/bitcoin/bitcoin/pull/10794/commits/5927f7f1d6c7eb1bc7fb7a39dfda277ad5e4474c#diff-eff7adeaec73a769788bb78858815c91R483

jonasschnelli:
* Overhauled once again. Added the processing logic to ensure blocks are processed in order (makes it much simpler to process by "the other side").
* Added a new main signal for the processing (reusing BlockChecked seems wrong).
* Priority requests are now pushed through signal

This is now tested on my SPV branch and could be the first reviewable step towards light clients / wallet process separation.

jtimon:
Needs rebase

nopara73:
@ryanofsky Reflecting to your usefullness concern:  
I wrote a full block downloading SPV wallet ([HiddenWallet](https://github.com/nopara73/HiddenWallet/)), the main reason is, because without such architecture [private transaction retrieval is hard](https://groups.google.com/forum/#!msg/bitcoinj/Ys13qkTwcNg/9qxnhwnkeoIJ), if not impossible.  
When some form of light functionality gets into Core I would personally consider porting HiddenWallet to on top of Core, since I cannot compete with Core's speed and stability.

ryanofsky:
That's really interesting. @nopara73, if I understand you correctly, you are saying that if the `requestblocks` JSON-RPC API from this PR is added to bitcoin core, then [HiddenWallet](https://github.com/nopara73/HiddenWallet/) would be able to call it to download blocks instead of depending on [NBitcoin](https://github.com/MetacoSA/NBitcoin)?

nopara73:
@ryanofsky 

> JSON-RPC API from this PR is added to bitcoin core, then HiddenWallet would be able to call it to download blocks instead of depending on NBitcoin?  

Yes.

jonasschnelli:
Rebased

src/rpc/blockchain.cpp
```diff
@@ -1587,6 +1592,60 @@ UniValue savemempool(const JSONRPCRequest& request)
     return NullUniValue;
 }
 
+UniValue requestblocks(const JSONRPCRequest& request)
+{
+    if (request.fHelp || request.params.size() < 1 || request.params.size() > 3)
```
promag: Should be `> 2`?

src/rpc/blockchain.cpp
```diff
@@ -645,6 +649,7 @@ UniValue getblockheader(const JSONRPCRequest& request)
             "{\n"
             "  \"hash\" : \"hash\",     (string) the block hash (same as provided)\n"
             "  \"confirmations\" : n,   (numeric) The number of confirmations, or -1 if the block is not on the main chain\n"
+            "  \"validated\" : n,       (boolean) True if the block has been validated\n"
```
promag: Should update `getblockheader` test?
https://github.com/bitcoin/bitcoin/blob/84fa64598e8b7d6901c546e938fb0eeda625a8e8/test/functional/blockchain.py#L173-L198

src/net_processing.h
```diff
@@ -81,4 +83,18 @@ bool GetNodeStateStats(NodeId nodeid, CNodeStateStats &stats);
 /** Increase a node's misbehavior score. */
 void Misbehaving(NodeId nodeid, int howmuch);
 
+/**
+ * Prioritize a block for downloading
+ * Blocks requested with priority will be downloaded and processed first
+ * Downloaded blocks will not trigger ActivateBestChain
+ */
+void AddPriorityDownload(std::vector<const CBlockIndex*>& blocksToDownload);
```
promag: Arg reference should be const:
```cpp
const std::vector<const CBlockIndex*>& blocks_to_download
```

src/net_processing.cpp
```diff
@@ -3647,6 +3699,74 @@ bool PeerLogicValidation::SendMessages(CNode* pto, std::atomic<bool>& interruptM
...
+
+bool FlushPriorityDownloads() {
+    LOCK(cs_main);
+    bool ret = blocksToDownloadFirst.empty();
+    blocksToDownloadFirst.clear();
+    return !ret;
+}
+
+size_t CountPriorityDownloads() {
+    // return a copy
```
promag: Copy of?

src/net_processing.cpp
```diff
@@ -3647,6 +3699,74 @@ bool PeerLogicValidation::SendMessages(CNode* pto, std::atomic<bool>& interruptM
...
+        }
+        else if (r.pindex->nStatus & BLOCK_HAVE_DATA) {
+            CBlock loadBlock;
+            if (!ReadBlockFromDisk(loadBlock, r.pindex, Params().GetConsensus())) {
+                throw std::runtime_error(std::string(__func__) + "Can't read block from disk");
+            }
+            currentBlock = std::make_shared<const CBlock>(loadBlock);
+        }
+        else {
+            break;
```
promag: Add comment explaining this case?

src/net_processing.cpp
```diff
@@ -3647,6 +3699,74 @@ bool PeerLogicValidation::SendMessages(CNode* pto, std::atomic<bool>& interruptM
...
+void AddPriorityDownload(std::vector<const CBlockIndex*>& blocksToDownload) {
+    LOCK(cs_main);
+    for (const CBlockIndex* pindex: blocksToDownload) {
+        // we add blocks regardless of duplicates
+        blocksToDownloadFirst.push_back({pindex, false});
+    }
+}
+
+void ProcessPriorityRequests(const std::shared_ptr<CBlock> blockRef) {
+    LOCK(cs_main);
```
promag: `cs_main` is locked throughout. It is enough to lock when dequeuing right? If dequeuing fails then return.

src/net_processing.cpp
```diff
@@ -3647,6 +3699,74 @@ bool PeerLogicValidation::SendMessages(CNode* pto, std::atomic<bool>& interruptM
...
+            }
+            currentBlock = std::make_shared<const CBlock>(loadBlock);
+        }
+        else {
+            break;
+        }
+
+        // allow processing through signal
+        GetMainSignals().ProcessPriorityRequest(currentBlock, r.pindex);
+        LogPrint(BCLog::NET, "process priority block request (%s) height=%d\n", r.pindex->GetBlockHash().ToString(), r.pindex->nHeight);
```
promag: Move log before signal?

src/net_processing.cpp
```diff
@@ -3647,6 +3699,74 @@ bool PeerLogicValidation::SendMessages(CNode* pto, std::atomic<bool>& interruptM
...
+
+        // remove processed block from queue
+        it = blocksToDownloadFirst.erase(std::remove_if(blocksToDownloadFirst.begin(), blocksToDownloadFirst.end(), [&r](const PriorityBlockRequest &rB) {
```
promag: Is removing duplicates here really necessary? Blocks with `BLOCK_HAVE_DATA` will be skipped.

src/rpc/blockchain.cpp
```diff
@@ -1587,6 +1592,60 @@ UniValue savemempool(const JSONRPCRequest& request)
...
+            "\nPriorize blocks downloads.\n"
+            "\nArguments:\n"
+            "1. action            (string, required) the action to execute\n"
+            "                                        add  = add new blocks to the priority download\n"
+            "                                        flush = flush the queue (blocks in-flight will still be downloaded)\n"
+            "                                        status = get info about the queue\n"
+            "2. blockhashes       (array, optional) the hashes of the blocks to download\n"
+            "\nResult:\n"
+            "   add: <null>\n"
+            "   flush: <true|false> (if the the queue wasn't empty)\n"
```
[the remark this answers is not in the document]
promag: What is the purpose of the return value?

src/net_processing.h
```diff
@@ -81,4 +83,18 @@ bool GetNodeStateStats(NodeId nodeid, CNodeStateStats &stats);
...
+void ProcessPriorityRequests(const std::shared_ptr<CBlock> block);
+bool FlushPriorityDownloads();
+size_t CountPriorityDownloads();
+void ProcessPriorityRequests(const std::shared_ptr<CBlock> block);
```
promag: Remove, duplicate declaration.

src/net_processing.cpp
```diff
@@ -473,9 +498,26 @@ void FindNextBlocksToDownload(NodeId nodeid, unsigned int count, std::vector<con
     // Make sure pindexBestKnownBlock is up to date, we'll need it.
     ProcessBlockAvailability(nodeid);
 
+    if (!blocksToDownloadFirst.empty()) {
+        for (const PriorityBlockRequest &r: blocksToDownloadFirst) {
+            if (r.downloaded) continue;
+            if (r.pindex && state->pindexBestKnownBlock != NULL && state->pindexBestKnownBlock->nHeight >= r.pindex->nHeight && !mapBlocksInFlight.count(r.pindex->GetBlockHash())) {
```
promag: Split in multiple tests. Suggestion for loop body:
```cpp
if (!r.pindex) continue;
if (state->pindexBestKnownBlock == nullptr) continue;
if (state->pindexBestKnownBlock->nHeight < r.pindex->nHeight) continue;
if (mapBlocksInFlight.count(r.pindex->GetBlockHash())) continue;

vBlocks.push_back(r.pindex);

if (vBlocks.size() == count) break;
```
IMHO more readable and more easier to add comments.

promag: Otherwise, replace `NULL` with nullptr and use `break` instead of `return true`.

promag (commented):
LGTM, will play a bit, meanwhile some comments below.

Test that flush empties the queue:
 - disconnect node
 - requestblocks add ["valid hash"]
 - requestblocks flush
 - requestblocks status
 - assert count equal 0

Test that no block is enqueued if one is unknown:
 - requestblocks add ["valid hash", "invalid hash"]
 - assert the above error
 - requestblocks status
 - assert count equal 0

src/net_processing.cpp
```diff
@@ -136,6 +139,13 @@ namespace {
     MapRelay mapRelay;
     /** Expiration-time ordered list of (expire time, relay map entry) pairs, protected by cs_main). */
     std::deque<std::pair<int64_t, MapRelay::iterator>> vRelayExpiration;
+
+    struct PriorityBlockRequest {
```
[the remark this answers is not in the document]
promag: Same for other structs.

src/rpc/blockchain.cpp
```diff
@@ -1587,6 +1592,60 @@ UniValue savemempool(const JSONRPCRequest& request)
...
+            "   status: {\"count\": \"<amount of blocks in the queue>\"}\n"
+            "\nExamples:\n"
+            + HelpExampleCli("requestblocks", "add, \"'[\"<blockhash>\"]'\"")
+            + HelpExampleRpc("requestblocks", "add, \"'[\"<blockhash>\"]'\"")
+            );
+
+    if (request.params[0].get_str() == "flush") {
+        return UniValue(FlushPriorityDownloads());
```
promag: Assert `request.params[1]` is null. Same for status.

nopara73:
# Simple RPC App Example

Just to be sure I understand correctly this PR and the application I am considering will work as I expect to work:

How would the wallet behave? Best effort or disabled?  

So let's say is it possible to write a software, that uses Bitcoin Core's wallet with autorequestblocks=0 in the following way:  

1. Sync up the header.  
2. Create a wallet, save the creation time (or block height).  
3. Subscribe for new header notifications.  
4. Every time a new header comes call `requestblock add blockhash`

So the wallet, from the information it has available executes the rpc commands properly based on best effort, right? And is not supposed to fail.  

This gives me the possibility to implement leapfrogging for example. (Enable the user to say: hey, I didn't do or wait for any transaction in the last 3 months and I have no intention to sync it up, it's ok if I get the blocks from now on.)  

# Other Concerns

1. Since this mode cannot have up to date utxo set, how do you broadcast transactions? How do you make sure it's valid?  
2. How do you accept transactions to your mempool without making sure it spends an utxo? Are you also propagating them?  
3. What would getblockcount return?

src/net_processing.cpp
jonasschnelli: I think it's okay after, but should say "processed" instead of "process".

src/net_processing.cpp
jonasschnelli: With the current implementation, `blocksToDownloadFirst` can have multiple of the same block. I'd say it should remain here...

src/rpc/blockchain.cpp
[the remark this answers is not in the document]
jonasschnelli: You can determine if you have flushed something or not.

src/rpc/blockchain.cpp
```diff
@@ -1587,6 +1592,60 @@ UniValue savemempool(const JSONRPCRequest& request)
...
+    }
+    else if (request.params[0].get_str() == "status") {
```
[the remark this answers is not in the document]
jonasschnelli: Why?

src/rpc/blockchain.cpp
jonasschnelli: done

nopara73:
@jonasschnelli I'd like to make the last couple of things clear before I compile and test it, just to know what to expect.  

# 1

> there is a commit available in #9483 7ca1a87 that does allow broadcasting transactions without an utxo set.

Meaning, `sendrawtransaction` will always return null?

# 2

> This PR does not touch the wallet (that will come later, would be to large of a PR).  

Meaning, (1) disables the wallet, (2) makes the wallet rpcs return sometimes incorrect results, (3) you don't know what the rpcs would return?

jonasschnelli:
> Meaning, sendrawtransaction will always return null?

No. It will return the txid (as expected). But not in this PR (that would be with something like https://github.com/bitcoin/bitcoin/commit/7ca1a8738a878078c0d1546f7743a05ee474dd1b).

> Meaning, (1) disables the wallet, (2) makes the wallet rpcs return sometimes incorrect results, (3) you don't know what the rpcs would return?

The wallet runs the same way as in master.
All wallet commands do work exactly the same way as the do in standard IBD mode.
Blocks processed with the new `requestblocks` will be ignored by the wallet (otherwise this PR would be too large).

The scope of this PR is to add the necessary network processing infrastructure to later add an SPV mode to the wallet.

ryanofsky:
>> It would be useful to get concept ACKs from @gmaxwell and @sipa to make sure concerns from the earlier design discussion (https://botbot.me/freenode/bitcoin-core-dev/msg/88437543/) are addressed now.

> Agree about required conceptual ACKs from other devs.

Last I looked code was in good shape, but I don't think there was ever more design feedback on this. Maybe it should be brought up at an IRC meeting?

nopara73:
This will be especially useful when combined with BIP158 (Compact Block Filters for Light Clients.) 
https://github.com/bitcoin/bips/blob/master/bip-0158.mediawiki

I imagine, based on this PR someone will eventually make a Bitcoin Core light client after Core starts to serve filters. At least that'll be a low hanging fruit.

Sjors:
Though I'm not sure what it would do on its own, perhaps the `PriorityBlockRequest` could be a separate PR?

Same for `requestblocks` RPC; could that be useful in pruned nodes to retrieve a pruned block?

dmoney0546:
This is dmarcus Eason and we need to talk about this because Daniel have it
right and I need it fast

jonasschnelli:
Rebased

Sjors:
I rebased it here: https://github.com/Sjors/bitcoin/commits/2018/08/spv-rpc

@ryanofsky said:
> I still can't figure out if new RPC is actually supposed to be useful for something other than debugging/testing

The bigger picture for me is #9483, the ability for users to get started while IBD happens in the background.

However, a use case I have in mind for `requestblocks add` is https://github.com/ElementsProject/lightning/issues/1297, in particular the ability for a lightning client to re-download historic blocks from a pruned node. This actually works now.

`-autorequestblocks` is useful when running on a phone (iOs: #11720, Android: [ABCore](https://github.com/greenaddress/abcore)). A native application could use the RPC and only allow fetching blocks when on wifi. Being able to toggle automatic download via RPC would be nice, but can wait.

Being able to fetch blocks by height range (or up to a certain height) instead of hashes would be nice, but that's something that can be built on top of this.

I noticed there's no log entry when you `requestblocks add`.

Can you explain when `priorityRequest` is `true`?

Did some light testing and couldn't find problems, but more people should try :-)

test/functional/rpc_requestblocks.py
```diff
@@ -0,0 +1,62 @@
...
+            timeout-=1
+        assert(timeout>0)
+
+        # block must now be available
+        block = self.nodes[1].getblock(node0bbhash, True)
+        assert_equal(block['hash'], node0bbhash)
+        assert_equal(block['validated'], False)
+
+        #prevblock must not be available
+        assert_raises_rpc_error(-1, "Block not found on disk", self.nodes[1].getblock, block['previousblockhash'])
```
Sjors: Maybe do a prune here and then refetch the old block?

src/rpc/blockchain.cpp
```diff
@@ -1626,6 +1630,60 @@ UniValue savemempool(const JSONRPCRequest& request)
...
+                }
+                blocksToDownload.push_back(mi->second);
+            }
+        }
+
+        AddPriorityDownload(blocksToDownload);
+        return NullUniValue;
+    }
+    else {
+        throw JSONRPCError(RPC_INVALID_PARAMETER, "Unkown action");
```
practicalswift: Typo found by `codespell`: Unkown

Issue #10616, opened 2017-06-16
Consider building with -flto


sipa:
GCC since version 4.5 has supporting LTO building. In this mode, compiling cpp files to object files does not emit the final assembly code yet, but an intermediate representation which is used to produce the assembly at link time. This allows the compiler to perform inter-module optimizations; most notably it permits inlining across module boundaries.

Advantages:
* Smaller binaries (code from headers does not get duplicated into multiple objects anymore).
* Faster code (more optimization opportunities)
* Lower build-time memory requirements

Downsides:
* Slower building
* Requires modern compilers
* Debugging sometimes gets harder
* Before GCC 4.8: double size object files

I believe we should consider producing release binaries created using LTO, if the compiler is sufficiently modern. It categorically removes the concern about whether code should go in header files or not (before, moving implementations to .h files would sometimes be needed for performance reasons - with LTO that is not a good reason anymore).

Opinions?

luke-jr:
I think using dynamic libraries would be better *unless* there's a significant performance improvement from LTO...

gmaxwell:
@luke-jr this more or less completely orthogonal from dynamic libraries.

luke-jr:
I was under the impression LTO was fundamentally incompatible with dynamic libraries, due to its optimisation methods?

sipa:
LTO code can statically or dynamically link non-LTO code.

laanwj:
Seems like a good idea to me.

Static/dynamc linking is orthogonal, although one option would be to build the depends with LTO as well, then with static linking the entire executable can be optimized by the linker, also calls crossing library boundaries. But I"m not sure this is worth the extra build time - the important dependencies from a performance point of view are secp256k1 and leveldb and those are already internal.

gmaxwell:
@theuni Is there anything special in the build scripts that would get in the way of this?

theuni:
@gmaxwell It depends on what the definition of LTO is here. As @laanwj said, taking this as far as possible, the dependencies can also be built with LTO, which ends in one final massive link. There were a few issues with that last I tried, most having to do with ODR violations (bdb and qt, iirc).

For just LTO'ing our own builds, I think it should just work if the compiler is capable.

We could start by adding an "--enable-lto" configure option and letting people take it for a spin.

TheBlueMatt:
[...]

PR #10800, opened 2017-07-12
build: add lto configure option

theuni:
Addresses #10616. Please see the commit messages and note that AR/RANLIB may need to be set manually!

This is testing the waters for LTO. There are a few unknowns:
- A boost LTO issue was addressed here, but others may remain for different libs/versions
- ar/ranlib detection will likely turn out to be annoying
- Not yet tested with Gitian

I propose adding this as-is (maybe after Gitian testing), and seeking feedback. I suspect that there will be quite a few platforms/toolchains where it's broken, but there's only so much we can do about it.


jonasschnelli:
Built successfully on my gitian machine (haven't tested the build yes though): https://bitcoin.jonasschnelli.ch/build/213

laanwj:
Bleh, the boost work-around is really nasty. It breaks my use-case with cloudabi, where fs::path is not actually convertible from/to a string losslessly.

MarcoFalke:
@theuni This can be closed given https://github.com/bitcoin/bitcoin/issues/10616#issuecomment-372027171 ? Otherwise needs rebase

laanwj:
Closing, let me know when it needs to be reopened.

PR #10804, opened 2017-07-12
Add histunspent RPC

promag:
This PR introduces `histunspent` RPC. It calculates an histogram for the current unspent transaction output amounts.

This allows to have a better perspective of how the total balance is distributed. This also avoids transmitting the whole UTXO to create the histogram on client side, specially when it has thousands of unspents.

For the moment, the histogram bins are defined with an array of amounts. Later there can be other constructors to define the bins.

Example:
```
./bitcoin-cli -regtest  listunspent | grep amount
    "amount": 50.00000000,
    "amount": 50.00000000,
    "amount": 4.00000000,
    "amount": 45.99996160,
    "amount": 50.00000000,
    "amount": 50.00000000,
    "amount": 36.99996160,
    "amount": 13.00000000,
    "amount": 50.00000000,

./bitcoin-cli -regtest  histunspent '[0,1,5,10,15,100]'
{
  "bins": [
    0, 
    1, 
    0, 
    1, 
    7
  ],
  "ranges": [
    0.00000000, 
    1.00000000, 
    5.00000000, 
    10.00000000, 
    15.00000000, 
    100.00000000
  ]
}
```

Note for reviewers, WIP and missing tests.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -2664,6 +2664,100 @@ UniValue listunspent(const JSONRPCRequest& request)
     return results;
 }
 
+UniValue histunspent(const JSONRPCRequest& request)
+{
+    CWallet * const pwallet = GetWalletForJSONRPCRequest(request);
+    if (!EnsureWalletIsAvailable(pwallet, request.fHelp)) {
+        return NullUniValue;
+    }
+
+    if (request.fHelp || request.params.size() > 5)
+        throw std::runtime_error(
+            "histunspent [ranges] ( min_depth max_depth )\n"
+            "\nReturns an histogram of unspent transaction output amounts\n"
+            "with between min_depth and max_depth (inclusive) confirmations.\n"
+            "\nArguments:\n"
+            "1. ranges             (array) A json array of amounts that define the histogram bins\n"
+            "2. min_depth          (numeric, optional, default=1) The minimum confirmations to filter\n"
+            "3. max_depth          (numeric, optional, default=9999999) The maximum confirmations to filter\n"
+            "\nResult\n"
+            "{\n"
+            "  \"bins\" : \"[]\",           (array) A json array of counts of amounts in each bin\n"
+            "  \"ranges\" : \"[]\",         (array) A json array of amounts that define the histogram bins\n"
+            "}\n"
+
+            "\nExamples\n"
+            + HelpExampleCli("histunspent", "\"[0,1,10,100,1000,10000]\"")
+        );
+
+    std::vector<unsigned int> bins;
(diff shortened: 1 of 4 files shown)


jnewbery:
My immediate reaction to this is the same as https://github.com/bitcoin/bitcoin/pull/10757#issuecomment-313714215 and https://github.com/bitcoin/bitcoin/pull/9733#issuecomment-291981292. I'm not a fan of these statistical RPCs, which seem to me to be very interesting, but really only provide niche functionality. My issue is that they add maintenance burden for features that are only likely to be used by very few people. For me, they're better served by a fork of the project or a patch set. To my mind, this project should try to minimize functionality to what's essential for end users.

I understand that I may be in the minority in that view, and that it's probably discouraging for contributors for proposed PRs to be rejected, so I'm not going to push the point.

promag:
I generally agree with you, this is easily done out but the problem is when the UTXO set is very large:
 - listunspent can take minutes to complete;
 - the result is a huge array;
 - no pagination support
 - locks held too much time.

The goal with this RPC, is to create a flexible UTXO counting to solve the above problems in one-shot.

Typically this is useful for online wallets to monitor the UTXO set, to trigger sweep transactions, to understand if there are good unspents for the upcoming transactions to have the least input count (hence low fee), etc..

promag:
>  For me, they're better served by a fork of the project or a patch set. To my mind, this project should try to minimize functionality to what's essential for end users.

@jnewbery IMO there is space for a basic statistical feature like this in core, which is by far one of the easiest around in code.

laanwj:
I don't share @jnewbery's general resent against statistical calls. Sometimes the information to compute the statistics is available only server-side.

In this case, however, all the information is available in the `listunspent` result. So this can just as well be computed in an external script. Building it into bitcoind is an unnecessary burden.

promag:
> all the information is available in the listunspent

@laanwj sure but it's not that feasible when it's a huge data set, even with `listunspent` query options.

promag:
Closing since the general opinion is to compute client side. I guess the above bottlenecks should be tackled instead.

src/wallet/rpcwallet.cpp
```diff
@@ -2664,6 +2664,100 @@ UniValue listunspent(const JSONRPCRequest& request)
...
+    std::vector<unsigned int> bins;
+    std::map<CAmount, unsigned int> ranges;
+    CAmount min_amount = 0;
+    CAmount max_amount = MAX_MONEY;
+
+    if (!request.params[0].isNull()) {
+        const std::vector<UniValue>& values = request.params[0].getValues();
+
+        if (values.size() < 2) {
+            // throw invalid range (minimum 2 amounts);
```
pedrobranco: 👀

Issue #10670, opened 2017-06-26
Master doesn't build on OpenBSD (leveldb)


laanwj:
It looks like OpenBSD's `eg++` which is based on 4.9.4 (or rather, the binutils based on 2.17) doesn't support the CRC32 instructions introduced with the recent leveldb upgrade:
```
Making all in src
gmake[1]: Entering directory '/home/user/bitcoin/src'
gmake[2]: Entering directory '/home/user/bitcoin/src'
gmake[3]: Entering directory '/home/user/bitcoin'
gmake[3]: Leaving directory '/home/user/bitcoin'
eg++ -std=c++11 -DHAVE_CONFIG_H -I. -I../src/config  -U_FORTIFY_SOURCE -D_FORTIFY_SOURCE=2 -I./leveldb -DOS_OPENBSD -DLEVELDB_ATOMIC_PRESENT -D__STDC_LIMIT_MACROS  -DLEVELDB_PLATFORM_POSIX -I./leveldb/include -I./leveldb/helpers/memenv -DLEVELDB_PLATFORM_POSIX_SSE -DHAVE_BUILD_INFO -D__STDC_FORMAT_MACROS -Wstack-protector -fstack-protector-all  -fPIE -msse4.2 -g -O2 -Wall -Wextra -Wformat -Wvla -Wformat-security -Wno-unused-parameter -MT leveldb/port/leveldb_libleveldb_sse42_a-port_posix_sse.o -MD -MP -MF leveldb/port/.deps/leveldb_libleveldb_sse42_a-port_posix_sse.Tpo -c -o leveldb/port/leveldb_libleveldb_sse42_a-port_posix_sse.o `test -f 'leveldb/port/port_posix_sse.cc' || echo './'`leveldb/port/port_posix_sse.cc
leveldb/port/port_posix_sse.cc: In function 'uint32_t leveldb::port::AcceleratedCRC32C(uint32_t, const char*, size_t)':
leveldb/port/port_posix_sse.cc:59:15: warning: 'ecx' may be used uninitialized in this function [-Wmaybe-uninitialized]
   return (ecx & (1 << 20)) != 0;
               ^
leveldb/port/port_posix_sse.cc:57:26: note: 'ecx' was declared here
   unsigned int eax, ebx, ecx, edx;
                          ^
/tmp//ccBAimlE.s: Assembler messages:
/tmp//ccBAimlE.s:95: Error: no such instruction: `crc32b -1(%rbp),%eax'
[...]

PR #10806, opened 2017-07-12
build: verify that the assembler can handle crc32 functions

theuni:
Also, enable crc32 even if -msse4.2 wasn't added by us, as long as it works. This allows custom flags (such as -march=native) to work as expected.

Addresses #10670.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -246,9 +246,32 @@ if test "x$CXXFLAGS_overridden" = "xno"; then
   # Check for optional instruction set support. Enabling these does _not_ imply that all code will
   # be compiled with them, rather that specific objects/libs may use them after checking for runtime
   # compatibility.
-  AX_CHECK_COMPILE_FLAG([-msse4.2],[[enable_sse42=yes; SSE42_CXXFLAGS="-msse4.2"]],,[[$CXXFLAG_WERROR]])
+  AX_CHECK_COMPILE_FLAG([-msse4.2],[[SSE42_CXXFLAGS="-msse4.2"]],,[[$CXXFLAG_WERROR]])
 
 fi
+
+TEMP_CXXFLAGS="$CXXFLAGS"
+CXXFLAGS="$CXXFLAGS $SSE42_CXXFLAGS"
+AC_MSG_CHECKING(for assembler crc32 support)
+AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
+    #include <stdint.h>
+    #if defined(_MSC_VER)
+    #include <intrin.h>
+    #elif defined(__GNUC__) && defined(__SSE4_2__)
+    #include <nmmintrin.h>
+    #endif
+  ]],[[
+    uint64_t l = 0;
+    l = _mm_crc32_u8(l, 0);
+    l = _mm_crc32_u32(l, 0);
+    l = _mm_crc32_u64(l, 0);
+    return l;
+  ]])],
+ [ AC_MSG_RESULT(yes); enable_hwcrc32=yes],
+ [ AC_MSG_RESULT(no)]
+)
+CXXFLAGS="$TEMP_CXXFLAGS"
+
 CPPFLAGS="$CPPFLAGS -DHAVE_BUILD_INFO -D__STDC_FORMAT_MACROS"
 
 AC_ARG_WITH([utils],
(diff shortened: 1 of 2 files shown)


gmaxwell (approved):
utACK

sipa:
@theuni I'm confused why the old code wouldn't build the sse library with the correct flags?

theuni:
@sipa This addresses 2 things:

1. Disables crc32 on platforms where the assembler can't handle it
2. Enables crc32 if cflags/cxxflags are overridden but the intrinsics still work.

I think 2 is what you're asking about? If so, it wouldn't work before because, even though the user has specified -msse4.2 (or -march=native, or whatever it takes to get the intrinsics working), our build wouldn't have the necessary make option set here: https://github.com/bitcoin/bitcoin/blob/master/src/Makefile.leveldb.include#L145, so the necessary define wouldn't be added.

laanwj:
Going to test.

laanwj:
Interesting. On OpenBSD 6.1:
```
configure:18285: checking whether C++ compiler accepts -msse4.2  
configure:18304: eg++ -std=c++11 -c -g -O2 -Wall -Wextra -Wformat -Wvla -Wformat-security -Wno-unused-parameter -Werror -msse4.2  conftest.cpp >&5
configure:18304: $? = 0
configure:18313: result: yes
configure:18326: checking for assembler crc32 support  
configure:18351: eg++ -std=c++11 -c -g -O2 -Wall -Wextra -Wformat -Wvla -Wformat-security -Wno-unused-parameter -msse4.2  conftest.cpp >&5
conftest.cpp: In function 'int main()':
conftest.cpp:34:14: warning: variable 'l' set but not used [-Wunused-but-set-variable]
     uint64_t l = 0;
              ^
configure:18351: $? = 0
configure:18352: result: yes
configure:20040: checking for pk
```
Then when running gmake:
```
leveldb/port/port_posix_sse.cc: In function 'uint32_t leveldb::port::AcceleratedCRC32C(uint32_t, const char*, size_t)':
leveldb/port/port_posix_sse.cc:59:15: warning: 'ecx' may be used uninitialized in this function [-Wmaybe-uninitialized]
   return (ecx & (1 << 20)) != 0;
               ^
leveldb/port/port_posix_sse.cc:57:26: note: 'ecx' was declared here
   unsigned int eax, ebx, ecx, edx;
                          ^
/tmp//cc5NhjN4.s: Assembler messages:
/tmp//cc5NhjN4.s:95: Error: no such instruction: `crc32b -1(%rbp),%eax'
/tmp//cc5NhjN4.s:121: Error: no such instruction: `crc32q -8(%rbp),%rax'
/tmp//cc5NhjN4.s:148: Error: no such instruction: `crc32b -1(%rbp),%eax'
:312: Error: no such instruction: `crc32l 0(%rbp),%eax'
gmake[2]: *** [Makefile:4923: leveldb/port/leveldb_libleveldb_sse42_a-port_posix_sse.o] Error 1
```
So somehow it works while running configure, but not while building *confused*.

theuni:
@laanwj Grr, I checked an asm dump of the test to make sure that the crc32s are actually emitted, but it looks like I forgot to test with -O2. Indeed with optims on, they're optimized out. I'll add something to use the result.

laanwj:
@theuni this solves it:
```diff
diff --git a/configure.ac b/configure.ac
index 5c3dac648..6b5d891fa 100644
--- a/configure.ac
+++ b/configure.ac
@@ -262,9 +262,10 @@ AC_COMPILE_IFELSE([AC_LANG_PROGRAM([[
     #endif
   ]],[[
     uint64_t l = 0;
-    l = _mm_crc32_u8(0, 0);
-    l = _mm_crc32_u32(0, 0);
-    l = _mm_crc32_u64(0, 0);
+    l = _mm_crc32_u8(l, 0);
+    l = _mm_crc32_u32(l, 0);
+    l = _mm_crc32_u64(l, 0);
+    return l;
   ]])],
  [ AC_MSG_RESULT(yes); enable_hwcrc32=yes],
  [ AC_MSG_RESULT(no)]
```

theuni:
@laanwj Aha. I was messing with printf to force it to stay around, I didn't realize that the test function was "int main()". That makes total sense and is really good to know!

Pushed with your fix.

laanwj:
Tested ACK d34d77a

PR #10810, opened 2017-07-12
missing white space in function arg

greenaddress:


jnewbery:
utACK 7bf33a8fb06fbfd15704021528990d1ac70fee15

promag:
NACK, I believe linting fixes should be avoided. Regarding this PR, there are lots of other cases.

jonasschnelli:
Agree with @promag, don't do such changes, the slow down everything while not really improving things.
utACK 69a4339ef509da633e83fa9968ac646c58b95390

greenaddress:
@promag @jonasschnelli thanks for feedback, I was reviewing some code in the area and noticed the missing white space and thought about a mini PR - will avoid in the future.

PR #10814, opened 2017-07-13
Change type of op to agree with type of MAX_OPCODE.

ReneNyffenegger:
MAX_OPCODE is defined 'unsigned int' in script/script.h.
Changing the variable op's type from int to unsigned int eliminates
a compiler warning.


fanquake:
This is part of #10808

PR #10822, opened 2017-07-14
RPC: Also serve txo from gettxout (not just utxo and mempool)

jtimon:
Expand gettxout to also serve spent txo if pruning or lack of -txindex don't get in the way.

If the output is spent, it will be slower to load, but there's no reason not to serve this if the data is there.

TODO (functional tests):

- [x] Test success (ideally more than one example)
- [x] Test error no txid
- [x] Test error no n for this txid
- [x] Test spent/unspent
- [x] Test txindex error
- [ ] Test prunning error

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -46,6 +46,33 @@ static CUpdatedBlock latestblock;
 
 extern void TxToJSON(const CTransaction& tx, const uint256 hashBlock, UniValue& entry);
 
+static bool ReadTxoFromOutpoint(Coin& coin, bool& is_spent, const CBlockIndex* pindex, const COutPoint& out, bool include_spent)
+{
+    is_spent = false;
+    if (pcoinsTip->GetCoin(out, coin)) {
+        return true;
+    } else if (!include_spent) {
+        return false;
+    }
+    is_spent = true;
+
+    uint256 hashblock;
+    CTransactionRef tx_ref;
+    if (!GetBlockchainTx(out.hash, tx_ref, Params().GetConsensus(), hashblock, true)) {
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string(fTxIndex ? "No such blockchain transaction"
+            : "No such utxo blockchain transaction. Use -txindex to enable spent output queries"));
+    }
+    if (out.n >= tx_ref->vout.size()) {
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string(strprintf("No output %d for tx %s", out.n, out.hash.GetHex())));
+    }
+    const CTxOut& prevoutput = tx_ref->vout[out.n];
+    coin.out = prevoutput;
+    coin.fCoinBase = tx_ref->IsCoinBase();
+    coin.nHeight = mapBlockIndex[hashblock]->nHeight;
+
+    return true;
+}
+
 double GetDifficulty(const CBlockIndex* blockindex)
 {
     if (blockindex == NULL)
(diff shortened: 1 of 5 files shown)


sipa:
I don't think we should be expanding the impact of `-txindex`. This functionality is already available through `getrawtransaction` anyway.

jtimon:
> I don't think we should be expanding the impact of -txindex.

Is there a problem with that? This is just RPC anyway.
I also think making -txindex more useful may increase the incentives to run an archive full node.

My reasoning (with some errors corrected below it) is here: https://github.com/bitcoin/bitcoin/pull/9806#issuecomment-313578569

> This functionality is already available through getrawtransaction anyway.

By that logic, shouldn't we just fully remove gettxout?
That would be bad for gettxoutbyscript and gettxoutbyaddress though, I think.

gettxout is faster for utxo, so if one wants a txo (not knowing if stxo/utxo a priori), one could call gettxout first and only if it fails call getrawtransation. But I think a single call would be more convenient.
If not, I guess we need getoutpointfromscript and getoupointfrom address call too to serve the full txo.
One could, for example, call gettxoutbyaddress and if it fails, also call gettxoutbyaddress and getrawtransaction. It would be more cumbersome, but it would also work.

src/validation.cpp
```diff
@@ -955,6 +948,23 @@ bool GetTransaction(const uint256 &hash, CTransactionRef &txOut, const Consensus
...
+/**
+ * Return transaction in txOut, and if it was found inside a block,
+ * its hash is placed in hashBlock. Look in the mempool first.
+*/
+bool GetTransaction(const uint256 &hash, CTransactionRef &txOut, const Consensus::Params& consensusParams, uint256 &hashBlock, bool fAllowSlow)
+{
+    LOCK(cs_main);
+
+    CTransactionRef ptx = mempool.get(hash);
+    if (ptx)
```
promag:
```cpp
if (ptx) {
```

test/functional/wallet.py
```diff
@@ -19,6 +19,7 @@ def __init__(self):
         self.setup_clean_chain = True
         self.num_nodes = 4
         self.extra_args = [['-usehd={:d}'.format(i%2==0)] for i in range(4)]
+        self.extra_args[0].append('-txindex')
```
promag: @jnewbery this is why I prefer isolated tests. This is unrelated to all these tests except the new.

src/rpc/blockchain.cpp
```diff
@@ -987,6 +1016,12 @@ UniValue gettxout(const JSONRPCRequest& request)
     if (request.params.size() > 2)
         fMempool = request.params[2].get_bool();
 
+    bool include_spent = false;
+    if (request.params.size() > 3) {
+        include_spent = request.params[3].get_bool();
+    }
```
promag: Disallow `include_spent` if `include_mempool`:
```cpp
    if (fMempool && include_spent) {
        throw JSONRPCError(RPC_INVALID_PARAMETER, "...");
    }
```
If you add this, update test below.

src/validation.cpp
```diff
@@ -955,6 +948,23 @@ bool GetTransaction(const uint256 &hash, CTransactionRef &txOut, const Consensus
     return false;
 }
 
+/**
+ * Return transaction in txOut, and if it was found inside a block,
+ * its hash is placed in hashBlock. Look in the mempool first.
+*/
+bool GetTransaction(const uint256 &hash, CTransactionRef &txOut, const Consensus::Params& consensusParams, uint256 &hashBlock, bool fAllowSlow)
```
promag: snake_case.

jtimon:
Needs rebase and has nits but since I didn't heard any enthusiasm for more functionality regarding -txindex and @sipa was negative about it, I am closing this for now.
@promag thanks for the review and your nits will be taken into account if I ever reopen this.

PR #10827, opened 2017-07-14
fixed grammar error

Thecave3:
Just found a typing error in the italian translation, i've fixed it


MeshCollider:
> Important: We do not accept translation changes as GitHub pull requests because the next pull from Transifex would automatically overwrite them again.

Please take a look at the README and submit the change to Trabsifex instead

Thecave3:
Ok, Thank you and sorry for that then ;)

PR #10831, opened 2017-07-15
Batch flushing operations to the walletdb during top up and increase keypool size.

gmaxwell:
This carries the walletdb object from top-up through GenerateNewKey/DeriveNewChildKey/CWallet::AddKeyPubKey, which allows us to avoid the flush on destruction until the top up finishes instead of flushing the wallet for every key.

This speeds up adding keys by well over 10x on my laptop (actually something like 17x), I wouldn't be surprised if it were an even bigger speedup on spinning rust.

Then it increases the keypool size to 1000. I would have preferred to use 10,000 but in the case where the user creates a new wallet and then turns on encryption it seems kind of dumb to have >400KB of marked-used born unencrypted keys just laying around.

(Thanks to Matt for cluesticking me on how to bypass the crypter spaghetti)

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -3213,7 +3213,9 @@ bool CWallet::TopUpKeyPool(unsigned int kpSize)
             } else {
                 setExternalKeyPool.insert(nEnd);
             }
-            LogPrintf("keypool added key %d, size=%u (%u internal), new key is %s\n", nEnd, setInternalKeyPool.size() + setExternalKeyPool.size(), setInternalKeyPool.size(), internal ? "internal" : "external");
+        }
+        if (missingInternal + missingExternal > 0) {
+            LogPrintf("keypool added %d keys (%d internal), size=%u (%u internal)\n", missingInternal + missingExternal, missingInternal, setInternalKeyPool.size() + setExternalKeyPool.size(), setInternalKeyPool.size());
         }
     }
     return true;
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -106,8 +106,9 @@ CPubKey CWallet::GenerateNewKey(CWalletDB &walletdb, bool internal)
     }
 
     // Compressed public keys were introduced in version 0.6.0
-    if (fCompressed)
+    if (fCompressed) {
         SetMinVersion(FEATURE_COMPRPUBKEY);
+    }
 
     CPubKey pubkey = secret.GetPubKey();
     assert(secret.VerifyPubKey(pubkey));
@@ -115,8 +116,9 @@ CPubKey CWallet::GenerateNewKey(CWalletDB &walletdb, bool internal)
     mapKeyMetadata[pubkey.GetID()] = metadata;
     UpdateTimeFirstKey(nCreationTime);
 
-    if (!AddKeyPubKey(secret, pubkey))
+    if (!AddKeyPubKeyWithDB(walletdb, secret, pubkey)) {
         throw std::runtime_error(std::string(__func__) + ": AddKey failed");
+    }
     return pubkey;
 }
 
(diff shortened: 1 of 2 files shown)


gmaxwell:
@pstratem  you might have some interest in reviewing this.

sipa:
utACK ff9a291b30d7464c1ff0990b1cab36d6119b2ae5

instagibbs:
tACK https://github.com/bitcoin/bitcoin/pull/10831/commits/ff9a291b30d7464c1ff0990b1cab36d6119b2ae5

with 1000 keypool size, topup went from 20 seconds to 1.6.

TheBlueMatt:
utACK ff9a291b30d7464c1ff0990b1cab36d6119b2ae5 (needs rebase, though).

gmaxwell:
@TheBlueMatt  rebased, please re-review

src/wallet/wallet.cpp
```diff
@@ -162,33 +164,49 @@ void CWallet::DeriveNewChildKey(CKeyMetadata& metadata, CKey& secret, bool inter
...
+        return walletdb.WriteKey(pubkey,
                                                  secret.GetPrivKey(),
                                                  mapKeyMetadata[pubkey.GetID()]);
     }
     return true;
 }
 
+bool CWallet::AddKeyPubKey(const CKey& secret, const CPubKey &pubkey)
+{
+    CWalletDB walletdb(*dbw);
```
TheBlueMatt: It seems strange that we're creating a db transaction, only to throw it out in a second. I think its safe, but it might be nicer to let AddKeyPubKeyWithDB create a CWalletDB if it needs one automagically (eg by passing the walletdb in as a pointer).

src/wallet/wallet.cpp
```diff
@@ -162,33 +164,49 @@ void CWallet::DeriveNewChildKey(CKeyMetadata& metadata, CKey& secret, bool inter
...
+    }
     script = GetScriptForRawPubKey(pubkey);
-    if (HaveWatchOnly(script))
+    if (HaveWatchOnly(script)) {
         RemoveWatchOnly(script);
+    }
 
     if (!IsCrypted()) {
-        return CWalletDB(*dbw).WriteKey(pubkey,
+        return walletdb.WriteKey(pubkey,
```
promag: Another approach is to use `dbw` as the storage of the transaction. For instance, `CWalletDBWrapper` could keep a `CBD` instance, referenced counted by `CWalletDB`. This could be thread safe too.

promag: I'm happy to write such alternative if there is interest.

TheBlueMatt:
@promag sadly I think it's a bit late for that for 15, it might be a decent change later, however.

src/wallet/wallet.cpp
gmaxwell: This isn't called when we don't need one; the case where we already have a dbwallet calls the WithDB version.  (and changing the prototype of the function would mean changing the function it overrides, which a db handle doesn't even make sense for, and 29 callsites) Unless I'm misunderstanding you.

src/wallet/wallet.cpp
gmaxwell: Seems like a reasonable thing to do, but I didn't even want to consider anything that complex-- I wanted to eliminate the complaint that having many keys is 'slow' for 0.15 because of the bad interactions with hd-wallet and rescan when the keypool isn't big.

promag:
@TheBlueMatt another PR then.

utACK 3986271.

sipa:
reutACK 3986271c0accfe896c66a48d3ed884a7553b6413

src/wallet/wallet.cpp
```diff
@@ -3183,8 +3201,9 @@ bool CWallet::TopUpKeyPool(unsigned int kpSize)
                 nEnd = std::max(nEnd, *(setExternalKeyPool.rbegin()) + 1);
             }
 
-            if (!walletdb.WritePool(nEnd, CKeyPool(GenerateNewKey(internal), internal)))
+            if (!walletdb.WritePool(nEnd, CKeyPool(GenerateNewKey(walletdb, internal), internal))) {
                 throw std::runtime_error(std::string(__func__) + ": writing generated key failed");
+            }
 
             if (internal) {
                 setInternalKeyPool.insert(nEnd);
```
laanwj: If we increase the default keypool size, the `LogPrintf("keypool added key %d` message below shouldn't be logged for every key. Or at least moved to a debug class.

src/wallet/wallet.cpp
```diff
@@ -100,27 +100,29 @@ CPubKey CWallet::GenerateNewKey(bool internal)
...
     if (IsHDEnabled()) {
-        DeriveNewChildKey(metadata, secret, (CanSupportFeature(FEATURE_HD_SPLIT) ? internal : false));
+        DeriveNewChildKey(walletdb, metadata, secret, (CanSupportFeature(FEATURE_HD_SPLIT) ? internal : false));
     } else {
         secret.MakeNewKey(fCompressed);
     }
 
     // Compressed public keys were introduced in version 0.6.0
-    if (fCompressed)
+    if (fCompressed) {
```
promag: Unrelated, unless you want to fix all style issues?

instagibbs: I don't think we should nit people doing partial style issue fixes unless it affects reviewability.

instagibbs:
reACK https://github.com/bitcoin/bitcoin/pull/10831/commits/4f811059fab9a9121befae09f66b9b12244051f0

src/wallet/wallet.cpp
laanwj: Yes, this seems to be a "damned if you do, damned if you don't" thing.
- If you deliberately make style fixes to code you touch, you get comments that it's unrelated
- If you deliberately don't make style fixes to code you touch, you get questions why you didn't update the code for the style guidelines
I think both are valid, unless it makes review-ability in which case it could still be done in a separate commit in the same PR.

jonasschnelli:
Nice!
utACK 4f811059fab9a9121befae09f66b9b12244051f0.

src/wallet/wallet.cpp
```diff
@@ -162,33 +164,49 @@ void CWallet::DeriveNewChildKey(CKeyMetadata& metadata, CKey& secret, bool inter
...
+
+    bool needsDB = !pwalletdbEncryption;
+    if (needsDB) {
+        pwalletdbEncryption = &walletdb;
+    }
+    if (!CCryptoKeyStore::AddKeyPubKey(secret, pubkey)) {
+        if (needsDB) pwalletdbEncryption = NULL;
         return false;
+    }
+    if (needsDB) pwalletdbEncryption = NULL;
```
laanwj: I don't understand this code.
Given that `pwalletdbEncryption` is not used in `CCryptoKeyStore::AddKeyPubKey`, could this code be moved before `if (!CCryptoKeyStore::AddKeyPubKey`? This would avoid some awkward duplication here.
Or is it used indirectly? This looks really sneaky. Please add a comment if you need to keep it at least, so that no one else gets my bad idea.

gmaxwell: It's used-- alas,  the whole purpose of that pre-existing terrible variable is because CCryptoKeyStore::AddKeyPubKey  calls AddCryptedKey which on actual wallets is overridden to the function below.  CCryptoKeyStore rightfully has no concept of wallet databases, so the variable is used to tunnel through it, to get it to the AddCryptedKey override below.  This infrastructure was preexisting because the encryption setup needs it.    I can add a comment explaining it.  Some people have suggested larger refactors to remove this knitted maze, but it's clearly out of scope for now.

gmaxwell: Added a comment

src/wallet/wallet.cpp
gmaxwell: Aggregated to a summary line.

src/wallet/wallet.cpp
gmaxwell: I have been trying to limit them to things that will show up in the same patch hunks, and only changes which I'm pretty sure won't break review.   I'm happy to do what other people want, though I prefer to at least fix braces (because I dislike reviewing the risky looking unbraced code myself)-- but ultimately I just need clear guidance to avoid the damned if/don't.

gmaxwell:
(new revisions are just due to adding a comment in the middle of the patch stack)

src/wallet/wallet.cpp
TheBlueMatt: I was confused, sorry.

TheBlueMatt:
postumous utACK b0e8e2de8408cbaed9d70914c67b4c9f11397cb7 (lets clean up some of this stuff in 16).

PR #10844, opened 2017-07-16
Use range based for loop

ReneNyffenegger:
Instead of iterating over 0 .. 1 and then deciding on an actual desired
value, use a range based for loop for the desired value.

### DIFF
--- a/src/test/dbwrapper_tests.cpp
+++ b/src/test/dbwrapper_tests.cpp
@@ -24,8 +24,7 @@ BOOST_FIXTURE_TEST_SUITE(dbwrapper_tests, BasicTestingSetup)
 BOOST_AUTO_TEST_CASE(dbwrapper)
 {
     // Perform tests both obfuscated and non-obfuscated.
-    for (int i = 0; i < 2; i++) {
-        bool obfuscate = (bool)i;
+    for (bool obfuscate : {false, true}) {
         fs::path ph = fs::temp_directory_path() / fs::unique_path();
         CDBWrapper dbw(ph, (1 << 20), true, false, obfuscate);
         char key = 'k';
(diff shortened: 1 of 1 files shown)


dcousens (approved):

src/test/dbwrapper_tests.cpp
```diff
@@ -24,8 +24,8 @@ BOOST_FIXTURE_TEST_SUITE(dbwrapper_tests, BasicTestingSetup)
 BOOST_AUTO_TEST_CASE(dbwrapper)
 {
     // Perform tests both obfuscated and non-obfuscated.
-    for (int i = 0; i < 2; i++) {
-        bool obfuscate = (bool)i;
+    for (bool obfuscate: { false, true } )
```
practicalswift: Nit: Please use the form `for (int x: {1, 2})` instead of `for (int x: { 1, 2 } )`.

The former is the dominant form used in the repo:

``` shell
$ git grep -E "\{(true|false|0|1|2)" -- "*.cpp" | wc -l
112
$ git grep -E "\{ (true|false|0|1|2)" -- "*.cpp" | wc -l
9
```

practicalswift:
utACK ab005dd40dfc3e0e4ec126bbfc8144cd8a42e741

MarcoFalke:
Make sure to adhere to the developer notes when touching code.

https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md#developer-notes

There is even a link to a clang format script which should take care of
white space and braces for you.

ReneNyffenegger:
I was unaware of the clang format script. Thaks for the pointer. I let it change the commit, and squashed and force-pushed it.

TheBlueMatt:
utACK d0413c670b4e5dc79d5cc1bc35571fca745c9a24

sipa:
utACK d0413c670b4e5dc79d5cc1bc35571fca745c9a24

MarcoFalke:
utACK d0413c6

PR #10849, opened 2017-07-17
Multiwallet: simplest endpoint support

jonasschnelli:
Alternative for #10829 and #10650.
It adds the most simplest form of wallet based endpoint support (`/wallet/<filename>`).
No v1 and no node/wallet endpoint split.


laanwj:
Looks good to me too, utACK https://github.com/bitcoin/bitcoin/pull/10849/commits/c8115e31325312061682bf966b4475359cbc5215.

ryanofsky:
This has no documentation. If we think this approach is a good idea, I would like to see in what way it can be explained to users. Is there going to be a `doc/JSPON-RPC-interface.md` to complement https://github.com/bitcoin/bitcoin/blob/master/doc/REST-interface.md? Will endpoint be considered experimental (unchanging semantics but possibly temporary) or unstable (possibly changing semantic and possibly temporary)?

Also, from the other issue I think something like `-rpcwallet=filename` would be better than `-usewallet=filename`. It's hard to see how "use" could convey any meaning when you would expect any command line option that you specify to be used.

And, also from other issue I think '`?wallet=filename` makes more sense than `/wallet/filename` as more extensible and conventional way of passing options (as opposed to required arguments) in urls.

laanwj:
> And, also from other issue I think '?wallet=filename makes more sense than /wallet/filename as more extensible and conventional way of passing options (as opposed to required arguments) in urls.

Definitely no! Let's not repeat this: JSON-RPC uses POST only, and with POST it is weird to have `?a=b` URL-encoded arguments in the URL as well. It also implies more arguments could be added later, which is not true. This is only to select the wallet, which is inherently hierarchical.
(also e.g. ?a=b syntax isn't deterministic, it's possible to add arguments that are ignored, change the order of the arguments, etc, whereas a /wallet/<name>/ URL is simply a fixed thing)

ryanofsky:
> POST it is weird to have ?a=b URL-encoded arguments in the URL as well

Don't know your source for this but this is done all the time.

> It also implies more arguments could be added later, which is not true.

I don't know why it it wouldn't be true, but ok. I can think of lots of RPC options you might want to add  outside the RPC request: timeouts, completions handlers, etc.

> ?a=b syntax isn't deterministic, it's possible to add arguments that are ignored, change the order of the arguments, etc, whereas a /wallet// URL is simply a fixed thing

Not true, you don't have to ignore any argument, but it is good practice to accept arguments in any order.

laanwj:
> I don't know why it it wouldn't be true, but ok. I can think of lots of RPC options you might want to add outside the RPC request: timeouts, completions handlers, etc.

Those really shouldn't be passed in the URL.

ryanofsky:
> Those really shouldn't be passed in the URL. The URL locates something, such as the wallet in this case.

They shouldn't be part of the uri-path, but you would have to explain more why they shouldn't be options in a query string...

gmaxwell:
ryanofsky: Among other reasons the path approach will already work with most but not all software,  e.g. you change your RPC endpoint url to be one with the patch, and tada-- done.  I believe a query string would not.

I could see an argument that something like a timeout might make sense as a query like parameter, but a wallet seems like a pretty perfect 'navigation' like element.

laanwj:
> Is there going to be a doc/JSPON-RPC-interface.md to complement https://github.com/bitcoin/bitcoin/blob/master/doc/REST-interface.md? 

And yes, there should be. Documentation is good.
But that's not a requirement for the feature freeze. This is an experimental API, it could be changed completely for next release, so not having it documented 100% on first release is not a problem.

ryanofsky:
> you change your RPC endpoint url to be one with the patch, and tada-- done

This is true for this PR but _not_ #10650 which is a big reason why I think this PR is better and less user-hostile than #10650.

@laanwj, my bigger point is about extensibility . We want to add a wallet option right now. Great, so that can be as simple as adding `?wallet=filename` to the query string to provide the nice user benefit Greg cites. And maybe you have a problem with timeouts or completion handlers, but lets say we want to add a `?network=name` parameter in the future (maybe to add extra way to specify that an operation is done on mainnet or testnet, or in some dystopia where there are multiple chains, but no need to go there...). The point is you want to be able to add some new option. With query strings you have a standard way to do this. `?wallet=filename&network=name&timeout=30s` works just as well as `?timeout=30s&wallet=filename&network=name` as established by long-running convention.

If we use paths instead, are we requiring `/wallet/filename/network/name` or `/network/name/filename` or `/wallet/filename?network=name`, or something else completely arbitrary based on historical accident? Or because a filename is a "location" but a network name isn't?

I guess another thing that is motivating me is that I would like to see less uri-path space annexed for this one multiwallet feature. Right now json-rpc calls go directly to `/` handler and REST calls go directly to `/rest` handler. I'd be happier if more of the path space could be free for other applications like a web interface. If all this stuff was moved under an `/rpc` path that would also seem like an improvement to me.

Also, so my other points doesn't get lost above, it would be great to see something done on documentation and `-usewallet` if we are taking this endpoint approach.

gmaxwell:
> I'd be happier if more of the path space could be free for other applications like a web interface. If all this stuff was moved under an /rpc path

It's under /wallet/ at least, so I think it won't have any clashes with other use.

As far as usewallet-- I think we intend that argument to eventually specify all wallets that get loaded, not just specific to rpc.  e.g. it would also tell it which ones to load in the UI when that support is complete.

ryanofsky:
> As far as usewallet-- I think we intend that argument to eventually specify all wallets that get loaded, not just specific to rpc. e.g. it would also tell it which ones to load in the UI when that support is complete.

That's interesting. I thought the UI was would just tabs or a dropdown to choose between any of the loaded wallets. But I guess the option of having the `bitcoin-qt` executable (not `bitcoin-cli`) accept a `-usewallet=filename` instead of a more descriptive `-showwallet=filename` or `-restrictwallet=filename` (if UI will support multiwallet display) would be that you can specify `-usewallet` in your `bitcoind.conf` and have it affect both `bitcoin-cli` and `bitcoin-qt`

src/bitcoin-cli.cpp
```diff
@@ -241,7 +242,20 @@ UniValue CallRPC(const std::string& strMethod, const UniValue& params)
     assert(output_buffer);
     evbuffer_add(output_buffer, strRequest.data(), strRequest.size());
 
-    int r = evhttp_make_request(evcon.get(), req.get(), EVHTTP_REQ_POST, "/");
+    // check if we should use a special wallet endpoint
+    std::string endpoint;
```
jnewbery: bug: if no wallet is specified then the endpoint will be "" instead of "/"

result: bitcoin-cli doesn't work without a `-usewallet` parameter:

```
→ bitcoin-cli getinfo
error: no response from server
```

Fix is:

```c++
    std::string endpoint = "/";
```

jnewbery:
It's a shame that this PR doesn't prevent individual wallet endpoints from accessing server-level RPCs. However that can be added at a later date.

I have a branch here: https://github.com/jnewbery/bitcoin/tree/pr10849 that does a couple of things:

- fixes https://github.com/bitcoin/bitcoin/pull/10849#discussion_r127724767 (in commit *bugfix - allow bitcoin-cli to work without -wallet parameter*)
- prevents `bitcoin-cli` from reading `-wallet` arguments from the conf file and removes the `-usewallet` argument. That means that to run a command on a wallet in multiwallet mode, the bitcoin-cli user must specify `-wallet` on the command line.

Feel free to squash the first commit and cherry-pick the next two if you think they're useful.

jonasschnelli:
Force push fixed the endpoint issue (thanks @jnewbery). 
I think we should keep it minimal with a hope this can slip in before 0.15 freeze.

Fixes can be done later.

jnewbery:
> Fixes can be done later.

Agree. Multiwallet interface should be considered unstable for v0.15.

achow101:
So this will remove the concept of "default wallet" from the RPCs? Any attempt to use a wallet RPC without specifying a wallet will fail, right?

jonasschnelli:
@achow101: only if you call a wallet RPC and if you run with multiple wallets.

achow101:
@jonasschnelli Ok, cool!

utACK 6b9faf747003995417d6a66fad64d2537c371092

test/functional/multiwallet.py
```diff
@@ -0,0 +1,47 @@
...
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import *
+
+class MultiWalletTest(BitcoinTestFramework):
+
+    def __init__(self):
+        super().__init__()
+        self.setup_clean_chain = True
+        self.num_nodes = 1
+        self.extra_args = [['-wallet=w1', '-wallet=w2', '-wallet=w3']]
```
theuni: Could you change one of these to just 'w' to catch possible length extension bugs? If we're asking for w2, we want to make sure that we don't get w. on accident.

theuni:
Code review utACK 6b9faf747003995417d6a66fad64d2537c371092. No opinion on the endpoint discussion.

sipa (commented):
utACK 6b9faf747003995417d6a66fad64d2537c371092

I like this endpoints based approach, as it's simple for clients to adopt (just make the URI configurable, which immediately prepares it for future changes that move things to other hosts/ports), and does not force adopting named arguments.

As for not introducing 'v1' or separating off node calls, I think this is fine. Once we do introduce a new API, it'll need more thought about which calls are available where (ideally, reasonable clients only need access to one URI).

MeshCollider:
utACK https://github.com/bitcoin/bitcoin/commit/6b9faf747003995417d6a66fad64d2537c371092 looks good to me

morcos:
utACK 6b9faf7

However, this badly needs documentation for the user.

Also I am not happy with the fact that passing multiple -usewallet arguments to the commandline results in the last wallet getting used, that should be an error.

jnewbery:
@morcos - does #10868 (for merge after v0.15) resolve your issue with `-usewallet`?

Suggestions for where to put the documentation? I'm happy to work on that.

NicolasDorier:
This is incompatible with RPC batching. :( (you can't batch a /wallets/ request with one which does not use it)

NicolasDorier:
Ok disregard my comment. I thought that adding /wallet/w1 to a non wallet RPC method would fail. This is not the case, so no issue!

test/functional/multiwallet.py
```diff
@@ -0,0 +1,47 @@
...
+        w1.generate(1)
+
+        #accessing wallet RPC without using wallet endpoint fails
+        assert_raises_jsonrpc(-32601, "Method not found", self.nodes[0].getwalletinfo)
+
+        #check w1 wallet balance
+        walletinfo = w1.getwalletinfo()
+        assert_equal(walletinfo['immature_balance'], 50)
+
+        #check w1 wallet balance
```
MarcoFalke: nit: w2

maflcko (commented):
Post-merge utACK 6b9faf7

PR #10851, opened 2017-07-17
depends: fix fontconfig with newer glibc

theuni:
Fixes build with newer gcc.

Without this, depends builds fail with gcc7, maybe gcc6.

### DIFF
--- a/depends/packages/fontconfig.mk
+++ b/depends/packages/fontconfig.mk
@@ -13,7 +13,13 @@ define $(package)_config_cmds
   $($(package)_autoconf)
 endef
 
+# 2.12.1 uses CHAR_WIDTH which is reserved and clashes with some glibc versions, but newer versions of fontconfig
+# have broken makefiles which needlessly attempt to re-generate headers with gperf.
+# Instead, change all uses of CHAR_WIDTH, and disable the rule that forces header re-generation.
+# This can be removed once the upstream build is fixed.
 define $(package)_build_cmds
+  sed -i 's/CHAR_WIDTH/CHARWIDTH/g' fontconfig/fontconfig.h src/fcobjshash.gperf src/fcobjs.h src/fcobjshash.h && \
+  sed -i 's/fcobjshash.h: fcobjshash.gperf/fcobjshash.h:/' src/Makefile && \
   $(MAKE)
 endef
 


gmaxwell:
Sounds fine to me.

achow101:
This fails to build for me with this warning:
```
/home/andy/bitcoin4/depends/work/build/x86_64-pc-linux-gnu/fontconfig/2.12.4-5ee293e84f5/missing: line 81: gperf: command not found
WARNING: 'gperf' is missing on your system.
         You might have modified some files without having the proper
         tools for further handling them.  Check the 'README' file, it
         often tells you about the needed prerequisites for installing
         this package.  You may also peek at any GNU archive site, in
         case some other package contains this missing 'gperf' program.
```

It looks like travis failed with the same error

theuni:
grr! Upstream borked this. I'll just cherry-pick the patch into our current version instead.

laanwj:
Hehe, that's what you get when you say "fixes build", you get reports of new build problems.

More seriously: this is only used for linking against, right? does raising the fontconfig version raise the minimum fontconfig version on the user's system?

theuni:
@laanwj Yes, this is only for linking against.

That's a good question. We should make sure to check before bumping.

jonasschnelli:
Gitian build errors:

#### Linux:
```
/home/ubuntu/build/bitcoin/depends/work/build/i686-pc-linux-gnu/fontconfig/2.12.4-e57b06ea870/missing: line 81: gperf: command not found
WARNING: 'gperf' is missing on your system.
```

laanwj:
This seems more trouble than it's worth for 0.15.

Could we maybe backport just the build fix for gcc6/7?

MarcoFalke:
This should also be backported to 0.14.3.

laanwj:
Not in the current state at least...

theuni:
@laanwj Yes, will do that today.

laanwj:
Travis is a timeout, and probably unrelated, respinning

MarcoFalke:
Can confirm that 3498a8d018fa02de17e1f9c4cda49f4c7edb3c6c compiles on gcc7.1.1

laanwj:
Changed the PR title to reflect what the commit does :)

PR #10853, opened 2017-07-17
[tests] Fix RPC failure testing (again)

jnewbery:
I did this a few months ago (here: #9707), but a few new examples have crept back in.

When testing RPC failures, the test case should always assert the error value and message, to ensure that the failure was for the correct reason. Not doing that can hide bugs in the test code and mean that the test is not testing the correct behaviour.

RPC failure testing should use the utility function `assert_raises_jsonrpc()` (renamed in the final commit of this PR to `assert_raises_rpc_error()`.

This PR does the following:
- changes all remaining instances of tests directly testing on `JSONRPCException` to calls to `assert_raises_jsonrpc()`
- prevents `assert_raises_message()` from being called with `JSONRPCException`
- scripted-diff changes `assert_raises_jsonrpc()` to `assert_raises_rpc_error()`

### DIFF
--- a/test/functional/bip68-sequence.py
+++ b/test/functional/bip68-sequence.py
@@ -83,7 +83,7 @@ class BIP68Test(BitcoinTestFramework):
         tx2.vout = [CTxOut(int(value-self.relayfee*COIN), CScript([b'a']))]
         tx2.rehash()
 
-        assert_raises_jsonrpc(-26, NOT_FINAL_ERROR, self.nodes[0].sendrawtransaction, ToHex(tx2))
+        assert_raises_rpc_error(-26, NOT_FINAL_ERROR, self.nodes[0].sendrawtransaction, ToHex(tx2))
 
         # Setting the version back down to 1 should disable the sequence lock,
         # so this should be accepted.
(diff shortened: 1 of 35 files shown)


jnewbery:
Rebased and scripted-diff Travis check fixed.

MarcoFalke:
Not sure if it is worth to go through the rebases due to renaming of the method. Mind to drop the last commit for now?

jnewbery:
I wanted the name change for a couple of reasons:

- `assert_raises_jsonrpc()` doesn't make much sense. It's raising an error, so it should include that in the function name. Hopefully changing it to `assert_raises_rpc_error()` will mean people notice the function and use it instead of `assert_raises_message()` or try to roll their own.
- with `--usecli` (see https://github.com/bitcoin/bitcoin/pull/10798), I'm adding a new mode to the test framework, where RPCs are sent via `bitcoin-cli` rather than using JSON over HTTP. In that context, `assert_raises_rpc_error()` makes more sense.

It's a scripted diff changing one function name, so rebasing shouldn't be too difficult.

jnewbery:
rebased

jnewbery:
rebased

MarcoFalke:
Needs rebase

jnewbery:
rebased

MarcoFalke:
Needs silent merge conflicts solved.

jnewbery:
Thanks. scripted-diff rerun and repushed.

MarcoFalke:
utACK 47ba8cf71e9a3a6f7452488fc9ab3b350f0fae36

PR #10858, opened 2017-07-17
[RPC] Add "errors" field to getblockchaininfo and unify "errors" field in get*info RPCs

achow101:
The `getblockchaininfo` output does not contain the `errors` field which the `getinfo`, `getmininginfo`, and `getnetworkinfo` RPCs have. It should have it as the errors pertain to the blockchain. This PR adds that field.

This PR also unifies the help text for the `errors` field and its output position so that all of the `get*info` commands are consistent.

`getnetworkinfo`'s `errors` field is named `warnings`. I did not change this even though it is inconsistent since this naming has been in use for a long time.

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -24,6 +24,7 @@
 #include "util.h"
 #include "utilstrencodings.h"
 #include "hash.h"
+#include "warnings.h"
 
 #include <stdint.h>
 
@@ -1162,7 +1163,7 @@ UniValue getblockchaininfo(const JSONRPCRequest& request)
             "        }\n"
             "     }\n"
             "  }\n"
-            "  \"warnings\" : \"...\",         (string) any network and blockchain errors.\n"
+            "  \"warnings\" : \"...\",         (string) any network and blockchain warnings.\n"
             "}\n"
             "\nExamples:\n"
             + HelpExampleCli("getblockchaininfo", "")
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -1162,6 +1162,7 @@ UniValue getblockchaininfo(const JSONRPCRequest& request)
             "        }\n"
             "     }\n"
             "  }\n"
+            "  \"warnings\" : \"...\",         (string) any network and blockchain errors.\n"
             "}\n"
             "\nExamples:\n"
             + HelpExampleCli("getblockchaininfo", "")
@@ -1201,6 +1202,7 @@ UniValue getblockchaininfo(const JSONRPCRequest& request)
 
         obj.push_back(Pair("pruneheight",        block->nHeight));
     }
+    obj.push_back(Pair("warnings", GetWarnings("statusbar")));
     return obj;
 }
 
--- a/src/rpc/mining.cpp
+++ b/src/rpc/mining.cpp
@@ -202,7 +202,8 @@ UniValue getmininginfo(const JSONRPCRequest& request)
             "  \"networkhashps\": nnn,      (numeric) The network hashes per second\n"
             "  \"pooledtx\": n              (numeric) The size of the mempool\n"
             "  \"chain\": \"xxxx\",           (string) current network name as defined in BIP70 (main, test, regtest)\n"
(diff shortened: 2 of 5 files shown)


laanwj:
IMO, errors does not need to be duplicated on every `get*info` call. It's fine to have it on only one.

achow101:
I have changed the field to be `warnings` instead of errors. However I left `getmininginfo`'s `errors` field as is so that we don't break api compatibility.

jnewbery:
I think we can be more aggressive in improving and harmonizing RPCs, as long as we give plenty of warning and time for clients to upgrade. How about changing `getmininginfo` to return a `warning` field in the next version (currently 0.16), but add a command line argument to have it return the information in an `errors` field. Then remove that option in the following version.

achow101:
Rebased.

@jnewbery I don't like the idea of adding special command line args for dictating whether something is shown in a specific RPC.

jnewbery:
@achow101 - do you disagree with changing RPCs in general? I think we need to have some way of evolving the API over releases otherwise mistakes made in the past are set in stone forever.

If you do agree that the API needs to change over releases, can you suggest a better, safer way of doing it than making the versioning configurable with a command line option? We could obviously add a version path to the RPC endpoint like many REST APIs implement, but I think that's overkill at this stage (and I don't know of anyone who's volunteered implementing it)

achow101:
@jnewbery I agree that the API should change. However I don't think that we should give each deprecated thing its own command line argument to enable or disable it. I think we could do some sort of versioning with different endpoints like what multiwallet does with multiple wallets. So we could add a `/deprecated/` endpoint for which any deprecated behavior would exist behind before being removed in the future. If you want to access deprecated stuff, you can use that endpoint, and ideally, non-deprecated stuff would also work behind that endpoint as well.

jnewbery:
> we could add a /deprecated/ endpoint for which any deprecated behavior would exist behind before being removed in the future. If you want to access deprecated stuff, you can use that endpoint, and ideally, non-deprecated stuff would also work behind that endpoint as well.

I don't think that's ideal. The user would then:

1. upgrade, find that an RPC that they use is deprecated and then have to update their client software to access the new endpoint. Updating client software is more difficult than updating a Bitcoin Core command line option, especially for businesses where the client software might be maintained by a different team.
2. then continue using the `/deprecated` endpoint for everything, because why wouldn't they? At that point we've lost all benefit of having a staged end-of-life for deprecated RPCs.

achow101:
@jnewbery Ok, then.. How about a `-deprecated=<rpc>` option? So instead of a switch for every deprecated thing, the user can specify that some or all RPCs should use the deprecated behavior.

jnewbery:
> How about a -deprecated=<rpc> option?

Yes, sounds sensible. I think `-deprecatedrpcmethod` is probably a bit more descriptive. We could just rename the new option from #11031 .

achow101:
Rebased onto #11031 to use `-deprecatedrpc` for showing `errors` instead of `warnings` for `getmininginfo`. I wasn't sure what to do with the help text, so I just left the `errors` field there with a deprecation warning.

jnewbery:
`p2p-versionbits-warning.py` requires update to use `['warnings']` instead of `['errors']`. Otherwise looks good to me.

achow101:
Fixed `p2p-versionbits-warning.py`

achow101:
Rebased to master

jnewbery:
Tested ACK 395cef7601479b97f5794b0c98067c859f00fc7f

src/rpc/blockchain.cpp
```diff
@@ -1162,6 +1163,7 @@ UniValue getblockchaininfo(const JSONRPCRequest& request)
             "        }\n"
             "     }\n"
             "  }\n"
+            "  \"warnings\" : \"...\",         (string) any network and blockchain warnings.\n"
```
jonasschnelli: nit: I think here is one whitespace missing between text (`warning[...]`and description (`(string)[...]`)

jonasschnelli:
utACK 395cef7601479b97f5794b0c98067c859f00fc7f

PR #10859, opened 2017-07-18
RPC: gettxout: Slightly improve doc and tests

jtimon:
Slightly related to https://github.com/bitcoin/bitcoin/pull/10822 in the sense that I felt the documentation and testing wasn't as good as it could be while writing it.

Ping @sipa since we discussed this on IRC.


src/rpc/blockchain.cpp
```diff
@@ -944,9 +944,10 @@ UniValue gettxout(const JSONRPCRequest& request)
             "gettxout \"txid\" n ( include_mempool )\n"
             "\nReturns details about an unspent transaction output.\n"
             "\nArguments:\n"
-            "1. \"txid\"       (string, required) The transaction id\n"
-            "2. n              (numeric, required) vout number\n"
-            "3. include_mempool  (boolean, optional) Whether to include the mempool\n"
+            "1. \"txid\"             (string, required) The transaction id\n"
+            "2. \"n\"                (numeric, required) vout number\n"
+            "3. \"include_mempool\"  (boolean, optional) Whether to include the mempool. Default: true."
+            "     Note that an unspent output that is spent in the memepool won't appear.\n"
```
sipa: Wow, such meme, many pool.

jtimon: hahaha, yeah let's leave a memepool structure for another PR. fixed

jtimon:
btw, hopefully the nit was solved

sipa:
utACK 7ec368927480c809d27de186690f9d6b6c2e9a9f

test/functional/wallet.py
```diff
@@ -56,6 +56,15 @@ def run_test(self):
         assert_equal(len(self.nodes[1].listunspent()), 1)
         assert_equal(len(self.nodes[2].listunspent()), 0)
 
+        self.log.info("test gettxout")
+        confirmed_txid, confirmed_index = utxos[0]["txid"], utxos[0]["vout"]
+        # First, outputs that are unspent both in the chain and in the
+        # mainpool should appear with or without include_mempool
```
MarcoFalke: mempool

MarcoFalke:
utACK 7ec368927480c809d27de186690f9d6b6c2e9a9f after fixup

jtimon:
Fixed @MarcoFalke 's nit

test/functional/wallet.py
```diff
@@ -56,6 +56,15 @@ def run_test(self):
...
 
+        self.log.info("test gettxout")
+        confirmed_txid, confirmed_index = utxos[0]["txid"], utxos[0]["vout"]
+        # First, outputs that are unspent both in the chain and in the
+        # mempool should appear with or without include_mempool
+        txout = self.nodes[0].gettxout(txid=confirmed_txid, n=confirmed_index, include_mempool=False)
+        assert_equal(txout['value'], 50)
+        txout = self.nodes[0].gettxout(txid=confirmed_txid, n=confirmed_index, include_mempool=True)
+        assert_equal(txout['value'], 50)
+
```
MarcoFalke: nit: There is trailing whitespace, causing issues with some editors. We don't have this rule in the developer notes, but good to keep in mind for the future.

MarcoFalke:
re-utACK 6d2d2eb

src/rpc/blockchain.cpp
```diff
@@ -944,9 +944,10 @@ UniValue gettxout(const JSONRPCRequest& request)
             "gettxout \"txid\" n ( include_mempool )\n"
             "\nReturns details about an unspent transaction output.\n"
             "\nArguments:\n"
-            "1. \"txid\"       (string, required) The transaction id\n"
-            "2. n              (numeric, required) vout number\n"
-            "3. include_mempool  (boolean, optional) Whether to include the mempool\n"
+            "1. \"txid\"             (string, required) The transaction id\n"
+            "2. \"n\"                (numeric, required) vout number\n"
+            "3. \"include_mempool\"  (boolean, optional) Whether to include the mempool. Default: true."
+            "     Note that an unspent output that is spent in the mempool won't appear.\n"
```
promag: Mixed feelings when reading the note, I thought an unspent output is never spent weather the transaction is in the mempool or not, otherwise is not an unspent output.

Suggestion, remove `unspent`: `Note that an output that is spent in the mempool won't appear`.

test/functional/wallet.py
jtimon: Perhaps it is time to add it to the notes. Is is save right now to have your editor configured to remove all trailing whitespaces in any file you save in the project now?
I remember when that wouldn't be save in the sense that your +14-5 commit would turn into +60-50 or something.

src/rpc/blockchain.cpp
promag: @MarcoFalke it's merged but WDYT?

MarcoFalke: "unspent output" refers to the chainstate, whereas "spent in the mempool" has no effect on the chainstate. Imo it is fine.

PR #10863, opened 2017-07-18
Fix leaking a socket handle if call to SetSocketNonBlocking failed

bytting:


corebob:
There is no leak here. My mistake. Closing

PR #10867, opened 2017-07-18
Allow only a single -usewallet argument

jonasschnelli:
Currently, one can specify multiple `-usewallet` arguments in bitcoin-cli (Including `bitcoin.conf`) where the last passed-in `-usewallet` will be used. This can result in executing critical actions on the wrong wallet.

This PR halts bitcoin-cli if multiple `-usewallet`-arguments where passed-in.

Tagged for 0.15 (in case we consider this a bug-fix).

Reported by @morcos (https://github.com/bitcoin/bitcoin/pull/10849#issuecomment-316065717)

### DIFF
--- a/src/bitcoin-cli.cpp
+++ b/src/bitcoin-cli.cpp
@@ -300,8 +300,12 @@ int CommandLineRPC(int argc, char *argv[])
             while (std::getline(std::cin,line))
                 args.push_back(line);
         }
-        if (args.size() < 1)
+        if (args.size() < 1) {
             throw std::runtime_error("too few parameters (need at least command)");
+        }
+        if (gArgs.GetArgs("-usewallet").size() > 1) {
+            throw std::runtime_error("Only one -usewallet arguments is allowed");
+        }
         std::string strMethod = args[0];
         args.erase(args.begin()); // Remove trailing method name from arguments vector
 


jnewbery:
You talk in #10868 about having a default wallet specified in .conf by using `-usewallet`. I think this PR would mean that if you did have a `-usewallet` in .conf, then you'd never be able to issue commands to any of the other wallets (since trying to specify a `-usewallet` override on the command line would trigger your `"Only one -usewallet arguments is allowed"` throw). Is that the intention?

jonasschnelli:
> You talk in #10868 about having a default wallet specified in .conf by using -usewallet. I think this PR would mean that if you did have a -usewallet in .conf, then you'd never be able to issue commands to any of the other wallets (since trying to specify a -usewallet override on the command line would trigger your "Only one -usewallet arguments is allowed" throw). Is that the intention?

Yes. I think that would make sense.
If a user defines a default wallet via explicitly place a `-usewallet` in `bitcoin.conf`, then rejecting an override via `"Only one -usewallet arguments is allowed"` seems reasonable.
IMO this would be the right level of encourage / discourage to use a default wallet (allow it if you always want to use it, disallow it if you want to mix usage between specifying a wallet and having a default wallet).

jnewbery:
In that case, I don't think this is a useful change. It would mean that users would be able to load a wallet and then not do anything with it. I think it'd be a really frustrating experience to see a wallet loaded by bitcoin, but the only way to access it is to stop the server, change the .conf file and then restart.

jonasschnelli:
> In that case, I don't think this is a useful change. It would mean that users would be able to load a wallet and then not do anything with it. I think it'd be a really frustrating experience to see a wallet loaded by bitcoin, but the only way to access it is to stop the server, change the .conf file and then restart.

Not sure if I am wrong, but...
1. The wallet does also something if you can't access it via bitcoin-cli, a) it syncs, b) you can access it with other tools (RPC, maybe later Qt).
2. If you have decided to use a default wallet via `-usewallet` in your conf, and then want to access other wallets *with bitcoin-cli*, then just remove `-usewallet` from the conf (no need to restart the server). This seems to be the right level of encourage/discourage for the using a default wallet.

jnewbery:
Yes, you're right. You can do this by changing .conf without restarting bitcoind, although I think that generally people shouldn't do that. Having inconsistent .conf/running config can be confusing.

morcos:
Prefer #10868

jonasschnelli:
Closing. Either we do that globally or not.

PR #10868, opened 2017-07-18
Remove -usewallet

jnewbery:
For merge after v0.15 (**or perhaps as a bugfix for v0.15**)

1. Adds better validation of bitcoin-cli wallet argument to make sure it is only specified once, and that is only ever picked up from the command line, not from a config file (which could have bad results if someone was writing a script and forgot a command line argument).
2. Renames -usewallet to -wallet

If there is zero or one `-wallet` argument in bitcoin.conf (ie single wallet mode), **there is no change from pre-0.15 behaviour**. No `-wallet` argument is required for bitcoin.cli.

If there is more than one `-wallet` argument in bitcoin.conf (ie multiwallet), then a single command-line `-wallet` argument must be provided to bitcoin.cli.

It is always an error to specify more than one `-wallet` command line argument, and the request will be rejected by bitcoin.cli on the client side.

### DIFF
--- a/src/bitcoin-cli.cpp
+++ b/src/bitcoin-cli.cpp
@@ -46,7 +46,7 @@ std::string HelpMessageCli()
     strUsage += HelpMessageOpt("-rpcpassword=<pw>", _("Password for JSON-RPC connections"));
     strUsage += HelpMessageOpt("-rpcclienttimeout=<n>", strprintf(_("Timeout in seconds during HTTP requests, or 0 for no timeout. (default: %d)"), DEFAULT_HTTP_CLIENT_TIMEOUT));
     strUsage += HelpMessageOpt("-stdin", _("Read extra arguments from standard input, one per line until EOF/Ctrl-D (recommended for sensitive information such as passphrases)"));
-    strUsage += HelpMessageOpt("-usewallet=<walletname>", _("Send RPC for non-default wallet on RPC server (argument is wallet filename in bitcoind directory, required if bitcoind/-Qt runs with multiple wallets)"));
+    strUsage += HelpMessageOpt("-wallet=<walletname>", _("Send RPC for non-default wallet on RPC server (argument is wallet filename in bitcoind directory, required if bitcoind/-Qt runs with multiple wallets)"));
 
     return strUsage;
 }
@@ -103,8 +103,27 @@ static int AppInitRPC(int argc, char* argv[])
         fprintf(stderr, "Error: Specified data directory \"%s\" does not exist.\n", GetArg("-datadir", "").c_str());
         return EXIT_FAILURE;
     }
+    // For bitcoin-cli, the -wallet parameter can only be specified from the
+    // command line, not in the conf file. This is because the conf file can
+    // contain multiple -wallet args, so it's not possible to determine which
+    // one to use.
+    //
+    // Store the command line -wallet arg here so we can restore it after
+    // parsing the conf file.
+    std::string wallet;
+    if (gArgs.GetArgs("-wallet").size() > 1) {
+        fprintf(stderr, "Error: Only one -wallet option may be used.\n");
+        return EXIT_FAILURE;
+    } else if (gArgs.GetArgs("-wallet").size() == 1) {
+        wallet = (gArgs.GetArgs("-wallet"))[0];
+    }
     try {
         ReadConfigFile(GetArg("-conf", BITCOIN_CONF_FILENAME));
+        // restore the command line -wallet argument
+        gArgs.ClearArg("-wallet");
+        if (!wallet.empty()) {
+            gArgs.SoftSetArg("-wallet", wallet);
+        }
     } catch (const std::exception& e) {
         fprintf(stderr,"Error reading configuration file: %s\n", e.what());
         return EXIT_FAILURE;
@@ -244,7 +263,7 @@ UniValue CallRPC(const std::string& strMethod, const UniValue& params)
 
     // check if we should use a special wallet endpoint
     std::string endpoint = "/";
-    std::string walletName = GetArg("-usewallet", "");
+    std::string walletName = GetArg("-wallet", "");
     if (!walletName.empty()) {
         char *encodedURI = evhttp_uriencode(walletName.c_str(), walletName.size(), false);
         if (encodedURI) {
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -482,7 +482,13 @@ void ArgsManager::ForceSetArg(const std::string& strArg, const std::string& strV
     mapMultiArgs[strArg].push_back(strValue);
 }
 
-
+bool ArgsManager::ClearArg(const std::string& strArg)
+{
+    LOCK(cs_args);
+    return mapArgs.erase(strArg) | mapMultiArgs.erase(strArg);
+    // bitwise or is deliberate. erase() returns number of items erased. We don't want
+    // to short-circuit the mapMultiArgs erase().
+}
 
 static const int screenWidth = 79;
 static const int optIndent = 2;
--- a/src/util.h
+++ b/src/util.h
@@ -259,6 +259,14 @@ public:
     // Forces an arg setting. Called by SoftSetArg() if the arg hasn't already
     // been set. Also called directly in testing.
     void ForceSetArg(const std::string& strArg, const std::string& strValue);
+
+    /**
+     * Clears an argument if it is already set
+     *
+     * @param strArg Argument to set (e.g. "-foo")
+     * @return true if argument gets cleared, false if there was no argument
+     */
+    bool ClearArg(const std::string& strArg);
 };
 
 extern ArgsManager gArgs;


jnewbery:
@ryanofsky - should address your concern about `usewallet`

ryanofsky (commented):
Thanks for doing this. I think this is a better interface, and it also addresses @morcos comment about conflicting `-usewallet` values. Implementation could and probably should be simplified, though.

For anybody reading the PR description, this might sound pretty complicated, but the description is mostly  referring to existing behavior that isn't changing. In particular none of what's described about how bitcoind interprets arguments or handles RPCs is changing in this PR. This change only does two things:

1. Adds better validation of `bitcoin-cli` wallet argument to make sure it is only specified once, and that is only ever picked up from the command line, not from a config file (which could have bad results if someone was writing a script and forgot a command line argument).
2. Renames `-usewallet` to `-wallet`

The first part is arguably a bugfix, so maybe this could be included in 0.15.

jnewbery:
Thanks for the feedback. You're right - the PR description was overly long and confusing for something that's actually very simple. I've reworked it using your words.

Yes, perhaps this could go in 0.15 as a bugfix.

Will address your technical feedback soon.

sipa:
If we want to replace `-usewallet` with `-wallet` for `bitcoin-cli`, it should go in 0.15 or we'd be breaking backward compatibility in later?

jnewbery:
> If we want to replace -usewallet with -wallet for bitcoin-cli, it should go in 0.15 or we'd be breaking backward compatibility in later?

Yes, but as I've mentioned before, I'm not very concerned about backward compatibility for an interface that should be marked as experimental.

That said, I'm happy for this to go in v0.15 as a bugfix.

sipa:
> I'm not very concerned about backward compatibility for an interface that should be marked as experimental.

Still no reason to go intentionally plan to break it.

jnewbery:
> Still no reason to go intentionally plan to break it.

Sorry, I'm confused about what you're advocating for? I've already said I'm happy for this to go into v0.15 as a bugfix.

sipa:
@jnewbery I think we should either decide to have it, and get it in now, or not at all.

jnewbery:
ok, my vote is for having it now.

aside: on a philosophical level I disagree that we should be forever bound to an interface that hasn't been fully agreed or designed. Presumably any users that are active enough to enable multiwallet in its first release and update all their client software are perfectly able to change `usewallet` to `wallet` at a later date.

src/bitcoin-cli.cpp
```diff
@@ -103,8 +103,31 @@ static int AppInitRPC(int argc, char* argv[])
         fprintf(stderr, "Error: Specified data directory \"%s\" does not exist.\n", GetArg("-datadir", "").c_str());
         return EXIT_FAILURE;
     }
+    // For bitcoin-cli, the -wallet parameter can only be specified from the
+    // command line, not in the conf file. This is because the conf file can
+    // contain multiple -wallet args, so it's not possible to determine which
+    // one to use.  Furthermore, if bitcoin-cli used the -wallet arg from the
+    // conf file, we'd get 'trapped' in wallet mode and not be able to issue
```
jnewbery: Thanks. Fixed

src/util.cpp
```diff
@@ -482,7 +482,21 @@ void ArgsManager::ForceSetArg(const std::string& strArg, const std::string& strV
     mapMultiArgs[strArg].push_back(strValue);
 }
 
-
+bool ArgsManager::ClearArg(const std::string& strArg)
+{
+    LOCK(cs_args);
```
jnewbery: Neat! Thanks. Fixed

src/bitcoin-cli.cpp
```diff
@@ -103,8 +103,31 @@ static int AppInitRPC(int argc, char* argv[])
...
+    // commands to the node.
+    //
+    // Store the command line -wallet arg here so we can restore it after
+    // parsing the conf file.
+    bool use_wallet = false;
```
jnewbery: Yes, much better. Thanks

src/bitcoin-cli.cpp
```diff
@@ -103,8 +103,31 @@ static int AppInitRPC(int argc, char* argv[])
...
+        fprintf(stderr, "Error: Only one -wallet option may be used.\n");
+        return EXIT_FAILURE;
+    } else if (gArgs.GetArgs("-wallet").size() == 1) {
+        wallet_name = (gArgs.GetArgs("-wallet"))[0];
+        use_wallet = true;
+    }
     try {
         ReadConfigFile(GetArg("-conf", BITCOIN_CONF_FILENAME));
+        // restore the command line -wallet argument
+        gArgs.ClearArg("-wallet");
```
jnewbery: I prefer not to set this as a global or pass it through `main()`, `CommandLineRPC()` and `CallRPC()`. It makes sense to me to set up all the config in `AppInitRPC()` and have it contained in the standard `gArgs` global.

Happy to reconsider if making changes to util.cpp is a blocker for this PR.

jnewbery:
Addressed @ryanofsky's comments

src/bitcoin-cli.cpp
ryanofsky: > Happy to reconsider if making changes to util.cpp is a blocker for this PR.

Not a big deal to me. If someone else objects to the util.cpp change though, we should just be aware that it isn't a critical part of the PR and there are various ways to get the same behavior.

ryanofsky (commented):
utACK 561d0190d326c4bb75d9db03cabbeea10b9ec5cb

gmaxwell:
But I want to be able to tell my daemon to load wallets one two and three, while make my bitcoin-cli default to wallet one.

ryanofsky:
> But I want to be able to tell my daemon to load wallets one two and three, while make my bitcoin-cli default to wallet one.

We could allow this, but the thought is that it's dangerous because maybe you'll forget to add an explicit -wallet=filename argument in one of the cases where you don't intend to use your default wallet (maybe in a single bitcoin-cli invocation as part of a larger script) and then you will accidentally perform an action on the wrong wallet.

gmaxwell:
@ryanofsky  That danger doesn't hold w the only allow one argument change.

jnewbery:
> But I want to be able to tell my daemon to load wallets one two and three, while make my bitcoin-cli default to wallet one.

Then write a wrapper around your cli :slightly_smiling_face: 

I agree with @ryanofsky - not having a default wallet is *much* safer for all users.

gmaxwell:
A setting I explicitly set isn't a default though. I agree it shouldn't have a default.  The argument you're making would apply identically to the testnet vs mainnet switch (which is also multiple wallets and also unsafe if you get the wrong one).

ryanofsky:
Yeah, I'm not actually against allowing users to explicitly set a default wallet. I was just trying to explain what the rationale was.

@gmaxwell, are you suggesting with the one-argument limit that you might want to allow users to specify a default wallet in the conf file, but then disallow specifying a second wallet on the command line that would override it? That seems like a slightly weird behavior, but I guess could have some use cases and would be safe by john's definition.

jnewbery:
I agree - the `testnet` / `mainnet` switch in the .conf file is dangerous and has tripped lots of people up, including me several times. Having to specify it explicitly it for each bitcoin-cli call would be safer (but irritating).

I'm not even clear on what exact functionality you want - are you saying `-usewallet` in the conf file should be the default? But that it can be overridden by `-usewallet` on the command line?

ryanofsky:
If we are ok with explicitly set default wallets, maybe it would best to have an "-defaultwallet" config setting that gets implemented on server side rather than client side. This would allow for consistent behavior even across different clients.

sipa:
What is the rationale for renaming -usewallet to -wallet (over just rejecting duplicate -usewallet and/or -defaultwallet)?

ryanofsky:
> What is the rationale for renaming -usewallet to -wallet (over just rejecting duplicate -usewallet and/or -defaultwallet)?

From my point of view I think the consistency is good. If we just have a consistent "-wallet=filename" option accepted by `bitcoind`, `bitcoin-qt`, `bitcoin-cli` tools, it is harder to screw up using them. Especially if we want to have support for default wallets like Greg wants, it seems like asking for trouble expecting users to remember "-usewallet" instead of "-wallet" on the command line for this one tool, and punishing them if they forget by not showing any error and falling back to another wallet written in some config file.

sipa:
I'm not sure it's a great thing for consistency, if it needs such exceptional processing (ignoring something in the conf file is unique...).

jnewbery:
Exactly what @ryanofsky said.

And less importantly, because `-use` is content free. All arguments are used.

ryanofsky:
> I'm not sure it's a great thing for consistency, if it needs such exceptional processing (ignoring something in the conf file is unique...).

I don't necessarily think it should be ignored in the conf file. That's really just a quirk of the way this PR is implemented. I think my preferred implementation would check to see if a command line argument is specified, and if so override anything in the config file (obviously). Else if no command line argument is specified, and the configuration file has more than one value specified, exit with an error message, otherwise use the value from the config file. Behavior would be indistinguishable in the common case where bitcoind and bitcoincli share the same config file.

I think John favored current approach though to prioritize safety just always get the wallet name from the command line when an RPC call is being made.

jnewbery:
> That's really just a quirk of the way this PR is implemented.

Correct - in terms of implementation, it's simpler to just clear out the .conf file wallet argument (since wallet is a multimap). The end result is exactly as Russ has described.

I'm happy to review alternative implementations if that's confused people.

sipa:
I think my objection goes away if -wallet in the config file is not being ignored, but there are just rules that prevent specifying it twice on the command line.

sipa:
After some offline discussion, Concept ACK. I misunderstood how little effect "ignore the config file" had on the semantics of -wallet.

I believe the behaviour of the -wallet option can effectively be modelled as "the command-line -wallet arguments replace the config file ones as a whole, rather than appending to them". I think that's fine.

Having a `-defaultwallet` can be an independent feature that is added later, or not.

ryanofsky:
Sounds good. I'm happy with this PR as it is, as well as the other options that have been discussed. Note that the current PR doesn't exactly model command line -wallet options "replacing config file ones as a whole", because it clears config wallet options even when no command line option specified. The change I was talking about earlier would give you that behavior (and potentially less safe results when bitcoind and bitcoin-cli read different config files). FWIW, I implemented it here https://github.com/ryanofsky/bitcoin/commit/b701085e0c35e9e09f44d8e9ba8303617e445ca3

jnewbery:
@ryanofsky - I think your branch is much nicer (modulo the two things we just chatted about). Mind if I cherry pick it here?

achow101:
ACK 561d0190d326c4bb75d9db03cabbeea10b9ec5cb

jonasschnelli:
Agree with @jnewbery that @ryanofsky ryanofsky/bitcoin@b701085 is preferable.

I like the current `-usewallet` for the reasons that users can/will/must clearly distinct between running with multiple wallets (d/QT `-wallet`) and accessing a specific wallet (`-usewallet`).

I'm aware of the different layers (d/Qt vs `bitcoin-cli`) but from the users perspective, differentiation of an argument between the daemon and the cli is difficult to understand (once its running multiple wallets, once its selecting a specific wallet). Having two identical arguments that have total different effects is not ideal (even cross binaries).

I think for 0.15 we should merge the multiple `-usewallet` arguments blocker: #10867 

Also, if we want to go with a default wallet (in future and we already do this for Qt [Index 0]), we could reuse the `-usewallet` argument for the daemon and/or-only Qt.

ryanofsky:
#10867 doesn't address usability problem I referred to above where user forgets "use" prefix in a bitcoin-cli call, their `-wallet=filename` is ignored by bitcoin-cli, resulting in server error or the RPC call ignoring specified filename and accessing wrong wallet if server configured with a different single wallet.

Having `-usewallet` vs `-wallet` distinction was briefly necessary while the wallet endpoint didn't allow node calls, but now having two differently spelled options is just adding a footgun and unnecessary source of confusion.

Anyway, I believe John is currently working on an improved version of ryanofsky/bitcoin@b701085 that actually shrinks it down further, removes global variable, and fixes bug where I forgot to remove "use" prefix (footgun!).

jonasschnelli:
I can't follow the potential source of confusion. Please help me to understand.

### The current master approach is:
* Daemon (server) uses `-wallet` to specify (additional) wallets to run with.
* CLI (client) uses `-usewallet` to specify which of the available wallets (specified though daemons `-wallet`) should be used.
* `bitcoin.conf`'s `-wallet` arguments do not affect `bitcoin-cli`.
* `bitcoin.conf`'s `-usewallet` arguments does allow to define a default wallet (sounds okay to me if someone adds this intentionally, the term `use` makes it clear).
* providing an `-usewallet` argument in CLI will "override" the default bitcoin.conf `-usewallet` directive (can be blocked by merging #10867).

### This PR would lead to:
* Daemon and client do both use `-wallet` (once for defining multiple wallets, once for defining a single wallet to use for executing a command, which are two completely different use-cases).
* Disabling the option to have a default wallet by explicitly setting `-usewallet` in `bitcoin.conf` (may be okay)
* Add code to detect and reject the `-wallet` arguments in `bitcoin.conf` and only accept them via command line (?)

IMO from the users perspective, the current master approach seems much more logical.
Also, I don't see the footgun source (both approaches have risks selecting the wrong wallet).

jonasschnelli:
> A simple example of confusion (mentioned twice above) is user forgets to add "use" prefix to bitcoin-cli and the argument gets ignored.

If he forgets the `use` prefix and uses `-wallet` instead, the call would not be executed in multiwallet environment because he had not specified a wallet.

I don't think we should avoid setting a default wallet via `-usewallet` in `bitcoin.conf`, this seems to me after a desirable use-cases and, because it's different to `-wallet` and has the term `-use` in it, I think its acceptable.

If we don't want that, then we should ignore -usewallet in the conffile-reader logic.

> Ideal scenario to me is our tools consistently accept -wallet=filename to denote which wallet to use. 

But that would not be true for bitcoind/Qt because you don't specify which wallet (single) to use (you do more specify which wallets you want to have accessible). Especially for Qt that's not true (there always index 0 will be used).

I fail to see the advantages of having a single `-wallet` argument for both tools because IMO they have different implications. I think it is a source of confusion.

ryanofsky:
> If he forgets the use prefix and uses -wallet instead, the call would not be executed in multiwallet environment because he had not specified a wallet.

Yes, that is what I meant by "-wallet=filename is ignored by bitcoin-cli, resulting in server error or the RPC call ignoring specified filename and accessing wrong wallet if server configured with a different single wallet." The worse case of the two is the case where server is single wallet and filename provided just gets silently ignored instead of checked on the server. Anyway, even if we disagree about how bad it is for bitcoin-cli to silently ignore "-wallet" options, I think we can agree that in the most common use-case (calling bitcoin-cli with a wallet filename on command line and no custom default config), having differently spelled `-usewallet` and `-wallet` arguments that get used by some commands and ignored by others does add some new opportunities for confusion and errors even if we disagree about how bad the confusion and errors are. And in any case (I will explain below) adding confusion to command line interface to support a default wallet feature is unnecessary.

> I don't think we should avoid setting a default wallet via -usewallet in bitcoin.conf, this seems to me after a desirable use-cases and, because it's different to -wallet and has the term -use in it, I think its acceptable.

I agree that being able to set default wallet is useful. All I'm saying is that better way to do it is to call the default wallet option `-defaultwallet` instead of `-usewallet` and implement it server side instead of client side. Implementing it server side would have advantage of working with clients other than `bitcoin-cli`. It would avoid adding unnecessary complexity to command line in the common case where it isn't used, and avoid the complicated interactions you described in your five bullet points when it is used. Finally, it could interact better with per-user default wallets (which are necessarily implemented server-side) which luke-jr at least has mentioned use cases for if we decide to add support for them in the future.

jnewbery:
I think I've sown unnecessary confusion with this PR. All that it's doing is:

- If bitcoind is running multiwallet, bitcoin-cli must be called with exactly one `-wallet` command line argument to access wallet RPCs.

That's it. With singlewallet mode, everything runs exactly as before. With non-wallet RPCs, everything runs exactly as before. If more than one `-wallet` argument is provided on the command line, the call fails.

Having two different arguments would make this more confusing, especially if one of them is called `usewallet`. How do users know that `usewallet` means *bitcoin-cli uses this but not bitcoind* and `wallet` means *bitcoind uses this but not bitcoin-cli*? How do they know that `usewallet` in .conf means *default wallet* and on the command line means *override default wallet*? How do they know that `usewallet` is the default wallet if you're using bitcoin-cli but there is no default wallet if you're using RPC over HTTP? They don't, because the `use` is `usewallet` is ambiguous enough to be meaningless.

morcos:
Concept ACK

ryanofsky:
https://github.com/ryanofsky/bitcoin/commit/4ae1c6858e55f6f715918fe3c82403a8c8e69828 is another implementation of this PR with identical behavior, which might be a little easier to discuss.

Sadly, though, treating command line arguments exactly the same as config file arguments might be an inviolable principle in bitcoin that would make these approaches acceptable. So users might have to deal with differently spelled wallet options for different commands. Not ideal from my point of view (and of course predictably avoidable if we just treated wallet filenames like normal parameters https://github.com/bitcoin/bitcoin/pull/10829#issuecomment-315941661), but fine.

jnewbery:
As far as I can tell, this PR is functionally equivalent to the command-line `-wallet` replacing the .conf file `wallet` args, which is standard for all arguments. It has to be handled slightly carefully because of the way mapmultiargs work, and if we really wanted to be formal about "the command-line -wallet arguments replace the config-file one as a whole, rather than appending to", then we'd need bitcoin-cli to have knowledge of which RPCs are wallet calls and which are server calls.

I think @ryanofsky's branch is as good as this, although I haven't manually tested it yet.

laanwj:
I prefer not to overload `-wallet` both for bitcoind, which has "what wallets to load", and bitcoin-cli which has "what wallet to choose".
These meanings are different enough. Also one option is multi-value, the other single-value. This requires special casing to make work, which I don't think is worth it.
(if bitcoind and bitcoin-cli read different configuration files, this issue wouldn't exist as there would be no overlap in namespaces - but alas)

The current approach of introducing a new argument name for bitcoin-cli to select the wallet (now `-rpcwallet`) has less potential of confusing, and doesn't require special casing of argument handling, so is preferable to me.

ryanofsky:
Should be closed? (#10883 is merged)

jnewbery:
I still don't understand why `-rpcwallet` is preferred. bitcoind is a node where `-wallet` means "this is the wallet I wish to use". Since 0.15, multiple wallets are allowed. bitcoin-cli is a command line interface where `wallet` would mean "this is the wallet I wish to issue a command to". Obviously only one `wallet` can be used in this context. So yes, perhaps this is a little bit overloaded, but only in the most intuitive and expected way.

As for special-casing, the only way `wallet` is treated functionally differently is that in bitcoin-cli a `wallet` on the command-line *overrides* `wallet` in .conf, but for bitcoind `wallet` on the command-line *appends* to the `wallets` in bitcoin.conf. That's really the only way it could work, since appending doesn't make sense for bitcoin-cli where only one `wallet` is allowed.

(and FWIW, command-line arguments *appending* to .conf arguments instead of replacing them is very counter-intuitive)

But yes, you're right - the decision has been made. I'll close this.

PR #10870, opened 2017-07-18
[Qt] Use wallet 0 in rpc console if running with multiple wallets

jonasschnelli:
Current master with multiwallet results in accessing wallet 0 in QT (send / receive / tx history / etc.), **but** the RPC console cannot access that wallet (only non-wallet commands work).

This is a quick solution to re-allow accessing the same wallet (Index 0) via RPC console in multiwallet.

The solutions design is not "state of the art" (should go over WalletModel). Ideally we work on an overall multiwallet support for the GUI (which then would remove this change).

I think we should consider this as a bugfix.

### DIFF
--- a/src/qt/rpcconsole.cpp
+++ b/src/qt/rpcconsole.cpp
@@ -25,6 +25,7 @@
 
 #ifdef ENABLE_WALLET
 #include <db_cxx.h>
+#include <wallet/wallet.h>
 #endif
 
 #include <QKeyEvent>
@@ -301,6 +302,14 @@ bool RPCConsole::RPCParseCommandLine(std::string &strResult, const std::string &
                             JSONRPCRequest req;
                             req.params = RPCConvertValues(stack.back()[0], std::vector<std::string>(stack.back().begin() + 1, stack.back().end()));
                             req.strMethod = stack.back()[0];
+#ifdef ENABLE_WALLET
+                            // TODO: Move this logic to WalletModel
+                            if (!vpwallets.empty()) {
+                                // in Qt, use always the wallet with index 0 when running with multiple wallets
+                                QByteArray encodedName = QUrl::toPercentEncoding(QString::fromStdString(vpwallets[0]->GetName()));
+                                req.URI = "/wallet/"+std::string(encodedName.constData(), encodedName.length());
+                            }
+#endif
                             lastResult = tableRPC.execute(req);
                         }
 


src/qt/rpcconsole.cpp
```diff
@@ -301,6 +302,14 @@ bool RPCConsole::RPCParseCommandLine(std::string &strResult, const std::string &
                             JSONRPCRequest req;
                             req.params = RPCConvertValues(stack.back()[0], std::vector<std::string>(stack.back().begin() + 1, stack.back().end()));
                             req.strMethod = stack.back()[0];
+#ifdef ENABLE_WALLET
+                            // TODO: Move this logic to WalletModel
+                            if (vpwallets.size() > 1)
+                            {
```
promag: Nit, same line as `if`.

promag (commented):
Tested ACK 2044595. Is this WIP (because of the TODO) or final?

src/qt/rpcconsole.cpp
jonasschnelli: I think we should keep the TODO to make sure we clean that up asap (once 0.15 has been split off).

laanwj: Wouldn't this be safe to do for vpwallets.size()==1 too? If so, I'd slightly prefer that, because it means this code is always used when wallet is enabled.

src/qt/rpcconsole.cpp
```diff
@@ -301,6 +302,14 @@ bool RPCConsole::RPCParseCommandLine(std::string &strResult, const std::string &
...
+                                // in Qt, use always the wallet with index 0 when running with multiple wallets
+                                req.URI = "/wallet/"+QUrl(QString::fromStdString(vpwallets[0]->GetName())).toString().toStdString();
```
laanwj: What about
```c++
QByteArray encodedName = QUrl::toPercentEncoding(QString::fromStdString(vpwallets[0]->GetName())));
req.URI = "wallet/"+std::string(encodedName.constData(), encodedName.length());
```
Still fairly ugly, but this makes it clear that the point is to URL-encode (percent encode) the wallet name. Not convinced that passing a string into `QUrl()` does the right thing in all cases.

jonasschnelli:
Implemented @laanwj's recommendation, fixed @promag's nit.

promag:
Tested ACK 9737572.

PR #10871, opened 2017-07-18
Handle getinfo in bitcoin-cli w/ -getinfo (revival of #8843)

achow101:
Since @laanwj doesn't want to maintain these changes anymore, I will.

This PR is a revival of #8843. I have addressed @jnewbery's comments.

Regarding atomicity, I don't think that is a concern here. This is explicitly a new API and those who use it will know that this is different and that it is not atomic.

### DIFF
--- a/src/bitcoin-cli.cpp
+++ b/src/bitcoin-cli.cpp
@@ -37,6 +37,7 @@ std::string HelpMessageCli()
     strUsage += HelpMessageOpt("-?", _("This help message"));
     strUsage += HelpMessageOpt("-conf=<file>", strprintf(_("Specify configuration file (default: %s)"), BITCOIN_CONF_FILENAME));
     strUsage += HelpMessageOpt("-datadir=<dir>", _("Specify data directory"));
+    strUsage += HelpMessageOpt("-getinfo", _("Get general information from the remote server. Note that unlike server-side RPC calls, the results of -getinfo is the result of multiple non-atomic requests. Some entries in the result may represent results from different states (e.g. wallet balance may be as of a different block from the chain state reported)"));
     AppendParamsHelpMessages(strUsage);
     strUsage += HelpMessageOpt("-named", strprintf(_("Pass named instead of positional arguments (default: %s)"), DEFAULT_NAMED));
     strUsage += HelpMessageOpt("-rpcconnect=<ip>", strprintf(_("Send commands to node running on <ip> (default: %s)"), DEFAULT_RPCCONNECT));
@@ -191,7 +192,96 @@ static void http_error_cb(enum evhttp_request_error err, void *ctx)
 }
 #endif
 
-static UniValue CallRPC(const std::string& strMethod, const UniValue& params)
+/** Class that handles the conversion from a command-line to a JSON-RPC request,
+ * as well as converting back to a JSON object that can be shown as result.
+ */
+class BaseRequestHandler
+{
+public:
+    virtual UniValue PrepareRequest(const std::string& method, const std::vector<std::string>& args) = 0;
+    virtual UniValue ProcessReply(const UniValue &batch_in) = 0;
+};
+
+/** Process getinfo requests */
+class GetinfoRequestHandler: public BaseRequestHandler
+{
+public:
+    const int ID_NETWORKINFO = 0;
+    const int ID_BLOCKCHAININFO = 1;
+    const int ID_WALLETINFO = 2;
+
+    /** Create a simulated `getinfo` request. */
+    UniValue PrepareRequest(const std::string& method, const std::vector<std::string>& args) override
+    {
+        UniValue result(UniValue::VARR);
+        result.push_back(JSONRPCRequestObj("getnetworkinfo", NullUniValue, ID_NETWORKINFO));
+        result.push_back(JSONRPCRequestObj("getblockchaininfo", NullUniValue, ID_BLOCKCHAININFO));
+        result.push_back(JSONRPCRequestObj("getwalletinfo", NullUniValue, ID_WALLETINFO));
+        return result;
+    }
+
+    /** Collect values from the batch and form a simulated `getinfo` reply. */
+    UniValue ProcessReply(const UniValue &batch_in) override
+    {
+        UniValue result(UniValue::VOBJ);
+        std::vector<UniValue> batch = JSONRPCProcessBatchReply(batch_in, 3);
+        // Errors in getnetworkinfo() and getblockchaininfo() are fatal, pass them on
+        // getwalletinfo() is allowed to fail in case there is no wallet.
+        if (!batch[ID_NETWORKINFO]["error"].isNull()) {
+            return batch[ID_NETWORKINFO];
+        }
+        if (!batch[ID_BLOCKCHAININFO]["error"].isNull()) {
+            return batch[ID_BLOCKCHAININFO];
+        }
+        result.pushKV("version", batch[ID_NETWORKINFO]["result"]["version"]);
+        result.pushKV("protocolversion", batch[ID_NETWORKINFO]["result"]["protocolversion"]);
+        if (!batch[ID_WALLETINFO].isNull()) {
+            result.pushKV("walletversion", batch[ID_WALLETINFO]["result"]["walletversion"]);
+            result.pushKV("balance", batch[ID_WALLETINFO]["result"]["balance"]);
+        }
+        result.pushKV("blocks", batch[ID_BLOCKCHAININFO]["result"]["blocks"]);
+        result.pushKV("timeoffset", batch[ID_NETWORKINFO]["result"]["timeoffset"]);
+        result.pushKV("connections", batch[ID_NETWORKINFO]["result"]["connections"]);
+        result.pushKV("proxy", batch[ID_NETWORKINFO]["result"]["networks"][0]["proxy"]);
+        result.pushKV("difficulty", batch[ID_BLOCKCHAININFO]["result"]["difficulty"]);
+        result.pushKV("testnet", UniValue(batch[ID_BLOCKCHAININFO]["result"]["chain"].get_str() == "test"));
+        if (!batch[ID_WALLETINFO].isNull()) {
+            result.pushKV("walletversion", batch[ID_WALLETINFO]["result"]["walletversion"]);
+            result.pushKV("balance", batch[ID_WALLETINFO]["result"]["balance"]);
+            result.pushKV("keypoololdest", batch[ID_WALLETINFO]["result"]["keypoololdest"]);
+            result.pushKV("keypoolsize", batch[ID_WALLETINFO]["result"]["keypoolsize"]);
+            if (!batch[ID_WALLETINFO]["result"]["unlocked_until"].isNull()) {
+                result.pushKV("unlocked_until", batch[ID_WALLETINFO]["result"]["unlocked_until"]);
+            }
+            result.pushKV("paytxfee", batch[ID_WALLETINFO]["result"]["paytxfee"]);
+        }
+        result.pushKV("relayfee", batch[ID_NETWORKINFO]["result"]["relayfee"]);
+        result.pushKV("warnings", batch[ID_NETWORKINFO]["result"]["warnings"]);
+        return JSONRPCReplyObj(result, NullUniValue, 1);
+    }
+};
+
+/** Process default single requests */
+class DefaultRequestHandler: public BaseRequestHandler {
+public:
+    UniValue PrepareRequest(const std::string& method, const std::vector<std::string>& args) override
+    {
+        UniValue params;
+        if(gArgs.GetBoolArg("-named", DEFAULT_NAMED)) {
+            params = RPCConvertNamedValues(method, args);
+        } else {
+            params = RPCConvertValues(method, args);
+        }
+        return JSONRPCRequestObj(method, params, 1);
+    }
+
+    UniValue ProcessReply(const UniValue &reply) override
+    {
+        return reply.get_obj();
+    }
+};
+
+static UniValue CallRPC(BaseRequestHandler *rh, const std::string& strMethod, const std::vector<std::string>& args)
 {
     std::string host;
     // In preference order, we choose the following for the port:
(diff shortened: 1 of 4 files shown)


jonasschnelli:
Concept ACK.
Can you add a tool test for that?

laanwj:
> Can you add a tool test for that?

A simple tool-test like -tx's wont work here as it requires a running bitcoind.
Better to wait for bitcoin-cli testing infrastructure to be in place I guess (#10798)?

achow101:
rebased

jnewbery:
bad rebase:

```
bitcoin-cli.cpp: In function ‘int CommandLineRPC(int, char**)’:
bitcoin-cli.cpp:393:41: error: ‘GetBoolArg’ was not declared in this scope
         if (GetBoolArg("-getinfo", false)) {
                                         ^
```

#10798 is now rebased and ready for merge. It should be fairly straightforward to rebase this on top of that to add a functional test.

jnewbery:
Test here if you want it:  https://github.com/jnewbery/bitcoin/tree/pr10871_test

Builds on top of #10798. Test currently fails because the output for `bitcoin-cli getinfo` and `bitcoin-cli -getinfo` is different in `keypoolsize` and `unlocked_until`. See https://github.com/bitcoin/bitcoin/pull/8843#discussion_r127599074 for information about `keypoolsize` discrepancy.

TheBlueMatt:
I think the docs should at least mention something about atomicity. Every other API to Bitcoin Core is atomic, so it feels very strange to add a new one which is not.

laanwj:
> I think the docs should at least mention something about atomicity. Every other API to Bitcoin Core is atomic, so it feels very strange to add a new one which is not.

I know you don't like this particular changej, but this is reaching FUD levels here, what *exactly* are you concerned about happening here? `-getinfo` pulls the following information:
```
getnetworkinfo:
"version"
"protocolversion"
"timeoffset"
"connections"
"proxy"
"relayfee"
"errors"

getblockchaininfo:
"blocks"
"difficulty"
"testnet"

getwalletinfo:
"walletversion"
"balance"
"keypoololdest"
"keypoolsize"
"unlocked_until"
"paytxfee"
```
Within a group, the values are as atomic as ever. Some of these don't change at runtime at all, some hardly change.

Can you indicate exactly between which two you're afraid non-atomicity is going to give a problem, in practice?

If not, let's please leave this behind us.

TheBlueMatt:
My only real atomicity concern is between blocks and balance - ie that you may call getinfo, see a balance and assume that it is current as of a given block height. It may be of lesser concern in just bitcoin-cli, as its not intended to be used in scripts, hence my note that I'd be ok with just documenting this behavior, though I could still see someone fucking it up in some way.

laanwj:
> My only real atomicity concern is between blocks and balance - ie that you may call getinfo, see a balance and assume that it is current as of a given block height.

Yes, that's a valid concern. Also outside this PR - there is currently no way to query up to which height the wallet has processed. Especially when the wallet and node are decoupled further w/ threads or even processes. We should add a `curheight` or such to `getwalletinfo`.

jnewbery:
> there is currently no way to query up to which height the wallet has processed

I believe that `getinfo` itself could be misleading here. `balance` could be for a different height than `blocks`.

There's been discussion of adding the wallet's best block to `getwalletinfo` on IRC: https://botbot.me/freenode/bitcoin-core-dev/2017-08-11/?msg=89723092&page=2 and I have a branch to do that, which I'm going to PR after #10286

jnewbery:
Is this still being actively worked on? Next steps would be:

- take the functional test in https://github.com/jnewbery/bitcoin/tree/pr10871_test
- fix the discrepencies between `getinfo` and `-getinfo`
- rebase

achow101:
I am still working on this, just got sidetracked with other stuff.

achow101:
Took the test and rebased.

> fix the discrepencies between `getinfo` and `-getinfo`

What discrepancies?

jnewbery:
> What discrepancies?

See above:

> Test currently fails because the output for bitcoin-cli getinfo and bitcoin-cli -getinfo is different in keypoolsize and unlocked_until. See #8843 (comment) for information about keypoolsize discrepancy.

Once #10838 is merged, then that's no longer an issue.

After #10838, I think the way to test this would be to collect the responses from `getwalletinfo`, `getblockchaininfo` and `getnetworkinfo`, merge those dictionaries (hint: https://stackoverflow.com/a/26853961) and then compare that to the response from `bitcoin-cli -getinfo`.

achow101:
> After #10838, I think the way to test this would be to collect the responses from `getwalletinfo`, `getblockchaininfo` and `getnetworkinfo`, merge those dictionaries (hint: https://stackoverflow.com/a/26853961) and then compare that to the response from `bitcoin-cli -getinfo`.

I think that we should be doing that instead of comparing against `getinfo`. This is explicitly a new thing which is similar to `getinfo` but not exactly replicating it.

achow101:
I modified the test to compare against the `get*info` RPCs instead of `getinfo`

src/rpc/protocol.cpp
```diff
@@ -135,3 +135,20 @@ void DeleteAuthCookie()
...
 
+std::vector<UniValue> JSONRPCProcessBatchReply(const UniValue &in, size_t num)
+{
+    if (!in.isArray()) {
+        throw std::runtime_error("Batch must be an array");
+    }
+    std::vector<UniValue> batch(num);
+    for (size_t i=0; i<in.size(); ++i) {
+        const UniValue &rec = in[i];
+        if (!rec.isObject())
```
jnewbery: please also add braces to these if blocks

jnewbery (commented):
one nit inline.

Please tidy up commits. First two commits should be squashed, and final two commits should be squashed into a second commit.

achow101:
Tidied up commits

src/rpc/protocol.cpp
achow101: Done

src/bitcoin-cli.cpp
```diff
@@ -191,7 +192,95 @@ static void http_error_cb(enum evhttp_request_error err, void *ctx)
...
+        return JSONRPCRequestObj(strMethod, params, 1);
+    }
+
+    UniValue ProcessReply(const UniValue &reply) override
+    {
+        return reply.get_obj();
+    }
+};
+
+UniValue CallRPC(BaseRequestHandler *rh, const std::string& strMethod, const std::vector<std::string>& args)
```
jnewbery: You've lost the static keyword for this function in your rebase.

src/bitcoin-cli.cpp
```diff
@@ -307,23 +396,24 @@ int CommandLineRPC(int argc, char *argv[])
             while (std::getline(std::cin,line))
                 args.push_back(line);
         }
+        std::unique_ptr<BaseRequestHandler> rh;
+        if (gArgs.GetBoolArg("-getinfo", false)) {
+            rh.reset(new GetinfoRequestHandler());
+            args.clear();
+            args.push_back("getinfo");
```
jnewbery: I think it would be clearer to not reset `args` here. The only reason you're doing this is to pass the `if (args.size() < 1)` check below and then extract `strMethod`.

Instead, just move the `if (args.size() < 1)` check into the else branch. `GetinfoRequestHandler.PrepareRequest()` doesn't require `strMethod` or `args`.

Taking this further, you could add a method to `DefaultRequestHandler` to update the `strMethod` and `args`, and then update the `PrepareRequest()` method to not take any arguments.

src/bitcoin-cli.cpp
```diff
@@ -191,7 +192,95 @@ static void http_error_cb(enum evhttp_request_error err, void *ctx)
...
+        result.pushKV("connections", batch[ID_NETWORKINFO]["result"]["connections"]);
+        result.pushKV("proxy", batch[ID_NETWORKINFO]["result"]["networks"][0]["proxy"]);
+        result.pushKV("difficulty", batch[ID_BLOCKCHAININFO]["result"]["difficulty"]);
+        result.pushKV("testnet", UniValue(batch[ID_BLOCKCHAININFO]["result"]["chain"].get_str() == "test"));
+        if (!batch[ID_WALLETINFO].isNull()) {
+            result.pushKV("walletversion", batch[ID_WALLETINFO]["result"]["walletversion"]);
+            result.pushKV("balance", batch[ID_WALLETINFO]["result"]["balance"]);
+            result.pushKV("keypoololdest", batch[ID_WALLETINFO]["result"]["keypoololdest"]);
+            result.pushKV("keypoolsize", batch[ID_WALLETINFO]["result"]["keypoolsize"]);
+            if (!batch[ID_WALLETINFO]["result"]["unlocked_until"].isNull())
```
jnewbery: nit: braces

src/bitcoin-cli.cpp
```diff
@@ -191,7 +192,95 @@ static void http_error_cb(enum evhttp_request_error err, void *ctx)
...
+/** Process getinfo requests */
+class GetinfoRequestHandler: public BaseRequestHandler
+{
+public:
+    const int ID_NETWORKINFO = 0;
+    const int ID_BLOCKCHAININFO = 1;
+    const int ID_WALLETINFO = 2;
+
+    /** Create a simulated `getinfo` request. */
+    UniValue PrepareRequest(const std::string& strMethod, const std::vector<std::string>& args) override
```
jnewbery: nit: function arguments should not be hungarian/camel case `strMethod` -> `method`

jnewbery (commented):
Some comments in line.

You seem to have lost the test in your commit squashing.

achow101:
Fixed the test failure.

src/bitcoin-cli.cpp
```diff
@@ -191,7 +192,96 @@ static void http_error_cb(enum evhttp_request_error err, void *ctx)
...
 #endif
 
-static UniValue CallRPC(const std::string& strMethod, const UniValue& params)
+/** Class that handles the conversion from a command-line to a JSON-RPC request,
+ * as well as converting back to a JSON object that can be shown as result.
+ */
+class BaseRequestHandler
+{
+public:
+    virtual UniValue PrepareRequest(const std::string& strMethod, const std::vector<std::string>& args) = 0;
```
jnewbery: nit `strMethod` -> `method`

src/bitcoin-cli.cpp
jnewbery: > nit: function arguments should not be hungarian/camel case strMethod -> method

not currently addressed

test/functional/bitcoin_cli.py
```diff
@@ -35,5 +35,30 @@ def run_test(self):
...
+        assert_equal(cli_get_info['proxy'], network_info['networks'][0]['proxy'])
+        assert_equal(cli_get_info['difficulty'], blockchain_info['difficulty'])
+        assert_equal(cli_get_info['testnet'], blockchain_info['chain'] == "test")
+        assert_equal(cli_get_info['balance'], wallet_info['balance'])
+        assert_equal(cli_get_info['keypoololdest'], wallet_info['keypoololdest'])
+        assert_equal(cli_get_info['keypoolsize'], wallet_info['keypoolsize'])
+        assert_equal(cli_get_info['paytxfee'], wallet_info['paytxfee'])
+        assert_equal(cli_get_info['relayfee'], network_info['relayfee'])
+
+        if 'unlocked_until' in wallet_info:
```
jnewbery: We know that the wallet isn't locked, so this won't be tested. I don't think there's any benefit in adding a test code that won't be executed (in fact it's slightly deleterious since a casual reader might think that we're actually testing this).

Either:
- lock the wallet and actually test this
- add a comment saying we're not testing the field because the wallet isn't locked.

I think either is fine.

jnewbery (commented):
Looks really good. Just a few nitty comments inline

src/bitcoin-cli.cpp
achow101: Done

src/bitcoin-cli.cpp
achow101: Done

test/functional/bitcoin_cli.py
achow101: Done

achow101:
Addressed nits (sorry for the delay).

jnewbery:
Tested ACK 5ca97e5f03d7352558991e8eb26bb58dc4eec453.

Thanks for sticking with this!

src/bitcoin-cli.cpp
```diff
@@ -37,6 +37,7 @@ std::string HelpMessageCli()
     strUsage += HelpMessageOpt("-?", _("This help message"));
     strUsage += HelpMessageOpt("-conf=<file>", strprintf(_("Specify configuration file (default: %s)"), BITCOIN_CONF_FILENAME));
     strUsage += HelpMessageOpt("-datadir=<dir>", _("Specify data directory"));
+    strUsage += HelpMessageOpt("-getinfo", _("Get general information from the remote server"));
```
TheBlueMatt: Are you going to update the help text to note something about balance/height atomicity? The old getinfo takes cs_main and cs_wallet and returns atomic information, this no longer does. getwalletinfo will likely eventually return something so that you can report if a result was atomic here (probably a follow-up to #10286).

achow101: What do you suggest that the text for such a note be?

TheBlueMatt:
Something like "Note that unlike server-side RPC calls, the results of getinfo is the result of multiple non-atomic requests. Some entries in the result may represent results from different states (eg wallet balance may be as of a different block from the chain state reported)"

achow101:
Added the atomicity note to the help text.

src/bitcoin-cli.cpp
laanwj: as getinfo is misleading here too, see https://github.com/bitcoin/bitcoin/pull/10871#issuecomment-324699582, I still don't see this as a strong point, but as it seems to bother @TheBlueMatt very much let's just add it...

laanwj:
utACK 5e69a43

src/bitcoin-cli.cpp
TheBlueMatt: @laanwj I do not believe that to be the case in any release nor on master. It will be the case after #10286, however there is no getinfo on master either....

musabi:
Thanks, helped.

crombiecrunch:
can someone please tell me how to apply this so that getinfo can work with the yiimp opensource pool files?

MarcoFalke:
@crombiecrunch Please submit a pull request upstream, replacing getinfo['balance'] with getwalletinfo['balance'] in the open source code.

PR #10872, opened 2017-07-18
Docs: Syntax highlight shell commands in the building notes

eklitzke:
This is a trivial change that adds syntax highlighting to all of the shell commands that use strings/variables/comments from the build docs (i.e. all of the shell commands that actually can be highlighted). I also made one minor change to one of the commands, encouraging the use of `sudo -i` instead of `sudo su` (the two are functionally equivalent).

### DIFF
--- a/doc/build-openbsd.md
+++ b/doc/build-openbsd.md
@@ -41,7 +41,7 @@ Do not use `pkg_add boost`! The boost version installed thus is compiled using t
 
 This makes it necessary to build boost, or at least the parts used by Bitcoin Core, manually:
 
-```
+```bash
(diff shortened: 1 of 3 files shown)


practicalswift:
Concept ACK

jonasschnelli:
Not sure about this.
One great thing Markup text files offer, is, that they are perfectly fine readable with no Markup render engine (in a standard text editor). Adding "tags" defeats that purpose IMO and may confuse pure text-editor/reader users.

practicalswift:
Good point @jonasschnelli. When considering also the non-rendered use-case I'll have to withdraw my previous "concept ACK" :-)

laanwj:
Seems there is no agreement to do this, closing.

PR #10873, opened 2017-07-18
Use a condition variable for shutdown notifications

eklitzke:
Currently bitcoind polls every 200ms to see if a shutdown has been requested. This causes the process to wake up frequently and unnecessarily when the network is idle. The change here adds a condition variable so that the main bitcoind thread can do a blocking wait on the condition variable, rather than polling. Without this patch (i.e. in master) if you strace bitcoind you will see a lot of nanosleep() calls, with this patch the process does just a blocking wait on a futex(). The motivation here is that I want the CPU to be able to stay powered down more on my local bitcoin node.

I elected to keep `fRequestShutdown` as an atomic variable, even though a new mutex is added for the condition variable, just to minimize code churn. I think there's a case to be made that it should be turned into a regular bool and then `ShutdownRequested()` should use the mutex, but I don't have a strong preference. To make it non-atomic and use the mutex would make the diff a little bigger, as there are a few places that read the atomic variable directly.

This is my first Bitcoin diff. I have read the contributions doc, but please let me know if I made any errors related to code style, contributing, etc.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -116,16 +116,38 @@ static const char* FEE_ESTIMATES_FILENAME="fee_estimates.dat";
 // shutdown thing.
 //
 
-std::atomic<bool> fRequestShutdown(false);
-std::atomic<bool> fDumpMempoolLater(false);
+// g_request_shutdown is true when bitcoin should be shutting down, and access
+// is guarded by g_shutdown_mut.
+static std::mutex g_shutdown_mut;
+static std::condition_variable g_shutdown_cond;
+static bool g_request_shutdown{false};
 
-void StartShutdown()
+static std::atomic<bool> g_dump_mempool_later(false);
+
+static void SetShutdown(bool value)
 {
-    fRequestShutdown = true;
+    {
+        std::lock_guard<std::mutex> lk(g_shutdown_mut);
+        g_request_shutdown = value;
+    }
+    g_shutdown_cond.notify_all();
 }
+
+void StartShutdown() {
+    SetShutdown(true);
+}
+
+// Threads can use this method to wait for a shutdown request, without polling.
+void WaitForShutdownNotify()
+{
+    std::unique_lock<std::mutex> lk(g_shutdown_mut);
+    g_shutdown_cond.wait(lk, []{ return g_request_shutdown; });
+}
+
 bool ShutdownRequested()
 {
-    return fRequestShutdown;
+    std::lock_guard<std::mutex> lk(g_shutdown_mut);
+    return g_request_shutdown;
 }
 
 /**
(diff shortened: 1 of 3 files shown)


sipa:
@eklitzke Thanks for your contribution! I think your patch is pretty much acceptable as-is. Since the (new) `fRequestShutdown` is static and very localized, I would argue to convert it to a non-atomic in one go.

As you're pretty much rewriting all uses of these variables anyway, I think you could adapt them to the new coding style guidelines too (e.g. `g_cs_shutdown`, `g_shutdown_condition`, `g_request_shutdown`). That's just a suggestion, not a requirement.

Concept ACK

practicalswift:
Concept ACK

Very nice first-time contribution! Thanks!

jonasschnelli:
Concept ACK.

src/init.cpp
```diff
@@ -283,7 +305,7 @@ void Shutdown()
  */
 static void HandleSIGTERM(int)
 {
-    fRequestShutdown = true;
+    StartShutdown();
```
laanwj: Unfortunately, a signal handler is not allowed to do anything besides set a flag.
This is the only reason why we still have the polling.

src/init.cpp
```diff
@@ -116,16 +116,38 @@ static const char* FEE_ESTIMATES_FILENAME="fee_estimates.dat";
...
 {
-    fRequestShutdown = true;
+    {
+        std::lock_guard<std::mutex> lk(g_shutdown_mut);
+        g_request_shutdown = value;
+    }
+    g_shutdown_cond.notify_all();
+}
+
+void StartShutdown() {
```
promag: Nit, new line for `{`.

src/init.cpp
```diff
@@ -116,16 +116,38 @@ static const char* FEE_ESTIMATES_FILENAME="fee_estimates.dat";
...
 
-void StartShutdown()
+static std::atomic<bool> g_dump_mempool_later(false);
+
+static void SetShutdown(bool value)
 {
-    fRequestShutdown = true;
+    {
+        std::lock_guard<std::mutex> lk(g_shutdown_mut);
+        g_request_shutdown = value;
```
promag: Add something like
```cpp
if (g_request_shutdown == value) {
  // log something
  return;
}
```

promag (commented):
Concept ACK.

laanwj:
Needs update for code review comments

MarcoFalke:
This is up for grabs, it seems

laanwj:
I'm closing it for now, then. The UNIX signals + pthread issue is not trivial to resolve. Instead of a condition variable to synchronize you could use a file descriptor pair in a specific way that is allowed in a signal handler on any platform. I'm not up to date on the specifics, though.

Let me know if you want to work on this again @eklitzke then we'll reopen.

PR #10874, opened 2017-07-19
[RPC] getblockchaininfo: Loop through the bip9 soft fork deployments instead of hard coding

achow101:
Instead of hard coding which deployment statistics should be listed in the `getblockchaininfo` output, loop through the available deployments (except testdummy) when displaying their deployment info.

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -1107,13 +1107,13 @@ static UniValue BIP9SoftForkDesc(const Consensus::Params& consensusParams, Conse
     return rv;
 }
 
-void BIP9SoftForkDescPushBack(UniValue& bip9_softforks, const std::string &name, const Consensus::Params& consensusParams, Consensus::DeploymentPos id)
+void BIP9SoftForkDescPushBack(UniValue& bip9_softforks, const Consensus::Params& consensusParams, Consensus::DeploymentPos id)
 {
     // Deployments with timeout value of 0 are hidden.
     // A timeout value of 0 guarantees a softfork will never be activated.
     // This is used when softfork codes are merged without specifying the deployment schedule.
     if (consensusParams.vDeployments[id].nTimeout > 0)
-        bip9_softforks.push_back(Pair(name, BIP9SoftForkDesc(consensusParams, id)));
+        bip9_softforks.push_back(Pair(VersionBitsDeploymentInfo[id].name, BIP9SoftForkDesc(consensusParams, id)));
 }
 
 UniValue getblockchaininfo(const JSONRPCRequest& request)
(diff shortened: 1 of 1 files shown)


src/rpc/blockchain.cpp
```diff
@@ -1185,8 +1185,11 @@ UniValue getblockchaininfo(const JSONRPCRequest& request)
     softforks.push_back(SoftForkDesc("bip34", 2, tip, consensusParams));
     softforks.push_back(SoftForkDesc("bip66", 3, tip, consensusParams));
     softforks.push_back(SoftForkDesc("bip65", 4, tip, consensusParams));
-    BIP9SoftForkDescPushBack(bip9_softforks, "csv", consensusParams, Consensus::DEPLOYMENT_CSV);
-    BIP9SoftForkDescPushBack(bip9_softforks, "segwit", consensusParams, Consensus::DEPLOYMENT_SEGWIT);
+    for (int BIP9SoftFork = Consensus::DEPLOYMENT_CSV; BIP9SoftFork != Consensus::MAX_VERSION_BITS_DEPLOYMENTS; ++BIP9SoftFork) {
```
sipa: Why convert to int and back to `DeploymentPos`?

jonasschnelli:
utACK 4048d77fb462268f374edd420a827d91be8af11f

jnewbery:
I'm not sure this change is worth it when there are just two bip9 softforks. This is a net +ve LOC change for the same functionality.

Perhaps worth revisiting if we get more bip9 softforks in future, or am I missing some benefit?

achow101:
It's fine to do this in the future when we have more soft forks. I had only made the change since I needed to add stats about BIP 91 to one of my branches and the fact that the forks needed to be hard coded in to be displayed with `getblockchaininfo` bothered me.

promag:
ACK 4048d77.

Nit, make it so that the name is accessed in `BIP9SoftForkDescPushBack()`, so the loop would be:
```cpp
for (int pos = Consensus::DEPLOYMENT_CSV; pos != Consensus::MAX_VERSION_BITS_DEPLOYMENTS; ++pos) {
    BIP9SoftForkDescPushBack(bip9_softforks, consensusParams, static_cast<Consensus::DeploymentPos>(pos));
}
```

laanwj:
> Nit, make it so that the name is accessed in BIP9SoftForkDescPushBack(), so the loop would be:

I also prefer that to casting to int and back in the loop header.

laanwj:
Ping @achow101

achow101:
pong @laanwj

laanwj:
@achow101 can you respond to @promag's nit? If you don't want to do it that's fine, but if you don't respond to review comments your PR will be in limbo.

achow101:
Done.

src/rpc/blockchain.cpp
```diff
@@ -1185,8 +1185,10 @@ UniValue getblockchaininfo(const JSONRPCRequest& request)
     softforks.push_back(SoftForkDesc("bip34", 2, tip, consensusParams));
     softforks.push_back(SoftForkDesc("bip66", 3, tip, consensusParams));
     softforks.push_back(SoftForkDesc("bip65", 4, tip, consensusParams));
-    BIP9SoftForkDescPushBack(bip9_softforks, "csv", consensusParams, Consensus::DEPLOYMENT_CSV);
-    BIP9SoftForkDescPushBack(bip9_softforks, "segwit", consensusParams, Consensus::DEPLOYMENT_SEGWIT);
+    for (int pos = Consensus::DEPLOYMENT_CSV; pos != Consensus::MAX_VERSION_BITS_DEPLOYMENTS; ++pos) {
+        BIP9SoftForkDescPushBack(bip9_softforks, consensusParams, static_cast<Consensus::DeploymentPos>(pos));
+
```
laanwj: please don't add an empty line here

promag:
Thanks, much better.

utACK 8cf8b06.

src/rpc/blockchain.cpp
achow101: Oops, removed.

laanwj:
utACK e4d0af4

PR #10875, opened 2017-07-19
BIP 91 deployment parameters

achow101:
This PR contains the deployment parameters for BIP 91. This PR is intended to be a place where people can easily get the BIP 91 deployment parameters for things like shutting up the unknown block version warning and monitoring the deployment information with `getblockchaininfo`.

I do not expect this to be released as by the time a release is made, BIP 91 should be long gone.

### DIFF
--- a/src/consensus/params.h
+++ b/src/consensus/params.h
@@ -17,6 +17,7 @@ enum DeploymentPos
     DEPLOYMENT_TESTDUMMY,
     DEPLOYMENT_CSV, // Deployment of BIP68, BIP112, and BIP113.
     DEPLOYMENT_SEGWIT, // Deployment of BIP141, BIP143, and BIP147.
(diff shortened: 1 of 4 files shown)


src/consensus/params.h
```diff
@@ -17,6 +17,7 @@ enum DeploymentPos
     DEPLOYMENT_TESTDUMMY,
     DEPLOYMENT_CSV, // Deployment of BIP68, BIP112, and BIP113.
     DEPLOYMENT_SEGWIT, // Deployment of BIP141, BIP143, and BIP147.
+    DEPLOYMENT_SEGWIT2X, // Deployment of BIP 91
```
jameshilliard: DEPLOYMENT_SEGSIGNAL is the parameter used for BIP91 by itself

PR #10876, opened 2017-07-19
What do you mean a bug in the latest version?

Z-IDDQD:
Indexing doesn't help - after about a few minutes appears again.
v. 0.14.2 (x64)


sipa:
See https://bitcoin.stackexchange.com/questions/56795/warning-unknown-block-versions-being-mined-its-possible-unknown-rules-are-in/56796#56796.

Also, why is this a pull request?

PR #10877, opened 2017-07-19
[rpc] Verbose flags for chaining and scripting

kallewoof:
[Reviewer hint: use [?w=1](https://github.com/bitcoin/bitcoin/pull/10877/files?w=1) to avoid seeing a bunch of indentation changes.]

Right now, some RPC commands return JSON data along with transaction hex strings, which means scripting and/or chaining needs to do JSON processing before being able to use the resulting transaction data. For use cases where you don't need the extra info, this PR adds a verbose=false mode (default true) which returns only the transaction hex string.

This means you can now do things like this:
```Bash
./bitcoin-cli -regtest -named signrawtransactionwithwallet hexstring=$(
    ./bitcoin-cli -regtest -named fundrawtransaction hexstring=$(
        ./bitcoin-cli -regtest createrawtransaction "[]" "{\"$(
            ./bitcoin-cli -regtest getnewaddress)\": 10.0}") verbose=false) verbose=false
```

This is also useful in scripts, where you perform some operation on a transaction and store the tx itself in a variable. Before you would need e.g. `jq` or some other tool to do this.

Edit: `signrawtransaction` has been deprecated; I chose to add verbose to the new commands `withkey`/`withwallet` but not to `signrawtransaction`.

### DIFF
--- a/src/rpc/client.cpp
+++ b/src/rpc/client.cpp
@@ -96,6 +96,7 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "signrawtransaction", 2, "privkeys" },
     { "signrawtransactionwithkey", 1, "privkeys" },
     { "signrawtransactionwithkey", 2, "prevtxs" },
+    { "signrawtransactionwithkey", 4, "verbose" },
     { "signrawtransactionwithwallet", 1, "prevtxs" },
     { "signrawtransactionwithwallet", 3, "verbose" },
     { "sendrawtransaction", 1, "allowhighfees" },
--- a/src/rpc/client.cpp
+++ b/src/rpc/client.cpp
@@ -97,6 +97,7 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "signrawtransactionwithkey", 1, "privkeys" },
     { "signrawtransactionwithkey", 2, "prevtxs" },
     { "signrawtransactionwithwallet", 1, "prevtxs" },
+    { "signrawtransactionwithwallet", 3, "verbose" },
     { "sendrawtransaction", 1, "allowhighfees" },
     { "combinerawtransaction", 0, "txs" },
     { "fundrawtransaction", 1, "options" },
--- a/src/rpc/client.cpp
+++ b/src/rpc/client.cpp
@@ -101,6 +101,7 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "combinerawtransaction", 0, "txs" },
     { "fundrawtransaction", 1, "options" },
     { "fundrawtransaction", 2, "iswitness" },
+    { "fundrawtransaction", 3, "verbose" },
     { "gettxout", 1, "n" },
     { "gettxout", 2, "include_mempool" },
     { "gettxoutproof", 0, "txids" },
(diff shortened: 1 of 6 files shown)


jonasschnelli:
Code wise, this is more of a refactor (much more lines are touch for refactoring then the intended feature).

Wouldn't' the non verbose mode (== directly return hex) include some risks because one my miss important errors (sent back through the "error" json array/object)?

kallewoof:
@jonasschnelli For `fundrawtransaction`, you will no longer see the resulting fee, which is a pity, but if you don't need to know it, it's fine. For `signrawtransaction`, you will not see if it was only partially signed, which may trip you in certain cases, but I don't think it would end up in lost money or be dangerous (you just have a transaction that the network will reject).

jonasschnelli:
@kallewoof your right about `fundrawtransaction` (I mixed it up with `bumpfee`). But what about `signrawtransaction` Errors like `TxInErrorToJSON(txin, vErrors, "Input not found or already spent");` or the possible script sig errors?

I think its probably acceptable to suppress those errors, but maybe we should mention that in the help description.

kallewoof:
I'm actually confused why the sign code doesn't throw an error in the case you mentioned. Why would it continue when the tx appears to be broken? Hm..

promag:
Does this sounds more like a bash mode for cli than verbose option?

promag:
Another suggestion to ease chaining and eventually reduce the transmitted data is to specify a [json path](http://jsonpath.com/) to filter the output. For instance:
```
bitcoin-cli -rpcfilter '*.amount' listunspent
```
Compared to jq, it has the advantage of server side filtering. But I guess it is too much for core.

kallewoof:
> Does this sounds more like a bash mode for cli than verbose option?

I chose verbose because if you look at e.g. `createrawtransaction` it responds with the hex only (the same style), whereas the ones here will respond with a JSON object with extra information. That to me was actually a bit unexpected, and it still trips me up on occasion when I'm trying to do something (I somehow think the results will be the tx hex when they're a json object).

promag:
IMO verbose is unclear and usually means more detailed logs which is unrelated.

Also, I think we should discourage chaining these calls because the middle results/errors are important and should be acknowledge/logged/handled/whatever.

As you said, it's still doable using other tools.

Not sure about others, but for me it's a NACK.

kallewoof:
Verbose is not restricted to logs, in my world. The world itself simply means "using or expressed in more words than are needed" which I think fits very nicely with the use case here (you don't need the complete flag for signing, for instance; it's just a helpful thing to have).

Yes, it's doable using other tools, but this dependency is rather superfluous considering the relatively small addition (style fixes outweighs the code itself) means *less* work for the node (a string rather than a JSON object).

It's ultimately ~2-3 lines per RPC call to remove a potential dependency on external tools. I think that's worth something personally.

jnewbery:
Concept ACK. I think chaining commands together without intermediate error checking like this is definitely dangerous and poor practice, but there are probably legitimate use-cases for testing, etc.

Thinking out loud: verbosity is a very common option for RPCs. Perhaps we should consider making it a query string in the URL rather than adding parameters to every RPC that requires it. Maybe that's not practical, but worth considering.

kallewoof:
Rebased, and dropped `signrawtransaction` patch and added `signrawtransactionwith*` support instead.

kallewoof:
One NACK, one concept ACK. I think it's not worth keeping this around.

PR #10878, opened 2017-07-19
Docs: Fix Markdown formatting issues in init.md

dongcarl:
Trivial commit that fixes Markdown line breaks in `docs/init.md`. Markdown line breaks take the form of two spaces, which is hard to spot when viewing raw text but visible when previewing on GitHub. Line 72-73 of `docs/init.md` did not conform to the rest the rest of the documentation, and is corrected in this PR.

### DIFF
--- a/doc/init.md
+++ b/doc/init.md
@@ -10,14 +10,14 @@ can be found in the contrib/init folder.
     contrib/init/bitcoind.conf:       Upstart service configuration file
     contrib/init/bitcoind.init:       CentOS compatible SysV style init script
 
-1. Service User
+Service User
 ---------------------------------
 
 All three Linux startup configurations assume the existence of a "bitcoin" user
 and group.  They must be created before attempting to use these scripts.
 The OS X configuration assumes bitcoind will be set up for the current user.
 
-2. Configuration
+Configuration
 ---------------------------------
 
 At a bare minimum, bitcoind requires that the rpcpassword setting be set
@@ -46,10 +46,10 @@ relative to the data directory. `wallet` *only* supports relative paths.
 For an example configuration file that describes the configuration settings,
 see `contrib/debian/examples/bitcoin.conf`.
 
-3. Paths
+Paths
 ---------------------------------
 
-3a) Linux
+### Linux
 
 All three configurations assume several paths that might need to be adjusted.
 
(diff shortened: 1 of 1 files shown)


practicalswift:
ACK 72a184a780f29d7833d900d728c0c3a10f460969

jonasschnelli:
ACK 72a184a780f29d7833d900d728c0c3a10f460969

promag (approved):
ACK 72a184a.

laanwj:
This whole file looks strange in the github renderer: https://github.com/dongcarl/bitcoin/blob/72a184a780f29d7833d900d728c0c3a10f460969/doc/init.md

For example, it doesn't quite catch the section headers.

dongcarl:
@laanwj Right. There's the option to use "##", but then it wouldn't look as nice in plain text. I'm sure it's a glitch in the renderer.

Another question: I want to expand on Section 3 (Paths), and give example cli commands in different UNIX distros to do things described in the section such as:

1. Create bitcoin user and group (with optional flag to disable home directory creation)
2. Make PID and data directory owned by bitcoin user and group
3. Make configuration file and data directory only readable by bitcoin user and group

laanwj:
> Right. There's the option to use "##", but then it wouldn't look as nice in plain text. I'm sure it's a glitch in the renderer.

`---------------` works ok for defining titles in other files, the reason might be that the titles are prefixed with a number and a dot.

flack:
@laanwj `-------` works for headlines, but it's not anything mentioned in github's documentation, more like cmpatibility to the old Daring Fireball specification. Right now, the headlines with numbers get rendered as `<ol>` tags, so my guess would be that the parser goes through the document line by line and when it matches the list syntax, it doesn't look for headlines any more, but instead treats the `----` as a new element which gets rendered as `<hr>`. There are two ways around it:

1. Use `#`:

```
#1. Service User
``` 

2. Escape the `.` character:


```
1\. Service User
----------------
``` 

Although it's kind of hard to decide which one is less ugly :-)

Alternatively, why not write 

```
1: Service User
----------------
``` 

instead?

laanwj:
In none of the other documents we have section numbering, so it would also be ok to remove the numbers completely, at least assuming they're not used to reference anything.

dongcarl:
@laanwj https://github.com/bitcoin/bitcoin/pull/10878/commits/d201e40c55b0c02e3bec715ded06ec40f5c06fc4

dongcarl:
Travis failed on something unrelated to the docs... My changes shouldn't have broken anything.

laanwj:
ACK d201e40

> My changes shouldn't have broken anything.

Yup. Welcome in the world of automated testing :/

PR #10880, opened 2017-07-19
Update .gitignore to ignore more test files

eklitzke:
Without this change, after building bitcoin master and testing it, I see the following untracked files:

```
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	qa/
	src/test/buildenv.py
	src/wallet.backup
```

I've tried to keep the `.gitignore` lines in the appropriate sections of the file.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -49,6 +49,7 @@ src/qt/test/moc*.cpp
 .libs
 .*.swp
 *.*~*
+*.backup
 *.bak
 *.rej
 *.orig
@@ -97,9 +98,11 @@ total.coverage/
 coverage_percent.txt
 
 #build tests
+qa/
 linux-coverage-build
 linux-build
 win32-build
+src/test/buildenv.py
 test/config.ini
 test/cache/*
 


practicalswift:
Concept ACK

.gitignore
```diff
@@ -97,9 +98,11 @@ total.coverage/
 coverage_percent.txt
 
 #build tests
+qa/
```
jonasschnelli: `qa/` should no longer be created in current master. If this is still the case, we should rather find the source why this was created then just ignoring it.

.gitignore
```diff
@@ -49,6 +49,7 @@ src/qt/test/moc*.cpp
 .libs
 .*.swp
 *.*~*
+*.backup
```
jonasschnelli: I guess this would not hurt, though I think we should cleanup the backup file (unlink them) in `wallet_test.cpp`.

promag:
Concept ACK.

.gitignore
laanwj: Not only that, but we shouldn't be creating a `wallet.backup` file in the src directory at all. It should go in the temporary directory created for the other test files.

Sounds like a really bad thing now, someone could confuse it for a real wallet backup, for one.

Thanks for noticing @eklitzke.

laanwj: This line can go after #10927

.gitignore
laanwj: I haven't seen any instances of it being created in new trees, so most probably if you have this, it's a remainder from an old build.

MarcoFalke:
@eklitzke Are you still working on this?

eklitzke:
It looks like I get a clean environment when building on master, including when running the test suite. Previously I had built Bitcoin from a release tag. For this reason, I'm closing the PR.

PR #10882, opened 2017-07-19
[Do not merge] Stop advancing best block and shutdown node if keypool drops below critical threshold

jnewbery:
- Add `keypool_critical` (configurable). If the number of keys in the keypool drops below this limit while the wallet is rescanning, shutdown the node. Do not shutdown the node if the wallet is 'current', ie it is receiving `BlockConnected` calls from the node.
- Add `keypool_min` (configurable). If the number of keys in the keypool drops below this limit, stop advancing the wallet's best block. This forces the wallet to rescan from the point that it dropped below the limit the next time that it starts up. This is a toggle controlled by `m_update_best_block`, which doesn't get unset until the wallet has rescanned with the keypool above keypool_min.
- Add `bypasskeypoolcritical` (command line argument). This disables the keypool_critical behavior, so the user has a chance to top up their keypool.
- don't allow user actions like `getnewaddress` to cause the keypool to drop below the critical limit (return an error telling the user to unlock and topup their keypool).

This is a simpler version of #10830 , which caused the node to stop sync'ing if the keypool dropped below a certain limit. It is built on top of #11022 which does the following:

- if a key in the keypool is used, mark all keys in the keypool up to that key as used
- try to top up the keypool when keys from the keypool are used.

This PR couldn't be merged for v0.15 because there are some edge cases that make this dangerous and could result in users not being able to start up the node without onerous recovery steps.

### DIFF
--- a/src/qt/addresstablemodel.cpp
+++ b/src/qt/addresstablemodel.cpp
@@ -369,17 +369,14 @@ QString AddressTableModel::addRow(const QString &type, const QString &label, con
     {
         // Generate a new address to associate with given label
         CPubKey newKey;
-        if(!wallet->GetKeyFromPool(newKey))
-        {
+        if (!wallet->GetKeyFromPool(newKey, false, true)) {
             WalletModel::UnlockContext ctx(walletModel->requestUnlock());
-            if(!ctx.isValid())
-            {
+            if(!ctx.isValid()) {
                 // Unlock wallet failed or was cancelled
                 editStatus = WALLET_UNLOCK_FAILURE;
                 return QString();
             }
-            if(!wallet->GetKeyFromPool(newKey))
-            {
+            if (!wallet->GetKeyFromPool(newKey, false, true)) {
                 editStatus = KEY_GENERATION_FAILURE;
                 return QString();
             }
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -161,8 +161,8 @@ UniValue getnewaddress(const JSONRPCRequest& request)
 
     // Generate a new key that is added to wallet
     CPubKey newKey;
-    if (!pwallet->GetKeyFromPool(newKey)) {
-        throw JSONRPCError(RPC_WALLET_KEYPOOL_RAN_OUT, "Error: Keypool ran out, please call keypoolrefill first");
+    if (!pwallet->GetKeyFromPool(newKey, false, true)) {
+        throw JSONRPCError(RPC_WALLET_KEYPOOL_RAN_OUT, "Error: Keypool is at critical level or has run out, please call keypoolrefill first");
     }
     CKeyID keyID = newKey.GetID();
 
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -885,8 +885,9 @@ bool CWallet::GetAccountPubkey(CPubKey &pubKey, std::string strAccount, bool bFo
 
     // Generate a new key
     if (bForceNew) {
-        if (!GetKeyFromPool(account.vchPubKey, false))
+        if (!GetKeyFromPool(account.vchPubKey, false, true)) {
             return false;
+        }
 
         SetAddressBook(account.vchPubKey.GetID(), strAccount, "receive");
         walletdb.WriteAccount(strAccount, account);
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -419,8 +419,14 @@ void CWallet::SetBestChain(const CBlockLocator& loc)
 {
     LOCK(cs_wallet); //nWalletMaxVersion
     unsigned int keypool_min = GetArg("-keypoolmin", DEFAULT_KEYPOOL_MIN);
-    if (IsHDEnabled() && !HasUnusedKeys(keypool_min)) {
+    if (IsHDEnabled() && (!m_update_best_block || !HasUnusedKeys(keypool_min))) {
         // If the keypool has dropped below -keypoolmin, then don't update the bestblock height. We can rescan later once the wallet is unlocked.
+
+        if (m_update_best_block) {
+            LogPrintf("Keypool has fallen below keypool_min (%s). Wallet will no longer watch for new transactions and best block height will not be advanced.\n", keypool_min);
+            LogPrintf("Unlock wallet, top up keypool and rescan to resume watching for new transactions.\n");
+            m_update_best_block = false;
+        }
         return;
     }
     CWalletDB walletdb(*dbw);
@@ -1649,6 +1655,23 @@ CBlockIndex* CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, bool f
         ShowProgress(_("Rescanning..."), 100); // hide progress dialog in GUI
 
         fScanningWallet = false;
+    }
+
+    // Update m_update_best_block if we've scanned from before the previous best block and keypool level is still above Keypool_min
+    CWalletDB walletdb(*dbw);
+    CBlockLocator loc;
+    walletdb.ReadBestBlock(loc);
+    CBlockIndex* pindexBestBlock = FindForkInGlobalIndex(chainActive, loc);
+    if (IsHDEnabled() && pindexStart && pindexStart->nHeight <= pindexBestBlock->nHeight) {
+        m_update_best_block = HasUnusedKeys(GetArg("-keypoolmin", DEFAULT_KEYPOOL_MIN));
+    }
+
+    // If the keypool has dropped below -keypoolmin, then stop updating the bestblock height. We can rescan later once the wallet is unlocked.
+    unsigned int keypool_min = GetArg("-keypoolmin", DEFAULT_KEYPOOL_MIN);
+    if (IsHDEnabled() && !HasUnusedKeys(keypool_min) && m_update_best_block) {
+        LogPrintf("Keypool has fallen below keypool_min (%s). Wallet will no longer watch for new transactions and best block height will not be advanced.\n", keypool_min);
+        LogPrintf("Unlock wallet, top up keypool and rescan to resume watching for new transactions.\n");
+        m_update_best_block = false;
     }
 
     // Check that we haven't dropped below the keypool_critical threshold.
@@ -4082,6 +4105,7 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
         }
 
     }
+
     // Try to top up keypool. No-op if the wallet is locked.
     walletInstance->TopUpKeyPool();
 
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -1644,6 +1644,10 @@ CBlockIndex* CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, bool f
 
         fScanningWallet = false;
     }
+
+    // Check that we haven't dropped below the keypool_critical threshold.
+    ShutdownIfKeypoolCritical();
+
     return ret;
 }
 
@@ -3641,6 +3645,25 @@ void CReserveKey::ReturnKey()
     vchPubKey = CPubKey();
 }
 
+void CWallet::ShutdownIfKeypoolCritical() {
+    LOCK(cs_wallet);
+    unsigned int keypool_critical = GetArg("-keypoolcritical", DEFAULT_KEYPOOL_CRITICAL);
+    if (IsHDEnabled() && !HasUnusedKeys(keypool_critical)) {
+        // if the remaining keypool size is below the gap limit, shutdown
+        LogPrintf("%s: Number of keys in keypool is below critical minimum. Shutting down. internal keypool: %d, external keypool: %d, keypool minimum: %d\n",
+                  __func__, setInternalKeyPool.size(), setExternalKeyPool.size(), keypool_critical);
+        const std::string error_msg = "Number of keys in keypool is below critical minimum and the wallet is encrypted. Bitcoin will now shutdown to avoid loss of funds.\n"
+                                      "This is probably because you are restoring an old backup wallet file which has not been topped up with the most recently "
+                                      "derived keys, and so you would not detect transactions involving those keys.\n"
+                                      "You can manually top-up your wallet keypool as follows:\n"
+                                      " - restart bitcoin with -keypoolcritical set to 0 (to prevent bitcoind from shutting down again)\n"
+                                      " - unlock the wallet using walletpassphrase (to top up the keypool)\n"
+                                      " - restart with -rescan. This will redownload the blockchain if you are running a pruned node.";
+        uiInterface.ThreadSafeMessageBox(error_msg, "", CClientUIInterface::MSG_ERROR);
+        StartShutdown();
+    }
+}
+
 void CWallet::MarkReserveKeysAsUsed(int64_t keypool_id)
 {
     AssertLockHeld(cs_wallet);
@@ -3921,6 +3944,7 @@ std::string CWallet::GetWalletHelpString(bool showDebug)
 
         strUsage += HelpMessageOpt("-dblogsize=<n>", strprintf("Flush wallet database activity from memory to disk log every <n> megabytes (default: %u)", DEFAULT_WALLET_DBLOGSIZE));
         strUsage += HelpMessageOpt("-flushwallet", strprintf("Run a thread to flush wallet periodically (default: %u)", DEFAULT_FLUSHWALLET));
+        strUsage += HelpMessageOpt("-keypoolcritical", strprintf(_("If the keypool drops below this number of keys and we are unable to generate new keys, shutdown (default: %u)"), DEFAULT_KEYPOOL_CRITICAL));
         strUsage += HelpMessageOpt("-privdb", strprintf("Sets the DB_PRIVATE flag in the wallet db environment (default: %u)", DEFAULT_WALLET_PRIVDB));
         strUsage += HelpMessageOpt("-walletrejectlongchains", strprintf(_("Wallet will not create transactions that violate mempool chain limits (default: %u)"), DEFAULT_WALLET_REJECT_LONG_CHAINS));
     }
@@ -4041,6 +4065,16 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
 
     RegisterValidationInterface(walletInstance);
 
+    // Make sure we have enough keys in our keypool if HD is enabled
+    if (walletInstance->IsHDEnabled()) {
+        unsigned int keypool_size = GetArg("-keypool", DEFAULT_KEYPOOL_SIZE);
+        unsigned int keypool_critical = GetArg("-keypoolcritical", DEFAULT_KEYPOOL_CRITICAL);
+        if (keypool_size < keypool_critical) {
+            LogPrintf("Parameter Interaction: keypool critical size (%d) must be smaller than keypool size (%d)\n", keypool_critical, keypool_size);
+            ForceSetArg("-keypool", std::to_string(keypool_critical));
+        }
+
+    }
     // Try to top up keypool. No-op if the wallet is locked.
     walletInstance->TopUpKeyPool();
 
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -417,6 +417,12 @@ bool CWallet::ChangeWalletPassphrase(const SecureString& strOldWalletPassphrase,
 
 void CWallet::SetBestChain(const CBlockLocator& loc)
 {
+    LOCK(cs_wallet); //nWalletMaxVersion
+    unsigned int keypool_min = GetArg("-keypoolmin", DEFAULT_KEYPOOL_MIN);
+    if (IsHDEnabled() && !HasUnusedKeys(keypool_min)) {
+        // If the keypool has dropped below -keypoolmin, then don't update the bestblock height. We can rescan later once the wallet is unlocked.
+        return;
+    }
     CWalletDB walletdb(*dbw);
     walletdb.WriteBestBlock(loc);
 }
@@ -3945,6 +3951,7 @@ std::string CWallet::GetWalletHelpString(bool showDebug)
         strUsage += HelpMessageOpt("-dblogsize=<n>", strprintf("Flush wallet database activity from memory to disk log every <n> megabytes (default: %u)", DEFAULT_WALLET_DBLOGSIZE));
         strUsage += HelpMessageOpt("-flushwallet", strprintf("Run a thread to flush wallet periodically (default: %u)", DEFAULT_FLUSHWALLET));
         strUsage += HelpMessageOpt("-keypoolcritical", strprintf(_("If the keypool drops below this number of keys and we are unable to generate new keys, shutdown (default: %u)"), DEFAULT_KEYPOOL_CRITICAL));
+        strUsage += HelpMessageOpt("-keypoolmin", strprintf(_("If the keypool drops below this number of keys and we are unable to generate new keys, don't advance the wallet's best block (default: %u)"), DEFAULT_KEYPOOL_MIN));
         strUsage += HelpMessageOpt("-privdb", strprintf("Sets the DB_PRIVATE flag in the wallet db environment (default: %u)", DEFAULT_WALLET_PRIVDB));
         strUsage += HelpMessageOpt("-walletrejectlongchains", strprintf(_("Wallet will not create transactions that violate mempool chain limits (default: %u)"), DEFAULT_WALLET_REJECT_LONG_CHAINS));
     }
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -419,7 +419,8 @@ void CWallet::SetBestChain(const CBlockLocator& loc)
 {
     LOCK(cs_wallet); //nWalletMaxVersion
     unsigned int keypool_min = GetArg("-keypoolmin", DEFAULT_KEYPOOL_MIN);
-    if (IsHDEnabled() && (!m_update_best_block || !HasUnusedKeys(keypool_min))) {
+    if (GetBoolArg("-bypasskeypoolcritical", false) || 
+        (IsHDEnabled() && (!m_update_best_block || !HasUnusedKeys(keypool_min)))) {
         // If the keypool has dropped below -keypoolmin, then don't update the bestblock height. We can rescan later once the wallet is unlocked.
 
         if (m_update_best_block) {
@@ -3677,7 +3678,9 @@ void CReserveKey::ReturnKey()
 void CWallet::ShutdownIfKeypoolCritical() {
     LOCK(cs_wallet);
     unsigned int keypool_critical = GetArg("-keypoolcritical", DEFAULT_KEYPOOL_CRITICAL);
-    if (IsHDEnabled() && !HasUnusedKeys(keypool_critical)) {
+    if (!GetBoolArg("-bypasskeypoolcritical", false) &&
+        IsHDEnabled() &&
+        !HasUnusedKeys(keypool_critical)) {
         // if the remaining keypool size is below the gap limit, shutdown
         LogPrintf("%s: Number of keys in keypool is below critical minimum. Shutting down. internal keypool: %d, external keypool: %d, keypool minimum: %d\n",
                   __func__, setInternalKeyPool.size(), setExternalKeyPool.size(), keypool_critical);
@@ -3685,9 +3688,11 @@ void CWallet::ShutdownIfKeypoolCritical() {
                                       "This is probably because you are restoring an old backup wallet file which has not been topped up with the most recently "
                                       "derived keys, and so you would not detect transactions involving those keys.\n"
                                       "You can manually top-up your wallet keypool as follows:\n"
-                                      " - restart bitcoin with -keypoolcritical set to 0 (to prevent bitcoind from shutting down again)\n"
+                                      " - restart bitcoin with -bypasskeypoolcritical (to prevent bitcoind from shutting down again)\n"
                                       " - unlock the wallet using walletpassphrase (to top up the keypool)\n"
-                                      " - restart with -rescan. This will redownload the blockchain if you are running a pruned node.";
+                                      " - restart without -bypasskeypoolcritical.\n"
+                                      "NOTE: if you have a pruned node that prunes blocks your wallet hasn't scanned yet when restarting with -bypasskeypoolcritical "
+                                      "then you may need to do a complete reindex. Consider raising the prune limit temporarily for both restarts to avoid this.";
         uiInterface.ThreadSafeMessageBox(error_msg, "", CClientUIInterface::MSG_ERROR);
         StartShutdown();
     }
@@ -3971,6 +3976,7 @@ std::string CWallet::GetWalletHelpString(bool showDebug)
     {
         strUsage += HelpMessageGroup(_("Wallet debugging/testing options:"));
 
+        strUsage += HelpMessageOpt("-bypasskeypoolcritical", _("Bypass keypool critical limit. Don't shutdown the node if keypool drops below keypoolcritical limit (but don't advance BestBlock)"));
         strUsage += HelpMessageOpt("-dblogsize=<n>", strprintf("Flush wallet database activity from memory to disk log every <n> megabytes (default: %u)", DEFAULT_WALLET_DBLOGSIZE));
         strUsage += HelpMessageOpt("-flushwallet", strprintf("Run a thread to flush wallet periodically (default: %u)", DEFAULT_FLUSHWALLET));
         strUsage += HelpMessageOpt("-keypoolcritical", strprintf(_("If the keypool drops below this number of keys and we are unable to generate new keys, shutdown (default: %u)"), DEFAULT_KEYPOOL_CRITICAL));
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -3444,11 +3444,21 @@ void CWallet::ReturnKey(int64_t nIndex, bool fInternal, const CPubKey& pubkey)
     LogPrintf("keypool return %d\n", nIndex);
 }
 
-bool CWallet::GetKeyFromPool(CPubKey& result, bool internal)
+bool CWallet::GetKeyFromPool(CPubKey& result, bool internal, bool fail_on_critical)
 {
     CKeyPool keypool;
     {
         LOCK(cs_wallet);
+
+        // First, try to top up the keypool (this is a no-op if wallet is encrypted and locked)
+        TopUpKeyPool();
+        if (fail_on_critical && IsHDEnabled() &&
+            !HasUnusedKeys(GetArg("-keypoolcritical", DEFAULT_KEYPOOL_CRITICAL) + 1)) {
+            // If getting a key from the pool would result in us dropping to
+            // the keypool_critical threshold, fail
+            return false;
+        }
+
         int64_t nIndex = 0;
         ReserveKeyFromKeyPool(nIndex, keypool, internal);
         if (nIndex == -1)
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -40,7 +40,13 @@ extern unsigned int nTxConfirmTarget;
 extern bool bSpendZeroConfChange;
 extern bool fWalletRbf;
 
+//! Default size for the keypool. Always try to top up the keypool to have this
+//  many keys. For an HD split wallet, have this many keys in each of the
+//  internal/external chains
 static const unsigned int DEFAULT_KEYPOOL_SIZE = 1000;
+//! Don't update wallet's best block if keypool falls below this size (to avoid
+//  not detecting transactions)
+static const unsigned int DEFAULT_KEYPOOL_MIN = 500;
 //! Shut down if the keypool falls below this size
 static const unsigned int DEFAULT_KEYPOOL_CRITICAL = 500;
 //! -paytxfee default
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -665,6 +665,7 @@ private:
     static std::atomic<bool> fFlushScheduled;
     std::atomic<bool> fAbortRescan;
     std::atomic<bool> fScanningWallet;
+    bool m_update_best_block;
 
     /**
      * Select a set of coins such that nValueRet >= nTargetValue and at least
@@ -800,6 +801,7 @@ public:
         nRelockTime = 0;
         fAbortRescan = false;
         fScanningWallet = false;
+        m_update_best_block = true;
     }
 
     std::map<uint256, CWalletTx> mapWallet;
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -987,7 +987,7 @@ public:
     void ReserveKeyFromKeyPool(int64_t& nIndex, CKeyPool& keypool, bool fRequestedInternal);
     void KeepKey(int64_t nIndex);
     void ReturnKey(int64_t nIndex, bool fInternal, const CPubKey& pubkey);
-    bool GetKeyFromPool(CPubKey &key, bool internal = false);
+    bool GetKeyFromPool(CPubKey &key, bool internal = false, bool fail_on_critical = false);
     int64_t GetOldestKeyPoolTime();
     void ShutdownIfKeypoolCritical();
     /**
--- a/src/wallet/wallet.h
+++ b/src/wallet/wallet.h
@@ -41,6 +41,8 @@ extern bool bSpendZeroConfChange;
 extern bool fWalletRbf;
 
 static const unsigned int DEFAULT_KEYPOOL_SIZE = 1000;
+//! Shut down if the keypool falls below this size
+static const unsigned int DEFAULT_KEYPOOL_CRITICAL = 500;
 //! -paytxfee default
 static const CAmount DEFAULT_TRANSACTION_FEE = 0;
 //! -fallbackfee default
@@ -979,6 +981,7 @@ public:
     void ReturnKey(int64_t nIndex, bool fInternal, const CPubKey& pubkey);
     bool GetKeyFromPool(CPubKey &key, bool internal = false);
     int64_t GetOldestKeyPoolTime();
+    void ShutdownIfKeypoolCritical();
     /**
      * Marks all keys in the keypool up to and including reserve_key as used.
      */
--- a/src/qt/paymentserver.cpp
+++ b/src/qt/paymentserver.cpp
@@ -648,7 +648,7 @@ void PaymentServer::fetchPaymentACK(CWallet* wallet, SendCoinsRecipient recipien
     }
     else {
         CPubKey newKey;
-        if (wallet->GetKeyFromPool(newKey)) {
+        if (wallet->GetKeyFromPool(newKey, false, true)) {
             CKeyID keyID = newKey.GetID();
             wallet->SetAddressBook(keyID, strAccount, "refund");
 
--- a/test/functional/keypool-topup.py
+++ b/test/functional/keypool-topup.py
@@ -6,15 +6,23 @@
 
 Two nodes. Node1 is under test. Node0 is providing transactions and generating blocks.
 
+Repeat test twice, once with encrypted wallet and once with unencrypted wallet:
+
 - Start node1, shutdown and backup wallet.
 - Generate 110 keys (enough to drain the keypool). Store key 90 (in the initial keypool) and key 110 (beyond the initial keypool). Send funds to key 90 and key 110.
 - Stop node1, clear the datadir, move wallet file back into the datadir and restart node1.
-- connect node1 to node0. Verify that they sync and node1 receives its funds."""
+- if wallet is unencrypted:
+    - connect node1 to node0. Verify that they sync and node1 receives its funds.
+- else wallet is encrypted:
+    - connect node1 to node0. Verify that node1 shuts down to avoid loss of funds.
+"""
+import os
 import shutil
 
-from test_framework.test_framework import BitcoinTestFramework
+from test_framework.test_framework import BitcoinTestFramework, BITCOIND_PROC_WAIT_TIMEOUT
 from test_framework.util import (
     assert_equal,
+    assert_raises_jsonrpc,
     connect_nodes_bi,
     sync_blocks,
 )
@@ -24,15 +32,27 @@ class KeypoolRestoreTest(BitcoinTestFramework):
         super().__init__()
         self.setup_clean_chain = True
         self.num_nodes = 2
-        self.extra_args = [['-usehd=0'], ['-usehd=1', '-keypool=100', '-keypoolmin=20']]
+        self.extra_args = [['-usehd=0'], ['-usehd=1', '-keypool=100', '-keypoolmin=20', '-keypoolcritical=20']]
 
     def run_test(self):
         self.tmpdir = self.options.tmpdir
         self.nodes[0].generate(101)
 
+        self.log.info("Test keypool restore (encrypted wallet)")
+        self._test_restore(encrypted=True)
+
+        self.log.info("Test keypool restore (unencrypted wallet)")
+        self._test_restore(encrypted=False)
+
+    def _test_restore(self, encrypted):
+
         self.log.info("Make backup of wallet")
 
-        self.stop_node(1)
+        if encrypted:
+            self.nodes[1].encryptwallet('test')
+            self.bitcoind_processes[1].wait(timeout=BITCOIND_PROC_WAIT_TIMEOUT)
+        else:
+            self.stop_node(1)
 
         shutil.copyfile(self.tmpdir + "/node1/regtest/wallet.dat", self.tmpdir + "/wallet.bak")
         self.nodes[1] = self.start_node(1, self.tmpdir, self.extra_args[1])
@@ -40,7 +60,13 @@ class KeypoolRestoreTest(BitcoinTestFramework):
 
         self.log.info("Generate keys for wallet")
 
-        for _ in range(90):
+        for _ in range(80):
+            self.nodes[1].getnewaddress()
+        if encrypted:
+            # Keypool can't top up because the wallet is encrypted
+            assert_raises_jsonrpc(-12, "Keypool is at critical level or has run out, please call keypoolrefill first", self.nodes[1].getnewaddress)
+            self.nodes[1].walletpassphrase("test", 10)
+        for _ in range(10):
             addr_oldpool = self.nodes[1].getnewaddress()
         for _ in range(20):
             addr_extpool = self.nodes[1].getnewaddress()
@@ -59,17 +85,28 @@ class KeypoolRestoreTest(BitcoinTestFramework):
 
         shutil.copyfile(self.tmpdir + "/wallet.bak", self.tmpdir + "/node1/regtest/wallet.dat")
 
-        self.log.info("Verify keypool is restored and balance is correct")
+        if encrypted:
+            self.log.info("Encrypted wallet - verify node shuts down to avoid loss of funds")
 
-        self.nodes[1] = self.start_node(1, self.tmpdir, self.extra_args[1])
-        connect_nodes_bi(self.nodes, 0, 1)
-        self.sync_all()
+            self.assert_start_raises_init_error(1, self.tmpdir, self.extra_args[1], expected_msg="Number of keys in keypool is below critical minimum and the wallet is encrypted. Bitcoin will now shutdown to avoid loss of funds.")
 
-        assert_equal(self.nodes[1].getbalance(), 15)
-        assert_equal(self.nodes[1].listtransactions()[0]['category'], "receive")
+            shutil.rmtree(self.tmpdir + "/node1/regtest/chainstate")
+            shutil.rmtree(self.tmpdir + "/node1/regtest/blocks")
+            os.remove(self.tmpdir + "/node1/regtest/wallet.dat")
+            self.nodes[1] = self.start_node(1, self.tmpdir, self.extra_args[1])
 
-        # Check that we have marked all keys up to the used keypool key as used
-        assert_equal(self.nodes[1].validateaddress(self.nodes[1].getnewaddress())['hdkeypath'], "m/0'/0'/111'")
+        else:
+            self.log.info("Unencrypted wallet - verify keypool is restored and balance is correct")
+
+            self.nodes[1] = self.start_node(1, self.tmpdir, self.extra_args[1])
+            connect_nodes_bi(self.nodes, 0, 1)
+            self.sync_all()
+
+            assert_equal(self.nodes[1].getbalance(), 15)
+            assert_equal(self.nodes[1].listtransactions()[0]['category'], "receive")
+
+            # Check that we have marked all keys up to the used keypool key as used
+            assert_equal(self.nodes[1].validateaddress(self.nodes[1].getnewaddress())['hdkeypath'], "m/0'/0'/111'")
 
 if __name__ == '__main__':
     KeypoolRestoreTest().main()
--- a/test/functional/keypool.py
+++ b/test/functional/keypool.py
@@ -27,7 +27,7 @@ class KeyPoolTest(BitcoinTestFramework):
         wallet_info = nodes[0].getwalletinfo()
         assert(addr_before_encrypting_data['hdmasterkeyid'] != wallet_info['hdmasterkeyid'])
         assert(addr_data['hdmasterkeyid'] == wallet_info['hdmasterkeyid'])
-        assert_raises_jsonrpc(-12, "Error: Keypool ran out, please call keypoolrefill first", nodes[0].getnewaddress)
+        assert_raises_jsonrpc(-12, "Keypool is at critical level or has run out, please call keypoolrefill first", nodes[0].getnewaddress)
 
         # put six (plus 2) new keys in the keypool (100% external-, +100% internal-keys, 1 in min)
         nodes[0].walletpassphrase('test', 12000)
@@ -57,7 +57,7 @@ class KeyPoolTest(BitcoinTestFramework):
         addr.add(nodes[0].getnewaddress())
         assert(len(addr) == 6)
         # the next one should fail
-        assert_raises_jsonrpc(-12, "Error: Keypool ran out, please call keypoolrefill first", nodes[0].getnewaddress)
+        assert_raises_jsonrpc(-12, "Keypool is at critical level or has run out, please call keypoolrefill first", nodes[0].getnewaddress)
 
         # refill keypool with three new addresses
         nodes[0].walletpassphrase('test', 1)
--- a/test/functional/keypool.py
+++ b/test/functional/keypool.py
@@ -44,19 +44,10 @@ class KeyPoolTest(BitcoinTestFramework):
         nodes[0].getrawchangeaddress()
         nodes[0].getrawchangeaddress()
         nodes[0].getrawchangeaddress()
-        addr = set()
         # the next one should fail
         assert_raises_jsonrpc(-12, "Keypool ran out", nodes[0].getrawchangeaddress)
 
-        # drain the external keys
-        addr.add(nodes[0].getnewaddress())
-        addr.add(nodes[0].getnewaddress())
-        addr.add(nodes[0].getnewaddress())
-        addr.add(nodes[0].getnewaddress())
-        addr.add(nodes[0].getnewaddress())
-        addr.add(nodes[0].getnewaddress())
-        assert(len(addr) == 6)
-        # the next one should fail
+        # getting an external key should also fail now
         assert_raises_jsonrpc(-12, "Keypool is at critical level or has run out, please call keypoolrefill first", nodes[0].getnewaddress)
 
         # refill keypool with three new addresses
@@ -68,9 +59,8 @@ class KeyPoolTest(BitcoinTestFramework):
         assert_equal(nodes[0].getwalletinfo()["unlocked_until"], 0)
 
         # drain them by mining
-        nodes[0].generate(1)
-        nodes[0].generate(1)
-        nodes[0].generate(1)
+        for _ in range(6):
+            nodes[0].generate(1)
         assert_raises_jsonrpc(-12, "Keypool ran out", nodes[0].generate, 1)
 
         nodes[0].walletpassphrase('test', 100)
--- a/test/functional/test_framework/test_framework.py
+++ b/test/functional/test_framework/test_framework.py
@@ -216,7 +216,7 @@ class BitcoinTestFramework(object):
         datadir = os.path.join(dirname, "node" + str(i))
         if binary is None:
             binary = os.getenv("BITCOIND", "bitcoind")
-        args = [binary, "-datadir=" + datadir, "-server", "-keypool=1", "-discover=0", "-rest", "-logtimemicros", "-debug", "-debugexclude=libevent", "-debugexclude=leveldb", "-mocktime=" + str(self.mocktime), "-uacomment=testnode%d" % i]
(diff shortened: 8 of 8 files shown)


test/functional/keypool-restore.py
```diff
@@ -0,0 +1,157 @@
...
+
+from test_framework.test_framework import BitcoinTestFramework, BITCOIND_PROC_WAIT_TIMEOUT
+from test_framework.util import assert_equal, connect_nodes_bi
+
+class KeypoolRestoreTest(BitcoinTestFramework):
+    def __init__(self):
+        super().__init__()
+        self.setup_clean_chain = True
+        self.num_nodes = 2
+        self.extra_args = [['-usehd=0'], ['-usehd=1', '-keypool=100', '-keypoolemin=20']]
```
ryanofsky: keypoolemin seems misspelled

instagibbs:
There really needs to be some sort of recourse presented to the user upon shutdown.

test/functional/keypool-restore.py
```diff
@@ -0,0 +1,157 @@
...
+
+        # txdata_f = self.nodes[0].fundrawtransaction(txdata)
+        # txdata_s = self.nodes[0].signrawtransaction(txdata_f['hex'])
+        # self.nodes[0].sendrawtransaction(txdata_s['hex'])
+
+        # self.sync_all()
+        # self.nodes[0].generate(1)
+        self.sync_all()
+
+        assert_equal(self.nodes[1].getbalance(), 15)
```
ryanofsky: On test failure happening this line, I think the reason is that encryptwallet generates an entirely new HD master key, so `addr_enc_oldpool` and `addr_enc_extpool` come from a new HD master key which is not part of "hd.bak"

Restoring "hd.enc.bak" instead of "hd.bak" makes this check pass: https://github.com/ryanofsky/bitcoin/commit/93349ad5d915ee152a1a84920f18cbe0cab8b036

jnewbery: Thank you for this insight! Yes - that's what was causing the test to fail.

I've made some progress with the test. I've pushed what I've got so far, and hope to finish it off tomorrow morning.

src/wallet/wallet.cpp
```diff
@@ -1004,6 +1037,26 @@ bool CWallet::AddToWalletIfInvolvingMe(const CTransactionRef& ptx, const CBlockI
         if (fExisted && !fUpdate) return false;
         if (fExisted || IsMine(tx) || IsFromMe(tx))
         {
+            /* Check if any keys in the wallet keypool that were supposed to be unused
+             * have appeared in a new transaction. If so, remove those keys from the keypool.
+             * This can happen when restoring an old wallet backup that does not contain
+             * the mostly recently created transactions from newer versions of the wallet.
+             */
+            std::set<CKeyID> keyPool;
+            GetAllReserveKeys(keyPool);
```
ryanofsky: It seems like this code would more efficient and maybe simpler if instead of making a set of CKeyID's here, we made a map from CKeyID -> (keypool_index, is_internal) and passed it into MarkReserveKeysAsUsed, so the first two loops in that function could be removed.

src/wallet/wallet.cpp
```diff
@@ -3523,18 +3576,78 @@ void CReserveKey::ReturnKey()
...
+                    throw std::runtime_error(strprintf("%s: read failed for index %d", __func__, id));
+                }
+
+                KeepKey(id);
+                it = setKeyPool->erase(it);
+            }
+        }
+    }
+
+    if (IsHDEnabled() && !TopUpKeyPool()) {
```
ryanofsky: @sipa's comment about only topping up hd key pools would seem to apply here and maybe to CheckKeypoolMinSize: https://github.com/bitcoin/bitcoin/pull/10240#pullrequestreview-49644483

src/wallet/wallet.cpp
```diff
@@ -3523,18 +3576,78 @@ void CReserveKey::ReturnKey()
...
     }
 }
 
+void CWallet::MarkReserveKeysAsUsed(const CKeyID& keyId)
+{
+    AssertLockHeld(cs_wallet);
+    CWalletDB walletdb(*dbw);
+    for (std::set<int64_t> *setKeyPool : {&setInternalKeyPool, &setExternalKeyPool}) {
+        int64_t foundIndex = -1;
+        for (const int64_t& id : *setKeyPool) {
```
ryanofsky: Maybe use `index` instead of `id` here and other places to distinguish from pool indices from key ids.

src/wallet/wallet.cpp
jnewbery: Yes, you're right that this would be much more efficient. However, `GetAllReserveKeys()` is also called elsewhere, which would also need to be modified to accept a map, so I'd prefer not to change it as part of this PR.

This can be fixed in a follow-up PR unless you think the performance in `MarkReserveKeysAsUsed()` is unacceptable.

src/wallet/wallet.cpp
```diff
@@ -3523,18 +3576,78 @@ void CReserveKey::ReturnKey()
     vchPubKey = CPubKey();
 }
 
+void CWallet::CheckKeypoolMinSize() {
+    unsigned int keypool_min = GetArg("-keypoolmin", DEFAULT_KEYPOOL_MIN);
+    if (IsHDEnabled() && (setInternalKeyPool.size() < keypool_min || (setExternalKeyPool.size() < keypool_min))) {
```
jnewbery: I think you're right. Fixed

src/wallet/wallet.cpp
```diff
@@ -3523,18 +3576,78 @@ void CReserveKey::ReturnKey()
...
+        // if the remaining keypool size is below the gap limit, shutdown
+        LogPrintf("%s: Keypool is too small. Shutting down. internal keypool: %d, external keypool: %d, keypool minimum: %d\n",
+                  __func__, setInternalKeyPool.size(), setExternalKeyPool.size(), keypool_min);
+        const static std::string error_msg = "Keypool is too small. Shutting down";
```
jnewbery: yup

src/wallet/wallet.cpp
jnewbery: done

src/wallet/wallet.cpp
```diff
@@ -3523,18 +3576,78 @@ void CReserveKey::ReturnKey()
...
+
+        auto it = std::begin(*setKeyPool);
+        // mark all keys up to the found key as used
+        if (foundIndex >= 0) {
+            while (it != std::end(*setKeyPool)) {
+                const int64_t& id = *(it);
+                if (id > foundIndex) break; // set*KeyPool is ordered
+
+                CKeyPool keypool;
+                if (!walletdb.ReadPool(id, keypool)) {
```
jnewbery: removed

src/wallet/wallet.cpp
jnewbery: Yes. Applies here. I don't think I need to change `CheckKeypoolMinSize()` though. The node should only be shutdown for HD wallets.

src/wallet/wallet.cpp
```diff
@@ -3954,6 +4036,25 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
...
+    // Make sure we have enough keys in our keypool if HD is enabled
+    if (walletInstance->IsHDEnabled()) {
+        unsigned int keypool_size = GetArg("-keypool", DEFAULT_KEYPOOL_SIZE);
+        unsigned int keypool_min = GetArg("-keypoolmin", DEFAULT_KEYPOOL_MIN);
+        if (walletInstance->IsCrypted()) {
+            if (keypool_size < keypool_min) {
+                LogPrintf("Parameter Interaction: keypool size (%d) must be larger than keypool minimum size for encrypted wallets (%d)\n", keypool_size, keypool_min);
+                SoftSetArg("-keypool", std::to_string(keypool_min));
+            }
+            InitWarning(_("You are using an encrypted HD wallet. You may miss incoming or outgoing transactions."));
```
jnewbery: Updated. Let me know what you think of the new text.

src/wallet/wallet.cpp
```diff
@@ -3954,6 +4036,25 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
...
+        } else {
+            if (keypool_size < keypool_min && keypool_size < DEFAULT_KEYPOOL_MIN) {
```
jnewbery: yes - fixed

src/wallet/wallet.cpp
```diff
@@ -3954,6 +4036,25 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
...
+                InitWarning(_("Your keypool size is below the recommended limit for HD rescans. You may miss incoming or outgoing transactions."));
```
jnewbery: Updated. Let me know what you think of the new text.

jnewbery:
Thanks for the review @ryanofsky . I've addressed all of your concerns except the `GetAllReserveKeys()` refactor which can be done later.

All the changes are in separate fixup commits which can be squashed later.

jnewbery:
Test is fixed. It required #10703 since the test involves node1 emitting warnings to stderr.

src/wallet/wallet.cpp
```diff
@@ -3954,6 +4031,27 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
 
     RegisterValidationInterface(walletInstance);
 
+    // Make sure we have enough keys in our keypool if HD is enabled
+    if (walletInstance->IsHDEnabled()) {
+        unsigned int keypool_size = GetArg("-keypool", DEFAULT_KEYPOOL_SIZE);
+        unsigned int keypool_min = GetArg("-keypoolmin", DEFAULT_KEYPOOL_MIN);
+        if (walletInstance->IsCrypted()) {
+            if (keypool_size < keypool_min) {
+                LogPrintf("Parameter Interaction: keypool size (%d) must be larger than keypool minimum size for encrypted wallets (%d)\n", keypool_size, keypool_min);
```
ryanofsky: If keypool size was provided nothing actually happens here so the log message seems misleading. Maybe change SoftSetArg to ForceSetArg or only log this if SoftSetArg returns true.

src/wallet/wallet.cpp
```diff
@@ -3954,6 +4031,27 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
...
+                SoftSetArg("-keypool", std::to_string(keypool_min));
+            }
+            InitWarning(strprintf(_("You are using an encrypted HD wallet. If you are restoring an old HD wallet that has not been topped up with the most recently "
```
ryanofsky: Maybe explicitly warn about shutdowns, too

src/wallet/wallet.cpp
```diff
@@ -3954,6 +4031,27 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
...
+                                    "derived keys your wallet may not detect transactions involving those keys. You should manually top-up your wallet keypool.")));
+        } else {
+            if (keypool_size < keypool_min) {
+                InitWarning(strprintf(_("Your keypool is configured to store %d keys, which is below the keypool minimum size of %d. Using a larger keypool will make it less "
+                                        "likely that your wallet will be missing transactions and funds if it is restored from an old backup."), keypool_size, keypool_min));
+            }
+            walletInstance->TopUpKeyPool();
+        }
+        walletInstance->CheckKeypoolMinSize();
```
instagibbs: We need to hold the wallet lock here or we fail the assertion at wallet/wallet.cpp:830 in  `CanSupportFeature`.

test/functional/keypool-restore.py
```diff
@@ -0,0 +1,113 @@
...
+)
+
+class KeypoolRestoreTest(BitcoinTestFramework):
+    def __init__(self):
+        super().__init__()
+        self.setup_clean_chain = True
+        self.num_nodes = 2
+        self.extra_args = [['-usehd=0'], ['-usehd=1', '-keypool=100', '-keypoolmin=20']]
+
+    def run_test(self):
```
ryanofsky: Maybe split this up into test_restore(encrypted=False/True) or test_encrypted_restore/test_unencrypted_restore calls to make the test less repetitive and simpler to understand.

ryanofsky (commented):
utACK 965594a5528ce74e6a1d51232da729c166d176d3, left minor suggestions but this seems reasonable.

I do think [using a map](https://github.com/bitcoin/bitcoin/pull/10882#discussion_r128380261) would probably simplify the code (even without changing the GetAllReserveKeys method), not just make it more efficient, so it might be something to reconsider if you end up making more changes for another reason.

Also maybe squash the commit history. I don't think it it's helpful at all. Would just keep the MOVEONLY commit, the #10703 commit, and then combine everything else in another commit.

jnewbery:
Squashed down to 4 commits. Will address @instagibbs and @ryanofsky feedback next.

src/wallet/wallet.cpp
jnewbery: is it ok to lock in `CheckKeypoolMinSize()` instead?

instagibbs: tI think so. You'd be calling for the lock twice in the other path, but that's ok right?

jnewbery: yes, I don't think there's a problem doing that.

src/wallet/wallet.cpp
```diff
@@ -3523,18 +3576,73 @@ void CReserveKey::ReturnKey()
...
+                LogPrintf("%s: Found keypool index %d\n", __func__, index);
+                foundIndex = index;
+                if (!keypool.fInternal) {
+                    SetAddressBook(keyId, "", "receive");
+                }
+                break;
+            }
+        }
+
+        auto it = std::begin(*setKeyPool);
```
jnewbery: done

src/wallet/wallet.cpp
```diff
@@ -3523,18 +3576,73 @@ void CReserveKey::ReturnKey()
...
+                const int64_t& index = *(it);
+                if (index > foundIndex) break; // set*KeyPool is ordered
+
+                KeepKey(index);
+                it = setKeyPool->erase(it);
+            }
+        }
+    }
+
+    if (!TopUpKeyPool()) {
```
jnewbery: done

src/wallet/wallet.cpp
jnewbery: done

src/wallet/wallet.cpp
jnewbery: done

test/functional/keypool-restore.py
jnewbery: done

gmaxwell:
Can we make this also suppress the relock while scanning ... so that it's viable to just unlock and have it stay unlocked until its done-ish?

What will this do if, without any rescan, I exhaust all the keys in the wallet... will it falsely trigger?

src/wallet/wallet.cpp
```diff
@@ -3523,18 +3582,68 @@ void CReserveKey::ReturnKey()
     vchPubKey = CPubKey();
 }
 
+void CWallet::CheckKeypoolMinSize() {
```
instagibbs: Method name should reflect the fact it will shut down upon failure.

jnewbery: suggested name?

jnewbery:
I've implemented @ryanofsky's suggestion for simplifying `MarkReserveKeysAsUsed()` and improved the error message if the node shuts down.

instagibbs:
light tACK

Can confirm the new directions upon failure to topup using Crypted lead to recovery of proper index position.

src/wallet/wallet.cpp
```diff
@@ -3954,6 +4019,25 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
...
+                ForceSetArg("-keypool", std::to_string(keypool_min));
+            }
+        } else {
+            if (keypool_size < keypool_min) {
+                InitWarning(strprintf(_("Your keypool is configured to store %d keys, which is below the keypool minimum size of %d. Using a larger keypool will make it less "
+                                        "likely that your wallet will be missing transactions and funds if it is restored from an old backup."), keypool_size, keypool_min));
+            }
+            walletInstance->TopUpKeyPool();
+        }
+        walletInstance->CheckKeypoolMinSize();
```
instagibbs: The new error message doesn't quite work here, but it's also quite unlikely to happen, so maybe moot.

src/wallet/wallet.cpp
```diff
@@ -3523,30 +3584,65 @@ void CReserveKey::ReturnKey()
...
+        // if the remaining keypool size is below the gap limit, shutdown
+        LogPrintf("%s: Number of keys in keypool is below minimum. Shutting down. internal keypool: %d, external keypool: %d, keypool minimum: %d\n",
+                  __func__, setInternalKeyPool.size(), setExternalKeyPool.size(), keypool_min);
+        const std::string error_msg = "Number of keys in keypool is below minimum and the wallet is encrytped. Bitcoin will now shutdown to avoid loss of funds.\n"
+                                      "This is probably because you are restoring an old backup wallet file which has not been topped up with the most recently "
+                                      "derived keys, and so you would not detect transactions involving those keys.\n"
+                                      "You can manually top-up your wallet keypool as follows:\n"
+                                      " - restart bitcoin with -keypoolmin set to 0\n"
+                                      " - unlock the wallet using walletpassphrase\n"
+                                      " - restart with -rescan. This will redownload the blockchain if you are running a pruned node.";
```
instagibbs: Just noting in case I'm underestimating the likelihood/cost: unlikely but this could be painful if they're using more than 1500 keys, as this might trigger multiple times causing them to re-download the chain multiple times.

This will go away with proper long-term fix.

src/wallet/wallet.cpp
```diff
@@ -3589,9 +3589,15 @@ void CWallet::CheckKeypoolMinSize() {
...
+        LogPrintf("%s: Number of keys in keypool is below minimum. Shutting down. internal keypool: %d, external keypool: %d, keypool minimum: %d\n",
                   __func__, setInternalKeyPool.size(), setExternalKeyPool.size(), keypool_min);
-        const std::string error_msg = "Keypool is too small. Shutting down";
+        const std::string error_msg = "Number of keys in keypool is below minimum and the wallet is encrytped. Bitcoin will now shutdown to avoid loss of funds.\n"
+                                      "This is probably because you are restoring an old backup wallet file which has not been topped up with the most recently "
+                                      "derived keys, and so you would not detect transactions involving those keys.\n"
+                                      "You can manually top-up your wallet keypool as follows:\n"
+                                      " - restart bitcoin with -keypoolmin set to 0\n"
+                                      " - unlock the wallet using walletpassphrase\n"
+                                      " - restart with -rescan. This will redownload the blockchain if you are running a pruned node.";
```
ryanofsky: This seems safe, but practically speaking shouldn't a manual rescan be unnecessary in most cases, because if the node shuts down when the keypool size is low, restarting should automatically rescan from `ReadBestBlock`.

Relatedly, maybe this PR should explicitly avoid calling `WriteBestBlock` when the keypool size is too low, in case shutdown takes a long time or -keypoolmin is set to 0 to recover.

gmaxwell: > Relatedly, maybe this PR should explicitly avoid calling WriteBestBlock when the keypool size is too low, 

ACK. Very much so.  If not for having no mechanism for telling you that your wallet has fallen behind, this would more or less eliminate the need for shutdown.  though it's weird that you'd lower the minimum to bypass the shutdown and inadvertently update the height.

ryanofsky: Maybe there should be two flags:

`-keypoolmin` as the threshold for calling `WriteBestBlock`
`-keypoolcriticalmin` or `-keypoolshutdownmin` as the threshold for shutting down

Both flags could be set to 500. This still wouldn't provide a safe recovery path that avoids a complete rescan (we would need an RPC call or another mechanism that could rescan from bestblock after unlocking), but that could be added in the future, and in the meantime this would provide a safer way to avoid shutdowns if a rescan can't be done right away.

src/wallet/wallet.cpp
jnewbery: sorry - I don't understand this comment. Can you explain what you mean by the new error message not quite working?

src/wallet/wallet.cpp
jnewbery: What's special about 1500?

instagibbs: Ok my math is off, I think 1000, I just meant the 2nd time this could occur.

src/wallet/wallet.cpp
instagibbs: mistaken, ignore

jnewbery:
Added two commits:

1. rename current option `keypoolcritical` - if the keypool falls below this, then shutdown the node
2. add option `keypoolmin` - if the keypool falls below this, don't update the wallet's best block.

If people think that's conceptually the right approach, then I'll squash down into sensible commits.

src/wallet/wallet.cpp
```diff
@@ -3523,30 +3589,65 @@ void CReserveKey::ReturnKey()
     vchPubKey = CPubKey();
 }
 
-static void LoadReserveKeysToSet(std::set<CKeyID>& setAddress, const std::set<int64_t>& setKeyPool, CWalletDB& walletdb) {
-    for (const int64_t& id : setKeyPool)
+void CWallet::CheckKeypoolMinSize() {
```
ryanofsky: Could s/CheckKeypoolMinSize/CheckKeypoolCriticalSize

There was also an old comment about renaming this. https://github.com/bitcoin/bitcoin/pull/10882#discussion_r128611234. Could go with something with something like `MaybeShutdownForLaggingWallet`

src/wallet/wallet.cpp
```diff
@@ -3834,6 +3903,7 @@ std::string CWallet::GetWalletHelpString(bool showDebug)
 
         strUsage += HelpMessageOpt("-dblogsize=<n>", strprintf("Flush wallet database activity from memory to disk log every <n> megabytes (default: %u)", DEFAULT_WALLET_DBLOGSIZE));
         strUsage += HelpMessageOpt("-flushwallet", strprintf("Run a thread to flush wallet periodically (default: %u)", DEFAULT_FLUSHWALLET));
+        strUsage += HelpMessageOpt("-keypoolcritical", strprintf(_("If the keypool drops below this number of keys and we are unable to generate new keys, shutdown (default: %u)"), DEFAULT_KEYPOOL_CRITICAL));
```
ryanofsky: Need to add `-keypoolmin` I think too.

src/wallet/wallet.cpp
```diff
@@ -382,6 +415,11 @@ bool CWallet::ChangeWalletPassphrase(const SecureString& strOldWalletPassphrase,
 
 void CWallet::SetBestChain(const CBlockLocator& loc)
 {
+    unsigned int keypool_min = GetArg("-keypoolmin", DEFAULT_KEYPOOL_MIN);
+    if (IsHDEnabled() && ((CanSupportFeature(FEATURE_HD_SPLIT) && setInternalKeyPool.size() < keypool_min) || (setExternalKeyPool.size() < keypool_min))) {
```
jnewbery: As discussed, if the user is trying to restore from an old non-HD wallet and they drop below keypoolmin/keypoolcritical, there's very little they can do (since topup will generate new random keys). I'll leave this in for now, but if you think we definitely should shutdown or stop advancing best block for non-HD wallets, let me know.

Unless you strongly feel that I should add the helper method, I'll leave this as explicitly coded in both places.

src/wallet/wallet.cpp
jnewbery: Good point. Changed name to `ShutdownIfKeypoolCritical()`

src/wallet/wallet.cpp
jnewbery: done

src/wallet/wallet.h
```diff
@@ -988,7 +996,9 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
     void ReturnKey(int64_t nIndex, bool fInternal);
     bool GetKeyFromPool(CPubKey &key, bool internal = false);
     int64_t GetOldestKeyPoolTime();
-    void GetAllReserveKeys(std::set<CKeyID>& setAddress) const;
+    void CheckKeypoolMinSize();
+    void MarkReserveKeysAsUsed(const int64_t keypool_index, const bool internal);
+    void GetAllReserveKeys(std::map<CKeyID, std::pair<int64_t, bool>>& mapKeyPool) const;
```
jnewbery: done

jnewbery:
@ryanofsky I've addressed all of your comments. Let me know if you're happy with the last three commits and I'll squash them down for the benefit of other reviewers,

src/wallet/wallet.cpp
```diff
@@ -3523,33 +3588,70 @@ void CReserveKey::ReturnKey()
...
+
+        KeepKey(index);
+        it = setKeyPool->erase(it);
+    }
+}
+
+std::map<CKeyID, ReserveKey> CWallet::GetAllReserveKeys() const
+{
+    std::map<CKeyID, ReserveKey> mapKeyPool;
+    mapKeyPool.clear();
```
ryanofsky: Should already be empty

ryanofsky (commented):
utACK 89af21ddcd7837c45cf8e261711d227e4f8448db

src/wallet/wallet.cpp
jnewbery: dropped

jnewbery:
All comments should now be addressed and changes squashed down into 6 commits:

- *Allow tests to pass when stderr is non-empty* is #10703 
- *MOVEONLY move CAffectedKeysVisitor* is a code move
- *Add ReserveKey struct and return it from GetAllReserveKeys()* is a slight refactor of existing code
- *Add keypoolcritical* adds the node shutdown functionality
- *Add keypoolmin* adds the don't-advance-wallet-best-block functionality
- *add keypool restore functional test* adds tests

Ready for wider review.

jnewbery:
Locking issue has been resolved. This should now pass Travis.

I've added a new commit *don't hold cs_LastBlockFile while calling setBestChain*

src/wallet/wallet.cpp
```diff
@@ -3568,15 +3596,50 @@ void CReserveKey::ReturnKey()
...
+        if (!walletdb.ReadPool(index, keypool))
             throw std::runtime_error(std::string(__func__) + ": read failed");
         assert(keypool.vchPubKey.IsValid());
         CKeyID keyID = keypool.vchPubKey.GetID();
-        mapKeyPool[keyID] = ReserveKey {id, keypool.fInternal};
+        mapKeyPool[keyID] = ReserveKey {index, keypool.fInternal};
+    }
+}
+
+void CWallet::MarkReserveKeysAsUsed(const ReserveKey reserve_key)
```
ryanofsky: In commit "[wallet] Add keypoolcritical"

Maybe add & if this isn't an intentional optimization, it's a little more conventional in c++ to pass structs by reference instead of value to avoid potentially slow copies.

src/wallet/wallet.cpp
```diff
@@ -3568,15 +3596,50 @@ void CReserveKey::ReturnKey()
...
+    for (const int64_t& index : setKeyPool)
     {
         CKeyPool keypool;
-        if (!walletdb.ReadPool(id, keypool))
+        if (!walletdb.ReadPool(index, keypool))
             throw std::runtime_error(std::string(__func__) + ": read failed");
         assert(keypool.vchPubKey.IsValid());
         CKeyID keyID = keypool.vchPubKey.GetID();
-        mapKeyPool[keyID] = ReserveKey {id, keypool.fInternal};
+        mapKeyPool[keyID] = ReserveKey {index, keypool.fInternal};
```
ryanofsky: In commit "[wallet] Add keypoolcritical"

Could make PR a little smaller by moving these changes to previous commit.

ryanofsky (commented):
utACK 91f3fb95d9e197a3f7758e1a4f37df003affd28f. Changes since last review: new lock in SetBestChain and unlock in FlushStateToDisk, changed MarkReserveKeysAsUsed arguments, changed LoadReserveKeysToMap error string

I also posted a comment in https://github.com/bitcoin/bitcoin/pull/10238#issuecomment-317817700 about potential future optimizations.

ryanofsky:
@jnewbery, can you explain the reason for dependency on #10703? It does seem useful to have travis continue checking for unexpected error messages (see https://botbot.me/freenode/bitcoin-core-dev/msg/89036779/)

MarcoFalke:
After discussion on irc:

Would it be possible to pipe the stderr into a SpooledTempFile instead? This way one can check that the error string matches.
The issue with the current approach is that it is active for all tests, even though none of the other tests needs it. This might hide serious bugs by accident.

Finally, please remove the execute bit from util.py

jnewbery:
Reworked with stderr capture/checking done within the test.

Also addressed @MarcoFalke's nit and rebased.

test/functional/keypool-restore.py
```diff
@@ -0,0 +1,121 @@
...
+            with tempfile.SpooledTemporaryFile(max_size=2**16) as log_stderr:
+                self.nodes[1] = self.start_node(1, self.tmpdir, self.extra_args[1], stderr=log_stderr)
+                connect_nodes(self.nodes[0], 1)
+
+                # node1 should shutdown because it can't topup its keypool
+                self.bitcoind_processes[1].wait(2)
+
+                log_stderr.seek(0)
+                stderr = log_stderr.read().decode('utf-8')
+                assert "Number of keys in keypool is below critical minimum and the wallet is encrytped. Bitcoin will now shutdown to avoid loss of funds." in stderr
```
laanwj: "encrytped"

sipa (commented):
utACK ec59b2eaaccd4d42d98345e9719265f102494e46

test/functional/keypool-restore.py
jnewbery: fixed

src/wallet/wallet.cpp
```diff
@@ -1038,6 +1039,33 @@ bool CWallet::AddToWalletIfInvolvingMe(const CTransactionRef& ptx, const CBlockI
...
         {
+            /* Check if any keys in the wallet keypool that were supposed to be unused
+             * have appeared in a new transaction. If so, remove those keys from the keypool.
+             * This can happen when restoring an old wallet backup that does not contain
+             * the mostly recently created transactions from newer versions of the wallet.
+             */
+            std::map<CKeyID, ReserveKey> keyPool = GetAllReserveKeys();
+
+            // loop though all outputs
+            for(const CTxOut& txout: tx.vout) {
```
jnewbery: fixed

TheBlueMatt:
As it is this looks to me to be a major performance regression due to calling GetAllReserveKeys in every matching AddToWalletIfInvolvingMe. #10238 is a relatively simple patch to avoid the massive disk churn, but would need rebase. Alternatively, it would be an equivalently simple change to just keep the reserved keys map created in GetAllReserveKeys in memory, and probably even remove more lines than it adds.

ryanofsky:
> As it is this looks to me to be a major performance regression due to calling GetAllReserveKeys in every matching AddToWalletIfInvolvingMe. #10238 is a relatively simple patch to avoid the massive disk churn, but would need rebase. Alternatively, it would be an equivalently simple change to just keep the reserved keys map created in GetAllReserveKeys in memory, and probably even remove more lines than it adds.

https://github.com/ryanofsky/bitcoin/commit/79e2411e6fcc544326a02de2b0b5756c17c3531f is a straightforward, though not very good-looking way of avoiding GetAllReserveKeys calls on each matching transaction. Probably a nicer approach would be to keep an updated map in memory like you suggest, but I wanted to post this just in case the other approach leads to bugs or delays.

src/wallet/wallet.h
```diff
@@ -82,6 +82,7 @@ class CTxMemPool;
 class CBlockPolicyEstimator;
 class CWalletTx;
 struct FeeCalculation;
+struct ReserveKey { int64_t index; bool internal; };
```
morcos: could you rename this to be something that isn't so easily confused with `CReserveKey`

morcos (commented):
This seems a bit like a hack, but what about instead of the sequence of operations currently recommended:
```  
" - restart bitcoin with -keypoolcritical set to 0\n"
" - unlock the wallet using walletpassphrase\n"
" - restart with -rescan. This will redownload the blockchain if you are running "
```

We instead changed the check for updating SetBestBlock to recognize the condition of the first restart and continue to not update the best block in the wallet.  This could detect a separate command line argument such as `-bypasskeypoolcritical` which is used in that check and the shutdown check.

Then the second restart would not require a rescan.

So instructions are:
```
" - restart bitcoin with -bypasskeypoolcritical\n"
" - unlock the wallet using walletpassphrase\n"
" - restart without -bypasskeypoolcritical. This might redownload the blockchain if you are running"
``` 

Suppose there is still a pruning edge case?

TheBlueMatt:
I did https://github.com/TheBlueMatt/bitcoin/commits/2017-07-keypool-topup-no-perf-regression to fix the performance regression, it looks fine to me (note minor changes in "[wallet] Add keypoolcritical" which is a bit slower - now must read keypool before erasing, but it shouldn't be nearly as bad and #10238 should fix it) and is really pretty simple.

ryanofsky:
> We instead changed the check for updating SetBestBlock to recognize the condition of the first restart and continue to not update the best block in the wallet. This could detect a separate command line argument such as -bypasskeypoolcritical which is used in that check and the shutdown check.

Maybe we could avoid the need for a -bypasskeypoolcritical command line option with a CWallet "bool m_update_best_block" member that starts off true, gets set to false whenever the number of keys falls below GetArg("-keypoolmin"), and only gets set to back to true when there is a rescan from at or before the best block while there are more than -keypoolmin keys.

sipa:
> As it is this looks to me to be a major performance regression due to calling GetAllReserveKeys in every matching AddToWalletIfInvolvingMe. #10238 is a relatively simple patch to avoid the massive disk churn, but would need rebase.

Ouch, I wasn't aware that `GetAllReserveKeys` read all the keys from disk. #10283 looks like the right solution, and shouldn't be hard to rebase?

TheBlueMatt:
I don't believe #10238 fixes the issue directly (doesn't provide the bidirectional mapping, I was incorrect), but is also a nice cleanup.

src/wallet/wallet.h
jnewbery: any suggestions? Is `ReserveKeyIndex` better? `ReserveKeyIndexChain`?

jnewbery:
Thanks @TheBlueMatt . I've force pushed your branch here.

@morcos - I've added your suggested `bypasskeypoolcritical` option. Can you take a look at the *[wallet] add bypasskeypoolcritical* commit and see if it captures what you were suggesting accurately?

@ryanofsky - I think your solution would involve storing a new key in the wallet DB, since at the moment the only way to rescan is to restart the node. Am I understanding correctly?

ryanofsky:
> @ryanofsky - I think your solution would involve storing a new key in the wallet DB, since at the moment the only way to rescan is to restart the node. Am I understanding correctly?

No, you do not need to need to add anything new to the database, best block is all you need. I included the part about setting `m_update_best_block` back to true again when there's a rescan with sufficient keys just because in the future we will have good rescan RPC. For now, the only way to get `m_update_best_block` set back to true would be to restart like the instructions say (or import a key in some clever way to trigger a rescan).

src/wallet/wallet.cpp
```diff
@@ -1006,6 +1046,33 @@ bool CWallet::AddToWalletIfInvolvingMe(const CTransactionRef& ptx, const CBlockI
         if (fExisted && !fUpdate) return false;
         if (fExisted || IsMine(tx) || IsFromMe(tx))
         {
+            /* Check if any keys in the wallet keypool that were supposed to be unused
+             * have appeared in a new transaction. If so, remove those keys from the keypool.
+             * This can happen when restoring an old wallet backup that does not contain
+             * the mostly recently created transactions from newer versions of the wallet.
+             */
+            std::map<CKeyID, int64_t> keyPool = GetAllReserveKeys();
```
morcos: Access the member directly to avoid a copy?

Also this will avoid running out of the keypool inside of one scriptPubKey :)

src/wallet/wallet.cpp
```diff
@@ -3530,38 +3624,53 @@ void CReserveKey::KeepKey()
...
+        LogPrintf("%s: Number of keys in keypool is below critical minimum. Shutting down. internal keypool: %d, external keypool: %d, keypool minimum: %d\n",
+                  __func__, setInternalKeyPool.size(), setExternalKeyPool.size(), keypool_critical);
+        const std::string error_msg = "Number of keys in keypool is below critical minimum and the wallet is encrypted. Bitcoin will now shutdown to avoid loss of funds.\n"
+                                      "This is probably because you are restoring an old backup wallet file which has not been topped up with the most recently "
+                                      "derived keys, and so you would not detect transactions involving those keys.\n"
+                                      "You can manually top-up your wallet keypool as follows:\n"
+                                      " - restart bitcoin with -bypasskeypoolcritical\n"
+                                      " - unlock the wallet using walletpassphrase\n"
+                                      " - restart without -bypasskeypoolcritical. This may redownload the blockchain if you are running a pruned node.";
```
morcos: Maybe:  "- restart without -bypasskeypoolcritical." followed by:
"NOTE: if you have a pruned node that prunes blocks your wallet hasn't scanned yet when restarting with -bypasskeypoolcritical then you may need to do a complete reindex.  Consider raising the prune limit temporarily for both restarts to avoid this."

morcos (approved):
This is looking pretty good to me.

src/wallet/wallet.cpp
jnewbery: I believe Matt's commit 64dc2dddf52a48c9cef158951f46395683a26595 resolves your concern about copying. `GetAllReserveKeys()` returns a const reference 

I don't understand your comment about running out of the keypool. Is it still an issue?

src/wallet/wallet.cpp
jnewbery: Yes, makes sense. Changed

jnewbery:
@TheBlueMatt - I don't fully understand the implications of `m_pool_key_to_id` and returning a const references to the internal cache in `GetAllReserveKeys()`. For example, Cory has suggested that this might not be thread-safe, but perhaps that's ok in wallet code?

Reviewers who have previously ACKed this PR - please be aware of the additional commits.

morcos:
@jnewbery I just kind of assumed it was protected by cs_wallet, but we should verify.  

No my concern still exists.  It returns a const reference but then you immediately copy it to your local variable.  Also in that same code path you call TopUpKeyPool so now there will be more reserved keys but the next time you lookup you are still looking up in your stale copy

src/wallet/wallet.cpp
morcos: No my concern still exists. It returns a const reference but then you immediately copy it to your local variable. Also in that same code path you call TopUpKeyPool so now there will be more reserved keys but the next time you lookup you are still looking up in your stale copy

TheBlueMatt:
@jnewbery heh, yea, I just generally always assume everything in wallet takes LOCK2(cs_main, cs_wallet). The GetAllReserveKeys call in rpcdump as well as AddToWalletIfInvolvingMe both take/require cs_wallet, so it looks fine to me?

jnewbery:
@morcos - please look at commit *fixup: don't make local copy of m_pool_key_to_id*. I think that addresses your point above.

@TheBlueMatt - yes, all access to `m_pool_key_to_id` is protected by cs_wallet. The only place that isn't explicit is in `LoadKeyPool()`. I've added an `AssertLockHeld()` in *fixup: explicitly check lock is held in LoadKeyPool*.

I'll squash both commits if you're happy with them.

morcos:
@jnewbery looks good i think!

src/wallet/wallet.cpp
```diff
@@ -3595,6 +3623,48 @@ void CReserveKey::ReturnKey()
...
+        const std::string error_msg = "Number of keys in keypool is below critical minimum and the wallet is encrypted. Bitcoin will now shutdown to avoid loss of funds.\n"
+                                      "This is probably because you are restoring an old backup wallet file which has not been topped up with the most recently "
+                                      "derived keys, and so you would not detect transactions involving those keys.\n"
+                                      "You can manually top-up your wallet keypool as follows:\n"
+                                      " - restart bitcoin with -keypoolcritical set to 0\n"
+                                      " - unlock the wallet using walletpassphrase\n"
+                                      " - restart with -rescan. This will redownload the blockchain if you are running a pruned node.";
+        uiInterface.ThreadSafeMessageBox(error_msg, "", CClientUIInterface::MSG_ERROR);
+        StartShutdown();
+        throw std::runtime_error(error_msg);
```
TheBlueMatt: Wait, why is this safe? I'm both highly skeptical that throwing here doesn't blow up all kinds of potential callers and that this is neccessary. Better to just call StartShutdown() and then avoid writing wallet best block.

src/wallet/wallet.cpp
```diff
@@ -3968,6 +4039,25 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
...
+        unsigned int keypool_size = GetArg("-keypool", DEFAULT_KEYPOOL_SIZE);
+        unsigned int keypool_critical = GetArg("-keypoolcritical", DEFAULT_KEYPOOL_CRITICAL);
+        if (walletInstance->IsCrypted()) {
+            if (keypool_size < keypool_critical) {
+                LogPrintf("Parameter Interaction: keypool size (%d) must be larger than keypool critical size for encrypted wallets (%d)\n", keypool_size, keypool_critical);
+                ForceSetArg("-keypool", std::to_string(keypool_critical));
+            }
+        } else {
+            if (keypool_size < keypool_critical) {
+                InitWarning(strprintf(_("Your keypool is configured to store %d keys, which is below the keypool critical size of %d. Using a larger keypool will make it less "
```
TheBlueMatt: I think this needs to be more than a warning, no? There is no check in ShutdownIfKeypoolCritical to only exit if we failed to top up (ie if we hit this warning, the first time we see ShutdownIfKeypoolCritical we will exit, no?

src/wallet/wallet.cpp
```diff
@@ -3922,6 +3927,7 @@ std::string CWallet::GetWalletHelpString(bool showDebug)
     {
         strUsage += HelpMessageGroup(_("Wallet debugging/testing options:"));
 
+        strUsage += HelpMessageOpt("-bypasskeypoolcritical", _("Bypass keypool critical limit. Don't shutdown the node if keypool drops below keypoolcritical limit (but don't advance BestBlock)"));
```
TheBlueMatt: The doc is wrong (we do advance the best block), but I prefer the doc version, so maybe just remove the check in SetBestChain.

morcos: huh? @TheBlueMatt

src/wallet/wallet.cpp
morcos: agreed, or we need the parameter interaction to increase keypool to keypool_critical? or something?

src/wallet/wallet.cpp
TheBlueMatt: Oh, guess the doc is right, but just super confusing (double negative-ish, but not really).Maybe "but *still* don't advance"?

src/wallet/wallet.cpp
TheBlueMatt: Yea, think I'd vote for a paramter interaction.

jnewbery: If we hit this warning, then the wallet is not encrypted, and `TopUpKeyPool()` should succeed.

If the wallet is encrypted we'll hit the branch above and we force set keypool to keypool_critical.

jnewbery: oh, I see. Yes, I'll change the parameter interaction to set keypool to keypool_critical for all wallets.

jnewbery:
@TheBlueMatt I've pushed a fixup commit that addresses your two comments. If you're happy I'll squash.

jnewbery:
@TheBlueMatt commits squashed. Ready for final review.

morcos:
utACK 7cca786

jnewbery:
Unrebased at @morcos's request. Presquahed branch is here: https://github.com/jnewbery/bitcoin/releases/tag/pr10882.4

jnewbery:
rebased on master. Don't ask

TheBlueMatt:
utACK 85079e6ab3e7d0596c010df22256ccb3c2a9be03 (minus the test bits)

src/wallet/wallet.cpp
```diff
@@ -3984,6 +4053,19 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
 
     RegisterValidationInterface(walletInstance);
 
+    // Make sure we have enough keys in our keypool if HD is enabled
+    if (walletInstance->IsHDEnabled()) {
+        unsigned int keypool_size = GetArg("-keypool", DEFAULT_KEYPOOL_SIZE);
+        unsigned int keypool_critical = GetArg("-keypoolcritical", DEFAULT_KEYPOOL_CRITICAL);
+        if (keypool_size < keypool_critical) {
```
instagibbs: nit: current logic is "larger or equal to"

instagibbs:
utACK https://github.com/bitcoin/bitcoin/pull/10882/commits/6b2a63828ed8d36ae8ce45b3d2ccb18645ab03ab

gmaxwell:
I'm still not clear on what prevent a forced shutdown under this pattern: start a new node, sync,  encrypt wallet,  run getnewaddress 501 times, get a payment.    Where is it distinguishing the keypool going under critical while behind vs while current?

jnewbery:
@gmaxwell - currently `getnewaddress` will fail if the keypool has run out. How about we change it so that it fails if it hits the `keypool_critical` limit?

gmaxwell:
@jnewbery That seems reasonable to me!

jnewbery:
I've added three new commits:

1. *add HasUnusedKeys() helper* is just a refactor
2. *return error from getnewaddress if keypool drops to critical* prevents `getnewaddress` from returning a new address if the keypool drops to critical
3. *add checks for keypool critical* does the same for the other places that `GetKeyFromPool()` is called

I'm not sure if we want (3). It can be omitted if not required.

src/wallet/wallet.cpp
```diff
@@ -417,6 +417,12 @@ bool CWallet::ChangeWalletPassphrase(const SecureString& strOldWalletPassphrase,
 
 void CWallet::SetBestChain(const CBlockLocator& loc)
 {
+    LOCK(cs_wallet); //nWalletMaxVersion
+    unsigned int keypool_min = GetArg("-keypoolmin", DEFAULT_KEYPOOL_MIN);
+    if (IsHDEnabled() && ((CanSupportFeature(FEATURE_HD_SPLIT) && setInternalKeyPool.size() < keypool_min) || (setExternalKeyPool.size() < keypool_min))) {
```
ryanofsky: In commit "[wallet] Add keypoolmin"

As alluded to in https://github.com/bitcoin/bitcoin/pull/10882#issuecomment-318182682 and https://github.com/bitcoin/bitcoin/pull/10882#issuecomment-318322124, this check is not as safe as it appears, because if keypool gets drained below the minimum, then is refilled later after unlocking, there might be missing transactions which will not get rescanned in the interval of time where the keypool was below the minimum. I would at least add a comment here mentioning this, if you don't want to add an `m_update_best_block` bool to fix it.

By the way, another advantage of adding a `m_update_best_block` bool is that it would let us print a log message when the keypool falls below the minimum (right now there is no indication, and we probably don't want to add a log statement here since it would spam the log).

ryanofsky (commented):
utACK 85079e6ab3e7d0596c010df22256ccb3c2a9be03. This seems fine to merge. I made some suggestions but they could be addressed separately.

src/qt/addresstablemodel.cpp
```diff
@@ -369,17 +370,14 @@ QString AddressTableModel::addRow(const QString &type, const QString &label, con
     {
         // Generate a new address to associate with given label
         CPubKey newKey;
-        if(!wallet->GetKeyFromPool(newKey))
-        {
+        if (!wallet->GetKeyFromPool(newKey) || (wallet->IsHDEnabled() && !wallet->HasUnusedKeys(GetArg("-keypoolcritical", DEFAULT_KEYPOOL_CRITICAL)))) {
```
ryanofsky: In commit "[wallet] add checks for keypool critical"

It seems like a less repetitive (and safer?) approach would be to make GetKeyFromPool return false when number of keys is below critical, and maybe have an override if needed.

src/wallet/rpcwallet.cpp
```diff
@@ -161,8 +161,8 @@ UniValue getnewaddress(const JSONRPCRequest& request)
 
     // Generate a new key that is added to wallet
     CPubKey newKey;
-    if (!pwallet->GetKeyFromPool(newKey)) {
-        throw JSONRPCError(RPC_WALLET_KEYPOOL_RAN_OUT, "Error: Keypool ran out, please call keypoolrefill first");
+    if (!pwallet->GetKeyFromPool(newKey) || (pwallet->IsHDEnabled() && !pwallet->HasUnusedKeys(GetArg("-keypoolcritical", DEFAULT_KEYPOOL_CRITICAL)))) {
```
ryanofsky: In commit "[wallet] return error from getnewaddress if keypool drops to critical"

Should keypool size be checked before calling GetKeyFromPool instead of after, to avoid getting a key and throwing it away?

ryanofsky: In commit "[wallet] return error from getnewaddress if keypool drops to critical"

Maybe compare against -keypoolmin instead of -keypoolcritical to be more conservative, and because otherwise this change will have no effect (because the node will be shut down) unless `-bypasskeypoolcritical` happens to be in use. Also I think it's good just for sake of simplicity if `-bypasskeypoolcritical` and `-keypoolcritical=0` are synonyms for the same thing. Here you are introducing a subtle difference.

Either way, `-keypoolmin` or `-keypoolcritical` documentation could be updated to mention this new behavior, too.

ryanofsky (commented):
utACK 98a1e38a658e6c500b0cb9f979118f8f21733aa0. Only change since last review is GetKeyFromPool being disallowed on keypool critical condition.

ryanofsky:
> I'm not sure if we want (3). It can be omitted if not required.

Seems good to me to disallow GetKeyFromPool() generally like this is doing. `getrawchangeaddress` RPC is another thing to consider disabling.

src/wallet/wallet.h
```diff
@@ -973,10 +974,10 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
     bool TopUpKeyPool(unsigned int kpSize = 0);
     void ReserveKeyFromKeyPool(int64_t& nIndex, CKeyPool& keypool, bool fRequestedInternal);
     void KeepKey(int64_t nIndex);
-    void ReturnKey(int64_t nIndex, bool fInternal);
+    void ReturnKey(int64_t nIndex, bool fInternal, const CPubKey& pubkey);
     bool GetKeyFromPool(CPubKey &key, bool internal = false);
     int64_t GetOldestKeyPoolTime();
-    void GetAllReserveKeys(std::set<CKeyID>& setAddress) const;
+    const std::map<CKeyID, int64_t>& GetAllReserveKeys() const { return m_pool_key_to_id; }
```
jnewbery: agree - changed

src/wallet/wallet.cpp
jnewbery: yes - fixed

src/wallet/wallet.cpp
```diff
@@ -3611,6 +3638,47 @@ void CReserveKey::ReturnKey()
...
+    if (IsHDEnabled() && ((CanSupportFeature(FEATURE_HD_SPLIT) && setInternalKeyPool.size() < keypool_critical) || (setExternalKeyPool.size() < keypool_critical))) {
+        // if the remaining keypool size is below the gap limit, shutdown
+        LogPrintf("%s: Number of keys in keypool is below critical minimum. Shutting down. internal keypool: %d, external keypool: %d, keypool minimum: %d\n",
+                  __func__, setInternalKeyPool.size(), setExternalKeyPool.size(), keypool_critical);
+        const std::string error_msg = "Number of keys in keypool is below critical minimum and the wallet is encrypted. Bitcoin will now shutdown to avoid loss of funds.\n"
+                                      "This is probably because you are restoring an old backup wallet file which has not been topped up with the most recently "
+                                      "derived keys, and so you would not detect transactions involving those keys.\n"
+                                      "You can manually top-up your wallet keypool as follows:\n"
+                                      " - restart bitcoin with -keypoolcritical set to 0\n"
```
jnewbery: added

src/wallet/wallet.cpp
```diff
@@ -3611,6 +3638,47 @@ void CReserveKey::ReturnKey()
...
+                                      " - unlock the wallet using walletpassphrase\n"
```
jnewbery: added

src/wallet/wallet.cpp
jnewbery: Ah, your allusion was a bit subtle for me. I thought your original comment was "this is an alternative way to do it" rather than "the current way is unsafe. You should do this instead"

Can you take a look at https://github.com/jnewbery/bitcoin/commit/145a06521bd458e214cd9c8d163c9a1ebbd63189 . Is that what you had in mind?

With that change, I think we can remove `bypasskeypoolcritical` and update the recovery instructions to say `set -keypool_critical=0`. Do you agree?

ryanofsky: Thread https://github.com/bitcoin/bitcoin/pull/10882#discussion_r131245392

Posted a question in that commit, but I think this seems basically right. Few things

- I don't think this change necessarily needs to be part of this PR if it will delay the merge. Could be done as a followup.
- It seems like it would be a little safer (though maybe equivalent) and more logical to change `m_update_best_block` from true to false in `AddToWalletIfInvolvingMe` between `TopUpKeyPool` and `ShutdownIfKeypoolCritical`, since similar checks are being done there, and it would make it obvious that if there's any dip in keypool size while transactions are being processed, the flag will flip.
- When I talked about this with Alex he wanted to keep `-bypasskeypoolcritical` even if `-keypool_critical=0` can be made safe & equivalent, because it makes purpose of the flag more explicit.
- Would be good to use the HasUnusedKeys helper to simplify the logic.

jnewbery:
Pushed a new branch, which I think should address all of @ryanofsky's review comments.

Old branch here for comparison: https://github.com/jnewbery/bitcoin/releases/tag/pr10882.5

src/qt/addresstablemodel.cpp
jnewbery: ok, done

src/wallet/rpcwallet.cpp
jnewbery: I originally wasn't doing this because `GetKeyFromPool()` calls into `ReserveKeyFromKeyPool()` which tops up the keypool. I didn't want to fail `GetKeyFromPool()` when the wallet is unlocked and able to top up the keypool.

I've now added a `TopUpKeyPool()` call in `GetKeyFromPool()`, so we try to top up before checking that the keypool hasn't fallen to the critical level.

jnewbery: Changed the check to `HasUnusedKeys(GetArg("-keypoolcritical", DEFAULT_KEYPOOL_CRITICAL) + 1)`, which will prevent us from dropping to the keypool_critical threshold.

src/wallet/wallet.cpp
jnewbery: Done (in fact I've added the change to  `m_update_best_block` false in `AddToWalletIfInvolvingMe()` and `SetBestChain()`)

test/functional/keypool-restore.py
```diff
@@ -0,0 +1,121 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Test HD Wallet keypool restore function.
+
+Two nodes. Node1 is under test. Node0 is providing transactions and generating blocks.
+
+Repeat test twice, once with encrypted wallet and once with unencrypted wallet:
```
jnewbery: I think this can be added later.

laanwj:
Various issues with Travis have appeared after the last push.
(I've tried restarting the build and it didn't change anything)

ryanofsky:
I think this should just be rolled back to [`85079e6`](https://github.com/bitcoin/bitcoin/commits/85079e6ab3e7d0596c010df22256ccb3c2a9be03), which got a bunch of ACKs, and was straightforward, and did the job. The improvements that have been made since then are nice, but fairly minor, and backwards compatible, so they could be done in a separate PR.

laanwj:
@ryanofsky Agree. In any case, we'll need to do *something* here today if this is still to make 0.15.0rc1.

jonasschnelli:
Agree with @ryanofsky.
utACK the older https://github.com/bitcoin/bitcoin/commit/85079e6ab3e7d0596c010df22256ccb3c2a9be03

jnewbery:
ok, reverted to https://github.com/bitcoin/bitcoin/commit/85079e6ab3e7d0596c010df22256ccb3c2a9be03

Other branch is here: https://github.com/jnewbery/bitcoin/releases/tag/pr10882.6 . In that branch I fixed one problem with travis (I accidentally removed the `bypasscritical` help text in the rebase). Other problem was `import-rescan.py` failing, which I wasn't able to fix.

ryanofsky:
Seems like this can be merged now.

ryanofsky:
From discussion in IRC, this isn't going to be merged as is because it will caused bad upgrade experience (immediately shutting down) for older locked HD wallets with less than 500 keys. John is implementing a solution "remove [ShutdownIfKeypoolCritical call] from CreateWalletFromFile(), and only call it from AddToWalletIfInvolvingMe() if that function was called by SyncTransaction() (and not by ScanForWalletTransactions())" https://botbot.me/freenode/bitcoin-core-dev/msg/89554129/.

This change seems fine to me, though I wonder if it won't just delay the shutdown instead of causing an immediate shutdown?

Another way to solve this problem to would be delay `-keypoolcritical` enforcement on an older database until it is unlocked. E.g. add a new `ENFORCE_KEYPOOL_CRITICAL` flag to the database, and only enforce the critical minimum when it is present. If not present, add the flag to the database when the wallet is unlocked (or on startup if the wallet is unencrypted).

jnewbery:
re-ordered commits. The only difference with the new branch is renaming keypool-restore.py to keypool-topup.py, and calling `TopUpKeyPool()` on wallet load for all wallets (not just HD wallets). Old branch is here: https://github.com/jnewbery/bitcoin/releases/tag/pr10882.7

jnewbery:
As requested by @gmaxwell's I've pulled out the basic keypool mark-used/topup functionality into #11022. This PR is a superset of that and also contains the keypool_min/keypool_critical functionality to stop the node/prevent best block from advancing if the keypool drops below a threshold.

ryanofsky:
Note to reviewers, this is now based on #11022, so changes in commits before "Addkeypoolmin" and "Add keypoolcritical" can be discussed there.

src/wallet/wallet.cpp
```diff
@@ -1644,6 +1644,10 @@ CBlockIndex* CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, bool f
 
         fScanningWallet = false;
     }
+
+    // Check that we haven't dropped below the keypool_critical threshold.
+    ShutdownIfKeypoolCritical();
```
ryanofsky: In commit "[wallet] Add keypoolcritical"

I'm confused by this. It seems like this is the only call to ShutdownIfKeypoolCritical() in the whole PR, and it's in the rescan function, so the node will no longer shut down if the wallet is locked and the keypool gradually drains, unless the user manually triggers a rescan or restarts the node? If this is behavior that's intended, it should be documented in the -keypoolcritical help. It might also be good to update the PR description (which is currently out of date) to list the changes here and say what problems are fixed and not fixed by them.

jnewbery: Yes, that's the current implementation:

`07:58 < gmaxwell> jnewbery: Can we distinguish the case where we are current vs rescanning? If so, then we just shouldn't perform the shutdown when we're current.`

(https://botbot.me/freenode/bitcoin-core-dev/2017-08-08/?msg=89552785&page=2)

We'll need to discuss this in the meeting today. There doesn't seem to be agreement about what this PR should even be trying to do.

src/wallet/rpcwallet.cpp
```diff
@@ -161,8 +161,8 @@ UniValue getnewaddress(const JSONRPCRequest& request)
 
     // Generate a new key that is added to wallet
     CPubKey newKey;
-    if (!pwallet->GetKeyFromPool(newKey)) {
-        throw JSONRPCError(RPC_WALLET_KEYPOOL_RAN_OUT, "Error: Keypool ran out, please call keypoolrefill first");
+    if (!pwallet->GetKeyFromPool(newKey, false, true)) {
```
ryanofsky: In commit "[wallet] Don't allow getnewaddress to drop keypool to critical."

Do you want to make the same change to getrawchangeaddress? (as mentioned https://github.com/bitcoin/bitcoin/pull/10882#issuecomment-320083591)

MarcoFalke:
Cleared the 0.15.0 milestone for now. Let's focus on #11022 first.

src/wallet/wallet.cpp
ryanofsky: Thread https://github.com/bitcoin/bitcoin/pull/10882#discussion_r132535000

That's fine, but it seems that current -keypoolcritical help description is pretty misleading. It would be good to update it, as well as the PR description.

src/wallet/wallet.cpp
ryanofsky: Thread https://github.com/bitcoin/bitcoin/pull/10882#discussion_r131033130

I don't see a change, should log print just be changed to "smaller or equal to"?

src/wallet/wallet.cpp
```diff
@@ -1651,6 +1657,23 @@ CBlockIndex* CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, bool f
...
+    if (IsHDEnabled() && pindexStart && pindexStart->nHeight <= pindexBestBlock->nHeight) {
+        m_update_best_block = HasUnusedKeys(GetArg("-keypoolmin", DEFAULT_KEYPOOL_MIN));
+    }
+
+    // If the keypool has dropped below -keypoolmin, then stop updating the bestblock height. We can rescan later once the wallet is unlocked.
+    unsigned int keypool_min = GetArg("-keypoolmin", DEFAULT_KEYPOOL_MIN);
+    if (IsHDEnabled() && !HasUnusedKeys(keypool_min) && m_update_best_block) {
+        LogPrintf("Keypool has fallen below keypool_min (%s). Wallet will no longer watch for new transactions and best block height will not be advanced.\n", keypool_min);
+        LogPrintf("Unlock wallet, top up keypool and rescan to resume watching for new transactions.\n");
+        m_update_best_block = false;
```
ryanofsky: In commit "[wallet] Add m_update_best_block"

As mentioned https://github.com/bitcoin/bitcoin/pull/10882#discussion_r131463645, I think the transition from true -> false m_update_best_block should just happen after the topup in AddToWalletIfInvolvingMe. So this logic, and the logic in SetBestChain could be removed. SetBestChain would only read m_update_best_block, not write to it or interact with the keypool.

jnewbery:
rebased on master and updated PR notes. Shouldn't be merged without further work to make sure this is safe.

@gmaxwell - could you comment here on the edge cases here (for future reference).

jnewbery:
Closing for now. There were several edge-cases discovered that meant that this approach needs a bit more careful consideration. @gmaxwell is going to add some notes to this PR so if we do pick this later, we don't need to rediscover the edge-cases.

droark:
@jnewbery - Just happened to come across this PR while going through some commits. Did you ever get the edge case notes?

jnewbery:
I'm not sure if Greg collected them together. You can probably find them if you scrape through the IRC logs.

Issue #10269, opened 2017-04-24
Windows build appears broken on WSL (buntu okay)


AllanDoensen:
Get the latest WSL on windows 10.
clone a clean copy of core.
This is on Windows 10. Ubuntu looks okay.

allan@DESKTOP-U9EV6QE:~$ git clone https://github.com/Bitcoin/Bitcoin.git
Cloning into 'Bitcoin'...
remote: Counting objects: 91026, done.
remote: Compressing objects: 100% (3/3), done.
remote: Total 91026 (delta 5), reused 4 (delta 4), pack-reused 91019
Receiving objects: 100% (91026/91026), 79.40 MiB | 137.00 KiB/s, done.
Resolving deltas: 100% (67842/67842), done.
Checking connectivity... done.
Checking out files: 100% (1491/1491), done.
allan@DESKTOP-U9EV6QE:~$ cd
Bitcoin/          .cache/           core/             Desktop/          .subversion/
BitcoinUnlimited/ .ccache/          .dbus/            .gconf/
bu2/              .config/          .dbus-keyrings/   .mozilla/
allan@DESKTOP-U9EV6QE:~$ cd Bitcoin
allan@DESKTOP-U9EV6QE:~/Bitcoin$ cd depends/
allan@DESKTOP-U9EV6QE:~/Bitcoin/depends$ make HOST=x86_64-w64-mingw32
Fetching ccache-3.3.3.tar.bz2 from https://samba.org/ftp/ccache
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   261  100   261    0     0     28      0  0:00:09  0:00:09 --:--:--    55
100  337k  100  337k    0     0  20729      0  0:00:16  0:00:16 --:--:-- 77660
/home/allan/Bitcoin/depends/work/download/native_ccache-3.3.3/ccache-3.3.3.tar.bz2.temp: OK
Extracting native_ccache...
/home/allan/Bitcoin/depends/sources/ccache-3.3.3.tar.bz2: OK
Preprocessing native_ccache...
Configuring native_ccache...
/bin/sh: 1: Syntax error: "(" unexpected
make: *** [/home/allan/Bitcoin/depends/work/build/x86_64-w64-mingw32/native_ccache/3.3.3-43ca3e6a03b/./.stamp_configured] Error 2
allan@DESKTOP-U9EV6QE:~/Bitcoin/depends$

[...]

PR #10884, opened 2017-07-19
Added fix for wrong PATH to build-windows.md

eliahuhorwitz:
Added fix for wrong PATH as seen here:
https://github.com/bitcoin/bitcoin/issues/10269

### DIFF
--- a/doc/build-windows.md
+++ b/doc/build-windows.md
@@ -69,6 +69,7 @@ Then build using:
 
     cd depends
     make HOST=x86_64-w64-mingw32
+    #if the above line fails run this "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games"
     cd ..
     ./autogen.sh # not required when building from tarball
     CONFIG_SITE=$PWD/depends/x86_64-w64-mingw32/share/config.site ./configure --prefix=/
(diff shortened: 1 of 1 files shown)


doc/build-windows.md
```diff
@@ -69,6 +69,7 @@ Then build using:
 
     cd depends
     make HOST=x86_64-w64-mingw32
+    #if the above line fails run this "PATH=/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/usr/games:/usr/local/games"
```
promag: Could we change the instructions to never change?

doc/build-windows.md
```diff
@@ -84,6 +85,7 @@ Then build using:
 
     cd depends
     make HOST=i686-w64-mingw32
+
```
promag: Unrelated?

promag (commented):
Missing explanation on PR description.

Nit, PR title and commit message should mention it's a fix. Suggestion: `Fix windows build instructions`.

doc/build-windows.md
eliahuhorwitz: Ta, my bad. Although I guess the 32 bit install will have the same problem so maybe add it here as well?

laanwj: That would be repeating the same thing twice, maybe it would be better to add a *warning* section describing the problem, instead of buried in the instructions to type? (and possibly refer to that here)

eliahuhorwitz: Great, ill do that later, there are a few more things I need to add

fanquake:
Closing in favour of #11244.

PR #10888, opened 2017-07-20
range-based loops and const qualifications in net.cpp

benma:
Plus a use of std::copy() instead of manual copying.

(The loop on line 117 is already done in #10493).

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -135,11 +135,10 @@ static std::vector<CAddress> convertSeed6(const std::vector<SeedSpec6> &vSeedsIn
     const int64_t nOneWeek = 7*24*60*60;
     std::vector<CAddress> vSeedsOut;
     vSeedsOut.reserve(vSeedsIn.size());
-    for (std::vector<SeedSpec6>::const_iterator i(vSeedsIn.begin()); i != vSeedsIn.end(); ++i)
-    {
+    for (const auto& seed_in : vSeedsIn) {
         struct in6_addr ip;
-        memcpy(&ip, i->addr, sizeof(ip));
-        CAddress addr(CService(ip, i->port), NODE_NETWORK);
+        memcpy(&ip, seed_in.addr, sizeof(ip));
+        CAddress addr(CService(ip, seed_in.port), NODE_NETWORK);
         addr.nTime = GetTime() - GetRand(nOneWeek) - nOneWeek;
         vSeedsOut.push_back(addr);
     }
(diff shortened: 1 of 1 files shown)


src/net.cpp
```diff
@@ -305,16 +305,16 @@ CNode* CConnman::FindNode(const CNetAddr& ip)
...
     return NULL;
 }
 
 CNode* CConnman::FindNode(const CSubNet& subNet)
 {
     LOCK(cs_vNodes);
     for (CNode* pnode : vNodes)
-    if (subNet.Match((CNetAddr)pnode->addr))
-        return (pnode);
+        if (subNet.Match((CNetAddr)pnode->addr))
```
jonasschnelli: Please switch to brackets here.

practicalswift:
Concept ACK (strong ACK! :-))

src/net.cpp
benma: Done.

dcousens (approved):

corebob:
How about "improving" those FindNode functions to something like this,

```
auto iter = std::find_if(vNodes.begin(), vNodes.end(), [&](CNode *node) {
        return node->GetAddrName() == addrName;
    });

    return iter == vNodes.end() ? NULL : *iter;
```

 I believe C++11 allows this

practicalswift:
@corebob I guess it is a matter of taste, but personally I find the current range-based for loop form more straightforward and easier to read:

```c++
     for (CNode* pnode : vNodes) {
          if (pnode->GetAddrName() == addrName) {
              return pnode;
          }
      }
      return NULL;
```

corebob:
@practicalswift 
I pretty much agree. I think pro arguments could be made for getting rid of a raw loop, and getting rid of an extra exit point. Though probably not worth it in this case.

benma:
I am usually in favor of using `algorithm`, as explicit names (`find_if`) and lambdas are more descriptive than loops etc., but in this case, I don't like the return value conversion, so it indeed looks more complicated than before. Too bad.

benma:
Rebased.

benma:
Can someone please retrigger CI?

sipa:
utACK a9144d535122a4d08348e477315b5a94368e6dfc

Travis passed after restart.

src/net.cpp
```diff
@@ -135,11 +135,11 @@ static std::vector<CAddress> convertSeed6(const std::vector<SeedSpec6> &vSeedsIn
     const int64_t nOneWeek = 7*24*60*60;
     std::vector<CAddress> vSeedsOut;
     vSeedsOut.reserve(vSeedsIn.size());
-    for (std::vector<SeedSpec6>::const_iterator i(vSeedsIn.begin()); i != vSeedsIn.end(); ++i)
+    for (const auto seed_in : vSeedsIn)
```
MarcoFalke: `const SeedSpec6& seed_in` makes more sense to me as it prevents a copy, no?

maflcko (commented):
utACK a9144d535.

Since this is refactoring only you also might want to use clang format on the changed lines. (See the developer notes for a one-liner to fixup the formatting)

src/net.cpp
benma: Correct. Done!

benma:
@MarcoFalke I also applied clang-format. Nice tool. Maybe it should go into the CI for all future changes ;)

MarcoFalke:
re-utACK 05cae8a

benma:
Can this be merged?

PR #10891, opened 2017-07-21
[RPC] Make ImportAddress works with segwit scriptPubKey

NicolasDorier:
`ImportAddress` rpc functions does not work for segwit `scirptPubKey`. Worst, it fails without error.

Reason is https://github.com/bitcoin/bitcoin/blob/041dad94b047313a17edb0324742cf80cfd550f1/src/wallet/rpcdump.cpp#L202 , if `ExtractDestination` fails, nothing is done.

### DIFF
--- a/src/script/standard.cpp
+++ b/src/script/standard.cpp
@@ -202,6 +202,16 @@ bool ExtractDestination(const CScript& scriptPubKey, CTxDestination& addressRet)
         addressRet = CScriptID(uint160(vSolutions[0]));
         return true;
     }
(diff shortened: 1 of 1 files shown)


NicolasDorier:
Closing for now, I suspect my problem comes from elsewhere, will reopen once I understand better the situation.

sipa:
Can you try whether #10788 helps?

NicolasDorier:
nop, does not help.

NicolasDorier:
trying to write a test for reproducing.

PR #10892, opened 2017-07-21
Replace traditional for with ranged for in block and transaction primitives

bytting:
Replace traditional for with ranged for in block and transaction primitives to improve readability

### DIFF
--- a/src/primitives/block.cpp
+++ b/src/primitives/block.cpp
@@ -25,9 +25,8 @@ std::string CBlock::ToString() const
         hashMerkleRoot.ToString(),
         nTime, nBits, nNonce,
         vtx.size());
-    for (unsigned int i = 0; i < vtx.size(); i++)
-    {
-        s << "  " << vtx[i]->ToString() << "\n";
+    for (const auto& tx : vtx) {
+        s << "  " << tx->ToString() << "\n";
     }
     return s.str();
 }
--- a/src/primitives/transaction.cpp
+++ b/src/primitives/transaction.cpp
@@ -83,10 +83,9 @@ CTransaction::CTransaction(CMutableTransaction &&tx) : nVersion(tx.nVersion), vi
 CAmount CTransaction::GetValueOut() const
 {
     CAmount nValueOut = 0;
-    for (std::vector<CTxOut>::const_iterator it(vout.begin()); it != vout.end(); ++it)
-    {
-        nValueOut += it->nValue;
-        if (!MoneyRange(it->nValue) || !MoneyRange(nValueOut))
+    for (const auto& tx_out : vout) {
+        nValueOut += tx_out.nValue;
+        if (!MoneyRange(tx_out.nValue) || !MoneyRange(nValueOut))
             throw std::runtime_error(std::string(__func__) + ": value out of range");
     }
     return nValueOut;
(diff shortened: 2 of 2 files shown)


dcousens (approved):

src/primitives/block.cpp
```diff
@@ -25,9 +25,8 @@ std::string CBlock::ToString() const
         hashMerkleRoot.ToString(),
         nTime, nBits, nNonce,
         vtx.size());
-    for (unsigned int i = 0; i < vtx.size(); i++)
-    {
-        s << "  " << vtx[i]->ToString() << "\n";
+    for (const auto &tx : vtx) {
```
benma: Nit: `const auto& tx` seems to be the normal way to format it.

corebob: Yeah, I was unsure about that. It seems counter intuitive to me since it is the variable that is a reference, not the type.

Anyway, I'll look into it so that it is consistent with the rest of the core base.

Thanks.

benma: I did check, `auto&` is much more common in the codebase (60 times vs. sth. like 5 times).

> It seems counter intuitive to me since it is the variable that is a reference, not the type.

It's actually the type. It's apparent if you look at function signatures.

`void foo(const std::string&);` is a valid function signature (first argument is a reference type).

corebob: Hm, when I look at your last patch #10916 you actually do the same.

> first argument is a reference type

I would call that an unnamed reference to type std::string :trollface:

I guess it's not that bad for arguments and return types though, I'll look into it.

benma: > Hm, when I look at your last patch #10916 you actually do the same.

You mean using `const auto& var`, not `const auto &var`? Makes sense that I use what I advocate :)

corebob: > for (const CKeyID &keyid : GetKeys()) {

benma: Good catch! In my defense, that line was there like that before and is one of the 5 or so instances like that I refered to earlier. I'll fix it in a few hours when I am back home.

corebob: But how would you fix this?
```c++
int *a, *b, &c = value;
```

benma:
utACK 72f00608d0a130a1488dc49df1073abe01d49519

sipa:
In C++ syntax, elements like `[]`, `*`, and `&` belong with the variable,
not the type. An example is `int *a, b`, which declares `b` as an `int`,
not as a pointer to `int`.

However, in terms of style recommendation, this is controversial, and the
codebase contains plenty of examples of left and right 'hugging'. I
wouldn't worry about using one over the other.

benma:
I guess you're right that it doesn't matter, I just thought it made sense to stick with the prelevant style in the code. Personally I put it with the type because of things like `std::vector<Foo*>` and func signatures.

ryanofsky:
Currently we have [PointerAlignment: Left](https://clang.llvm.org/docs/ClangFormatStyleOptions.html) in our clang-format file. This puts pointers and reference symbols next to the type instead of the variable. Should probably remove the setting or change it if this style isn't actually preferred.

sipa:
utACK

PR #10894, opened 2017-07-21
Add CChain object for headers-only chain

jonasschnelli:
This PR introduces a `CChain` object for the headers chain. It allows to better inspect the headers chain (things like reorg detection). It would simplify a lot of things for a possible light client (SPV) mode.

Slightly improves the headers-chain fork reporting in `getchaintips`.

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -60,6 +60,7 @@ CCriticalSection cs_main;
 
 BlockMap mapBlockIndex;
 CChain chainActive;
+CChain headersChainActive;
 CBlockIndex *pindexBestHeader = NULL;
 CWaitableCriticalSection csBestBlock;
 CConditionVariable cvBlockChange;
@@ -2613,7 +2614,10 @@ static CBlockIndex* AddToBlockIndex(const CBlockHeader& block)
     pindexNew->nChainWork = (pindexNew->pprev ? pindexNew->pprev->nChainWork : 0) + GetBlockProof(*pindexNew);
     pindexNew->RaiseValidity(BLOCK_VALID_TREE);
     if (pindexBestHeader == NULL || pindexBestHeader->nChainWork < pindexNew->nChainWork)
+    {
         pindexBestHeader = pindexNew;
+        headersChainActive.SetTip(pindexNew);
+    }
 
     setDirtyBlockIndex.insert(pindexNew);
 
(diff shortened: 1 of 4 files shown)


src/validation.cpp
```diff
@@ -2613,7 +2614,10 @@ static CBlockIndex* AddToBlockIndex(const CBlockHeader& block)
     pindexNew->nChainWork = (pindexNew->pprev ? pindexNew->pprev->nChainWork : 0) + GetBlockProof(*pindexNew);
     pindexNew->RaiseValidity(BLOCK_VALID_TREE);
     if (pindexBestHeader == NULL || pindexBestHeader->nChainWork < pindexNew->nChainWork)
+    {
```
MeshCollider: Nit, brace on new line

src/validation.cpp
```diff
@@ -3483,7 +3487,10 @@ bool static LoadBlockIndexDB(const CChainParams& chainparams)
         if (pindex->pprev)
             pindex->BuildSkip();
         if (pindex->IsValid(BLOCK_VALID_TREE) && (pindexBestHeader == NULL || CBlockIndexWorkComparator()(pindexBestHeader, pindex)))
+        {
```
MeshCollider: Same here

meshcollider (commented):
utACK https://github.com/bitcoin/bitcoin/commit/890c935d187401dc9e95f2a2c992a5aec4f438ad
Couple of nits

sipa:
It is unnecessary to add a new CChain object (3.5 MiB of RAM) for this. You can test whether `pindex` is an ancestor of `pindexBestHeader` using 'pindexBestHeader->GetAncestor(pindex->nHeight) == pindex`.

jonasschnelli:
@sipa:
For the light client headers based fork detection the `FindFork()` method would be very handy. Maybe have a look at the "sketch" SPV code for the wallet: https://github.com/bitcoin/bitcoin/pull/9483/files#diff-b2bb174788c7409b671c46ccc86034bdR1182

jonasschnelli:
If one runs in full non validation mode (light client mode), the memory consumption would be ~the same, because only the headers chain will require memory (the `chainActive` will be empty).

sipa:
@jonasschnelli You can use FindCommonAncestor between two CBlockIndex objects if you don't have a CChain, instead of FindFork.

jonasschnelli:
Thanks @sipa. Makes sense.
Closing for now.

danra:
@jonasschnelli @sipa If this is no longer relevant I suggest removing it from the "Client-Mode (SPV)" project to reduce noise (especially since there are *already* two other, somewhat overlapping, PRs in the project)

PR #10895, opened 2017-07-21
[wallet] Fix wallet memory leaks in cases of unsuccessful CreateWalletFromFile calls

practicalswift:
Fix wallet memory leaks in cases of unsuccessful `CreateWalletFromFile` calls.


bpay:
This sort of thing should be handled by unique_ptr or some other form of RAII.

jonasschnelli:
Agree with @bpay: Should use a unique_ptr for `CWallet *tempWallet = new CWallet(std::move(dbw));` (same way like the line above: `std::unique_ptr<CWalletDBWrapper> dbw(new CWalletDBWrapper(&bitdb, walletFile));`)

PR #10602, opened 2017-06-15
Make clang-format use C++11 features (e.g. A<A<int>> instead of A<A<int> >)

practicalswift:
Make `clang-format` use C++11 features (e.g. `A<A<int>>` instead of `A<A<int> >`).

PR #10440, opened 2017-05-22
[tests] Add libFuzzer support

practicalswift:
Add `libFuzzer` support.

As discussed in [issue #10364](https://github.com/bitcoin/bitcoin/issues/10364#issuecomment-300000902).

See http://llvm.org/docs/LibFuzzer.html#fuzzer-usage for usage instructions.

PR #10898, opened 2017-07-21
Fix invalid checks (NULL checks after dereference, redundant checks, etc.)

practicalswift:
Contains:
* Remove redundant `NULL` checks after throwing `new`
* Remove redundant check (`!ecc` is always true)
* Remove duplicate `uriParts.size() > 0` check
* Use two boolean literals instead of re-using variable

### DIFF
--- a/src/bitcoin-tx.cpp
+++ b/src/bitcoin-tx.cpp
@@ -682,10 +682,10 @@ static void MutateTx(CMutableTransaction& tx, const std::string& command,
     else if (command == "outaddr")
         MutateTxAddOutAddr(tx, commandVal);
     else if (command == "outpubkey") {
-        if (!ecc) { ecc.reset(new Secp256k1Init()); }
+        ecc.reset(new Secp256k1Init());
         MutateTxAddOutPubKey(tx, commandVal);
     } else if (command == "outmultisig") {
-        if (!ecc) { ecc.reset(new Secp256k1Init()); }
+        ecc.reset(new Secp256k1Init());
         MutateTxAddOutMultiSig(tx, commandVal);
     } else if (command == "outscript")
         MutateTxAddOutScript(tx, commandVal);
@@ -693,7 +693,7 @@ static void MutateTx(CMutableTransaction& tx, const std::string& command,
         MutateTxAddOutData(tx, commandVal);
 
     else if (command == "sign") {
-        if (!ecc) { ecc.reset(new Secp256k1Init()); }
+        ecc.reset(new Secp256k1Init());
         MutateTxSign(tx, commandVal);
     }
 
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -371,19 +371,17 @@ void MaybeSetPeerAsAnnouncingHeaderAndIDs(NodeId nodeid, CConnman* connman) {
             }
         }
         connman->ForNode(nodeid, [connman](CNode* pfrom){
-            bool fAnnounceUsingCMPCTBLOCK = false;
             uint64_t nCMPCTBLOCKVersion = (pfrom->GetLocalServices() & NODE_WITNESS) ? 2 : 1;
             if (lNodesAnnouncingHeaderAndIDs.size() >= 3) {
                 // As per BIP152, we only get 3 of our peers to announce
                 // blocks using compact encodings.
-                connman->ForNode(lNodesAnnouncingHeaderAndIDs.front(), [connman, fAnnounceUsingCMPCTBLOCK, nCMPCTBLOCKVersion](CNode* pnodeStop){
-                    connman->PushMessage(pnodeStop, CNetMsgMaker(pnodeStop->GetSendVersion()).Make(NetMsgType::SENDCMPCT, fAnnounceUsingCMPCTBLOCK, nCMPCTBLOCKVersion));
+                connman->ForNode(lNodesAnnouncingHeaderAndIDs.front(), [connman, nCMPCTBLOCKVersion](CNode* pnodeStop){
+                    connman->PushMessage(pnodeStop, CNetMsgMaker(pnodeStop->GetSendVersion()).Make(NetMsgType::SENDCMPCT, /*fAnnounceUsingCMPCTBLOCK=*/false, nCMPCTBLOCKVersion));
                     return true;
                 });
                 lNodesAnnouncingHeaderAndIDs.pop_front();
             }
-            fAnnounceUsingCMPCTBLOCK = true;
-            connman->PushMessage(pfrom, CNetMsgMaker(pfrom->GetSendVersion()).Make(NetMsgType::SENDCMPCT, fAnnounceUsingCMPCTBLOCK, nCMPCTBLOCKVersion));
+            connman->PushMessage(pfrom, CNetMsgMaker(pfrom->GetSendVersion()).Make(NetMsgType::SENDCMPCT, /*fAnnounceUsingCMPCTBLOCK=*/true, nCMPCTBLOCKVersion));
             lNodesAnnouncingHeaderAndIDs.push_back(pfrom->GetId());
             return true;
         });
--- a/src/qt/walletview.cpp
+++ b/src/qt/walletview.cpp
@@ -122,8 +122,8 @@ void WalletView::setWalletModel(WalletModel *_walletModel)
     overviewPage->setWalletModel(_walletModel);
     receiveCoinsPage->setModel(_walletModel);
     sendCoinsPage->setModel(_walletModel);
-    usedReceivingAddressesPage->setModel(_walletModel->getAddressTableModel());
-    usedSendingAddressesPage->setModel(_walletModel->getAddressTableModel());
+    usedReceivingAddressesPage->setModel(_walletModel ? _walletModel->getAddressTableModel() : nullptr);
+    usedSendingAddressesPage->setModel(_walletModel ? _walletModel->getAddressTableModel() : nullptr);
 
     if (_walletModel)
     {
--- a/src/rest.cpp
+++ b/src/rest.cpp
@@ -409,10 +409,8 @@ static bool rest_getutxos(HTTPRequest* req, const std::string& strURIPart)
 
     if (uriParts.size() > 0)
     {
-
         //inputs is sent over URI scheme (/rest/getutxos/checkmempool/txid1-n/txid2-n/...)
-        if (uriParts.size() > 0 && uriParts[0] == "checkmempool")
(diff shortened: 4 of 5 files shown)


src/rest.cpp
```diff
@@ -427,9 +427,8 @@ static bool rest_getutxos(HTTPRequest* req, const std::string& strURIPart)
 
     if (uriParts.size() > 0)
     {
-
         //inputs is sent over URI scheme (/rest/getutxos/checkmempool/txid1-n/txid2-n/...)
-        if (uriParts.size() > 0 && uriParts[0] == "checkmempool")
+        if (uriParts[0] == "checkmempool")
             fCheckMemPool = true;
```
benma: add braces or put on same line according do the dev notes.

src/net_processing.cpp
```diff
@@ -418,19 +418,17 @@ void MaybeSetPeerAsAnnouncingHeaderAndIDs(NodeId nodeid, CConnman& connman) {
...
         connman.ForNode(nodeid, [&connman](CNode* pfrom){
-            bool fAnnounceUsingCMPCTBLOCK = false;
             uint64_t nCMPCTBLOCKVersion = (pfrom->GetLocalServices() & NODE_WITNESS) ? 2 : 1;
             if (lNodesAnnouncingHeaderAndIDs.size() >= 3) {
                 // As per BIP152, we only get 3 of our peers to announce
                 // blocks using compact encodings.
-                connman.ForNode(lNodesAnnouncingHeaderAndIDs.front(), [&connman, fAnnounceUsingCMPCTBLOCK, nCMPCTBLOCKVersion](CNode* pnodeStop){
-                    connman.PushMessage(pnodeStop, CNetMsgMaker(pnodeStop->GetSendVersion()).Make(NetMsgType::SENDCMPCT, fAnnounceUsingCMPCTBLOCK, nCMPCTBLOCKVersion));
+                connman.ForNode(lNodesAnnouncingHeaderAndIDs.front(), [&connman, nCMPCTBLOCKVersion](CNode* pnodeStop) {
+                    connman.PushMessage(pnodeStop, CNetMsgMaker(pnodeStop->GetSendVersion()).Make(NetMsgType::SENDCMPCT, false, nCMPCTBLOCKVersion));
```
benma: Now the code is less readable, as I have to guess or look up what `false` means here. I'd restore the var in this case.

src/bitcoin-tx.cpp
```diff
@@ -687,18 +687,18 @@ static void MutateTx(CMutableTransaction& tx, const std::string& command,
     else if (command == "outaddr")
         MutateTxAddOutAddr(tx, commandVal);
     else if (command == "outpubkey") {
-        if (!ecc) { ecc.reset(new Secp256k1Init()); }
+        ecc.reset(new Secp256k1Init());
```
benma: 1. The usage of `Secp256k1Init` is pretty obscure here. Why is it needed for those few commands but not others?
2. Would it be a better idea to move the Secp256k1Init into the functions that are called? As close as possible to where it is actually required, which is not hard to see now.

Pinging @sipa, the author of this.

src/qt/walletview.cpp
```diff
@@ -122,11 +122,12 @@ void WalletView::setWalletModel(WalletModel *_walletModel)
     overviewPage->setWalletModel(_walletModel);
     receiveCoinsPage->setModel(_walletModel);
     sendCoinsPage->setModel(_walletModel);
-    usedReceivingAddressesPage->setModel(_walletModel->getAddressTableModel());
-    usedSendingAddressesPage->setModel(_walletModel->getAddressTableModel());
 
     if (_walletModel)
```
benma: After a quick research it looks like `_walletModel` is never passed as NULL. So it would be good to change the argument from a pointer to a reference and remove this check.

benma (changes requested):

src/net_processing.cpp
practicalswift: Fixed by adding two comments :-)

practicalswift:
@benma Thanks for reviewing! Good feedback. All items addressed. Looks good? :-)

src/net_processing.cpp
```diff
@@ -418,19 +418,17 @@ void MaybeSetPeerAsAnnouncingHeaderAndIDs(NodeId nodeid, CConnman& connman) {
...
         connman.ForNode(nodeid, [&connman](CNode* pfrom){
-            bool fAnnounceUsingCMPCTBLOCK = false;
             uint64_t nCMPCTBLOCKVersion = (pfrom->GetLocalServices() & NODE_WITNESS) ? 2 : 1;
             if (lNodesAnnouncingHeaderAndIDs.size() >= 3) {
                 // As per BIP152, we only get 3 of our peers to announce
                 // blocks using compact encodings.
-                connman.ForNode(lNodesAnnouncingHeaderAndIDs.front(), [&connman, fAnnounceUsingCMPCTBLOCK, nCMPCTBLOCKVersion](CNode* pnodeStop){
-                    connman.PushMessage(pnodeStop, CNetMsgMaker(pnodeStop->GetSendVersion()).Make(NetMsgType::SENDCMPCT, fAnnounceUsingCMPCTBLOCK, nCMPCTBLOCKVersion));
+                connman.ForNode(lNodesAnnouncingHeaderAndIDs.front(), [&connman, nCMPCTBLOCKVersion](CNode* pnodeStop){
+                    connman.PushMessage(pnodeStop, CNetMsgMaker(pnodeStop->GetSendVersion()).Make(NetMsgType::SENDCMPCT, /*fAnnounceUsingCMPCTBLOCK=*/false, nCMPCTBLOCKVersion));
```
benma: I am not sure whether I like this 😆 

Maybe nicer to define two constants for true/false, or an enum?

I commend the effort to reduce the number of lines, but in some cases, being descriptive is worth it.

utACK anyway, I'll leave the decision to you.

benma:
utACK da5f591c03e04cfc159aa9d8ed1036576f51aa25

src/bitcoin-tx.cpp
sipa: Some libsecp256k1 calls require a context, and some don't.

practicalswift:
Rebased!

practicalswift:
Rebased! :-)

src/qt/walletview.h
```diff
@@ -48,7 +48,7 @@ class WalletView : public QStackedWidget
         The wallet model represents a bitcoin wallet, and offers access to the list of transactions, address book and sending
         functionality.
     */
-    void setWalletModel(WalletModel *walletModel);
+    void setWalletModel(WalletModel& walletModel);
```
laanwj: Please don't change the `setWalletModel()` interface. It is supposed to be consistent over all objects, and setting a `NULL` model is meaningful: it should be possible to use the view classes without a model for simple testing.

practicalswift:
@laanwj I've now removed the `Move NULL check prior to dereference` commit.

Please note that setting a NULL model will result in a null pointer dereference in current `master`. Should I fix that?

What about:

```diff
diff --git a/src/qt/walletview.cpp b/src/qt/walletview.cpp
index 971f5e0..a56a400 100644
--- a/src/qt/walletview.cpp
+++ b/src/qt/walletview.cpp
@@ -122,8 +122,8 @@ void WalletView::setWalletModel(WalletModel *_walletModel)
     overviewPage->setWalletModel(_walletModel);
     receiveCoinsPage->setModel(_walletModel);
     sendCoinsPage->setModel(_walletModel);
-    usedReceivingAddressesPage->setModel(_walletModel->getAddressTableModel());
-    usedSendingAddressesPage->setModel(_walletModel->getAddressTableModel());
+    usedReceivingAddressesPage->setModel(_walletModel ? _walletModel->getAddressTableModel() : nullptr);
+    usedSendingAddressesPage->setModel(_walletModel ? _walletModel->getAddressTableModel() : nullptr);
```

laanwj:
@practicalshift Yes that should be fixed, looks good to me!

practicalswift:
@laanwj I've now added a NULL check for `_walletModel`. Looks good now? :-)

ryanofsky (commented):
utACK 76fed838f381a6efba175f3650ec7e9dc73016d3

@practicalswift, how are you finding these changes? Are you using a linter tool? Or methodically going through the source code in some way and reviewing it?

practicalswift:
@ryanofsky I use a combination of manual code review, static analysis and fuzzing to find potential issues. In the Apple Swift project most of my results have been from fuzzing (using my own compiler fuzzer): [see my `apple/swift` contributions here](https://github.com/apple/swift/pulls?utf8=%E2%9C%93&q=is%3Apr%20author%3Apracticalswift%20is%3Aclosed%20crash).

ryanofsky:
Thanks. It could be educational (and also nice to give credit) if you mentioned the specific static analysis & fuzzing tools were that used in your commit messages and PR descriptions.

promag:
Agree with @ryanofsky, also for useful when reviewing!

JeremyRubin:
Echoing @ryanofsky, would even be useful to fill this info in on already merged PRs (people often refer to them long after the fact).

practicalswift:
@ryanofsky @promag @JeremyRubin  Sure! Thanks for showing interest! That is appreciated.

Here are pointers to some of the open source tools that I use regularly:
* clang's Thread Safety Analysis - see PR #11226 (**please review!**) + #10866 (**please review!**)
* `libFuzzer` - see PR #10440 (**please review!**)
* `afl-fuzz` - see PR #10415
* `valgrind` - see PR #11035 (**please review!**)
* `clang-tidy` - see PR #10735
* `clang-format` - see PR #10602
* `pycodestyle` - see PR #9508
* clang's `AddressSanitizer` (ASan: `-fsanitize=address`) 
* clang's `LeakSanitizer` (LSan: `-fsanitize=leak`)
* clang's `MemorySanitizer` (MSan: `-fsanitize=memory`)
* clang's `ThreadSanitizer` (TSan: `-fsanitize=thread`)
* clang's `UndefinedBehaviorSanitizer` (UBSan: `-fsanitize=undefined`)
* `flintpp`
* `oclint`
* `cppcheck`

The list is not at all exhaustive - I try to use as many tools as possible to increase diversity in the classes of issues covered. In addition to the above I use some custom/proprietary analysis tools/scripts, but they are not that interesting or general.

If you want to help me, please review the open PR:s mentioned above to increase the chance of them getting merged :-)

ryanofsky:
Can this be merged? Has two utACKs and is a small, minor change.

laanwj:
Looks good to me, utACK 76fed83

PR #10900, opened 2017-07-21
[0.14] Enforce segsignal activation height and rules

achow101:
This PR adds a segsignal compatible rule that all blocks after block 477120 (segsignal's activation height) must signal for segwit so long as segwit is still in the `started` state. This will make sure that miners are enforcing the rules that they have chosen to activate.

This PR is on the 0.14 branch so that a 0.14.3 release can be quickly made which includes this change. It can also be easily rebased onto master if needed.

### DIFF
--- a/qa/rpc-tests/segsignal.py
+++ b/qa/rpc-tests/segsignal.py
@@ -1,0 +1,68 @@
+#!/usr/bin/env python3
+# Copyright (c) 2015-2016 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+from test_framework.test_framework import ComparisonTestFramework
+from test_framework.util import *
+from test_framework.blocktools import create_coinbase, create_block
+
+class SegsignalUASFTest(ComparisonTestFramework):
+
+    def __init__(self):
+        super().__init__()
+        self.setup_clean_chain = True
+        self.num_nodes = 1
+
+    def run_test(self):
+        self.tip = int("0x" + self.nodes[0].getbestblockhash(), 0)
+        self.height = 1  # height of the next block to build
+        self.last_block_time = int(time.time())
+
+        # Mine 1439 non segwit blocks
+        self.generate_blocks(1439, 0x20000000)
+
+        # Check that segwit is still started
+        assert_equal(self.get_bip9_status('segwit')['status'], 'started')
+
+        # Mine a non-segwit block. this should be rejected
+        self.generate_blocks(1, 0x20000000, 'bad-no-segwit')
+
+        # Now mine a segwit block. this should not be rejected
+        self.generate_blocks(1, 0x20000002)
+
+        # Mine another non-segwit block, this should be rejected
+        self.generate_blocks(1, 0x20000000, 'bad-no-segwit')
+
+        # Mine 143 more blocks. should still be started
+        print(self.nodes[0].getblockchaininfo())
+        self.generate_blocks(142, 0x20000002)
+        assert_equal(self.get_bip9_status('segwit')['status'], 'started')
+
+        # Mine 1 more block, should now be locked in
+        self.generate_blocks(1, 0x20000002)
+        print(self.nodes[0].getblockchaininfo())
+        assert_equal(self.get_bip9_status('segwit')['status'], 'locked_in')
+
+        # Mine a non-segwit block, should be accepted
+        self.generate_blocks(1, 0x20000000)
+
+    def generate_blocks(self, number, version, error = None):
+        for i in range(number):
+            block = create_block(self.tip, create_coinbase(self.height), self.last_block_time + 1)
+            block.nVersion = version
+            block.rehash()
+            block.solve()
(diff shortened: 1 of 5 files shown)


morcos:
Strong Concept ACK

However I don't think this should be built on the 0.14 branch but directly on 0.14.2
I don't think we want any of the other backports.
If we merge this, there will be no time for significant testing and we want the decision on whether to run this or keep running 0.14.2 to be based solely on a user's opinion on whether BIP91 represents the new consensus rules.

I'd advocate for releasing this as perhaps 0.14.2-segsignal or something else outside of the usual version numbering scheme.

In future versions, I think we should be able to drop all this transient BIP91 nonsense.
This code/patch is hopefully only necessary to protect users for the next 3 weeks until BIP 141 (segwit) activates.

achow101:
@morcos I have rebased it so that it is just a patch on top of 0.14.2

qa/rpc-tests/segsignal.py
```diff
@@ -0,0 +1,68 @@
...
+        self.generate_blocks(142, 0x20000002)
+        assert_equal(self.get_bip9_status('segwit')['status'], 'started')
+
+        # Mine 1 more block, should now be locked in
+        self.generate_blocks(1, 0x20000002)
+        print(self.nodes[0].getblockchaininfo())
+        assert_equal(self.get_bip9_status('segwit')['status'], 'locked_in')
+
+        # Mine a non-segwit block, should be accepted
+        self.generate_blocks(1, 0x20000000)
```
jimmysong: Not sure if this is a test case we necessarily care about, but maybe generate another 143 blocks and check that a non-segwit-signaling block is still accepted?

jimmysong (commented):
Tested ACK

Thank you for getting this change in so quickly. Just one minor point on the rpc test.

lichtamberg:
utAck - yes please, merge BIP148 into core! 🎉 😄

qa/rpc-tests/segsignal.py
achow101: I don't think that test case matters as it is already covered by this one here.

src/validation.cpp
```diff
@@ -1851,6 +1851,17 @@ bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockIndex* pin
         flags |= SCRIPT_VERIFY_NULLDUMMY;
     }
 
+    // SEGSIGNAL mandatory segwit signalling.
+    if (pindex->nHeight >= chainparams.GetConsensus().BIP91Height &&
+        VersionBitsState(pindex->pprev, chainparams.GetConsensus(), Consensus::DEPLOYMENT_SEGWIT,    versionbitscache) == THRESHOLD_STARTED)
```
jamesob: nit: whitespace before last param

qa/rpc-tests/segsignal.py
jimmysong: Wouldn't the bip9 status be "active"? Just suggesting you check behavior in that last state for completeness.

luke-jr:
Concept ACK, agree with @morcos on a new branch for 0.14.2.1 or such (we've used a fourth number for this kind of purpose before). Code looks okay too, but I didn't test (utACK).

andyrowe:
NACK.  I think it's absurdly late in the game to be attempting to throw together a minimally tested segwit2x compatibility fix given the current state of the network.

petertodd:
Concept ACK @morcos's suggestions.

Similar situation w/ the previous BIP148 pull-req: Core users have good reason to want this rule, but whether or not it represents Bitcoin is still undecided. Unlike the BIP148 situation, we're incredibly short on time, so a non-standard version like -segsignal or .1 (@luke-jr's suggestion) makes sense.

tmornini:
I'd encourage naming the release 0.14.2-segsignal rather than 0.14.2.1.

Rational: There's nothing in 0.14.2.1 to indicate that it's NOT a bug fix of 0.14.2.

Consider if a decision is made to release a BIP148 patch against 0.14.2. What might it be called? 0.14.2-BIP148 is perfectly clear, whereas 0.14.2.2 could easily be a follow-on release to 0.14.2.1.

Avoid confusion at all costs, and always call a thing a thing. :-)

felipelalli:
ACK right now.

jameshilliard:
utAck

luke-jr:
@tmornini Softforks are always deployed as a minor release. In ordinary circumstances, BIP148 or this would be 0.14.3. The only reason this can't really be that, is because we don't have time to do a full RC cycle to test the other fixes (which are being skipped). Historically such time-sensitive releases have used a fourth number, hence 0.14.2.1.

gomedera:
This change could be released as 14.3 and whatever 14.3 meant to be could be released as 14.4. Problem solved.

earonesty:
It's too late now, but any consensus change where it may be possible that the change is unsafe today, but may be safer in the future should be released as a user configurable option early on, not patched in at the last minute.

kallewoof:
utACK b6abab653300da5152d9d9cfa5f8582e0b4b85d4

Leviathn:
Concept ACK in line with @morcos @luke-jr & @petertodd

achow101:
BIP 91 is no longer active now that segwit has locked in, so I am closing this.

PR #10903, opened 2017-07-22
Add configuration files for a Docker-based development environment

jamesob:
This PR introduces a number of configuration files which facilitate use of a Docker environment for local development and bitcoind execution. I've found this method to be an easy way to quickly setup a local dev environment; the use of a container also allows us to prevent polluting the host system with Bitcoin-specific dependencies.

The base image that I reference in the Dockerfile is maintained by Docker [in an "official" image repository](https://hub.docker.com/_/python/). It's based on Debian 8 "Jessie" and comes with Python 3.6.2 for running the tests.

After obtaining the Docker dependencies (as described in `docs/build-docker.md`), building and testing bitcoin from scratch can be done in a single command:
```
docker-compose run bitcoind test
```

During the docker image build (`docker-compose build bitcoind`), system dependencies are installed per `docs/build-unix.md` (including all non-GUI optional dependencies) but Bitcoin itself isn't compiled. This is because `docker-compose` attaches the source tree to the container as a mounted volume after build, which allows changes we make to the source on the host machine to propagate immediately to a running container. The entrypoint script (`./bin/docker_entrypoint.sh`) is responsible for handling bitcoind compilation as well as providing the `test*` commands. Clang is installed and used as the default compiler.

You'll notice that a few files (`Dockerfile`, `bin/`, `docker-compose.yaml`) are introduced to the top-level of the directory structure. I wasn't eager to do this -- I started out in `contrib/`, but in order for `docker-compose` to mount the entire source tree, commands have to be executed from the project root.

This is a modest suggestion and I'm happy to help incorporate feedback.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,35 @@
+# A base Docker image running Python 3.6.2 and Debian 8 "Jessie"
+# See: https://github.com/docker-library/python/blob/master/3.6/jessie/Dockerfile
+FROM python:3.6.2-jessie
+
+RUN apt-get update && \
+    apt-get install -y \
+    libboost-system-dev libboost-filesystem-dev libboost-chrono-dev \
+    libboost-program-options-dev libboost-test-dev libboost-thread-dev \
+    build-essential libtool autotools-dev automake pkg-config libssl-dev \
+    libevent-dev bsdmainutils \
+    libzmq3 miniupnpc \
+    # Install clang for potential use over g++.
+    clang
+
+# The Bitcoin source code will be located at /code and mounted by 
+# docker-compose.
+ENV HOME /code
+WORKDIR /code
+
+# Install Berkeley DB 4.8 for use with wallet functionality.
+ENV BDB_PREFIX /db4
+ENV BDB_VERSION db-4.8.30.NC
+RUN mkdir -p $BDB_PREFIX && \
+    wget --no-check-certificate \
+        "https://download.oracle.com/berkeley-db/${BDB_VERSION}.tar.gz" \
+        -O ${BDB_VERSION}.tar.gz && \
+    echo "12edc0df75bf9abd7f82f821795bcee50f42cb2e5f76a6a281b85732798364ef  ${BDB_VERSION}.tar.gz" \
+        | sha256sum -c && \
+    tar -xzvf ${BDB_VERSION}.tar.gz -C $BDB_PREFIX && \
+    rm ${BDB_VERSION}.tar.gz && \
+    ${BDB_PREFIX}/${BDB_VERSION}/dist/configure \
+        --enable-cxx --disable-shared --with-pic --prefix=$BDB_PREFIX && \
+    make install
+
+ENTRYPOINT ["./bin/docker_entrypoint.sh"]
--- a/doc/build-docker.md
+++ b/doc/build-docker.md
@@ -1,0 +1,64 @@
+# Docker Build Instructions and Notes
+
+Bitcoind can be easily built on any platform which hosts a 
+[Docker](https://en.wikipedia.org/wiki/Docker_(software)) daemon. This is
+likely the lowest-effort way to get a bitcoin development environment on your
+machine.
+
+The Docker development environment does not facilitate development of the
+GUI.
+
+## Preparation
+
+1. [Install Docker](https://www.docker.com/community-edition#/download).
+2. If on Linux, install Docker Compose with `pip install docker-compose`.
+3. Change directory to the root of this repository.
+4. Run `docker-compose run bitcoind` to compile bitcoin.
+
+Bitcoin's source tree will be
+[mounted](https://docs.docker.com/compose/compose-file/compose-file-v2/#volumes-volume_driver) 
+into the Docker container, so any changes you make to the source tree on
+your host computer will be immediately reflected within the Docker container.
+
+## Commands
+
+### Run all tests
+
+```sh
+docker-compose run --rm bitcoind test
+```
+
+### Run C++ tests
+
+```sh
+docker-compose run --rm bitcoind test-cpp [args ...]
+```
+
+### Run Python tests
+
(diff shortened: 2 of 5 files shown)


Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+
+# The Bitcoin source code will be located at /code and mounted by 
+# docker-compose.
+ENV HOME /code
+WORKDIR /code
+
+# Install Berkeley DB 4.8 for use with wallet functionality.
+ENV BDB_PREFIX /db4
+RUN mkdir -p $BDB_PREFIX
+RUN wget 'http://download.oracle.com/berkeley-db/db-4.8.30.NC.tar.gz'
```
emanuelb: 1. it better to use https link with --no-check-certificate option (until download.oracle.com have valid SSL cert)
2. it better to use -O parameter to specify filename of downloaded file (there was issues in filename auto-detection of wget in the past)
3. use environment variable for 'db-4.8.30.NC' string and use it instead below (can be updated easily in future)

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN echo '12edc0df75bf9abd7f82f821795bcee50f42cb2e5f76a6a281b85732798364ef  db-4.8.30.NC.tar.gz' \
+    | sha256sum -c | grep 'db-4.8.30.NC.tar.gz: OK'
+RUN tar -xzvf db-4.8.30.NC.tar.gz -C $BDB_PREFIX
```
emanuelb: remove the 'db-4.8.30.NC.tar.gz' file after extract operation.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN apt-get install -y clang
+
+# The Bitcoin source code will be located at /code and mounted by 
+# docker-compose.
+ENV HOME /code
+WORKDIR /code
+
+# Install Berkeley DB 4.8 for use with wallet functionality.
+ENV BDB_PREFIX /db4
+RUN mkdir -p $BDB_PREFIX
```
emanuelb: use && instead of multiple RUN commands to avoid unneeded layers.

emanuelb (changes requested):
added comments about stuff that better to change in Dockerfile.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN wget 'http://download.oracle.com/berkeley-db/db-4.8.30.NC.tar.gz'
+RUN echo '12edc0df75bf9abd7f82f821795bcee50f42cb2e5f76a6a281b85732798364ef  db-4.8.30.NC.tar.gz' \
+    | sha256sum -c | grep 'db-4.8.30.NC.tar.gz: OK'
+RUN tar -xzvf db-4.8.30.NC.tar.gz -C $BDB_PREFIX
+RUN ls ${BDB_PREFIX}
```
stevendlander: Unnecessary `ls` on the BDB directory.

Dockerfile
```diff
@@ -0,0 +1,33 @@
...
+RUN apt-get update && \
+    apt-get install -y \
+    libboost-system-dev libboost-filesystem-dev libboost-chrono-dev \
+    libboost-program-options-dev libboost-test-dev libboost-thread-dev \
+    build-essential libtool autotools-dev automake pkg-config libssl-dev \
+    libevent-dev bsdmainutils \
+    libzmq3 miniupnpc
+
+# Install clang for potential use over g++.
+RUN apt-get install -y clang
```
stevendlander: This should also be rolled up with the above apt install block, with the comment above it.

```docker
RUN apt-get update && apt-get install -y \
    libboost-system-dev \
    ...
    miniupnpc \
    # Install clang for potential use over g++.
    clang
```

doc/build-docker.md
```diff
@@ -0,0 +1,50 @@
+# Docker Build Instructions and Notes
+
+Bitcoind can be easily built on any platform which hosts a 
+[Docker](https://en.wikipedia.org/wiki/Docker_(software)) daemon. This is
+likely the lowest-effort way to get a bitcoin development environment on your
+machine.
+
+The Docker development enviornment does not facilitate development of the
```
stevendlander: nit: `environment`

jamesob:
@emanuelb @stevendlander thanks for the feedback. I've incorporated that as well as a few changes to make running bitcoind itself more straightforward.

greenaddress:
I didn't review and for now neutral on the concept but why jessie and not stretch?

greenaddress:
Also rather than having lots of && which make it very unreadable why not using COPY for a bash script file and then RUN on it? the script would be reusable to setup dependency on Jessie/Stretch non docker too.

promag:
I understand why you want this but a `Dockerfile` in the root usually is used to build the service container. The built image doesn't even have bitcoind. Also, `docker-compose.yml` has 1 service only, with the intent to avoid calling docker with those parameters.

The usage here is for development, so I would rename `Dockerfile` to `Dockerfile.dev` (or something), remove `docker-compose.yml` (less dependencies) and update the instructions. 

However I dunno if this is something worth to maintain here, or something you could put in another repo.

I can review if this receives concept ACK.

jamesob:
Those are all good points. I think closing this PR and creating a personal repo is probably the best route.

Does anyone think it would be worth filing a separate PR to put the db-4 installation script into `contrib/` or elsewhere?

Sjors:
@jamesob did you ever end up updating your local repo?

I created yet another Dockerfile: https://github.com/Sjors/docker-bitcoin-core

It borrows a lot of code from @NicolasDorier's [project](https://github.com/NicolasDorier/docker-bitcoin), but it builds from source. It lets you build any tag or commit (only tried it for the latest master though). Only support modern BDB. It also lets you attach.

I tend to agree that there are too many ways to do this to warrant an "official" Docker file, unless perhaps if they end being used to replace Gitian, or maybe for Travis.

PR #10907, opened 2017-07-22
Prefer iterators arrow operator over iterator dereference

bytting:
Prefer iterators arrow operator over iterator dereference to improve readability

### DIFF
--- a/src/addrman.cpp
+++ b/src/addrman.cpp
@@ -71,10 +71,10 @@ CAddrInfo* CAddrMan::Find(const CNetAddr& addr, int* pnId)
     if (it == mapAddr.end())
         return NULL;
     if (pnId)
-        *pnId = (*it).second;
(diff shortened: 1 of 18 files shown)


dcousens (approved):

practicalswift:
Should we extend this to doing it globally? If so, the following might be of help:

```
git grep -E '\(\*[a-zA-Z0-9]+\)\.(first|second)' # 80 occurrences of 
                                                 # which 7 in validation.cpp
```

Ideally such a change should probably be made as an scripted-diff.

corebob:
@practicalswift 
That only works for iterators to pairs, still that seems to be a common structure here.
I'm not quite comfortable with automating this but if someone wants to go for that, that's fine with me.

laanwj:
Although I agree with you with regard to readability, I'm not convinced this is worth the risk to do as a global change. Sorry, closing.

bytting:
@laanwj Fair enough

PR #10913, opened 2017-07-23
fixing a couple of typos

ghost:


MeshCollider:
NACK, these corrections aren't even grammatically correct

PR #10920, opened 2017-07-24
[qt] Fix potential memory leak in newPossibleKey(ChangeCWallet *wallet)

practicalswift:
Fix potential memory leak in `newPossibleKey(ChangeCWallet *wallet)`.

### DIFF
--- a/src/qt/walletmodeltransaction.cpp
+++ b/src/qt/walletmodeltransaction.cpp
@@ -10,7 +10,6 @@
 WalletModelTransaction::WalletModelTransaction(const QList<SendCoinsRecipient> &_recipients) :
     recipients(_recipients),
     walletTransaction(0),
-    keyChange(0),
     fee(0)
 {
     walletTransaction = new CWalletTx();
@@ -18,7 +17,6 @@ WalletModelTransaction::WalletModelTransaction(const QList<SendCoinsRecipient> &
 
 WalletModelTransaction::~WalletModelTransaction()
 {
-    delete keyChange;
     delete walletTransaction;
 }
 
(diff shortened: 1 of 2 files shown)


sipa:
Is it possible to change this to use a `unique_ptr` instead?

promag:
:+1: @sipa suggestion.

Correct me if I'm wrong, but from what I've seen, `WalletModelTransaction::keyChange` pointer is initialized to null, and only has a value when `newPossibleKeyChange` is called. So, in the first call, this would `delete NULL`.

Moreover, in the `~WalletModelTransaction` it also `delete keyChange` which can be null if `newPossibleKeyChange` is not called.

practicalswift:
@sipa Sure, I'll fix!

@promag We're doing `delete NULL` all over the codebase assuming it is a no-op :-)

promag:
Yeah it's safe, can be avoided though.

luke-jr:
But there's no reason to avoid it..

practicalswift:
Now using `unique_ptr`. Please review :-)

jonasschnelli:
utACK 1ce39c59c22a5878b5af6a4ff821db83c988e8b7

ryanofsky (commented):
utACK 446e2610b0cd6535e623952bf22ea09ce0ed8aaf. FWIW #10244 (commit "Remove most direct bitcoin calls from qt/walletmodel.cpp") takes this cleanup further, combining key and transaction into a pending transaction object and referencing that through a unique pointer.

promag:
utACK 446e261.

laanwj:
utACK 446e261

Issue #10736, opened 2017-07-03
Feature Request: The client should have a separate option for the storage location of the database data


fresheneesz:
After wondering why my bitcoin client was taking literally weeks to download the blockchain, I found out that it was going so ultra slow because I was storing the data directory on an external hard drive. After using a symlink to put the database data on my SSD and keep the blockchain stored on my external HD (via the symlink), my blockchain validation sped up by at least 6 fold. Details here: https://github.com/bitcoin/bitcoin/issues/10647 . 

Anyways, because this can be so important for a non-frustrating experience for users who don't have enough SSD space, I'm requesting that bitcoin core store the database data on the same drive as the program executables, and give the user the option of only where to put the blockchain data (what is currently the `blocks` directory within the data directory). Alternatively, we could give the user separate options for where to store the blockchain and where to store everything else, with some info about why the storage location is important. 

If I'd wanted to buy something with bitcoins in the last month, I simply wouldn't have been able to because my client was taking forfuckingever to sync up. Its really important to make these kinds of performance improvements idiot-proof so users with less technical savvy can use bitcoin core without rage quitting.

corebob:
> store the database data on the same drive as the program executables

This shouldn't be necessary. I have everything on a secondary drive and have no issues with speed.
It's probably the USB that causes the problem

sipa:
> The client should store the database data on the same drive as the program

That's not generally possible. On UNIX systems, the filesystem with binaries may not be writable by the user.

I do however think it's useful to allow the chainstate to be stored in another directory from the blockchain. The former mostly needs high IOPS, the latter only needs serial read/writes.

fresheneesz:
Alright, I changed the name of the request to be a separate option for storage location. 

@corebob Sipa also mentioned that USB is probably the issue. Do you have any information you could point me to that substantiates this? I don't see any logical reason why USB would cause a bottleneck here.

corebob:
@fresheneesz 
Not at the moment, it was just an assumption I made based on your description.

jimhashhq:
[...]

Issue #10647, opened 2017-06-21
Bitcoin Core not using configured database cache size and taking way too long to sync


fresheneesz:
### Describe the issue

My bitcoin core wallet is taking a ridiculously long time to catch up to the blockchain. I've been running it on and off for a MONTH now, and I'm still 41 weeks behind. Looking at my Task Manager, I see that my network is being almost totally unutilized at 0.1Mbps, tho my disk is being used at 1.5 MBps. It also says the Bitcoin Core (GUI node for Bitcoin) is only using 280 MB of memory, even tho I have `-dbcache=700` set in the config (which I can see it recognizes in the options since it lists it under "Active command-line options that override above options"). 

I have also tried removing `-dbcache=700` from my bitcoin.conf and instead setting 1000 MB as the database cache option in the GUI options window (and restarting of course). Neither the bitcoin.conf nor the database cache size option in the options seem to affect how much memory is allocated to bitcoin core, even tho I have at least 800MB free. 

### What version of bitcoin-core are you using?
Bitcoin Core version v0.14.1 (64-bit), official

### Machine specs:
- OS: Windows 8.1 64-bit
- CPU: Intel Core i7-4702HQ  @ 2.2 GHz
- RAM: 8 GB
- Disk size: 4TB HD for the blockchain, 250 GB SSD for the program and other (smaller) data
- Network Speed: 50 Mbps

### Any extra information that might be useful in the debugging process.
My debug.log (note the weird whitespace is verbatim):

```
2017-06-21 19:59:18 



















2017-06-21 19:59:18 Bitcoin version v0.14.2
2017-06-21 19:59:18 InitParameterInteraction: parameter interaction: -whitelistforcerelay=1 -> setting -whitelistrelay=1
2017-06-21 19:59:18 Assuming ancestors of block 00000000000000000013176bf8d7dfeab4e1db31dc93bc311b436e82ab226b90 have valid signatures.
2017-06-21 19:59:18 GUI: "registerShutdownBlockReason: Successfully registered: Bitcoin Core didn't yet exit safely..."
2017-06-21 19:59:18 Default data directory C:\Users\fresheneesz\AppData\Roaming\Bitcoin
2017-06-21 19:59:18 Using data directory H:\Bitcoin
2017-06-21 19:59:18 Using config file H:\Bitcoin\bitcoin.conf
2017-06-21 19:59:18 Using at most 125 automatic connections (2048 file descriptors available)
[...]

Issue #10787, opened 2017-07-10
ioctl error when opening database on external hard drive


dszakallas:
### Describe the issue
IO error when opening block database on NTFS partition of external hard drive on macOS.

### Can you reliably reproduce the issue?
Yes.

#### If so, please list the steps to reproduce below:
1. Delete database:
```
david@mac:/Volumes/SAMSUNG/Ledgers% rm -rf BitCoinCore
```
2.
Open BitCoin-Qt. On *Welcome Screen* choose *Use a custom data directory:* `/Volumes/SAMSUNG/Ledgers/BitCoinCore` and click OK.
3.
Wild alert box appears:
![image](https://user-images.githubusercontent.com/5807322/28039971-70b67136-65c4-11e7-8764-59ff4e9c7bab.png)
Click OK.

### Expected behaviour
Should start creating the block database I guess. Never made it through this step.

### Actual behaviour
![image](https://user-images.githubusercontent.com/5807322/28040006-8949dc88-65c4-11e7-9ddc-eff66eaeee44.png)
:(

Contents of debug.log:
```
2017-07-10 21:05:54 Bitcoin version v0.14.2
2017-07-10 21:05:54 InitParameterInteraction: parameter interaction: -whitelistforcerelay=1 -> setting -whitelistrelay=1
2017-07-10 21:05:54 Assuming ancestors of block 00000000000000000013176bf8d7dfeab4e1db31dc93bc311b436e82ab226b90 have valid signatures.
2017-07-10 21:05:54 Default data directory /Users/david/Library/Application Support/Bitcoin
2017-07-10 21:05:54 Using data directory /Volumes/SAMSUNG/Ledgers/BitCoinCore
2017-07-10 21:05:54 Using config file /Volumes/SAMSUNG/Ledgers/BitCoinCore/bitcoin.conf
2017-07-10 21:05:54 Using at most 125 automatic connections (2560 file descriptors available)
2017-07-10 21:05:54 Using 32 MiB out of 32 requested for signature cache, able to store 1048576 elements
2017-07-10 21:05:54 Using 2 threads for script verification
2017-07-10 21:05:54 scheduler thread start
2017-07-10 21:05:54 Using BerkeleyDB version Berkeley DB 4.8.30: (April  9, 2010)
2017-07-10 21:05:54 Using wallet wallet.dat
2017-07-10 21:05:54 init message: Verifying wallet...
[...]

PR #10922, opened 2017-07-24
New file-partition.md doc describing how to partition files to ensure fast initial blockchain synchronization..

jimhashhq:
After native build from source on Mac OS, my initial attempts to synchronize the blockchain were very very slow.  Upon finding [Issue Sync Taking Too Long](https://github.com/bitcoin/bitcoin/issues/10647), I found discussion by all and comments by @sipa in particular to be very useful, and reorganized $datadir folders on my local macOS build/install and summarized steps taken in file-partition.md doc.  These comments might find their audience more appropriately elsewhere, please feel free to suggest, thank you very much.
-jimhash
Note: This looks to be logged as issue: https://github.com/bitcoin/bitcoin/issues/10736

### DIFF
--- a/doc/file-partition.md
+++ b/doc/file-partition.md
@@ -45,7 +45,7 @@ Finally, set up soft links to restore the original folder structure:
       
           ln -s /Users/coinadm/local/bitcoin/index /Volumes/WD-Passport-Mac/bitcoin/blocks/index 
 
-6) Replace the original index folder location with a soft link:
+6) Replace the original blocks folder location with a soft link:
       
           ln -s /Volumes/WD-Passport-Mac/bitcoin/blocks /Users/coinadm/local/bitcoin/data/blocks
 
--- a/doc/file-partition.md
+++ b/doc/file-partition.md
@@ -4,14 +4,14 @@ Since the blockchain is around ~140GB, storage of large files on an external dri
 | Original Location       | New Location | Capacity Needs | i/o Frequency  |
 | ----------------------- | ------------ | -------------- | -------------- |
 | ${datadir}/blocks/index | ${datadir}   | low            | high           |
-| ${datadir}/blocks       | ${EXTERAL}   | high           | low            |
+| ${datadir}/blocks       | ${EXTERNAL}  | high           | low            |
 | ${datadir}/chainstate   | n/a          | low            | high           |
 
 Note: the chainstate folder will not move if these instructions are followed closely.  Like the index folder, chainstate contains very high frequency LevelDB files and must remain on a fast (internal) disk if reasonable syncronization time is to be expected.
 
 Change "coinadm" in the following as appropriate for your Bitcoin administrative account.
 
-1) Start bitcoind with --datadir pointing to your internal drive:
+1) Start bitcoind with -datadir pointing to your internal drive:
 
         bitcoind -daemon -conf=macos-bitcoin.conf -datadir=/Users/coinadm/local/bitcoin/data
 
@@ -25,7 +25,7 @@ Let it run for a few minutes, or long enough that it has started syncho-
 
        bitcoin-cli stop
 
-3) Move the "index" folder up one level (so that it may remain on the internal drive). Like the 
+3) Move the "index" folder up one level, so that it may remain on the internal drive: 
       
           mv -f /Users/coinadm/local/bitcoin/data/blocks/index /Users/coinadm/local/bitcoin/data/
 
@@ -39,7 +39,7 @@ Finally, set up soft links to restore the original folder structure:
 | Folder Name              | Link Name           |
 | ------------------------ | ------------------- |
 | ${EXTERNAL}/blocks/index | ${datadir}/../index |
-| ${datadir}/blocks        | ${EXTERAL}/blocks   |
+| ${datadir}/blocks        | ${EXTERNAL}/blocks  |
 
 5) Replace the original index folder location with a soft link:
       
--- a/doc/file-partition.md
+++ b/doc/file-partition.md
@@ -1,8 +1,12 @@
-Since the blockchain is around 130GB many developers may find it 
-convenient to store the blockchain on an external drive.  This
-document describes how to partitioning of datadir files between 
-the high-frequency/low-capacity "index" files and the low-frequency/high-capacity "blocks" files. 
-Examples are given for macOS, but Linux / Windows should be similar. 
+# File Partition:
+Since the blockchain is around 130GB, storage of large files on an external drive is convenient.  
+If this is not done properly though, the external drive will also contain high i/o-frequency LevelDB index
+files, protracting time for initial blockchain synchronization. This document describes how partition datadir files between the high-frequency/low-capacity "index" files and the low-frequency/high-capacity "blocks" files. Examples are given for macOS, but Linux / Windows should be similar. These instructions result in the following physical folder rearrangement:
+
+| Original Location       | New Location | Capacity Needs | IO Frequency        |
+| ----------------------- | ------------ | -------------- | ------------------- |
+| ${datadir}/blocks/index | ${datadir}   | low            | high                |
+| ${datadir}/blocks       | ${EXTERAL}   | high           | low                 |
 
 Change "username" in the follows as appropriate for your macOS Bitcoin
 administrative account ("coinadm" in this example conf file).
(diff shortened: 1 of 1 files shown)


sipa:
I think you missed something. There are three directories that matter:
* $DATADIR/blocks
* $DATADIR/blocks/index
* $DATADIR/chainstate

The first is high-bandwidth, low IO. The second has hardly any activity at all. The third is where all activity happens, and is critical for performance.

You should not separate the blocks from the blocks/index directory, as things may get ugly if they're out of sync. You can however put the chainstate on a faster/smaller device.

jimhashhq:
Hi @sipa, thank you very much for the prompt follow up.  Yes, the "chainstate" (in addition to "index") LevelDB files need to stay on the internal disk for reasonable synchronization performance.  The "file-partition.md" notes I made should describe:

- Installing $datadir to local internal disk (includig "chainstate").
- Start/stop bitcoind (though maybe this isn't necessary).
- Move the "index" files up one folder level. 
- Move the "blocks" folder off the internal disk.
- Go to the new external "blocks" folder and soft-link back to the new index location.
- Go to the internal "data" location and link the now external "blocks" folder.

Here, "chainstate" remains on the internal disk as never needs to move.

Again these "file-partition.md" notes are not the direct route I took, but (if correct) might save the next person considerable delay in initial sync.

To your point even though these low-capacity/high-IOfrequency LevelDb and high-capacity/low-IOfrequency blockchain files are physically separated on separate disks, they are kept in sync via the soft links created.

_Maybe ultimately a more natural configuration of these files would be to have the "index" folder up a level to start with._  That's what makes moving these folders confusing in my opinion.

Thank you again very much for the prompt feedback, very flattering, thanks.

jimhashhq:
I updated the doc to describe the need to keep "chainstate" on a fast internal disk as per @sipa, thank you.  Also, thanks @fanquake for adding the "Docs & Output" label.  This issue is probably also related to installation and/or configuration, but I did not see a label for that in the bitcoin project, perhaps that might be a useful label category?  Thank you both.

jimhashhq:
"file-partition.md"has been updated to highlight the need to keep "chainstate" folder on a fast (internal) drive in addition to "index" files as per @sipa.  Thanks @fanquake for adding the label.   This issue seems also related to "installation" and/or "configuration"; I did not see either of these as bitcoin project label categories, was thinking they might prove useful as well.  Thank you both for your guidance.

doc/file-partition.md
```diff
@@ -0,0 +1,53 @@
...
+
+Let it run for a few minutes, or long enough that it has started syncho-
+   nizing.  Either tail the debug.log, or watch the blocks folder:
+   
+        ls -l /Users/coinadm/local/bitcoin/data/blocks/blk00000.dat
+        -rw-------  1 coinadm  staff  134191893 Jul 27 11:31 /Users/coinadm/local/data/blocks/blk00000.dat
+
+2) Stop bitcoind, so that we can rearrange some datadir folders:
+
+       kill -QUIT `cat /Volumes/WD-Passport-Mac/bitcoin/data/bitcoind.pid`
```
achow101: Instead of killing the process, you should use `bitcoin-cli stop`.

doc/file-partition.md
```diff
@@ -0,0 +1,53 @@
...
+
+5) Replace the original index folder location with a soft link:
+      
+          ln -s /Users/coinadm/local/bitcoin/index /Volumes/WD-Passport-Mac/bitcoin/blocks/index 
+
+6) Replace the original index folder location with a soft link:
+      
+          ln -s /Volumes/WD-Passport-Mac/bitcoin/blocks /Users/coinadm/local/bitcoin/data/blocks
+
+ 注意 - Nota - Note - ध्यान दें - ﻢﻠﺣﻮﻇﺓ - метка
```
achow101: What's with the muptiple languages here?

jimhashhq: Again, good question, thank you.  To me, the "key finding" here (if any, really) is simply that the "index" folder by default is nested *within* the "blocks" folder, unlike the "chainstate" folder which is a sibling folder of "blocks".  This *slightly* complicates moving the "blocks" folder off of the internal disk to an external disk; my apologies for repeating the obvious.  Database administrators out there (me included) might argue that this nested folder configuration is less desirable as it complicates physical separation of high-capacity/low-frequency block files from the lower-capacity/higher-frequency index (and chainstate) LevelDB files.
When I open the hood to my car there are warnings labels on the radiator cap, etc., etc.., and these labels are in multiple languages to point out appropriate cautions to naive vehicle operators who may have never looked under the hood or checked a radiator.  I'm taking inspiration from this and wanted to (hopefully) say "Note" in 1/2 dozen or so most common languages by usage.  Also, I want to be especially friendly in this day and age; I feel like we could all use it.
In closing, I hope it doesn't sound like I am trying to "make a mountain out of a molehill" here, it's not that at all; I just wanted to share my personal experiences in hopes of further facilitating ease of use of the system.  Overall I find the system to be very easy to work with and well thought out.

sipa:
Your document is still suggesting to split the blocks/ directory from the blocks/index/ directory. Please don't do that; it's dangerous (they need to be in sync), and unnecessary (the blocks/index/ directory hardly sees any I/O). You should just suggest to move the chainstate/ to a faster drive compared to blocks/.

doc/file-partition.md
```diff
@@ -0,0 +1,53 @@
+# File Partition:
+Since the blockchain is around ~140GB, storage of large files on an external drive is convenient.  If this is not done properly, the external drive will also contain high i/o-frequency LevelDB index files, protracting time for initial blockchain synchronization. This document describes how partition datadir files between the high-frequency/low-capacity "index" files and the low-frequency/high-capacity "blocks" files. Examples are given for macOS, but Linux / Windows should be similar. These instructions result in the following physical folder rearrangement:
+
+| Original Location       | New Location | Capacity Needs | i/o Frequency  |
+| ----------------------- | ------------ | -------------- | -------------- |
+| ${datadir}/blocks/index | ${datadir}   | low            | high           |
+| ${datadir}/blocks       | ${EXTERAL}   | high           | low            |
+| ${datadir}/chainstate   | n/a          | low            | high           |
+
```
jimhashhq: There seems to be some agreement that the last two rows in the above table are correct.  At issue still is whether the *first row* is correct or not -- it does reflect my experience, and is the impetus for this doc.  I identified this possible re-configuration pitfall (which I mistakenly made myself) via bitcoind file usage reported by the following command:
     lsof -p `cat ${datadir}/bitcoind.pid` | grep ldb\$
This motivated me to move the subordinate/child "index" folder back onto the internal drive and set up the soft links described here.  I did not "go the extra mile" though and verify i/o frequency demands of these index LeveDB files.  If someone could chime in and corroborate either way that would be appreciated.  Is there a blockchain indexing or synchronization test that we can run with the index files both *on* the external drive and *off-of* the external drive while capturing i/o frequency statistics as well as wall clock time?  Arguably this is currently "seat of my pants"; I first moved the entire blocks folder (including the index subfolder) to the external disk for internal capacity reasons, and it synched super super slowly.  Then I looked at open files as described above and spotted those "LevelDB" index folder files.  Moving them as described here sped things up to the point that performance seemed to match that with the default configuration of the blocks folder on the internal drive.  Basically I was just doing a simplistic `ls -lrt blk*.dat` every so often and watching how fast the blk*.dat files were growing in both cases.

A possible misperception here is that this doc is meant to be some sort of performance advice.  It is not, rather it's really just meant to be a note to help other developers/analysts who (like me) want to work on very low end commodity hardware yet perform initial synchronization quickly, without the large-capacity internal disk space requirements.

Neither are these meant as instructions to backup the blockchain for portability between bitcoin development instances either; as @sipa points out, the "blocks" folder 'needs to be kept in sync with the "index" folder', and the blocks folder is useless by itself for sure.  

Missing from this document is the (?reasonable?) expectation that bitcoind not be started until the external disk is mounted and likewise that the external disk not be dismounted/ejected until bitcoind is shut down.   To this point, I have not tried to see what happens, if after running in the configuration suggested here, the operator/developer accidentally tries to start bitcoin w/o the the external storage plugged in; I would hope that the index folder files are not corrupted if the blocks folder is not accessible.

This is merely intended to be more of a "pitfalls to avoid" type of document, rather than some sort of tuning advice.  I hope that this perspective is of use.

doc/file-partition.md
```diff
@@ -0,0 +1,53 @@
...
+   
+          mkdir /Volumes/WD-Passport-Mac/bitcoin 
+          mv -f /Users/coinadm/local/bitcoin/data/blocks /Volumes/WD-Passport-Mac/bitcoin 
+
+Finally, set up soft links to restore the original folder structure:
+   
+| Folder Name              | Link Name           |
+| ------------------------ | ------------------- |
+| ${EXTERNAL}/blocks/index | ${datadir}/../index |
+| ${datadir}/blocks        | ${EXTERAL}/blocks   |
```
flack: typo: `EXTERAL`

laanwj:
> (the blocks/index/ directory hardly sees any I/O). 

Except with `-txindex` I guess :(

laanwj:
> Please don't do that; it's dangerous (they need to be in sync)

Also: sometimes there's the problem with the leveldb not supporting the filesystem that the `blocks/` directory points to. I think this happens with some network filesystems. If `blocks/index` is on the same partition that will never work :/

jimhashhq:
My apologies for not getting back sooner, I was ill but am feeling better.
I think my intent here was just to share/communicate my experiences with "symlink(ing) out the block-files (the large part)" as mentioned in issue #10787 referenced above, thanks.
I was hoping this experience might prove useful to others on low end commodity hardware who wish to store the blocks on an external USB.  Thanks!

doc/file-partition.md
```diff
@@ -0,0 +1,53 @@
...
+| Folder Name              | Link Name           |
+| ------------------------ | ------------------- |
+| ${EXTERNAL}/blocks/index | ${datadir}/../index |
+| ${datadir}/blocks        | ${EXTERNAL}/blocks  |
+
+5) Replace the original index folder location with a soft link:
+      
+          ln -s /Users/coinadm/local/bitcoin/index /Volumes/WD-Passport-Mac/bitcoin/blocks/index 
+
+6) Replace the original index folder location with a soft link:
```
arowser: should be "block folder"?

jimhashhq: Corrected, thanks!

laanwj:
> I was hoping this experience might prove useful to others on low end commodity hardware who wish to store the blocks on an external USB. Thanks!

Why close?

jimhashhq:
I got my GitHub workflow wires crossed, reopening thank you.

PR #10931, opened 2017-07-26
Fix misleading "Method not found" multiwallet errors

ryanofsky:
Raise RPC_WALLET_NOT_SPECIFIED instead of RPC_METHOD_NOT_FOUND when a required
wallet filename was not specified in an RPC call.

Also raise more specific RPC_WALLET_NOT_FOUND error instead of
RPC_INVALID_PARAMETER in case an invalid wallet was specified, for consistency.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -43,7 +43,7 @@ CWallet *GetWalletForJSONRPCRequest(const JSONRPCRequest& request)
                 return pwallet;
             }
         }
-        throw JSONRPCError(RPC_INVALID_PARAMETER, "Requested wallet does not exist or is not loaded");
+        throw JSONRPCError(RPC_WALLET_NOT_FOUND, "Requested wallet does not exist or is not loaded");
     }
     return ::vpwallets.size() == 1 || (request.fHelp && ::vpwallets.size() > 0) ? ::vpwallets[0] : nullptr;
 }
@@ -57,13 +57,14 @@ std::string HelpRequiringPassphrase(CWallet * const pwallet)
 
 bool EnsureWalletIsAvailable(CWallet * const pwallet, bool avoidException)
 {
-    if (!pwallet) {
-        if (!avoidException)
-            throw JSONRPCError(RPC_METHOD_NOT_FOUND, "Method not found (disabled)");
-        else
-            return false;
+    if (pwallet) return true;
+    if (avoidException) return false;
+    if (::vpwallets.empty()) {
+        // Wallet RPC methods are disabled if no wallets are loaded.
+        throw JSONRPCError(RPC_METHOD_NOT_FOUND, "Method not found (disabled)");
     }
-    return true;
+    throw JSONRPCError(RPC_WALLET_NOT_SPECIFIED,
+        "Wallet file not specified (must request wallet RPC through /wallet/<filename> uri-path).");
 }
 
 void EnsureWalletIsUnlocked(CWallet * const pwallet)
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -60,8 +60,13 @@ bool EnsureWalletIsAvailable(CWallet * const pwallet, bool avoidException)
     if (pwallet) return true;
     if (avoidException) return false;
     if (::vpwallets.empty()) {
-        // Wallet RPC methods are disabled if no wallets are loaded.
-        throw JSONRPCError(RPC_METHOD_NOT_FOUND, "Method not found (disabled)");
+        // Note: It isn't currently possible to trigger this error because
+        // wallet RPC methods aren't registered unless a wallet is loaded. But
+        // this error is being kept as a precaution, because it's possible in
+        // the future that wallet RPC methods might get or remain registered
+        // when no wallets are loaded.
+        throw JSONRPCError(
+            RPC_METHOD_NOT_FOUND, "Method not found (wallet method is disabled because no wallet is loaded)");
     }
     throw JSONRPCError(RPC_WALLET_NOT_SPECIFIED,
         "Wallet file not specified (must request wallet RPC through /wallet/<filename> uri-path).");
--- a/src/bitcoin-cli.cpp
+++ b/src/bitcoin-cli.cpp
@@ -337,6 +337,10 @@ int CommandLineRPC(int argc, char *argv[])
 
                         if (errMsg.isStr())
                             strPrint += "error message:\n"+errMsg.get_str();
+
+                        if (errCode.isNum() && errCode.get_int() == RPC_WALLET_NOT_SPECIFIED) {
+                            strPrint += "\nTry adding \"-rpcwallet=<filename>\" option to bitcoin-cli command line.";
+                        }
                     }
                 } else {
                     // Result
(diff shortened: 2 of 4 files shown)


ryanofsky:
Should tag for 0.15 since this fixes a new error and wouldn't be backwards compatible if it was in a later release.

sipa:
utACK 5bb4fe6b3f7c8ced61902615edee708667cbae33

src/wallet/rpcwallet.cpp
```diff
@@ -57,13 +57,15 @@ std::string HelpRequiringPassphrase(CWallet * const pwallet)
...
-            throw JSONRPCError(RPC_METHOD_NOT_FOUND, "Method not found (disabled)");
-        else
-            return false;
-    }
-    return true;
+    if (pwallet) return true;
+    if (avoidException) return false;
+    if (::vpwallets.empty()) {
+        // Wallet RPC methods are disabled if no wallets are loaded.
+        throw JSONRPCError(RPC_METHOD_NOT_FOUND, "Method not found (disabled)");
```
jnewbery: Any chance of changing the error message to `Method not found (wallet disabled)`?

jnewbery (commented):
Definite concept ACK, and agree that this should go into v0.15 as a bugfix.

I'll test tomorrow.

src/wallet/rpcwallet.cpp
```diff
@@ -57,13 +57,15 @@ std::string HelpRequiringPassphrase(CWallet * const pwallet)
...
+    }
+    throw JSONRPCError(RPC_WALLET_NOT_SPECIFIED, "Wallet file not specified (please provide bitcoin-cli "
```
jonasschnelli: Not sure if we should mention a specific RPC client (bitcoin-cli) on the server side.

jonasschnelli:
utACK 5bb4fe6b3f7c8ced61902615edee708667cbae33

src/wallet/rpcwallet.cpp
laanwj: Or "no wallets loaded", I think that's more relevant now, especially when runtime loading/unloading is added later

src/wallet/rpcwallet.cpp
laanwj: Agree with @jonasschnelli, we shouldn't mention bitcoin-cli in server errors.
Nothing prevents you from catching this specific error code in the client and giving a specific error though, if you want to be user-friendly.

src/wallet/rpcwallet.cpp
ryanofsky: Thread https://github.com/bitcoin/bitcoin/pull/10931#discussion_r129523302

Ok, added in 4cfa4442aa53d9c86d1596e05dcd1608d378120f

ryanofsky (commented):
Added a commit 5bb4fe6b3f7c8ced61902615edee708667cbae33 -> 4cfa4442aa53d9c86d1596e05dcd1608d378120f ([pr/multierr.1](https://github.com/ryanofsky/bitcoin/commits/pr/multierr.1) -> [pr/multierr.2](https://github.com/ryanofsky/bitcoin/commits/pr/multierr.2), [compare](https://github.com/ryanofsky/bitcoin/compare/pr/multierr.1...pr/multierr.2))

gmaxwell (approved):
utACK

achow101:
utACK 4cfa4442aa53d9c86d1596e05dcd1608d378120f

src/wallet/rpcwallet.cpp
jnewbery: Actually it appears that it's not currently possible to hit this error. If bitcoind is started with `disablewallet` then the wallet RPC commands aren't registered and any call to a wallet RPC will be rejected here: https://github.com/bitcoin/bitcoin/blob/5c8eb7916de7030d6610df53977b3068342c8f24/src/rpc/server.cpp#L495. I don't think there's currently a way to start bitcoind with the wallet component loaded but no wallets loaded.

Code looks good and will be required when dynamic wallet loading is implemented, but is not currently testable. Suggest you leave the code in, but add a comment.

src/wallet/rpcwallet.cpp
jnewbery: I almost wrote the same comment as Jonas about server not having bitcoin-cli specific messages, but Russ has convinced me. I think it's ok to have a bitcoin-cli message here.

jnewbery:
Tested ACK 4cfa4442aa53d9c86d1596e05dcd1608d378120f

src/wallet/rpcwallet.cpp
laanwj: NACK. I still disagree. People use RPC from many other programs than just bitcoin-cli. Mentioning a -cli argument is just confusing. If you want to add documentation in an error message, you should document how people should change their API usage instead.

> how the server mentioning a bitcoin-cli option is different than server providing bitcoin-cli command lines in the help text of all rpc methods

That's IMO different because it's a clearly isolated context that mentions use of bitcoin-cli, not a random error.

src/wallet/rpcwallet.cpp
```diff
@@ -57,13 +57,16 @@ std::string HelpRequiringPassphrase(CWallet * const pwallet)
...
-    }
-    return true;
+    if (pwallet) return true;
+    if (avoidException) return false;
+    if (::vpwallets.empty()) {
+        // Wallet RPC methods are disabled if no wallets are loaded.
+        throw JSONRPCError(
+            RPC_METHOD_NOT_FOUND, "Method not found (wallet method is disabled because no wallet is loaded)");
+    }
+    throw JSONRPCError(RPC_WALLET_NOT_SPECIFIED, "Wallet file not specified (please provide bitcoin-cli "
```
laanwj: Somehow this got hidden: https://github.com/bitcoin/bitcoin/pull/10931#discussion_r129526990

I still disagree with mentioning -cli here. If there is to be extenxsive documentation here, this should mention how to change the API usage, as people will be using this API from many applications and not just bitcoin-cli.

jonasschnelli: Agree with @laanwj. Please remove the `bitcoin-cli` part.
By design, it makes no sense to report client specific error message in a pure JSON RPC server implementation.

ryanofsky (commented):
Added 2 commits 4cfa4442aa53d9c86d1596e05dcd1608d378120f -> 6b5c343a49a25085bf1d6300cea2d40cebc993b6 ([pr/multierr.2](https://github.com/ryanofsky/bitcoin/commits/pr/multierr.2) -> [pr/multierr.3](https://github.com/ryanofsky/bitcoin/commits/pr/multierr.3), [compare](https://github.com/ryanofsky/bitcoin/compare/pr/multierr.2...pr/multierr.3))

Squashed 6b5c343a49a25085bf1d6300cea2d40cebc993b6 -> df389bca2048f5814387987463fe3298766bc6b6 ([pr/multierr.3](https://github.com/ryanofsky/bitcoin/commits/pr/multierr.3) -> [pr/multierr.4](https://github.com/ryanofsky/bitcoin/commits/pr/multierr.4))

src/wallet/rpcwallet.cpp
laanwj: In a way, yes, though I don't think this is much less maintainable; we have those error codes for a reason, might as well use them to print custom text at the client side. We could at some point provide a better help text for other codes as well, where it makes sense.
In any case thanks for implementing this.

laanwj:
utACK df389bc

PR #10939, opened 2017-07-26
[init] Check non-emptiness of -blocknotify command prior to executing

practicalswift:
Check that `-blocknotify` command is non-empty before executing.

To make the `BlockNotifyCallback(...)` (`-blocknotify`) behaviour consistent with that of:
* `AlertNotify(...)` (`-alertnotify`)
* `AddToWallet(...)` (`-walletnotify`)

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -547,9 +547,10 @@ static void BlockNotifyCallback(bool initialSync, const CBlockIndex *pBlockIndex
         return;
 
     std::string strCmd = gArgs.GetArg("-blocknotify", "");
-
-    boost::replace_all(strCmd, "%s", pBlockIndex->GetBlockHash().GetHex());
-    boost::thread t(runCommand, strCmd); // thread runs free
+    if (!strCmd.empty()) {
+        boost::replace_all(strCmd, "%s", pBlockIndex->GetBlockHash().GetHex());
+        boost::thread t(runCommand, strCmd); // thread runs free
+    }
 }
 
 static bool fHaveGenesis = false;
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -988,7 +988,7 @@ bool CWallet::AddToWallet(const CWalletTx& wtxIn, bool fFlushOnClose)
     // notify an external script when a wallet transaction comes in or is updated
     std::string strCmd = gArgs.GetArg("-walletnotify", "");
 
-    if ( !strCmd.empty())
+    if (!strCmd.empty())
     {
         boost::replace_all(strCmd, "%s", wtxIn.GetHash().GetHex());
         boost::thread t(runCommand, strCmd); // thread runs free
--- a/src/util.cpp
+++ b/src/util.cpp
@@ -812,6 +812,7 @@ fs::path GetSpecialFolderPath(int nFolder, bool fCreate)
 
 void runCommand(const std::string& strCommand)
 {
+    if (strCommand.empty()) return;
     int nErr = ::system(strCommand.c_str());
     if (nErr)
         LogPrintf("runCommand error: system(%s) returned %d\n", strCommand, nErr);


src/wallet/wallet.cpp
```diff
@@ -940,7 +940,7 @@ bool CWallet::AddToWallet(const CWalletTx& wtxIn, bool fFlushOnClose)
     // notify an external script when a wallet transaction comes in or is updated
     std::string strCmd = GetArg("-walletnotify", "");
 
-    if ( !strCmd.empty())
+    if (!strCmd.empty())
```
promag: Unrelated.

src/init.cpp
```diff
@@ -545,9 +545,10 @@ static void BlockNotifyCallback(bool initialSync, const CBlockIndex *pBlockIndex
         return;
 
     std::string strCmd = GetArg("-blocknotify", "");
-
-    boost::replace_all(strCmd, "%s", pBlockIndex->GetBlockHash().GetHex());
-    boost::thread t(runCommand, strCmd); // thread runs free
+    if (!strCmd.empty()) {
```
promag: No need to test because in `init.cpp` there's:
```cpp
    if (IsArgSet("-blocknotify"))
        uiInterface.NotifyBlockTip.connect(BlockNotifyCallback);
```

promag (commented):
Partial ACK. IMO the check should be in init.cpp:
```
if (!GetArg("-blocknotify", "").empty()) {
    uiInterface.NotifyBlockTip.connect(BlockNotifyCallback);
}
```
That way it's checked one time only.

src/wallet/wallet.cpp
practicalswift: That change was intentional to show how the logic is handled in `AddToWallet`.

src/init.cpp
practicalswift: Does `IsArgSet(…)` test for emptiness?

practicalswift:
@promag It is better to do the check as close as possible to the `RunCommand(…)` since `BlockNotifyCallback(…)` might be invoked from other places in the future.

src/init.cpp
promag: Ignore if you want to keep the test here.

promag:
Side note, should we treat empty args as invalid?

practicalswift:
@promag In the case of `-{alert,block,wallet}notify`: yes - no need to execute `""` :-)

laanwj:
> Side note, should we treat empty args as invalid?

Depends on the specific argument. For some things, an empty string might be a valid value, for others it's not.

For executing a command it makes sense to skip it when the string is empty.
What does `system("")` do? spawn a shell?

BTW: Why not handle this in `runCommand`? that seems the shortest path to making this consistent.

laanwj:
> What does system("") do? spawn a shell?

It already does nothing! Same for `system("     ");` etc.
Why do we need this change?

promag:
@laanwj the only reason to test `""` in each is to add some custom logging. Otherwise agree with you (despite the fact that we can avoid connecting to the signal in init).

laanwj:
Not convinced that "custom logging" makes sense here. The user provided an empty command, which non-ambiguously means they don't want it to do anything. It doesn't do anything.

(maybe skipping could be considered a small optimization, but could leave it up to the user to just not provide the argument at all so that the signal doesn't get registered?)

practicalswift:
@laanwj My original idea was to make `-blocknotify`, `-alertnotify` and `-walletnotify` behave identically with regards to thread creation and `execve(…)`-calls in case of an `""` argument. Not a big issue, just that interface consistency is nice :-)

promag:
@laanwj IMO empty commands should be invalid. The same goes for when a command does not exists. Not providing the command should be the correct way to run nothing.

For instance, with `find` it fails:
```
$ find . -exec \;
find: -exec: no command specified
```

jnewbery:
<slightly offtopic>

@promag - I'm with you on this one. I'd like there to be better systematic validation of command line-arguments.

I think you've spoken about argument validation on a few issues. Perhaps we should raise this in a future meeting to get concept feedback?

promag:
Yes @jnewbery, IMO it can be improved. We should validate command line arguments in the same way we validate RPC arguments. Again, for me `""` is an invalid command.

luke-jr (approved):
utACK

luke-jr:
We shouldn't make the empty command invalid so quickly: if one has `blocknotify=foobar` in the `bitcoin.conf` file, they may wish to use `-blocknotify=` on the commandline to temporarily disable it.

(The alternative, `-noblocknotify`, is probably broken...)

promag:
> they may wish to use -blocknotify= 

They can remove the line or comment it out?

luke-jr:
Better if no change to the config file is needed for such a temporary purpose.

jnewbery:
@luke-jr raises a good point. Being able to override and temporarily disable .conf file options from the command line is very useful. This PR allows that (supplying `blocknotify=` won't cause an error).

Has anyone tested what happens with `noblocknotify`?

laanwj:
> We shouldn't make the empty command invalid so quickly: if one has blocknotify=foobar in the bitcoin.conf file, they may wish to use -blocknotify= on the commandline to temporarily disable it.

Good point! And the first good argument for doing this at all. Being able to overriding the blocknotify command with "" (say, from the command line) to disable it is useful. In that case you don't want an empty command to be executed.

utACK cffe85f

PR #10941, opened 2017-07-27
Add blocknotify and walletnotify functional tests

promag:
This patch adds the missing functional tests for `-blocknotify` and `-walletnotify` notifications. The `-alertnotify` test file `forknotify.py` is renamed to `notifications.py` to accommodate the new tests. Credits to @jnewbery for this cleanup and unification.

### DIFF
--- a/test/functional/notifications.py
+++ b/test/functional/notifications.py
@@ -2,23 +2,29 @@
 # Copyright (c) 2014-2016 The Bitcoin Core developers
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
-"""Test the -alertnotify and -blocknotify options."""
+"""Test the -alertnotify, -blocknotify and -walletnotify options."""
 import os
 
 from test_framework.test_framework import BitcoinTestFramework
-from test_framework.util import assert_equal, wait_until
+from test_framework.util import assert_equal, wait_until, connect_nodes_bi
 
 class NotificationsTest(BitcoinTestFramework):
     def set_test_params(self):
         self.num_nodes = 2
+        self.setup_clean_chain = True
 
     def setup_network(self):
         self.alert_filename = os.path.join(self.options.tmpdir, "alert.txt")
-        self.block_filename = os.path.join(self.options.tmpdir, 'blocks.txt')
+        self.block_filename = os.path.join(self.options.tmpdir, "blocks.txt")
+        self.tx_filename = os.path.join(self.options.tmpdir, "transactions.txt")
+
+        # -alertnotify and -blocknotify on node0, walletnotify on node1
         self.extra_args = [["-blockversion=2",
                             "-alertnotify=echo %%s >> %s" % self.alert_filename,
                             "-blocknotify=echo %%s >> %s" % self.block_filename],
-                           ["-blockversion=211"]]
+                           ["-blockversion=211",
+                            "-rescan",
+                            "-walletnotify=echo %%s >> %s" % self.tx_filename]]
         super().setup_network()
 
     def run_test(self):
@@ -32,6 +38,28 @@ class NotificationsTest(BitcoinTestFramework):
         # file content should equal the generated blocks hashes
         with open(self.block_filename, 'r') as f:
             assert_equal(sorted(blocks), sorted(f.read().splitlines()))
+
+        self.log.info("test -walletnotify")
+        # wait at most 10 seconds for expected file size before reading the content
+        wait_until(lambda: os.path.isfile(self.tx_filename) and os.stat(self.tx_filename).st_size >= (block_count * 65), timeout=10)
+
+        # file content should equal the generated transaction hashes
+        txids_rpc = list(map(lambda t: t['txid'], self.nodes[1].listtransactions("*", block_count)))
+        with open(self.tx_filename, 'r') as f:
+            assert_equal(sorted(txids_rpc), sorted(f.read().splitlines()))
+        os.remove(self.tx_filename)
+
+        self.log.info("test -walletnotify after rescan")
+        # restart node to rescan to force wallet notifications
+        self.restart_node(1)
+        connect_nodes_bi(self.nodes, 0, 1)
+
+        wait_until(lambda: os.path.isfile(self.tx_filename) and os.stat(self.tx_filename).st_size >= (block_count * 65), timeout=10)
+
+        # file content should equal the generated transaction hashes
+        txids_rpc = list(map(lambda t: t['txid'], self.nodes[1].listtransactions("*", block_count)))
+        with open(self.tx_filename, 'r') as f:
+            assert_equal(sorted(txids_rpc), sorted(f.read().splitlines()))
 
         # Mine another 41 up-version blocks. -alertnotify should trigger on the 51st.
         self.log.info("test -alertnotify")
--- a/test/functional/notifications.py
+++ b/test/functional/notifications.py
@@ -2,25 +2,40 @@
 # Copyright (c) 2014-2016 The Bitcoin Core developers
 # Distributed under the MIT software license, see the accompanying
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
-"""Test the -alertnotify option."""
+"""Test the -alertnotify and -blocknotify options."""
 import os
 
 from test_framework.test_framework import BitcoinTestFramework
 from test_framework.util import assert_equal, wait_until
 
-class ForkNotifyTest(BitcoinTestFramework):
+class NotificationsTest(BitcoinTestFramework):
     def set_test_params(self):
         self.num_nodes = 2
 
     def setup_network(self):
         self.alert_filename = os.path.join(self.options.tmpdir, "alert.txt")
-        self.extra_args = [["-alertnotify=echo %%s >> %s" % self.alert_filename],
+        self.block_filename = os.path.join(self.options.tmpdir, 'blocks.txt')
+        self.extra_args = [["-blockversion=2",
+                            "-alertnotify=echo %%s >> %s" % self.alert_filename,
+                            "-blocknotify=echo %%s >> %s" % self.block_filename],
                            ["-blockversion=211"]]
         super().setup_network()
 
     def run_test(self):
-        # Mine 51 up-version blocks. -alertnotify should trigger on the 51st.
-        self.nodes[1].generate(51)
+        self.log.info("test -blocknotify")
+        block_count = 10
+        blocks = self.nodes[1].generate(block_count)
+
+        # wait at most 10 seconds for expected file size before reading the content
+        wait_until(lambda: os.path.isfile(self.block_filename) and os.stat(self.block_filename).st_size >= (block_count * 65), timeout=10)
+
+        # file content should equal the generated blocks hashes
+        with open(self.block_filename, 'r') as f:
+            assert_equal(sorted(blocks), sorted(f.read().splitlines()))
+
+        # Mine another 41 up-version blocks. -alertnotify should trigger on the 51st.
+        self.log.info("test -alertnotify")
+        self.nodes[1].generate(41)
         self.sync_all()
 
         # Give bitcoind 10 seconds to write the alert notification
@@ -40,4 +55,4 @@ class ForkNotifyTest(BitcoinTestFramework):
         assert_equal(alert_text, alert_text2)
 
 if __name__ == '__main__':
-    ForkNotifyTest().main()
+    NotificationsTest().main()
--- a/test/functional/forknotify.py
+++ b/test/functional/forknotify.py
@@ -4,9 +4,9 @@
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 """Test the -alertnotify option."""
 import os
-import time
 
 from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal, wait_until
 
 class ForkNotifyTest(BitcoinTestFramework):
     def set_test_params(self):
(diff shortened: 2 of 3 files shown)


test/functional/blocknotify.py
```diff
@@ -0,0 +1,42 @@
...
+
+        self.path = path
+        self.setup_clean_chain = False
+        self.num_nodes = 1
+        self.extra_args = [['-blocknotify=echo %%s >> %s' % path]]
+
+    def run_test(self):
+        # generate 5 blocks
+        blocks = self.nodes[0].generate(5)
+        sleep(1)
```
promag: Without these sometimes the file doesn't have all blocks because `runCommand()` is aync. Any idea to avoid the sleep without pooling or blocking reads?

test/functional/blocknotify.py
```diff
@@ -0,0 +1,42 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Test blocknotify.
+
+Verify that a bitcoind node calls the given command for each block
+"""
+from os import close, remove
```
practicalswift: Replace with `import os`. `os.close` and `os.remove` is used below :-)

test/functional/blocknotify.py
```diff
@@ -0,0 +1,42 @@
...
+from tempfile import mkstemp
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal, assert_raises_jsonrpc
```
practicalswift: Remove `assert_raises_jsonrpc`. Not used :-)

test/functional/blocknotify.py
```diff
@@ -0,0 +1,42 @@
...
+from time import sleep
+
+class BlockNotifyTest(BitcoinTestFramework):
+
+    def __init__(self):
+        super().__init__()
+
+        # get a temporary file path to store the notified blocks
+        (fd, path) = mkstemp(text = True)
```
practicalswift: Nit: `fd, path = mkstemp(text = True)` will do :-)

practicalswift:
Concept ACK :-)

test/functional/blocknotify.py
promag: Lol yes. What do you guys prefer? Import module or symbols?

practicalswift: We have a winner:

```
$ git grep -E "(import os|from os)" -- "*.py" | cut -f2 -d: | sort | uniq -c
     25 import os
      2 import os.path
```

promag:
@practicalswift thanks for the review. Added a few comments and fixed the nits.

test/functional/blocknotify.py
```diff
@@ -0,0 +1,49 @@
...
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal
+
+class BlockNotifyTest(BitcoinTestFramework):
+
+    def __init__(self):
+        super().__init__()
+
+        # get a temporary file path to store the notified blocks
+        fd, path = tempfile.mkstemp(text = True)
```
MarcoFalke: Please use `self.options.tmpdir`, so that files are contained in a single tree for each test.

Also nit: Please no whitespace for named args. (`text=True`)

test/functional/blocknotify.py
```diff
@@ -0,0 +1,49 @@
...
+
+        # the command will append the block hash to file
+        self.extra_args = [['-blocknotify=echo -n %%s >> %s' % path]]
+
+    def run_test(self):
+        # generate 3 blocks
+        blocks = self.nodes[0].generate(3)
+
+        # wait while the command runs asynchronously
+        time.sleep(3)
```
MarcoFalke: Please no fixed sleeps. This might be too much time on a fast system or too less time on a slow io system.

You could wrap it into a while loop with timeout on failure and break on success. Alternatively there should be some helper function `wait_until` if I am not mistaken.

maflcko (commented):
Concept ACK, but needs travis fixed:

```
WARNING! The following scripts are not being run: ['blocknotify.py']. Check the test lists in test_runner.py.

File: test/functional/blocknotify.py [hunk 7]
```diff
@@ -0,0 +1,49 @@
...
+
+        # file content should equal the generated blocks hashes
+        with open(self.path, 'r') as f:
+            assert_equal(''.join(blocks), f.read())
+
```
jnewbery: nit: whitespace.

Consider using a linter to catch nits.

test/functional/blocknotify.py
```diff
@@ -0,0 +1,49 @@
...
+class BlockNotifyTest(BitcoinTestFramework):
+
+    def __init__(self):
+        super().__init__()
+
+        # get a temporary file path to store the notified blocks
+        fd, path = tempfile.mkstemp(text = True)
+        os.close(fd)
+
+        self.path = path
```
jnewbery: can assign `self.path` above rather than assign to a local variable `path`

jnewbery (commented):
A couple of nits to add to @MarcoFalke's

You'll need to add `blocknotify.py` to `BASE_SCRIPTS` in test_runner.py

test/functional/blocknotify.py
promag: `self.options` is not defined at this point.

promag:
@MarcoFalke done. However can't get tmpdir in the constructor. Any other suggestion?

MarcoFalke:
You could try to overwrite setup_network:

```py
def setup_network():
    self.extra_args = [bla]
    super().setup_network()
```

promag:
@MarcoFalke ty, all comments addressed.

practicalswift:
utACK d2c730c77f9b4011040e30f2bc2a4c8c6deb13f2

test/functional/blocknotify.py
```diff
@@ -0,0 +1,47 @@
...
+
+        # the command will append the notified block hash to file
+        self.extra_args = [['-blocknotify=echo %%s >> %s' % self.path]]
+        super().setup_network()
+
+    def run_test(self):
+        block_count = 10
+        blocks = self.nodes[0].generate(block_count)
+
+        # wait for expected data length
```
jnewbery: A few comments on this loop:

- I'd make the sleep longer (say 0.1 seconds). I'm also not sure you need the back-off. Just a `time.sleep(0.1)` should be fine.
- The total time allowed for the file to be written is 0.45 seconds. That needs to be longer (up to 10 seconds) since Travis can block i/o for many seconds at a time. See #10072.
- Please make the comment clear about the total time allowed for the loop.

jnewbery (commented):
tested ACK with one change needed to the timing of the while loop

promag:
@jnewbery done. Also rebased.

MarcoFalke:
Needs rebase

promag:
@MarcoFalke done.

test/functional/blocknotify.py
```diff
@@ -0,0 +1,47 @@
...
+        self.extra_args = [['-blocknotify=echo %%s >> %s' % self.path]]
+        super().setup_network()
+
+    def run_test(self):
+        block_count = 10
+        blocks = self.nodes[0].generate(block_count)
+
+        # wait at most 10 seconds for expected file size before reading the content
+        attempts = 0
+        while os.stat(self.path).st_size < (block_count * 65) and attempts < 100:
```
MarcoFalke: Maybe wait until the file is created? I'd really suggest using some sort of wrapper, like the `wait_until` from `mininode`.

```py
wait_until(lambda: os.path.isfile(self.path), timeout=1)
wait_until(lambda: os.stat(self.path).st_size >= (block_count * 65), timeout=10)

**Comment by jnewbery** [comment]
I still believe that this test should be merged with #10966 and the `forknotify.py` test. They're testing very similar functionality and there's a lot of shared code between the three tests.

File: test/functional/blocknotify.py (hunk 10 above)
Comment (MarcoFalke) [inline test/functional/blocknotify.py]: Just to make sure, that this is causing failures on my machine:

```py
  File "./test/functional/blocknotify.py", line 38, in run_test
    while os.stat(self.path).st_size < (block_count * 65) and attempts < 100:
FileNotFoundError: [Errno 2] No such file or directory: '/tmp/bitcoin_test_runner_20170815_171120/blocknotify_621/blocks.txt'

Comment (MarcoFalke) [reply test/functional/blocknotify.py]: ping @promag: Are you still working on this?

File: test/functional/blocknotify.py [hunk 11]
```diff
@@ -0,0 +1,47 @@
...
+        super().__init__()
+
+        self.setup_clean_chain = False
+        self.num_nodes = 1
+
+    def setup_network(self):
+        self.path = os.path.join(self.options.tmpdir, 'blocks.txt')
+
+        # the command will append the notified block hash to file
+        self.extra_args = [['-blocknotify=echo %%s >> %s' % self.path]]
```
ryanofsky: Maybe escape path with `shlex.quote(self.path)`

test/functional/blocknotify.py
ryanofsky: > ping @promag: Are you still working on this?

I'm not seeing any errors when I run the test, but this is probably caused by a race between the start of this loop and the first notification. Could be fixed by adding a `not os.path.exists(self.path)` check here, or adding `with open(self.path, "w"): pass` to `setup_network`.

ryanofsky (commented):
Tested ACK 2b03bc1fb844980ffc67c6eaa52b034e978a9826 as long as the race condition is fixed. 

I agree with others the test setup should be merged with  walletnotify test setup in #10966 so code is not duplicated, but I don't think it matters if this is done in one or two prs.

test/functional/blocknotify.py
MarcoFalke: Indeed. The code I posted a few comments above fixed it for me.

c.f.

```py
wait_until(lambda: os.path.isfile(self.path), timeout=1)
wait_until(lambda: os.stat(self.path).st_size >= (block_count * 65), timeout=10)

**Comment by promag** [comment]
As suggested I'll merge the test and use `waut_until`.

File: test/functional/blocknotify.py (hunk 11 above)
Comment (promag) [reply test/functional/blocknotify.py]: `tmpdir` is used elsewhere (should be secure) and `blocks.txt` doesn't need to be quoted. Is there a strong argument for it?

**Comment by promag** [comment]
Rebased and updated as per suggestions. Still in a separate test file because IMO that should go to a different PR since `forknotify.py` needs to be changed.

File: test/functional/blocknotify.py (hunk 11 above)
Comment (promag) [reply test/functional/blocknotify.py]: Of you mean to quote the command `echo %%s >> %s`?

Comment (ryanofsky) [reply test/functional/blocknotify.py]: Quoting the entire command would be incorrect because it is passed literally to [system(3)](http://man7.org/linux/man-pages/man3/system.3.html). Quoting the path is correct because it is being passed as a shell token. Anyway just suggested this in case you wanted the code to be more correct. There are no practical implications (security or otherwise) either way.

Comment (promag) [reply test/functional/blocknotify.py]: Right, otherwise it wouldn't work 😃 I'm going for simplicity if you don't mind.

**Comment by jnewbery** [comment]
Please see https://github.com/jnewbery/bitcoin/tree/fork_notify. This adds blocknotify and walletnotify to the existint fork_notify.py test script.

File: test/functional/blocknotify.py [hunk 12]
```diff
@@ -0,0 +1,40 @@
...
+        block_count = 10
+        blocks = self.nodes[0].generate(block_count)
+
+        wait_until(lambda: os.path.isfile(self.path), timeout=1)
+        # wait at most 10 seconds for expected file size before reading the content
+        wait_until(lambda: os.stat(self.path).st_size >= (block_count * 65), timeout=10)
+
+        # file content should equal the generated blocks hashes
+        with open(self.path, 'r') as f:
+            assert_equal(blocks, f.read().splitlines())
```
MarcoFalke: Hmm. I don't think boost::thread guarantees that the thread run in the order they are constructed. Am I wrong?

The arrays should have the same content, but a different order.

maflcko (commented):
utACK e7408a444 but needs clarification

test/functional/blocknotify.py
promag: Correct, in `-walletnotify` (#10966) there's:
```cpp
            # txids are sorted because there is no order guarantee in notifications
            assert_equal(sorted(txids_rpc), sorted(txids_file))
```
So I'll do the same here for correctness.

promag:
Picked your branch and added a commit that adds `-rescan` to `node[1]` (needs squash).

@MarcoFalke the block sort was already on the @jnewbery branch.

jnewbery:
Tested ACK 6d9c1593497e9f295b5f335c7c7a0df7966d7590. As you've pointed out, should be squashed prior to merge.

promag:
@jnewbery Done.

mess110:
Is the name of the test class still correct? I think it should be something like `NotificationsTest`

test/functional/notifications.py
```diff
@@ -0,0 +1,86 @@
+#!/usr/bin/env python3
+# Copyright (c) 2014-2016 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Test the -alertnotify and -blocknotify options."""
```
mess110: You are also testing -walletnotify, could you please update the comment?

test/functional/notifications.py
```diff
@@ -0,0 +1,86 @@
...
+import os
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal, wait_until, connect_nodes_bi
+
+class ForkNotifyTest(BitcoinTestFramework):
```
mess110: Is the class name still correct? I think it should be something like `NotificationsTest`

promag:
@mess110 Done, thank you.

mess110:
ACK https://github.com/bitcoin/bitcoin/pull/10941/commits/857b32b4b280f13997cf2fa471802ad6a27075fb

(or the squashed version of this)

promag:
I think makes sense to have individual commits.

laanwj:
> I think makes sense to have individual commits.

Certainly if multiple authors are involved. And the different commits concern different aspects.

What we want to avoid with squashing is sequential commits like "fix the previous commit" which are internal to the author's working and are only confusing when reading the history. But that's not the case here.

PR #10946, opened 2017-07-28
Add chainwork to getchaintxstats

sipa:
This makes it easier to create and verify the `nMinimumChainWork` data in chain parameters.

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -1528,6 +1528,7 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
     ret.push_back(Pair("time", (int64_t)pindex->nTime));
     ret.push_back(Pair("txcount", (int64_t)pindex->nChainTx));
     ret.push_back(Pair("txrate", ((double)nTxDiff) / nTimeDiff));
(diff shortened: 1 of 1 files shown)


gmaxwell:
Hm? getblockchaininfo gives it to you, and this is what the readme tells you to use.

sipa:
It only gives it you for the tip. `getchaintxstats` works for any block.

promag (commented):
Can you add the new field in the functional test in `test/functional/blockchain.py`?

jnewbery:
What's the reason for wanting chainwork for any arbitrary block? Are you expecting to use nMinimumChainWork and defaultAssumeValid from the same block? Or is it so you can verify you have the exact same value as someone else who's run the release process?

sipa:
@jnewbery Yes, ability to repeatably come up with the same numbers, even if the exact value doesn't matter.

This probably belongs more in `getblockheader` than in `getchaintxstats`, as it's not really a statistic about transactions. Closing.

achow101:
`getblockheader` already includes the chainwork for a block.

sipa:
Oops, totally unnecessary in that case.

PR #10947, opened 2017-07-28
[Wallet] Bare segwit scriptPubKey should not considered change by the wallet

NicolasDorier:
As the test describe, segwit scriptPubKey are considered as change. It means that if you import the scriptPubKey, then send money to it, and call `listtransactions`, you will not see it. (contrary to p2sh scriptPubKey)

This PR makes it possible to use watchonly on segwit scriptPubKey.

### DIFF
--- a/src/base58.cpp
+++ b/src/base58.cpp
@@ -222,6 +222,8 @@ public:
 
     bool operator()(const CKeyID& id) const { return addr->Set(id); }
     bool operator()(const CScriptID& id) const { return addr->Set(id); }
+    bool operator()(const CWitKeyID& id) const { return false; }
+    bool operator()(const CWitScriptID& id) const { return false; }
     bool operator()(const CNoDestination& no) const { return false; }
 };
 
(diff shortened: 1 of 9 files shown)


NicolasDorier:
Not working after restart.... working on it.

NicolasDorier:
For this PR to work, I would need to save `CWitKeyID` and `CWitKScriptID` from the `mapAddressBook` into `CWalletDB`. Problem is, now `CBitcoinAddress` is used to serialized `TxDestination` into string. This mean that `CBitcoinAddress` would need to support Bech32 or completely removed into another system (a new visitor?).

I think this PR will become useful for later, so I am not closing it.

ryanofsky:
It seems like the code changes here are superceded by #11167, but maybe the new python test is useful to keep. Is this correct?

NicolasDorier:
@ryanofsky the test have been included by sipa in another PR (I don't remember when though).

NicolasDorier:
Closing this sipa already fixed it and included the test.

NicolasDorier:
Was merged on https://github.com/bitcoin/bitcoin/commit/06eaca63139ac02abf82b0245b42493c9b383905

PR #10948, opened 2017-07-28
p2p: Hardcoded seeds update pre-0.15 branch

laanwj:
Just the usual
```bash
    curl -s http://bitcoin.sipa.be/seeds.txt.gz | gzip -dc > seeds_main.txt
    python3 makeseeds.py < seeds_main.txt > nodes_main.txt
    python3 generate-seeds.py . > ../../src/chainparamsseeds.h
```


fanquake:
utACK f5b0980

achow101:
utACK f5b09807e5eb21a72602a4a24a3892ebcf1ae244

gmaxwell:
Needs an update to include 0.14 nodes.

-PATTERN_AGENT = re.compile(r"^(/Satoshi:0.12.(0|1|99)/|/Satoshi:0.13.(0|1|2|99)/)$")
+PATTERN_AGENT = re.compile(r"^(/Satoshi:0.13.(1|2|99)/|/Satoshi:0.14.(0|1|2|99)/)$")

My proposal leaves  out 0.13.0 due to no segwit support; they're under 8% of the 0.13.1+ nodes.

laanwj:
After the version bump, master will be 0.15.99 - shouldn't we include 0.15.x as well? Otherwise we're likely to forget it if we do another seeds update after the version bump for some reason.

MarcoFalke:
According to the release process, the seed update is only done before every major release. So it likely doesn't matter.

laanwj:
Ok...

gmaxwell:
@laanwj I would have been fine sticking 0.15 stuff in there, but I think it doesn't matter.

gmaxwell (approved):
utACK

MarcoFalke:
ACK f0acedd. (I ran the same commands and a subset of nodes was identical)

jonasschnelli:
Tested the ipv4 addresses (as indication if the source data is up to date):
Connect (timeout 6s), handshake and get headers with locator at height 161280.
420 of the 512 passed this test (~82%).

partial tested ACK: f0aceddf88d7d680d3b55d302b8b2935f4b1c53f

laanwj:
Thanks for testing @MarcoFalke @jonasschnelli !

> I would have been fine sticking 0.15 stuff in there, but I think it doesn't matter.

Yes, I agree it doesn't matter, I'm a little bit scared that the version filtering regex will bite us some day by accident, thanks for reminding me of it anyhow.

PR #10952, opened 2017-07-29
[wallet] Remove vchDefaultKey and have better first run detection

achow101:
Removes vchDefaultKey which was only used for first run detection. Improves wallet first run detection by checking to see if any keys were read from the database.

This also fixes a (rather contrived) case where an encrypted non-HD wallet has corruption such that the default key is no longer valid and is loaded into a Core version that supports HD wallets. This causes a runtime exception since a new hd master key is generated as the software believes the wallet file is newly created but cannot add the generated key to the wallet since it is encrypted. I was only able to replicate this error by creating a non-hd wallet, encrypting it, then editing the wallet using `db_dump` and `db_load` before loading the wallet with hd enabled. This problem has been reported by [two](https://bitcointalk.org/index.php?topic=1993244.0) [users](https://bitcointalk.org/index.php?topic=1746976.msg17511261#msg17511261) so it is something that can happen, although that raises the question of "what corrupted the default key".

~P.S. I don't know what's up with the whitespace changes. I think my text editor is doing something stupid but I don't think those are important enough to attempt undoing them.~ Undid those

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -3114,9 +3114,11 @@ DBErrors CWallet::LoadWallet(bool& fFirstRunRet)
         }
     }
 
+    // This wallet is in its first run if all of these are empty
+    fFirstRunRet = mapKeys.empty() && mapCryptedKeys.empty() && mapWatchKeys.empty() && setWatchOnly.empty() && mapScripts.empty();
+
     if (nLoadWalletRet != DB_LOAD_OK)
         return nLoadWalletRet;
-    fFirstRunRet = !vchDefaultKey.IsValid();
 
     uiInterface.LoadWallet(this);
 
@@ -3126,7 +3128,6 @@ DBErrors CWallet::LoadWallet(bool& fFirstRunRet)
 DBErrors CWallet::ZapSelectTx(std::vector<uint256>& vHashIn, std::vector<uint256>& vHashOut)
 {
     AssertLockHeld(cs_wallet); // mapWallet
-    vchDefaultKey = CPubKey();
     DBErrors nZapSelectTxRet = CWalletDB(*dbw,"cr+").ZapSelectTx(vHashIn, vHashOut);
     for (uint256 hash : vHashOut)
         mapWallet.erase(hash);
@@ -3155,7 +3156,6 @@ DBErrors CWallet::ZapSelectTx(std::vector<uint256>& vHashIn, std::vector<uint256
 
 DBErrors CWallet::ZapWalletTx(std::vector<CWalletTx>& vWtx)
 {
-    vchDefaultKey = CPubKey();
     DBErrors nZapWalletTxRet = CWalletDB(*dbw,"cr+").ZapWalletTx(vWtx);
     if (nZapWalletTxRet == DB_NEED_REWRITE)
     {
@@ -3229,14 +3229,6 @@ const std::string& CWallet::GetAccountName(const CScript& scriptPubKey) const
     // associated with the default account ("").
     const static std::string DEFAULT_ACCOUNT_NAME;
     return DEFAULT_ACCOUNT_NAME;
-}
-
-bool CWallet::SetDefaultKey(const CPubKey &vchPubKey)
-{
-    if (!CWalletDB(*dbw).WriteDefaultKey(vchPubKey))
-        return false;
-    vchDefaultKey = vchPubKey;
-    return true;
 }
 
 /**
@@ -4014,13 +4006,11 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
             if (!walletInstance->SetHDMasterKey(masterPubKey))
                 throw std::runtime_error(std::string(__func__) + ": Storing master key failed");
         }
-        CPubKey newDefaultKey;
-        if (walletInstance->GetKeyFromPool(newDefaultKey, false)) {
-            walletInstance->SetDefaultKey(newDefaultKey);
-            if (!walletInstance->SetAddressBook(walletInstance->vchDefaultKey.GetID(), "", "receive")) {
-                InitError(_("Cannot write default address") += "\n");
-                return nullptr;
-            }
+
+        // Top up the keypool
+        if (!walletInstance->TopUpKeyPool()) {
+            InitError(_("Unable to generate initial keys") += "\n");
+            return NULL;
         }
 
         walletInstance->SetBestChain(chainActive.GetLocator());
--- a/src/wallet/walletdb.cpp
+++ b/src/wallet/walletdb.cpp
@@ -130,11 +130,6 @@ bool CWalletDB::WriteOrderPosNext(int64_t nOrderPosNext)
     return WriteIC(std::string("orderposnext"), nOrderPosNext);
 }
 
-bool CWalletDB::WriteDefaultKey(const CPubKey& vchPubKey)
-{
-    return WriteIC(std::string("defaultkey"), vchPubKey);
-}
-
 bool CWalletDB::ReadPool(int64_t nPool, CKeyPool& keypool)
 {
     return batch.Read(std::make_pair(std::string("pool"), nPool), keypool);
@@ -452,7 +447,14 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
         }
         else if (strType == "defaultkey")
         {
-            ssValue >> pwallet->vchDefaultKey;
+            // We don't want or need the default key, but if there is one set,
+            // we want to make sure that it is valid so that we can detect corruption
+            CPubKey vchPubKey;
+            ssValue >> vchPubKey;
+            if (!vchPubKey.IsValid()) {
+                strErr = "Error reading wallet database: Default Key corrupt";
+                return false;
+            }
         }
         else if (strType == "pool")
         {
@@ -522,7 +524,6 @@ bool CWalletDB::IsKeyType(const std::string& strType)
 
 DBErrors CWalletDB::LoadWallet(CWallet* pwallet)
 {
-    pwallet->vchDefaultKey = CPubKey();
     CWalletScanState wss;
     bool fNoncriticalErrors = false;
     DBErrors result = DB_LOAD_OK;
@@ -565,7 +566,7 @@ DBErrors CWalletDB::LoadWallet(CWallet* pwallet)
             {
                 // losing keys is considered a catastrophic error, anything else
                 // we assume the user can live with:
-                if (IsKeyType(strType))
+                if (IsKeyType(strType) || strType == "defaultkey")
                     result = DB_CORRUPT;
                 else
                 {
--- a/src/wallet/crypter.h
+++ b/src/wallet/crypter.h
@@ -113,7 +113,6 @@ public:
 class CCryptoKeyStore : public CBasicKeyStore
 {
 private:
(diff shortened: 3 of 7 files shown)


gmaxwell:
So you still need to write a default key unless we're going to bump the wallet version again for this, because otherwise you'll make it backwards incompatible with software without this patch.

Otherwise, Concept ACK!

achow101:
How about bump the wallet version again? If there are any other wallet changes that need a version bump, we could do them all at the same time.

gmaxwell:
I would be fine with that but I think it might be easier to get this bugfix merged without it.

TheBlueMatt:
Sounds like maybe something that we should look into for 15, then, though I'd really like to look into the source of the corruption here, rather than patching over it.

achow101:
@gmaxwell I added `WriteDefaultKey` back in so that a default key will be written to maintain backwards compatibility. The key will not be read in.

gmaxwell (approved):
utACK

TheBlueMatt:
I really dont think this is the right way to fix this issue. Can we instead ensure that, if the wallet has keys but vchDefaultKey is invalid, an error is printed like "your wallet is corrupted, seems you disk is silently corrupting things", and then make sure -salvagewallet is able to properly recover from this state?

achow101:
I have updated this with a few changes.

I removed the defaultkey writing again. New wallets will no longer have a default key. If a defaultkey is found in the wallet, it must be valid, otherwise it will exit with a wallet corruption error.

It turns out that `-salvagewallet` may have been the cause of all of these issues. `-salvagewallet` apparently does not write a default key to the wallet, which with an encrypted wallet and with HD enabled, will cause the aforementioned runtime exception and crash. If a user does have a corrupted default key, then `-salvagewallet` will remove it, as it did in the past.

src/wallet/wallet.cpp
```diff
@@ -3958,11 +3947,13 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
...
         }
-        CPubKey newDefaultKey;
-        if (walletInstance->GetKeyFromPool(newDefaultKey, false)) {
-            walletInstance->SetDefaultKey(newDefaultKey);
-            if (!walletInstance->SetAddressBook(walletInstance->vchDefaultKey.GetID(), "", "receive")) {
-                InitError(_("Cannot write default address") += "\n");
+
+        // Generate an address for the address book and to fill the keypool
+        CPubKey key;
+        if (walletInstance->GetKeyFromPool(key, false)) {
```
sipa: Can you call `walletInstance->TopUpKeyPool();` here instead of this entire section? No need to waste a key and an address book entry for this.

src/wallet/walletdb.cpp
```diff
@@ -452,7 +447,13 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
         }
         else if (strType == "defaultkey")
         {
-            ssValue >> pwallet->vchDefaultKey;
+            CPubKey vchPubKey;
+            ssKey >> vchPubKey;
+            if (!vchPubKey.IsValid())
+            {
+                strErr = "Error reading wallet database: Default Key corrupt";
```
sipa: "defaultkey" is not part of `IsKeyType`, so this failure will just be a non-critical warning upon loading. I assume that's intentional?

src/wallet/walletdb.cpp
```diff
@@ -452,7 +447,13 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
         }
         else if (strType == "defaultkey")
         {
-            ssValue >> pwallet->vchDefaultKey;
+            CPubKey vchPubKey;
+            ssKey >> vchPubKey;
+            if (!vchPubKey.IsValid())
+            {
```
sipa: Code style nit: `{` on the same like as `if`.

src/wallet/wallet.cpp
achow101: Done

src/wallet/walletdb.cpp
achow101: Done

src/wallet/walletdb.cpp
achow101: It should be a critical warning and exit with an error. I have fixed this.

MarcoFalke:
AssertionError in wallet-hd:

```py
AssertionError: not(m/0'/0'/0' == m/0'/0'/1')

**Comment by achow101** [comment]
Can this be tagged for 0.15.0? It's a bug fix for a problem with `-salvagewallet` which will result in further corruption and unusability of people's wallets.

**Comment by sipa** [comment]
Tagged for 0.15, as it fixes a bug when using `-savagewallet` with encrypted wallets.

**Comment by TheBlueMatt** [comment]
Because one of the two cases of corruption in the wild was that the default key was simply deleted, can you also add a if(wallet version < x && !have_default_key)... in the load?

**Comment by achow101** [comment]
@TheBlueMatt What for?

**Comment by TheBlueMatt** [comment]
@achow101 because its a trivial check and its a type of corruption we've seen in wallets on the wild...it costs nothing to add a check for it.

**Comment by sipa** [comment]
@TheBlueMatt The current code requires that if a default key is present, it must be valid. Do you think an extra check is needed to verify that it hasn't been accidentally deleted?

**Comment by TheBlueMatt** [comment]
Considering that we've seen it in the wild and its clear indication that
there is something wrong with your hardware/wallet storage, I'd say it
would be prudent.

**Comment by achow101** [comment]
@TheBlueMatt I meant what do you want it to check for and error on? That below a certain version number you must have a valid default key? If that is the case, that isn't possible since this PR does not change the version number, thus all wallets will have to have a valid default key.

Furthermore, I believe the corruption that we are seeing in the wild is a result of `-savagewallet` not passing the default key through to the new wallet file, not because of hardware corruption or the like.

**Comment by TheBlueMatt** [comment]
Ahh, OK, I missed the salvagewallet-possible-culprit. PR looks good
as-is to me.

**Comment by laanwj** [comment]
Nit: please also remove the `"defaultkey"` parsing in `walletdb.cpp` `ReadKeyValue`

**Comment by achow101** [comment]
@laanwj That was left in due address @TheBlueMatt's concerns about warning users about possible corruption. It was modified so that the default key must either exist and be valid or not exist at all. The reason for this is that if it exists and is not valid, then corruption has occurred and the user will be warned accordingly.

**Comment by laanwj** [comment]
That's true, could have used a comment, it now looks a bit silly just throwing away the result. But I agree.

**Comment by laanwj** [comment]
I tested this with an old wallet and got "Error: Error loading wallet.dat: Wallet corrupted", will check why.

File: src/wallet/walletdb.cpp [hunk 4]
```diff
@@ -452,7 +447,12 @@ ReadKeyValue(CWallet* pwallet, CDataStream& ssKey, CDataStream& ssValue,
         }
         else if (strType == "defaultkey")
         {
-            ssValue >> pwallet->vchDefaultKey;
+            CPubKey vchPubKey;
+            ssKey >> vchPubKey;
```
laanwj: This should be `ssValue >> vchPubKey`. Otherwise this will fail for every wallet with a default key record!

jnewbery:
Intuitively for me, the logic for determining whether this is a first run should remain in the `CWallet::LoadWallet()` function. It makes sense for me that `CWalletDB` remains as a modifier object for the wallet, and simply reads/writes data to the wallet database. Is it possible to leave the first_run logic in `CWallet::LoadWallet()` (by checking that `mapKeys` is empty or similar)?

I'm not all that familiar with wallet code, so I may be wrong about this. I'll defer to those who have a better understanding of the wallet model.

I also don't think this necessarily needs to go into v0.15 if we don't have time to fully review it. The problems with `salvagewallet` are long-standing and not made any worse in v0.15.

src/wallet/walletdb.cpp
achow101: Oops. Fixed.

achow101:
>  Is it possible to leave the first_run logic in CWallet::LoadWallet() (by checking that mapKeys is empty or similar)?

That would probably be better to do.

> I also don't think this necessarily needs to go into v0.15 if we don't have time to fully review it. The problems with `salvagewallet` are long-standing and not made any worse in v0.15.

0.13 broke `salvagewallet` with HD wallets. Anyone who uses `salvagewallet` on an encrypted wallet is currently going to run into problems just starting their wallet in the first place.

jnewbery:
> 0.13 broke salvagewallet with HD wallets.

To be clear, I'm not arguing against this PR - it's obviously good and we should do it. My point is that we don't need to rush it for v0.15. The number of users who haven't yet upgraded to v0.13, but will upgrade to v0.15 before we release v0.15.1 is going to be ~ zero. Not including the fix in v0.15 doesn't make the problem any worse.

laanwj:
> Intuitively for me, the logic for determining whether this is a first run should remain in the CWallet::LoadWallet() function.

I thought about this too and came to the opposite conclusion: wouldn't the database layer be the foremost authority on whether it just created a new database file, or opened an existing one? I think this PR goes the right direction.

jnewbery:
I guess it depends on what we mean by 'first run'. If it's 'the first time we create a file', then yes I agree that the database layer is the authority. If it's 'a wallet without any of these objects', then I would have thought the right place would be at the wallet layer.

But that's just my intuition. You know this code better than I do, so if you think this is the right direction, I won't object.

laanwj:
> I guess it depends on what we mean by 'first run'. If it's 'the first time we create a file', then yes I agree that the database layer is the authority. If it's 'a wallet without any of these objects', then I would have thought the right place would be at the wallet layer.

CWalletDB is not just the low-level database though - it also manages the format. After all it iterates over application-specific key/value pairs in `CWalletDB::LoadWallet`. So any "application format" checking belongs there too, IMO. Counting "a wallet without any of these objects" belongs naturally in the place where all the objects are iterated over.

`wallet/db.cpp` has the low-level, application-independent database handling. If you'd argue that that is an abstraction level too low to put this, I'd agree.

I also can't help that all these classes and objects are named confusingly.

laanwj:
>To be clear, I'm not arguing against this PR - it's obviously good and we should do it. My point is that we don't need to rush it for v0.15. 

I'd prefer to drop it for 0.15.0 too, to be honest.

achow101:
I changed the detection back to `CWallet::LoadWallet()` since I think it is more logical to be there. That was my original plan, I just couldn't figure out how since I forgot that it has access to `mapKeys`, etc.

jnewbery:
Looks good. Tested ACK b8aa9729a5dbeb883857b14918aa6c6d2b71eea4 (although needs rebase)

jonasschnelli:
Needs rebase.

achow101:
rebased

jnewbery:
utACK 1a7bc5aac60a37b6f9f0632546467d716a3fd5e4

sipa:
There seems to be a problem in `keypool_topup.py`.

achow101:
Fixed that I think. It's an off-by-one since default key is no longer created and consuming a key from the keypool.

laanwj:
> It's an off-by-one since default key is no longer created and consuming a key from the keypool.

Finally :-)

jonasschnelli:
Great to see the default key vanish...
utACK e53615b443894fb96d5eb885b3812776c1b1033b

laanwj:
Removing 0.15.1 milestone, as it can create problems at the moment when opening a new wallet with older versions.

PR #10705, opened 2017-06-29
Trivial: spelling fixes

ka7:
patch contains some spelling fixes ( just in comments ) as found by a bot ( http://www.misfix.org, https://github.com/ka7/misspell_fixer ).

PR #10956, opened 2017-07-30
Fix typos

practicalswift:
Fix some typos not covered by #10705.

### DIFF
--- a/contrib/linearize/README.md
+++ b/contrib/linearize/README.md
@@ -46,7 +46,7 @@ linearize-hashes.py.
 (Default: `1000*1000*1000 bytes`)
 * `netmagic`: Network magic number.
 * `out_of_order_cache_sz`: If out-of-order blocks are being read, the block can
-be written to a cache so that the blockchain doesn't have to be seeked again.
+be written to a cache so that the blockchain doesn't have to be sought again.
 This option specifies the cache size. (Default: `100*1000*1000 bytes`)
 * `rev_hash_bytes`: If true, the block hash list written by linearize-hashes.py
 will be byte-reversed when read by linearize-data.py. See the linearize-hashes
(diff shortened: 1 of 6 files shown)


jonasschnelli:
Trivial no-code-change ACK c8cd2bf05c81508e173bf1e9f7a7f234b9d0e695

promag (commented):
Search and fix:
 - targetting;
 - parametres
 - discardability (?)

practicalswift:
@promag Thanks! Added the "parameters" fix.

Regarding "targeting" - part of `leveldb` and should be reported upstreams.

Regarding "discardability" - I think that one is OK.

promag:
utACK e061983.

contrib/linearize/README.md
```diff
@@ -46,7 +46,7 @@ linearize-hashes.py.
 (Default: `1000*1000*1000 bytes`)
 * `netmagic`: Network magic number.
 * `out_of_order_cache_sz`: If out-of-order blocks are being read, the block can
-be written to a cache so that the blockchain doesn't have to be seeked again.
```
eklitzke: I would keep this as-is, since "seek" has a technical meaning (referring to the system call of the same name). It's not necessarily grammatical English, but I'd consider it grammatical programmer-English.

practicalswift: Sure, but is it referring to the system call in this case?

practicalswift:
Added corrections for three misspellings of the word "transaction" :-)

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/10956/commits/9d5e98ff80479628492329b3b6bdce56d255baa0

fanquake:
utACK 9d5e98f

contrib/linearize/README.md
jnewbery: how about "...so that we don't need to seek through the blockchain again"?

jnewbery:
utACK with one comment.

PR #10961, opened 2017-07-31
Improve readability of DecodeBase58Check(...)

practicalswift:
Use the more readable form ...

```c++
&vchRet[vchRet.size() - 4]
```

... instead of ...

```c++
&v.end()[-n]
```

Has the added benefit of eliminating a spurious static analyzer warning about improper use of negative values.

### DIFF
--- a/src/base58.cpp
+++ b/src/base58.cpp
@@ -136,7 +136,7 @@ bool DecodeBase58Check(const char* psz, std::vector<unsigned char>& vchRet)
     }
     // re-calculate the checksum, ensure it matches the included 4-byte checksum
     uint256 hash = Hash(vchRet.begin(), vchRet.end() - 4);
-    if (memcmp(&hash, &vchRet.end()[-4], 4) != 0) {
+    if (memcmp(&hash, &vchRet[vchRet.size() - 4], 4) != 0) {
         vchRet.clear();
         return false;
     }


promag:
utACK 6ab2ce8.

jonasschnelli:
This seems to be an over-cosmetic change... not sure if its worth.
utACK 6ab2ce83d9268fc81ac3bec41365ce2d2adfd349

src/base58.cpp
```diff
@@ -136,7 +136,7 @@ bool DecodeBase58Check(const char* psz, std::vector<unsigned char>& vchRet)
     }
     // re-calculate the checksum, ensure it matches the included 4-byte checksum
     uint256 hash = Hash(vchRet.begin(), vchRet.end() - 4);
-    if (memcmp(&hash, &vchRet.end()[-4], 4) != 0) {
+    if (memcmp(&hash, vchRet.data() + vchRet.size() - 4, 4) != 0) {
```
laanwj: I guess there's some reason `vchRet.end() - 4` doesn't work (pointer versus iterator conflict or something like that)?

MeshCollider: Yeah `vchRet.end()-4` would still be an iterator I believe, so you'd have to do some ugly thing like `&*(vchRet.end()-4)` to make it into a pointer iirc?

sipa: How about `vchRet[vchRet.size() - 4]` ?

MeshCollider: Believe that would work, and yeah that's a lot clearer imo 👍

practicalswift:
@laanwj @MeshCollider @sipa Thanks for reviewing! Changed to `&vchRet[vchRet.size() - 4]`. Looks good? :-)

MeshCollider:
Just confirmed this should be identical behavior, ACK https://github.com/bitcoin/bitcoin/pull/10961/commits/c6a995e7e5e6ae37dc1684f6060038e96864e947

practicalswift:
Travis have completed without errors, but GitHub still reports "1 pending check".

practicalswift:
Seems like Travis CI is stuck. Anyone able to re-start Travis? :-)

ryanofsky (commented):
utACK c6a995e7e5e6ae37dc1684f6060038e96864e947

sipa:
utACK c6a995e7e5e6ae37dc1684f6060038e96864e947

PR #10964, opened 2017-07-31
Pass SendCoinsRecipient (208 bytes) by reference

practicalswift:
Pass `SendCoinsRecipient` (208 bytes) by reference.

Avoid passing big parameters by value.

### DIFF
--- a/src/qt/paymentserver.cpp
+++ b/src/qt/paymentserver.cpp
@@ -620,7 +620,7 @@ void PaymentServer::fetchRequest(const QUrl& url)
     netManager->get(netRequest);
 }
 
-void PaymentServer::fetchPaymentACK(CWallet* wallet, SendCoinsRecipient recipient, QByteArray transaction)
+void PaymentServer::fetchPaymentACK(CWallet* wallet, const SendCoinsRecipient& recipient, QByteArray transaction)
 {
     const payments::PaymentDetails& details = recipient.paymentRequest.getDetails();
     if (!details.has_payment_url())
--- a/src/qt/paymentserver.h
+++ b/src/qt/paymentserver.h
@@ -113,7 +113,7 @@ public Q_SLOTS:
     void uiReady();
 
     // Submit Payment message to a merchant, get back PaymentACK:
-    void fetchPaymentACK(CWallet* wallet, SendCoinsRecipient recipient, QByteArray transaction);
+    void fetchPaymentACK(CWallet* wallet, const SendCoinsRecipient& recipient, QByteArray transaction);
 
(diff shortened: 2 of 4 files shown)


promag (commented):
These are unrelated structures. How did you came to this change? I mean, aren't there more cases like these?

Anyway, IMO split in 2 commits, one for each structure and remove the sizeof() from PR title and commit messages.

practicalswift:
@promag No, these two are the only two structures over 128 bytes which are passed by value. Let me know if you find any counterexample that falsifies that claim :-)

benma:
utACK 346d8b096e1a93d673ca6e8090ed3e6f53dd4be7

I'd be happy if you could document how you identified the refactorings you are doing (with the exact command used in case of static checkers) in the commit message.

Do you have a document somewhere that details all of the checks you are doing? Would be handy to have them in one place.

src/net.cpp
```diff
@@ -2254,7 +2254,7 @@ bool CConnman::InitBinds(const std::vector<CService>& binds, const std::vector<C
     return fBound;
 }
 
-bool CConnman::Start(CScheduler& scheduler, Options connOptions)
+bool CConnman::Start(CScheduler& scheduler, const Options& connOptions)
```
theuni: Please don't pass this by reference. This was intentional in case any unique_ptr's end up being passed in. Please just std::move it in init instead.

practicalswift: Thanks for the clarification! Removing!

practicalswift:
Build error unrelated. May I request a re-build? :-)

src/net.cpp
benma: @theuni can you elaborate? Why worry about it when there is no unique_ptr inside now? Even if there was, passing the struct by const reference prevents Connman from taking ownership.

theuni: @benma Setting the unique_ptr thing aside, generally speaking, I prefer to pass anything with allocated mem by value rather than const reference if I know that it's going to be stored by the caller.

For example, we know that Start() is going to keep a local copies of vSeedNodes/vWhiteBinds/etc. If you pass by const reference, you guarantee the copies. But if you pass by value, you give the caller the opportunity to std::move them in instead.

That said, #10977 changes this to a const reference also. I don't care too strongly, so I'll just shut up about it :)

practicalswift: @theuni Didn't realize that I changed the same thing in two PRs - sorry :-)

jonasschnelli:
trivial utACK d3d946a294988ca3ab2b2c8848da5b393fbbcf42

PR #10968, opened 2017-08-01
Add instructions for parallel gitian builds.

coblee:
This makes builds much faster if you have a multi-core machine.


laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/10968/commits/36928b069561501c9d239830ed59d27dd9814325

(that's awful, it only picks a single-core VM by default?)

coblee:
Yes by default, it's only a single-core VM. And gbuild defaults to "-j 2".
Now builds are 4 times faster on my machine.

doc/release-process.md
```diff
@@ -111,16 +111,16 @@ The gbuild invocations below <b>DO NOT DO THIS</b> by default.
 ### Build and sign Bitcoin Core for Linux, Windows, and OS X:
 
     pushd ./gitian-builder
-    ./bin/gbuild --memory 3000 --commit bitcoin=v${VERSION} ../bitcoin/contrib/gitian-descriptors/gitian-linux.yml
+    ./bin/gbuild --num-make `nproc` --memory 3000 --commit bitcoin=v${VERSION} ../bitcoin/contrib/gitian-descriptors/gitian-linux.yml
```
laanwj: This is a bit dangerous - in many cases, number of concurrent C++ compiles is limited by memory, not number of cores

PRabahy: https://github.com/bitcoin/bitcoin/blob/master/contrib/gitian-build.sh appears to use -j instead of --num-make. Is there a difference?

coblee: -j and --num-make are the same thing.

@laanwj what's a better way then? I do know that builds are much faster on my quad-core mbp when I run it with "--num-make 4"

laanwj: You could just provide the default of `--num-make 2` here, which works with `--memory 3000`
Doubling one of the numbers means you have to double the other one too, at least for KVM, for LXC the memory limit is  ignored which is why you didn't notice anything.

coblee:
@laanwj I made the fix you recommended. Change `num-make` to 2

jonasschnelli:
utACK 393e77637c65f73c65de8462454d2582a5e359a7

fanquake:
utACK 393e776

MarcoFalke:
utACK 393e776, but I am not entirely happy about the binary files in our source code repo. At least it seems pngcrush can reduce the size by 30k... Would you mind fixing that up? (c.f. 7a482e4ac)

fanquake:
ping @coblee Could you make Marcos change so we can merge this?

coblee:
Done and pngcrush'd 2 other images.

MarcoFalke:
Looks like my compression was slightly more efficient. (I used `contrib/devtools/optimize-pngs.py`)

Also, rewriting existing files in the git repo will keep the old versions around. Probably not worth trying to reduce the size...

I reverted the change to the existing files and force pushed the commit to the litecoin repo. I hope this is ok.

coblee:
Sure, np

fanquake:
ACK e93ff71

PR #10975, opened 2017-08-02
[script] Return early if no valid opcodes found in CountWitnessSigOps(...)

practicalswift:
Return early if no valid opcodes found in `CountWitnessSigOps(...)`.

Prior to this commit this was the only place in the codebase where `GetOp(...)` was called without checking the return value.

### DIFF
--- a/src/script/interpreter.cpp
+++ b/src/script/interpreter.cpp
@@ -1557,9 +1557,15 @@ size_t CountWitnessSigOps(const CScript& scriptSig, const CScript& scriptPubKey,
     if (scriptPubKey.IsPayToScriptHash() && scriptSig.IsPushOnly()) {
         CScript::const_iterator pc = scriptSig.begin();
         std::vector<unsigned char> data;
+        bool hasValidOp = false;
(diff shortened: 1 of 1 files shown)


src/script/interpreter.cpp
```diff
@@ -1557,9 +1557,15 @@ size_t CountWitnessSigOps(const CScript& scriptSig, const CScript& scriptPubKey,
...
+        bool hasValidOp = false;
         while (pc < scriptSig.end()) {
             opcodetype opcode;
-            scriptSig.GetOp(pc, opcode, data);
+            if (scriptSig.GetOp(pc, opcode, data)) {
+                hasValidOp = true;
+            }
+        }
+        if (!hasValidOp) {
+            return 0;
```
promag: This is not an obvious early return, because one has to go deeper in `IsWitnessProgram`.

promag (commented):
Concept ACK.

promag:
Nit, reword to use your usual prefix `script:`.

PR #10973, opened 2017-08-01
Refactor: separate wallet from node

ryanofsky:
This PR is the last in a chain of PRs (#14437, #14711, and #15288) that make the wallet code access node state through an abstract [`Chain`](https://github.com/ryanofsky/bitcoin/blob/pr/wipc-sep/src/interfaces/chain.h) class in [`src/interfaces/`](https://github.com/ryanofsky/bitcoin/tree/pr/wipc-sep/src/interfaces) instead of using global variables like `cs_main`, `chainActive`, and `g_connman`. After this PR, wallet code no longer accesses global variables declared outside the wallet directory, and no longer calls functions accessing those globals (as verified by the `hide-globals` script in #10244).

This PR and the previous PRs have been refactoring changes that do not affect behavior. Previous PRs have consisted of lots of mechanical changes like:

```diff
-    wtx.nTimeReceived = GetAdjustedTime();
+    wtx.nTimeReceived = m_chain->getAdjustedTime();
```

This PR is smaller, but less mechanical. It replaces last few bits of wallet code that access node state directly (through `CValidationInterface`, `CRPCTable`, and `CCoinsViewMemPool` interfaces) with code that uses the `Chain` interface.

These changes allow followup PR #10102 (multiprocess gui & wallet PR) to work without any significant updates to wallet code. Additionally they:

* Provide a single place to describe the interface between wallet and node code.
* Can make better wallet testing possible, because the `Chain` object consists of virtual methods that can be overloaded for mocking. (This could be used to test edge cases in the rescan code, for example).

PR #10976, opened 2017-08-02
[MOVEONLY] Move some static functions out of wallet.h/cpp

ryanofsky:
This just moves some static wallet fee and init functions out of `wallet/wallet.cpp` and into new `wallet/fees.cpp` and `wallet/init.cpp` source files. There is one commit updating declarations and callers, followed by two MOVEONLY commits actually moving the function bodies.

This change is desirable because wallet.h/cpp are monolithic and hard to navigate, so pulling things out and grouping together pieces of related functionality should improve the organization.

Another motivation is the wallet process separation work in https://github.com/bitcoin/bitcoin/pull/10973, where (at least initially) parameter parsing and fee estimation are still done in the main process rather than the wallet process, and having functions that run in different processes scrambled up throughout wallet.cpp is unnecessarily confusing.

### DIFF
--- a/src/wallet/init.h
+++ b/src/wallet/init.h
@@ -1,0 +1,25 @@
+// Copyright (c) 2009-2010 Satoshi Nakamoto
+// Copyright (c) 2009-2017 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#ifndef BITCOIN_WALLET_INIT_H
+#define BITCOIN_WALLET_INIT_H
+
+#include <string>
+
+//! Return the wallets help message.
+std::string GetWalletHelpString(bool showDebug);
+
+//! Wallets parameter interaction
+bool WalletParameterInteraction();
+
+//! Responsible for reading and validating the -wallet arguments and verifying the wallet database.
+//  This function will perform salvage on the wallet if requested, as long as only one wallet is
+//  being loaded (CWallet::ParameterInteraction forbids -salvagewallet, -zapwallettxes or -upgradewallet with multiwallet).
+bool WalletVerify();
+
+//! Load wallet databases.
+bool InitLoadWallet();
+
+#endif // BITCOIN_WALLET_INIT_H
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -8,7 +8,6 @@
 #include "chain.h"
 #include "consensus/validation.h"
 #include "core_io.h"
-#include "init.h"
 #include "httpserver.h"
 #include "validation.h"
 #include "net.h"
@@ -26,6 +25,8 @@
 #include "wallet/feebumper.h"
 #include "wallet/wallet.h"
 #include "wallet/walletdb.h"
+
+#include <init.h>  // For StartShutdown
 
 #include <stdint.h>
 
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -162,6 +162,8 @@ BITCOIN_CORE_H = \
   wallet/crypter.h \
   wallet/db.h \
   wallet/feebumper.h \
+  wallet/fees.h \
+  wallet/init.h \
   wallet/rpcwallet.h \
   wallet/wallet.h \
   wallet/walletdb.h \
@@ -239,6 +241,8 @@ libbitcoin_wallet_a_SOURCES = \
   wallet/crypter.cpp \
   wallet/db.cpp \
   wallet/feebumper.cpp \
+  wallet/fees.cpp \
+  wallet/init.cpp \
   wallet/rpcdump.cpp \
   wallet/rpcwallet.cpp \
   wallet/wallet.cpp \
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -43,6 +43,7 @@
 #include "utilmoneystr.h"
 #include "validationinterface.h"
 #ifdef ENABLE_WALLET
+#include "wallet/init.h"
 #include "wallet/wallet.h"
 #endif
 #include "warnings.h"
@@ -420,7 +421,7 @@ std::string HelpMessage(HelpMessageMode mode)
     strUsage += HelpMessageOpt("-maxuploadtarget=<n>", strprintf(_("Tries to keep outbound traffic under the given target (in MiB per 24h), 0 = no limit (default: %d)"), DEFAULT_MAX_UPLOAD_TARGET));
 
 #ifdef ENABLE_WALLET
-    strUsage += CWallet::GetWalletHelpString(showDebug);
+    strUsage += GetWalletHelpString(showDebug);
 #endif
 
 #if ENABLE_ZMQ
@@ -1035,7 +1036,7 @@ bool AppInitParameterInteraction()
         if (!ParseMoney(gArgs.GetArg("-minrelaytxfee", ""), n)) {
             return InitError(AmountErrMsg("minrelaytxfee", gArgs.GetArg("-minrelaytxfee", "")));
         }
-        // High fee check is done afterward in CWallet::ParameterInteraction()
+        // High fee check is done afterward in WalletParameterInteraction()
         ::minRelayTxFee = CFeeRate(n);
     } else if (incrementalRelayFee > ::minRelayTxFee) {
         // Allow only setting incrementalRelayFee to control both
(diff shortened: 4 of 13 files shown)


ryanofsky:
@laanwj, if you think this change is a good idea, it might be worth merging before 0.15 is branched, since it moves some functions around. I don't think it conflicts with other tagged prs, but if it does I'd be happy to rebase this on top of them.

jnewbery:
Seems to be largely the same as #10767 ?

ryanofsky:
> Seems to be largely the same as #10767 ?

This covers fee code as well as init code, and I think the cleanups in #10767 are good but also that it's nice to start with a PR that's 100% MOVEONLY (or as close as possible). This way both the cleanup changes and moveonly changes are simpler to evaluate, and the MOVEONLY changes have (hopefully) a chance of getting merged before the branch.

src/wallet/rpcwallet.cpp
```diff
@@ -27,6 +26,8 @@
 #include "wallet/wallet.h"
 #include "wallet/walletdb.h"
 
+#include <init.h>  // For StartShutdown
```
jnewbery: why?

jnewbery (commented):
ACK 4298d90a9e3ea04e74d85ee995bf58e6aa8a0587.

I'd prefer to name the new files `walletinit.cpp` and `walletinit.h` just to avoid any ambiguity with `src/init.cpp`

jonasschnelli:
Concept ACK

promag (commented):
The only static function staying is `CreateWalletFromFile()` which is a factory for `CWallet` and requires access to private members, so makes sense to keep it there.

Agree with @jnewbery regarding file names, but to be consistent to `rpcwallet.*`, `initwallet.*` sounds better.

ACK 4298d90.

jnewbery:
@laanwj - as @ryanofsky says, this is another move/refactor PR that would be good to get in before branching v0.15, to help with backporting. But maybe there are already too many of those competing for your attention!

(I'd still like the files to be renamed `initwallet.{cpp,h}` before this gets merged)

MarcoFalke:
We can always backport the commits right after release of 0.15.0 to the 0.15 branch. No need to rush, imo.

src/wallet/init.h
```diff
@@ -0,0 +1,25 @@
...
+
+//! Wallets parameter interaction
+bool WalletParameterInteraction();
+
+//! Responsible for reading and validating the -wallet arguments and verifying the wallet database.
+//  This function will perform salvage on the wallet if requested, as long as only one wallet is
+//  being loaded (CWallet::ParameterInteraction forbids -salvagewallet, -zapwallettxes or -upgradewallet with multiwallet).
+bool WalletVerify();
+
+//! Load wallet databases
```
ryanofsky: Added in aee0e0f6857a18d0e93825a2bb475c55446cfa85

ryanofsky (commented):
Rebased 4298d90a9e3ea04e74d85ee995bf58e6aa8a0587 -> 5461f69fe9e503be734c33d7b57035214bafba53 ([pr/wmove.2](https://github.com/ryanofsky/bitcoin/commits/pr/wmove.2) -> [pr/wmove.3](https://github.com/ryanofsky/bitcoin/commits/pr/wmove.3)) because of conflicts with gArgs PR.
Added 1 commit 5461f69fe9e503be734c33d7b57035214bafba53 -> aee0e0f6857a18d0e93825a2bb475c55446cfa85 ([pr/wmove.3](https://github.com/ryanofsky/bitcoin/commits/pr/wmove.3) -> [pr/wmove.4](https://github.com/ryanofsky/bitcoin/commits/pr/wmove.4), [compare](https://github.com/ryanofsky/bitcoin/compare/pr/wmove.3...pr/wmove.4))
Squashed aee0e0f6857a18d0e93825a2bb475c55446cfa85 -> f01103c1e0a204fc7f40a06755f6c3adb5480cf8 ([pr/wmove.4](https://github.com/ryanofsky/bitcoin/commits/pr/wmove.4) -> [pr/wmove.5](https://github.com/ryanofsky/bitcoin/commits/pr/wmove.5))


> I'd prefer to name the new files walletinit.cpp and walletinit.h just to avoid any ambiguity with src/init.cpp

> to be consistent to rpcwallet.\*, initwallet.\* sounds better.

I don't think this is a good rationale for renaming. Including relative to current directory is generally bad practice in c and c++, because it makes code harder to understand, and can lead to accidental breakages or strange requirements like filenames throughout the directory tree having to be globally unique. In bitcoin, we mostly avoid including relative to current directory (hence `#include "policy/fees.h"` and `#include "wallet/feebumper.h"` instead of `#include "fees.h"` and `#include "feebumper.h"`). In other large codebases, including relative to current directory is [forbidden outright](https://google.github.io/styleguide/cppguide.html#Names_and_Order_of_Includes).

In the `src/wallet` directory, there are currently 7 header files. 5 of them have no redundant `wallet` prefix or suffix, 1 of them has a redundant `wallet` prefix, and 1 has a redundant `wallet` suffix. In other bitcoin source directories like `src/policy`, `src/primitives`, and `src/rpc`, there are no redundant prefixes or suffixes at all. I would prefer also not to add redundant prefixes or suffixes here.

jnewbery:
> it makes code harder to understand

Can you explain this?

In fact, there are only a couple of places where non-test files are not named uniquely within the directory tree:

```
→ find -name *cpp -printf '%f\n' | sort | uniq -c | sort
...
      1 zmqpublishnotifier.cpp
      2 base58.cpp
      2 crypto_tests.cpp
      2 lockedpool.cpp
      2 net.cpp
      2 protocol.cpp
→ find -name base58.cpp
./src/base58.cpp
./src/bench/base58.cpp
→ find -name crypto_tests.cpp
./src/wallet/test/crypto_tests.cpp
./src/test/crypto_tests.cpp
→ find -name lockedpool.cpp
./src/support/lockedpool.cpp
./src/bench/lockedpool.cpp
→ find -name net.cpp
./src/net.cpp
./src/rpc/net.cpp
→ find -name protocol.cpp
./src/rpc/protocol.cpp
./src/protocol.cpp
```

I prefer `walletinit.cpp`, but I'll defer to yours and others' C++ expertise on this.

ryanofsky:
> > it makes code harder to understand

> Can you explain this?

`#include "policy/fees.h"` is easier to understand than `#include "fees.h"` because it provides more context. 

> In fact, there are only a couple of places where non-test files are not named uniquely within the directory tree:

I've never worked on any project that required source filenames to be globally unique, so I'm not surprised that you found source filenames are not globally unique in bitcoin.

laanwj:
> I don't think this is a good rationale for renaming. Including relative to current directory is generally bad practice in c and c++, because it makes code harder to understand,

Unfortunately, relative inclusion is what we do in bitcoin by using `#include ""` instead of `#include <>` **everywhere**. Heck, there have been well-meaning PRs that change to `#include ""` for "consistency" (see #10752). 

I agree it's generally a bad idea. It can lead to confusion, as well as makes the compiler do a lot of 'probing' where files are (so slows down compilation). But not something we're going to get rid of from one moment to another.

So: right now if you were to include "init.h" in "wallet.h", it gets "wallet/init.h" instead of the main one (which is what happens in rpcwallet). Yes, this is confusing. I'd also prefer unique naming of header files for now, for that reason. Another vote to rename it to `walletinit.h`.

promag:
I also prefer `<>` includes with full project path instead of prefixes or whatever.

laanwj:
> I also prefer <> includes with full project path instead of prefixes or whatever.

The funniest thing is that that's how we're using `#include ""`. I would guess a script-diff that replaces `"..."` in `#include` lines with `<...>` would build with only few changes (maybe some exceptions such as including generated test data). We hardly, if ever use "" for actual relative include.

ryanofsky:
Could this be merged? It has 2 code review acks and a concept ack and is mostly moveonly.

#11053 in it's current form or one of the variations proposed in https://github.com/bitcoin/bitcoin/pull/11053#issuecomment-323069831 should moot whatever concerns there might be about includes going forward.

jnewbery:
I withdraw my objection to the filename `init.cpp`. utACK

src/wallet/init.h
```diff
@@ -0,0 +1,25 @@
...
+
+//! Return the wallets help message.
+std::string GetWalletHelpString(bool showDebug);
+
+//! Wallets parameter interaction
+bool WalletParameterInteraction();
+
+//! Responsible for reading and validating the -wallet arguments and verifying the wallet database.
+//  This function will perform salvage on the wallet if requested, as long as only one wallet is
+//  being loaded (CWallet::ParameterInteraction forbids -salvagewallet, -zapwallettxes or -upgradewallet with multiwallet).
```
MarcoFalke: nit: doc string needs fixup

MarcoFalke:
Post merge utACK f01103c1e0a204fc7f40a06755f6c3adb5480cf8. Might want to address the nit sometime

PR #10980, opened 2017-08-03
[Refactor] Decouple CKeyStore from CWatchOnlyStore

NicolasDorier:
Breaking `CKeyStore` into two parts, one being the `CKeyStore`, the other `CWatchOnlyStore`.

The future `externalhd` would then just have to implement a new CKeyStore.

See https://github.com/bitcoin/bitcoin/pull/9728#issuecomment-322953243

### DIFF
--- a/src/keystore.cpp
+++ b/src/keystore.cpp
@@ -17,16 +17,21 @@ bool CBasicKeyStore::GetPubKey(const CKeyID &address, CPubKey &vchPubKeyOut) con
 {
     CKey key;
     if (!GetKey(address, key)) {
-        LOCK(cs_KeyStore);
-        WatchKeyMap::const_iterator it = mapWatchKeys.find(address);
-        if (it != mapWatchKeys.end()) {
-            vchPubKeyOut = it->second;
-            return true;
-        }
         return false;
     }
     vchPubKeyOut = key.GetPubKey();
     return true;
+}
+
+bool CWatchOnlyStore::GetPubKey(const CKeyID &address, CPubKey &vchPubKeyOut) const
+{
+    LOCK(cs_KeyStore);
+    WatchKeyMap::const_iterator it = mapWatchKeys.find(address);
+    if (it != mapWatchKeys.end()) {
+        vchPubKeyOut = it->second;
+        return true;
+    }
+    return false;
 }
 
 bool CBasicKeyStore::AddKeyPubKey(const CKey& key, const CPubKey &pubkey)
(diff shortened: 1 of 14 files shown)


NicolasDorier:
This is the first step to achieve what is described on https://github.com/bitcoin/bitcoin/pull/9728#issuecomment-322953243

instagibbs:
Will review.

instagibbs (commented):
concept ACK, needs rebase

jtimon:
Concept ACK, needs rebase

sipa:
Concept NACK, I think this is a step in the wrong direction. The concept of watching should be removed from the keystore entirely.

NicolasDorier:
@sipa what do you mean? this is precisely the intent of this PR: Removing WatchOnlyStore from CKeyStore. (see the CWatchOnlyStore class)

sipa:
@NicolasDorier There's no need for a separate store for that, watching should be determined by the wallet, not the keys or anything related to signing.

I'm working on a design doc for how to evolve the watching/signing code and segwit/HD/... support. I'll let you know when it's in a usable state.

ryanofsky:
> I'm working on a design doc for how to evolve the watching/signing code and segwit/HD/... support. I'll let you know when it's in a usable state.

Doc was posted as https://gist.github.com/sipa/125cfa1615946d0c3f3eec2ad7f250a2#future-design

It does make this PR look like a step in a different direction if the goal is to have unified representation and handling of watchonly and signing keys.

It also isn't obviously to me how this PR would simplify https://github.com/instagibbs/bitcoin/commits/externalhd2 or #9728. @NicolasDorier it would be good if you could explain more why this change is useful, and whether you think anything in sipa's design might make it more difficult to support external hd wallets.

NicolasDorier:
@ryanofsky this PR was an attempt to fix my immediate problem on externalhd2. I was trying to fix the mess incrementally, one PR after another.

The design doc has a different approach: Complete redesign + one time migration. I agree with the general direction it takes (I need to re read it later though), it just seems to me it will be quite hard to do incrementally.

NicolasDorier:
@ryanofsky  Just additional information about what I was thinking. Currently `externalhd2` works fine but generate the address as "watch-only". This is long time ago so I might not remember all details.

The problem is that core consider now "IS_MINE" only if you can sign. But nevertheless, the watchonly addresses generated by externalhd2 should be considered IS_MINE even if I can't sign. (Because I want Core to behave exactly as if it was a normal wallet, except without it the ability to sign with the generated keys)

My solution was to create a new implementation of CKeyStore for external hd keys. 
The old CKeyStore would still be used in non-external hd mode. 
And I would keep the WatchOnlyStore separately from those classes. (As the WatchOnlyStore still need to exist independently from the CKeyStore implementation use)

Because generated addresses would come from the CKeyStore, it would be considered "IS_MINE".

NicolasDorier:
closing this, my approach of slowly refactoring does not seem to get traction.

PR #10988, opened 2017-08-04
qt: Increase BLOCK_CHAIN_SIZE constants

laanwj:
- Increase `BLOCK_CHAIN_SIZE` from 120GB to 150GB
- Increase `CHAIN_STATE_SIZE` from 2GB to 4GB

I took the local sizes of the blocks and chainstate directory, and added a bit extra to accomodate the near future (15GB for the chain and 1GB for the chainstate).

### DIFF
--- a/src/qt/intro.cpp
+++ b/src/qt/intro.cpp
@@ -22,9 +22,9 @@
 
 static const uint64_t GB_BYTES = 1000000000LL;
 /* Minimum free space (in GB) needed for data directory */
-static const uint64_t BLOCK_CHAIN_SIZE = 120;
+static const uint64_t BLOCK_CHAIN_SIZE = 150;
 /* Minimum free space (in GB) needed for data directory when pruned; Does not include prune target */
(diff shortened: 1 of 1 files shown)


src/qt/intro.cpp
```diff
@@ -22,9 +22,9 @@
 
 static const uint64_t GB_BYTES = 1000000000LL;
 /* Minimum free space (in GB) needed for data directory */
-static const uint64_t BLOCK_CHAIN_SIZE = 120;
+static const uint64_t BLOCK_CHAIN_SIZE = 140;
```
jonasschnelli: Github PR description says 150MB and I think 150MB is more realistic.

achow101: MB? Looks like a GB to me :stuck_out_tongue:

sipa: I wish.

laanwj: Eh yes, thanks for noticing. Updated to be really 150 exabyte.

PR #10990, opened 2017-08-04
0 locktime issue

gandrewstone:
transactions are being created with a 0 locktime but the intention is to use a locktime that is near the current block to prevent fee sniping

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -2537,7 +2537,9 @@ bool CWallet::FundTransaction(CMutableTransaction& tx, CAmount& nFeeRet, int& nC
         }
     }
 
-
+    // If the transaction has not provided a locktime, let's use the default locktime chosen in CreateTransaction
+    // so that the miner fee sniping prevention implemented there works
+    if (tx.nLockTime == 0) tx.nLockTime = wtx.tx->nLockTime;
     return true;
 }
 
--- a/test/functional/fundrawtransaction.py
+++ b/test/functional/fundrawtransaction.py
@@ -81,6 +81,37 @@ class RawTransactionsTest(BitcoinTestFramework):
         dec_tx  = self.nodes[2].decoderawtransaction(rawtxfund['hex'])
         assert(len(dec_tx['vin']) > 0) #test that we have enough inputs
 
+        #############################
+        # test preserving nLockTime #
+        #############################
+        inputs  = [ ]
+        outputs = { self.nodes[0].getnewaddress() : 1.0 }
+        rawtx   = self.nodes[2].createrawtransaction(inputs, outputs,1234)
+        dec_tx  = self.nodes[2].decoderawtransaction(rawtx)
+        rawtxfund = self.nodes[2].fundrawtransaction(rawtx)
+        fee = rawtxfund['fee']
+        dec_tx  = self.nodes[2].decoderawtransaction(rawtxfund['hex'])
+        assert(dec_tx["locktime"] == 1234)
+
+        ################################
+        # test using default nLockTime #
+        ################################
+        blockcount =  self.nodes[0].getblockcount()
+        inputs  = [ ]
+        outputs = { self.nodes[0].getnewaddress() : 1.0 }
+        rawtx   = self.nodes[2].createrawtransaction(inputs, outputs)
+        dec_tx  = self.nodes[2].decoderawtransaction(rawtx)
+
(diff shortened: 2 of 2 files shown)


src/wallet/wallet.cpp
```diff
@@ -2537,7 +2537,9 @@ bool CWallet::FundTransaction(CMutableTransaction& tx, CAmount& nFeeRet, int& nC
         }
     }
 
-
+    // If the transaction has not provided a locktime, let's use the default locktime chosen in CreateTransaction
+    // so that the miner fee sniping prevention implemented there works
+    if (tx.nLockTime == 0) tx.nLockTime = wtx.tx->nLockTime;
```
promag: Nit, newline after.

test/functional/fundrawtransaction.py
```diff
@@ -81,6 +81,37 @@ def run_test(self):
...
+
+        # there's a random chance of an earlier locktime so iterate a few times
+        for i in range(0,20):
+            rawtxfund = self.nodes[2].fundrawtransaction(rawtx)
+            fee = rawtxfund['fee']
+            dec_tx  = self.nodes[2].decoderawtransaction(rawtxfund['hex'])
+            if dec_tx["locktime"] == blockcount:
+                break
+            assert(dec_tx["locktime"] > 0)
+            assert(i<18)  # incrediably unlikely to never produce the current blockcount
```
promag: Nit, `i < 18` (add spaces).

test/functional/fundrawtransaction.py
```diff
@@ -81,6 +81,37 @@ def run_test(self):
...
+        # test using default nLockTime #
+        ################################
+        blockcount =  self.nodes[0].getblockcount()
+        inputs  = [ ]
+        outputs = { self.nodes[0].getnewaddress() : 1.0 }
+        rawtx   = self.nodes[2].createrawtransaction(inputs, outputs)
+        dec_tx  = self.nodes[2].decoderawtransaction(rawtx)
+
+        # there's a random chance of an earlier locktime so iterate a few times
+        for i in range(0,20):
```
promag: Nit, `(0, 20)` (add space).

test/functional/fundrawtransaction.py
```diff
@@ -81,6 +81,37 @@ def run_test(self):
         dec_tx  = self.nodes[2].decoderawtransaction(rawtxfund['hex'])
         assert(len(dec_tx['vin']) > 0) #test that we have enough inputs
 
+        #############################
+        # test preserving nLockTime #
+        #############################
+        inputs  = [ ]
+        outputs = { self.nodes[0].getnewaddress() : 1.0 }
+        rawtx   = self.nodes[2].createrawtransaction(inputs, outputs,1234)
```
promag: Nit, `, 1234` (add space).

promag (commented):
Nit, improve commit message as it is too long.

achow101:
NACK

The nature of the raw transaction RPCs is that the user is given absolute control over what the transaction will be. We shouldn't be doing anything for the user automatically unless they explicitly asked for that to happen.

gandrewstone:
@achow101 you'd want the nLockTime value to inform the nSequence then...

achow101:
@gandrewstone Yes, so?

gandrewstone:
so its not.  Pick this PR or make another... it took me 5 minutes to point this very small issue out but that's all the time I'm inclined to spend on it.

achow101:
@gandrewstone Your PR does not do that. Your PR is not making the "nLockTime value to inform the nSequence". You don't even touch the nSequence., I agree that nLockTime should inform and partially dictate what the nSequence is going to be, but your PR doesn't do that. It just puts a lock time on transactions when they use `fundrawtransaction` and we don't want that to happen as it is doing something which the user did not tell it to do and that defeats the purpose of the raw transaction RPCs.

jnewbery:
@gandrewstone thanks for your contribution. As @achow101 says, the `_rawtransaction` RPCs are tools for advanced users and `fundrawtransaction` shouldn't be changing the locktime value unexpectedly under the user's feet.

I'm happy to review any future PR where nLockTime informs the nSequence value, but I'm also a NACK for this PR.

jonasschnelli:
I agree with @achow101 and @jnewbery. Auto nLockTime adjustment during fundrawtransaction is not what we should do at this point. I agree a 0 value is not ideal, but changing it during a frt is not what a user would expect.

laanwj:
Ok, agreement seems to close this.

Issue #8266, opened 2016-06-26
Versionbits should have an ignore button


gmaxwell:
It should be possible to configure a node to ignore particular versionbit activations and not warn about them-- because they're not relevant to you.

Config could be something like:
vbaddignore=5:806400   # bit:height 

It should also be possible to preset ignores as network parameters.

Testnet now has a testcase for this as there a long run of blocks with undefined bit 28 set.

It'll be useful to have this in place ahead of the wallet making use of VB warnings in the future (e.g. hiding unconfirmed txn and delaying confirmation indicators when a warning is active).

dcousens:
Does the `height` exactly correlate with the BIP9's `starttime` or is it a generic parameter to ignore only from a particular point in time?

gmaxwell:
It must be either the lock-in time or activation time (I don't care which, they're the same but for the 2016 block offset). It would ignore only that particular use of the bit, if the same bit is recycled and used again, it should still warn.

So it's more like a "not after" rather than a not before. Right now the bip9 logic warns forever once an unknown bit is activated. So this proposes the ability to ignore a specific activation event (which is a height/bit or blockhash/bit pair), but continue to warn on future activation events using that bit.

This lets you look at it and go "oh, this is OP_CHECKSPATULAVERIFY, I don't care about that, won't impact my usage... ignore"   but later when MULCHBABIES softfork happens on the same bit, you'll get warned again and be able to respond with a pitchfork.

pinheadmz:
I think this issue can be closed since the warnings were removed in https://github.com/bitcoin/bitcoin/pull/15471

PR #10994, opened 2017-08-05
Add option to avoid warning on certain network upgrades

ajtowns:
The "-vbignore=bit:start:end" option can be used to prevent bitcoin
from alerting the user of known but uninteresting proposed upgrades
that are being signalled via the version field.

This hopefully provides the feature requested in issue #8266, allow version bits to have an "ignore button". It requires specifying start and end times rather than a final block height, which should map better to BIP 9 implementations. It also allows multiple deployments on a single bit to be ignored independently -- if the bit is used in between any deployments, it will still generate warnings.

This setting will also avoid the bit from triggering the "Unknown block versions being mined" warning during the set period.

### DIFF
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -5,6 +5,7 @@
 
 #include <chainparams.h>
 #include <consensus/merkle.h>
+#include <versionbits.h>
 
 #include <tinyformat.h>
 #include <util.h>
@@ -171,6 +172,8 @@ public:
                         //   (the tx=... number in the SetBestChain debug.log lines)
             3.1         // * estimated number of transactions per second after that timestamp
         };
+
+        vIgnoreBits.clear();
     }
 };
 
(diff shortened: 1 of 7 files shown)


promag (commented):
Missing update to help message.

laanwj:
Concept ACK

Ping @gmaxwell which requested this.

ajtowns:
Rebased against master; updated help text; moved structure definition to versionbits.h; use emplace_back instead of push_back.

Added additional commit that removes the BIP109/bit 28 warning on testnet by default, without the user needing to specify the parameter.

laanwj:
It looks like interest in this got lost a bit due to recent drama, because it is not sure how feasible BIP9 still is in the first place.

But anyhow ping @gmaxwell (again, sorry).

ajtowns:
Rebased; use an enum class instead of a bool for use_ignorebits argument to ComputeBlockVersion.

FWIW, I think version bits would still be useful for BIP8 or BIP148/BIP91-style UASF's; full node users can require miners to signal support (or to not signal support if they want to opt-out of a change) so as to allow two potential chains to be easily distinguished by SPV clients.

ajtowns:
Going to close this until there's some clarity on what method might get used for soft-forks in future (bip 9, bip 8, something else)...

PR #9402, opened 2016-12-21
Allow per network configuration file

rebroad:
Fixes #9374

PR #10996, opened 2017-08-05
Add per-network config file network.conf

ajtowns:
Adds a per-net network.conf as per #9374 ; this is pretty much the same as #9402 which was closed for lack of interest.

It's somewhat interesting in the context of #10994 because it provides somewhere to stick the network-specific vbignore settings. Unless I'm missing something, I don't think #10267 provides an easy way of doing per-network config files, so wouldn't work well for that particular use case.

This patch doesn't include any tests, so is marked as WIP. There's some test code in #10267 that would presumably be able to be adapted pretty easily.

### DIFF
--- a/src/util.h
+++ b/src/util.h
@@ -174,7 +174,7 @@ bool TryCreateDirectories(const fs::path& p);
 fs::path GetDefaultDataDir();
 const fs::path &GetDataDir(bool fNetSpecific = true);
 void ClearDatadirCache();
-fs::path GetConfigFile(const std::string& confPath, bool fNetSpecific = false);
+fs::path GetConfigFile(const std::string& confPath, bool fNetSpecific);
 #ifndef WIN32
 fs::path GetPidFile();
 void CreatePidFile(const fs::path &path, pid_t pid);
@@ -204,7 +204,7 @@ protected:
     std::map<std::string, std::vector<std::string>> mapMultiArgs;
 public:
     void ParseParameters(int argc, const char*const argv[]);
-    void ReadConfigFile(const std::string& confPath, bool fNetSpecific = false);
+    void ReadConfigFile(const std::string& confPath, bool fNetSpecific);
 
     /**
      * Return a vector of strings of the given argument
--- a/src/util.h
+++ b/src/util.h
@@ -56,6 +56,7 @@ extern std::atomic<bool> fReopenDebugLog;
 extern CTranslationInterface translationInterface;
 
 extern const char * const BITCOIN_CONF_FILENAME;
+extern const char * const BITCOIN_NETCONF_FILENAME;
 extern const char * const BITCOIN_PID_FILENAME;
 
 extern std::atomic<uint32_t> logCategories;
@@ -173,7 +174,7 @@ bool TryCreateDirectories(const fs::path& p);
 fs::path GetDefaultDataDir();
 const fs::path &GetDataDir(bool fNetSpecific = true);
 void ClearDatadirCache();
-fs::path GetConfigFile(const std::string& confPath);
+fs::path GetConfigFile(const std::string& confPath, bool fNetSpecific = false);
 #ifndef WIN32
 fs::path GetPidFile();
 void CreatePidFile(const fs::path &path, pid_t pid);
@@ -203,7 +204,7 @@ protected:
     std::map<std::string, std::vector<std::string>> mapMultiArgs;
 public:
     void ParseParameters(int argc, const char*const argv[]);
-    void ReadConfigFile(const std::string& confPath);
+    void ReadConfigFile(const std::string& confPath, bool fNetSpecific = false);
 
     /**
      * Return a vector of strings of the given argument
--- a/src/bitcoin-cli.cpp
+++ b/src/bitcoin-cli.cpp
@@ -107,7 +107,7 @@ static int AppInitRPC(int argc, char* argv[])
         return EXIT_FAILURE;
     }
     try {
-        gArgs.ReadConfigFile(gArgs.GetArg("-conf", BITCOIN_CONF_FILENAME));
+        gArgs.ReadConfigFile(gArgs.GetArg("-conf", BITCOIN_CONF_FILENAME), false);
     } catch (const std::exception& e) {
         fprintf(stderr,"Error reading configuration file: %s\n", e.what());
         return EXIT_FAILURE;
@@ -326,8 +326,7 @@ static UniValue CallRPC(BaseRequestHandler *rh, const std::string& strMethod, co
         if (!GetAuthCookie(&strRPCUserColonPass)) {
             throw std::runtime_error(strprintf(
                 _("Could not locate RPC credentials. No authentication cookie could be found, and RPC password is not set.  See -rpcpassword and -stdinrpcpass.  Configuration file: (%s)"),
-                    GetConfigFile(gArgs.GetArg("-conf", BITCOIN_CONF_FILENAME)).string().c_str()));
-
+                    GetConfigFile(gArgs.GetArg("-conf", BITCOIN_CONF_FILENAME), false).string().c_str()));
         }
     } else {
         strRPCUserColonPass = gArgs.GetArg("-rpcuser", "") + ":" + gArgs.GetArg("-rpcpassword", "");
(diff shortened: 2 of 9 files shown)


laanwj:
Concept ACK

doc/man/bitcoin-cli.1
```diff
@@ -22,6 +22,10 @@ This help message
 .IP
 Specify configuration file (default: bitcoin.conf)
 .HP
+\fB\-netconf=\fR<file>
```
laanwj: I wonder - is there any practical point in being able to override this?

ajtowns: It could be useful if you wanted to temporarily restart bitcoind without using the network.conf file. You could have a bitcoin-alt.conf that specified netconf=network-alt.conf so that you could load bitcoind with -conf=bitcoin-alt.conf and have that load network specific alternative configurations too. Not sure how realistic/valuable those are, though. Not having the option would be inconsistent; and having the option makes the name of the config file at least a little discoverable...

ajtowns:
Commit 7a71b09 adds basic functional tests; it makes use of -netconf being a modifiable parameter to point bitcoind at different files in different ways and checks that has different results.

Commit 8b48aa5 cleans up the code a little, but also allows you to specify an absolute path for network.conf if you want to for some reason (not that doing so actually makes a lot of sense).

Commit 70f5388 removes the default for fNetSpecific for calls to Get/ReadConfigFile, which seems sensible since the default would need to be false to be backwards compatible with the old calls for bitcoin.conf, but is the opposite to the default for the same parameter for GetDataDir().

ajtowns:
One possibly confusing behaviour: possible to set "-testnet" or "-regtest" in a network.conf where those parameters would be wrong; if you do so, they won't take effect or cause problems because ChainNameFromCommandLine() has already been called and isn't called again; but if some future change to the code does try to call ChainNameFromCommandLine() sometime after startup, and a network.conf sets a wrong parameter, it could blow up in your face.

I'm not sure if this is worth fixing...

(Adding a "static const string *result;" to ChainNameFromCommandLine() to memoise it at first call, and just reuse the result without reexamining the arguments later seems to be sufficient to fix the problem if it is worth fixing. Giving an error if regtest/testnet is specified in network.conf would probably be better, but seems like it violates the abstraction layer a fair bit...)

jnewbery:
I've wanted to rebase and revive #9402 for some time, but I was waiting for #10267, both because this conflicts heavily with that PR, and also because I think once #10267 is merged, then this PR becomes more straightforward.

ghost (approved):

ajtowns:
@jnewbery Hmm, looking at #10267 in a bit more detail, the current patch seems to only allow a single includeconf directive to have an effect anyway, rather than multiple includes or recursive includes; so you could actually get the same effect as that patch with this one by just giving an absolute path to the netconf directory here. It doesn't look to me like there's really much difficulty in combining the two patches either once one or the other gets merged (I'm not seeing much room for either to get much simpler, though -- you still have to finish loading bitcoin.conf then load the chain params to work out which network you're on before you can load the right network.conf).

ajtowns:
Squashed most of the commits, and rebased to current master.

doc/man/bitcoin-cli.1
MeshCollider: Aren't the doc/man/ files automatically generated anyway so shouldn't be modified directly?

ajtowns: @MeshCollider Yep, that's correct; worse the help text for bitcoin-cli is missing. Will fix.

ajtowns:
Rebased to current master. Also: no longer committing changes to manpages, functional test updated due to API changes on master, help text for bitcoin-cli --netconf added, and minor code style fixes.

ajtowns:
Rebased to fix test_runner conflicts

ajtowns:
@MeshCollider In normal usage, -netconf isn't needed -- you just add the network.conf file to .bitcoin/testnet3/ (or wherever) if you want to use it, and you don't if you don't want to use it. The main benefits to having the option are (a) it makes it easier to test (can just specify different config files to check different behaviour, rather than having to move files around), and (b) it means it shows up when you say --help so people have some chance of knowing it exists. I think that's enough to justify having the option, but happy to drop it if people think that makes it simpler.

MeshCollider:
Upon discussion on IRC: I suggested using network.conf if `-conf` is not specified but using the `-conf` if it is and not using network.conf at all. This removes the need for the netconf argument. Reasoning is that is a user specified a config file they probably want to use it on the network they specify as well, so we really only need to change the default behaviour. But unsure if that's too limiting

sipa:
Concept ACK

ajtowns:
@MeshCollider see https://github.com/ajtowns/bitcoin/pull/2 . not a big fan of repeating the conditionals still, but it's not too bad.

src/util.h
```diff
@@ -200,7 +201,7 @@ class ArgsManager
     std::map<std::string, std::vector<std::string>> mapMultiArgs;
 public:
     void ParseParameters(int argc, const char*const argv[]);
-    void ReadConfigFile(const std::string& confPath);
+    void ReadConfigFile(const std::string& confPath, bool fNetSpecific);
```
promag: Same as above.

src/util.h
```diff
@@ -171,7 +172,7 @@ bool TryCreateDirectories(const fs::path& p);
 fs::path GetDefaultDataDir();
 const fs::path &GetDataDir(bool fNetSpecific = true);
 void ClearDatadirCache();
-fs::path GetConfigFile(const std::string& confPath);
+fs::path GetConfigFile(const std::string& confPath, bool fNetSpecific);
```
promag: Set default value `bool fNetSpecific = false`? In that case remove changes where it's false.

promag (commented):
Meanwhile some nits.

src/util.h
ajtowns: The "Be explicit whether config file is net specific" commit removes the fNetSpecific=false default argument, because otherwise it would have the opposite default to GetDataDir() which seems confusing to me.

test/functional/netconf.py
```diff
@@ -0,0 +1,82 @@
...
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""
+Tests the network-specific config file
+"""
+
+import os
+from test_framework.test_framework import BitcoinTestFramework
+
+class IncludeConfTest(BitcoinTestFramework):
```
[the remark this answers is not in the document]
ajtowns: Changed.

test/functional/netconf.py
```diff
@@ -0,0 +1,82 @@
...
+        #   order (command line, bitcoin.conf, network.conf). tested
+        #   using uacomment sub-version.
+
+        # by default, loads network.conf
+        self.check_subversion("main", "net")
+        self.check_maxmempool(111)
+
+        # otherwise loads specified network conf
+        self.stop_node(0)
+        self.start_node(0, ["-conf=bitcoin-altnet.conf"])
```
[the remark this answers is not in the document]
ajtowns: Changed.

test/functional/netconf.py
```diff
@@ -0,0 +1,82 @@
...
+        self.check_subversion("mainalt", "altnet", "extra")
+        self.check_maxmempool(333)
+
+        # netconf on command line overrides specification in bitcoin.conf
+        self.stop_node(0)
+        self.start_node(0, ["-conf=bitcoin-altnet.conf", "-netconf=network.conf"])
```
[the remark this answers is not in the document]
ajtowns: Changed.

test/functional/netconf.py
```diff
@@ -0,0 +1,82 @@
...
+        self.check_subversion("mainalt", "net")
+        self.check_maxmempool(333)
+
+        # if file doesn't exist, everything is fine
+        self.stop_node(0)
+        self.start_node(0, ["-netconf=no-network.conf"])
```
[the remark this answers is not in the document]
ajtowns: Changed.

test/functional/netconf.py
```diff
@@ -0,0 +1,82 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""
+Tests the network-specific config file
```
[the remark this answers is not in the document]
ajtowns: Changed.

ajtowns:
Rebased, small improvements to test case, nit fixes.

src/util.h
ajtowns: Err, the implication being that just dropping/not merging that commit is how you'd get the fNetSpecific=false default.

ajtowns:
Rebased to fix test_runner.py conflict, renamed test case to conform to naming convention of #11796

MeshCollider:
This is now superseded by https://github.com/bitcoin/bitcoin/pull/11862

ajtowns:
Abandoning in favour of #11862

PR #10997, opened 2017-08-06
RPC: Add option -stdinrpcpass to bitcoin-cli to allow RPC password to be read from standard input

jharvell:
Add a new command-line option to bitcoin-cli that allows the RPC password to be read from standard intput.  The purpose of this option is to allow secure RPC password input to bitcoin-cli through an external program that is capable of disabling terminal echo.

This option works similarly to the existing -stdin option, and also works when combined with that option.

I have also written a simple ncurses based program that disables echo, gets input from the terminal and writes to standard output.  I couldn't find an existing askpass program that doesn't require graphics libraries, since they are primarily used for getting passwords in a graphics environment.  Unless someone can point out a suitable existing askpass program, I plan to submit my ncurses program to the contrib directory separately from this pull request.

### DIFF
--- a/src/bitcoin-cli.cpp
+++ b/src/bitcoin-cli.cpp
@@ -45,7 +45,8 @@ std::string HelpMessageCli()
     strUsage += HelpMessageOpt("-rpcuser=<user>", _("Username for JSON-RPC connections"));
     strUsage += HelpMessageOpt("-rpcpassword=<pw>", _("Password for JSON-RPC connections"));
     strUsage += HelpMessageOpt("-rpcclienttimeout=<n>", strprintf(_("Timeout in seconds during HTTP requests, or 0 for no timeout. (default: %d)"), DEFAULT_HTTP_CLIENT_TIMEOUT));
-    strUsage += HelpMessageOpt("-stdin", _("Read extra arguments from standard input, one per line until EOF/Ctrl-D (recommended for sensitive information such as passphrases)"));
+    strUsage += HelpMessageOpt("-stdinrpcpass", strprintf(_("Read RPC password from standard input as a single line.  When combined with -stdin, the first line from standard input is used for the RPC password.")));
+    strUsage += HelpMessageOpt("-stdin", _("Read extra arguments from standard input, one per line until EOF/Ctrl-D (recommended for sensitive information such as passphrases).  When combined with -stdinrpcpass, the first line from standard input is used for the RPC password."));
     strUsage += HelpMessageOpt("-rpcwallet=<walletname>", _("Send RPC for non-default wallet on RPC server (argument is wallet filename in bitcoind directory, required if bitcoind/-Qt runs with multiple wallets)"));
 
     return strUsage;
@@ -190,7 +191,7 @@ static void http_error_cb(enum evhttp_request_error err, void *ctx)
 }
 #endif
 
-UniValue CallRPC(const std::string& strMethod, const UniValue& params)
+static UniValue CallRPC(const std::string& strMethod, const UniValue& params)
 {
     std::string host;
     // In preference order, we choose the following for the port:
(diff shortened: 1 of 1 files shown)


jharvell:
I created an askpass utility program (https://github.com/jharvell/askpass)  since I didn't find one to my liking.  In particular, this one supports mult-line input which would be needed for -stdinrpcpass combined with -stdin.

src/bitcoin-cli.cpp
```diff
@@ -293,6 +302,13 @@ int CommandLineRPC(int argc, char *argv[])
             argc--;
             argv++;
         }
+        std::string rpcPass;
+        if (GetBoolArg("-stdinrpcpass", false)) {
+            if(!std::getline(std::cin,rpcPass)) {
+                std::cerr<<"error: -stdinrpcpass specified but failed to read from standard input\n";
+                return EXIT_FAILURE;
+            }
```
laanwj: Suggestion: `SetArg("-rpcpassword", rpcPass)` here, I think this avoids most of the other changes.

src/bitcoin-cli.cpp
```diff
@@ -293,6 +302,13 @@ int CommandLineRPC(int argc, char *argv[])
             argc--;
             argv++;
         }
+        std::string rpcPass;
+        if (GetBoolArg("-stdinrpcpass", false)) {
+            if(!std::getline(std::cin,rpcPass)) {
+                std::cerr<<"error: -stdinrpcpass specified but failed to read from standard input\n";
```
laanwj: Please use `throw std::runtime_error("-stdinrpcpass specified but failed to read from standard input")`, this will do automatically the right thing (like adding error: in front, adding a newline and setting the return code).

jharvell: done

src/bitcoin-cli.cpp
jharvell: done

jharvell:
I realized I had committed with the wrong author email.  So I did a rebase where I amended the author email.  Both commits are the same content, just now with the correct author email (and hash).

src/bitcoin-cli.cpp
```diff
@@ -222,12 +223,18 @@ UniValue CallRPC(const std::string& strMethod, const UniValue& params)
...
-
+                _("Could not locate RPC credentials. No authentication cookie could be found, and RPC password is not set.  See -rpcpassword and -stdinrpcpass.  Configuration file: (%s)"),
+                GetConfigFile(GetArg("-conf", BITCOIN_CONF_FILENAME)).string().c_str()));
         }
     } else {
-        strRPCUserColonPass = GetArg("-rpcuser", "") + ":" + GetArg("-rpcpassword", "");
+        strRPCUserColonPass = GetArg("-rpcuser", strRPCUserColonPass);
+        if (strRPCUserColonPass.empty()) {
+            throw std::runtime_error(strprintf(
+                _("Could not locate RPC credentials. No authentication cookie could be found, and RPC user is not set.  See -rpcuser.  Configuration file: (%s)"),
```
laanwj: I don't understand this change. I think it's better to leave it out as it's not relevant to adding `-stdinrpcpass`, it also duplicates the credentials error which is a bit ugly.

jharvell: Without this change, it is possible to specify an RPC password (either through explicit option for via -stdinrpcpass) without specifying an RPC user.  In this case, the authentication token is ":<rpcpass>".  I had assumed that this cold never match and was an error.  Now that I think about it, perhaps this was an intentional mechanism to allow an empty RPC user.

In any case, I will remove this change.

jharvell: done.

Also, let me know if/when to rebase to combine these three commits into one.

laanwj:
Looks good to me now utACK. Would be nice if someone could test, though.

> Also, let me know if you want me to do one last rebase and squash to a single commit if/before you merge to master.

Yes please.

jonasschnelli:
utACK 4bd8775ec3ea1d32a15aad9ceb6fcf21694df9d6 (yes, please squash).

jharvell:
squashed to single commit

laanwj:
Tested ACK 79191f5

src/bitcoin-cli.cpp
```diff
@@ -293,6 +294,12 @@ int CommandLineRPC(int argc, char *argv[])
             argc--;
             argv++;
         }
+        std::string rpcPass;
+        if (gArgs.GetBoolArg("-stdinrpcpass", false)) {
+            if(!std::getline(std::cin,rpcPass))
```
promag: Nit, missing spaces after `if` and `,` and missing `{`.

@laanwj I'm writing a test for bitcoin-cli, maybe add a commit there to fix these?

PR #10999, opened 2017-08-07
Fix amounts formatting in `decoderawtransaction`

laanwj:
With this, the amounts returned in `decoderawtransaction` will be padded to 8 digits like anywhere else in the API.

This is accomplished by using `ValueFromAmount` in `TxToUniv`, instead of `FormatMoney` which it currently (mistakingly) uses. The `FormatMoney` function is only for debugging/logging use!

To avoid dependency issues, `ValueFromAmount` is moved to `core_write.cpp`, where it also fits better. I don't move `AmountFromValue` to `core_read.cpp` at the same time, as this would have more impact due to the RPCError dependency there.

(n.b.: large number of changed files is solely due to the util_tests JSONs needing update)


doc/developer-notes.md
```diff
@@ -564,7 +564,7 @@ A few guidelines for introducing and reviewing new RPC interfaces:
   - *Exception*: AmountToValue can parse amounts as string. This was introduced because many JSON
```
jnewbery: Please also update `AmountToValue` here.

jnewbery (changes requested):
Seems reasonable to me. Tested ACK https://github.com/bitcoin/bitcoin/pull/10999/commits/29f42b0fcbd34b7ac2ef982d94ce82adb55981b3

You say in the PR text that "The FormatMoney function is only for debugging/logging use!", but I don't see any mention of that in the source. Can you add a comment to the declaration for `FormatMoney()` in `utilmoneystr.h` to make that clear?

Note that `FormatMoney()` is used in a couple of other places:

- the `errors` field in the `bumpfee` return object
- `COutput::ToString()`. We don't actually use `ToString()` anywhere so we could probably just remove that function.

gmaxwell (approved):
utACK. Thanks.

promag:
ACK 29f42b0.

But why does it have to be padded? 🙄

laanwj:
> You say in the PR text that "The FormatMoney function is only for debugging/logging use!", but I don't see any mention of that in the source. Can you add a comment to the declaration for FormatMoney() in utilmoneystr.h to make that clear?

It's fine to use it for user-facing things. `COutput::ToString()` is clearly for logging/debugging.

> the errors field in the bumpfee return object

It's impossible to use AmountToValue there because it's interposed into a string. Also that's just a text message for human consumption. Nothing should parse it. Same for other error messages.

> But why does it have to be padded? :roll_eyes:

We don't want to assume any intelligence at the side of the parser. It should be possible to just remove the '.' and have integer units.
This is the API that we standardized on a long time ago, so it must be used everywhere for consistency, we are not going to change this.

doc/developer-notes.md
laanwj: Thanks, updated.

jnewbery (approved):
Tested ACk https://github.com/bitcoin/bitcoin/pull/10999/commits/ce076383a8578626a7eac37533cba26dece1c877

Thanks for the clarifying comment.

achow101:
utACK ce076383a8578626a7eac37533cba26dece1c877

PR #11001, opened 2017-08-07
[tests] Test disconnecting unsupported service bits logic.

jnewbery:
In v0.15, we disconnect nodes that send us version messages with
unsupported service bits (1 << 5 and 1 << 7). This commit adds a test
that bitcoind will disconnect those nodes before August 1st 2018, and won't
disconnect those nodes after August 1st 2018.

@sdaftuar @TheBlueMatt

### DIFF
--- a/test/functional/p2p-leaktests.py
+++ b/test/functional/p2p-leaktests.py
@@ -9,7 +9,10 @@ received a VERACK.
 
 This test connects to a node and sends it a few messages, trying to intice it
 into sending us something it shouldn't.
-"""
+
+Also test that nodes that send unsupported service bits to bitcoind are disconnected
+and don't receive a VERACK. Unsupported service bits are currently 1 << 5 and
+1 << 7 (until August 1st 2018)."""
 
 from test_framework.mininode import *
 from test_framework.test_framework import BitcoinTestFramework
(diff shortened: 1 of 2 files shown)


jnewbery:
@laanwj - should be tagged v0.15 since it's a regression test for #10982

TheBlueMatt:
Thanks, looks good to me.

MarcoFalke:
utACK 5e35cd94c18d81cbcc7fdece484822841ded402c

bmagn:
Why not relax the rule to instead allow for a maximum of e.g. 3 connections to nodes with the service bits 5 and 7? That way the network may stay fully connected both before and after a 2x HF in three months.

MarcoFalke:
This was meant as a minimal fix/diff to be included in the first release candidate of 0.15.0, which was already due last Sunday. Beside the additional review required for more complex changes to p2p logic, it is unlikely that a more involved fix makes it into 0.15.0. Code cleanup for 0.16.0 is welcome, though.

TheBlueMatt:
The 2x/Bitcoin networks will stay trivially connected through the next few months. There is no chance that everyone running 0.13/0.14 nodes will upgrade to 15 that quickly - we've never seen upgrades anywhere near that fast in the past, even when there was more incentive to do so.

jnewbery:
Please can we lock this PR? This PR added a new functional test case is not the place to discuss the merits/demerits of the node disconnection behaviour in 10982. Thanks!

PR #11002, opened 2017-08-07
[wallet] return correct error code from resendwallettransaction

jnewbery:
New code in #10995 uses `RPC_INVALID_REQUEST`. According to the comment in rpc/protocol.h:
```
// RPC_INVALID_REQUEST is internally mapped to HTTP_BAD_REQUEST (400).
// It should not be used for application-layer errors.
```
Change the returned error code to `RPC_WALLET_ERROR`

#11000 will need to be updated to test for the correct error code.


MarcoFalke:
utACK 055d95f842e3659c41ad55101df8dce4865cd68f

promag (commented):
Tested ACK 055d95f.

laanwj:
Good catch.

PR #11005, opened 2017-08-08
Tests: Add a lint check for trailing whitespace

eklitzke:
This adds a new CHECK_DOC check that looks for newly introduced trailing
whitespace. Existing trailing whitespace (of which there is plenty!)
will not trigger an error.

This is written in a generic way so that new lint-*.sh scripts can be
added to contrib/devtools/, as I'd like to contribute additional lint
checks in the future.

Example of what a lint failure looks like in Travis: https://travis-ci.org/eklitzke/bitcoin/jobs/262052953

### DIFF
--- a/contrib/devtools/lint-whitespace.sh
+++ b/contrib/devtools/lint-whitespace.sh
@@ -1,0 +1,47 @@
+#!/bin/bash
+#
+# Copyright (c) 2017 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+#
+# Check for new lines in diff that introduce trailing whitespace.
+
+# We can't run this check unless we know the commit range for the PR.
+if [ -z "${TRAVIS_COMMIT_RANGE}" ]; then
+  exit 0
+fi
+
+showdiff() {
+  if ! git diff "${TRAVIS_COMMIT_RANGE}" --; then
+    echo "Failed to get a diff"
+    exit 1
+  fi
+}
+
+# Do a first pass, and if no trailing whitespace was found then exit early.
+if ! showdiff | egrep -q '^\+.*\s+$'; then
+  exit
+fi
+
+echo "This diff appears to have added new lines with trailing whitespace."
+echo "The following changes were suspected:"
+
+FILENAME=""
+SEEN=0
+
+while read -r line; do
+  if [[ "$line" =~ ^diff ]]; then
+    FILENAME="$line"
+    SEEN=0
+  else
+    if [ "$SEEN" -eq 0 ]; then
+      # The first time a file is seen with trailing whitespace, we print the
+      # filename (preceded by a newline).
+      echo
+      echo "$FILENAME"
+      SEEN=1
+    fi
+    echo "$line"
+  fi
+done < <(showdiff | egrep '^(diff --git |\+.*\s+$)')
+exit 1
--- a/contrib/devtools/lint-whitespace.sh
+++ b/contrib/devtools/lint-whitespace.sh
@@ -12,14 +12,14 @@ if [ -z "${TRAVIS_COMMIT_RANGE}" ]; then
 fi
 
 showdiff() {
-  if ! git diff "${TRAVIS_COMMIT_RANGE}" --; then
+  if ! git diff -U0 "${TRAVIS_COMMIT_RANGE}" --; then
     echo "Failed to get a diff"
     exit 1
   fi
 }
 
 # Do a first pass, and if no trailing whitespace was found then exit early.
-if ! showdiff | egrep -q '^\+.*\s+$'; then
+if ! showdiff | grep -E -q '^\+.*\s+$'; then
   exit
 fi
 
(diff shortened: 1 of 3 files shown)


contrib/devtools/lint-whitespace.sh
```diff
@@ -0,0 +1,47 @@
...
+#
+# Check for new lines in diff that introduce trailing whitespace.
+
+# We can't run this check unless we know the commit range for the PR.
+if [ -z "${TRAVIS_COMMIT_RANGE}" ]; then
+  exit 0
+fi
+
+showdiff() {
+  if ! git diff "${TRAVIS_COMMIT_RANGE}" --; then
```
MarcoFalke: I assume you need to set the context lines to 0 to prevent false positives?

`-U0`

maflcko (commented):
Concept ACK, as trailing white space causes issues with some editors that remove it automatically on save from the whole file, even from untouched lines.

contrib/devtools/lint-whitespace.sh
```diff
@@ -0,0 +1,47 @@
...
+    echo "Failed to get a diff"
+    exit 1
+  fi
+}
+
+# Do a first pass, and if no trailing whitespace was found then exit early.
+if ! showdiff | egrep -q '^\+.*\s+$'; then
```
practicalswift: Nit: Use grep -E instead of egrep which is deprecated :-)

contrib/devtools/lint-whitespace.sh
```diff
@@ -0,0 +1,47 @@
...
+    if [ "$SEEN" -eq 0 ]; then
+      # The first time a file is seen with trailing whitespace, we print the
+      # filename (preceded by a newline).
+      echo
+      echo "$FILENAME"
+      SEEN=1
+    fi
+    echo "$line"
+  fi
+done < <(showdiff | egrep '^(diff --git |\+.*\s+$)')
```
practicalswift: Nit: Use `grep -E` instead of `egrep` which is deprecated :-)

practicalswift:
Concept ACK

contrib/devtools/lint-whitespace.sh
eklitzke: That doesn't hurt, although I think it's correct as is. Context lines are shown with a leading space when they have no changes, so a context line that starts with `+` will actually have a leading space.

contrib/devtools/lint-whitespace.sh
eklitzke: Serious nit, haha.

eklitzke:
@MarcoFalke @practicalswift Thanks for the suggestions, I've made these changes.

eklitzke:
If this is merged, I'll submit some whitespace-only PRs to clean up various files in the project (probably starting with markdown files, and then moving onto code if that goes well). I feel like it's not worth sending out any whitespace PRs though until this has landed, since otherwise there could be new regressions.

fanquake:
@eklitzke Whitespace only PRs aren't likely to be accepted. Basically they disrupt other work and patches, and create rebasing issues, without providing much value. See the first point in the [dev notes](https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md). "Do not submit patches solely to modify the style of existing code." 

We also have [clang-format rules](https://github.com/bitcoin/bitcoin/blob/master/src/.clang-format) that can be used against new/modified code, which should handle formatting.

eklitzke:
@fanquake Noted. Although I did have problems with `clang-format`, since many of the existing files are not formatted properly, so running those files through `clang-format` caused needless code churn. In any event, this check will prevent the introduction of new trailing whitespace.

kallewoof (approved):

MarcoFalke:

practicalswift:
Can we ignore violations introduced in the directories of our imported dependencies?

I'm thinking about:
* `src/leveldb/`
* `src/secp256k1/`
* `src/univalue/`

Also, what about adding a similar check for introductions of new `\t`:s in files with suffixes `*.cpp`, `*.h`, `*.md`, `*.py` and `*.sh`? Again excluding the imported dependencies.

MarcoFalke:
@eklitzke Are you still working on this?

contrib/devtools/lint-whitespace.sh
```diff
@@ -0,0 +1,47 @@
...
+#
+# Copyright (c) 2017 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+#
+# Check for new lines in diff that introduce trailing whitespace.
+
+# We can't run this check unless we know the commit range for the PR.
+if [ -z "${TRAVIS_COMMIT_RANGE}" ]; then
+  exit 0
```
ryanofsky: Why exit as if the check was successful instead of showing an error?

ryanofsky (commented):
utACK d8a7add0a1f31388eb1235ae11315f00dcf43be3 if the the right files / directories are excluded. It looks like git has a [pathspec](https://git-scm.com/docs/user-manual.html#user-manual-aiddefpathspecapathspec) feature that let you easily exclude paths with `:!` syntax.

Other things that could be done in future prs:

- Incorporate the scripted diff check `contrib/devtools/commit-script-check.sh` into this linting framework instead of running it separately.
- Make it possible to disable or pass options to the linter with directives in the commit message.

MarcoFalke:
Closing due to inactivity. Please ping me to get this reopened.

This is up for grab by other contributors.

PR #6990, opened 2015-11-11
http: speed up shutdown

laanwj:
Pardon a bit of iteration. This continues/fixes #6719.

`event_base_loopexit` was not doing what I expected it to. What I expected was that it sets a timeout, given that no other pending events it would exit in N seconds. However, what it does was delay the event loop exit with N seconds, even if nothing is pending.

Solve it in a different way: give the event loop thread time to exit out of itself, and if it doesn't, send loopbreak (and then join it).

This speeds up the RPC tests a lot, each exit incurred a 10 second overhead, with this change there should be no shutdown overhead in the common case and up to two seconds if the event loop is blocking.

As a bonus this breaks dependency on the global boost::thread_group, as the HTTP server minds its own offspring.

Time for rpctests with this patch:

```
real    7m50.215s
user    0m30.152s
sys     0m11.752s
```

Without:

```
real    13m12.069s
user    0m30.412s
sys     0m11.168s
```

PR #6719, opened 2015-09-24
Make HTTP server shutdown more graceful

laanwj:
Shutting down the HTTP server currently breaks off all current requests. This can create a race condition with RPC `stop` command, where the calling process never receives confirmation.

This change removes the listening sockets on shutdown so that no new requests can come in, but no longer breaks off requests in progress.

Meant to fix #6717.

PR #11006, opened 2017-08-08
Improve shutdown process

promag:
Improve the shutdown time by not having to wait up to 2 seconds.

Here is a comparison running `wallet.py` function tests before this PR:
```
2017-08-08 03:25:20.881000 TestFramework (INFO): Initializing test directory /var/folders/1v/8_69hby54nj2k3n6fywt44x80000gn/T/testq_ramjjr
2017-08-08 03:25:23.853000 TestFramework (INFO): Mining blocks...
2017-08-08 03:25:24.132000 TestFramework (INFO): test getmemoryinfo
2017-08-08 03:25:24.559000 TestFramework (INFO): test gettxout
2017-08-08 03:25:59.858000 TestFramework (INFO): check -rescan
2017-08-08 03:26:07.735000 TestFramework (INFO): check -reindex
2017-08-08 03:26:15.751000 TestFramework (INFO): check -zapwallettxes=1
2017-08-08 03:26:24.105000 TestFramework (INFO): check -zapwallettxes=2
2017-08-08 03:26:36.694000 TestFramework (INFO): Stopping nodes
2017-08-08 03:26:43.599000 TestFramework (INFO): Cleaning up
2017-08-08 03:26:43.612000 TestFramework (INFO): Tests successful
```
After:
```
2017-08-08 03:24:04.319000 TestFramework (INFO): Initializing test directory /var/folders/1v/8_69hby54nj2k3n6fywt44x80000gn/T/testoqeyi50_
2017-08-08 03:24:07.035000 TestFramework (INFO): Mining blocks...
2017-08-08 03:24:07.317000 TestFramework (INFO): test getmemoryinfo
2017-08-08 03:24:07.763000 TestFramework (INFO): test gettxout
2017-08-08 03:24:25.715000 TestFramework (INFO): check -rescan
2017-08-08 03:24:27.792000 TestFramework (INFO): check -reindex
2017-08-08 03:24:29.797000 TestFramework (INFO): check -zapwallettxes=1
2017-08-08 03:24:32.207000 TestFramework (INFO): check -zapwallettxes=2
2017-08-08 03:24:36.812000 TestFramework (INFO): Stopping nodes
2017-08-08 03:24:37.915000 TestFramework (INFO): Cleaning up
2017-08-08 03:24:37.927000 TestFramework (INFO): Tests successful
```
This largely improves the time spent in Travis (under evaluation).

### DIFF
--- a/src/httpserver.cpp
+++ b/src/httpserver.cpp
@@ -481,6 +481,8 @@ void StopHTTPServer()
     }
     if (eventBase) {
         LogPrint(BCLog::HTTP, "Waiting for HTTP event thread to exit\n");
+        // Exit the event loop as soon as there are no active events.
+        event_base_loopexit(eventBase, nullptr);
         // Give event loop a few seconds to exit (to send back last RPC responses), then break it
         // Before this was solved with event_base_loopexit, but that didn't work as expected in
         // at least libevent 2.0.21 and always introduced a delay. In libevent


promag:
With this change the time should be significantly lower than 5 seconds. This was motivated by [@jnewbery comment](https://github.com/bitcoin/bitcoin/pull/11000#pullrequestreview-54752223).

Keep in mind that this is still WIP it requires some more testing.

promag:
For the record, running with version libevent@2.1.8.

If there is a minimum version required, can it be defined, like boost?

jnewbery:
This PR should just have 2195bb0e579336f2b9adfb9a2e23957e794f5621. You can remove the other commits.

You could potentially protect this change with a preprocessor:

```
#if LIBEVENT_VERSION_NUMBER >= 0x02xxxxxx
```

I don't know what version of libevent the `event_base_loopexit` behaviour was fixed. Perhaps @laanwj can offer some clues (he's the author of #6990).

Definitely seems like a good win for test run times.

src/httpserver.cpp
```diff
@@ -481,16 +479,7 @@ void StopHTTPServer()
...
-        // Before this was solved with event_base_loopexit, but that didn't work as expected in
-        // at least libevent 2.0.21 and always introduced a delay. In libevent
-        // master that appears to be solved, so in the future that solution
-        // could be used again (if desirable).
-        // (see discussion in https://github.com/bitcoin/bitcoin/pull/6990)
-        if (threadResult.valid() && threadResult.wait_for(std::chrono::milliseconds(2000)) == std::future_status::timeout) {
-            LogPrintf("HTTP event loop did not exit within allotted time, sending loopbreak\n");
-            event_base_loopbreak(eventBase);
-        }
+        event_base_loopexit(eventBase, nullptr);
```
laanwj: I don't get it - you don't give any rationale for reverting this and remove the comment.

laanwj:
> I don't know what version of libevent the event_base_loopexit behaviour was fixed. Perhaps @laanwj can offer some clues (he's the author of #6990).

No, I don't know, and we certainly can't assume that everyone is building against a version where this is fixed.

src/httpserver.cpp
laanwj: As I understand it, removing the timeout here (resulting in instant shutdown) means that responses from RPC `stop` might not come back, as the http event handler is killed before that could finish.

promag: No, that happens when evhttp_free is called.

promag:
@jnewbery yes extra commits were mistakenly pushed when testing with other branch. Will fix.

src/httpserver.cpp
laanwj: I'm sure that at least used to be the case, we had problems here in the past, this is why this particular shutdown sequence was changed so many times.

promag:
@laanwj what about keeping the older code for older libevents?

laanwj:
> @laanwj what about keeping the older code for older libevents?

Yes, I'd say that is a requirement.

promag:
@laanwj @jnewbery will try to dissect which version has bad behaviour.

Will add a comment before `event_base_loopexit`.

promag:
With `libevent@2.0.21` (the version that supposedly has issues with `event_base_loopexit` everything seems to work (at least on my system). Maybe the problem was in other place? I'll dig a bit more.

laanwj:
>  everything seems to work (at least on my system)

As I remember the intermittent errors on issuing `stop`, might well not happen locally in good-weather conditions. As we all know, travis VMs run on overbooked, extremely busy servers, so it's more likely to happen there. Maybe it can be simulated by adding an artificial delay somewhere, my hunch would be to have the client side read the RPC response to `stop` after a delay.

jnewbery:
@laanwj - forgive me, I'm trying to piece together the history of changes to the shutdown procedure:

- In #6719, you changed from using `event_base_loopbreak()` to `event_base_loopexit()` with a timeout. That was to make sure that the server shut down gracefully and didn't just drop responses to pending requests.
- In #6990, you say that '`event_base_loopexit` was not doing what I expected it to. What I expected was that it sets a timeout, given that no other pending events it would exit in N seconds. However, what it does was delay the event loop exit with N seconds, even if nothing is pending.' I can't understand the difference between those two statements. That PR switches from calling `event_base_loopexit()` to call `event_base_loopbreak()` after a 2 second delay.
- Later in the same PR, you say 'it appears that with libevent master, this issue doesn't exist.
`event_base_loopexit` does actually do what I expect it to in the OP.' Looking at the documentation, if `event_base_loopexit` is called with a time value, then it will always wait for that long before exiting. If it's called without a time value, it'll exit after the next base loop iteration (http://www.wangafu.net/~nickm/libevent-2.1/doxygen/html/event_8h.html#ab89f0c907906f784fc80a3107016c3dc)

So I think what we want is to just call `event_base_loopexit()` without a time value. That would allow the current loop to exit, so shutdown should be graceful. I can't tell from the history or your notes whether you tried that before. Is there any reason that wouldn't work? (specifically, I'm just looking for your issue with `event_base_loopexit` was)

jnewbery:
Oh, and Concept ACK:

locally with this PR:
```
TEST                           | STATUS    | DURATION

abandonconflict.py             | ✓ Passed  | 4 s 
...
zmq_test.py                    | ✓ Passed  | 2 s 

ALL                            | ✓ Passed  | 630 s (accumulated)
Runtime: 162 s
```
locally without this PR:
```
TEST                           | STATUS    | DURATION

abandonconflict.py             | ✓ Passed  | 14 s
...
zmq_test.py                    | ✓ Passed  | 7 s 

ALL                            | ✓ Passed  | 985 s (accumulated)
```

Seems like a huge win (40% over the BASE_SCRIPTS test suite)

promag:
Nice @jnewbery. Furthermore, the the docs say
```
tv	the amount of time after which the loop should terminate, or NULL to exit after running all currently active events.
```
And while the RPC response is being handled, there is an active event.

promag:
IMO we should just add the loop exit, and keep the break after the timeout.

jnewbery:
> IMO we should just add the loop exit, and keep the break after the timeout.

Sounds reasonable if you're able to test this against old versions of libevent (and pending any further background info from @laanwj)

promag:
@laanwj you mean:
```diff
diff --git a/test/functional/test_framework/authproxy.py b/test/functional/test_framework/authproxy.py
index b3671cbdc..5a2f9c514 100644
--- a/test/functional/test_framework/authproxy.py
+++ b/test/functional/test_framework/authproxy.py
@@ -166,6 +166,8 @@ class AuthServiceProxy(object):
     def _get_response(self):
         req_start_time = time.time()
         try:
+            if self._service_name == 'stop':
+                time.sleep(3)
             http_response = self.__conn.getresponse()
         except socket.timeout as e:
             raise JSONRPCException({
```
All looks good, even with `libevent@2.0.21`.

laanwj:
utACK 793667a, my concerns have been addressed.

PR #11007, opened 2017-08-08
wallet: Fix potential memory leak when loading a corrupted wallet file

practicalswift:
Fix potential memory leak when loading a corrupted wallet file.

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -3937,15 +3937,12 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
         uiInterface.InitMessage(_("Zapping all transactions from wallet..."));
 
         std::unique_ptr<CWalletDBWrapper> dbw(new CWalletDBWrapper(&bitdb, walletFile));
-        CWallet *tempWallet = new CWallet(std::move(dbw));
+        std::unique_ptr<CWallet> tempWallet(new CWallet(std::move(dbw)));
         DBErrors nZapWalletRet = tempWallet->ZapWalletTx(vWtx);
         if (nZapWalletRet != DB_LOAD_OK) {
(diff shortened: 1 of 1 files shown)


src/wallet/wallet.cpp
```diff
@@ -3886,15 +3886,12 @@ CWallet* CWallet::CreateWalletFromFile(const std::string walletFile)
         uiInterface.InitMessage(_("Zapping all transactions from wallet..."));
 
         std::unique_ptr<CWalletDBWrapper> dbw(new CWalletDBWrapper(&bitdb, walletFile));
-        CWallet *tempWallet = new CWallet(std::move(dbw));
+        std::unique_ptr<CWallet> tempWallet(new CWallet(std::move(dbw)));
```
eklitzke: nit: If you move the previous line into this one, you can omit the `std::move` operation.

gmaxwell:
Please avoid titling PR's with things like "Fix memory leak" for things where there isn't an actual leak in practice. (e.g. because it exits).  Changes like this are good but they mess up people trying to figure out if an issue they are encounter has been fixed ("I have a memory leak, oh they just fixed one. They's probably it.").

gmaxwell (approved):
utACK

practicalswift:
@gmaxwell That's a good point! I've now changed to "Fix potential memory leak" - works or still room for improvement? :-)

MarcoFalke:
utACK c8b4b80df8346739c9a52113fd6673db1574e1f3

practicalswift:
Rebased!

src/wallet/wallet.cpp
jonasschnelli: I think @eklitzke is right here that it would be good to avoid that std::move.

promag: Too long?

PR #11008, opened 2017-08-08
Enable disablesafemode by default.

gmaxwell:
Safemode is almost useless as is-- it only triggers in limited
 cases most of which aren't even concerning. There have been
 several proposals to remove it. But as a simpler, safer, and
 more flexible first case, simply deactivate it by default.

Anyone who wants it can re-enable and know what they've signed up for.


gmaxwell:
The fact this passes travis shows the functionality has no tests, too.

achow101:
utACK

laanwj:
Seems a good precursor to #10563 "Remove safe mode".

instagibbs:
utACK

laanwj:
utACK

PR #11009, opened 2017-08-08
Add information about the next state in getblockchaininfo rpc request

rawodb:


rawodb:
Will resubmit from different branch.

PR #11012, opened 2017-08-09
Make sure to clean up mapBlockSource if we've already seen the block

theuni:
Otherwise we may leave them dangling.

Credit TheBlueMatt.

### DIFF
--- a/src/net_processing.cpp
+++ b/src/net_processing.cpp
@@ -2134,9 +2134,12 @@ bool static ProcessMessage(CNode* pfrom, const std::string& strCommand, CDataStr
             }
             bool fNewBlock = false;
             ProcessNewBlock(chainparams, pblock, true, &fNewBlock);
-            if (fNewBlock)
+            if (fNewBlock) {
                 pfrom->nLastBlockTime = GetTime();
-
+            } else {
+                LOCK(cs_main);
+                mapBlockSource.erase(pblock->GetHash());
+            }
             LOCK(cs_main); // hold cs_main for CBlockIndex::IsValid()
             if (pindex->IsValid(BLOCK_VALID_TRANSACTIONS)) {
                 // Clear download state for this block, which is in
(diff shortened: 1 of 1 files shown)


src/net_processing.cpp
```diff
@@ -2134,9 +2134,12 @@ bool static ProcessMessage(CNode* pfrom, const std::string& strCommand, CDataStr
             }
             bool fNewBlock = false;
             ProcessNewBlock(chainparams, pblock, true, &fNewBlock);
-            if (fNewBlock)
+            if (fNewBlock) {
                 pfrom->nLastBlockTime = GetTime();
-
+            } else {
+                LOCK(cs_main);
```
promag: Nit, move the lock up and remove the one below?

promag (commented):
Nit, reword `Remove existing blocks from mapBlockSource`?

laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/11012/commits/3f8fa7f58b68a8ed596c62d1edb85a3455a5a724

src/net_processing.cpp
```diff
@@ -2390,8 +2397,12 @@ bool static ProcessMessage(CNode* pfrom, const std::string& strCommand, CDataStr
         }
         bool fNewBlock = false;
         ProcessNewBlock(chainparams, pblock, forceProcessing, &fNewBlock);
-        if (fNewBlock)
+        if (fNewBlock) {
```
laanwj: Would be nice to factor this out to a function, as we're repeating exactly the same code three times.

theuni: Would you like me to do that here? Either way works for me.

laanwj: I don't care strongly, I'm fine with having this as quick fix only. But it feels a tad brittle.

theuni: I believe that @TheBlueMatt had plans to clean this up further (or remove the need) in the future, so if it's all the same, I'd prefer to leave it as-is for 0.15.

jnewbery:
utACK 3f8fa7f58b68a8ed596c62d1edb85a3455a5a724

promag:
utACK 3f8fa7f. Should we open an issue to clean this in the future as @theuni suggests?

Issue #3141, opened 2013-10-24
Automatic filter should be delayed until input is paused


kroese:
When searching for transactions, the grid begins to query the database while typing, which is fine. But when using very large wallets, this will freeze the interface so much that it becomes impossible to finish typing. Adding a very small delay would solve this issue, since it would allow me to finish my input before the client freezes.

For example, when I try to look up transactions greater than 999 BTC, it will first start searching for all transactions > 9 BTC, and then > 99 BTC, and finally 999 BTC, even though I'm typing 999 as fast as I possibly can.

jonasschnelli:
Solved by #11015. Closing.

PR #11015, opened 2017-08-09
[Qt] Add delay before filtering transactions

lclc:
As discussed in https://github.com/bitcoin/bitcoin/issues/3141.

This adds a QTimer pause of 200ms before start to filter so it should be possible to filter big data sets easier.

### DIFF
--- a/src/qt/transactionview.cpp
+++ b/src/qt/transactionview.cpp
@@ -33,6 +33,7 @@
 #include <QScrollBar>
 #include <QSignalMapper>
 #include <QTableView>
+#include <QTimer>
 #include <QUrl>
 #include <QVBoxLayout>
 
@@ -112,6 +113,17 @@ TransactionView::TransactionView(const PlatformStyle *platformStyle, QWidget *pa
     amountWidget->setValidator(new QDoubleValidator(0, 1e20, 8, this));
     hlayout->addWidget(amountWidget);
 
+    // Delay before filtering transactions in ms
+    static const int input_filter_delay = 200;
+
+    QTimer* amount_typing_delay = new QTimer(this);
+    amount_typing_delay->setSingleShot(true);
+    amount_typing_delay->setInterval(input_filter_delay);
+
+    QTimer* prefix_typing_delay = new QTimer(this);
+    prefix_typing_delay->setSingleShot(true);
+    prefix_typing_delay->setInterval(input_filter_delay);
+
     QVBoxLayout *vlayout = new QVBoxLayout(this);
     vlayout->setContentsMargins(0,0,0,0);
     vlayout->setSpacing(0);
@@ -173,8 +185,10 @@ TransactionView::TransactionView(const PlatformStyle *platformStyle, QWidget *pa
     connect(dateWidget, SIGNAL(activated(int)), this, SLOT(chooseDate(int)));
     connect(typeWidget, SIGNAL(activated(int)), this, SLOT(chooseType(int)));
     connect(watchOnlyWidget, SIGNAL(activated(int)), this, SLOT(chooseWatchonly(int)));
-    connect(addressWidget, SIGNAL(textChanged(QString)), this, SLOT(changedPrefix(QString)));
-    connect(amountWidget, SIGNAL(textChanged(QString)), this, SLOT(changedAmount(QString)));
+    connect(amountWidget, SIGNAL(textChanged(QString)), amount_typing_delay, SLOT(start()));
+    connect(amount_typing_delay, SIGNAL(timeout()), this, SLOT(changedAmount()));
+    connect(addressWidget, SIGNAL(textChanged(QString)), prefix_typing_delay, SLOT(start()));
+    connect(prefix_typing_delay, SIGNAL(timeout()), this, SLOT(changedPrefix()));
 
     connect(view, SIGNAL(doubleClicked(QModelIndex)), this, SIGNAL(doubleClicked(QModelIndex)));
     connect(view, SIGNAL(customContextMenuRequested(QPoint)), this, SLOT(contextualMenu(QPoint)));
@@ -312,20 +326,19 @@ void TransactionView::chooseWatchonly(int idx)
         (TransactionFilterProxy::WatchOnlyFilter)watchOnlyWidget->itemData(idx).toInt());
 }
 
-void TransactionView::changedPrefix(const QString &prefix)
+void TransactionView::changedPrefix()
 {
     if(!transactionProxyModel)
         return;
-    transactionProxyModel->setAddressPrefix(prefix);
+    transactionProxyModel->setAddressPrefix(addressWidget->text());
 }
 
-void TransactionView::changedAmount(const QString &amount)
+void TransactionView::changedAmount()
 {
     if(!transactionProxyModel)
         return;
     CAmount amount_parsed = 0;
-    if(BitcoinUnits::parse(model->getOptionsModel()->getDisplayUnit(), amount, &amount_parsed))
-    {
+    if (BitcoinUnits::parse(model->getOptionsModel()->getDisplayUnit(), amountWidget->text(), &amount_parsed)) {
         transactionProxyModel->setMinAmount(amount_parsed);
     }
     else
--- a/src/qt/transactionview.h
+++ b/src/qt/transactionview.h
@@ -112,8 +112,8 @@ public Q_SLOTS:
     void chooseDate(int idx);
     void chooseType(int idx);
     void chooseWatchonly(int idx);
-    void changedPrefix(const QString &prefix);
-    void changedAmount(const QString &amount);
+    void changedAmount();
+    void changedPrefix();
     void exportClicked();
     void focusTransaction(const QModelIndex&);
 


src/qt/transactionview.cpp
```diff
@@ -173,8 +180,10 @@ TransactionView::TransactionView(const PlatformStyle *platformStyle, QWidget *pa
     connect(dateWidget, SIGNAL(activated(int)), this, SLOT(chooseDate(int)));
     connect(typeWidget, SIGNAL(activated(int)), this, SLOT(chooseType(int)));
     connect(watchOnlyWidget, SIGNAL(activated(int)), this, SLOT(chooseWatchonly(int)));
-    connect(addressWidget, SIGNAL(textChanged(QString)), this, SLOT(changedPrefix(QString)));
-    connect(amountWidget, SIGNAL(textChanged(QString)), this, SLOT(changedAmount(QString)));
+    connect(addressWidget, SIGNAL(textChanged(QString)), this, SLOT(changedPrefix()));
+    connect(inputDelayPrefix, SIGNAL(timeout()), this, SLOT(filterByPrefix()));
+    connect(amountWidget, SIGNAL(textChanged(QString)), this, SLOT(changedAmount()));
```
promag: Connect to start `QTimer::start()` slot instead, and configure the interval above.

src/qt/transactionview.cpp
```diff
@@ -173,8 +180,10 @@ TransactionView::TransactionView(const PlatformStyle *platformStyle, QWidget *pa
     connect(dateWidget, SIGNAL(activated(int)), this, SLOT(chooseDate(int)));
     connect(typeWidget, SIGNAL(activated(int)), this, SLOT(chooseType(int)));
     connect(watchOnlyWidget, SIGNAL(activated(int)), this, SLOT(chooseWatchonly(int)));
-    connect(addressWidget, SIGNAL(textChanged(QString)), this, SLOT(changedPrefix(QString)));
-    connect(amountWidget, SIGNAL(textChanged(QString)), this, SLOT(changedAmount(QString)));
+    connect(addressWidget, SIGNAL(textChanged(QString)), this, SLOT(changedPrefix()));
```
promag: Connect to start `QTimer::start()` slot instead, and configure the interval above.

src/qt/transactionview.cpp
```diff
@@ -312,19 +321,29 @@ void TransactionView::chooseWatchonly(int idx)
         (TransactionFilterProxy::WatchOnlyFilter)watchOnlyWidget->itemData(idx).toInt());
 }
 
-void TransactionView::changedPrefix(const QString &prefix)
+void TransactionView::changedPrefix()
```
promag: Can be removed.

src/qt/transactionview.cpp
```diff
@@ -312,19 +321,29 @@ void TransactionView::chooseWatchonly(int idx)
...
+
+void TransactionView::filterByPrefix()
 {
     if(!transactionProxyModel)
         return;
-    transactionProxyModel->setAddressPrefix(prefix);
+    transactionProxyModel->setAddressPrefix(addressWidget->text());
+}
+
+void TransactionView::changedAmount()
```
promag: Can be removed.

promag (commented):
Should the interval be configurable? At least it's duplicated so have a const variable?

promag:
Nit, reword commit and PR title to `qt: Add delay before filtering transactions`?

jonasschnelli:
Concept ACK

lclc:
Thanks a lot for the fast review and suggestions for improvement @promag. I've updated the commit accordingly.

> Should the interval be configurable?

I don't think many people want to change that. If we see 200ms is not optimal we can still change it later.
Adding this as a config flag option would probably mostly add code that nobody uses in the end - but if more people think it's a good idea I can implement it.

sipa:
Don't make things configurable if there is no advice for what to set it to.

src/qt/transactionview.h
```diff
@@ -69,6 +70,12 @@ class TransactionView : public QWidget
     QLineEdit *addressWidget;
     QLineEdit *amountWidget;
 
+    QTimer *inputDelayPrefix;
+    QTimer *inputDelayAmount;
+
+    /* Delay before before filtering transactions in ms */
+    static const int inputFilterDelay = 200;
```
promag: Move to `.cpp`?

src/qt/transactionview.h
```diff
@@ -69,6 +70,12 @@ class TransactionView : public QWidget
     QLineEdit *addressWidget;
     QLineEdit *amountWidget;
 
+    QTimer *inputDelayPrefix;
```
promag: The convention is to prefix with `m_` in new code.

promag (commented):
Agree with @sipa. A variable is enough to add meaning to the duration.

src/qt/transactionview.h
lclc: I also like the m_ convention but I don't see it in any of the other Qt code.
Is the plan to slowly change to m_ over time or maybe take the effort to change (part of) the full source code at ones? Two different styles in the same file is not so nice I think.

promag: @lclc unfortunately there's mixed styles all over the place. Last time that was discussed, that I know of, is to use the defined convention in new or "touched" code.

lclc: Ok, thanks. I'll update.

lclc:
Improved as suggested by @promag.

src/qt/transactionview.h
```diff
@@ -69,6 +70,9 @@ class TransactionView : public QWidget
     QLineEdit *addressWidget;
     QLineEdit *amountWidget;
 
+    QTimer *m_inputDelayPrefix;
```
benma: please use snake_case as per developer notes.

src/qt/transactionview.h
```diff
@@ -112,8 +116,8 @@ public Q_SLOTS:
     void chooseDate(int idx);
     void chooseType(int idx);
     void chooseWatchonly(int idx);
-    void changedPrefix(const QString &prefix);
-    void changedAmount(const QString &amount);
+    void changedPrefix();
```
benma: can be moved to the private Q_SLOTS section. All the others can as well, mabye in a separate commit?

benma (commented):
ACK 0587d2e36a0a64535e8d296fb97fcb7a258637fe

I tested that the issue existed and that this PR fixes it.

src/qt/transactionview.h
jonasschnelli: To also drop my opinion here (though I don't care): Use the scheme of the surrounding code to not confuse new developers. New classes or complete rewrites may be different.

sipa: From the developer notes:

> When writing patches, favor the new style over attempting to mimick the surrounding style, except for move-only commits.

jonasschnelli: Hmm.. I see. My advice is against the developer notes then. It just looked confusing to have two single instance variables with sneak case where the rest is without `m_` and camel case. But I guess we have to start somewhere to adopt the style rules.

laanwj:
Concept ACK, thanks!

> Don't make things configurable if there is no advice for what to set it to.

Indeed, just make a choice, there's no use in making this configurable. No one will bother and it will just add more things to test.

src/qt/transactionview.h
lclc: I plan to look at Q_SLOTS in general in a separate PR. There are a few other cases in different files where some slots are public that wouldn't need to be.

lclc:
Thanks a lot for clarifying, review and testing @benma. I changed the commit according to the developer notes.

src/qt/transactionview.h
benma: +1

benma:
ACK 6d073a2d1c85ecc523d13c5f47514daa5de4757c

src/qt/transactionview.h
```diff
@@ -69,6 +70,9 @@ class TransactionView : public QWidget
     QLineEdit *addressWidget;
     QLineEdit *amountWidget;
 
+    QTimer *m_input_delay_prefix;
```
promag: There is no need to have these here..

promag:
utACK 6d073a2.

jonasschnelli:
I couldn't really test this (all the machines where to fast). The delay seems very short (with 200ms). Will the timer be reset when entering the next char <200ms (I think so otherwise this would not work)?.
Maybe @promag can do a test?

promag:
@jonasschnelli yes start() resets.

src/qt/transactionview.cpp
```diff
@@ -112,6 +113,17 @@ TransactionView::TransactionView(const PlatformStyle *platformStyle, QWidget *pa
     amountWidget->setValidator(new QDoubleValidator(0, 1e20, 8, this));
     hlayout->addWidget(amountWidget);
 
+    /* Delay before before filtering transactions in ms */
+    static const int inputFilterDelay = 200;
```
promag: Use snake case. Comment to indicate unit?

FWIW there is this [global constant](https://github.com/bitcoin/bitcoin/blame/master/src/qt/guiconstants.h#L9) cc @sipa.

src/qt/transactionview.cpp
```diff
@@ -312,19 +326,19 @@ void TransactionView::chooseWatchonly(int idx)
...
 }
 
-void TransactionView::changedAmount(const QString &amount)
+void TransactionView::changedAmount()
 {
     if(!transactionProxyModel)
         return;
     CAmount amount_parsed = 0;
-    if(BitcoinUnits::parse(model->getOptionsModel()->getDisplayUnit(), amount, &amount_parsed))
+    if(BitcoinUnits::parse(model->getOptionsModel()->getDisplayUnit(), amountWidget->text(), &amount_parsed))
```
promag: Missing space after `if`. Move below `{` to this line.

src/qt/transactionview.h
promag: Please remove.

src/qt/transactionview.h
```diff
@@ -69,6 +70,9 @@ class TransactionView : public QWidget
...
+    QTimer *m_input_delay_amount;
```
promag: Same as above.

src/qt/transactionview.h
promag: Agree, all is done in this PR context.

src/qt/transactionview.h
```diff
@@ -112,8 +116,8 @@ public Q_SLOTS:
     void chooseDate(int idx);
     void chooseType(int idx);
     void chooseWatchonly(int idx);
-    void changedPrefix(const QString &prefix);
-    void changedAmount(const QString &amount);
+    void changedPrefix();
+    void changedAmount();
```
promag: Nit, I would sort the slots.

src/qt/transactionview.cpp
```diff
@@ -112,6 +113,17 @@ TransactionView::TransactionView(const PlatformStyle *platformStyle, QWidget *pa
     amountWidget->setValidator(new QDoubleValidator(0, 1e20, 8, this));
     hlayout->addWidget(amountWidget);
 
+    /* Delay before before filtering transactions in ms */
+    static const int inputFilterDelay = 200;
+
+    m_input_delay_amount = new QTimer(this);
```
promag: Nit, change to `QTimer* amount_typing_delay = new QTimer(this);`

src/qt/transactionview.cpp
```diff
@@ -112,6 +113,17 @@ TransactionView::TransactionView(const PlatformStyle *platformStyle, QWidget *pa
     amountWidget->setValidator(new QDoubleValidator(0, 1e20, 8, this));
     hlayout->addWidget(amountWidget);
 
+    /* Delay before before filtering transactions in ms */
```
promag: Comments is this file are `// ...`.

promag (changes requested):
For reference https://wiki.qt.io/Delay_action_to_wait_for_user_interaction.

jonasschnelli:
@lclc, could you address @promag's points? I'd like to get this in soon.

lclc:
I'll do it on Tuesday night.

Am 07.09.2017 17:52 schrieb "Jonas Schnelli" <notifications@github.com>:

lclc:
Updated.

Thanks for the reviews @promag - this version is now quite a bit simpler and cleaner by now than the first one :)

promag:
utACK a82976d. Will test in a couple of hours.

promag:
Tested ACK a82976d.

flack:
If the motivation is to make filtering large data sets easier, why not apply the delay only when the data set is sufficiently large? For other cases, it just reduces responsiveness. Or is the size of the data set not known when the UI is constructed?

promag:
200ms is very reasonable. See https://ux.stackexchange.com/a/38545.

jonasschnelli:
Tested ACK a82976d96009680ab61bcea763f15c7594b9afb0

src/qt/transactionview.cpp
```diff
@@ -112,6 +113,17 @@ TransactionView::TransactionView(const PlatformStyle *platformStyle, QWidget *pa
     amountWidget->setValidator(new QDoubleValidator(0, 1e20, 8, this));
     hlayout->addWidget(amountWidget);
 
+    // Delay before before filtering transactions in ms
```
jonasschnelli: nit: comment (before before)

lclc: Thanks, fixed.

promag:
Tested ACK 7b137ac.

PR #11017, opened 2017-08-09
[wallet] Close DB on error.

kallewoof:
This PR intends to plug some leaks. It specifically implements adherence to the requirement in BDB to close a handle which failed to open (https://docs.oracle.com/cd/E17276_01/html/api_reference/C/dbopen.html):

> The `DB->open()` method returns a non-zero error value on failure and 0 on success. If `DB->open()` fails, the `DB->close()` method must be called to discard the DB handle.

### DIFF
--- a/src/wallet/db.cpp
+++ b/src/wallet/db.cpp
@@ -101,8 +101,10 @@ bool CDBEnv::Open(const fs::path& pathIn)
                              DB_RECOVER |
                              nEnvFlags,
                          S_IRUSR | S_IWUSR);
-    if (ret != 0)
+    if (ret != 0) {
+        dbenv->close(0);
         return error("CDBEnv::Open: Error %d opening database environment: %s\n", ret, DbEnv::strerror(ret));
+    }
 
     fDbEnvInit = true;
     fMockDb = false;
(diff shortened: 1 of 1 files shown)


practicalswift:
Concept ACK! Nice find 👍

gmaxwell (approved):
utACK

MarcoFalke:
utACK 03bc719a85cb4928cb4b43d0bc4142f72cb01b23

paveljanik:
utACK 03bc719

src/wallet/db.cpp
```diff
@@ -536,8 +538,10 @@ bool CDB::Rewrite(CWalletDBWrapper& dbw, const char* pszSkip)
                         env->CloseDb(strFile);
                         if (pdbCopy->close(0))
                             fSuccess = false;
-                        delete pdbCopy;
+                    } else {
+                        pdbCopy->close(0);
                     }
+                    delete pdbCopy;
```
laanwj: Shouldn't it always close pdbCopy before deleting it, not just on failure?

kallewoof: Line #537 closes for fSuccess being false and #542 for true. Am I missing a case?

promag (commented):
utACK 03bc719.

BTW, how did you catch this?

kallewoof:
@promag I ran bitcoind through Instruments on a mac, and noticed I think 40k leaks (not 40 kb, 40 thousand occurrences..) on startup. One of the traces for leaked memory was the DB open call.

jonasschnelli:
utACK 03bc719a85cb4928cb4b43d0bc4142f72cb01b23

PR #11019, opened 2017-08-10
[wallet] Abandon transactions that fail to go into the mempool

kallewoof:
This PR addresses the case where a wallet transaction fails to be accepted to the mempool when committed. This would keep the transaction in the wallet, to be rebroadcast at some undetermined time in the future, with the consequence that the change output would be cloaked. Even adding the mempool transactions into a new block will not immediately resolve the problem, due to the wallet not rebroadcasting the transaction immediately (even to itself) for (I think) privacy reasons.

Although there is a `walletrejectlongchain` option, the behavior when this is turned off is confusing and unpredictable, as noted above, and the `CommitTransaction` code will do the exact same thing for other mempool add failures, whichever those may be.

Edit: ~~~open question and currently investigating: will the coin selection algorithm inadvertently pick UTXOs that would result in a rejection of the transaction even if the user has other UTXOs available that wouldn't result in a too long chain? I have tried to trigger this case but the coin select is always picking the "right" UTXO that won't result in failure.~~~

Edit 2: In https://github.com/bitcoin/bitcoin/blame/master/src/wallet/wallet.cpp#L2457 the very last attempt to select coins increases the ancestor cap to num max, so above concern about coin select sometimes inadvertently picking an unspendable-due-to-chain-length tx is not a problem.

### DIFF
--- a/src/wallet/feebumper.cpp
+++ b/src/wallet/feebumper.cpp
@@ -266,7 +266,6 @@ Result CommitTransaction(CWallet* wallet, const uint256& txid, CMutableTransacti
     wtxBumped.fFromMe = true;
     CValidationState state;
     if (!wallet->CommitTransaction(wtxBumped, reservekey, g_connman.get(), state)) {
-        // NOTE: CommitTransaction never returns false, so this should never happen.
         errors.push_back(strprintf("The transaction was rejected: %s", state.GetRejectReason()));
         return Result::WALLET_ERROR;
     }
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -3036,7 +3036,8 @@ bool CWallet::CommitTransaction(CWalletTx& wtxNew, CReserveKey& reservekey, CCon
             // Broadcast
             if (!wtx.AcceptToMemoryPool(maxTxFee, state)) {
                 LogPrintf("CommitTransaction(): Transaction cannot be broadcast immediately, %s\n", state.GetRejectReason());
-                // TODO: if we expect the failure to be long term or permanent, instead delete wtx from the wallet and return failure.
+                AbandonTransaction(wtxNew.GetHash());
+                return false;
             } else {
                 wtx.RelayWalletTransaction(connman);
             }
--- a/test/functional/test_runner.py
+++ b/test/functional/test_runner.py
@@ -128,6 +128,7 @@ BASE_SCRIPTS= [
     'uacomment.py',
     'p2p-acceptblock.py',
     'feature_logging.py',
+    'wallet-longchain.py',
 ]
 
 EXTENDED_SCRIPTS = [
--- a/test/functional/wallet-longchain.py
+++ b/test/functional/wallet-longchain.py
@@ -1,0 +1,67 @@
+#!/usr/bin/env python3
+# Copyright (c) 2017 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Test long chained transactions in the wallet code."""
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import (connect_nodes_bi,
+                                 assert_equal,
+                                )
+
+class WalletTestLongchain(BitcoinTestFramework):
+
+    def set_test_params(self):
+        self.setup_clean_chain = True
+        self.num_nodes = 4
+
+    def setup_network(self):
+        self.add_nodes(3)
+        self.start_node(0)
+        self.start_node(1)
+        self.start_node(2)
+        connect_nodes_bi(self.nodes,0,1)
+        connect_nodes_bi(self.nodes,1,2)
+        connect_nodes_bi(self.nodes,0,2)
+        self.sync_all([self.nodes[0:3]])
+
+    def run_test(self):
+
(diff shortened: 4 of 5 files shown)


gmaxwell:
> immediately resolve the problem, due to the wallet not rebroadcasting the transaction immediately (even to itself) for (I think) privacy reasons.

It resolves it shortly.  Why do you think immediate is required?

You change makes it impossible to have the wallet simply queue transactions in long chains for sending when it becomes possible.

kallewoof:
Shortly is several minutes, which isn't really acceptable in some cases.

The code does indeed prevent you from adding one additional UTXO to an existing chain, but that's all you get from the current system. You don't get those UTXOs for further chaining as they are cloaked, as noted above. You would get the exact same results by adding 1 to the max chain length.

NicolasDorier:
@kallewoof trying to reword, correct me if wrong.

Currently situation:
Imagine that the unconf chain limit is 3 transactions and you have 1 UTXO

```
sendtoaddress abc 10 #TX1 OK
sendtoaddress abc 10 #TX2 OK
sendtoaddress abc 10 #TX3 OK

sendtoaddress abc 10 #TX4 BOOM chain-too-long
```

TX4 is still added to the wallet and will be rebroadcasted later.
Note that now, `getbalance` will returns 0, whatever the amount of the initial UTXO.

```
sendtoaddress abc 10 #TX5 BOOM insufficient funds
```
TX5 will not be rebroadcasted later.

With @kallewoof PR we have:

```
sendtoaddress abc 10 #TX1 OK
sendtoaddress abc 10 #TX2 OK
sendtoaddress abc 10 #TX3 OK

sendtoaddress abc 10 #TX4 BOOM chain-too-long
``` 

With TX4 not rebroadcasted by the wallet later, and `getbalance` does not returns 0.

I think it makes lots of sense. The current situation only save into the wallet one transaction on top of the longest chain, and the price is that the user see his balance to 0 when he still have funds. The current situation is also counter intuitive: why TX4 get queue by the wallet for later broadcast, but not TX5 ?

kallewoof:
Yes, the core issue is that a transaction that did not make it into the mempool but stayed in the wallet is invisible to the wallet code until it is put into the mempool at some undetermined later time.

Perhaps the better approach here is to make an additional coin view on top of the mempool called wallet view or something, which *would* include these to-be-mempooled transactions.

This would allow for arbitrary length tx chains, presuming the wallet code for rebroadcasting is able to deal with ordering, and presuming users will not be thrown off by transactions taking a lot longer to get into the blockchain (100 txs would at minimum take ~5 blocks, as only 20 or so txs in one chain would be accepted at a time).

kallewoof:
I am also wondering how very-long-chaining will affect existing UTXOs. Presuming the coin select picks indiscriminately (which does NOT seem to be the case, but I can't pinpoint why), having a long chain and sending more would "lock up" more and more UTXOs into that chain, locking more and more of your funds into an ever growing chain of transactions that have to go into the blockchain.

The more I think about it, the more I lean towards not allowing chains that are longer than what would go into the mempool, period. I.e. what this PR does.

sdaftuar:
> change makes it impossible to have the wallet simply queue transactions in long chains for sending when it becomes possible.

I agree with @gmaxwell; I believe there are use cases which prompted supporting the current behavior of allowing transactions to be queued up (for instance, if you have lots and lots of utxos, and your transactions are not in a hurry, then it may be fine to have some of them fail to get added to the mempool and become unavailable for further spending). 

I also think that in the future we may want to build more wallet features around the idea of, say, starting with a low fee (which may not make it into the mempool) and then auto-fee-bumping according to some schedule until it gets confirmed.  So I wouldn't want to preclude these kinds of strategies.

As an aside -- with this patch, don't we risk clogging up a user's wallet if they are trying to send in a loop and bump into this behavior?  Like if they get an error and just try again, hoping to get different inputs -- but if they don't have enough utxo's they get the same result, a transaction eating up space in their wallet which has been abandoned.  That seems like a bad outcome as well....  I'm not sure if a problem like this is unique to this approach but I'm generally wary of promoting more use of `AbandonTransaction()`; in general I feel like that should be for emergencies only.  I think to implement something like this behavior it would make more sense to add a way to remove a transaction from the wallet.

kallewoof:
@sdaftuar 
> I believe there are use cases which prompted supporting the current behavior of allowing transactions to be queued up (for instance, if you have lots and lots of utxos, and your transactions are not in a hurry, then it may be fine to have some of them fail to get added to the mempool and become unavailable for further spending).

I'm (still) confused about this statement, as the current behavior will only let you queue up one extra transaction, as if you increased mempool chain limit by 1. Beyond that, the UTXOs are hidden from the user when selecting coins, and the user will simply not be able to add to the chain anymore.

> I also think that in the future we may want to build more wallet features around the idea of, say, starting with a low fee (which may not make it into the mempool) and then auto-fee-bumping according to some schedule until it gets confirmed. So I wouldn't want to preclude these kinds of strategies.

If it's so low it doesn't make it into the mempool, what is the purpose of the transaction existing at all? It will never make it into the mempool as is, so you could just as well abandon it in favor of the higher-fee variant that comes along later. Again, I may be missing some subtlety here.

kallewoof:
No idea why `preciousblock.py` is failing on `setup_nodes()`. Definitely not related to this PR.

PR #11020, opened 2017-08-10
[wallet] getbalance: Add option to include non-mempool UTXOs

kallewoof:
Currently, if the wallet generates a transaction that cannot currently go into the mempool (e.g. due to too-long-mempool-chain), the wallet UTXOs related to this will vanish from `getbalance`, giving the appearance that the user has less funds than they actually do.

From a "get spendable balance" perspective, this is perfectly valid, but users may sometimes want to see their actual balance, regardless of whether they can spend it or not at that time.

This PR adds an `include_unspendable` option (default=false) to `getbalance` which, when `true`, will consider non-mempool transactions as trusted, and thus display these in the tally.

Note: this flag does nothing when a user specifies an account (i.e. `GetLegacyBalance`), and the legacy balance in fact already does what `include_unspendable=true` does, given the right arguments.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -832,41 +832,63 @@ static UniValue getbalance(const JSONRPCRequest& request)
         return NullUniValue;
     }
 
-    if (request.fHelp || (request.params.size() > 3 && IsDeprecatedRPCEnabled("accounts")) || (request.params.size() != 0 && !IsDeprecatedRPCEnabled("accounts")))
-        throw std::runtime_error(
-            "getbalance ( \"account\" minconf include_watchonly )\n"
-            "\nIf account is not specified, returns the server's total available balance.\n"
-            "The available balance is what the wallet considers currently spendable, and is\n"
-            "thus affected by options which limit spendability such as -spendzeroconfchange.\n"
-            "If account is specified (DEPRECATED), returns the balance in the account.\n"
-            "Note that the account \"\" is not the same as leaving the parameter out.\n"
-            "The server total may be different to the balance in the default \"\" account.\n"
-            "\nArguments:\n"
-            "1. \"account\"         (string, optional) DEPRECATED. This argument will be removed in V0.18. \n"
-            "                     To use this deprecated argument, start bitcoind with -deprecatedrpc=accounts. The account string may be given as a\n"
-            "                     specific account name to find the balance associated with wallet keys in\n"
-            "                     a named account, or as the empty string (\"\") to find the balance\n"
-            "                     associated with wallet keys not in any named account, or as \"*\" to find\n"
-            "                     the balance associated with all wallet keys regardless of account.\n"
-            "                     When this option is specified, it calculates the balance in a different\n"
-            "                     way than when it is not specified, and which can count spends twice when\n"
-            "                     there are conflicting pending transactions (such as those created by\n"
-            "                     the bumpfee command), temporarily resulting in low or even negative\n"
-            "                     balances. In general, account balance calculation is not considered\n"
-            "                     reliable and has resulted in confusing outcomes, so it is recommended to\n"
-            "                     avoid passing this argument.\n"
-            "2. minconf           (numeric, optional, default=1) DEPRECATED. Only valid when an account is specified. This argument will be removed in V0.18. To use this deprecated argument, start bitcoind with -deprecatedrpc=accounts. Only include transactions confirmed at least this many times.\n"
-            "3. include_watchonly (bool, optional, default=false) DEPRECATED. Only valid when an account is specified. This argument will be removed in V0.18. To use this deprecated argument, start bitcoind with -deprecatedrpc=accounts. Also include balance in watch-only addresses (see 'importaddress')\n"
-            "\nResult:\n"
-            "amount              (numeric) The total amount in " + CURRENCY_UNIT + " received for this account.\n"
-            "\nExamples:\n"
-            "\nThe total amount in the wallet with 1 or more confirmations\n"
-            + HelpExampleCli("getbalance", "") +
-            "\nThe total amount in the wallet at least 6 blocks confirmed\n"
-            + HelpExampleCli("getbalance", "\"*\" 6") +
-            "\nAs a json rpc call\n"
-            + HelpExampleRpc("getbalance", "\"*\", 6")
-        );
+    if (request.fHelp || (request.params.size() > 3 && IsDeprecatedRPCEnabled("accounts")) || (request.params.size() > 1 && !IsDeprecatedRPCEnabled("accounts"))) {
+        if (IsDeprecatedRPCEnabled("accounts")) {
+            throw std::runtime_error(
+                "getbalance ( \"account\" minconf include_watchonly )\n"
+                "\nIf account is not specified, returns the server's total available balance.\n"
+                "The available balance is what the wallet considers currently spendable, and is\n"
+                "thus affected by options which limit spendability such as -spendzeroconfchange.\n"
+                "If account is specified (DEPRECATED), returns the balance in the account.\n"
+                "Note that the account \"\" is not the same as leaving the parameter out.\n"
+                "The server total may be different to the balance in the default \"\" account.\n"
+                "\nArguments:\n"
+                "1. \"account\"         (string, optional) DEPRECATED. This argument will be removed in V0.18. \n"
+                "                     To use this deprecated argument, start bitcoind with -deprecatedrpc=accounts. The account string may be given as a\n"
+                "                     specific account name to find the balance associated with wallet keys in\n"
+                "                     a named account, or as the empty string (\"\") to find the balance\n"
+                "                     associated with wallet keys not in any named account, or as \"*\" to find\n"
+                "                     the balance associated with all wallet keys regardless of account.\n"
+                "                     When this option is specified, it calculates the balance in a different\n"
+                "                     way than when it is not specified, and which can count spends twice when\n"
+                "                     there are conflicting pending transactions (such as those created by\n"
+                "                     the bumpfee command), temporarily resulting in low or even negative\n"
+                "                     balances. In general, account balance calculation is not considered\n"
+                "                     reliable and has resulted in confusing outcomes, so it is recommended to\n"
+                "                     avoid passing this argument.\n"
+                "2. minconf           (numeric, optional, default=1) DEPRECATED. Only valid when an account is specified. This argument will be removed in V0.18. To use this deprecated argument, start bitcoind with -deprecatedrpc=accounts. Only include transactions confirmed at least this many times.\n"
+                "3. include_watchonly (bool, optional, default=false) DEPRECATED. Only valid when an account is specified. This argument will be removed in V0.18. To use this deprecated argument, start bitcoind with -deprecatedrpc=accounts. Also include balance in watch-only addresses (see 'importaddress')\n"
+                "\nResult:\n"
+                "amount              (numeric) The total amount in " + CURRENCY_UNIT + " received for this account.\n"
+                "\nExamples:\n"
+                "\nThe total amount in the wallet with 1 or more confirmations\n"
+                + HelpExampleCli("getbalance", "") +
+                "\nThe total amount in the wallet at least 6 blocks confirmed\n"
+                + HelpExampleCli("getbalance", "\"*\" 6") +
+                "\nAs a json rpc call\n"
+                + HelpExampleRpc("getbalance", "\"*\", 6")
+            );
+        } else {
+            throw std::runtime_error(
+                "getbalance ( include_unavailable )\n"
+                "\nReturns the wallet's total available balance.\n"
+                "The available balance is what the wallet considers currently spendable, and is\n"
+                "thus affected by options which limit spendability such as -spendzeroconfchange.\n"
+                "\nArguments:\n"
+                "1. include_unavailable (bool, optional, default=false) Also include balance for UTXOs which are currently not (but possibly in the future)\n"
+                "                                                       spendable.\n"
+                "\nResult:\n"
+                "amount              (numeric) The total amount in " + CURRENCY_UNIT + " received.\n"
+                "\nExamples:\n"
+                "\nThe total amount in the wallet\n"
+                + HelpExampleCli("getbalance", "") +
+                "\nThe total amount in the wallet, including currently unspendable UTXO:s\n"
+                + HelpExampleCli("getbalance", "true") +
+                "\nAs a json rpc call\n"
+                + HelpExampleRpc("getbalance", "true")
+            );
+        }
+    }
 
     // Make sure the results are valid at least up to the most recent block
     // the user could have gotten from another RPC command prior to now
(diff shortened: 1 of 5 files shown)


kallewoof:
Not sure why `zapwallettxes.py` is failing. It's not touching anything I changed (and the test passes on my fork's travis & when running the test locally / via `make check` ...). @jnewbery thoughts?

MarcoFalke:
The zapwallettxes' failure is intermittent. Respinned travis.

jnewbery:
I'd hoped that #10330 had fixed the zapwallettxes intermittency. @MarcoFalke do you have an example of it failing (the failure on this PR is no longer available since the task was restarted in Travis).

(apologies for being off-topic)

MarcoFalke:
@jnewbery Sorry I didn't keep track of the exact failure this time, I assume it was one of the known issues that popped up recently.

src/wallet/wallet.h
```diff
@@ -940,7 +940,7 @@ class CWallet : public CCryptoKeyStore, public CValidationInterface
     void ResendWalletTransactions(int64_t nBestBlockTime, CConnman* connman) override;
     // ResendWalletTransactionsBefore may only be called if fBroadcastTransactions!
     std::vector<uint256> ResendWalletTransactionsBefore(int64_t nTime, CConnman* connman);
-    CAmount GetBalance() const;
+    CAmount GetBalance(bool ignoreUnspendable = true) const;
```
promag: Nit, snake case.

src/wallet/wallet.h
```diff
@@ -465,7 +465,7 @@ class CWalletTx : public CMerkleTx
     bool IsEquivalentTo(const CWalletTx& tx) const;
 
     bool InMempool() const;
-    bool IsTrusted() const;
+    bool IsTrusted(bool ignoreUnspendable = true) const;
```
promag: Nit, snake case.

test/functional/wallet.py
```diff
@@ -392,6 +392,8 @@ def run_test(self):
         extra_txid = self.nodes[0].sendtoaddress(sending_addr, Decimal('0.0001'))
         assert(extra_txid not in self.nodes[0].getrawmempool())
         assert(extra_txid in [tx["txid"] for tx in self.nodes[0].listtransactions()])
+        # The change output for extra_txid should be in the balance if we include unspendable
+        assert(self.nodes[0].getbalance() < self.nodes[0].getbalance(include_unspendable=True))
```
promag: Assert exact value?

src/wallet/wallet.cpp
```diff
@@ -1823,7 +1823,7 @@ bool CWalletTx::InMempool() const
     return mempool.exists(GetHash());
 }
 
-bool CWalletTx::IsTrusted() const
+bool CWalletTx::IsTrusted(bool ignoreUnspendable) const
```
promag: Nit, snake case.

src/wallet/wallet.cpp
```diff
@@ -1837,7 +1837,7 @@ bool CWalletTx::IsTrusted() const
         return false;
 
     // Don't trust unconfirmed transactions from us unless they are in the mempool.
```
promag: Update comment.

promag (commented):
Should add the same option to `getunconfirmedbalance`?

src/wallet/rpcwallet.cpp
```diff
@@ -755,6 +755,8 @@ UniValue getbalance(const JSONRPCRequest& request)
             "                     avoid passing this argument.\n"
             "2. minconf           (numeric, optional, default=1) Only include transactions confirmed at least this many times.\n"
             "3. include_watchonly (bool, optional, default=false) Also include balance in watch-only addresses (see 'importaddress')\n"
+            "4. include_unspendable (bool, optional, default=false) Also include balance for UTXOs which are currently not (but\n"
```
jnewbery: I think this argument name is a bit confusing. Perhaps `include_not_yet_spendable`. It's a bit of a mouthful, but 'unspendable' sounds too permanent to me.

src/wallet/rpcwallet.cpp
```diff
@@ -768,8 +770,10 @@ UniValue getbalance(const JSONRPCRequest& request)
 
     LOCK2(cs_main, pwallet->cs_wallet);
 
-    if (request.params.size() == 0)
-        return  ValueFromAmount(pwallet->GetBalance());
+    bool ignoreUnspendable = request.params[3].isNull() || !request.params[3].get_bool();
+    if ((request.params[0].isNull() || (request.params[0].get_str() == "" && !ignoreUnspendable)) && request.params[1].isNull() && request.params[2].isNull()) {
```
jnewbery: This is confusing for me. `include_unspendable` only has effect if:

- `account` is not specified; and
- `minconf` is not specified; and
- `include_watchonly` is not specified

Additionally, if `account` is specified as `""` (ie balance not associated with any account), the returned balance will use `GetBalance()` if `include_unspendable` is not set, but `GetLegacyBalance()` if it is set.

Can you explain the reasoning here?

src/wallet/rpcwallet.cpp
```diff
@@ -768,8 +770,10 @@ UniValue getbalance(const JSONRPCRequest& request)
 
     LOCK2(cs_main, pwallet->cs_wallet);
 
-    if (request.params.size() == 0)
-        return  ValueFromAmount(pwallet->GetBalance());
+    bool ignoreUnspendable = request.params[3].isNull() || !request.params[3].get_bool();
```
jnewbery: I find it confusing that the API uses `include_unspendable`, but internally you use `ignoreUnspendable` which is the inverse. Any reason not to be consistent?

test/functional/wallet.py
jnewbery: Asserting on exact values can make the test more brittle. I think it's fine to test strict lesser than here.

src/wallet/wallet.cpp
jnewbery: Again, I think the name `ignoreUnspendable` could be made clearer.

jnewbery (commented):
Concept ACK. A few questions about the interface and naming inline.

src/wallet/wallet.h
kallewoof: Not sure what you mean -- most other argument names are written in this style, e.g. redeemScript, mapKeyBirth, etc.

src/wallet/rpcwallet.cpp
kallewoof: I was hoping the description of the argument would alleviate that, but you may be right.

src/wallet/rpcwallet.cpp
kallewoof: It ended up that way. You're right, will fix.

src/wallet/rpcwallet.cpp
kallewoof: The argument is the last in the list, and there's no way to give a default value for `account` because it's not being JSON parsed (you can't pass "null" as it would become the string "null"). Using named arguments, you can do this, but otherwise I have to make the "special case" of you passing defaults in.

The legacy balance seems to not be very used and possibly even about to be deprecated, so I didn't feel the need to update it. That, and it already kind of does what the PR is doing when you do a 0 `minconf` and `account="*"`.

test/functional/wallet.py
promag: > Asserting on exact values can make the test more brittle.

Is that a bad thing?

kallewoof:
@promag @jnewbery Thanks for the review!

* https://github.com/bitcoin/bitcoin/pull/11020/commits/28e8271424b8642e825889bfaa8e51ae6a2a60dd fixes comment
* https://github.com/bitcoin/bitcoin/pull/11020/commits/df7228d5dec184c92f7190265afb2c380da0a9ae fixes the inconsistency in variable names, but currently retains the `include_unspendable` name. I'm not sure I agree with the longer version @jnewbery suggested, but I'm up for changing this to something more intuitive if people feel it's worthwhile.
* https://github.com/bitcoin/bitcoin/pull/11020/commits/80acc02b0e77791605eb8f0cb20263986ea87399 adds the feature to `getunconfirmedbalance` (tested in https://github.com/bitcoin/bitcoin/pull/11020/commits/36423983a1b1a473151056bb5d4e3159975c9b18). I was not entirely sure about the logic on this one, but I went with `!IsTrusted() & depth=0 & (includeUnspendable || InMempool())`, changed from `!IsTrusted() & depth=0 & InMempool()`. This means some balance will overlap between `getbalance(include_unspendable=true)` and `getunconfirmedbalance(include_unspendable=true)` but I think that's ok.

test/functional/wallet.py
kallewoof: It's bad if the test will intermittently fail for unrelated reasons (rounding errors, for example).

src/wallet/rpcwallet.cpp
jnewbery: I find it surprising that `minconf` and `watchonly` only have effect if account is specified, and now `include_unspendable` only has effect if account is not specified. There doesn't appear to be any user documentation explaining that.

I'm afraid I don't know the history or motivation for that. @ryanofsky made changes in this area recently. Perhaps he can shed light on this.

kallewoof:
zapwallettxes.py again, @jnewbery: https://pastebin.com/70bQ7CLq

test/functional/wallet.py
promag: That should not happen. There are lots of exact asserts through out this file.

src/wallet/rpcwallet.cpp
promag: Also, please align other arguments as this is currently the longest.

promag:
IMO from the wallet point of view those UTXO should always be part of the balance, and in that case this should be a bug fix and not require more options. WDYT?

kallewoof:
Yes `getbalance` is available balance, I think, so I don't think it should default to including these. Of course, there is the option of making the coin select include these coins even though they're not in the mempool, in which case `getbalance` could include them, but that's for another time (and PR).

Hm, `include_unavailable` seems like a better name for it than `include_unspendable`.

src/wallet/rpcwallet.cpp
kallewoof: Actually, `include_watchonly` is slightly longer, but I think realigning is a good idea yeah.

kallewoof:
Rebased this, but now it is not available unless the user enables deprecated account. I could

1. Add a new "index 0" parameter to the non-deprecated variant, replacing `"account"`, and allowing up to 1 argument for the new style. This obviously could be potentially confusing.
2. Keep the index and allow up to 4 indices in new style, but require that items at index 0..2 are all null. Fine for `-named` style but ughy otherwise.

Ping @jnewbery.

jnewbery:
I think I prefer option (1). Overloading parameters can be confusing, but I think it's better in this case because:

- the meaning of argument 0 is not ambiguous - it's fixed by setting the `deprecaterrpc` argument.
- `include_unavailable` is a bool and `account` is a string
- the `account`, `minconf` and `include_watchonly` parameters will be removed in the next release.
- going with option (2) will lock us in to having 3 dummy arguments for future releases.

kallewoof:
@jnewbery See b93fbc4 -- I split up the descriptions based on deprecation, and fixed several text issues (e.g. RPC examples using the deprecated stuff). I think this change alone is worth a PR, to be honest.

FWIW I also removed `include_unavailable` from the deprecated variant since we're removing it anyway.

src/wallet/rpcwallet.cpp
```diff
@@ -832,48 +832,74 @@ static UniValue getbalance(const JSONRPCRequest& request)
...
-            "\nThe total amount in the wallet at least 6 blocks confirmed\n"
-            + HelpExampleCli("getbalance", "\"*\" 6") +
-            "\nAs a json rpc call\n"
-            + HelpExampleRpc("getbalance", "\"*\", 6")
-        );
+    if (request.fHelp || (request.params.size() > 3 && IsDeprecatedRPCEnabled("accounts")) || (request.params.size() > 1 && !IsDeprecatedRPCEnabled("accounts"))) {
+        if (IsDeprecatedRPCEnabled("accounts")) {
+            throw std::runtime_error(
+                "getbalance ( \"account\" minconf include_watchonly )\n"
+                "\nIf account is not specified, returns the server's total available balance.\n"
```
jnewbery: nit: since you're changing the help text for this and `getunconfirmedbalance`, perhaps correct it to say "the *wallet's* ..." instead of "the *server's* ..."

src/wallet/rpcwallet.cpp
```diff
@@ -832,48 +832,74 @@ static UniValue getbalance(const JSONRPCRequest& request)
...
 
     // Make sure the results are valid at least up to the most recent block
     // the user could have gotten from another RPC command prior to now
     pwallet->BlockUntilSyncedToCurrentChain();
 
     LOCK2(cs_main, pwallet->cs_wallet);
 
+    // params[0] is not converted from JSON until account stuff is removed, so we need to manually convert it here
+    // TODO: add param[0] to clients.cpp for getbalance and switch to get_bool() here, once account stuff is removed
+    const bool include_unavailable = !IsDeprecatedRPCEnabled("accounts") && !request.params[0].isNull() && (request.params[0].get_str() == "1" || request.params[0].get_str() == "true");
```
jnewbery: It's a bit ugly to replicate the univalue parsing here, but I think it's ok since it's only going to be in here for one release.

I think you should remove the `request.params[0].get_str() == "1"` (1 isn't interpreted as true in univalue). You may also throw an error if `request.params[0].get_str()` is not `false` or `true` (so passing `True` for example doesn't silently fail and leave `include_unavailable` as false.

src/wallet/wallet.cpp
```diff
@@ -2141,31 +2142,31 @@ void CWallet::ResendWalletTransactions(int64_t nBestBlockTime, CConnman* connman
...
+CAmount CWallet::GetUnconfirmedBalance(bool includeUnavailable) const
 {
     CAmount nTotal = 0;
     {
         LOCK2(cs_main, cs_wallet);
         for (const auto& entry : mapWallet)
         {
             const CWalletTx* pcoin = &entry.second;
-            if (!pcoin->IsTrusted() && pcoin->GetDepthInMainChain() == 0 && pcoin->InMempool())
+            if (!pcoin->IsTrusted() && pcoin->GetDepthInMainChain() == 0 && (includeUnavailable || pcoin->InMempool()))
```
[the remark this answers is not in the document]
kallewoof: I'm not sure. I'll add tests to check that this is not the case.

src/wallet/rpcwallet.cpp
```diff
@@ -832,48 +832,74 @@ static UniValue getbalance(const JSONRPCRequest& request)
...
+        }
+    }
 
     // Make sure the results are valid at least up to the most recent block
     // the user could have gotten from another RPC command prior to now
     pwallet->BlockUntilSyncedToCurrentChain();
 
     LOCK2(cs_main, pwallet->cs_wallet);
 
+    // params[0] is not converted from JSON until account stuff is removed, so we need to manually convert it here
```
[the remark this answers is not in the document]
kallewoof: Good point!

src/wallet/rpcwallet.cpp
kallewoof: I thought `1` was true. E.g. `getrawtransaction <id> 1`. But that seems to be a special case for `getrawtransaction`, actually. Changing as you suggest.

kallewoof:
@jnewbery 
> Should this call IsTrusted() with includeUnavailable? If not, coins that are in the mempool could show up in both the balance and the unconfirmed balance.

Looking at the tests (it was a while since I wrote this), I think the idea was that it should be counted in both (see wallet_basic.py changes). Is this problematic?

jnewbery:
Needs rebase.

@kallewoof - how urgently do you want this? If it can wait until after v0.17, then it'll probably be easier and a smaller diff once the accounts API code has been removed. I intend to do that as soon as 0.17 is branched.

kallewoof:
@jnewbery Not urgent. I will wait until 0.17 branches off.

kallewoof:
Happy birthday. Closing due to lack of interest.

jnewbery:
I commit to reviewing this after it's rebased on #13825

kallewoof:
@jnewbery OK!

kallewoof:
@jnewbery I began to rebase this, but self-reflectively decided this is a bit too invasive for the improvement it provides. I think it's a bit weird that 'getbalance' will suddenly drop significantly just because you ran out of chain length, but I'm no longer confident this is the right approach. Sorry for wasting your time!

jnewbery:
> Sorry for wasting your time!

Not wasted. No need to apologise!

PR #11022, opened 2017-08-10
Basic keypool topup

jnewbery:
This PR contains the first part of #10882 :

- if a key from the keypool is used, mark all keys up to that key as used, and then try to top up the keypool
- top up the keypool on startup

Notably, it does not stop the node or prevent the best block from advancing if the keypool drops below a threshold (which means that transactions may be missed and funds lost if restoring from an old HD wallet backup).

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -12,6 +12,7 @@
 #include "consensus/consensus.h"
 #include "consensus/validation.h"
 #include "fs.h"
+#include "init.h"
 #include "key.h"
 #include "keystore.h"
 #include "validation.h"
@@ -1053,6 +1054,30 @@ bool CWallet::AddToWalletIfInvolvingMe(const CTransactionRef& ptx, const CBlockI
         if (fExisted && !fUpdate) return false;
         if (fExisted || IsMine(tx) || IsFromMe(tx))
         {
+            /* Check if any keys in the wallet keypool that were supposed to be unused
+             * have appeared in a new transaction. If so, remove those keys from the keypool.
+             * This can happen when restoring an old wallet backup that does not contain
+             * the mostly recently created transactions from newer versions of the wallet.
+             */
+
+            // loop though all outputs
+            for (const CTxOut& txout: tx.vout) {
+                // extract addresses and check if they match with an unused keypool key
+                std::vector<CKeyID> vAffected;
+                CAffectedKeysVisitor(*this, vAffected).Process(txout.scriptPubKey);
+                for (const CKeyID &keyid : vAffected) {
+                    std::map<CKeyID, int64_t>::const_iterator mi = m_pool_key_to_index.find(keyid);
+                    if (mi != m_pool_key_to_index.end()) {
+                        LogPrintf("%s: Detected a used keypool key, mark all keypool key up to this key as used\n", __func__);
+                        MarkReserveKeysAsUsed(mi->second);
+
+                        if (!TopUpKeyPool()) {
+                            LogPrintf("%s: Topping up keypool failed (locked wallet)\n", __func__);
+                        }
+                    }
+                }
+            }
+
             CWalletTx wtx(this, ptx);
 
             // Get merkle branch if transaction was found in a block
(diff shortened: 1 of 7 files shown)


test/functional/test_framework/test_framework.py
```diff
@@ -209,7 +209,7 @@ def start_node(self, i, dirname, extra_args=None, rpchost=None, timewait=None, b
         datadir = os.path.join(dirname, "node" + str(i))
         if binary is None:
             binary = os.getenv("BITCOIND", "bitcoind")
-        args = [binary, "-datadir=" + datadir, "-server", "-keypool=1", "-discover=0", "-rest", "-logtimemicros", "-debug", "-debugexclude=libevent", "-debugexclude=leveldb", "-mocktime=" + str(self.mocktime), "-uacomment=testnode%d" % i]
+        args = [binary, "-datadir=" + datadir, "-server", "-keypool=1", "-keypoolcritical=0", "-discover=0", "-rest", "-logtimemicros", "-debug", "-debugexclude=libevent", "-debugexclude=leveldb", "-mocktime=" + str(self.mocktime), "-uacomment=testnode%d" % i]
```
ryanofsky: In commit "[wallet] keypool mark-used and topup"

It looks like these test changes got added to the wrong commit.

jnewbery: Thanks Russ. You're right, this was in the wrong commit and shouldn't be in this PR. Fixed

src/wallet/wallet.cpp
```diff
@@ -3611,6 +3636,28 @@ void CReserveKey::ReturnKey()
...
+    CWalletDB walletdb(*dbw);
+    while (it != std::end(*setKeyPool)) {
+        const int64_t& index = *(it);
+        if (index > keypool_id) break; // set*KeyPool is ordered
+
+        CKeyPool keypool;
+        if (walletdb.ReadPool(index, keypool)) { //TODO: This should be unnecessary
+            m_pool_key_to_index.erase(keypool.vchPubKey.GetID());
+        }
+        walletdb.ErasePool(index);
```
ryanofsky: In commit "[wallet] keypool mark-used and topup"

An older version of this PR was calling KeepKey instead of doing this manually. It'd be nice to either call KeepKey here, or add the "keypool keep" log print here from that method. Logging about the key being kept would help debugging and make this code clearer.

ryanofsky (commented):
Slightly tested ACK d34957e17e8c9740104533aaf4a896e93548c87e. Confirmed disabling topup breaks the new test. Would be nice to get this merged, because this makes updating unlocked wallets a lot safer, and because merging it would considerably simplify #10882.

gmaxwell (approved):
utACK

sipa:
utACK d34957e17e8c9740104533aaf4a896e93548c87e

instagibbs (approved):
utACK

jnewbery:
Thanks for the code review @ryanofsky . This branch now has some ACKs so I'm going to hold off making any additional changes (unless others think them necessary)

I'll happily review and ACK a cleanup PR after v0.15

achow101:
utACK d34957e17e8c9740104533aaf4a896e93548c87e

jnewbery:
:tada:

paveljanik:
This brings two new warnings:

```
wallet/wallet.cpp:3668:38: warning: comparison of integers of different signs: 'size_type' (aka 'unsigned long') and 'int' [-Wsign-compare]
    return setExternalKeyPool.size() >= min_keys && (setInternalKeyPool.size() >= min_keys || !CanSupportFeature(FEATURE_HD_SPLIT));
           ~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~
wallet/wallet.cpp:3668:80: warning: comparison of integers of different signs: 'size_type' (aka 'unsigned long') and 'int' [-Wsign-compare]
    return setExternalKeyPool.size() >= min_keys && (setInternalKeyPool.size() >= min_keys || !CanSupportFeature(FEATURE_HD_SPLIT));
                                                     ~~~~~~~~~~~~~~~~~~~~~~~~~ ^  ~~~~~~~~
2 warnings generated.
```

jnewbery:
Thanks @paveljanik . This was pointed out by @ryanofsky before but I didn't fix it because this PR had already collected a few ACKs

 Cleanup commits here: #11044 . Not required for v0.15.

PR #11023, opened 2017-08-10
[tests] Add option to attach a python debugger if functional test fails

jnewbery:
Adds a simple option to the test_framework to attach pdb if the test fails.

Helpful for catching and debugging intermittent failures: Run the test in a loop with this option. The first failure will cause execution to pause and nodes will be left running for interactive debugging.

@sdaftuar

### DIFF
--- a/test/functional/test_framework/test_framework.py
+++ b/test/functional/test_framework/test_framework.py
@@ -11,6 +11,7 @@ import http.client
 import logging
 import optparse
 import os
+import pdb
 import shutil
 import subprocess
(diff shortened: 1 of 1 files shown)


laanwj:
This could certainly be useful to inspect state after a test failed, utACK.

sdaftuar:
I'm using this right now to debug a test.  ACK.

MarcoFalke:
utACK cc5d38f4b826b897dd818bb22c7f5d881718fd35

PR #11025, opened 2017-08-10
qa: Fix inv race in example_test

maflcko:
There have been intermittent test failures on this script.

```py
  File "./test/functional/example_test.py", line 216, in run_test
    assert_equal(block, 1)
AssertionError: not(2 == 1)
```

Probably the simplest way to fix them is overriding the `on_inv` method, so that no "colliding" getdata for the blocks are sent out.

Additionally, all getdata are now sent in a single message.

### DIFF
--- a/test/functional/example_test.py
+++ b/test/functional/example_test.py
@@ -58,6 +58,10 @@ class BaseNode(NodeConnCB):
         message.block.calc_sha256()
         self.block_receive_map[message.block.sha256] += 1
 
+    def on_inv(self, conn, message):
+        """Override the standard on_inv callback"""
+        pass
+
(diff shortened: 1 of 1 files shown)


jnewbery:
Seems reasonable. utACK faa76d1b79cc63177c8e165c9a2dc761aac7ccf4

Thanks!

sdaftuar:
utACK

Perhaps we should get rid of the default `on_inv` behavior altogether in `NodeConnCB`?  I don't think there's a good reason for its existence.

MarcoFalke:
Good suggestion. Though, I am going to leave it for later, as it requires a bit more review than this small fix.

PR #11026, opened 2017-08-10
Bugfix: Use testnet RequireStandard for -acceptnonstdtxn default

luke-jr:
Fixes a bug introduced in #8855

`-acceptnonstdtxn` is a valid option only for testnet/regtest (in Core), and the help message reflects that. Currently, however, it is buggy in two ways:

1. It uses mainnet to get the default value, which doesn't make sense since the option is never available for mainnet, and the only time the option is available, is when the default is the opposite.
2. It uses the value of "require standard" directly as the default for "accept non-standard transactions", but these concepts are opposites: a negation must be performed to transform one to the other.

Note the combination of these bugs results in the correct boolean output, but the logic to get there is completely wrong.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -479,7 +479,7 @@ std::string HelpMessage(HelpMessageMode mode)
 
     strUsage += HelpMessageGroup(_("Node relay options:"));
     if (showDebug) {
-        strUsage += HelpMessageOpt("-acceptnonstdtxn", strprintf("Relay and mine \"non-standard\" transactions (%sdefault: %u)", "testnet/regtest only; ", defaultChainParams->RequireStandard()));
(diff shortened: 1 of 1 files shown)


sdaftuar:
utACK

sipa:
Care to elaborate a bit?

luke-jr:
`-acceptnonstdtxn` is a valid option only for testnet/regtest (in Core), and the help message reflects that. Currently, however, it is buggy in two ways:

1. It uses mainnet to get the default value, which doesn't make sense since the option is never available for mainnet, and the only time the option is available, is when the default is the opposite.
2. It uses the value of "require standard" directly as the default for "accept non-standard transactions", but these concepts are opposites: a negation must be performed to transform one to the other.

Note the combination of these bugs results in the correct boolean output, but the logic to get there is completely wrong.

achow101:
utACK 4aa25089b4392a9a17ba5b3a8ea22e893d11d876

MeshCollider:
utACK https://github.com/bitcoin/bitcoin/pull/11026/commits/4aa25089b4392a9a17ba5b3a8ea22e893d11d876

PR #11027, opened 2017-08-10
[RPC] Only return hex field once in getrawtransaction

achow101:
The hex is already returned in `TxToUniv()`, no need to give it out a second time in getrawtransaction itself.

### DIFF
--- a/src/core_io.h
+++ b/src/core_io.h
@@ -31,6 +31,6 @@ UniValue ValueFromAmount(const CAmount& amount);
 std::string FormatScript(const CScript& script);
 std::string EncodeHexTx(const CTransaction& tx, const int serializeFlags = 0);
 void ScriptPubKeyToUniv(const CScript& scriptPubKey, UniValue& out, bool fIncludeHex);
-void TxToUniv(const CTransaction& tx, const uint256& hashBlock, UniValue& entry);
+void TxToUniv(const CTransaction& tx, const uint256& hashBlock, UniValue& entry, bool include_hex = true, int serialize_flags = 0);
 
 #endif // BITCOIN_CORE_IO_H
--- a/src/core_write.cpp
+++ b/src/core_write.cpp
@@ -153,7 +153,7 @@ void ScriptPubKeyToUniv(const CScript& scriptPubKey,
     out.pushKV("addresses", a);
 }
 
-void TxToUniv(const CTransaction& tx, const uint256& hashBlock, UniValue& entry)
+void TxToUniv(const CTransaction& tx, const uint256& hashBlock, UniValue& entry, bool include_hex, int serialize_flags)
 {
     entry.pushKV("txid", tx.GetHash().GetHex());
     entry.pushKV("hash", tx.GetWitnessHash().GetHex());
@@ -207,5 +207,7 @@ void TxToUniv(const CTransaction& tx, const uint256& hashBlock, UniValue& entry)
     if (!hashBlock.IsNull())
         entry.pushKV("blockhash", hashBlock.GetHex());
 
-    entry.pushKV("hex", EncodeHexTx(tx)); // the hex-encoded transaction. used the name "hex" to be consistent with the verbose output of "getrawtransaction".
+    if (include_hex) {
+        entry.pushKV("hex", EncodeHexTx(tx, serialize_flags)); // the hex-encoded transaction. used the name "hex" to be consistent with the verbose output of "getrawtransaction".
+    }
 }
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -125,7 +125,7 @@ UniValue blockToJSON(const CBlock& block, const CBlockIndex* blockindex, bool tx
         if(txDetails)
         {
             UniValue objTx(UniValue::VOBJ);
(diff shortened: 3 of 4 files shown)


sdaftuar:
ACK

achow101:
@MarcoFalke Hmm. That might be a problem since `RPCSerializationFlags` are not passed to `EncodeHexTx` in `TxToUniv`.

sipa:
> That might be a problem since `RPCSerializationFlags` are not passed to `EncodeHexTx` in `TxToUniv`.

That seems to be a bug on its own.

achow101:
Another thing is that not everything that uses `TxToUniv` needs to have hex output, e.g. `decoderawtransaction`

achow101:
I added a commit which passes the `RPCSerializationFlags` to `TxToUniv` and a boolean for whether to give `hex` output or not.

src/core_io.h
```diff
@@ -31,6 +31,6 @@ UniValue ValueFromAmount(const CAmount& amount);
 std::string FormatScript(const CScript& script);
 std::string EncodeHexTx(const CTransaction& tx, const int serializeFlags = 0);
 void ScriptPubKeyToUniv(const CScript& scriptPubKey, UniValue& out, bool fIncludeHex);
-void TxToUniv(const CTransaction& tx, const uint256& hashBlock, UniValue& entry);
+void TxToUniv(const CTransaction& tx, const uint256& hashBlock, UniValue& entry, const int serializeFlags = 0, bool include_hex = true);
```
sipa: Marking a by-value parameter as `const` is redundant. Also, style nit on `serializeFlags`.

sipa (commented):
utACK

src/core_io.h
achow101: Fixed. It was copy and pasted from two lines above in `EncodeHexTx` :)

MarcoFalke:
> That seems to be a bug on its own.

I don't think so?

According to the doc, `-rpcserialversion` sets the serialization of raw transaction or block hex returned in        **non-verbose** mode.

promag (commented):
utACK 9381d86.

There are only 2 more calls to `TxToUniv` so I wonder if it pays to have default arguments.

achow101:
@MarcoFalke I interpret that as talking about when verbose mode means that you wouldn't get the hex back, e.g. in `getblock`. If we are returning hex, then I think we should be consistent and all hex should follow whatever the `RPCSerializationFlags` are.

gmaxwell (approved):
ACK.  This needs a 0.15 tag: it's a 0.15 regression and it results in invalid JSON.

src/core_io.h
```diff
@@ -31,6 +31,6 @@ UniValue ValueFromAmount(const CAmount& amount);
 std::string FormatScript(const CScript& script);
 std::string EncodeHexTx(const CTransaction& tx, const int serializeFlags = 0);
 void ScriptPubKeyToUniv(const CScript& scriptPubKey, UniValue& out, bool fIncludeHex);
-void TxToUniv(const CTransaction& tx, const uint256& hashBlock, UniValue& entry);
```
laanwj: Don't care about the order, though I'd prefer not using default arguments. Easy enough to just add these on all call sites, there aren't many.

jnewbery:
Strictly speaking, I don't think this is invalid JSON, although client behaviour is undefined if object names are not unique (https://tools.ietf.org/html/rfc7159#section-4). But yes, this is a regression in 0.15 and should be treated as a bug fix.

Perhaps we should add JSON object checking to rpcserver.cpp (in `JSONRPCExecOne()`) and/or authproxy.py (in `__call__()`) to verify the validity of our returned JSON.

jnewbery:
Concept ACK

> Another thing is that not everything that uses TxToUniv needs to have hex output, e.g. decoderawtransaction
I added a commit which passes ... a boolean for whether to give hex output or not.

I don't think you should complicate the interface to `TxToUniv` to specify which fields to return. Either filter them out at the caller, or just return the hex in `decoderawtransaction`. Is there any harm in returning the hex in that RPC?

achow101:
Swapped the arguments.

>  Either filter them out at the caller

I looked at that but I couldn't find anything in Univalue for removing things from it.

> or just return the hex in `decoderawtransaction`. Is there any harm in returning the hex in that RPC?

Since you are already providing it the hex, it's not useful to be giving the hex back out to the user.

jtimon:
utACK e029c6e

> There are only 2 more calls to TxToUniv so I wonder if it pays to have default arguments.

+1

promag:
Wild option 🙄:
```
void TxToUniv(const CTransaction& tx, const uint256& hashBlock, UniValue& entry);
void TxToUnivWithHex(const CTransaction& tx, const uint256& hashBlock, int serialize_flags, UniValue& entry);
```

laanwj:
> Since you are already providing it the hex, it's not useful to be giving the hex back out to the user.

I was also really surprised to see the hex being echoed back in `decoderawtransaction` output. That's just a waste of bandwidth especially for large transactions.

>  Either filter them out at the caller

No, don't do this. univalue objects aren't meant to be written to/processed after initial generation. The library is not optimized for that. If you need any kind of editing use an intermedaite data structure.

PR #11032, opened 2017-08-11
[qa] Fix block message processing error in sendheaders.py

sdaftuar:
Introduced in #10169 

@jnewbery

### DIFF
--- a/test/functional/sendheaders.py
+++ b/test/functional/sendheaders.py
@@ -121,9 +121,6 @@ class TestNode(NodeConnCB):
             message.headers[-1].calc_sha256()
             self.last_blockhash_announced = message.headers[-1].sha256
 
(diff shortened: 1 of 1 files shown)


MarcoFalke:
utACK 38bb21bb409d4da4ffe6f81f9b060a75334d6fb6

jnewbery:
good catch!

The `last_message["block"]` variable isn't actually used at all in this test (and `last_block` wasn't used before my refactor which introduced this bug), so we can actually just remove the `on_block()` method override entirely.

sdaftuar:
@jnewbery Done.

promag:
utACK f1bf311.

fanquake:
utACK f1bf311

PR #11039, opened 2017-08-13
Avoid second mapWallet lookup

promag:
All calls to `mapWallet.count()` have the intent to detect if a `txid` exists and most are followed by a second lookup to retrieve the `CWalletTx`.

This PR replaces all `mapWallet.count()` calls with `mapWallet.find()` to avoid the second lookup.

### DIFF
--- a/src/wallet/feebumper.cpp
+++ b/src/wallet/feebumper.cpp
@@ -76,12 +76,12 @@ CFeeBumper::CFeeBumper(const CWallet *pWallet, const uint256 txidIn, const CCoin
     vErrors.clear();
     bumpedTxid.SetNull();
     AssertLockHeld(pWallet->cs_wallet);
-    if (!pWallet->mapWallet.count(txid)) {
+    auto it = pWallet->mapWallet.find(txid);
+    if (it == pWallet->mapWallet.end()) {
         vErrors.push_back("Invalid or non-wallet transaction id");
         currentResult = BumpFeeResult::INVALID_ADDRESS_OR_KEY;
         return;
     }
-    auto it = pWallet->mapWallet.find(txid);
     const CWalletTx& wtx = it->second;
 
     if (!preconditionChecks(pWallet, wtx)) {
@@ -241,12 +241,13 @@ bool CFeeBumper::commit(CWallet *pWallet)
     if (!vErrors.empty() || currentResult != BumpFeeResult::OK) {
         return false;
     }
-    if (txid.IsNull() || !pWallet->mapWallet.count(txid)) {
+    auto it = txid.IsNull() ? pWallet->mapWallet.end() : pWallet->mapWallet.find(txid);
+    if (it == pWallet->mapWallet.end()) {
         vErrors.push_back("Invalid or non-wallet transaction id");
         currentResult = BumpFeeResult::MISC_ERROR;
         return false;
     }
-    CWalletTx& oldWtx = pWallet->mapWallet[txid];
+    CWalletTx& oldWtx = it->second;
 
     // make sure the transaction still has no descendants and hasn't been mined in the meantime
     if (!preconditionChecks(pWallet, oldWtx)) {
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -1874,10 +1874,11 @@ UniValue listsinceblock(const JSONRPCRequest& request)
             throw JSONRPCError(RPC_INTERNAL_ERROR, "Can't read block from disk");
         }
         for (const CTransactionRef& tx : block.vtx) {
-            if (pwallet->mapWallet.count(tx->GetHash()) > 0) {
+            auto it = pwallet->mapWallet.find(tx->GetHash());
+            if (it != pwallet->mapWallet.end()) {
                 // We want all transactions regardless of confirmation count to appear here,
                 // even negative confirmation ones, hence the big negative.
-                ListTransactions(pwallet, pwallet->mapWallet[tx->GetHash()], "*", -100000000, true, removed, filter);
+                ListTransactions(pwallet, it->second, "*", -100000000, true, removed, filter);
             }
         }
         paltindex = paltindex->pprev;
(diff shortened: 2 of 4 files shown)


promag (commented):
Missing `mapWallet[]` cases.

jonasschnelli:
General Concept ACK. I guess extending to cover `mapWallet[]` would make sense?

promag:
Yes already on it.

src/wallet/feebumper.cpp
```diff
@@ -241,12 +241,13 @@ bool CFeeBumper::commit(CWallet *pWallet)
     if (!vErrors.empty() || currentResult != BumpFeeResult::OK) {
         return false;
     }
-    if (txid.IsNull() || !pWallet->mapWallet.count(txid)) {
+    auto it = pWallet->mapWallet.find(txid);
+    if (txid.IsNull() || it == pWallet->mapWallet.end()) {
```
ryanofsky: Maybe something for a followup commit, but I don't understand the txid IsNull check. If it's expected that there could be a null txid in mapWallet, there should be a comment explaining why that would happen, otherwise the check should be dropped or replaced with an assert or other error indicating that wallet state is corrupted.

src/wallet/rpcwallet.cpp
```diff
@@ -2011,7 +2013,7 @@ UniValue abandontransaction(const JSONRPCRequest& request)
     uint256 hash;
     hash.SetHex(request.params[0].get_str());
 
-    if (!pwallet->mapWallet.count(hash)) {
+    if (pwallet->mapWallet.find(hash) == pwallet->mapWallet.end()) {
```
ryanofsky: Seems pointlessly verbose.

src/wallet/wallet.cpp
```diff
@@ -1017,7 +1019,7 @@ bool CWallet::AddToWalletIfInvolvingMe(const CTransactionRef& ptx, const CBlockI
             }
         }
 
-        bool fExisted = mapWallet.count(tx.GetHash()) != 0;
+        bool fExisted = mapWallet.find(tx.GetHash()) != mapWallet.end();
```
ryanofsky: Also unnecessarily verbose.

ryanofsky (commented):
utACK 8e1b843103235d5ded84af5d25ee1530da448d53

> I guess extending to cover mapWallet[] would make sense?

Not sure specifically what is meant here, but #9680 replaces [] lookups with at() lookups to avoid inadvertently adding entries to the map when the txid doesn't exist.

src/wallet/rpcwallet.cpp
promag: Yeah, and there is no second lookup here so..

src/wallet/wallet.cpp
promag: Yeah, and there is no second lookup here so..

src/wallet/feebumper.cpp
promag: @ryanofsky I guess it's to prevent a unnecessary `count()` in the old code. What about:
```cpp
auto it = txid.IsNull() ? pWallet->mapWallet.end() : pWallet->mapWallet.find(txid);
if (it == pWallet->mapWallet.end()) {
```

promag:
@jonasschnelli leaving mapWallet[] cases out because of @ryanofsky suggestion above.

src/wallet/feebumper.cpp
ryanofsky: Thread https://github.com/bitcoin/bitcoin/pull/11039#discussion_r133029806

Good idea. I guess if the purpose of the IsNull call was to serve as a minor optimization, the intent is at still discernible, though it's also still unclear. And your new change would match previous behavior more closely, so that's good too.

I think ideally IsNull check would be dropped or actually explained, but that seems like something for a followup. Your new change would be good to add to this PR.

ryanofsky (commented):
utACK 8f2f1e0458d263dc9b51caad0adc0246e3580114. Only changes since previous are what was discussed in comments.

src/wallet/rpcwallet.cpp
```diff
@@ -1957,10 +1958,11 @@ UniValue gettransaction(const JSONRPCRequest& request)
             filter = filter | ISMINE_WATCH_ONLY;
 
     UniValue entry(UniValue::VOBJ);
-    if (!pwallet->mapWallet.count(hash)) {
+    auto it = pwallet->mapWallet.find(hash);
+    if (it == pwallet->mapWallet.end()) {
         throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid or non-wallet transaction id");
     }
-    const CWalletTx& wtx = pwallet->mapWallet[hash];
+    const CWalletTx& wtx = it->second;
```
luke-jr: IMO also use this in the other cases you're using `it->second`.

luke-jr (approved):
utACK, with nit

laanwj:
utACK 8f2f1e0

paveljanik:
This brought three new -Wshadow warnings:
```
  CXX      wallet/libbitcoin_wallet_a-wallet.o
wallet/wallet.cpp:1125:14: warning: declaration shadows a local variable [-Wshadow]
        auto it = mapWallet.find(now);
             ^
wallet/wallet.cpp:1112:10: note: previous declaration is here
    auto it = mapWallet.find(hashTx);
         ^
wallet/wallet.cpp:1152:22: warning: declaration shadows a local variable [-Wshadow]
                auto it = mapWallet.find(txin.prevout.hash);
                     ^
wallet/wallet.cpp:1125:14: note: previous declaration is here
        auto it = mapWallet.find(now);
             ^
wallet/wallet.cpp:1215:22: warning: declaration shadows a local variable [-Wshadow]
                auto it = mapWallet.find(txin.prevout.hash);
                     ^
wallet/wallet.cpp:1193:14: note: previous declaration is here
        auto it = mapWallet.find(now);
             ^
3 warnings generated.
```

PR #11042, opened 2017-08-13
[Trivial] update copyright years in comments

terryschmidt:


fanquake:
Thanks, however we have tools for doing this, see contrib/devtools/copyright_header.py, and will increment at the end of the year where required.

MarcoFalke:
Imo it is fine to update prior to every major release. This way it can be approximately inferred from the downloaded source code when the file was last changed.

PR #11043, opened 2017-08-14
Use std::unique_ptr (C++11) where possible

practicalswift:
Use `std::unique_ptr` (C++11) where possible.

Rationale:
1. Avoid resource leaks (specifically: forgetting to `delete` an object created using `new`)
2. Avoid undefined behaviour (specifically: double `delete`:s)

**Note to reviewers:** Please let me know if I've missed any obvious `std::unique_ptr` candidates. Hopefully this PR should cover all the trivial cases.

### DIFF
--- a/src/httprpc.cpp
+++ b/src/httprpc.cpp
@@ -238,7 +238,7 @@ bool StartHTTPRPC()
     RegisterHTTPHandler("/wallet/", false, HTTPReq_JSONRPC);
 #endif
     assert(EventBase());
-    httpRPCTimerInterface = std::unique_ptr<HTTPRPCTimerInterface>(new HTTPRPCTimerInterface(EventBase()));
+    httpRPCTimerInterface = MakeUnique<HTTPRPCTimerInterface>(EventBase());
     RPCSetTimerInterface(httpRPCTimerInterface.get());
     return true;
 }
--- a/src/httprpc.cpp
+++ b/src/httprpc.cpp
@@ -62,7 +62,7 @@ private:
 /* Pre-base64-encoded authentication token */
 static std::string strRPCUserColonPass;
 /* Stored RPC timer interface (for unregistration) */
-static HTTPRPCTimerInterface* httpRPCTimerInterface = nullptr;
+static std::unique_ptr<HTTPRPCTimerInterface> httpRPCTimerInterface;
 
 static void JSONErrorReply(HTTPRequest* req, const UniValue& objError, const UniValue& id)
 {
@@ -238,8 +238,8 @@ bool StartHTTPRPC()
     RegisterHTTPHandler("/wallet/", false, HTTPReq_JSONRPC);
 #endif
     assert(EventBase());
-    httpRPCTimerInterface = new HTTPRPCTimerInterface(EventBase());
-    RPCSetTimerInterface(httpRPCTimerInterface);
+    httpRPCTimerInterface = std::unique_ptr<HTTPRPCTimerInterface>(new HTTPRPCTimerInterface(EventBase()));
+    RPCSetTimerInterface(httpRPCTimerInterface.get());
     return true;
 }
 
@@ -253,8 +253,7 @@ void StopHTTPRPC()
     LogPrint(BCLog::RPC, "Stopping HTTP RPC server\n");
     UnregisterHTTPHandler("/", true);
     if (httpRPCTimerInterface) {
-        RPCUnsetTimerInterface(httpRPCTimerInterface);
-        delete httpRPCTimerInterface;
-        httpRPCTimerInterface = nullptr;
+        RPCUnsetTimerInterface(httpRPCTimerInterface.get());
+        httpRPCTimerInterface.reset();
     }
 }
--- a/src/policy/fees.cpp
+++ b/src/policy/fees.cpp
@@ -548,16 +548,13 @@ CBlockPolicyEstimator::CBlockPolicyEstimator()
     bucketMap[INF_FEERATE] = bucketIndex;
     assert(bucketMap.size() == buckets.size());
 
-    feeStats = new TxConfirmStats(buckets, bucketMap, MED_BLOCK_PERIODS, MED_DECAY, MED_SCALE);
-    shortStats = new TxConfirmStats(buckets, bucketMap, SHORT_BLOCK_PERIODS, SHORT_DECAY, SHORT_SCALE);
-    longStats = new TxConfirmStats(buckets, bucketMap, LONG_BLOCK_PERIODS, LONG_DECAY, LONG_SCALE);
+    feeStats = std::unique_ptr<TxConfirmStats>(new TxConfirmStats(buckets, bucketMap, MED_BLOCK_PERIODS, MED_DECAY, MED_SCALE));
+    shortStats = std::unique_ptr<TxConfirmStats>(new TxConfirmStats(buckets, bucketMap, SHORT_BLOCK_PERIODS, SHORT_DECAY, SHORT_SCALE));
+    longStats = std::unique_ptr<TxConfirmStats>(new TxConfirmStats(buckets, bucketMap, LONG_BLOCK_PERIODS, LONG_DECAY, LONG_SCALE));
 }
 
 CBlockPolicyEstimator::~CBlockPolicyEstimator()
 {
-    delete feeStats;
-    delete shortStats;
-    delete longStats;
 }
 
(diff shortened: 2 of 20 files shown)


src/policy/fees.cpp
```diff
@@ -997,12 +994,9 @@ bool CBlockPolicyEstimator::Read(CAutoFile& filein)
             }
 
             // Destroy old TxConfirmStats and point to new ones that already reference buckets and bucketMap
-            delete feeStats;
-            delete shortStats;
-            delete longStats;
-            feeStats = fileFeeStats.release();
-            shortStats = fileShortStats.release();
-            longStats = fileLongStats.release();
+            feeStats.reset(fileFeeStats.release());
```
sipa: How about

```c++
feeStats = std::move(fileFeeStats);
shortStats = std::move(fileShortStats);
longStats = std::move(fileLongStats);
```

practicalswift:
@sipa Nice catch! Fixed! :-)

ryanofsky (commented):
It would be nice to replace all the new

```std::unique_ptr<ClassName>(new ClassName(...))```

occurrences this is adding with simpler

```make_unique<ClassName>(...)```

calls.

[`std::make_unique`](http://en.cppreference.com/w/cpp/memory/unique_ptr/make_unique) isn't available until c++14, so we can't use it yet, but it is easy enough to write a substitute. For example (from #10973):

```cpp
//! Substitute for C++14 std::make_unique.
template <typename T, typename... Args>
std::unique_ptr<T> MakeUnique(Args&&... args)
{
    return std::unique_ptr<T>(new T(std::forward<Args>(args)...));
}
```

src/wallet/db.cpp
```diff
@@ -487,7 +487,7 @@ bool CDB::Rewrite(CWalletDBWrapper& dbw, const char* pszSkip)
                 std::string strFileRes = strFile + ".rewrite";
                 { // surround usage of db with extra {}
                     CDB db(dbw, "r");
-                    Db* pdbCopy = new Db(env->dbenv.get(), 0);
+                    std::unique_ptr<Db> pdbCopy = std::unique_ptr<Db>(new Db(env->dbenv.get(), 0));
```
ryanofsky: In commit "Use unique_ptr for pdbCopy (Db)"

Is this fixing a potential resource / memory leak? It seems like the pointer was only being freed conditionally below. Maybe update commit description to say how this is changing behavior, if it is changing behavior.

src/test/test_bitcoin.h
```diff
@@ -50,7 +50,7 @@ struct BasicTestingSetup {
  */
 class CConnman;
 struct TestingSetup: public BasicTestingSetup {
-    CCoinsViewDB *pcoinsdbview;
+    std::unique_ptr<CCoinsViewDB> pcoinsdbview;
```
ryanofsky: In commit "Use unique_ptr for pcoinscatcher/pcoinsdbview/pcoinsTip/pblocktree"

Can this variable just be deleted like the one in rpcnestedtests.h, since it is shadowing a global with the same name? If it can't be deleted, it would be good to have a comment noting that it does shadow the global (and maybe explaining why).

ryanofsky (commented):
utACK 5a92b9d89509fa17ea341da3794b34ac330669a9

laanwj:
> Avoid resource leaks (specifically: forgetting to delete an object created using new)

Just want to add that this is not a panacea. In a few places I've explicitly used manual pointer allocation/deallocation instead of unique_ptr (for globals) because the order of deallocation matters, so to explicitly control the lifespan. So in some of these cases, forgetting explicit `.clear()` is as bad as leaking and the 'parachute' that automatic deallocation provides isn't worth anything (could even result in spurious crashes at shutdown).

bytting:
@ryanovsky Didn't you declare MakeUnique somewhere else already?

I like this patch, and encourage anyone who wants more meat on the bone to read Sutter's Mill articles about smart pointers

https://herbsutter.com/gotw/

practicalswift:
@laanwj Absolutely true! My goal with this PR is to address only the trivial/safe cases :-)

I know about the dependency on deconstruction order related to `LogPrintStr(…)` (`FILE* fileout`, `boost::mutex* mutexDebugLog` and `std::list<std::string>* vMsgsBeforeOpenLog`) - so these where intentionally excluded. Were you thinking about to those when mentioning spurious crashes at shutdown? :-)

laanwj:
> Were you thinking about to those when mentioning spurious crashes at shutdown? :-)

No, I'll comment inline.

src/httprpc.cpp
```diff
@@ -253,8 +253,6 @@ void StopHTTPRPC()
     LogPrint(BCLog::RPC, "Stopping HTTP RPC server\n");
     UnregisterHTTPHandler("/", true);
     if (httpRPCTimerInterface) {
-        RPCUnsetTimerInterface(httpRPCTimerInterface);
-        delete httpRPCTimerInterface;
-        httpRPCTimerInterface = 0;
+        RPCUnsetTimerInterface(httpRPCTimerInterface.get());
```
laanwj: add `httpRPCTimerInterface.reset()` after this

practicalswift: @laanwj Thanks for reviewing! Fixed!

src/wallet/test/wallet_test_fixture.cpp
```diff
@@ -17,18 +17,16 @@ WalletTestingSetup::WalletTestingSetup(const std::string& chainName):
...
 
     RegisterWalletRPCCommands(tableRPC);
 }
 
 WalletTestingSetup::~WalletTestingSetup()
 {
-    UnregisterValidationInterface(pwalletMain);
-    delete pwalletMain;
-    pwalletMain = nullptr;
+    UnregisterValidationInterface(pwalletMain.get());
```
laanwj: add `pwalletMain.reset()` (this is not part of the WalletTestingSetup but a global, so needs to be cleaned up here and now)

laanwj:
I found two in a cursory look-over. There might be more instances where explicit `.reset` is important (where there was an explicit `delete` before), please be careful in reviewing this.

> LogPrintStr(…) (FILE* fileout, boost::mutex* mutexDebugLog and std::list<std::string>* vMsgsBeforeOpenLog)
> so these were intentionally excluded

Agree with leaving these changes out for that "magic", just too easy to inadvertently break something there.

src/wallet/test/wallet_test_fixture.cpp
practicalswift: Thanks! Fixed!

practicalswift:
@ryanofsky Thanks for reviewing! Great feedback - all items addressed.

ryanofsky (commented):
utACK 6b7863325b74a6b45aa77a1fc9f3586520783f5b. Changes since previous were just all the suggested changes (restoring some explicit deletes instead of relying on destructors, removing pcoinsdbview shadowing, adding MakeUnique) plus rebasing.

practicalswift:
Rebased!

laanwj:
Needs rebase again as #11007 was a subset of this.

practicalswift:
@laanwj Done! :-)

practicalswift:
Rebased! :-)

practicalswift:
Anyone willing to review? I think this one should be ready for merge :-)

ryanofsky (commented):
utACK 3d7109971002358fbf62477068dc06f67234d2b4. Only change since previous review is rebase and changing a few 0s to nullptrs.

sipa:
utACK 3d7109971002358fbf62477068dc06f67234d2b4

practicalswift:
Rebased!

practicalswift:
Rebased!

ryanofsky (commented):
0a38fb91025cb3a190aae5d8c44093a5d599f8c1. Only change is rebase and new MakeUnique in wallet/db.cpp

practicalswift:
Thanks for the re-review @ryanofsky!

Are we getting close to merge? :-)

MarcoFalke:
> Are we getting close to merge? :-)

We might want to wait until the 0.15.1 stuff is merged, to ease backports. Also, noting that some commits should be squashed before merge.

promag:
utACK 0a38fb9.

practicalswift:
@MarcoFalke Sounds good! Should I squash into one commit now or should I wait for further instructions?

practicalswift:
Rebased! :-)

laanwj:
Looks good to me now, utACK a357293

PR #11046, opened 2017-08-14
Replace boost sleep_for with C++11 equivalent

jonasschnelli:
Replaces `boost::thread::sleep_for` with C++ equivalent.
Removes ugly boost `sleep_for` version check and compatibility control structure.

Not meant for 0.15.

### DIFF
--- a/src/test/scheduler_tests.cpp
+++ b/src/test/scheduler_tests.cpp
@@ -11,6 +11,9 @@
 #include <boost/thread.hpp>
 #include <boost/test/unit_test.hpp>
 
+#include <chrono>
+#include <thread>
+
 BOOST_AUTO_TEST_SUITE(scheduler_tests)
 
 static void microTask(CScheduler& s, boost::mutex& mutex, int& counter, int delta, boost::chrono::system_clock::time_point rescheduleTime)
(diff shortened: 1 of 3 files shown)


laanwj:
Nice.
The difference is that this one is not interruptible. Not sure if that still makes a difference, though.
@theuni?

practicalswift:
Very nice! Concept strong ACK :-)

XingZheZ:
Great!

XingZheZ:
可能还有细节需要考虑。

theuni:
Yes, we can't do this just yet. This would make MilliSleep non-interruptible, meaning that a few threads (see upnp for an easy example) would hang around for several minutes before dying.

Due to the renewed interest (see #10923) as well, I'll finish up the std thread/mutex/condvar conversion asap so that we can proceed with changes like these. There aren't many left now.

src/test/scheduler_tests.cpp
```diff
@@ -11,6 +11,9 @@
 #include <boost/thread.hpp>
 #include <boost/test/unit_test.hpp>
 
+#include <chrono>
```
promag: Remove, it's included in thread header.

src/test/scheduler_tests.cpp
```diff
@@ -11,6 +11,9 @@
...
+#include <thread>
```
promag: Include before boost headers.

src/utiltime.cpp
```diff
@@ -10,9 +10,10 @@
 #include "utiltime.h"
 
 #include <atomic>
+#include <chrono>
```
promag: Remove.

laanwj:
Closing for now, as we can't do this yet. Please rebase and reopen once the dependencies went in.

MarcoFalke:
After f26866b9caf9ff27d129c703f51b8c2922000243, this should now be possible to do. See here: #16117

PR #11047, opened 2017-08-14
[tests] rename functional tests

jnewbery:
This may be a non-starter because of the churn caused, but I thought I'd put it out there and see what people think. Not that git is usually smart enough to figure out filename only changes, so rebasing shouldn't be an issue.

This PR changes the functional tests to have a consistent naming scheme:

- tests for individual RPC methods are named `rpc_...`
- tests for interfaces (REST, ZMQ, RPC features) are named `interface_...`
- tests that explicitly test the p2p interface are named `p2p_...`
- tests for wallet features are named `wallet_...`
- tests for mining features are named `mining_...`
- tests for mempool behaviour are named `mempool_...`
- tests for full features that aren't wallet/mining/mempool are named `feature_...`

Rationale: it's sometimes difficult for new contributors to know what's already covered by existing tests and where new tests should be added. Naming in a consistent fashion makes it easier to see what's already covered at a glance.


ryanofsky (commented):
I like this because there have been times I haven't known how to name a new test based on the existing names. Two suggestions:

- Could more consistently abbreviate transaction as either tx or txn
- Maybe rename to `category/test.py` instead of `category_test.py` to make it easier to distinguish between test utitilities (like combine_logs.py and test_runner.py) and tests, and to make the utilities more discoverable.

promag:
> Maybe rename to `category/test.py` instead of `category_test.py`.

:+1:

MarcoFalke:
Concept ACK. At this point we have so many test scripts that some sort of naming scheme or grouping really makes sense.

jonasschnelli:
I would have preferred folder but I guess that makes things complicated.
utACK deda19ea50b49cebe7e62caf3770989e46af505f

MarcoFalke:
Would you mind adding the naming rules from the OP to the QA developer
notes?

jnewbery:
@jonasschnelli folders is possible. This is just a concept at the moment. @ryanofsky also prefers folders. I'm happy to take feedback on what scheme people prefer.

laanwj:
> I would have preferred folders but I guess that makes things complicated.

I'd dread the resulting python mess when using folders as it conflicts with modules.
Also having them in one directory makes it easier to find all the tests.
So I slightly prefer this naming scheme.

MeshCollider:
I also prefer the prefix style over the folders

mess110:
For now, I think prefix solves the problem of grouping and provides some extra clarity.

jnewbery:
Updated with more consistent naming. `test_runner.py` and `README.md` have also been updated.

jnewbery:
reimplemented as a scripted-diff, since this is going to need rebase for every change in `/test/functional`.

MarcoFalke:
Can you remove the scripted diff tags? It takes longer to review the script than to look at the git diff. Though, feel free to keep the script in the commit body if that is helpful to you.

You might want to ask if anyone objects those changes on Thursday. If not, we can merge on Friday.

jnewbery:
Rebased with the final commit not implemented as a scripted-diff.

test/functional/README.md
```diff
@@ -27,6 +27,20 @@ don't have test cases for.
...
+    - `feature` for tests for full features that aren't wallet/mining/mempool, eg `feature_rbf.py`
+    - `interface` for tests for other interfaces (REST, ZMQ, etc), eg `interface_rest.py`
+    - `mempool` for tests for mempool behaviour, eg `mempool_reorg.py`
+    - `mining` for tests for mining features, eg `mining_prioritisetransaction.py`
+    - `p2p` for tests that explicitly test the p2p interface, eg `p2p_disconnect_ban.py`
+    - `rpc` for tests for individual RPC methods or features, eg `rpc_listtransactions.py`
+    - `wallet` for tests for wallet features, eg `wallet_keypool.py`
+- use an underscore to separate words
+    - exception: for tests for specific RPCs or command line options which don't include underscores, name the test after the exact RPC or argument name, eg `rpc_decodescript.py`, not `rpc_decode_script.py`
+- Don't use the redundant work `test` in the name, eg `interface_zmq.py`, not `interface_zmq_test.py`
```
ajtowns: Nit: "redundant word" not work

test/functional/test_runner.py
```diff
@@ -54,104 +54,104 @@
...
-    'p2p-compactblocks.py',
-    'segwit.py',
+    'p2p_block.py',
+    'rpc_fundrawtransaction.py',
+    'p2p_compactblocks.py',
+    'p2p_segwit_activation.py',
     # vv Tests less than 2m vv
-    'wallet.py',
-    'wallet-accounts.py',
-    'p2p-segwit.py',
```
ajtowns: This seems backwards: segwit.py becomes p2p_segwit_activation, but p2p-segwit.py becomes feature_segwit?

test/functional/test_runner.py
```diff
@@ -54,104 +54,104 @@
...
-    'merkle_blocks.py',
-    'receivedby.py',
-    'abandonconflict.py',
-    'bip68-112-113-p2p.py',
-    'rawtransactions.py',
-    'reindex.py',
+    'rpc_txoutproof.py',
+    'wallet_listreceivedby.py',
+    'wallet_abandonconflict.py',
+    'p2p_csv_activation.py',
```
ajtowns: This doesn't seem to test p2p protocol aspects, so maybe should be feature_csv_activation? (Or feature_bip68_112_113_activation?)

test/functional/test_runner.py
```diff
@@ -54,104 +54,104 @@
...
-    'wallet-hd.py',
-    'walletbackup.py',
+    'wallet_hd.py',
+    'wallet_backup.py',
     # vv Tests less than 5m vv
-    'p2p-fullblocktest.py',
-    'fundrawtransaction.py',
-    'p2p-compactblocks.py',
-    'segwit.py',
+    'p2p_block.py',
```
ajtowns: p2p_block tests block acceptance and reorg behaviour rather than the p2p protocol so should probably be in the feature_ namespace

ajtowns:
ACK with a few nits (see other comments).

MarcoFalke:
Will be closing in a couple of weeks, if this sits still.

jnewbery:
Closing in favour of #11774. Thanks @ajtowns!

PR #11049, opened 2017-08-14
coincontrol can filter for segwit inputs, expose fundraw option

instagibbs:
This is useful for external signers such as Ledger where only one type of input is allowed for any given transaction to be signed ~~or in the case where you want to explicitly make a non-malleable transaction.~~

### DIFF
--- a/src/wallet/coincontrol.cpp
+++ b/src/wallet/coincontrol.cpp
@@ -1,0 +1,17 @@
+// Copyright (c) 2011-2017 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "coincontrol.h"
+
+bool InputTypeFromString(const std::string& input_string, InputType& input_type) {
+    static const std::map<std::string, InputType> input_types = {
+        {"LEGACY", InputType::LEGACY},
+        {"SEGWIT", InputType::SEGWIT},
+        {"ALL", InputType::ALL},
+    };
+    auto it = input_types.find(input_string);
+    if (it == input_types.end()) return false;
+    input_type = it->second;
+    return true;
+}
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -6,6 +6,7 @@
 #include "amount.h"
 #include "base58.h"
 #include "chain.h"
+#include "coincontrol.h"
 #include "consensus/validation.h"
 #include "core_io.h"
 #include "init.h"
@@ -2811,6 +2812,10 @@ UniValue fundrawtransaction(const JSONRPCRequest& request)
                             "         \"UNSET\"\n"
                             "         \"ECONOMICAL\"\n"
                             "         \"CONSERVATIVE\"\n"
+                            "     \"input_type\"             (string, optional, default=ALL) The type of inputs used to fund the transaction, must be one of:\n"
+                            "         \"ALL\"                Only use non-segwit inputs\n"
+                            "         \"LEGACY\"             Only use segwit inputs\n"
+                            "         \"SEGWIT\"             Use all input types\n"
                             "   }\n"
                             "                         for backward compatibility: passing in a true instead of an object will result in {\"includeWatching\":true}\n"
                             "\nResult:\n"
@@ -2860,6 +2865,7 @@ UniValue fundrawtransaction(const JSONRPCRequest& request)
                 {"replaceable", UniValueType(UniValue::VBOOL)},
                 {"conf_target", UniValueType(UniValue::VNUM)},
                 {"estimate_mode", UniValueType(UniValue::VSTR)},
+                {"input_type", UniValueType(UniValue::VSTR)},
             },
             true, true);
 
@@ -2905,6 +2911,11 @@ UniValue fundrawtransaction(const JSONRPCRequest& request)
             }
             if (!FeeModeFromString(options["estimate_mode"].get_str(), coinControl.m_fee_mode)) {
                 throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid estimate_mode parameter");
+            }
+        }
+        if (options.exists("input_type")) {
+            if (!InputTypeFromString(options["input_type"].get_str(), coinControl.m_input_type)) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid input_type parameter");
             }
         }
       }
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -2215,6 +2215,25 @@ void CWallet::AvailableCoins(std::vector<COutput> &vCoins, bool fOnlySafe, const
                     continue;
                 }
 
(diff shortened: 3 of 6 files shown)


promag:
Drop `Desired` from enum?

sipa:
Concept ACK

src/wallet/wallet.h
```diff
@@ -1223,4 +1223,12 @@ bool CWallet::DummySignTx(CMutableTransaction &txNew, const ContainerType &coins
     return true;
 }
 
+enum class InputType {
+    LEGACY, //! Only use non-segwit inputs
+    SEGWIT, //! Only use segwit inputs
+    ALL,    //! Use any inputs
```
promag: `//! Use all input types`

promag (commented):
Concept ACK.

What about `unspent_types` (instead of `input_type`) and use `txnouttype` instead of adding the new `InputType` enum (which also adds *legacy*, both not well defined terms).

`unspent_types` should an array of strings, each parsed to a `txnouttype` value. I think RPC interface could be as low level as this.

instagibbs:
@promag I see negative value to the user to have them iterate what type of segwit inputs they'd like, especially as that list grows. What the user will typically care about is whether or not the resulting transaction can be malleated.

instagibbs:
travis failure seems unrelated, `p2p-leaktest.py`

src/wallet/rpcwallet.cpp
```diff
@@ -2811,6 +2811,10 @@ UniValue fundrawtransaction(const JSONRPCRequest& request)
                             "         \"UNSET\"\n"
                             "         \"ECONOMICAL\"\n"
                             "         \"CONSERVATIVE\"\n"
+                            "     \"input_type\"             (string, optional, default=ALL) The type of inputs used to fund the transaction, must be one of:\n"
+                            "         \"ALL\"\n"
```
jonasschnelli: Maybe re-use the comments from https://github.com/bitcoin/bitcoin/pull/11049/files#diff-12635a58447c65585f51d32b7e04075bR1227 here?

src/wallet/wallet.h
```diff
@@ -1223,4 +1223,12 @@ bool CWallet::DummySignTx(CMutableTransaction &txNew, const ContainerType &coins
     return true;
 }
 
+enum class InputType {
```
jonasschnelli: I think that enum class belong to `coincontrol.h`

src/wallet/wallet.h
```diff
@@ -1223,4 +1223,12 @@ bool CWallet::DummySignTx(CMutableTransaction &txNew, const ContainerType &coins
...
+    LEGACY, //! Only use non-segwit inputs
+    SEGWIT, //! Only use segwit inputs
+    ALL,    //! Use all input types
+};
+
+bool InputTypeFromString(const std::string& mode_string, InputType& input_type);
```
jonasschnelli: Same here: could be defined and implemented in `coincontrol.h/cpp`

src/wallet/wallet.cpp
```diff
@@ -2215,6 +2215,26 @@ void CWallet::AvailableCoins(std::vector<COutput> &vCoins, bool fOnlySafe, const
                     continue;
                 }
 
+
+                if (coinControl) {
+                    std::vector<std::vector<unsigned char> > vSolutions;
+                    txnouttype whichType;
```
jonasschnelli: Maybe it would be saner to set a default value in case `Solver()` returns false and hasn't set `whichType`?

jonasschnelli:
Concept ACK

Leviathn:
Concept ACK

NicolasDorier:
Concept NACK.

What you are interested is not SEGWIT output, what you are interested is third party `malleabilitySafe` coins.
All segwit coin are not malleabilitySafe. A segwit coin is malleability safe in the following condition:

1. The coin is confirmed
2. The unconfirmed parent coins are all segwit coins, are not meant to be RBF, and are trusted.

instagibbs:
@NicolasDorier I can remove that reasoning from the PR description, but it's still a use-case that would exist even with a more "complete" anti-malleability setting.

instagibbs:
fixed up @jonasschnelli 's concerns

src/wallet/wallet.cpp
```diff
@@ -2215,6 +2215,26 @@ void CWallet::AvailableCoins(std::vector<COutput> &vCoins, bool fOnlySafe, const
                     continue;
                 }
 
+
+                if (coinControl) {
```
promag: Correct me if I'm wrong but this block will incur in a substantial performance penalty? What about:
```cpp
if (coinControl && coinControler->m_input_type != InputType::ALL) {
```

src/wallet/coincontrol.cpp
```diff
@@ -0,0 +1,17 @@
+// Copyright (c) 2011-2017 The Bitcoin Core developers
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "coincontrol.h"
+
+bool InputTypeFromString(const std::string& input_string, InputType& input_type) {
+    static const std::map<std::string, InputType> input_types = {
+    {"LEGACY", InputType::LEGACY},
```
promag: Indent.

promag (commented):
IMO just be able to filter the unspent type is great. Something like:
```python
fundrawtransaction(rawtx, {"unspent_type":"*"});            # default (all)
fundrawtransaction(rawtx, {"unspent_type":"segwit_v0_*"});  # witness_v0_keyhash and
                                                            # witness_v0_scripthash
fundrawtransaction(rawtx, {"unspent_type":"multisig"});
```
See https://github.com/bitcoin/bitcoin/blame/master/src/script/standard.cpp#L25-L32.

So, no new type, no new "legacy", scalable and really "raw" as the call states it is.

In any case, some comments.

instagibbs:
@promag I'd really rather not have regex-y inputs presented to users unless there's a compelling use-case. Open to suggestions of course, I likely missed some.

Fixed comments.

src/wallet/rpcwallet.cpp
```diff
@@ -2907,6 +2913,11 @@ UniValue fundrawtransaction(const JSONRPCRequest& request)
                 throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid estimate_mode parameter");
             }
         }
+        if (options.exists("input_type")) {
+            if (!InputTypeFromString(options["input_type"].get_str(), coinControl.m_input_type)) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid input_type parameter");
```
promag: Missing test for this error.

NicolasDorier:
@instagibbs I don't see why someone would want to filter by type on fundraw?

Is it so that the ultimate signer don't have to know how to use legacy signing?

instagibbs:
Closing for now, as for my use-case it's actually not needed: At least for Ledger  one can sign by simply "signing twice", segwit inputs then non-segwit inputs. No need to special case coin selection. h/t @greenaddress

PR #11056, opened 2017-08-15
disable jni in builds

instagibbs:
Suggested by @theuni 

Unneeded and was causing travis issues downstream.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1261,7 +1261,7 @@ if test x$need_bundled_univalue = xyes; then
   AC_CONFIG_SUBDIRS([src/univalue])
 fi
 
-ac_configure_args="${ac_configure_args} --disable-shared --with-pic --with-bignum=no --enable-module-recovery"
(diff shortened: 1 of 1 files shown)


theuni:
utACK. @instagibbs had an unrelated build failure due to (auto) jni detection in libsecp256k1. Regardless of that problem (which should be addressed), we should be deterministic in the downstream features we enable/build.

laanwj:
What is jni and how would it possibly have any influence on us?

sipa:
libsecp256k1 has (optional) JNI bindings

jonasschnelli:
utACK 844b73e486ca146f3f9f3060e878814752c25fd5 though I think the libsecp256k1 JNI (java native interface) default should not be auto ("no" as default makes more sense IMO).

laanwj:
If the answer is that it doesn't affect us in any way, then it should be disabled, I agree. If anything it's a waste of resources building.

utACK.

fanquake:
ACK https://github.com/bitcoin/bitcoin/commit/844b73e486ca146f3f9f3060e878814752c25fd5

PR #11067, opened 2017-08-16
[qa] TestNode: Add wait_until_stopped helper method

maflcko:
This adds a helper method `wait_until_stopped` to the `TestNode` class. This should prevent numerous `time.sleep()` over all places.

Additionally, the timeout behavior is restored. (Was removed by the introduction of `TestNode`.)
This should prevent tests from running indefinitely by accident.

### DIFF
--- a/test/functional/test_framework/test_node.py
+++ b/test/functional/test_framework/test_node.py
@@ -17,8 +17,11 @@ from .util import (
     assert_equal,
     get_rpc_proxy,
     rpc_url,
+    wait_until,
 )
 from .authproxy import JSONRPCException
+
+BITCOIND_PROC_WAIT_TIMEOUT = 60
 
 class TestNode():
     """A class for representing a bitcoind node under test.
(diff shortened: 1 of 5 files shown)


jnewbery:
Concept ACK, but I'm concerned about defining a `wait_until` in util.py, when there's already a `wait_until` definied in mininode.py. This is a problem since wildcard imports are used in a lot of tests, so behavior will be unpredictable.

jnewbery:
Rebase on #11068?

jnewbery:
Looks great. Tested ACK fa7d395048c9395bbf23439c2986751db560aa8b

MarcoFalke:
Solved merge conflict fa7d39504 ... fa72a0011

test/functional/test_framework/test_node.py
```diff
@@ -121,25 +124,27 @@ def is_node_stopped(self):
...
-            return True
-        return False
+        if return_code is None:
+            return False
+
+        # process has stopped. Assert that it didn't return an error code.
+        assert_equal(return_code, 0)
+        self.running = False
+        self.process = None
+        self.rpc_connected = False
```
jnewbery: I think this method should also set `self.rpc` to None

jnewbery (commented):
Tested ACK fa72a0011bead99f60e6442b79ddad407d3a3df2.

One nit inline. Otherwise looks good.

test/functional/test_framework/test_node.py
MarcoFalke: @jnewbery Done

test/functional/test_framework/test_node.py
```diff
@@ -17,9 +17,12 @@
     assert_equal,
     get_rpc_proxy,
     rpc_url,
+    wait_until,
 )
 from .authproxy import JSONRPCException
 
+BITCOIND_PROC_WAIT_TIMEOUT = 60
```
jnewbery: This constant is now only used in one place. Perhaps remove it as a constant and add it as an integer literal in the definition of `wait_until_stopped()` with a comment.

If not, perhaps move this constant assignment down to just above the call to `wait_until_stopped()`

MarcoFalke: I prefer to put constants at the top of a file and my personal taste is that it is ugly to place them between or inside methods/functions.

jnewbery: That's reasonable

jnewbery:
utACK faa8d9581affde35a8242c63fca8a52004d4f943

PR #11068, opened 2017-08-16
qa: Move wait_until to util

maflcko:
This moves `wait_until` to `util.py` to make it generally available to python tests.
Also, `wait_until` now takes an optional lock that is acquired while testing the predicate.
Previously the lock was always acquired, even when it was not necessary, cf. `disconnect_ban.py`.

### DIFF
--- a/test/functional/bip65-cltv-p2p.py
+++ b/test/functional/bip65-cltv-p2p.py
@@ -109,7 +109,7 @@ class BIP65Test(BitcoinTestFramework):
         node0.send_and_ping(msg_block(block))
         assert_equal(int(self.nodes[0].getbestblockhash(), 16), tip)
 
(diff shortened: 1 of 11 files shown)


jnewbery:
I don't love this name, but we can tidy up in a later commit so concept ACK (since it's required for #11067)

MarcoFalke:
@jnewbery I pushed my preferred version. No need for unwanted intermediate
commits.

jnewbery:
Great! Tested ACK 08ce33f8e95efa81b37ddc6b3350462c61bbfd51 . I've long wanted to move the assert up into `wait_util()`, but for some reason I thought there were more call sites where the return value wasn't being asserted.

This is a good change. `wait_until()` is a pure utility function and should live in util.py. The caller passing in the lock seems fine to me.

Issue #9573, opened 2017-01-18
6696b46: Clang Static Analyzer Report


kallewoof:
BRANCH: master
COMMIT: 6696b4635ceb9b47aaa63244bff9032fa7b08354
ISSUE COUNT: 13
EXCLUDED: `src/bench`, `src/test`, `src/qt`

DISCLAIMER: These results have not been thoroughly confirmed, and may be improbable or flat out invalid, but it's worth having a list of these somewhere.

==============================================

SEMANTIC ISSUES [2 present, 0 confirmed]:
* [src/init.cpp:1524:58](https://github.com/bitcoin/bitcoin/blob/master/src/init.cpp#L1524:58): Variable `nStart` may be uninitialized when used here
* [src/rpc/server.cpp:496:29](https://github.com/bitcoin/bitcoin/blob/master/src/rpc/server.cpp#L496:29): Code will never be executed

LOGIC ERROR [6 present, 0 confirmed, 1 fixed]:
* [src/rpc/mining.cpp:551:20](https://github.com/bitcoin/bitcoin/blob/master/src/rpc/mining.cpp#L551:20): Access to field `nNonce` results in a dereference of a null pointer (loaded from variable `pblock`)
* [src/wallet/rpcdump.cpp:506:5](https://github.com/bitcoin/bitcoin/blob/master/src/wallet/rpcdump.cpp#L506:5): Access to field `nHeight` results in a dereference of a null pointer (loaded from variable `pindex`):
   1 [src/wallet/rpcdump.cpp:429:9](https://github.com/bitcoin/bitcoin/blob/master/src/wallet/rpcdump.cpp#L429:9): Assuming `fPruneMode` is 0
   2 [src/wallet/rpcdump.cpp:449:12](https://github.com/bitcoin/bitcoin/blob/master/src/wallet/rpcdump.cpp#L449:12): Loop body executed 0 times
   3 [src/wallet/rpcdump.cpp:499:27](https://github.com/bitcoin/bitcoin/blob/master/src/wallet/rpcdump.cpp#L499:27): Calling `CChain::Tip`
   4 [src/chain.h:429:5](https://github.com/bitcoin/bitcoin/blob/master/src/chain.h#L429:5): Entered call from `importwallet`
   5 [src/wallet/rpcdump.cpp:499:27](https://github.com/bitcoin/bitcoin/blob/master/src/wallet/rpcdump.cpp#L499:27): Returning from `CChain::Tip`
[...]

PR #11074, opened 2017-08-17
Assert that CWallet::SyncMetaData finds oldest transaction.

BitonicEelis:
Without this assert, the Clang static analyzer warns about subsequent dereferencing of copyFrom, because it can't be sure that it's not nullptr. See #9573.

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -569,6 +569,9 @@ void CWallet::SyncMetaData(std::pair<TxSpends::iterator, TxSpends::iterator> ran
             copyFrom = &mapWallet[hash];
         }
     }
+
+    assert(copyFrom);
+
     // Now copy data from copyFrom to rest:
     for (TxSpends::iterator it = range.first; it != range.second; ++it)
     {


src/wallet/wallet.cpp
```diff
@@ -569,6 +569,9 @@ void CWallet::SyncMetaData(std::pair<TxSpends::iterator, TxSpends::iterator> ran
             copyFrom = &mapWallet[hash];
         }
     }
+
+    assert(copyFrom);
```
promag: From what I've seen there is at least one element in the range:
```cpp
TxSpends::iterator it = range.first;
const CWalletTx* copyFrom = mapWallet[it->second];
for (; it != range.second; ++i) {
    ...
```
Maybe this assert should come first:
```cpp
assert(range.first != range.second);
```

BitonicEelis: That would not suffice to silence the analyzer, because copyFrom is also nullptr if the range is nonempty but n == nMinOrderPos in every iteration.

laanwj:
utACK 6c4042a

PR #11077, opened 2017-08-17
[tests] fix timeout issues from TestNode

jnewbery:
Fixes a couple of bugs from the introduction of TestNode:

- test scripts were no longer able to specify a custom timeout for
starting a node. Therefore tests with nodes that take a long time to
start up (eg pruning.py) would fail.
- the test for whether a node has failed on start up was broken
by changing 'assert x is None' to 'assert not x'. Since
subprocess.poll() can return None (indicating the node is still running)
or 0 (indicating the node exited with return code 0), this was a
regression.

### DIFF
--- a/test/functional/test_framework/test_node.py
+++ b/test/functional/test_framework/test_node.py
@@ -34,7 +34,11 @@ class TestNode():
         self.index = i
         self.datadir = os.path.join(dirname, "node" + str(i))
         self.rpchost = rpchost
-        self.rpc_timeout = timewait
+        if timewait:
+            self.rpc_timeout = timewait
+        else:
+            # Wait for up to 60 seconds for the RPC server to respond
+            self.rpc_timeout = 60
         if binary is None:
             self.binary = os.getenv("BITCOIND", "bitcoind")
         else:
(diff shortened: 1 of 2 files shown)


laanwj:
utACK, needs rebase

jnewbery:
Thanks @laanwj . Rebased

mess110:
Looks good, also ran tests locally

MarcoFalke:
post merge utACK 2b4ea52

test/functional/test_framework/test_node.py
```diff
@@ -34,7 +34,11 @@ def __init__(self, i, dirname, extra_args, rpchost, timewait, binary, stderr, mo
         self.index = i
         self.datadir = os.path.join(dirname, "node" + str(i))
         self.rpchost = rpchost
-        self.rpc_timeout = timewait
+        if timewait:
+            self.rpc_timeout = timewait
+        else:
+            # Wait for up to 60 seconds for the RPC server to respond
+            self.rpc_timeout = 60
```
MarcoFalke: Previously we used a default of `HTTP_TIMEOUT = 30`.

Why was this changed?

jnewbery: Longer is better (I think?). Travis is so unpredictable that one minute seemed appropriate.

We can tune it back down if you think that's better.

PR #7510, opened 2016-02-11
Read/write bitcoin_rw.conf for exposing shared Daemon/GUI options in the GUI

luke-jr:
This gives GUI options for maxuploadtarget and peerbloomfilters, while retaining the user-set values in the Daemon as well (unlike current GUI options). This is done by modifying a bitcoin_rw.conf which is loaded in addition to bitcoin.conf (before, because that's how our precedence is setup). There's about 200 LOC for making these updates, but it is all well-tested, and the worst-case scenario doesn't affect the user-managed bitcoin.conf file.

The ugliest part having to special-case additions to the OptionsModel stuff. It would be nice to avoid that, but I don't see any way to do so as long as an enum is being used as the interface there.

Issue #6461, opened 2015-07-21
Exposing prune option in GUI


sime:
With a pruned wallet possible in master I believe we should expose the option in the GUI.

The Main tab in preferences will offer a chance to toggle the node type and size.
<img width="612" alt="pruning_gui_off" src="https://cloud.githubusercontent.com/assets/216917/8766147/834fb842-2e2c-11e5-94b0-c342132243bf.png">
<img width="612" alt="pruning_gui_on" src="https://cloud.githubusercontent.com/assets/216917/8766148/8351a0c6-2e2c-11e5-841b-af02fd77ff44.png">

The prune size UI should not permit sizes less than 550.

jonasschnelli:
concept ACK.

ajweiss:
There are a couple of considerations to take into account here:

1. This only prunes the blockstore, it does not keep disk usage beneath the pruning target. I think this should be called out a little more clearly.
2. The pruning target is just that, a target, sometimes the target will be overshot a bit. Maybe this can be alluded to in the UI, or maybe it's for the release notes.
3. Enabling pruning essentially binds a given instance to its wallet.  That is, once pruned, you can no longer import addresses, keys or wallets without a full resync.  Since this is an hours long operation, I think there should probably be some warning.
4. Enabling pruning disables several node functions that support the network, such as historical block service and serving of proofs to SPV clients, most GUI users probably don't care, but it's a side effect.

While there are certainly differing opinions on this, I personally believe that the pruning feature at this stage is still somewhat experimental and has side effects that only people who are paying close attention will be aware of.  With time, some of these side effects will go away or be minimized, but if there's going to be a GUI element, I think it makes sense to call this out and direct the user to the release notes.  If someone turns it on and it results in an unexpected multi-hour sync being required, that's pretty bad UX.   

Honestly, I'd err on the side of requiring the user to manually supply the option in the config file as I believe users that are willing to do this are more likely to have read the release notes and fully understand the ramifications of turning it on.

luke-jr:
Probably it should be a button rather than a checkbox to signify the semi-irreversibility of the operation. For example, a button "Enable Pruning" which displays a dialogue box explaining all the risks/limitations of a pruned node (doesn't contribute to the network, tied to the current wallet, etc), and asks for a confirmation.

jonasschnelli:
+1 for @luke-jr dialog box.

sipa:
Agree with Luke.

ajweiss:
[...]

PR #11082, opened 2017-08-17
Add new bitcoin_rw.conf file that is used for settings modified by this software itself

luke-jr:
This is part of #7510, without the new GUI settings (ie, just the minimal framework for the RW conf file).


Sjors:
I'm not a big fan of multiple config files. I would prefer if QT just edited `bitcoin.conf` and tells the user to do so manually if things gets too complicated. See also #6461.

Sjors:
I tried just making `bitcoin.conf` writeable instead of having two files in #12833, but that seems to raise some objections. So in that case: Concept ACK. 

Can you rebase this? From my experience with the other PR that should be easy and it worked quite well.

luke-jr:
Rebased

Sjors:
tACK aac0501 (tested through #12833)

Sjors:
Needs another rebase (though maybe not worth the effort without more Concept ACKs) due to #11862.

ajtowns (commented):
ConceptACK. This approach isn't any worse than QSettings, and seems more consistent, and discoverable when anything weird is happening due to conflicting/unexpected settings.

src/init.cpp
```diff
@@ -332,6 +332,7 @@ std::string HelpMessage(HelpMessageMode mode)
     if (showDebug)
         strUsage += HelpMessageOpt("-blocksonly", strprintf(_("Whether to operate in a blocks only mode (default: %u)"), DEFAULT_BLOCKSONLY));
     strUsage += HelpMessageOpt("-conf=<file>", strprintf(_("Specify configuration file. Relative paths will be prefixed by datadir location. (default: %s)"), BITCOIN_CONF_FILENAME));
+    strUsage += HelpMessageOpt("-confrw=<file>", strprintf(_("Specify read/write configuration file. Relative paths will be prefixed by the network-specific datadir location. (default: %s)"), BITCOIN_RW_CONF_FILENAME));
```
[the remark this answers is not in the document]
ryanofsky: > "rwconfig" is an implementation detail; it might be better to describe it via it's purpose instead. "confui -- Specify configuration file that stores settings set in the UI" or something?

IMO, rwconf is nice because it would let us an add an RPC interface for updating settings and persisting them.

MarcoFalke:
Let me know when I can add this back to project 8, i.e. when it is ready for review.

src/util/system.cpp
```diff
@@ -979,6 +1028,215 @@ std::string ArgsManager::GetChainName() const
...
+    while (getline_with_eol(stream_in, s)) {
+        ++lineno;
+
+        have_eof_nl = (!s.empty()) && (*s.rbegin() == '\n');
+        n = s.find('#');
+        has_comment = (n != std::string::npos);
+        if (!has_comment) {
+            n = s.size();
+        }
+        n2 = s.find_last_not_of(ws_chars, n - 1);
```
practicalswift: An integer wraparound will occur here in the case of `n == 0`.

src/util/system.h
```diff
@@ -162,6 +169,10 @@ class ArgsManager
     bool ParseParameters(int argc, const char* const argv[], std::string& error);
     bool ReadConfigFiles(std::string& error, bool ignore_invalid_keys = false);
 
+    void ModifyRWConfigFile(const std::map<std::string, std::string>& settings_to_change);
+    void ModifyRWConfigFile(const std::string& setting_to_change, const std::string& new_value);
```
practicalswift: Make sure parameters name match between declaration and definition :-)

src/util/system.cpp
```diff
@@ -979,6 +1028,215 @@ std::string ArgsManager::GetChainName() const
     return CBaseChainParams::MAIN;
 }
 
+namespace {
+
+    // Like std::getline, but includes the EOL character in the result
+    static bool getline_with_eol(std::istream& streamIn, std::string& s)
```
practicalswift: `static` is redundant here

src/util/system.cpp
```diff
@@ -979,6 +1028,215 @@ std::string ArgsManager::GetChainName() const
...
+            i = streamIn.get();
+            if (i == std::char_traits<char>::eof()) {
+                break;
+            }
+            s.push_back(char(i));
+        }
+        return true;
+    }
+
+    static const char * const ModifyRWConfigFile_ws_chars = " \t\r\n";
```
practicalswift: `static` is redundant here.

src/util/system.cpp
```diff
@@ -979,6 +1028,215 @@ std::string ArgsManager::GetChainName() const
...
+
+    static void ModifyRWConfigFile_SanityCheck(const std::string& s)
```
practicalswift: Same here :-)

src/util/system.cpp
```diff
@@ -979,6 +1028,215 @@ std::string ArgsManager::GetChainName() const
...
+        static std::string ws_chars(ModifyRWConfigFile_ws_chars);
+        if (s.find_first_of(newline_chars) != std::string::npos) {
+            throw std::invalid_argument("New-line in config name/value");
+        }
+        if (ws_chars.find(*s.begin()) != std::string::npos || ws_chars.find(*s.rbegin()) != std::string::npos) {
+            throw std::invalid_argument("Config name/value has leading/trailing whitespace");
+        }
+    }
+
+    static void ModifyRWConfigFile_WriteRemaining(std::ostream& stream_out, const std::map<std::string, std::string>& settings_to_change, std::set<std::string>& setFound)
```
practicalswift: And here :-)

src/util/system.cpp
```diff
@@ -979,6 +1028,215 @@ std::string ArgsManager::GetChainName() const
...
+            if (setFound.find(key) != setFound.end()) {
+                continue;
+            }
+            setFound.insert(key);
+            ModifyRWConfigFile_SanityCheck(key);
+            ModifyRWConfigFile_SanityCheck(val);
+            stream_out << key << "=" << val << "\n";
+        }
+    }
+}
```
practicalswift: Nit: End anonymous namespace with `// namespace`

src/util/system.cpp
```diff
@@ -979,6 +1028,215 @@ std::string ArgsManager::GetChainName() const
...
+                // Rather than change a commented line, comment it out entirely (the existing comment may relate to the value) and replace it
+                stream_out << key << "=" << val << "\n";
+                linebegin.insert(linebegin.begin(), '#');
+            } else {
+                // Just modify the value in-line otherwise
+                n2 = s.find_first_not_of(ws_chars, n + 1);
+                if (n2 == std::string::npos) {
+                    n2 = n + 1;
+                }
+                s = s.substr(0, n2) + val;
```
practicalswift: Nit: Avoid extra allocations by doing:

```
s = s.substr(0, n2);
s += val;
```

luke-jr: I think the current style is more readable.

src/util/system.cpp
```diff
@@ -979,6 +1028,215 @@ std::string ArgsManager::GetChainName() const
...
+    // Like std::getline, but includes the EOL character in the result
+    static bool getline_with_eol(std::istream& streamIn, std::string& s)
+    {
+        int i;
+        i = streamIn.get();
+        if (i == std::char_traits<char>::eof()) {
+            return false;
+        }
+        s.clear();
+        s.push_back(char(i));
```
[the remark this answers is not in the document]
luke-jr: C-style casts aren't good practice in C++.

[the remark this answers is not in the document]
practicalswift: Agreed. `static_cast<char>(i)` then?

src/util/system.cpp
practicalswift: I agree. And readability is probably more important than allocation efficiency in this case :-)

src/util/system.cpp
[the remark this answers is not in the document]
luke-jr: There is no need for a cast at all. C++ allows construction of char just like any other type.

luke-jr:
Finally rebased (and ready for high-pri for review I think).

src/util/system.cpp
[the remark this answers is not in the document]
practicalswift: I'm not sure I follow the "no need for cast" statement. Do we agree on the following two statements? :-)

1. The functional cast expression `char(i)` is equivalent to the C-style cast expression `(char)i`
1. When a C-style cast expression is encountered, the compiler interprets it as the first named cast that satisfies the requirements of the respective cast operator in the order: a. `const_cast<T>(foo)`, b. `static_cast<T>(foo)`, …, etc.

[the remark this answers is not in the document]
sipa: For primitive types we've generally used C-style casts over C++-style ones (they're equivalent in that case, and much less syntactic burden).

Sjors (commented):
Thanks for the rebase! Works like a charm on my freshly rebased #12833.

Does `bitcoin-cli` really also need to parse this? I have a light preference for disallowing settings in `-confrw` that change how `bitcoin-cli` should behave, i.e.:
* `datadir`: this always need to be in QTSettings anyway
* RPC connection info (if someone really needs a custom host/port/binding and non-cookie authentication, they know how to edit `bitcoin.conf`)
* mainnet / testnet / regtest (the GUI can just have a toggle for this, `bitcoind` users know how to edit the config file, and/or can just use `-testnet`)

Reason for this preference is that it makes the life easier for other tools to parse the settings. E.g. wallet_tool #13926 might in the future want to parse bitcoin.conf, and external projects like c-lightning already do/did this.

luke-jr:
Rebased (and ready for high-prio review list)

src/init.cpp
```diff
@@ -1216,6 +1217,15 @@ bool AppInitMain(InitInterfaces& interfaces)
...
     }
 
+    fs::path rwconfig_file_path = GetRWConfigFile(gArgs.GetArg("-confrw", BITCOIN_RW_CONF_FILENAME));
+    if (fs::exists(rwconfig_file_path)) {
+        LogPrintf("R/W Config file: %s\n", rwconfig_file_path.string());
+    } else if (gArgs.IsArgSet("-confrw")) {
+        InitWarning(strprintf(_("The specified R/W config file %s does not exist\n"), rwconfig_file_path.string()));
+    } else {
+        LogPrintf("R/W Config file: %s (not found, skipping)\n", rwconfig_file_path.string());
+    }
```
Empact: nit: path implies file

ryanofsky (commented):
I started reviewing this, and the first few commits seem reasonable. But the last commit is a lot more complicated than I would have expected, and doesn't seem especially maintainable or user friendly.

It seems like all we need is basic storage for settings that can be updated in the gui and maybe over rpc. I'd think the simplest way to do to this would be to add a `UniValue m_rwsettings;` member to the `GlobalArgs` class, tweak `GetArg...()` methods to return these settings, and serialize/deserialize the settings member as needed to `<datadir>/settings.json` by calling existing univalue read and write methods.

Commit "Add new bitcoin_rw.conf file that is used for settings modified by this software itself" (31edb2c940caa3594d5235cc430f6fe5631c464a) is doing something more complicated:

- It adds a new `-confrw` option, for unclear reasons. Maybe to allow users to write settings outside the data dir? Or maybe to allow users control naming of the file, or toggle between different collections of settings within the datadir?
- Instead of storing settings in a simple, machine readable format, it stores them in full ini format with support for comments and other complicated parsing rules.
- Instead of using the existing ini parser, it adds a new parser, which will have to be kept in sync with the existing one.

I'm happy to do more review on this PR if this is the approach people want to take. But would like some confirmation this is actually the approach people want to take.

luke-jr:
Pretty sure UniValue is actually more code than this simple INI modification logic...

INI *is also* a simple machine-readable format, and we're already using it. The minimal complexity in modifications exists to *preserve* user edits (which UniValue doesn't support at all).

jonasschnelli:
There has been a quick discussion about UniValue vs INI files during the todays IRC meeting:
http://www.erisian.com.au/bitcoin-core-dev/log-2019-01-10.html#l-451

Sjors:
tl&dr of the IRC discussion seems to be "whatever is less code", so not much difference to the discussion here. I think there's something to be said for keeping the same format.

If we go for a different format, then I should repeat my point that I don't like how `bitcoin-cli`, and by extension other applications that use the RPC and rely on `bitcoin.conf` to figure out how to talk to it, can also process this file.

I rebased #12833 again.

luke-jr:
Rebased

Sjors:
At least one of the Travis machines failed with `test_bitcoin-qt: util/system.cpp:1213: void ArgsManager::EraseRWConfigFile(): Assertion `!rwconf_path.empty()' failed.`

I also rebased #12833; everything still works afaik.

achow101:
This currently only allows each config option to be specified once. However the normal bitcoin.conf file allows specifying some argument multiple times (e.g. `wallet`). It would be nice if this did that too.

luke-jr:
@achow101 Yes, but that seems better left for a subsequent PR. This is good enough for most use cases right now.

src/util/system.cpp
```diff
@@ -1004,6 +1004,14 @@ bool ArgsManager::ReadConfigFiles(std::string& error, bool ignore_invalid_keys)
         }
     }
 
+    // Check for -testnet or -regtest parameter (BaseParams() calls are only valid after this clause)
```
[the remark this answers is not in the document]
luke-jr: `bitcoin_rw.conf` is located in the network-specific data dir.

achow101:
@luke-jr Care to rebase?

jtimon:
Concept ACK

luke-jr:
Rebased

Sjors:
Rebased #12833 on top.

achow101:
The `bitcoin_rw.conf` file is being written to the network specific directories and not the datadir itself, so when using regtest or testnet mode, it won't be read and handled.

luke-jr:
It's being read from the network-specific directory too...

Sjors:
This is consistent with how QT stores network specific settings.

achow101:
> It's being read from the network-specific directory too...

It didn't seem to be working for me, maybe I was just doing something wrong.

luke-jr:
Trivial rebased

jonasschnelli:
What is the status of this? Shall we close it? It had a few concept ACKs.

Sjors:
I would still like to see this, or something equivalent...

luke-jr:
Forgot to push the last rebase... pushed.

jonasschnelli:
@luke-jr: pull is failing on CIs,... this PR could use some love and care.

luke-jr:
Rebased yet again

jnewbery:
I'm not sure this is needed now that we have the settings.json file. @luke-jr, can you explain why we should consider adding another settings file?

luke-jr:
So we can revert settings.json before 0.21 is released and we're locked in to supporting such a bad idea?

jnewbery:
NACK. Let's close this and move on.

Sjors:
I have no objection to swapping the new `settings.json` out for `bitcoin_rw.conf` before the release, if this gets enough review. But I probably won't review this, because I don't think the difference is worth it. I don't expect users to manually read or edit this, since they can already access `bitcoin.conf`.

The (rather large amount of) code in `ModifyRWConfigStream` presents entirely new file parsing. It would both be more useful, and easier to review, if it was also used to parse `bitcoin.conf`.

fanquake:
Now that we've branched off, and are unlikely to be reverting `settings.json`, I'm going to close this PR.

PR #11083, opened 2017-08-18
Fix combinerawtransaction RPC help result section

jonasnick:
Without this PR it looks like the RPC would return something like a dictionary. But it just returns the transaction in hex.


achow101:
ACK

MeshCollider:
utACK, Travis errors unrelated:
`The command "sudo apt-get update" failed 3 times.`

fanquake:
utACK f9ca0fe, restarted the failing tests.

sipa:
utACK. 0.15?

laanwj:
This is obviously correct, this just returns a string and besides, "naked" dictionary items aren't a thing.
utACK

PR #11087, opened 2017-08-18
Diagnose unsuitable outputs in lockunspent().

### LINKED ISSUE #2667: lockunspent returns true even for non-existent outputs
just use lockunspent with "txid":123 to reproduce

BitonicEelis:
Fixes #2667.

This is a simplified version of pull request #3574, which was abandoned by its author.

I added some tests as well.

### DIFF
--- a/src/wallet/rpcwallet.cpp
+++ b/src/wallet/rpcwallet.cpp
@@ -2447,12 +2447,15 @@ UniValue lockunspent(const JSONRPCRequest& request)
 
     RPCTypeCheckArgument(request.params[1], UniValue::VARR);
 
-    UniValue outputs = request.params[1].get_array();
-    for (unsigned int idx = 0; idx < outputs.size(); idx++) {
-        const UniValue& output = outputs[idx];
-        if (!output.isObject())
-            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, expected object");
-        const UniValue& o = output.get_obj();
+    const UniValue& output_params = request.params[1];
+
+    // Create and validate the COutPoints first.
+
+    std::vector<COutPoint> outputs;
+    outputs.reserve(output_params.size());
+
+    for (unsigned int idx = 0; idx < output_params.size(); idx++) {
+        const UniValue& o = output_params[idx].get_obj();
 
         RPCTypeCheckObj(o,
             {
@@ -2460,20 +2463,50 @@ UniValue lockunspent(const JSONRPCRequest& request)
                 {"vout", UniValueType(UniValue::VNUM)},
             });
 
-        std::string txid = find_value(o, "txid").get_str();
-        if (!IsHex(txid))
+        const std::string& txid = find_value(o, "txid").get_str();
+        if (!IsHex(txid)) {
             throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, expected hex txid");
+        }
 
-        int nOutput = find_value(o, "vout").get_int();
-        if (nOutput < 0)
+        const int nOutput = find_value(o, "vout").get_int();
+        if (nOutput < 0) {
             throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, vout must be positive");
+        }
 
-        COutPoint outpt(uint256S(txid), nOutput);
+        const COutPoint outpt(uint256S(txid), nOutput);
 
-        if (fUnlock)
-            pwallet->UnlockCoin(outpt);
-        else
-            pwallet->LockCoin(outpt);
+        const auto it = pwallet->mapWallet.find(outpt.hash);
+        if (it == pwallet->mapWallet.end()) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, unknown transaction");
+        }
+
+        const CWalletTx& trans = it->second;
+
+        if (outpt.n >= trans.tx->vout.size()) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, vout index out of bounds");
+        }
+
+        if (pwallet->IsSpent(outpt.hash, outpt.n)) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, expected unspent output");
+        }
+
+        const bool is_locked = pwallet->IsLockedCoin(outpt.hash, outpt.n);
+
+        if (fUnlock && !is_locked) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, expected locked output");
+        }
+
+        if (!fUnlock && is_locked) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, output already locked");
+        }
+
+        outputs.push_back(outpt);
+    }
+
+    // Atomically set (un)locked status for the outputs.
+    for (const COutPoint& outpt : outputs) {
+        if (fUnlock) pwallet->UnlockCoin(outpt);
+        else pwallet->LockCoin(outpt);
     }
 
     return true;
--- a/test/functional/wallet.py
+++ b/test/functional/wallet.py
@@ -100,11 +100,19 @@ class WalletTest(BitcoinTestFramework):
         # Exercise locking of unspent outputs
         unspent_0 = self.nodes[2].listunspent()[0]
         unspent_0 = {"txid": unspent_0["txid"], "vout": unspent_0["vout"]}
+        assert_raises_rpc_error(-8, "Invalid parameter, expected locked output", self.nodes[2].lockunspent, True, [unspent_0])
         self.nodes[2].lockunspent(False, [unspent_0])
+        assert_raises_rpc_error(-8, "Invalid parameter, output already locked", self.nodes[2].lockunspent, False, [unspent_0])
         assert_raises_rpc_error(-4, "Insufficient funds", self.nodes[2].sendtoaddress, self.nodes[2].getnewaddress(), 20)
         assert_equal([unspent_0], self.nodes[2].listlockunspent())
         self.nodes[2].lockunspent(True, [unspent_0])
         assert_equal(len(self.nodes[2].listlockunspent()), 0)
+        assert_raises_rpc_error(-8, "Invalid parameter, unknown transaction",
+                              self.nodes[2].lockunspent, False,
+                              [{"txid": "0000000000000000000000000000000000", "vout": 0}])
+        assert_raises_rpc_error(-8, "Invalid parameter, vout index out of bounds",
+                              self.nodes[2].lockunspent, False,
+                              [{"txid": unspent_0["txid"], "vout": 999}])
 
         # Have node1 generate 100 blocks (so node0 can recover the fee)
         self.nodes[1].generate(100)
@@ -142,6 +150,10 @@ class WalletTest(BitcoinTestFramework):
         assert_equal(self.nodes[0].getbalance(), 0)
         assert_equal(self.nodes[2].getbalance(), 94)
         assert_equal(self.nodes[2].getbalance("from1"), 94-21)
+
+        # Verify that a spent output cannot be locked anymore
+        spent_0 = {"txid": node0utxos[0]["txid"], "vout": node0utxos[0]["vout"]}
+        assert_raises_rpc_error(-8, "Invalid parameter, expected unspent output", self.nodes[0].lockunspent, False, [spent_0])
 
         # Send 10 BTC normal
         address = self.nodes[0].getnewaddress("test")


src/wallet/rpcwallet.cpp
```diff
@@ -2397,6 +2399,40 @@ UniValue lockunspent(const JSONRPCRequest& request)
...
+            const auto it = pwallet->mapWallet.find(outpt.hash);
+            if (it == pwallet->mapWallet.end())
+                throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, unknown transaction");
+
+            const CWalletTx& trans = it->second;
+
+            if (outpt.n >= trans.tx->vout.size())
+                throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, vout index out of bounds");
+
+            if (pwallet->IsMine(trans.tx->vout[outpt.n]) == ISMINE_NO)
```
promag: I think is must be either `ISMINE_SPENDABLE` or `ISMINE_WATCH_SOLVABLE`?

promag (commented):
Read developer notes and update accordingly.

src/wallet/rpcwallet.cpp
promag: Do we really have to test `IsMine` and all? In the wallet it's just a `std::set<COutPoint>`, so IMO is enough to check:
- if the input is valid output
- if the lock state is compatible with the argument.

But if there is interest in doing these tests, then perform the cheapest first.

src/wallet/rpcwallet.cpp
```diff
@@ -2397,6 +2399,40 @@ UniValue lockunspent(const JSONRPCRequest& request)
 
         COutPoint outpt(uint256S(txid), nOutput);
 
+        const bool isLocked = pwallet->IsLockedCoin(outpt.hash, outpt.n);
+
+        if (fUnlock)
```
BitonicEelis: Thanks for the review! Are you absolutely sure that attempting to lock an already locked output should be an error?

promag: IMO yes, it can indicate that it was locked by other client as such it is not yours to use?

BitonicEelis: Ok, makes sense!

src/wallet/rpcwallet.cpp
BitonicEelis: I'll remove the check for now. It can always be added later.

BitonicEelis:
Amended to:
- comply with variable naming policy
- reject attempts to lock already locked outputs
- remove MINE check

src/wallet/rpcwallet.cpp
```diff
@@ -2397,6 +2399,38 @@ UniValue lockunspent(const JSONRPCRequest& request)
 
         COutPoint outpt(uint256S(txid), nOutput);
 
+        const bool is_locked = pwallet->IsLockedCoin(outpt.hash, outpt.n);
+
+        if (fUnlock && !is_locked)
```
promag: Missing `{}` and `{` is one the same line. Same below. See [developer notes](https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md).

BitonicEelis: Will do! Thanks for caring about consistent style. :)

BitonicEelis:
Amended to use curly brackets in accordance with developer notes.

MeshCollider:
utACK https://github.com/bitcoin/bitcoin/pull/11087/commits/918b6db42b12d86a01bdb70a4fc50244aa2ce623

laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/11087/commits/918b6db42b12d86a01bdb70a4fc50244aa2ce623

test/functional/wallet.py
```diff
@@ -93,11 +93,17 @@ def run_test(self):
         # Exercise locking of unspent outputs
         unspent_0 = self.nodes[2].listunspent()[0]
         unspent_0 = {"txid": unspent_0["txid"], "vout": unspent_0["vout"]}
+        assert_raises_jsonrpc(-8, "expected locked output", self.nodes[2].lockunspent, True, [unspent_0])
```
jnewbery: Personally, I don't think the complete error message needs to be asserted, as long as the string is long enough to disambiguate. But also fine to change this to the complete string.

ryanofsky (commented):
utACK 918b6db42b12d86a01bdb70a4fc50244aa2ce623

You could modernize the code a little more: adding a range for loop instead of an index loop, and using UniValue directly instead of adding adding new find_value calls. Also I'm not sure if it's useful to make locking an already locked output or unlocked an already unlocked one errors. But overall, this change seems like an improvement and looks good to merge.

src/wallet/rpcwallet.cpp
```diff
@@ -2397,6 +2399,42 @@ UniValue lockunspent(const JSONRPCRequest& request)
...
+        if (fUnlock && !is_locked) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, expected locked output");
+        }
+
+        if (!fUnlock) {
+            if (is_locked) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, output already locked");
+            }
+
+            if (pwallet->IsSpent(outpt.hash, outpt.n)) {
```
jnewbery: Why are these three checks ("expected unspent", "unknown transaction" and "vout index out of bounds") only done when trying to lock? Isn't it equally invalid to try to unlock bad txouts? Can you place these three checks above the `const bool is_locked = pwallet->IsLockedCoin(outpt.hash, outpt.n);` line?

src/wallet/rpcwallet.cpp
```diff
@@ -2397,6 +2399,42 @@ UniValue lockunspent(const JSONRPCRequest& request)
...
+            const CWalletTx& trans = it->second;
+
+            if (outpt.n >= trans.tx->vout.size()) {
+                throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, vout index out of bounds");
+            }
+        }
+    }
+
+    // Atomically set (un)locked status for the outputs.
+    for (unsigned int idx = 0; idx < outputs.size(); idx++) {
```
jnewbery: You could save reparsing the json by saving a vector of `COutPoint`s in the first pass.

jnewbery (commented):
Looks good, and great test coverage. A couple of nits inline.

Maybe out of the scope of this PR, but this RPC returns `true` if called with `false` but no list of transactions. A user might expect that with that call all outputs have been locked, but in fact it's a no-op. I think that `if (request.params.size() == 1 && !fUnlock)` then we should return `false` or throw.

src/wallet/rpcwallet.cpp
Eelis: These loops don't do any json parsing, because the data is already in a UniValue. The `find_value` calls find the desired value in UniValue's internal std::vector of values.

src/wallet/rpcwallet.cpp
Eelis: The idea was that in all three of those cases (already spent output, unknown transaction, and vout index out of bounds), `IsLockedCoin` should already have returned false, so checking `!is_locked` should be enough.

src/wallet/rpcwallet.cpp
jnewbery: You're right. There's no reparsing the JSON, just rereading the string and reconstructing the COutPoint.

I think either way is fine. Whichever you prefer.

src/wallet/rpcwallet.cpp
jnewbery: I agree that the function will correctly return an error, but I think that it's better to return the more precise error if possible. `expected unspent output`, `unknown transaction` and `vout index out of bounds` are better to return than `expected locked output` (the last could infer that the outpoint exists but is unlocked).

Eelis: Makes sense, I'll make the errors more precise.

BitonicEelis:
Amended to:
- store COutPoints in vector instead of recomputing them
- include "invalid parameter:" error prefix in expectations in test
- make unlock errors more precise

@ryanofsky Unfortunately, range-for cannot be used to iterate over a UniValue because it does not provide begin()/end().

Thanks for the feedback!

src/wallet/rpcwallet.cpp
```diff
@@ -2389,33 +2389,72 @@ UniValue lockunspent(const JSONRPCRequest& request)
...
+
+    // Create and validate the COutPoints first.
+
+    std::vector<COutPoint> outputs;
+    outputs.reserve(output_params.size());
+
+    for (unsigned int idx = 0; idx < output_params.size(); idx++) {
+         const UniValue& param = output_params[idx];
+
+        if (!param.isObject()) {
```
jnewbery: You can remove this check entirely. Type checking is done by the `UniValue.get_xxx()` functions, and returns almost an identical error to that given by this explicit check.

With the explicit check:

```
→ bitcoin-cli -rpcwallet=w1 lockunspent true [\"string\"]
error code: -8
error message:
Invalid parameter, expected
```

without:

```
→ bitcoin-cli -rpcwallet=w1 lockunspent true [\"string\"]
error code: -1
error message:
JSON value is not an object as expected
```
The implicit `get_xxx()` check is used extensively in other RPCs.

You can also combine the line below with above:

```
const UniValue& o = request.params[1].get_obj();
```

src/wallet/rpcwallet.cpp
```diff
@@ -2389,33 +2389,72 @@ UniValue lockunspent(const JSONRPCRequest& request)
...
             throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, expected object");
-        const UniValue& o = output.get_obj();
+        }
+
+        const UniValue& o = param.get_obj();
 
         RPCTypeCheckObj(o,
```
jnewbery: In contrast to my comment above, it makes sense to keep these. `RPCTypeCheckObj` provides slightly more information (both the type expected and type received)

test/functional/wallet.py
```diff
@@ -144,6 +150,10 @@ def run_test(self):
         assert_equal(self.nodes[2].getbalance(), 94)
         assert_equal(self.nodes[2].getbalance("from1"), 94-21)
 
+        # Verify that a spent output cannot be locked anymore
+        spent_0 = {"txid": node0utxos[0]["txid"], "vout": node0utxos[0]["vout"]}
+        assert_raises_jsonrpc(-8, "Invalid parameter, expected unspent output", self.nodes[2].lockunspent, False, [spent_0])
```
jnewbery: This is the bug. `lockunspent()` should be called on `self.nodes[0]`, not `self.nodes[2]`

jnewbery (commented):
Looks great! One bug in the test code and one nit in the RPC method.

test/functional/wallet.py
```diff
@@ -101,11 +101,17 @@ def run_test(self):
...
         unspent_0 = {"txid": unspent_0["txid"], "vout": unspent_0["vout"]}
+        assert_raises_jsonrpc(-8, "Invalid parameter, expected locked output", self.nodes[2].lockunspent, True, [unspent_0])
         self.nodes[2].lockunspent(False, [unspent_0])
+        assert_raises_jsonrpc(-8, "Invalid parameter, output already locked", self.nodes[2].lockunspent, False, [unspent_0])
         assert_raises_jsonrpc(-4, "Insufficient funds", self.nodes[2].sendtoaddress, self.nodes[2].getnewaddress(), 20)
         assert_equal([unspent_0], self.nodes[2].listlockunspent())
         self.nodes[2].lockunspent(True, [unspent_0])
         assert_equal(len(self.nodes[2].listlockunspent()), 0)
+        assert_raises_jsonrpc(-8, "Invalid parameter, unknown transaction",
+          self.nodes[2].lockunspent, False, [{"txid": "0000000000000000000000000000000000", "vout": 0}])
```
jnewbery: I personally don't like this indentation style. It looks like something between a new line and a new code block.

Would you mind aligning this continuation line with the opening parens:

```python
    assert_raises_jsonrpc(-8, "Invalid parameter, unknown transaction",
                          self.nodes[2].lockunspent, False, [{"txid": "0000000000000000000000000000000000", "vout": 0}])
```

same below

test/functional/wallet.py
BitonicEelis: Thanks! This one stumped me because for some reason when I ran the wallet tests locally, they didn't fail..

BitonicEelis:
Amended to:

- remove isObject check
- indent continued function call lines to match opening parenthesis
- call lockunspent on correct node in test

test/functional/wallet.py
jnewbery: Yes! I saw the same thing. Very odd. h/t @sdaftuar for spotting the bug.

jnewbery:
Tested ACK fe2c95bda22c21252709cc1088d7e61a8c8b1df8. Looks great. Thanks for being receptive to all my feedback!

test/functional/wallet.py
```diff
@@ -144,6 +152,10 @@ def run_test(self):
         assert_equal(self.nodes[2].getbalance(), 94)
         assert_equal(self.nodes[2].getbalance("from1"), 94-21)
 
+        # Verify that a spent output cannot be locked anymore
+        spent_0 = {"txid": node0utxos[0]["txid"], "vout": node0utxos[0]["vout"]}
+        assert_raises_jsonrpc(-8, "Invalid parameter, expected unspent output", self.nodes[0].lockunspent, False, [spent_0])
```
MarcoFalke: Needs rebase and replace with "assert_raises_rpc_error"

src/wallet/rpcwallet.cpp
```diff
@@ -2389,33 +2389,66 @@ UniValue lockunspent(const JSONRPCRequest& request)
...
             {
                 {"txid", UniValueType(UniValue::VSTR)},
                 {"vout", UniValueType(UniValue::VNUM)},
             });
 
-        std::string txid = find_value(o, "txid").get_str();
-        if (!IsHex(txid))
+        const std::string txid = find_value(o, "txid").get_str();
```
ryanofsky: Could declare as `const std::string&` to avoid copying string.

ryanofsky (commented):
utACK fe2c95bda22c21252709cc1088d7e61a8c8b1df8. No significant changes since last review. Left minor comments (feel free to ignore them), but this does need to be rebased.

BitonicEelis:
Amended/rebased to:
- use `assert_raises_rpc_error`;
- use references to avoid copying a UniValue and std::string, as suggested by ryanofsky.

ryanofsky (commented):
utACK 28f8b6657764c7746645a6e75dfb09ffc0597322. Only changes since last review are rebase & adding const refs.

src/wallet/rpcwallet.cpp
```diff
@@ -2447,33 +2447,66 @@ UniValue lockunspent(const JSONRPCRequest& request)
...
+        if (!fUnlock && is_locked) {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "Invalid parameter, output already locked");
+        }
+
+        outputs.push_back(outpt);
+    }
+
+    // Atomically set (un)locked status for the outputs.
+    for (const COutPoint& outpt : outputs) {
+        if (fUnlock) pwallet->UnlockCoin(outpt);
```
sipa: Coding style nit; if with else always needs braces/indentation (I know the original code you're rewriting wasn't following this convention, but in new code we try to).

sipa (commented):
utACK 28f8b6657764c7746645a6e75dfb09ffc0597322

PR #11094, opened 2017-08-19
Docs: Hash in ZMQ hash is raw bytes, not hex

karelbilek:
Transaction hex cannot be in hexadecimal, that would be 64 bytes. In reality it's just raw bytes.


promag (approved):
ACK.

fanquake:
utACK 06a3aec

dcousens (approved):

laanwj:
utACK 4aa2508

PR #11095, opened 2017-08-19
Consensus: add const to temp int in base58.cpp

mdaiter:
Just caught my eye when looking over some code. Could be a const instead of a non-const.


promag (changes requested):
There are other variables that can be const all over the code, and in this file too.

You could improve the description and commit message.

As is NACK.

laanwj:
We don't regularly use `const` inside functions at all.
Maybe we should, but right now I don't think there is any reasoning why this value in base58 would be extra-important to warrant this change, sorry.

PR #7522, opened 2016-02-12
Bugfix: Only use git for build info if the repository is actually the right one

luke-jr:
Also adds ability to disable check with BITCOIN_GENBUILD_NO_GIT=1 in the environment

@theuni @Flowdalic @laanwj 

I believe this addresses all the possible use cases correctly.

(This commit is based on branch-0.10, so should merge cleanly into 0.10, 0.11, 0.12, and master branches.)

PR #11097, opened 2017-08-19
gitian: quick hack to fix version string in releases

theuni:
Credit: @luke-jr
Release version strings were broken in Gitian by #7522. This is a minimal fix suitable for 0.15.

After this, we should fix up version handling for good so that gitian packages the correct string in the release tarball, so that git is not required to get the tag name.

### DIFF
--- a/contrib/gitian-descriptors/gitian-linux.yml
+++ b/contrib/gitian-descriptors/gitian-linux.yml
@@ -132,7 +132,6 @@ script: |
   export PATH=${WRAP_DIR}:${PATH}
 
   # Create the release tarball using (arbitrarily) the first host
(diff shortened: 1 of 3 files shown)


luke-jr:
utACK

theuni:
I missed a removal of GIT_DIR in the osx descriptor. Pushed the fixed version, no other changes.

fanquake:
utACK 4452829. @jonasschnelli Can spin up some test gitian builds?

jonasschnelli:
Gitian build is here https://bitcoin.jonasschnelli.ch/build/267

laanwj:
utACK 4452829

Issue #11033, opened 2017-08-11
[QT] - Send-all Button


laudaa:
I think that a send-all button, which shouldn't be hard to implement, would be beneficial to the GUI. Generally, when people consolidate UTXO (for example) they select a number of inputs and then copy paste the full amount (or manually type it) into the amount field before sending. This is somewhat inconvenient, and other wallets (e.g. Copay, Mycelium) have this feature already.

Thanks.

jonasschnelli:
The current way (enter full available amount) and select "subtract fee from amount" is IMO relatively easy.
An extra button "send all" would do exactly that (type in the available balance and select the SFFA checkbox).
Could be handy though not sure...

laudaa:
The current way that it is set up is easy for sure, but it has an inconvenient and redundant step for a frequent action whilst the wallet should be aiming for ease of use/simplicity in that regard. The reasons for which I am suggesting it is that:
1) I find it inconvenient considering the number of times that I've done this, and I think others feel the same way.
2) I've noticed the feature implemented in other wallets (previously mentioned), and am quite fond of it.

aunyks:
@laudaa Just to be clear, this "Send-All Button" would completely empty the wallet right?

laudaa:
@aunyks If there were no inputs selected (coin control), then it would pull the total wallet balance. If there were inputs selected, then it would pull the total value of those inputs.

laanwj:
I'd like this feature too. I'm not sure what is exactly how you propose to offer this in the GUI but FWIW personally I'd prefer a button that fills in the total amount of all selected utxos in "amount" and auto-selects "subtract fee from amount" above an alternative send button.
This makes it immediately and intuitively clear what happens when the user clicks it.

laanwj:
The biggest challenge here is probably how to add this without creating an ugly wart big button in the GUI. Some subtle icon or such.

promag:
Concept ACK.

BTW should the same feature be available thru RPC? Like an alias for all instead of an amount?

promag:
> fills in the total amount of all selected utxos in "amount" and auto-selects "subtract fee from amount"

@laanwj that's not "atomic":
- can fail if meanwhile some utxo is spent
- can leave balance if meanwhile an utxo is new.

Should "send all button" prevent multiple recipients?

Or should the "send all button" be in each recipient and when activated it sets `recipient_amount = balance - sum_other_recipient_amounts` and check "subtract fee from amount"?

[...]

PR #11098, opened 2017-08-20
[Qt] Add spend all button to the SendCoinsDialog

CryptAxe:
Related to: https://github.com/bitcoin/bitcoin/issues/11033

### DIFF
--- a/src/qt/forms/sendcoinsdialog.ui
+++ b/src/qt/forms/sendcoinsdialog.ui
@@ -1224,6 +1224,13 @@
       </spacer>
      </item>
      <item>
+      <widget class="QPushButton" name="buttonSendAll">
+       <property name="text">
+        <string>Send All</string>
+       </property>
+      </widget>
+     </item>
+     <item>
       <layout class="QHBoxLayout" name="horizontalLayout_2">
        <property name="spacing">
         <number>3</number>
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -607,6 +607,34 @@ void SendCoinsDialog::on_buttonMinimizeFee_clicked()
     minimizeFeeSection(true);
 }
 
+void SendCoinsDialog::on_buttonSendAll_clicked()
+{
+    // Get CCoinControl instance if CoinControl is enabled or create a new one
+    CCoinControl coinControl;
+    if (model->getOptionsModel()->getCoinControlFeatures())
+        coinControl = *CoinControlDialog::coinControl;
+
+    // Calculate total amount to spend
+    CAmount nTotalAmount = model->getBalance(&coinControl);
+
+    // Calculate amount per entry
+    int nEntries = ui->entries->count();
+    CAmount nAmountPerEntry = (nEntries ? (nTotalAmount / nEntries) : nTotalAmount);
+
+    for(int i = 0; i < nEntries; ++i)
+    {
+        SendCoinsEntry *entry = qobject_cast<SendCoinsEntry*>(ui->entries->itemAt(i)->widget());
+        if(entry)
+        {
+            // Check subtract fee from amount checkbox
+            entry->checkSubtractFeeFromAmount();
+
+            // Set new amount for entry
+            entry->setAmount(nAmountPerEntry);
+        }
+    }
+}
+
 void SendCoinsDialog::setMinimumFee()
 {
     ui->radioCustomPerKilobyte->setChecked(true);
--- a/src/qt/sendcoinsdialog.h
+++ b/src/qt/sendcoinsdialog.h
@@ -75,6 +75,7 @@ private Q_SLOTS:
     void on_sendButton_clicked();
     void on_buttonChooseFee_clicked();
     void on_buttonMinimizeFee_clicked();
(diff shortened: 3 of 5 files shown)


MeshCollider:
I don't like the dividing equally between outputs behaviour, I think the button should only be possible on single entry spends, makes little sense to me otherwise. 
From discussion in #11033, @laanwj didn't just want a button added for appearance reasons.
Also nits, please update to snake_case and modify braces as per style guideline.

CryptAxe:
> I don't like the dividing equally between outputs behaviour, I think the button should only be possible on single entry spends, makes little sense to me otherwise.

Don't use the button with multiple outputs if you don't like it, no reason to stop other people from dividing their balance between multiple recipients. The way the SendCoinsDialog works right now assumes that there can be any number of recipients.

> From discussion in #11033, @laanwj didn't just want a button added for appearance reasons.

Feel free to suggest a better option. I put the button in an area so that it doesn't take up any vertical space and mess up the appearance of the entire page. It would be possible to put something on the SendCoinEntry(s) instead and send a signal to the SendCoinsDialog instance for example but that was more cluttered visually and in code when I tried it. 

> Also nits, please update to snake_case and modify braces as per style guideline.

You'll have to point out where you want me to use snake case...

promag:
> You'll have to point out where you want me to use snake case...

See [developer notes](https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md).

CryptAxe:
The best assumption that I can make from your response is that you want me to use snake case for variable names. I see that is in the developer notes now, but it is certainly not in line with most of the code that I see in Bitcoin core. Am I correct that we are now going to diverge from the style in the existing codebase, and require all new code to look different, creating a big mess of many different styles?

I don't know how much value a style guideline has besides maintaining clarity and uniformity of code, aren't we throwing that out the window here?

CryptAxe:
Why snake case, what led to that change? Who even made this decision and why would we diverge from the existing code in the first place? If this rule is followed we are making Bitcoin's code worse, it will be a mix of several styles (the new one being especially ugly in my opinion). The point of style guidelines are to maintain clarity and uniformity. There is no reason for Bitcoin to have style guidelines if they are going to change all the time. Do we really think 100% of the codebase is going to be re-written in the new style, or is it more likely that we will end up with an ugly mess as a result of this?

promag:
> Why snake case, what led to that change? Who even made this decision and why would we diverge from the existing code in the first place?

Some group of people did, git blame can point you in the right direction. But who cares, for me consistency wins.

> Do we really think 100% of the codebase is going to be re-written in the new style

Eventually.

> or is it more likely that we will end up with an ugly mess as a result of this?

I guess it is a mess because of the early lack of rules to follow and time to enforce those rules. Even nowadays, there is lot of new code that doesn't met those rules.

jonasschnelli:
Concept ACK.
I think the send all button needs to be in the send-to-box context (not in the global). Ideally close to where you enter the amount (it's mostly related to that IMO).

I think it should be `[amount] [denomination] [send-all] [subtract-from-amount]`.

Maybe we should call it not `send-all` because the max amount may change after you have pressed the button == you no longer sending "all". Maybe something like "select all", or "use max available" makes more sense.

bildschirmfoto 2017-09-07 um 10 28 10

CryptAxe:
Do you think a checkbox would be good for that? It looked a bit weird
having two of them next to eachother. I can't come up with anything
prettier besides a QLable with an icon people can click on, but we would
need to add an icon.

On Sep 7, 2017 10:33 AM, "Jonas Schnelli" <notifications@github.com> wrote:

Concept ACK.
I think the send all button needs to be in the send-to-box context (not in
the global). Ideally close to where you enter the amount (it's mostly
related to that IMO).

I think it should be [amount] [denomination] [send-all]
[subtract-from-amount].

Maybe we should call it not send-all because the max amount may change
after you have pressed the button == you no longer sending "all". Maybe
something like "select all", or "use max available" makes more sense.

[image: bildschirmfoto 2017-09-07 um 10 28 10]
<https://user-images.githubusercontent.com/178464/30176575-c43a67f6-93b7-11e7-8c2c-ec8b2156f820.png>

—
<https://github.com/bitcoin/bitcoin/pull/11098#issuecomment-327869628>, or mute
the thread

MeshCollider:
What's wrong with using a button like you're currently doing? Why are you talking about a checkbox?

CryptAxe:
Well I can but adding a push button where Jonas has the arrow pointing
would take up more vertical space and mess up the layout a bit. Should I
make some screenshots showing different options?

On Sep 7, 2017 7:37 PM, "MeshCollider" <notifications@github.com> wrote:

What's wrong with using a button like you're currently doing? Why are you
talking about a checkbox?

—
<https://github.com/bitcoin/bitcoin/pull/11098#issuecomment-327983132>, or mute
the thread

promag:
> Maybe something like "select all", or "use max available" makes more sense

Agree with @jonasschnelli, and that "button" should fill the amount field and toggle the checkbox. And I think it works well with multiple outputs: So when pressed the input will have the available amount minus other input amounts.

CryptAxe:
https://github.com/bitcoin/bitcoin/pull/11316 which improved this was merged so I'll close it.

Issue #11086, opened 2017-08-18
Autosave mempool to disk while running


shivaenigma:
Hi,

The feature to save mempool to disk on shutdown is great! This feature however, doesn't help where it could be most useful ie.. crashing/unplanned kill of bitcoind process. I had my bitcoin node crash after 30 days of running and the mempool loaded from disk was ofcourse of no use!

This is to request:

- Frequent autosaving of mempool to disk 
_AND/OR_
- RPC command to trigger save of mempool to disk

laanwj:
> Frequent autosaving of mempool to disk

This was part of the initial PR #8448 that implemented mempool saving to disk, but was deemed too much of an I/O overhead by default. It'd be okay as an optional setting.

> RPC command to trigger save of mempool to disk

Yeah or that...

gmaxwell:
I think RPC would be better.  Then you can just cron it to run whenever you want... 

I think the original PR got cut back more than it needed to just to deflect the wall of criticism being thrown at it.

As an aside, if your node is crashing we'd like to know about it, as that shouldn't happen!

shivaenigma:
@gmaxwell I will check if I can get something from the logs about the crash (it is a rare event though) !

PR #11099, opened 2017-08-20
[RPC][mempool]: Add savemempool RPC

greenaddress:
Adds a simple parameterless rpc command to dump the mempool.

Rationale:

Sometimes there can be a crash for whatever reason (bug, power loss, etc) causing the mempool.dat file to not be saved.

This change allows to script/cron the rpc call to have more regular saves to the file as well as cli/ad-hoc.

This should solve issue https://github.com/bitcoin/bitcoin/issues/11086

### DIFF
--- a/src/rpc/blockchain.cpp
+++ b/src/rpc/blockchain.cpp
@@ -1532,6 +1532,25 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
     return ret;
 }
 
+UniValue savemempool(const JSONRPCRequest& request)
+{
+    if (request.fHelp || request.params.size() != 0) {
+        throw std::runtime_error(
+            "savemempool\n"
+            "\nDumps the mempool to disk.\n"
+            "\nExamples:\n"
+            + HelpExampleCli("savemempool", "")
+            + HelpExampleRpc("savemempool", "")
+        );
+    }
+
+    if (!DumpMempool()) {
+        throw JSONRPCError(RPC_MISC_ERROR, "Unable to dump mempool to disk");
+    }
+
+    return NullUniValue;
+}
+
 static const CRPCCommand commands[] =
 { //  category              name                      actor (function)         argNames
   //  --------------------- ------------------------  -----------------------  ----------
@@ -1552,6 +1571,7 @@ static const CRPCCommand commands[] =
     { "blockchain",         "gettxout",               &gettxout,               {"txid","n","include_mempool"} },
     { "blockchain",         "gettxoutsetinfo",        &gettxoutsetinfo,        {} },
     { "blockchain",         "pruneblockchain",        &pruneblockchain,        {"height"} },
+    { "blockchain",         "savemempool",            &savemempool,            {} },
     { "blockchain",         "verifychain",            &verifychain,            {"checklevel","nblocks"} },
 
     { "blockchain",         "preciousblock",          &preciousblock,          {"blockhash"} },
--- a/test/functional/mempool_persist.py
+++ b/test/functional/mempool_persist.py
@@ -28,8 +28,14 @@ Test is as follows:
   - Restart node0 with -persistmempool. Verify that it has 5
     transactions in its mempool. This tests that -persistmempool=0
     does not overwrite a previously valid mempool stored on disk.
+  - Remove node0 mempool.dat and verify savemempool RPC recreates it
+    and verify that node1 can load it and has 5 transaction in its
+    mempool.
+  - Verify that savemempool throws when the RPC is called if
+    node1 can't write to disk.
 
 """
+import os
 import time
 
 from test_framework.test_framework import BitcoinTestFramework
@@ -77,6 +83,28 @@ class MempoolPersistTest(BitcoinTestFramework):
         self.stop_nodes()
         self.start_node(0)
         wait_until(lambda: len(self.nodes[0].getrawmempool()) == 5)
+
+        mempooldat0 = os.path.join(self.options.tmpdir, 'node0', 'regtest', 'mempool.dat')
+        mempooldat1 = os.path.join(self.options.tmpdir, 'node1', 'regtest', 'mempool.dat')
+        self.log.debug("Remove the mempool.dat file. Verify that savemempool to disk via RPC re-creates it")
+        os.remove(mempooldat0)
+        self.nodes[0].savemempool()
+        assert os.path.isfile(mempooldat0)
+
+        self.log.debug("Stop nodes, make node1 use mempool.dat from node0. Verify it has 5 transactions")
+        os.rename(mempooldat0, mempooldat1)
+        self.stop_nodes()
+        self.start_node(1, extra_args=[])
+        wait_until(lambda: len(self.nodes[1].getrawmempool()) == 5)
+
+        self.log.debug("Prevent bitcoind from writing mempool.dat to disk. Verify that `savemempool` fails")
+        # to test the exception we are setting bad permissions on a tmp file called mempool.dat.new
+        # which is an implementation detail that could change and break this test
+        mempooldotnew1 = mempooldat1 + '.new'
+        with os.fdopen(os.open(mempooldotnew1, os.O_CREAT, 0o000), 'w'):
+            pass
+        assert_raises_jsonrpc(-1, "Unable to dump mempool to disk", self.nodes[1].savemempool)
+        os.remove(mempooldotnew1)
 
 if __name__ == '__main__':
     MempoolPersistTest().main()
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -4332,7 +4332,7 @@ bool LoadMempool(void)
     return true;
 }
 
-void DumpMempool(void)
+bool DumpMempool(void)
 {
     int64_t start = GetTimeMicros();
 
@@ -4352,7 +4352,7 @@ void DumpMempool(void)
     try {
         FILE* filestr = fsbridge::fopen(GetDataDir() / "mempool.dat.new", "wb");
         if (!filestr) {
-            return;
+            return false;
         }
 
         CAutoFile file(filestr, SER_DISK, CLIENT_VERSION);
@@ -4376,7 +4376,9 @@ void DumpMempool(void)
         LogPrintf("Dumped mempool: %gs to copy, %gs to dump\n", (mid-start)*MICRO, (last-mid)*MICRO);
     } catch (const std::exception& e) {
         LogPrintf("Failed to dump mempool: %s. Continuing anyway.\n", e.what());
+        return false;
     }
+    return true;
 }
 
 //! Guess how far we are in the verification process at the given block index
--- a/src/validation.h
+++ b/src/validation.h
@@ -475,7 +475,7 @@ static const unsigned int REJECT_HIGHFEE = 0x100;
 CBlockFileInfo* GetBlockFileInfo(size_t n);
 
 /** Dump the mempool to disk. */
-void DumpMempool();
+bool DumpMempool();
 
 /** Load the mempool from disk. */
 bool LoadMempool();


src/rpc/blockchain.cpp
```diff
@@ -1531,6 +1531,21 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+        throw std::runtime_error(
+            "dumpmempool\n"
+            "\nDumps to disk the mempool (mempool.dat).\n"
+            "\nExamples:\n"
+            + HelpExampleCli("dumpmempool", "")
+            + HelpExampleRpc("dumpmempool", "")
+        );
+
+    DumpMempool();
+    return "Mempool dumped";
```
promag: This is not correct since it can fail.

src/rpc/blockchain.cpp
```diff
@@ -1531,6 +1531,21 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
     return ret;
 }
 
+UniValue dumpmempool(const JSONRPCRequest& request)
+{
+    if (request.fHelp || request.params.size() > 0)
+        throw std::runtime_error(
+            "dumpmempool\n"
+            "\nDumps to disk the mempool (mempool.dat).\n"
```
promag: Remove the filename?

promag (commented):
Please add a test.

The function can be changed to know if it failed or not.

promag:
Suggestion, reword `Add dumpmempool RPC`.

greenaddress:
@promag thanks, updated as for feedback and squashed

src/rpc/blockchain.cpp
```diff
@@ -1552,6 +1568,7 @@ static const CRPCCommand commands[] =
     { "blockchain",         "gettxoutsetinfo",        &gettxoutsetinfo,        true,  {} },
     { "blockchain",         "pruneblockchain",        &pruneblockchain,        true,  {"height"} },
     { "blockchain",         "verifychain",            &verifychain,            true,  {"checklevel","nblocks"} },
+    { "blockchain",         "dumpmempool",            &dumpmempool,            true,  {} },
```
promag: Sort, move after `gettxoutsetinfo`.

src/rpc/blockchain.cpp
```diff
@@ -1531,6 +1531,22 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+    if (request.fHelp || request.params.size() > 0)
+        throw std::runtime_error(
+            "dumpmempool\n"
+            "\nDumps to disk the mempool.\n"
+            "\nExamples:\n"
+            + HelpExampleCli("dumpmempool", "")
+            + HelpExampleRpc("dumpmempool", "")
+        );
+
+    if (!DumpMempool())
```
promag: Missing `{ }`.

src/rpc/blockchain.cpp
```diff
@@ -1531,6 +1531,22 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+        throw JSONRPCError(RPC_INTERNAL_ERROR, "Unable to dump mempool to disk");
+    return "Mempool dumped";
```
promag: Nit, new line before `return`.

promag (commented):
I wonder if it should throw `RPC_INTERNAL_ERROR`, since `DumpMempool` handles it's errors.

Please split in two commits:
- `Add return value to DumpMempool`
- `Add dumpmempool RPC`.

greenaddress:
@promag thanks, updated as for feedback and split in two commits

re: RPC_INTERNAL_ERROR - didn't immediately found something better, very open to suggestions

Anything else you would add to the test?

promag:
I meant maybe not throw, and have the bool in the response.

greenaddress:
I think the condition is exceptional and deserve a throw of an exception & consistent with the other rpcs (though I don't know if there's a more appropriate exception for it)

promag:
I almost agree with that if it wasn't the fact that the function doesn't throw.. but I have no strong opinion on that.

If the throw stays, then maybe `RPC_MISC_ERROR`. See what causes `RPC_INTERNAL_ERROR`, sound like really bad excpetions.

src/rpc/blockchain.cpp
```diff
@@ -1531,9 +1531,28 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
     return ret;
 }
 
+UniValue dumpmempool(const JSONRPCRequest& request)
+{
+    if (request.fHelp || request.params.size() > 0)
+        throw std::runtime_error(
+            "dumpmempool\n"
+            "\nDumps to disk the mempool.\n"
```
jonasschnelli: I'm not the one to complain about this: but should it be `"Dumps the mempool to disk"?

src/rpc/blockchain.cpp
```diff
@@ -1531,9 +1531,28 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+            "\nExamples:\n"
+            + HelpExampleCli("dumpmempool", "")
+            + HelpExampleRpc("dumpmempool", "")
+        );
+
+    if (!DumpMempool()) {
+        throw JSONRPCError(RPC_INTERNAL_ERROR, "Unable to dump mempool to disk");
+    }
+
+    return "Mempool dumped";
```
jonasschnelli: Because this is machine to machine communication, use `true` as ret val at this point?

jonasschnelli (changes requested):
Concept ACK

greenaddress:
thanks @jonasschnelli - good points

added a commit (which i think should be squashed at some point together with "Add dumpmempool RPC")

src/rpc/blockchain.cpp
```diff
@@ -1531,9 +1531,28 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+        throw std::runtime_error(
+            "dumpmempool\n"
+            "\nDumps the mempool to disk.\n"
+            "\nExamples:\n"
+            + HelpExampleCli("dumpmempool", "")
+            + HelpExampleRpc("dumpmempool", "")
+        );
+
+    if (!DumpMempool()) {
+        throw JSONRPCError(RPC_INTERNAL_ERROR, "Unable to dump mempool to disk");
```
promag: Maybe `RPC_MISC_ERROR`. See what causes `RPC_INTERNAL_ERROR`, sounds like really bad exceptions.

jonasschnelli:
utACK cd0ea487422028bec1f5df62ab4c57909c2bcc90

Maybe add an rational (possible use-cases) to the PR description.

greenaddress:
@promag updated the exception, thanks

@jonasschnelli updated the description. Please note your utACK points to a commit from @TheBlueMatt. GitHub also tells me there's some changes required, should I ignore that?

test/functional/mempool_dump.py
```diff
@@ -0,0 +1,43 @@
...
+class MempoolDumpTest(BitcoinTestFramework):
+
+    def __init__(self):
+        super().__init__()
+        self.num_nodes = 1
+        self.setup_clean_chain = True
+
+    def run_test(self):
+        self.log.debug("Verify mempool file doesn't exists")
+        mempooldat = self.options.tmpdir + "/node0/regtest/mempool.dat"
```
MarcoFalke: Use `os.path.join`, so that it works on other operating systems as well.

test/functional/mempool_dump.py
```diff
@@ -0,0 +1,43 @@
...
+        assert_raises(OSError, lambda: os.remove(mempooldat))
```
MarcoFalke: Might use `os.path.isfile` for better readability.

test/functional/mempool_dump.py
```diff
@@ -0,0 +1,43 @@
...
+        self.log.debug("Dump mempool to disk")
+        assert_equal(self.nodes[0].dumpmempool(), True)
```
MarcoFalke: nit: No need for assert_equal, you can just `assert dumpmempool()`

test/functional/mempool_dump.py
```diff
@@ -0,0 +1,43 @@
...
+        assert_equal(os.path.isfile(mempooldat), True)
+        os.rename(self.options.tmpdir + "/node0/regtest", self.options.tmpdir + "/node0/regtest_bk")
```
MarcoFalke: I don't think we allow the datadir to be moved while running. The tests should not rely on this.

MarcoFalke: You might be able to achieve the same error condition by changing the file permissions, though.

maflcko (commented):
Concept ACK, but I don't like the tests.

MarcoFalke:
You might even combine the tests with the existing mempool persist test.

greenaddress:
@MarcoFalke I tried with permissions and didn't work. I couldn't find another way to trigger the error.

Will address the other feedback now

test/functional/mempool_dump.py
```diff
@@ -0,0 +1,45 @@
...
+
+    def run_test(self):
+        self.log.debug("Verify mempool file doesn't exists")
+        regtestdir = os.path.join(self.options.tmpdir, 'node0', 'regtest')
+        mempooldat = os.path.join(regtestdir, 'mempool.dat')
+        assert not os.path.isfile(mempooldat)
+        self.log.debug("Dump mempool to disk")
+        assert self.nodes[0].dumpmempool()
+        assert os.path.isfile(mempooldat)
+        moveddir = os.path.join(self.options.tmpdir, 'node0', 'regtest_bk')
```
jnewbery: I'm not sure if this failure test is necessary. If bitcoind can't write to its data directory, I'm sure all kinds of bad things happen.

jnewbery:
> If the throw stays, then maybe RPC_MISC_ERROR. See what causes RPC_INTERNAL_ERROR, sound like really bad excpetions.

I think either is fine. bitcoind should always be able to write to its data directory, so I would say failure to do so counts as an internal error.

> Concept ACK, but I don't like the tests.

I also don't like the test framework messing with bitcoind's data directory. I don't think the failure testcase is necessary. If bitcoind can't write to its datadirectory, there are more pressing issues than the RPC raising the correct error.

Agree with @MarcoFalke that it might be a good idea to combine this with the mempool_persist.py test (if it can be done tidily).

test/functional/mempool_dump.py
greenaddress: I think it makes sense to check that there is no file, you dump and there is a file.

greenaddress:
@jnewbery thanks, addressed some feedback. 

Re: testing I also don't like messing with the datadir but unless someone has a better suggestions the options are keeping it or removing it.

jnewbery:
> the options are keeping it or removing it.

I vote for removing it

promag:
> bitcoind should always be able to write to its data directory

@jnewbery `DumpMempool()` ignores failures, treats them irrelevant, hence my initial point of removing the throw in the RPC.

test/functional/mempool_dump.py
MarcoFalke: >  I think it makes sense to check that there is no file, you dump and there is a file.

You do exactly this in the lines above. The comment is about `assert_raises_jsonrpc`, which raises due to the missing datadir. As mentioned earlier, this will likely introduce hard to reproduce test failures now or in the future.

With the current flakiness of the test framework in mind I am against adding such a failure test.

MarcoFalke: What you might try instead is to copy the dumped mempool from node0 to node1 and check that it can be loaded. (The nodes are not connected, but share the same blockchain)

Anyway, I just noticed you are using a clean blockchain and dump the empty mempool. Might be better to populate it a bit.

promag: A bit as in 1 transaction is enough? 😄

greenaddress: @MarcoFalke I made that change (about to push commit) and moved the test to mempool_persist as advised - however your suggestion still doesn't test the exception so I left that in, can easily remove the last 4 lines of the test if other agrees is better to not have it.

greenaddress:
@MarcoFalke I've merged the tests and added your suggestion "copy the dumped mempool from node0 to node1 and check that it can be loaded. (The nodes are not connected, but share the same blockchain)"

I left in the exception test for now.

Is now a good time to squash? shall I also rebase?

greenaddress:
Updated slightly the tests only: removed the ugly rename of the node directory in favor of using permissions to make it fail in the test

@MarcoFalke Permissions initially didn't work because there was a rename involved but it works if I use permissions on the file that gets renamed ('mempool.dat.new') so switched to that

I also removed from the test some redundant asserts re: os.remove doing its job (it already throws if it can't find a file or fails to remove it)

sipa:
utACK d0bfa6a11a66a8c08fe2630472546d0e60175fe3

test/functional/mempool_persist.py
```diff
@@ -85,5 +91,29 @@ def run_test(self):
         self.nodes.append(self.start_node(0, self.options.tmpdir))
         wait_until(lambda: len(self.nodes[0].getrawmempool()) == 5)
 
+        regtestdir0 = os.path.join(self.options.tmpdir, 'node0', 'regtest')
```
jnewbery: Intermediate `regtestdir0` variable not required. `os.path.join()` can take any number of arguments.

Same for `regtestdir1` below.

test/functional/mempool_persist.py
```diff
@@ -85,5 +91,29 @@ def run_test(self):
...
+        mempooldat0 = os.path.join(regtestdir0, 'mempool.dat')
+        regtestdir1 = os.path.join(self.options.tmpdir, 'node1', 'regtest')
+        mempooldat1 = os.path.join(regtestdir1, 'mempool.dat')
+        self.log.debug("Remove the mempool.dat file. Verify that dumpmempool to disk via RPC re-creates it")
```
jnewbery: Supernit: I like logging in the test script to be 'info' level, so when running the test locally, progress can be seen on the console by default.

test/functional/mempool_persist.py
```diff
@@ -85,5 +91,29 @@ def run_test(self):
...
+        os.remove(mempooldat0)
+        assert self.nodes[0].dumpmempool()
+        assert os.path.isfile(mempooldat0)
+
+        self.log.debug("Stop nodes, make node1 use mempool.dat from node0. Verify it has 5 transactions")
+        self.stop_nodes()
+        os.rename(mempooldat0, mempooldat1)
+        self.nodes.append(self.start_node(1, self.options.tmpdir))
+        # Give bitcoind a second to reload the mempool
+        time.sleep(1)
```
jnewbery: No need for `time.sleep()` if using `wait_until()` below.

jnewbery (commented):
I much, much prefer using file permissions on the `mempool.dat` file instead of deleting directories.

A few nits in the testcase, but otherwise looks great. Thanks for indulging all of my review comments :)

test/functional/mempool_persist.py
greenaddress: Makes sense. Will submit a fix for it

test/functional/mempool_persist.py
greenaddress: All the other logs in the test are debug. Either we change them all or I don't think it makes sense for me to just change mine.

test/functional/mempool_persist.py
greenaddress: I thought so too (and indeed it works without time.sleep()) - however the test has that as a pattern and I think we either remove it for the rest of the calls (at least within the same test) or it's inconsistent.

greenaddress:
@jnewbery thanks! Keep in mind we are not doing permissions on the mempool.dat file as that doesn't work (the dump gets done on a tmp file which then replace the mempool.dat and ignores permissions so we are acting on the tmp file instead)

test/functional/mempool_persist.py
jnewbery: I'm all for trying to match local style, but there's no benefit in copying bad patterns.

The `time.sleep` in L76 should be removed (and the `assert_equal()` in L78 replaced with a `wait_until()`) in a different PR.

The `time.sleep` in L85 serves a purpose - the mempool takes a while to load so we want to give it time before asserting it's empty. There's probably a better way of testing this, but we can't just replace the `assert_equal()` with a `wait_until()`.

jnewbery:
> Keep in mind we are not doing permissions on the mempool.dat file as that doesn't work (the dump gets done on a tmp file which then replace the mempool.dat and ignores permissions so we are acting on the tmp file instead)

Ah. Very crafty! Can you add a comment to the test to explain that (`mempool.dat.new` is an implementation detail - it could potentially be changed in future, breaking this test, so a comment would be helpful)

greenaddress:
@jnewbery I added the comment and removed the sleep. 
I didn't change the logging debug vs info as that imho should be done in a separate PR for all the log statements in the file (i.e. better to be consistent for now)

jnewbery:
New commits look great. Can you squash?

greenaddress:
@jnewbery done, didn't rebase however happy to if needed

deno49:
strictly q

test/functional/mempool_persist.py
```diff
@@ -85,5 +91,27 @@ def run_test(self):
...
         wait_until(lambda: len(self.nodes[0].getrawmempool()) == 5)
 
+        mempooldat0 = os.path.join(self.options.tmpdir, 'node0', 'regtest', 'mempool.dat')
+        mempooldat1 = os.path.join(self.options.tmpdir, 'node1', 'regtest', 'mempool.dat')
+        self.log.debug("Remove the mempool.dat file. Verify that dumpmempool to disk via RPC re-creates it")
+        os.remove(mempooldat0)
+        assert self.nodes[0].dumpmempool()
+        assert os.path.isfile(mempooldat0)
+
+        self.log.debug("Stop nodes, make node1 use mempool.dat from node0. Verify it has 5 transactions")
```
jnewbery: I think the intent of this section of the tests is to verify that the dumped mempool from the `dumpmempool` RPC contains all the transactions. Correct? The way it's written right now doesn't test this. The act of stopping node0 causes it to dump its mempool on shutdown, and hence this test is just a repeat of the test on line 77. I think you need to:

- run `dumpmempool` on a running node
- move that mempool.dat to a temp location
- shutdown the node
- move the mempool.dat back to the correct location
- start the node
- check that its mempool contains the correct transactions

test/functional/mempool_persist.py
```diff
@@ -85,5 +91,27 @@ def run_test(self):
...
+        self.stop_nodes()
+        os.rename(mempooldat0, mempooldat1)
+        self.nodes.append(self.start_node(1, self.options.tmpdir))
+        wait_until(lambda: len(self.nodes[1].getrawmempool()) == 5)
+
+        self.log.debug("Force core to fail to save file to disk. Check it errors.")
```
jnewbery: nit: debug message could be clearer. Perhaps:

```
"Prevent bitcoind from writing mempool.dat to disk. Verify that `dumpmempool` fails"
```

jnewbery (commented):
Looks really good. Just a couple more nits (sorry!)

jnewbery:
Needed slight rebase changes after #11121 (apologies - I keep making changes to the test_framework API).

I've done that and pushed here: https://github.com/jnewbery/bitcoin/tree/pr11099 . Feel free to reset your branch onto that commit.

greenaddress:
@jnewbery thank you! I've updated the PR with your rebase, no need to apologize especially as you did all the work :)

jnewbery:
Tested ACK 67d307f606f54a01b8dd6d79ca45d08171d20ccc. Looks great. Thanks for sticking with this.

Expect Travis to fail (for this and all other PRs) until #11215 is merged.

src/rpc/blockchain.cpp
```diff
@@ -1532,9 +1532,28 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
     return ret;
 }
 
+UniValue dumpmempool(const JSONRPCRequest& request)
+{
+    if (request.fHelp || request.params.size() != 0)
```
promag: Nit, add braces.

promag (commented):
There is only one test for `dumpmempool` failure, which is the `can't write to mempool.dat.new`. However there is another `return false;` inside the `catch` (which I can't understand why) that should be tested.

Also, `RenameOver()` doesn't throw, and IMO we should fail if the `mempool.dat` can't be updated.

MeshCollider:
utACK https://github.com/bitcoin/bitcoin/commit/67d307f606f54a01b8dd6d79ca45d08171d20ccc

luke-jr:
Suggest renaming it to `savemempool` or something.

The name `dumpmempool` to me suggested the same outcome as `getrawmempool`...

greenaddress:
@luke-jr changed the name, savemempool sounds better to me too.

@promag added brackets.

didn't squash them yet but happy to when people are happy with the changes

jonasschnelli (approved):
utACK 8ef87d253c88c2e19e875f9fec601159fb6da32f

jnewbery:
utACK 8ef87d253c88c2e19e875f9fec601159fb6da32f

MarcoFalke:
@greenaddress Will review after rebase

src/rpc/blockchain.cpp
```diff
@@ -1532,6 +1532,25 @@ UniValue getchaintxstats(const JSONRPCRequest& request)
...
+            + HelpExampleCli("savemempool", "")
+            + HelpExampleRpc("savemempool", "")
+        );
+    }
+
+    if (!DumpMempool()) {
+        throw JSONRPCError(RPC_MISC_ERROR, "Unable to dump mempool to disk");
+    }
+
+    return true;
```
MarcoFalke: This return value is not documented. And I think we shouldn't return anything on this rpc.

promag: Right, since on failure an exception is raised, on success it can be void.

test/functional/mempool_persist.py
```diff
@@ -86,9 +86,9 @@ def run_test(self):
 
         mempooldat0 = os.path.join(self.options.tmpdir, 'node0', 'regtest', 'mempool.dat')
         mempooldat1 = os.path.join(self.options.tmpdir, 'node1', 'regtest', 'mempool.dat')
-        self.log.debug("Remove the mempool.dat file. Verify that dumpmempool to disk via RPC re-creates it")
+        self.log.debug("Remove the mempool.dat file. Verify that savemempool to disk via RPC re-creates it")
         os.remove(mempooldat0)
-        assert self.nodes[0].dumpmempool()
+        assert self.nodes[0].savemempool()
```
promag: Remove assert.

src/rpc/blockchain.cpp
promag: Right, since on failure an exception is raised, on success it can be void.

test/functional/mempool_persist.py
```diff
@@ -78,5 +84,27 @@ def run_test(self):
         self.start_node(0)
         wait_until(lambda: len(self.nodes[0].getrawmempool()) == 5)
 
+        mempooldat0 = os.path.join(self.options.tmpdir, 'node0', 'regtest', 'mempool.dat')
+        mempooldat1 = os.path.join(self.options.tmpdir, 'node1', 'regtest', 'mempool.dat')
+        self.log.debug("Remove the mempool.dat file. Verify that savemempool to disk via RPC re-creates it")
+        os.remove(mempooldat0)
+        assert self.nodes[0].savemempool()
```
promag: No need to assert.

promag (commented):
Please see my above [comment](https://github.com/bitcoin/bitcoin/pull/11099#pullrequestreview-60254397).

src/rpc/blockchain.cpp
jnewbery: Yes, it appears that many RPCs that perform an action which is expected to succeed return `NullUniValue` on success. See several of the methods in src/rpc/net.cpp for example.

greenaddress:
rebased & squashed  and reworded the second commit to the new `savemempool` name - a lot of thanks for all feedback

@promag @jnewbery @MarcoFalke I changed the return value to NullUniValue (and accordingly updated the test)

promag:
utACK 1aa97ee.

jnewbery:
utACK 1aa97ee088ea03dd208be054c5ad9198c1f13329

MarcoFalke:
@greenaddress Thanks for sticking with the pull. Quite a pain to go through so much review iterations.

promag:
> Quite a pain to go through so much review iterations.

You mean fun? :trollface:

PR #11110, opened 2017-08-22
[script] Avoid implicit casts from bool to CScriptNum

practicalswift:
Avoid implicit casts from `bool` to `CScriptNum`.

Rationale: Explicit is better than implicit.

### DIFF
--- a/src/script/interpreter.cpp
+++ b/src/script/interpreter.cpp
@@ -756,8 +756,8 @@ bool EvalScript(std::vector<std::vector<unsigned char> >& stack, const CScript& 
                     case OP_1SUB:       bn -= bnOne; break;
                     case OP_NEGATE:     bn = -bn; break;
                     case OP_ABS:        if (bn < bnZero) bn = -bn; break;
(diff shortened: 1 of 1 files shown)


promag:
There is no implicit cast here, instead it uses `CScriptNum& operator=( const int64_t& rhs)`.

For that reason NACK as is.

practicalswift:
@promag The conversion from `bool` is clearly implicit, isn't it?

promag:
@practicalswift right, sorry, late night I thought you meant implicit construction.

ryanofsky (commented):
utACK 350f5b30533eebaf27d9a63c3d8d0bea3df3a291

Seems like a fine change, does what is described, don't really think it's good or bad.

laanwj:
I really don't like doing unnecessary changes in consensus code. The implicit case from bool to int64 is fully defined behavior.

IMO, this also, doesn't make the code easier to understand, just longer.

Sorry, going to close this. If there is overwhelming agreement to do this we should reopen, but otherwise, meh.

practicalswift:
OK :-)

PR #10969, opened 2017-08-01
Declare single-argument (non-converting) constructors "explicit"

practicalswift:
Declare single-argument (non-converting) constructors `explicit`.

In order to avoid unintended implicit conversions.

For a more thorough discussion, see ["C.46: By default, declare single-argument constructors explicit"](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c46-by-default-declare-single-argument-constructors-explicit) in the C++ Core Guidelines (Stroustrup & Sutter).

PR #11112, opened 2017-08-22
[developer-notes] By default, declare single-argument constructors "explicit"

practicalswift:
This is a follow-up to the now merged #10969.

Add recommendation:

> By default, declare single-argument constructors `explicit`.
> 
> - *Rationale*: This is a precaution to avoid unintended conversions that might arise when single-argument constructors are used as implicit conversion functions.
>


laanwj:
ACK f1708ef89a38efe857daeb4252c6dfab5b7c258e

MarcoFalke:
ACK f1708ef

sipa:
ACK

fanquake:
ACK f1708ef

PR #11116, opened 2017-08-23
[script] Unit tests for script/standard and IsMine functions.

jimpo:
Simply adding unit test coverage.

### DIFF
--- a/src/script/ismine.cpp
+++ b/src/script/ismine.cpp
@@ -46,6 +46,8 @@ isminetype IsMine(const CKeyStore &keystore, const CTxDestination& dest, bool& i
 
 isminetype IsMine(const CKeyStore &keystore, const CScript& scriptPubKey, bool& isInvalid, SigVersion sigversion)
 {
+    isInvalid = false;
+
     std::vector<valtype> vSolutions;
     txnouttype whichType;
     if (!Solver(scriptPubKey, whichType, vSolutions)) {
--- a/src/test/script_standard_tests.cpp
+++ b/src/test/script_standard_tests.cpp
@@ -3,6 +3,8 @@
 // file COPYING or http://www.opensource.org/licenses/mit-license.php.
 
 #include "key.h"
+#include "keystore.h"
+#include "script/ismine.h"
 #include "script/script.h"
 #include "script/script_error.h"
 #include "script/standard.h"
@@ -81,7 +83,6 @@ BOOST_AUTO_TEST_CASE(script_standard_Solver_success)
     BOOST_CHECK(solutions[4] == std::vector<unsigned char>({3}));
 
     // TX_NULL_DATA
-    solutions.clear();
     s.clear();
     s << OP_RETURN <<
         std::vector<unsigned char>({0}) <<
@@ -92,7 +93,6 @@ BOOST_AUTO_TEST_CASE(script_standard_Solver_success)
     BOOST_CHECK_EQUAL(solutions.size(), 0);
 
     // TX_WITNESS_V0_KEYHASH
-    solutions.clear();
     s.clear();
     s << OP_0 << ToByteVector(pubkeys[0].GetID());
     BOOST_CHECK(Solver(s, whichType, solutions));
@@ -102,9 +102,9 @@ BOOST_AUTO_TEST_CASE(script_standard_Solver_success)
 
     // TX_WITNESS_V0_SCRIPTHASH
     uint256 scriptHash;
-    CSHA256().Write(&redeemScript[0], redeemScript.size()).Finalize(scriptHash.begin());
+    CSHA256().Write(&redeemScript[0], redeemScript.size())
+        .Finalize(scriptHash.begin());
 
-    solutions.clear();
     s.clear();
     s << OP_0 << ToByteVector(scriptHash);
     BOOST_CHECK(Solver(s, whichType, solutions));
@@ -113,7 +113,6 @@ BOOST_AUTO_TEST_CASE(script_standard_Solver_success)
     BOOST_CHECK(solutions[0] == ToByteVector(scriptHash));
 
     // TX_NONSTANDARD
-    solutions.clear();
     s.clear();
     s << OP_9 << OP_ADD << OP_11 << OP_EQUAL;
     BOOST_CHECK(!Solver(s, whichType, solutions));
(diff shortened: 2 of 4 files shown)


src/script/ismine.cpp
```diff
@@ -46,6 +46,8 @@ isminetype IsMine(const CKeyStore &keystore, const CTxDestination& dest, bool& i
 
 isminetype IsMine(const CKeyStore &keystore, const CScript& scriptPubKey, bool& isInvalid, SigVersion sigversion)
 {
+    isInvalid = false;
```
promag: Remove.

promag (commented):
Remove periods from commit messages.

src/test/script_standard_tests.cpp
```diff
@@ -0,0 +1,752 @@
...
+        BOOST_CHECK(!isInvalid);
+    }
+
+    // P2WPKH compressed
+    {
+        CBasicKeyStore keystore;
+        keystore.AddKey(keys[0]);
+
+        CScript witnessScript;
+        witnessScript << OP_DUP << OP_HASH160 << ToByteVector(pubkeys[0].GetID()) << OP_EQUALVERIFY << OP_CHECKSIG;
```
jonasschnelli: This variable is unused.

src/script/ismine.cpp
jimpo: Could you please explain why? Since `isValid` is a return parameter, I think it should get set to `false` from an initial value of `true` if the script is not invalid.

gmaxwell:
> Without a6564cc, P2WPKH scripts where the pubkey is uncompressed are incorrectly

sounds correct to change; but how would a P2WPKH redeem script for such a key end up in the wallet in the first place?

jimpo:
@gmaxwell I haven't looked through the wallet code enough to know. I assume it would not happen, and it's just a bug in some internals that wouldn't be exercised, but I'm not 100% sure. I came across it through unit tests, not end-to-end testing of the wallet.

gmaxwell:
Great okay! just making sure you weren't aware of any.

sipa:
@jimpo If a6564cc has any effect, I think it's a sign there is a more severe issue, namely that the CKeyID and the CKey in the CBasicKeyStore's map are not in sync with each other (as the CKey stores the compressedness too).

src/test/script_standard_tests.cpp
```diff
@@ -0,0 +1,746 @@
...
+        scriptPubKey.clear();
+        scriptPubKey << ToByteVector(uncompressedPubkey0) << OP_CHECKSIG;
+
+        // Keystore does not have key
+        result = IsMine(keystore, scriptPubKey, isInvalid);
+        BOOST_CHECK_EQUAL(result, ISMINE_NO);
+        BOOST_CHECK(!isInvalid);
+
+        // Keystore has key
+        keystore.AddKeyPubKey(keys[0], uncompressedPubkey0);
```
sipa: You're adding a compressed CKey (`keys[0]`) with an uncompressed CPubKey, which is incorrect use of the function. You should first change the CKey's compressedness.

Perhaps this warrants a comment on AddKeyPubKey or even an assert. It should alleviate the need for a6564cca7d443789671778f769a24bbb5a4a0652 though.

jimpo: Thanks, will change.

jimpo:
Thanks for the review @sipa. I had forgotten that CKey tracks the compressedness. Updated the test and removed the change to CBasicKeyStore.

src/script/ismine.cpp
promag: Ok, got it.

gmaxwell:
utACK

src/test/script_standard_tests.cpp
```diff
@@ -0,0 +1,747 @@
...
+    BOOST_CHECK_EQUAL(whichType, TX_MULTISIG);
+    BOOST_CHECK_EQUAL(solutions.size(), 5);
+    BOOST_CHECK(solutions[0] == std::vector<unsigned char>({2}));
+    BOOST_CHECK(solutions[1] == ToByteVector(pubkeys[0]));
+    BOOST_CHECK(solutions[2] == ToByteVector(pubkeys[1]));
+    BOOST_CHECK(solutions[3] == ToByteVector(pubkeys[2]));
+    BOOST_CHECK(solutions[4] == std::vector<unsigned char>({3}));
+
+    // TX_NULL_DATA
+    solutions.clear();
```
MarcoFalke: why? It is the task of the solver to set the size of the vector to 0. Otherwise you could as well remove the `BOOST_CHECK_EQUAL(solutions.size(), 0);`, no?

maflcko (commented):
utACK 323dd0e

src/test/script_standard_tests.cpp
jimpo: Good point.

TheBlueMatt (commented):
utACK 184a07d5b6131c13238397db1b1e9a0a1c58c54f

src/test/script_standard_tests.cpp
```diff
@@ -0,0 +1,743 @@
...
+{
+    CKey keys[3];
+    CPubKey pubkeys[3];
+    for (int i = 0; i < 3; i++) {
+        keys[i].MakeNewKey(true);
+        pubkeys[i] = keys[i].GetPubKey();
+    }
+
+    CScript expected, result;
+    CTxDestination dest;
```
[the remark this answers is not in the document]
jimpo: Thanks, removed.

src/test/script_standard_tests.cpp
```diff
@@ -0,0 +1,740 @@
...
+    BOOST_CHECK(!ExtractDestination(s, address));
+
+    // TX_WITNESS_V0_KEYHASH
+    s.clear();
+    s << OP_0 << ToByteVector(pubkey);
+    BOOST_CHECK(!ExtractDestination(s, address));
+
+    // TX_WITNESS_V0_SCRIPTHASH
+    s.clear();
+    s << OP_0 << ToByteVector(CScriptID(redeemScript));
```
sipa: Posthumous review (and fixed in #11167): this does not construct a valid TX_WITNESS_V0_SCRIPTHASH but a TX_WITNESS_V0_KEYHASH (as the pushed size is 20, not 32)

PR #11117, opened 2017-08-23
Prepare for non-Base58 addresses

sipa:
This patch removes the need for the intermediary Base58 type `CBitcoinAddress`, by providing {`Encode`,`Decode`,`IsValid`}`Destination` functions that directly operate on the conversion between `std::string`s and `CTxDestination`.

As a side, it also fixes a number of indentation issues, and removes probably several unnecessary implicit `CTxDestination`<->`CBitcoinAddress` conversions.

This change is far from complete. In follow-ups I'd like to:
* Split off the specific address and key encoding logic from base58.h, and move it to a address.h or so.
* Replace `CTxDestination` with a non-`boost::variant` version (which can be more efficient as `boost::variant` allocates everything on the heap, and remove the need for `boost::get<...>` and `IsValidDestination` calls everywhere).
* Do the same for `CBitcoinSecret`, `CBitcoinExtKey`, and `CBitcoinExtPubKey`.

However, I've tried to keep this patch to be minimally invasive, but still enough to support non-Base58 addresses. Perhaps a smaller patch is possible to hack Bech32 support into `CBitcoinAddress`, but I would consider that a move in the wrong direction.

### DIFF
--- a/src/base58.cpp
+++ b/src/base58.cpp
@@ -212,6 +212,30 @@ int CBase58Data::CompareTo(const CBase58Data& b58) const
 
 namespace
 {
+/** base58-encoded Bitcoin addresses.
+ * Public-key-hash-addresses have version 0 (or 111 testnet).
+ * The data vector contains RIPEMD160(SHA256(pubkey)), where pubkey is the serialized public key.
+ * Script-hash-addresses have version 5 (or 196 testnet).
+ * The data vector contains RIPEMD160(SHA256(cscript)), where cscript is the serialized redemption script.
+ */
+class CBitcoinAddress : public CBase58Data {
+public:
+    bool Set(const CKeyID &id);
+    bool Set(const CScriptID &id);
+    bool Set(const CTxDestination &dest);
+    bool IsValid() const;
+    bool IsValid(const CChainParams &params) const;
+
+    CBitcoinAddress() {}
+    CBitcoinAddress(const CTxDestination &dest) { Set(dest); }
+    CBitcoinAddress(const std::string& strAddress) { SetString(strAddress); }
+    CBitcoinAddress(const char* pszAddress) { SetString(pszAddress); }
+
+    CTxDestination Get() const;
+    bool GetKeyID(CKeyID &keyID) const;
+    bool IsScript() const;
+};
+
 class CBitcoinAddressVisitor : public boost::static_visitor<bool>
 {
 private:
--- a/src/base58.cpp
+++ b/src/base58.cpp
@@ -318,3 +318,25 @@ bool CBitcoinSecret::SetString(const std::string& strSecret)
 {
     return SetString(strSecret.c_str());
 }
+
+std::string EncodeDestination(const CTxDestination& dest)
+{
+    CBitcoinAddress addr(dest);
+    if (!addr.IsValid()) return "";
+    return addr.ToString();
+}
+
+CTxDestination DecodeDestination(const std::string& str)
+{
+    return CBitcoinAddress(str).Get();
+}
+
+bool IsValidDestinationString(const std::string& str, const CChainParams& params)
+{
+    return CBitcoinAddress(str).IsValid(params);
+}
+
+bool IsValidDestinationString(const std::string& str)
+{
+    return CBitcoinAddress(str).IsValid();
+}
--- a/src/qt/addresstablemodel.cpp
+++ b/src/qt/addresstablemodel.cpp
@@ -82,14 +82,14 @@ public:
             LOCK(wallet->cs_wallet);
             for (const std::pair<CTxDestination, CAddressBookData>& item : wallet->mapAddressBook)
             {
-                const CBitcoinAddress& address = item.first;
-                bool fMine = IsMine(*wallet, address.Get());
+                const CTxDestination& address = item.first;
+                bool fMine = IsMine(*wallet, address);
                 AddressTableEntry::Type addressType = translateTransactionType(
                         QString::fromStdString(item.second.purpose), fMine);
                 const std::string& strName = item.second.name;
                 cachedAddressTable.append(AddressTableEntry(addressType,
                                   QString::fromStdString(strName),
-                                  QString::fromStdString(address.ToString())));
+                                  QString::fromStdString(EncodeDestination(address))));
             }
         }
         // qLowerBound() and qUpperBound() require our cachedAddressTable list to be sorted in asc order
@@ -246,7 +246,7 @@ bool AddressTableModel::setData(const QModelIndex &index, const QVariant &value,
     if(role == Qt::EditRole)
     {
         LOCK(wallet->cs_wallet); /* For SetAddressBook / DelAddressBook */
-        CTxDestination curAddress = CBitcoinAddress(rec->address.toStdString()).Get();
+        CTxDestination curAddress = DecodeDestination(rec->address.toStdString());
         if(index.column() == Label)
         {
             // Do nothing, if old label == new label
@@ -257,7 +257,7 @@ bool AddressTableModel::setData(const QModelIndex &index, const QVariant &value,
             }
             wallet->SetAddressBook(curAddress, value.toString().toStdString(), strPurpose);
         } else if(index.column() == Address) {
-            CTxDestination newAddress = CBitcoinAddress(value.toString().toStdString()).Get();
+            CTxDestination newAddress = DecodeDestination(value.toString().toStdString());
             // Refuse to set invalid address, set error status and return false
             if(boost::get<CNoDestination>(&newAddress))
             {
@@ -358,7 +358,7 @@ QString AddressTableModel::addRow(const QString &type, const QString &label, con
         // Check for duplicate addresses
         {
             LOCK(wallet->cs_wallet);
-            if(wallet->mapAddressBook.count(CBitcoinAddress(strAddress).Get()))
+            if(wallet->mapAddressBook.count(DecodeDestination(strAddress)))
             {
                 editStatus = DUPLICATE_ADDRESS;
                 return QString();
@@ -384,7 +384,7 @@ QString AddressTableModel::addRow(const QString &type, const QString &label, con
                 return QString();
             }
         }
-        strAddress = CBitcoinAddress(newKey.GetID()).ToString();
+        strAddress = EncodeDestination(newKey.GetID());
     }
     else
     {
@@ -394,7 +394,7 @@ QString AddressTableModel::addRow(const QString &type, const QString &label, con
     // Add entry
     {
         LOCK(wallet->cs_wallet);
-        wallet->SetAddressBook(CBitcoinAddress(strAddress).Get(), strLabel,
+        wallet->SetAddressBook(DecodeDestination(strAddress), strLabel,
                                (type == Send ? "send" : "receive"));
     }
     return QString::fromStdString(strAddress);
@@ -412,7 +412,7 @@ bool AddressTableModel::removeRows(int row, int count, const QModelIndex &parent
     }
     {
         LOCK(wallet->cs_wallet);
-        wallet->DelAddressBook(CBitcoinAddress(rec->address.toStdString()).Get());
+        wallet->DelAddressBook(DecodeDestination(rec->address.toStdString()));
     }
     return true;
 }
@@ -423,8 +423,8 @@ QString AddressTableModel::labelForAddress(const QString &address) const
 {
     {
         LOCK(wallet->cs_wallet);
-        CBitcoinAddress address_parsed(address.toStdString());
-        std::map<CTxDestination, CAddressBookData>::iterator mi = wallet->mapAddressBook.find(address_parsed.Get());
+        CTxDestination destination = DecodeDestination(address.toStdString());
+        std::map<CTxDestination, CAddressBookData>::iterator mi = wallet->mapAddressBook.find(destination);
         if (mi != wallet->mapAddressBook.end())
         {
             return QString::fromStdString(mi->second.name);
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -777,19 +777,18 @@ void SendCoinsDialog::coinControlChangeEdited(const QString& text)
         CoinControlDialog::coinControl->destChange = CNoDestination();
         ui->labelCoinControlChangeLabel->setStyleSheet("QLabel{color:red;}");
 
-        CBitcoinAddress addr = CBitcoinAddress(text.toStdString());
+        const CTxDestination dest = DecodeDestination(text.toStdString());
 
         if (text.isEmpty()) // Nothing entered
         {
             ui->labelCoinControlChangeLabel->setText("");
         }
-        else if (!addr.IsValid()) // Invalid address
+        else if (!IsValidDestination(dest)) // Invalid address
         {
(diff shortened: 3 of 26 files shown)


src/wallet/rpcwallet.cpp
```diff
@@ -275,24 +275,25 @@ UniValue setaccount(const JSONRPCRequest& request)
...
         strAccount = AccountFromValue(request.params[1]);
 
     // Only add the account if the address is yours.
-    if (IsMine(*pwallet, address.Get())) {
+    if (IsMine(*pwallet, dest)) {
         // Detect when changing the account of an address that is the 'unused current key' of another account:
-        if (pwallet->mapAddressBook.count(address.Get())) {
-            std::string strOldAccount = pwallet->mapAddressBook[address.Get()].name;
-            if (address == GetAccountAddress(pwallet, strOldAccount)) {
+        if (pwallet->mapAddressBook.count(dest)) {
```
promag: Nit, kind of unrelated, use `.find()` to avoid 2nd lookup below.

src/wallet/rpcwallet.cpp
```diff
@@ -451,9 +454,10 @@ UniValue sendtoaddress(const JSONRPCRequest& request)
 
     LOCK2(cs_main, pwallet->cs_wallet);
 
-    CBitcoinAddress address(request.params[0].get_str());
-    if (!address.IsValid())
+    CTxDestination dest = DecodeDestination(request.params[0].get_str());
+    if (!IsValidDestination(dest)) {
         throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid Bitcoin address");
```
promag: Follow up, should the error message be `Invalid destination address`?

src/wallet/rpcwallet.cpp
```diff
@@ -2695,18 +2704,20 @@ UniValue listunspent(const JSONRPCRequest& request)
...
-            if (!address.IsValid())
-                throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string("Invalid Bitcoin address: ")+input.get_str());
-            if (setAddress.count(address))
-                throw JSONRPCError(RPC_INVALID_PARAMETER, std::string("Invalid parameter, duplicated address: ")+input.get_str());
-           setAddress.insert(address);
+            CTxDestination dest = DecodeDestination(input.get_str());
+            if (!IsValidDestination(dest)) {
+                throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, std::string("Invalid Bitcoin address: ") + input.get_str());
+            }
+            if (destinations.count(dest)) {
```
promag: Replace with `.insert().second` like above.

promag (commented):
There is this pattern
```cpp
DecodeDestination();
if (!IsValidDestination()) {
```
What about:
```cpp
CTxDestination dest;
if (!DecodeDestination(str, dest)) {
```

Also, to avoid more usages of `CBitcoinAddress`, add commit to move it to `base58.cpp`?

src/wallet/rpcwallet.cpp
sipa: An address is just the string-encoded form of a CTxDestination, so no. I think the "destination" name should be internal.

src/wallet/rpcwallet.cpp
sipa: Trying to keep this as much of a straightforward change as possible, as it may need backporting.

src/wallet/rpcwallet.cpp
promag: Off topic, just `Invalid address` as it seems irrelevant to have `Bitcoin` in the error message?

sipa: Oh, sure!

sipa:
> What about:
> 
> ```c++
> CTxDestination dest;
> if (!DecodeDestination(str, dest)) {
> ```

There are plenty of cases where the result of decoding can just be passed to something else, and `Function(DecodeDestination(addr))` is far cleaner than `CTxDestination dest; DecodeDestination(addr, dest); Function(dest);`, so I prefer to keep that possible.

The current pattern is indeed a bit verbose, but once we rewrite `CTxDestination` to be a native type (rather than a `boost::variant`), it could at least simplify to `dest.IsValid()` or so.

promag:
@sipa sure, I just wanted to question that "style" because it's used elsewhere.

ACK 8363a2c.

jonasschnelli:
utACK 8363a2ccaffba793068db4f97184121730707104

laanwj:
Added 0.15.1 milestone

kallewoof (approved):
utACK 8363a2ccaffba793068db4f97184121730707104
LGTM. I skipped over QT stuff as I don't know it very well.

src/script/standard.cpp
```diff
@@ -317,3 +317,7 @@ CScript GetScriptForWitness(const CScript& redeemscript)
     ret << OP_0 << ToByteVector(hash);
     return ret;
 }
+
+bool IsValidDestination(const CTxDestination& dest) {
+    return dest.which() != 0;
```
theuni: Maybe add a quick comment that this is CNoDestination so it comes up when grepping for future changes.

src/script/standard.h
```diff
@@ -77,10 +77,13 @@ class CNoDestination {
  *  * CNoDestination: no destination set
  *  * CKeyID: TX_PUBKEYHASH destination
  *  * CScriptID: TX_SCRIPTHASH destination
- *  A CTxDestination is the internal data type encoded in a CBitcoinAddress
+ *  A CTxDestination is the internal data type encoded in a bitcoin address
  */
 typedef boost::variant<CNoDestination, CKeyID, CScriptID> CTxDestination;
 
+/** Check whether a CTxDestination is a CNoDestination. */
```
theuni: got it. ignore last nit.

src/wallet/rpcwallet.cpp
```diff
@@ -529,16 +533,16 @@ UniValue listaddressgroupings(const JSONRPCRequest& request)
...
         UniValue jsonGrouping(UniValue::VARR);
-        for (CTxDestination address : grouping)
+        for (const CTxDestination& address : grouping)
         {
             UniValue addressInfo(UniValue::VARR);
-            addressInfo.push_back(CBitcoinAddress(address).ToString());
+            addressInfo.push_back(EncodeDestination(address));
             addressInfo.push_back(ValueFromAmount(balances[address]));
             {
-                if (pwallet->mapAddressBook.find(CBitcoinAddress(address).Get()) != pwallet->mapAddressBook.end()) {
```
theuni: heh.

theuni (commented):
utACK other than the keyID issue.

src/qt/signverifymessagedialog.cpp
```diff
@@ -117,16 +117,14 @@ void SignVerifyMessageDialog::on_signMessageButton_SM_clicked()
...
+    if (!IsValidDestination(destination)) {
         ui->statusLabel_SM->setStyleSheet("QLabel { color: red; }");
         ui->statusLabel_SM->setText(tr("The entered address is invalid.") + QString(" ") + tr("Please check the address and try again."));
         return;
     }
-    CKeyID keyID;
-    if (!addr.GetKeyID(keyID))
-    {
+    const CKeyID* keyID = boost::get<CKeyID>(&destination);
+    if (keyID) {
```
promag: Good catch.

src/qt/addresstablemodel.cpp
```diff
@@ -82,14 +82,14 @@ class AddressTablePriv
             LOCK(wallet->cs_wallet);
             for (const std::pair<CTxDestination, CAddressBookData>& item : wallet->mapAddressBook)
             {
-                const CBitcoinAddress& address = item.first;
```
sipa: It was creating a temporary of type CBitcoinAddress by converting the CTxDestination initializer, using the implicit `CBitcoinAddress(const CTxDestination &dest)` constructor. That temporary's lifetime was then extended by taking a reference to it, named `address`.

src/wallet/rpcwallet.cpp
sipa: It took me a while to figure out what this `CBitcoinAddress` was doing. Answer: nothing at all.

src/qt/signverifymessagedialog.cpp
sipa: Wow, thanks :S

src/qt/addresstablemodel.cpp
theuni: Interesting, I didn't know an implicit conversion could be done when extending the lifetime of a const ref. Scary!

sipa:
Removed `CBitcoinAddress` entirely, using a commit from #11167.

promag (commented):
Since this is the base for Bech32, maybe move everything related to `CTxDestination` out of base58, including tests, to a new place. Also move `CTxDestination` definition and related functions from `src/script/standard.h`?

At least should be done in a follow up?

src/base58.cpp
```diff
@@ -12,6 +12,7 @@
 #include <string.h>
 #include <vector>
 #include <string>
+#include <algorithm>
```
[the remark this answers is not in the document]
sipa: Fixed.

src/qt/sendcoinsdialog.cpp
```diff
@@ -777,21 +777,20 @@ void SendCoinsDialog::coinControlChangeEdited(const QString& text)
...
 
-        CBitcoinAddress addr = CBitcoinAddress(text.toStdString());
+        CTxDestination destination = DecodeDestination(text.toStdString());
 
         if (text.isEmpty()) // Nothing entered
         {
             ui->labelCoinControlChangeLabel->setText("");
         }
-        else if (!addr.IsValid()) // Invalid address
+        else if (!IsValidDestination(destination)) // Invalid address
```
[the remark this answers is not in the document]
sipa: Not going to reflow this entire function in a PR intended for backporting.

src/rpc/misc.cpp
```diff
@@ -207,15 +208,14 @@ UniValue validateaddress(const JSONRPCRequest& request)
...
+    CTxDestination dest = DecodeDestination(request.params[0].get_str());
+    bool isValid = IsValidDestination(dest);
 
     UniValue ret(UniValue::VOBJ);
     ret.push_back(Pair("isvalid", isValid));
     if (isValid)
     {
-        CTxDestination dest = address.Get();
-        std::string currentAddress = address.ToString();
+        std::string currentAddress = EncodeDestination(dest);
```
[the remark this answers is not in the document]
sipa: There may be reasons not to do that. Bech32 addresses (added in a later PR) can be uppercase or lowercase, but we always want to store and look up in lowercase form. This is an easy way of doing that, and we should probably check that it's done consistently elsewhere too.

sipa:
> Since this is the base for Bech32, maybe move everything related to CTxDestination out of base58, including tests, to a new place.

Yes, read the PR description. I plan to do this, but not in a patch that's intended for backporting.

> Also move `CTxDestination` definition and related functions from `src/script/standard.h`?

Why? To where?

instagibbs:
utACK https://github.com/bitcoin/bitcoin/pull/11117/commits/87d5c28583374f1be6273c3620b29b0a505e0191

though I do not know the QT stuff well.

src/rpc/misc.cpp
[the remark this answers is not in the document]
promag: Agree, I'm +1 for clear data validation and/or sanitization. Although `encode(decode(data))` is a obscure way of doing it, not to mention `encode(decode(data))` should equal `data`?

promag:
> Removed CBitcoinAddress entirely, using a commit from #11167.

@sipa missing this commit as `CBitcoinAddress` is still around.

sipa:
@promag Yes, I removed it again.

This PR is just the part that is likely to break easily and need non-trivial rebasing.

src/qt/transactiondesc.cpp
```diff
@@ -91,9 +91,8 @@ QString TransactionDesc::toHTML(CWallet *wallet, CWalletTx &wtx, TransactionReco
         if (nNet > 0)
         {
             // Credit
-            if (CBitcoinAddress(rec->address).IsValid())
-            {
-                CTxDestination address = CBitcoinAddress(rec->address).Get();
+            if (IsValidDestination(rec->address)) {
+                CTxDestination address = DecodeDestination(rec->address);
```
promag: This incur in a double decoding since above `IsValidDestination` also decodes. Maybe swap:
```cpp
CTxDestination address = DecodeDestination(rec->address);
if (IsValidDestination(address)) {
```

src/script/standard.cpp
promag: Could use `!(operator==)` instead?

laanwj:
Not sure if we support this version, but I get the following compile error with clang 3.5.0:
```
/home/user/src/bitcoin/src/qt/signverifymessagedialog.cpp:235:40: error: invalid operands to binary expression ('CTxDestination' (aka 'variant<CNoDestination, CKeyID, CScriptID>') and 'CTxDestination')
    if (CTxDestination(pubkey.GetID()) != destination) {
```
Then a lot of note: candidate function not viable like:
```
/usr/include/x86_64-linux-gnu/qt5/QtCore/qchar.h:544:13: note: candidate function not viable: no known conversion from 'CTxDestination' (aka 'variant<CNoDestination, CKeyID, CScriptID>') to 'QChar' for 1st argument
inline bool operator!=(QChar c1, QChar c2) { return c1.unicode() != c2.unicode(); }
```

JeremyRubin:
@laanwj 

changing it to 


    if (!(CTxDestination(pubkey.GetID()) == destination)) {

probably fixes it. Ran into this while writing https://github.com/bitcoin/bitcoin/pull/9991

laanwj:
I confirm that @JeremyRubin's workaround makes it compile.

sipa:
@JeremyRubin Fixed

morcos:
utACK eebedca

sipa:
@laanwj Changed to `IsValidDestinationString`.

laanwj:
utACK 864cd2787

src/wallet/rpcwallet.cpp
```diff
@@ -454,9 +457,10 @@ UniValue sendtoaddress(const JSONRPCRequest& request)
     ObserveSafeMode();
     LOCK2(cs_main, pwallet->cs_wallet);
 
-    CBitcoinAddress address(request.params[0].get_str());
-    if (!address.IsValid())
-        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid Bitcoin address");
+    CTxDestination dest = DecodeDestination(request.params[0].get_str());
+    if (!IsValidDestination(dest)) {
+        throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Invalid address");
```
MarcoFalke: nit: Any reason why this was changed from "Invalid Bitcoin address" to "Invalid address" and the other places left as is. This is part of a error string, so shouldn't break anything I guess.

maflcko (commented):
(really late) post-merge utACK 864cd278741ab0e17e7d93eedee39417d4797b37

src/wallet/rpcwallet.cpp
promag: @achow101 regarding the need for error testing.

MarcoFalke:
Removing from backport

PR #11119, opened 2017-08-23
[doc] build-windows: Mention that only trusty works

maflcko:
This should prevent people from running into the issues to later find that there is no solution yet.


laanwj:
Yes!
(needs backport to 0.15 branch)

fanquake:
ACK fa14b67

achow101:
ACK

PR #11120, opened 2017-08-23
[RPC] verifytxoutproof returns object including blockhash

instagibbs:
Allows direct use of the proof to get block header related info without additional parsing, and more directly mirrors `gettxoutproof` arguments.


src/rpc/rawtransaction.cpp
```diff
@@ -258,19 +258,23 @@ UniValue verifytxoutproof(const JSONRPCRequest& request)
...
             "verifytxoutproof \"proof\"\n"
-            "\nVerifies that a proof points to a transaction in a block, returning the transaction it commits to\n"
-            "and throwing an RPC error if the block is not in our best chain\n"
+            "\nVerifies that a proof points to a transaction in a block, returning the blockhash and transactions it commits to\n"
+            ", throwing an RPC error if the block is not in our best chain\n"
             "\nArguments:\n"
             "1. \"proof\"    (string, required) The hex-encoded proof generated by gettxoutproof\n"
-            "\nResult:\n"
-            "[\"txid\"]      (array, strings) The txid(s) which the proof commits to, or empty array if the proof is invalid\n"
+            "\nResult: (or empty if proof is invalid)\n"
```
kallewoof: Remove `or` maybe? I.e. `Result: (empty if proof is invalid)\n`

kallewoof (commented):
utACK d77cded0b1d026f030ee424b89a01cfd6b7b9000

instagibbs:
updated with nit fix

src/rpc/rawtransaction.cpp
```diff
@@ -258,19 +258,23 @@ UniValue verifytxoutproof(const JSONRPCRequest& request)
     if (request.fHelp || request.params.size() != 1)
         throw std::runtime_error(
             "verifytxoutproof \"proof\"\n"
-            "\nVerifies that a proof points to a transaction in a block, returning the transaction it commits to\n"
-            "and throwing an RPC error if the block is not in our best chain\n"
+            "\nVerifies that a proof points to a transaction in a block, returning the blockhash and transactions it commits to\n"
+            ", throwing an RPC error if the block is not in our best chain\n"
```
promag: Nit, `,` is the above line?

src/rpc/rawtransaction.cpp
```diff
@@ -258,19 +258,23 @@ UniValue verifytxoutproof(const JSONRPCRequest& request)
     if (request.fHelp || request.params.size() != 1)
         throw std::runtime_error(
             "verifytxoutproof \"proof\"\n"
-            "\nVerifies that a proof points to a transaction in a block, returning the transaction it commits to\n"
-            "and throwing an RPC error if the block is not in our best chain\n"
+            "\nVerifies that a proof points to a transaction in a block, returning the blockhash and transactions it commits to\n"
```
promag: `... the block hash and ...`

src/rpc/rawtransaction.cpp
```diff
@@ -258,19 +258,23 @@ UniValue verifytxoutproof(const JSONRPCRequest& request)
...
+            ", throwing an RPC error if the block is not in our best chain\n"
             "\nArguments:\n"
             "1. \"proof\"    (string, required) The hex-encoded proof generated by gettxoutproof\n"
-            "\nResult:\n"
-            "[\"txid\"]      (array, strings) The txid(s) which the proof commits to, or empty array if the proof is invalid\n"
+            "\nResult: (empty if proof is invalid)\n"
```
promag: This is breaking change as the result is now an object instead of an array, needs release note.

src/rpc/rawtransaction.cpp
```diff
@@ -283,7 +287,10 @@ UniValue verifytxoutproof(const JSONRPCRequest& request)
         throw JSONRPCError(RPC_INVALID_ADDRESS_OR_KEY, "Block not found in chain");
 
     for (const uint256& hash : vMatch)
```
promag: Nit, add curly braces.

promag (commented):
Concept ACK.

src/rpc/rawtransaction.cpp
kallewoof: I was gonna mention that but chose not to. Agree though, I think that would look better.

instagibbs:
fixed @promag nits, curly-bracketed the surrounding statements since I'm touching it already

src/rpc/rawtransaction.cpp
```diff
@@ -258,32 +258,42 @@ UniValue verifytxoutproof(const JSONRPCRequest& request)
...
+    }
+
+    for (const uint256& hash : vMatch) {
+        txids.push_back(hash.GetHex());
+    }
 
-    for (const uint256& hash : vMatch)
-        res.push_back(hash.GetHex());
+    res.push_back(Pair("txids", txids));
+    res.push_back(Pair("blockhash", merkleBlock.header.GetHash().GetHex()));
```
promag: Push `blockhash` first.

kallewoof: I'm curious about the reason for this request. Alphabetical ordering?

promag: Also matches the documentation above.

instagibbs:
rebased, re-added legacy behavior under deprecation

instagibbs:
ping @kallewoof

src/rpc/rawtransaction.cpp
```diff
@@ -294,6 +294,8 @@ UniValue verifytxoutproof(const JSONRPCRequest& request)
             "    \"blockhash\"   (string) The blockhash the included proof commits to\n"
             "    [\"txid\"]      (array, strings) The txid(s) which the proof commits to\n"
             "}\n"
+            "\nResult: (DEPRECATED. To see this result in v0.17 instead, please start bitcoind with -deprecatedrpc=verifytxoutproof. Empty if proof is invalid)\n"
```
kallewoof: I'm not sure about the `in v0.17 instead` wording -- what about when bitcoin hits v0.18?

kallewoof (commented):
re-utACK a3716960de5555572408efdab4eba75484c8c1e3

src/rpc/rawtransaction.cpp
promag: I thought the same, but there are other occurrences in the code. I would remove the version though.

promag:
@instagibbs See https://github.com/bitcoin/bitcoin/pull/11872/files#diff-ef76fd6674f07db88c3422fdbf0bcf9fR65 for release note example.

src/rpc/rawtransaction.cpp
instagibbs: I'm a bit confused, if this is merged now, it definitely would be the way to see the old text for 0.17. Now, perhaps the note should also say when it will be removed?

instagibbs: Pushed a new wording, hopefully clearer and more pertinent to the user.

instagibbs:
fixed up deprecation message, added release notes

promag (commented):
Could add a test for the empty response `{}`, at least I don't see it?

doc/release-notes.md
```diff
@@ -62,6 +62,7 @@ RPC changes
 ### Low-level changes
 
 - The `fundrawtransaction` rpc will reject the previously deprecated `reserveChangeKey` option.
+- The `verifytxoutproof` rpc will output an object instead of an array unless `-deprecatedrpc=verifytxoutproof` is set
```
promag: Nit, missing period, sorry.

instagibbs:
fixed up release note nit. I think the only way to return `{}` is to have a non-matching hash header... too lazy to construct one in paste it in.

promag:
@instagibbs here is one
```
verifytxoutproof 0000000000000000000000000000000000000000000000000000000000000000000000000100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000
{
}
```
Maybe it should throw an error in this case?

kallewoof (approved):
utACK - looks good

instagibbs:
@promag yeah I'll catch the deprecated case of `[]` and throw an error for new behavior

instagibbs:
done

MarcoFalke:
Needs rebase to fix travis (sorry)

instagibbs:
rebased

kallewoof (approved):
utACK ec1695eedc545545e5273e19f1e1b80aaebecf28

promag:
utACK ec1695e.

MarcoFalke:
Needs rebase (after merge of `LookupBlockIndex`)

instagibbs:
rebased

kallewoof:
@instagibbs Travis is timing out. I tried restarting the two jobs that timed out but it still fails. Not sure what's up.

promag:
master should be fixed after #12681.

kallewoof:
@promag I think that's unrelated (and I don't think @instagibbs merged the commit that triggered the error that was fixed).

src/rpc/rawtransaction.cpp
```diff
@@ -285,24 +285,35 @@ UniValue verifytxoutproof(const JSONRPCRequest& request)
...
 
     std::vector<uint256> vMatch;
     std::vector<unsigned int> vIndex;
-    if (merkleBlock.txn.ExtractMatches(vMatch, vIndex) != merkleBlock.header.hashMerkleRoot)
-        return res;
+    if (merkleBlock.txn.ExtractMatches(vMatch, vIndex) != merkleBlock.header.hashMerkleRoot) {
+        if (IsDeprecatedRPCEnabled("verifytxoutproof")) {
+            return txids;
+        } else {
+            throw JSONRPCError(RPC_INVALID_PARAMETER, "No valid transaction commitments are found in the proof");
```
sipa: Is the RPC user at fault when this happens? If not, I don't think this should be an RPC error.

instagibbs: We already error out when the block is not found in the chain, which is just as much the users' fault.

I'll leave this up to others to chime in.

kallewoof:
re-utACK

src/rpc/rawtransaction.cpp
sipa: Ok, fair. Let's keep it this way then.

instagibbs:
rebased due to release notes conflict

instagibbs:
rebased again...

instagibbs:
rebased again for yet another release notes conflict

PR #11121, opened 2017-08-23
TestNode tidyups

jnewbery:
Some additional tidyups after the introduction of TestNode:

- commit 1 makes TestNode use the correct rpc timeout. This should have been included in #11077
- commit 2 separates `add_node()` from `start_node()` as originally discussed here: https://github.com/bitcoin/bitcoin/pull/10556#discussion_r121161453 with @kallewoof . The test writer no longer needs to assign to `self.nodes` when starting/stopping nodes.
- commit 3 adds a `set_test_params()` method, so individual tests don't need to override `__init__()` and call `super().__init__()`

### DIFF
--- a/test/functional/assumevalid.py
+++ b/test/functional/assumevalid.py
@@ -54,8 +54,7 @@ class BaseNode(NodeConnCB):
         self.send_message(headers_message)
 
 class AssumeValidTest(BitcoinTestFramework):
-    def __init__(self):
-        super().__init__()
+    def set_test_params(self):
         self.setup_clean_chain = True
         self.num_nodes = 3
 
--- a/test/functional/assumevalid.py
+++ b/test/functional/assumevalid.py
@@ -60,10 +60,11 @@ class AssumeValidTest(BitcoinTestFramework):
         self.num_nodes = 3
 
     def setup_network(self):
+        self.add_nodes(3, self.options.tmpdir)
         # Start node0. We don't start the other nodes yet since
         # we need to pre-mine a block with an invalid transaction
         # signature so we can pass in the block hash as assumevalid.
-        self.nodes = [self.start_node(0, self.options.tmpdir)]
+        self.start_node(0)
 
     def send_blocks_until_disconnected(self, node):
         """Keep sending blocks to the node until we're disconnected."""
@@ -162,15 +163,13 @@ class AssumeValidTest(BitcoinTestFramework):
             height += 1
 
         # Start node1 and node2 with assumevalid so they accept a block with a bad signature.
-        self.nodes.append(self.start_node(1, self.options.tmpdir,
-                                     ["-assumevalid=" + hex(block102.sha256)]))
+        self.start_node(1, extra_args=["-assumevalid=" + hex(block102.sha256)])
         node1 = BaseNode()  # connects to node1
         connections.append(NodeConn('127.0.0.1', p2p_port(1), self.nodes[1], node1))
         node1.add_connection(connections[1])
         node1.wait_for_verack()
 
-        self.nodes.append(self.start_node(2, self.options.tmpdir,
-                                     ["-assumevalid=" + hex(block102.sha256)]))
+        self.start_node(2, extra_args=["-assumevalid=" + hex(block102.sha256)])
         node2 = BaseNode()  # connects to node2
         connections.append(NodeConn('127.0.0.1', p2p_port(2), self.nodes[2], node2))
         node2.add_connection(connections[2])
--- a/test/functional/assumevalid.py
+++ b/test/functional/assumevalid.py
@@ -60,7 +60,7 @@ class AssumeValidTest(BitcoinTestFramework):
         self.num_nodes = 3
 
     def setup_network(self):
-        self.add_nodes(3, self.options.tmpdir)
+        self.add_nodes(3)
         # Start node0. We don't start the other nodes yet since
         # we need to pre-mine a block with an invalid transaction
         # signature so we can pass in the block hash as assumevalid.
--- a/test/functional/bumpfee.py
+++ b/test/functional/bumpfee.py
@@ -34,21 +34,18 @@ class BumpFeeTest(BitcoinTestFramework):
         super().__init__()
         self.num_nodes = 2
         self.setup_clean_chain = True
+        self.extra_args = [["-prematurewitness", "-walletprematurewitness", "-walletrbf={}".format(i)]
+                           for i in range(self.num_nodes)]
 
-    def setup_network(self, split=False):
-        extra_args = [["-prematurewitness", "-walletprematurewitness", "-walletrbf={}".format(i)]
-                      for i in range(self.num_nodes)]
-        self.nodes = self.start_nodes(self.num_nodes, self.options.tmpdir, extra_args)
-
+    def run_test(self):
         # Encrypt wallet for test_locked_wallet_fails test
         self.nodes[1].node_encrypt_wallet(WALLET_PASSPHRASE)
-        self.nodes[1] = self.start_node(1, self.options.tmpdir, extra_args[1])
+        self.start_node(1)
         self.nodes[1].walletpassphrase(WALLET_PASSPHRASE, WALLET_PASSPHRASE_TIMEOUT)
 
         connect_nodes_bi(self.nodes, 0, 1)
         self.sync_all()
 
-    def run_test(self):
         peer_node, rbf_node = self.nodes
         rbf_node_address = rbf_node.getnewaddress()
 
--- a/test/functional/bumpfee.py
+++ b/test/functional/bumpfee.py
@@ -30,8 +30,7 @@ WALLET_PASSPHRASE_TIMEOUT = 3600
 
 
 class BumpFeeTest(BitcoinTestFramework):
-    def __init__(self):
-        super().__init__()
+    def set_test_params(self):
         self.num_nodes = 2
         self.setup_clean_chain = True
         self.extra_args = [["-prematurewitness", "-walletprematurewitness", "-walletrbf={}".format(i)]
--- a/test/functional/pruning.py
+++ b/test/functional/pruning.py
@@ -57,7 +57,7 @@ class PruneTest(BitcoinTestFramework):
         sync_blocks(self.nodes[0:5])
 
     def setup_nodes(self):
-        self.add_nodes(self.num_nodes, self.options.tmpdir, self.extra_args, timewait=900)
+        self.add_nodes(self.num_nodes, self.extra_args, timewait=900)
         self.start_nodes()
 
     def create_big_chain(self):
--- a/test/functional/pruning.py
+++ b/test/functional/pruning.py
@@ -56,6 +56,10 @@ class PruneTest(BitcoinTestFramework):
         connect_nodes(self.nodes[0], 4)
         sync_blocks(self.nodes[0:5])
 
+    def setup_nodes(self):
+        self.add_nodes(self.num_nodes, self.options.tmpdir, self.extra_args, timewait=900)
+        self.start_nodes()
+
     def create_big_chain(self):
         # Start by creating some coinbases we can spend later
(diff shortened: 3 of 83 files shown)


test/functional/txn_clone.py
```diff
@@ -8,12 +8,6 @@
 from test_framework.util import *
 
 class TxnMallTest(BitcoinTestFramework):
-
-    def __init__(self):
-        super().__init__()
-        self.num_nodes = 4
-        self.setup_clean_chain = False
```
MarcoFalke: I prefer to explicitly set the network topology and initial state of the chain.

test/functional/test_framework/test_framework.py
```diff
@@ -202,51 +175,91 @@ def main(self):
             logging.shutdown()
             sys.exit(TEST_EXIT_FAILED)
 
-    # Public helper methods. These can be accessed by the subclass test scripts.
+    # Methods to override in subclass test scripts.
+    def set_test_params(self):
+        """Override this method to change default values for number of nodes, topology, etc"""
+        pass
```
MarcoFalke: should raise `RuntimeError('Not implemented')` to enforce the docstring.

maflcko (commented):
Concept ACK after quick glance

test/functional/test_framework/test_framework.py
```diff
@@ -202,51 +175,91 @@ def main(self):
...
-        return node
+    def run_test(self):
+        """Override this method to define test logic"""
+        raise NotImplementedError
 
-    def start_nodes(self, num_nodes, dirname, extra_args=None, rpchost=None, timewait=None, binary=None):
-        """Start multiple bitcoinds, return RPC connections to them"""
+    # Public helper methods. These can be accessed by the subclass test scripts.
+
+    def add_nodes(self, num_nodes, dirname, extra_args=None, rpchost=None, timewait=None, binary=None):
```
kallewoof: Since `self` knows about `self.options.tmpdir`, and all calls to `self.add_nodes` passes `self.options.tmpdir` as `dirname` argument, it seems clean to just give it a default `None` and to set to `self.options.tmpdir if None` in `add_nodes`.
Or perhaps remove it entirely and always use `self.options.tmpdir`. Will it ever be anything else?

test/functional/test_framework/test_node.py
```diff
@@ -61,9 +61,13 @@ def __getattr__(self, *args, **kwargs):
         assert self.rpc_connected and self.rpc is not None, "Error: no RPC connection"
         return self.rpc.__getattr__(*args, **kwargs)
 
-    def start(self):
+    def start(self, extra_args, stderr=None):
```
kallewoof: `extra_args=None, stderr=None` maybe?

test/functional/txn_clone.py
kallewoof: Faint agreement. It's easy to forget and start up 4 nodes when you only really need 2, for example. Not seeing the number of nodes makes that even harder to remember.

test/functional/wallet.py
```diff
@@ -7,28 +7,27 @@
...
-    def __init__(self):
-        super().__init__()
+    def set_test_params(self):
         self.setup_clean_chain = True
-        self.num_nodes = 4
         self.extra_args = [['-usehd={:d}'.format(i%2==0)] for i in range(4)]
 
     def setup_network(self):
-        self.nodes = self.start_nodes(3, self.options.tmpdir, self.extra_args[:3])
+        self.add_nodes(4, self.options.tmpdir, self.extra_args)
```
kallewoof: Why does it not just `add_nodes(3, ..)` here and then add/start/connect the fourth node below? That would remove a lot of necessary `[self.nodes[0:3]]` stuff.

kallewoof (commented):
utACK 06bde74f5ec8660975275f8594aee3a99064e327

test/functional/txn_clone.py
jnewbery: Sounds good. I've added a new commit that makes setting `self.num_nodes` mandatory. I haven't done the same for `self.setup_clean_chain`, since I think it makes more sense for that to have a default. If people feel strongly, it can be added in a future commit.

test/functional/test_framework/test_framework.py
jnewbery: Yes - done

test/functional/test_framework/test_framework.py
jnewbery: Good point. Thanks! Added this in a new commit.

test/functional/test_framework/test_node.py
jnewbery: `TestNode.start` is only ever called by `TestFramework.start_node`, which has its own default for `extra_args`, so not strictly necessary, but I agree having a default here makes it clearer. Added.

test/functional/wallet.py
jnewbery: Because there's not currently a method to add a single node. The model after this PR is to add all nodes at the top of the test and then start/stop them as required during the test.

We could potentially add an `add_node()` method to be called during the test, but lets leave that for a future PR.

jnewbery:
@MarcoFalke @kallewoof - thanks for the review. I think I've responded to all your comments.

New commits pushed, and rebased on master.

kallewoof:
utACK df824a9

test/functional/assumevalid.py
```diff
@@ -54,16 +54,16 @@ def send_header_for_blocks(self, new_blocks):
         self.send_message(headers_message)
 
 class AssumeValidTest(BitcoinTestFramework):
-    def __init__(self):
-        super().__init__()
+    def set_test_params(self):
         self.setup_clean_chain = True
```
promag: It seems there are more `setup_clean_chain = True` so why not make it the default?

promag (commented):
Have you thought of not having any node started automatically? There are some tests that start with `stop_node(0)`..

So each`run_test(self)` the nodes are explicitly started, `num_nodes` can go away, even `setup_clean_chain` would be a `start_node` argument.

test/functional/assumevalid.py
jnewbery: I haven't made any changes to `setup_clean_chain`. See earlier comment:

> I haven't done the same for self.setup_clean_chain, since I think it makes more sense for that to have a default. If people feel strongly, it can be added in a future commit.

Feel free to open a PR to change/remove the default if you think that's necessary. I don't think it's required for this PR.

jnewbery:
Thanks for the reivew @promag 

> Have you thought of not having any node started automatically? There are some tests that start with stop_node(0)

Tests that don't need their nodes to start should override the `setup_network()` or `setup_nodes()` methods. The vast majority of tests require nodes to be started at the top of the test, so that's the default.

> setup_clean_chain would be a start_node argument.

`setup_clean_chain` is a property of the pre-mined chain. I don't think we should change it to be a property of the individual node (if you disagree please open a separate PR to demonstrate the change).

test/functional/bumpfee.py
```diff
@@ -34,21 +34,18 @@ def __init__(self):
...
-        self.nodes = self.start_nodes(self.num_nodes, self.options.tmpdir, extra_args)
-
+    def run_test(self):
         # Encrypt wallet for test_locked_wallet_fails test
         self.nodes[1].node_encrypt_wallet(WALLET_PASSPHRASE)
-        self.nodes[1] = self.start_node(1, self.options.tmpdir, extra_args[1])
+        self.start_node(1)
         self.nodes[1].walletpassphrase(WALLET_PASSPHRASE, WALLET_PASSPHRASE_TIMEOUT)
 
         connect_nodes_bi(self.nodes, 0, 1)
```
MarcoFalke: This is already done by `setup_network`

jnewbery: L41 stops node 1 (the `encryptwallet` RPC shuts down the node), so I believe all the p2p connections are lost and we need to reconnect the nodes after restarting node 1.

I've tested removing these two lines and the test fails (the next call to `sync_all()` fails)

MarcoFalke: Sorry, missed that the node shuts down.

test/functional/pruning.py
```diff
@@ -307,15 +313,15 @@ def has_block(index):
 
         # stop node, start back up with auto-prune at 550MB, make sure still runs
         self.stop_node(node_number)
-        self.nodes[node_number] = self.start_node(node_number, self.options.tmpdir, ["-prune=550"], timewait=900)
+        self.start_node(node_number, extra_args=["-prune=550"])
```
MarcoFalke: You are removing `timewait=900` from all calls but the first. I assume this does not cause issues?

test/functional/wallet.py
```diff
@@ -208,8 +211,10 @@ def run_test(self):
         zeroValueTxid= decRawTx['txid']
         sendResp = self.nodes[1].sendrawtransaction(signedRawTx['hex'])
 
+        # self.sync_all([self.nodes[0:3]])
```
MarcoFalke: debug statement?

test/functional/wallet.py
```diff
@@ -208,8 +211,10 @@ def run_test(self):
...
         self.sync_all()
         self.nodes[1].generate(1) #mine a block
+        # self.sync_all([self.nodes[0:3]])
```
MarcoFalke: debug statement?

test/functional/wallet.py
```diff
@@ -247,14 +254,16 @@ def run_test(self):
...
         self.stop_nodes()
-        self.nodes = self.start_nodes(3, self.options.tmpdir)
+        self.start_node(0)
+        self.start_node(1)
+        self.start_node(2)
         connect_nodes_bi(self.nodes,0,1)
         connect_nodes_bi(self.nodes,1,2)
         connect_nodes_bi(self.nodes,0,2)
-        sync_blocks(self.nodes)
+        sync_blocks(self.nodes[0:2])
```
MarcoFalke: why not `self.nodes[:3]`?

test/functional/wallet.py
```diff
@@ -247,14 +254,16 @@ def run_test(self):
...
 
         self.nodes[0].generate(1)
-        sync_blocks(self.nodes)
+        sync_blocks(self.nodes[0:2])
```
MarcoFalke: why not `self.nodes[:3]`?

test/functional/zapwallettxes.py
```diff
@@ -20,11 +20,9 @@
                                  )
 
 class ZapWalletTXesTest (BitcoinTestFramework):
-
-    def __init__(self):
-        super().__init__()
+    def set_test_params(self):
         self.setup_clean_chain = True
-        self.num_nodes = 2
+        self.num_nodes = 3
```
MarcoFalke: why is this changed in a refactoring commit?

maflcko (commented):
utACK df824a92121eb68849e474e87098a441d4e52926

test/functional/pruning.py
jnewbery: `rpc_timeout` is a property of `TestNode` set when instantiating the python object, not when starting up the bitcoind node.

test/functional/wallet.py
jnewbery: derp. Fixed

test/functional/wallet.py
jnewbery: yep

test/functional/wallet.py
jnewbery: fixed

test/functional/wallet.py
jnewbery: fixed

test/functional/zapwallettxes.py
jnewbery: fixed

jnewbery:
Thanks @MarcoFalke . Review comments addressed.

MarcoFalke:
re-utACK 7148b74dc39110f53c665b94fa9d994c6ad6dc1c

PR #7550, opened 2016-02-17
rpc: Input-from-stdin mode for bitcoin-cli

laanwj:
Implements #7442 by adding an option `-stdin` which reads additional arguments from standard in, one per line.

For example

``` bash
$ echo -e "mysecretcode\n120" | src/bitcoin-cli -stdin walletpassphrase
$ echo -e "walletpassphrase\nmysecretcode\n120" | src/bitcoin-cli -stdin
$ src/bitcoin-cli -stdin walletpassphrase
mysecretcode
120
^D
```

This is the simplest implementation and avoids escaping issues by using newline as separator instead of space, I first had another implementation: https://github.com/laanwj/bitcoin/commit/1f73b8e27b57c8561840cddc9f69a97475d06e85 that reuses `parseCommandLine` from the GUI debug console, but I think this is more useful in practice as most use of cli is probably script-driven.

PR #11125, opened 2017-08-24
Add bitcoin-cli -stdin and -stdinrpcpass functional tests

promag:
This patch adds tests for `bitcoin-cli` options `-stdin` (#7550) and `-stdinrpcpass` #10997.

### DIFF
--- a/test/functional/bitcoin_cli.py
+++ b/test/functional/bitcoin_cli.py
@@ -4,7 +4,7 @@
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 """Test bitcoin-cli"""
 from test_framework.test_framework import BitcoinTestFramework
-from test_framework.util import assert_equal
+from test_framework.util import assert_equal, assert_raises_process_error, get_auth_cookie
 
 class TestBitcoinCli(BitcoinTestFramework):
 
@@ -16,16 +16,24 @@ class TestBitcoinCli(BitcoinTestFramework):
         """Main test logic"""
 
         self.log.info("Compare responses from gewalletinfo RPC and `bitcoin-cli getwalletinfo`")
-        cli_get_info = self.nodes[0].cli.getwalletinfo()
-        rpc_get_info = self.nodes[0].getwalletinfo()
-
-        assert_equal(cli_get_info, rpc_get_info)
+        cli_response = self.nodes[0].cli.getwalletinfo()
+        rpc_response = self.nodes[0].getwalletinfo()
+        assert_equal(cli_response, rpc_response)
 
         self.log.info("Compare responses from getblockchaininfo RPC and `bitcoin-cli getblockchaininfo`")
-        cli_get_info = self.nodes[0].cli.getblockchaininfo()
-        rpc_get_info = self.nodes[0].getblockchaininfo()
+        cli_response = self.nodes[0].cli.getblockchaininfo()
+        rpc_response = self.nodes[0].getblockchaininfo()
+        assert_equal(cli_response, rpc_response)
 
-        assert_equal(cli_get_info, rpc_get_info)
+        user, password = get_auth_cookie(self.nodes[0].datadir)
+
+        self.log.info("Test -stdinrpcpass option")
+        assert_equal(0, self.nodes[0].cli('-rpcuser=%s' % user, '-stdinrpcpass', input=password).getblockcount())
+        assert_raises_process_error(1, "incorrect rpcuser or rpcpassword", self.nodes[0].cli('-rpcuser=%s' % user, '-stdinrpcpass', input="foo").echo)
+
+        self.log.info("Test -stdin and -stdinrpcpass")
+        assert_equal(["foo", "bar"], self.nodes[0].cli('-rpcuser=%s' % user, '-stdin', '-stdinrpcpass', input=password + "\nfoo\nbar").echo())
+        assert_raises_process_error(1, "incorrect rpcuser or rpcpassword", self.nodes[0].cli('-rpcuser=%s' % user, '-stdin', '-stdinrpcpass', input="foo").echo)
 
 if __name__ == '__main__':
     TestBitcoinCli().main()
--- a/test/functional/test_framework/test_node.py
+++ b/test/functional/test_framework/test_node.py
@@ -181,5 +181,10 @@ class TestNodeCLI():
         if named_args:
             p_args += ["-named"]
         p_args += [command] + pos_args + named_args
-        cli_output = subprocess.check_output(p_args, input=self.input, universal_newlines=True)
-        return json.loads(cli_output, parse_float=decimal.Decimal)
+        process = subprocess.Popen(p_args, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE, universal_newlines=True)
+        cli_stdout, cli_stderr = process.communicate(input=self.input)
+        returncode = process.poll()
+        if returncode:
+            # Ignore cli_stdout, raise with cli_stderr
+            raise subprocess.CalledProcessError(returncode, self.binary, output=cli_stderr)
+        return json.loads(cli_stdout, parse_float=decimal.Decimal)
--- a/test/functional/test_framework/test_node.py
+++ b/test/functional/test_framework/test_node.py
@@ -155,8 +155,16 @@ class TestNodeCLI():
     """Interface to bitcoin-cli for an individual node"""
 
     def __init__(self, binary, datadir):
+        self.args = []
         self.binary = binary
         self.datadir = datadir
+        self.input = None
+
+    def __call__(self, *args, input=None):
+        # TestNodeCLI is callable with bitcoin-cli command-line args
+        self.args = [str(arg) for arg in args]
+        self.input = input
+        return self
 
     def __getattr__(self, command):
         def dispatcher(*args, **kwargs):
@@ -169,9 +177,9 @@ class TestNodeCLI():
         pos_args = [str(arg) for arg in args]
         named_args = [str(key) + "=" + str(value) for (key, value) in kwargs.items()]
         assert not (pos_args and named_args), "Cannot use positional arguments and named arguments in the same bitcoin-cli call"
-        p_args = [self.binary, "-datadir=" + self.datadir]
+        p_args = [self.binary, "-datadir=" + self.datadir] + self.args
         if named_args:
             p_args += ["-named"]
         p_args += [command] + pos_args + named_args
-        cli_output = subprocess.check_output(p_args, universal_newlines=True)
+        cli_output = subprocess.check_output(p_args, input=self.input, universal_newlines=True)
         return json.loads(cli_output, parse_float=decimal.Decimal)
--- a/test/functional/test_framework/util.py
+++ b/test/functional/test_framework/util.py
@@ -12,6 +12,7 @@ import logging
 import os
 import random
 import re
+from subprocess import CalledProcessError
 import time
 
 from . import coverage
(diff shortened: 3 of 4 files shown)


test/functional/test_framework/test_node.py
```diff
@@ -136,3 +136,7 @@ def node_encrypt_wallet(self, passphrase):
             time.sleep(0.1)
         self.rpc = None
         self.rpc_connected = False
+
+    def cli(self, args):
+        """Start the cli."""
+        return subprocess.Popen(['bitcoin-cli'] + args, stderr=self.stderr, stdin=subprocess.PIPE, stdout=subprocess.PIPE)
```
laanwj: We need a way to override what bitcoin-cli to use, akin to `BITCOIND` environment variable.

Which reminds me, didn't we have another PR to add bitcoin-cli testing support at some point?

laanwj:
Concept ACK

jnewbery:
Concept ACK, but I recommend you build on top of #10798 which already adds `bitcoin-cli` interface to TestNode.

test/functional/test_framework/test_node.py
MarcoFalke: You might be able to hack something up based on `TestNodeCLI` from current `master` branch.

jnewbery: This should already be possible with #10798 merged

promag:
I'll refactor now that #10798 is merged.

promag:
@jnewbery take a look at f275cf6, it allows to pass `-stdin` and `-stdinrpcpass`. Also, `TestNodeCLI` instances are only created when necessary.

Updated PR title and description accordingly.

jnewbery:
See commit here for an alternative way of doing this: https://github.com/jnewbery/bitcoin/commit/5bfd5e0d47f89f4c2e9becf266876c0c6cc62787

A few advantages:

- bitcoin-cli command line arguments are passed in as a string and fed through to `subprocess`, rather than being parsed by python
- It avoids the cludgy `input` magic key
- it maintains the `-datadir` argument, which I think we want to keep (without that, bitcoin-cli won't have access to the `rpcport` value in the `bitcoin.conf` file).

I've proposed that as a test for #10871. Let me know what you think.

promag:
> It avoids the cludgy input magic key

@jnewbery how do you pass the stdin content?

jnewbery:
ah sorry, I just looked at the commit at https://github.com/bitcoin/bitcoin/commit/f275cf65018681689e711af6bca202125557c434 and misunderstood how `input` was used. I'll re-review.

MarcoFalke:
utACK 5b4e3391f6a7dc9e7148eaee05e285f200f46927

test/functional/bitcoin_cli.py
```diff
@@ -17,10 +17,19 @@ def run_test(self):
...
         self.log.info("Compare responses from getinfo RPC and `bitcoin-cli getinfo`")
-        cli_get_info = self.nodes[0].cli.getinfo()
+        cli_get_info = self.nodes[0].cli().getinfo()
         rpc_get_info = self.nodes[0].getinfo()
-
         assert_equal(cli_get_info, rpc_get_info)
 
+        self.log.info("Test -stdinrpcpass option")
+        user, password = get_auth_cookie(self.nodes[0].datadir)
+        block_count = self.nodes[0].cli(rpcuser=user, rpcport=rpc_port(0), stdinrpcpass=None).getblockcount(input=password)
```
laanwj: I would prefer to make a temporary, empty datadir here instead of make it use the default. The tests should not touch `~/.bitcoin`.

promag:
Agree!

test/functional/test_framework/test_node.py
```diff
@@ -141,12 +139,22 @@ def node_encrypt_wallet(self, passphrase):
         self.rpc = None
         self.rpc_connected = False
 
+    def cli(self, **kwargs):
+        """Creates a configurable cli for the node.
+
+        If arguments is empty then -datadir is set to match the node
+        """
+        args = ["-" + str(key) + ("=" + str(value) if value else "") for (key, value) in kwargs.items()]
+        if not args:
```
jnewbery: I think you should add an error here. If args is specified, then `datadir` must be one of the args (otherwise `bitcoin-cli` will look in `~/.bitcoin`).

test/functional/bitcoin_cli.py
jnewbery: It seems weird to me that `input` is an argument to the RPC, rather than an argument to the bitcoin-cli instance.

It seems possible that an RPC will have `input` as a parameter. Rather less likely that bitcoin-cli will add an argument called `input`.

jnewbery (commented):
Looks generally good to me. A couple of nits inline.

test/functional/test_framework/test_node.py
laanwj: Another option would be to set `datadir=self.datadir` if it isn't overridden otherwise. As you say, there is no reason the argument should ever be *missing*.

test/functional/test_framework/test_node.py
```diff
@@ -51,8 +51,6 @@ def __init__(self, i, dirname, extra_args, rpchost, timewait, binary, stderr, mo
         self.extra_args = extra_args
         self.args = [self.binary, "-datadir=" + self.datadir, "-server", "-keypool=1", "-discover=0", "-rest", "-logtimemicros", "-debug", "-debugexclude=libevent", "-debugexclude=leveldb", "-mocktime=" + str(mocktime), "-uacomment=testnode%d" % i]
 
-        self.cli = TestNodeCLI(os.getenv("BITCOINCLI", "bitcoin-cli"), self.datadir)
```
MarcoFalke: Thinking more about this, I think we should keep `TestNode.cli` to be the corresponding interface to `TestNode.rpc`. The logic to hack a different datadir can be placed in the test that checks for `-stdinrpcpass`. This makes the TestNode code easier to understand, as the datadir overwrite is rarely needed... Likely only once?

test/functional/test_framework/test_node.py
```diff
@@ -155,13 +163,16 @@ def dispatcher(*args, **kwargs):
 
     def send_cli(self, command, *args, **kwargs):
         """Run bitcoin-cli command. Deserializes returned string as python object."""
-
+        input = None
+        if "input" in kwargs:
+            input = kwargs["input"]
+            del kwargs["input"]
```
ryanofsky: In commit "[test] Add support for custom arguments to TestNodeCLI"

You could replace these four lines with `input = kwargs.pop("input", None)`, or change the function signature to `def send_cli(self, command, *args, input=None, **kwargs)` to get the same result.

Probably though it is better just to pass input to the TestNodeCLI constructor as John suggests.

test/functional/test_framework/test_node.py
jnewbery: I think either approach is fine. Making it an error to not include it in args makes it more explicit to the test writer (ie datadir doesn't get set magically for them).

test/functional/test_framework/test_node.py
jnewbery: I also think I prefer keeping `TestNode.cli` as a class (and making it callable as per https://github.com/jnewbery/bitcoin/commit/5bfd5e0d47f89f4c2e9becf266876c0c6cc62787), but I'm happy to be convinced the other way.

@MarcoFalke - I'm not sure I fully understand your comment though. As things currently stand with this PR, the interface for changing `datadir` is to add an argument to `TestNode.cli()`. How would you propose to make that different?

MarcoFalke: In python every member is public, so changing the datadir is as easy as `self.nodes[0].cli.datadir = './new'  # Overwrite datadir, so we pass in rpcuser and password through stdin`

jnewbery: yes, that works. I think I still prefer making `TestNodeCLI` callable and adding `datadir` as an argument.

MarcoFalke: It is not like we need to pass in anything to `TestNodeCLI`. The `datadir` change is only needed for the test in this pull.

jnewbery: Have you looked at https://github.com/jnewbery/bitcoin/commit/5bfd5e0d47f89f4c2e9becf266876c0c6cc62787 ?
Passing arguments to `TestNodeCLI` seems like a good way to pass `bitcoin-cli` command line arguments.

(I'm not arguing forcefully for the change in https://github.com/jnewbery/bitcoin/commit/5bfd5e0d47f89f4c2e9becf266876c0c6cc62787 and I'm happy to be shown a different way to do this)

MarcoFalke: I did not.

This also help with selecting the rpcwallet, so making it callable is more useful than I thought.

MarcoFalke:
@promag Are you still working on this. I think what is missing is setting up the temp datadir and addressing @ryanofsky feedback..

promag:
Thanks all for the great feedback. These are the relevant changes:
- keep `-datadir`
- `input=...` moved to `.cli()`, although `-input` is not `bitcoin-cli` argument 🙄 
- added tests for invalid authentication

test/functional/test_framework/util.py
```diff
@@ -57,18 +58,44 @@ def assert_raises_message(exc, message, fun, *args, **kwds):
...
+        output (string), optional: [a substring of] the process output. Set to None if
+            checking the process output is not required.
+        fun (function): the function to call. This should execute a process.
+        args*: positional arguments for the function.
+        kwds**: named arguments for the function.
+    """
+    try:
+        fun(*args, **kwds)
+    except CalledProcessError as e:
+        if (returncode is not None) and (returncode != e.returncode):
```
jnewbery: You can remove `(returncode is not None)` here and `(output is not None)` below. The only reason I allowed `assert_raises_jsonrpc()` callers to not specify `returncode` or `output` was to be backwards compatible with existing tests which didn't specify those. As far as I'm concerned, tests should always assert return codes and error messages if they're expecting.

(we can probably remove the same from `assert_raises_jsonrpc()` as well in a separate PR now)

test/functional/test_framework/util.py
```diff
@@ -7,6 +7,7 @@
 from base64 import b64encode
 from binascii import hexlify, unhexlify
 from decimal import Decimal, ROUND_DOWN
+from subprocess import CalledProcessError
```
jnewbery: nit: my personal preference is for imports to be sorted by module name (ie `from subprocess...` should be between `import re` and `import time`)

test/functional/test_framework/util.py
```diff
@@ -57,18 +58,44 @@ def assert_raises_message(exc, message, fun, *args, **kwds):
     else:
         raise AssertionError("No exception raised")
 
+def assert_raises_process(returncode, output, fun, *args, **kwds):
```
jnewbery: nit: perhaps `assert_raises_process_error()`?

test/functional/bitcoin_cli.py
```diff
@@ -21,5 +21,15 @@ def run_test(self):
 
         assert_equal(cli_get_info, rpc_get_info)
 
+        user, password = get_auth_cookie(self.nodes[0].datadir)
+
+        self.log.info("Test -stdinrpcpass option")
+        assert_equal(0, self.nodes[0].cli('-rpcuser=%s' % user, '-stdinrpcpass', input=password).getblockcount())
+        assert_raises_process(1, "incorrect rpcuser or rpcpassword", self.nodes[0].cli('-rpcuser=%s' % user, '-stdinrpcpass', input="xpto").echo)
```
jnewbery: Thanks for teaching me some Portuguese: https://pt.wikipedia.org/wiki/X.P.T.O. 🙂

jnewbery (commented):
A few nits inline. Tested ACK ad730f55b182549c5e95d6cc09be4474a1c3152a

test/functional/test_framework/util.py
promag: My bad, that's also my preference.

test/functional/test_framework/util.py
promag: No problem.

test/functional/test_framework/util.py
promag: Great, will do.

PR #11129, opened 2017-08-24
[qa] util: Poll cookie file size before reading

maflcko:
I have been seeing empty cookie files on travis infrastructure, but couldn't reproduce the issue locally. I assume it takes a bit longer to flush the file on travis. Thus, a sleep should work around that.

### DIFF
--- a/test/functional/test_framework/util.py
+++ b/test/functional/test_framework/util.py
@@ -235,8 +235,10 @@ def get_datadir_path(dirname, n):
 def get_auth_cookie(datadir, n):
     user = None
     password = None
-    if os.path.isfile(os.path.join(datadir, "bitcoin.conf")):
-        with open(os.path.join(datadir, "bitcoin.conf"), 'r', encoding='utf8') as f:
+    file_conf = os.path.join(datadir, 'bitcoin.conf')
+    file_cookie = os.path.join(datadir, 'regtest', '.cookie')
+    if os.path.isfile(file_conf):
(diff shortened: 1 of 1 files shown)


test/functional/test_framework/util.py
```diff
@@ -235,17 +235,22 @@ def get_datadir_path(dirname, n):
...
                 if line.startswith("rpcpassword="):
                     assert password is None  # Ensure that there is only one rpcpassword line
                     password = line.split("=")[1].strip("\n")
-    if os.path.isfile(os.path.join(datadir, "regtest", ".cookie")):
-        with open(os.path.join(datadir, "regtest", ".cookie"), 'r') as f:
+    if os.path.isfile(file_cookie):
+        while os.path.getsize(file_cookie) != len('__cookie__:') + 64:
+            # Wait for the file to be flushed
+            time.sleep(0.05)
+        with open(file_cookie, 'r') as f:
```
laanwj: Concept ACK, though I don't like doing this polling in `get_auth_cookie` itself - my preference would be to throw a `AuthCookieIncomplete` error or such, catch that in `wait_for_rpc_connection` and keep the polling restricted to there.

laanwj:
Thinking about this some more, an even better option would be to use POSIX rename atomicity at the `bitcoind` side to create a working cookie atomically:
- Write `.cookie.tmp`, close file
- Rename `.cookie.tmp` to `.cookie`

PR #11131, opened 2017-08-25
rpc: Write authcookie atomically

laanwj:
Use POSIX rename atomicity at the `bitcoind` side to create a working
cookie atomically:

- Write `.cookie.tmp`, close file
- Rename `.cookie.tmp` to `.cookie`

This avoids clients reading invalid/partial cookies as in #11129. As such, this is an alternative to that PR.

### DIFF
--- a/src/rpc/protocol.cpp
+++ b/src/rpc/protocol.cpp
@@ -66,9 +66,14 @@ static const std::string COOKIEAUTH_USER = "__cookie__";
 /** Default name for auth cookie file */
 static const std::string COOKIEAUTH_FILE = ".cookie";
 
-fs::path GetAuthCookieFile()
+/** Get name of RPC authentication cookie file */
+static fs::path GetAuthCookieFile(bool temp=false)
 {
-    fs::path path(gArgs.GetArg("-rpccookiefile", COOKIEAUTH_FILE));
+    std::string arg = gArgs.GetArg("-rpccookiefile", COOKIEAUTH_FILE);
+    if (temp) {
+        arg += ".tmp";
+    }
+    fs::path path(arg);
     if (!path.is_complete()) path = GetDataDir() / path;
     return path;
 }
@@ -84,14 +89,20 @@ bool GenerateAuthCookie(std::string *cookie_out)
      * these are set to 077 in init.cpp unless overridden with -sysperms.
      */
     std::ofstream file;
-    fs::path filepath = GetAuthCookieFile();
-    file.open(filepath.string().c_str());
+    fs::path filepath_tmp = GetAuthCookieFile(true);
+    file.open(filepath_tmp.string().c_str());
     if (!file.is_open()) {
-        LogPrintf("Unable to open cookie authentication file %s for writing\n", filepath.string());
+        LogPrintf("Unable to open cookie authentication file %s for writing\n", filepath_tmp.string());
         return false;
     }
     file << cookie;
     file.close();
+
+    fs::path filepath = GetAuthCookieFile(false);
+    if (!RenameOver(filepath_tmp, filepath)) {
+        LogPrintf("Unable to rename cookie authentication file %s to %s\n", filepath_tmp.string(), filepath.string());
+        return false;
+    }
     LogPrintf("Generated RPC authentication cookie %s\n", filepath.string());
 
     if (cookie_out)
--- a/src/rpc/protocol.h
+++ b/src/rpc/protocol.h
@@ -91,8 +91,6 @@ UniValue JSONRPCReplyObj(const UniValue& result, const UniValue& error, const Un
 std::string JSONRPCReply(const UniValue& result, const UniValue& error, const UniValue& id);
 UniValue JSONRPCError(int code, const std::string& message);
 
-/** Get name of RPC authentication cookie file */
-fs::path GetAuthCookieFile();
 /** Generate a new RPC authentication cookie and write it to disk */
 bool GenerateAuthCookie(std::string *cookie_out);
 /** Read the RPC authentication cookie from disk */


MarcoFalke:
utACK 43e7dcb

On Aug 25, 2017 06:59, "Wladimir J. van der Laan" <notifications@github.com>
wrote:

Use POSIX rename atomicity at the bitcoind side to create a working
cookie atomically:

   - Write .cookie.tmp, close file
   - Rename .cookie.tmp to .cookie

This avoids clients reading invalid/partial cookies as in #11129
<https://github.com/bitcoin/bitcoin/pull/11129>. As such, this is an
alternative to that PR.
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/bitcoin/bitcoin/pull/11131
Commit Summary

   - rpc: Write authcookie atomically

File Changes

   - *M* src/rpc/protocol.cpp
   <https://github.com/bitcoin/bitcoin/pull/11131/files#diff-0> (23)
   - *M* src/rpc/protocol.h
   <https://github.com/bitcoin/bitcoin/pull/11131/files#diff-1> (2)

Patch Links:

   - https://github.com/bitcoin/bitcoin/pull/11131.patch
   - https://github.com/bitcoin/bitcoin/pull/11131.diff

—
<https://github.com/bitcoin/bitcoin/pull/11131>, or mute the thread

promag (commented):
Another option is to use [`fs::unique_path`](http://www.boost.org/doc/libs/1_53_0/libs/filesystem/doc/reference.html#unique_path).

Edit: Concept ACK!

laanwj:
> Another option is to use fs::unique_path.

I wasn't sure that would work in all cases - using a temporary file would potentially move over filesystem boundaries, right?

src/rpc/protocol.cpp
```diff
@@ -66,9 +69,13 @@ static const std::string COOKIEAUTH_USER = "__cookie__";
 /** Default name for auth cookie file */
 static const std::string COOKIEAUTH_FILE = ".cookie";
 
-fs::path GetAuthCookieFile()
+static fs::path GetAuthCookieFile(bool temp)
```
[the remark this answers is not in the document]
laanwj: Yeah, well, find a fool proof and portable way to append to a fs::path and get back me. After browsing stack overflow and boost documentation for a while, I deemed it easier and safer to just do it where the filename is constructed initially.

src/rpc/protocol.cpp
```diff
@@ -15,6 +15,9 @@
 #include <stdint.h>
 #include <fstream>
 
+/** Get name of RPC authentication cookie file */
+static fs::path GetAuthCookieFile(bool temp=false);
```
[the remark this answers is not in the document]
laanwj: Is it only used after the declaration? To minimize diff I didn't want to move the function around and had some compile issues, but that might have been for another reason, will try again.

promag:
I meant `GetDataDir() / fs::unique_path()`.

src/rpc/protocol.cpp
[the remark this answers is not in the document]
promag: I also saw all that, but with `GetDataDir() / fs::unique_path()` none of ths is really necessary.

laanwj:
> I meant GetDataDir() / fs::unique_path().

It's possible to place the cookie file outside of the data directory with an option. Also my gut feeling says it's better to create files with deterministic names in the data directory, what if e.g. boost accidentally generates a name used as a wallet name :) I don't see any advantages.

promag:
> what if e.g. boost accidentally generates a name used as a wallet name :)

I see you take serious precautions 😄

In the documentation there is http://www.boost.org/doc/libs/1_53_0/libs/filesystem/doc/reference.html#path-concatenation.

src/rpc/protocol.cpp
```diff
@@ -84,14 +91,20 @@ bool GenerateAuthCookie(std::string *cookie_out)
...
-    file.open(filepath.string().c_str());
+    fs::path filepath_tmp = GetAuthCookieFile(true);
+    file.open(filepath_tmp.string().c_str());
     if (!file.is_open()) {
-        LogPrintf("Unable to open cookie authentication file %s for writing\n", filepath.string());
+        LogPrintf("Unable to open cookie authentication file %s for writing\n", filepath_tmp.string());
         return false;
     }
     file << cookie;
     file.close();
```
promag: I've not tested but [here](http://www.microhowto.info/howto/atomically_rewrite_the_content_of_a_file.html) says:
> A common misconception is that calling close will necessarily perform an implicit fsync.

From the [docs](http://www.cplusplus.com/reference/fstream/ofstream/close/) there is no explicit mention to flush, only:

> Any pending output sequence is written to the file.

laanwj:
>  http://www.boost.org/doc/libs/1_53_0/libs/filesystem/doc/reference.html#path-concatenation.

Yes, but as I understoof, there is difference between boost_filesystem versions there. I don't really have a lot of time to dive into that, and I don't want this to come back with some strange compile error, and this works so...

src/rpc/protocol.cpp
promag: In other words, call `file.flush()` before close?

laanwj: You're having another misconception there: fsync is not necessary for other processes to see the result! Only to make sure that the data hits disk, in case of crashes etc. Which is not relevant here.

laanwj: Let's ask @theuni just in case.

src/rpc/protocol.cpp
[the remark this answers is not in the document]
MarcoFalke: I thought we always put the documentation above the declaration, but whatever, no need to change it back now.

src/rpc/protocol.cpp
sipa: I think the only concern is that libc may buffer the write as long as the file is open, in which case a flush would be useful to make it visible to the OS (and thus other processes). However, as you're closing the file anyway, I don't believe that matters.

gmaxwell (commented):
Concept ACK

MeshCollider:
utACK https://github.com/bitcoin/bitcoin/pull/11131/commits/82dd7195e1fb943f9cd45a48188f9678219c0206

> I don't want this to come back with some strange compile error, and this works so...

Agreed, this seems fine to me

sipa:
utACK 82dd7195e1fb943f9cd45a48188f9678219c0206

PR #11132, opened 2017-08-25
Document assumptions that are being made to avoid NULL pointer dereferences

practicalswift:
Document assumptions (via `assert(…)`:s) that are being made avoid `NULL` pointer dereferences.

Rationale:
* Make it clear to human reviewers and non-human static analyzers that what might look like potential `NULL` pointer dereferences are written the way they are intentionally (these cases are currently flagged by various static analyzers).

### DIFF
--- a/src/validation.cpp
+++ b/src/validation.cpp
@@ -1710,6 +1710,7 @@ static bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockInd
     // before the first had been spent.  Since those coinbases are sufficiently buried its no longer possible to create further
     // duplicate transactions descending from the known pairs either.
     // If we're on the known chain at height greater than where BIP34 activated, we can save the db accesses needed for the BIP30 check.
+    assert(pindex->pprev);
     CBlockIndex *pindexBIP34height = pindex->pprev->GetAncestor(chainparams.GetConsensus().BIP34Height);
     //Only continue to enforce if we're below BIP34 activation height or the block hash at that height doesn't correspond.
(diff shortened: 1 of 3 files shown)


src/validation.cpp
```diff
@@ -1843,6 +1844,7 @@ static bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockInd
         if (!pblocktree->WriteTxIndex(vPos))
             return AbortNode(state, "Failed to write transaction index");
 
+    assert(pindex->phashBlock && "pindex->phashBlock != nullptr assumed.");
```
sipa: Repeating the condition in a message is not very useful.

src/validation.cpp
```diff
@@ -1704,6 +1704,7 @@ static bool ConnectBlock(const CBlock& block, CValidationState& state, CBlockInd
     // before the first had been spent.  Since those coinbases are sufficiently buried its no longer possible to create further
     // duplicate transactions descending from the known pairs either.
     // If we're on the known chain at height greater than where BIP34 activated, we can save the db accesses needed for the BIP30 check.
+    assert(pindex->pprev && "pindex->pprev != nullptr assumed.");
```
sipa: Repeating the condition in a message is not very useful.

MeshCollider: Agree with @sipa, same thing as #11133

practicalswift:
@sipa Thanks for reviewing! Redundant comments removed. Looks good? :-)

dcousens (approved):

practicalswift:
Pushed an updated version. Now using `cassert` instead of `assert.h`.

laanwj:
utACK fdc3293

PR #11133, opened 2017-08-25
Document assumptions that are being made to avoid division by zero

practicalswift:
Document assumptions (via `assert(…)`:s) that are being made to avoid division by zero.

Rationale:
* Make it clear to human reviewers and non-human static analyzers that what might look like potential division by zero cases are written the way they are intentionally (these cases are currently flagged by various static analyzers).

### DIFF
--- a/src/policy/fees.cpp
+++ b/src/policy/fees.cpp
@@ -503,6 +503,7 @@ void TxConfirmStats::removeTx(unsigned int entryHeight, unsigned int nBestSeenHe
         }
     }
     if (!inBlock && (unsigned int)blocksAgo >= scale) { // Only counts as a failure if not confirmed for entire period
+        assert(scale != 0);
         unsigned int periodsAgo = blocksAgo / scale;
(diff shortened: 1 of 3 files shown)


MeshCollider:
Hmm does adding that text change the behavior of the static analyzers? Otherwise it seems pretty human readable to me already, they aren't exactly complicated asserts :)

sipa:
Agree with @MeshCollider.

danra:
utACK for asserting `!= 0` where added.
Agree with @MeshCollider that the extra text is redundant.

practicalswift:
@MeshCollider @sipa @danra Thanks for reviewing! Redundant comments removed. Looks good? :-)

MeshCollider:
Yep, utACK https://github.com/bitcoin/bitcoin/pull/11133/commits/55509f1a11003837714071d05ea878b340757a76

laanwj:
utACK 55509f1

PR #11135, opened 2017-08-25
Update developer notes with RPC response guidelines

promag:


jnewbery:
ACK - seems like a reasonable suggestion. Agree that this should be a SHOULD rather than a MUST.

mess110:
The main reason I can think of not to return a JSON object is if it returns something raw like a key or a number. Maybe there are also backward compatibility reasons why non JSON output is returned. Are there any other reasons?

Wouldn't it make it easier for clients to always expect a JSON object in the response? It would be future proof as well.

laanwj:
Backward compatibility is a strong reason. We don't change existing interfaces unless there is a a good reason, and just conforming to a developer-notes item is not one.

Also, it's slightly easier to use from the command-line with `bitcoin-cli` if something returns a bare value.

promag:
@laanwj the guideline is `Try to make the RPC response a JSON object`, it's not agains bare values. Beside that, I would bet RPC is mostly used with foreign client libraries than with `bitcoin-cli`.

laanwj:
> I would bet RPC is mostly used with foreign client libraries than with bitcoin-cli

That's not the point, isn't it? despite a bet what use form is more common, many people are using `bitcoin-cli` as a command line client.

promag:
No it isn't for sure, but I think we shouldn't have `bitcoin-cli` in mind when deciding API interface.

As for `bitcoin-cli`, I've previously suggested to have something like `-rpcoutput=...` to ease in those cases, but that's another topic.

Do you think I should improve the guideline?

luke-jr (approved):
ACK

MarcoFalke:
ACK 62ecce7

PR #11136, opened 2017-08-25
Docs: Add python3 to list of dependencies on some platforms

danra:
python3 is required for running the unit tests on macOS.

Without python3 on macOS, 'make check' fails because /test/util/bitcoin-util-test.py fails to find python3.

### DIFF
--- a/doc/build-osx.md
+++ b/doc/build-osx.md
@@ -16,7 +16,7 @@ Then install [Homebrew](https://brew.sh).
 Dependencies
 ----------------------
 
-    brew install automake berkeley-db4 libtool boost --c++11 miniupnpc openssl pkg-config protobuf qt libevent
+    brew install automake berkeley-db4 libtool boost --c++11 miniupnpc openssl pkg-config protobuf python3 qt libevent
 
 If you want to build the disk image with `make deploy` (.dmg / optional), you need RSVG
 
--- a/doc/build-unix.md
+++ b/doc/build-unix.md
@@ -65,7 +65,7 @@ Dependency Build Instructions: Ubuntu & Debian
 ----------------------------------------------
 Build requirements:
 
-    sudo apt-get install build-essential libtool autotools-dev automake pkg-config libssl-dev libevent-dev bsdmainutils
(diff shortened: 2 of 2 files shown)


MarcoFalke:
The dependencies for the functional tests are mentioned in https://github.com/danra/bitcoin/tree/master/test#dependencies

danra:
@MarcoFalke OK, but I don't think it makes sense for build_osx.md to recommend running `make check` on one hand, while not mentioning the dependencies for running it when listing the dependencies earlier.

MarcoFalke:
The readme in src/test mentions that the unit tests can be run with `make check`, but doesn't mention the python3 dependency. We might combine [the src/test readme](https://github.com/danra/bitcoin/blob/3f726c99f819f97f2ab21b94d34c6b3129cd883a/src/test/README.md#compilingrunning-unit-tests) into [the /test/ readme](https://github.com/danra/bitcoin/blob/3f726c99f819f97f2ab21b94d34c6b3129cd883a/test/README.md) and then link to the combined readme from the /doc/ subdir.

laanwj:
Concept ACK, as python 3 is required for the unit tests (not just the functional tests) we should just mention it as dependency.
(it used to be the case that `make check` could be done without python 3)

MarcoFalke:
Fine with me, but please also update the dependencies for ubuntu/fedora

danra:
@MarcoFalke I've only checked the dependencies on my macOS. Should I just add `python3` to the platforms you mentioned as well?

MarcoFalke:
Yes, python3 will be required on the other platforms as well.

MarcoFalke:
Yes, python3 will be required on the other platforms as well.

danra:
@MarcoFalke Done

MarcoFalke:
utACK 7bb5d30

jnewbery:
utACK 7bb5d3000ca4b74c3701cf49eaf63e5a37dfe7ac

fanquake:
utACK 7bb5d30

PR #11138, opened 2017-08-25
Compat: Simplify bswap_16 implementation

danra:
Simplify bswap_16 implementation on platforms which don't already have it defined.
This has no effect on the generated assembly; it just simplifies the source code.


laanwj:
Does this make any difference in the generated code with your compiler?

sipa:
utACK. Verified that in GCC 6.3.0 at `-Os`, `-O2` and `-O3` this does not make a difference in the generated code.

PR #11140, opened 2017-08-25
Trivial: Improve #endif comments

danra:
Improve the #endif comments for the '#if HAVE_DECL_BSWAP_XX == 0' preprocessor conditions, so each shows the full condition which it closes.

### DIFF
--- a/src/compat/byteswap.h
+++ b/src/compat/byteswap.h
@@ -37,7 +37,7 @@ inline uint16_t bswap_16(uint16_t x)
 {
     return (x >> 8) | ((x & 0x00ff) << 8);
 }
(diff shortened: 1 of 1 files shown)


danra:
Please rerun, looks like a spurious error

paveljanik:
ACK https://github.com/bitcoin/bitcoin/pull/11140/commits/ac1cf8d55db8fa9377a77581bfbb774a88e11dd1

promag:
Didn't check but are these the only that can be improved?

danra:
@promag I believe increasing the scope of this to cross-project improvement of endif comments would only result in this pull request never being merged.

PR #11142, opened 2017-08-25
Enhance getinfo with regtest information

wrapperband:
< Bitcoin-cli getinfo > outputs information such as whether the testnet is true. This change adds regtest to the list.


MarcoFalke:
According to the docstring, this information should be retrieved by other means.


```
  * @note Do not add or change anything in the information returned by this
  * method. `getinfo` exists for backwards-compatibility only. It combines
  * information from wildly different sources in the program, which is a mess,
  * and is thus planned to be deprecated eventually.
  *
  * Based on the source of the information, new information should be added to:
  * - `getblockchaininfo`,
  * - `getnetworkinfo` or
  * - `getwalletinfo`

jnewbery:
@wrapperband thanks for trying to help. As Marco says, getinfo is deprecated, so we don't want to add any additional fields to it. `getblockchaininfo()` in src/rpc/blockchain.cpp already returns a more generic `chain` field which tells the user which chain is being used (main, test or regtest).

wrapperband:
Ok, thanks.

PR #11143, opened 2017-08-25
Fix include path for bitcoin-config.h

danra:
All the other files in the repo which include bitcoin-config.h do so with the appropriate subfolder prefixed: config/bitcoin-config.h
The header should be included with the appropriate subfolder here as well.

### DIFF
--- a/src/Makefile.am
+++ b/src/Makefile.am
@@ -18,7 +18,6 @@ else
 LIBUNIVALUE = $(UNIVALUE_LIBS)
 endif
 
-BITCOIN_CONFIG_INCLUDES=-I$(builddir)/config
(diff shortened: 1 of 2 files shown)


laanwj:
If you do this, please also remove the `-I $srcdir/config` from the build system: https://github.com/bitcoin/bitcoin/blob/master/src/Makefile.am#L21, used in https://github.com/bitcoin/bitcoin/blob/master/src/Makefile.am#L253

MarcoFalke:
Please squash your commits when they are ready for merge. See  https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#squashing-commits

danra:
@MarcoFalke Done, thanks

PR #11144, opened 2017-08-25
Move local include to before system includes

danra:
Prevents accidental missing includes and hidden dependencies in the local file.


laanwj:
utACK

fanquake:
utACK eefc2f3

PR #9343, opened 2016-12-13
Don't create change at dust limit

morcos:
@cozz 
@gmaxwell 
@laanwj 

This might be controversial since it directly reverts behavior that was explicitly included in #4331 (merged as #5831)

That PR took the approach that if subtractFeeFromAmount was selected then it would be unexpected for the user to ever spend (net of sent - received) more bitcoins than the amount they requested to pay.  Therefore if any change was under the dust limit, instead of sending it to fee, the change was increased to the dust limit.  

I believe this is misguided.  Creating a change output at the dust limit is likely uneconomical.  There is a difference between having to draw a threshold below which all outputs are useless, and inverting that to say that an output just above that limit is useful.  Giving yourself change at the dust limit is no better (and maybe even worse) than not giving yourself change in the first place, so I don't believe it will adversely affect anyone who was not expecting to spend more coins than the pay amount.

The primary reason for this change is to make our wallet smarter about not creating uselessly small outputs.  To increase the change to the point where it was sufficiently big to be useful would require potentially removing too much from the recipients.

PR #11145, opened 2017-08-25
Fix rounding bug in calculation of minimum change

morcos:
Thanks to @juscamarena for reporting this.

Please backport to 0.15. 

There was a potential rounding error where the fee for the change added to the fee for the original tx could be less than the fee for the tx including change.

This is fixed in the first commit.  The second commit adds one more snippet of information in the fee calculation report.  I actually realized that there is more information that would be nice to report, but we can add that post 0.15.

An open question is whether we should be returning failure if the test in line 2885 is hit or just resetting pick_new_inputs and continuing.  Originally I made it a failure to avoid any possible infinite loops.  But the case hit here is an example of where that logic possibly backfired.

### DIFF
--- a/src/wallet/wallet.cpp
+++ b/src/wallet/wallet.cpp
@@ -2861,13 +2861,15 @@ bool CWallet::CreateTransaction(const std::vector<CRecipient>& vecSend, CWalletT
                     // new inputs. We now know we only need the smaller fee
                     // (because of reduced tx size) and so we should add a
                     // change output. Only try this once.
-                    CAmount fee_needed_for_change = GetMinimumFee(change_prototype_size, coin_control, ::mempool, ::feeEstimator, nullptr);
-                    CAmount minimum_value_for_change = GetDustThreshold(change_prototype_txout, discard_rate);
-                    CAmount max_excess_fee = fee_needed_for_change + minimum_value_for_change;
-                    if (nFeeRet > nFeeNeeded + max_excess_fee && nChangePosInOut == -1 && nSubtractFeeFromAmount == 0 && pick_new_inputs) {
-                        pick_new_inputs = false;
-                        nFeeRet = nFeeNeeded + fee_needed_for_change;
-                        continue;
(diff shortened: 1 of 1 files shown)


gmaxwell (approved):
utACK  (reviewed this earlier based on the report; this is the fix I was expecting.)

instagibbs (approved):
utACK

ryanofsky (commented):
utACK 6af49dddeaeec7f134e86d6f8cf839c55870b7ab.

Some code is shuffled around in the first commit, but only actual change in behavior seems to be computing minimum fee as `GetMinimumFee(nBytes + change_prototype_size + 2)` instead of `GetMinimumFee(nBytes) + GetMinimumFee(change_prototype_size)`.

Unclear whether it's the order of operations change or the +2 bytes which fixes the originally reported bug.

A test would be nice here but I think I say that every time I review CreateTransaction code (#9343, #10712)

gmaxwell:
> Unclear whether it's the order of operations change or the +2 bytes which fixes the originally reported bug.

Both.  GetMinimumFee(nBytes) + GetMinimumFee(change_prototype_size)  has a rounding error compared to GetMinimumFee(nBytes + change_prototype_size)  and the +2 deals with the case that the output count becomes large enough to increase the compact size.

juscamarena:
tACK, I no longer get the specific above issue. I ran into some other thing, that @morcos would probably better explain as he did as well.

sdaftuar:
utACK

kallewoof:
utACK 6af49dddeaeec7f134e86d6f8cf839c55870b7ab

gmaxwell (approved):
ACK.  (upgrading earlier utACK-- I left a node running for a few hours in regtest in a sending stress test and didn't manage to hit any issues.)

MeshCollider:
utACK https://github.com/bitcoin/bitcoin/commit/a54c7b94f8825e9b52fec9066fe7c1d5b6f53482

sipa:
utACK 6af49dddeaeec7f134e86d6f8cf839c55870b7ab

MarcoFalke:
This was backported. Removing label

PR #11149, opened 2017-08-25
Use prefix instead of postfix increment/decrement for non-trivial typ…

danra:
…es where the result isn't used (and for trivial types as well in the changed files, for consistency).

Follow Scott Meyers' advice in his More Effective C++ book, as well as cppcheck's performance messages, and use prefix instead of postfix increment/decrement for objects of non-trivial type, whenever the result of the increment isn't used, to avoid possible extra copies and maximize performance.

All the modified files fix this for some objects of non-trivial type. Increments/decrements of trivial types in these files were also converted to prefix instead of postfix (wherever the result isn't used) for consistency.

### DIFF
--- a/src/addrman.cpp
+++ b/src/addrman.cpp
@@ -121,7 +121,7 @@ void CAddrMan::Delete(int nId)
     vRandom.pop_back();
     mapAddr.erase(info);
     mapInfo.erase(nId);
-    nNew--;
+    --nNew;
 }
 
 void CAddrMan::ClearNew(int nUBucket, int nUBucketPos)
@@ -131,7 +131,7 @@ void CAddrMan::ClearNew(int nUBucket, int nUBucketPos)
         int nIdDelete = vvNew[nUBucket][nUBucketPos];
         CAddrInfo& infoDelete = mapInfo[nIdDelete];
         assert(infoDelete.nRefCount > 0);
-        infoDelete.nRefCount--;
+        --infoDelete.nRefCount;
         vvNew[nUBucket][nUBucketPos] = -1;
         if (infoDelete.nRefCount == 0) {
             Delete(nIdDelete);
@@ -142,14 +142,14 @@ void CAddrMan::ClearNew(int nUBucket, int nUBucketPos)
 void CAddrMan::MakeTried(CAddrInfo& info, int nId)
 {
     // remove the entry from all new buckets
-    for (int bucket = 0; bucket < ADDRMAN_NEW_BUCKET_COUNT; bucket++) {
+    for (int bucket = 0; bucket < ADDRMAN_NEW_BUCKET_COUNT; ++bucket) {
         int pos = info.GetBucketPosition(nKey, true, bucket);
         if (vvNew[bucket][pos] == nId) {
             vvNew[bucket][pos] = -1;
-            info.nRefCount--;
+            --info.nRefCount;
         }
     }
-    nNew--;
+    --nNew;
 
     assert(info.nRefCount == 0);
 
@@ -167,7 +167,7 @@ void CAddrMan::MakeTried(CAddrInfo& info, int nId)
         // Remove the to-be-evicted item from the tried set.
         infoOld.fInTried = false;
         vvTried[nKBucket][nKBucketPos] = -1;
-        nTried--;
+        --nTried;
 
         // find which new bucket it belongs to
         int nUBucket = infoOld.GetNewBucket(nKey);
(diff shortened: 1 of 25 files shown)


src/arith_uint256.h
```diff
@@ -33,21 +33,21 @@ class base_uint
     {
         static_assert(BITS/32 > 0 && BITS%32 == 0, "Template parameter BITS must be a positive multiple of 32.");
 
-        for (int i = 0; i < WIDTH; i++)
```
luke-jr: While you're touching these, can you add braces too (in a separate commit in the same PR)?

src/net.cpp
```diff
@@ -2384,13 +2384,13 @@ void CConnman::Interrupt()
     InterruptSocks5(true);
 
     if (semOutbound) {
-        for (int i=0; i<(nMaxOutbound + nMaxFeeler); i++) {
```
luke-jr: Add some spaces here.

luke-jr (approved):
utACK. Was there a script to automate this?

Maybe update the code format guidelines too...

promag (commented):
IMO this can turn to be one of the never merge PR. I agree with the change but I'm afraid this may force lots of rebases and there isn't that much value on the change.

danra:
@promag Would this be more manageable if this was split into single-file commits? I wouldn't mind doing that, but I'm just started out with pull requests to this repo and so far I got the impression that multiple small pull requests would be frowned upon.

danra:
@luke-jr cppcheck pointed to all the non-trivial object increments which should be optimized. Beyond that, there was no automation. Automation of this would probably require building a clang module or similar, since one can't blindly change all postfixes to prefixes, only those where the result is not used.

promag:
I think everyone agrees if you add a developer note with the advice, and keep an eye on PR's that add or touch code with postfix inc/dec.

As for this PR, wait for more feedback. If you could make this scripted then I wouldn't mind if it doesn't fix the code style as @luke-jr suggests.

danra:
@promag There is actually already such a note in the developer notes:

> ++i is preferred over i++.

Should I expand on the reasoning for this in the developer notes?

src/addrman.cpp
```diff
@@ -277,14 +277,14 @@ bool CAddrMan::Add_(const CAddress& addr, const CNetAddr& source, int64_t nTimeP
 
         // stochastic test: previous nRefCount == N: 2^N times harder to increase it
         int nFactor = 1;
-        for (int n = 0; n < pinfo->nRefCount; n++)
+        for (int n = 0; n < pinfo->nRefCount; ++n)
```
MarcoFalke: In the subject you mention "for non-trivial types". I always thought that int was a trivial type.

danra: I've explained this in the commit message body:

> All the modified files fix this for some objects of non-trivial type. Increments/decrements of trivial types in these files were also converted to prefix instead of postfix (wherever the result isn't used) for consistency.

But, I'll change the subject to reflect that as well.

achow101:
I tend toward NACKing this. It is a lot of code churn and will likely require a lot of other PRs to be rebased and work through a lot of merge conflict stuff. Also, shouldn't the compiler be optimizing out all of those anyways so it doesn't really matter?

danra:
@achow101 
>  Also, shouldn't the compiler be optimizing out all of those anyways so it doesn't really matter?

The experts say no. Isn't Scott Meyers' advice authoritative enough? :)

Also, see https://stackoverflow.com/questions/30036749/is-it-still-better-to-prefer-pre-increment-over-post-increment

My favorite quote from there:

> The point is, the compiler is free to change your i++ to ++i if it can prove that such transformation wont change the behavior of the program. For built-in types, proofs are easy. For class types, it could be really difficult (or practically expensive)!

promag:
NACK, although it's a better code and all, unless you show numbers of performance increase.

laanwj:
This should certainly be done in new code, but not sure about changing this for the entire existing code at once. Seems too much.

Do you see any differences in benchmarks/profiles to warrant this?

danra:
@laanwj I don't. Just thought that fixing some cppcheck performance messages as well as following the developer guidelines would be reason enough to introduce this fix.

MeshCollider:
Yeah because there are no real benefits to the change other than code style changes, its not worth the merge issues and even small risks imo, so I'd have to give this a NACK. Thanks though, nice in theory :)

MarcoFalke:
The last time someone tried those changes, we compared the binaries before
and after and nothing changed. Implying that the compiler can handle this.

laanwj:
Ok, closing this then. Sorry.

PR #11150, opened 2017-08-25
[tests] Add getmininginfo test

mess110:
Add `getmininginfo` functional test in `mining.py`

### DIFF
--- a/test/functional/mining.py
+++ b/test/functional/mining.py
@@ -4,16 +4,18 @@
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 """Test mining RPCs
 
+- getmininginfo
 - getblocktemplate proposal mode
 - submitblock"""
 
-from binascii import b2a_hex
 import copy
+from binascii import b2a_hex
+from decimal import Decimal
 
 from test_framework.blocktools import create_coinbase
-from test_framework.test_framework import BitcoinTestFramework
 from test_framework.mininode import CBlock
-from test_framework.util import *
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal, assert_raises_jsonrpc
 
 def b2x(b):
     return b2a_hex(b).decode('ascii')
@@ -33,6 +35,18 @@ class MiningTest(BitcoinTestFramework):
 
     def run_test(self):
         node = self.nodes[0]
+
+        self.log.info('getmininginfo')
+        mining_info = node.getmininginfo()
+        assert_equal(mining_info['blocks'], 200)
+        assert_equal(mining_info['chain'], 'regtest')
+        assert_equal(mining_info['currentblocksize'], 0)
+        assert_equal(mining_info['currentblocktx'], 0)
+        assert_equal(mining_info['currentblockweight'], 0)
+        assert_equal(mining_info['difficulty'], Decimal('4.656542373906925E-10'))
+        assert_equal(mining_info['networkhashps'], Decimal('0.003333333333333334'))
+        assert_equal(mining_info['pooledtx'], 0)
+
         # Mine a block to leave initial block download
         node.generate(1)
         tmpl = node.getblocktemplate()


test/functional/rpc_getmininginfo.py
```diff
@@ -0,0 +1,67 @@
...
+# Copyright (c) 2014-2016 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Test the getmininginfo API."""
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal
+from decimal import Decimal
+
+
```
promag: Remove newline.

test/functional/rpc_getmininginfo.py
```diff
@@ -0,0 +1,67 @@
...
+class GetMiningInfoTest(BitcoinTestFramework):
+    def __init__(self):
+        super().__init__()
+        self.num_nodes = 1
+        self.setup_clean_chain = False
+
+
```
promag: Remove newline.

test/functional/rpc_getmininginfo.py
```diff
@@ -0,0 +1,67 @@
...
+    def run_test(self):
+        self.mining_info = self.nodes[0].getmininginfo()
+
+        self._test_returns_all_the_keys()
+        self._test_values()
+        self._test_blocks_key_is_equal_to_getblockcount()
+
+
```
promag: Remove newline.

test/functional/rpc_getmininginfo.py
```diff
@@ -0,0 +1,67 @@
...
+            'difficulty',
+            'errors',
+            'networkhashps',
+            'pooledtx',
+        ]
+
+        for key in required_keys:
+            assert(key in self.mining_info)
+
+
```
promag: Remove newline.

test/functional/rpc_getmininginfo.py
```diff
@@ -0,0 +1,67 @@
...
+        assert_equal(self.mining_info['chain'], 'regtest')
+        assert_equal(self.mining_info['currentblocksize'], 0)
+        assert_equal(self.mining_info['currentblocktx'], 0)
+        assert_equal(self.mining_info['currentblockweight'], 0)
+        assert_equal(self.mining_info['difficulty'], difficulty)
+        assert_equal(self.mining_info['errors'], errors)
+        assert_equal(self.mining_info['networkhashps'], networkhashps)
+        assert_equal(self.mining_info['pooledtx'], 0)
+
+
```
promag: Remove newline.

test/functional/rpc_getmininginfo.py
```diff
@@ -0,0 +1,67 @@
...
+    def _test_blocks_key_is_equal_to_getblockcount(self):
+        assert_equal(self.nodes[0].getblockcount(), 200)
+        assert_equal(self.mining_info['blocks'], self.nodes[0].getblockcount())
+
+
```
promag: Remove newline.

test/functional/rpc_getmininginfo.py
```diff
@@ -0,0 +1,67 @@
...
+class GetMiningInfoTest(BitcoinTestFramework):
+    def __init__(self):
+        super().__init__()
+        self.num_nodes = 1
+        self.setup_clean_chain = False
+
+
+    def run_test(self):
+        self.mining_info = self.nodes[0].getmininginfo()
```
promag: Don't create `self.mining_info`, pass to functions below?

test/functional/rpc_getmininginfo.py
```diff
@@ -0,0 +1,67 @@
...
+
+        self._test_returns_all_the_keys()
+        self._test_values()
+        self._test_blocks_key_is_equal_to_getblockcount()
+
+
+    def _test_returns_all_the_keys(self):
```
promag: There are few with `_` prefix, remove?

test/functional/rpc_getmininginfo.py
```diff
@@ -0,0 +1,67 @@
...
+            'networkhashps',
+            'pooledtx',
+        ]
+
+        for key in required_keys:
+            assert(key in self.mining_info)
+
+
+    def _test_values(self):
+        difficulty = Decimal('4.656542373906925E-10')
```
promag: Inline below?

test/functional/rpc_getmininginfo.py
```diff
@@ -0,0 +1,67 @@
...
+        networkhashps = Decimal('0.003333333333333334')
```
promag: Same a above?

mess110:
@promag thanks for taking a look. Obviously won't leave 2 spaces in the future :smile:

test/functional/rpc_getmininginfo.py
```diff
@@ -0,0 +1,59 @@
+#!/usr/bin/env python3
+# Copyright (c) 2014-2017 The Bitcoin Core developers
+# Distributed under the MIT software license, see the accompanying
+# file COPYING or http://www.opensource.org/licenses/mit-license.php.
+"""Test the getmininginfo API."""
+
+from test_framework.test_framework import BitcoinTestFramework
+from test_framework.util import assert_equal
+from decimal import Decimal
```
jnewbery: supernit: place imports in PEP-8 ordering (std library first, then project imports)

test/functional/rpc_getmininginfo.py
```diff
@@ -0,0 +1,59 @@
...
+            'errors',
+            'networkhashps',
+            'pooledtx',
+        ]
+
+        for key in required_keys:
+            assert(key in mining_info)
+
+    def test_values(self, mining_info):
+        errors = 'This is a pre-release test build - use at your own risk -' \
```
jnewbery: The This is a pre-release test build - use at your own risk` error is only presented on non-release branches, so this assert will fail on all release branches. Just omit it.

jnewbery (commented):
Thanks for this @mess110 . As an alternative to making this an individual test case (which adds a few seconds overhead to running the test suite), you could add this as a sub-test at the top of the `mining.py` test script.

Whether or not to add a new test script or extend an old one is a judgement call. In this case, I think the `getmininginfo` RPC can be run at the start of `mining.py` without any side-effects so it's a good candidate for extending the existing test.

test/functional/rpc_getmininginfo.py
```diff
@@ -0,0 +1,59 @@
...
+            'currentblocktx',
+            'currentblockweight',
+            'difficulty',
+            'errors',
+            'networkhashps',
+            'pooledtx',
+        ]
+
+        for key in required_keys:
+            assert(key in mining_info)
```
MarcoFalke: No need to assert for keys in our tests. Python has you covered by throwing a KeyError in case you access a key that does not exist.

mess110:
@jnewbery thanks. I initially started in `mining.py` but I decided not to put it there because it would be "cleaner" to have its own test. In this particular case, I agree, the extra time to setup the test is not worth it

@MarcoFalke thanks for the review and hint

jnewbery:
Looks good to me. We could test the actual values returned by `getmininginfo` more thoroughly (ie are the values for size, weight, hashps, etc correct under different circumstances) in a different PR. Testing the interface correctness in this PR is better than no testing at all.

Tested ACK 4f2905b76b502a9de235dbe739e5fe504ef5bc22

test/functional/mining.py
```diff
@@ -4,16 +4,18 @@
...
 """Test mining RPCs
 
+- getmininginfo
 - getblocktemplate proposal mode
 - submitblock"""
 
-from binascii import b2a_hex
 import copy
+from binascii import b2a_hex
+from decimal import Decimal
```
promag: Nit, remove empty lines?

promag (commented):
utACK 4f2905b.

test/functional/mining.py
promag: Ignore as per PEP-8.

PR #11151, opened 2017-08-26
Fix header guards using reserved identifiers

danra:
Identifiers beginning with an underscore followed immediately by an uppercase letter are reserved.


gmaxwell (approved):
ACK. Oops. I should have caught this, I know they are reserved.

PR #11153, opened 2017-08-26
Add meshcollider Gitian PGP key

meshcollider:
Public PGP key used to sign Gitian builds, e.g. https://github.com/bitcoin-core/gitian.sigs/pull/566


laanwj:
Checked it's the same key used for the gitian signatures pull.
BTW: mind uploading this to a keyserver as well? I was unable to fetch it from there by id.

MeshCollider:
Yep, done :) It's also on keybase here: keybase.io/meshcollider

laanwj:
utack 8897b1b

PR #11156, opened 2017-08-26
Fix memory leaks in qt/guiutil.cpp

danra:
on macOS:
`listSnapshot` was leaking in `findStartupItemInList()`
`bitcoinAppUrl` was leaking in `[Get|Set]StartOnSystemStartup()`

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -780,47 +780,64 @@ bool SetStartOnSystemStartup(bool fAutoStart)
 LSSharedFileListItemRef findStartupItemInList(LSSharedFileListRef list, CFURLRef findUrl);
 LSSharedFileListItemRef findStartupItemInList(LSSharedFileListRef list, CFURLRef findUrl)
 {
-    // loop through the list of startup items and try to find the bitcoin app
     CFArrayRef listSnapshot = LSSharedFileListCopySnapshot(list, nullptr);
+    if (listSnapshot == nullptr) {
+        return nullptr;
+    }
+    
+    // loop through the list of startup items and try to find the bitcoin app
     for(int i = 0; i < CFArrayGetCount(listSnapshot); i++) {
         LSSharedFileListItemRef item = (LSSharedFileListItemRef)CFArrayGetValueAtIndex(listSnapshot, i);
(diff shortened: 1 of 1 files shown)


promag:
`CFArrayGetCount` dereferences `listSnapshot` and if it's `nullptr` then it crashes, so how have you tested this?

danra:
@promag Not sure I understand the question, do you mean
+    if (listSnapshot == nullptr) {
 +        return nullptr;
 +    }
added at 784-786?

This is just for safety, it's not really expected for `listSnapshot` to be `nullptr`. I think it makes sense to add this safety in the context of not doing `CFRelease` on `nullptr` at the end. You're right that even without the added lines above, in case of `listSnapshot == nullptr` there will probably be a crash even before we get to the `CFRelease`. Still, seems cleaner to me. 

Also, in the unlikely case that `listSnapshot` *is* set to `nullptr`, we will return `nullptr` instead of crashing, which is probably nicer behavior anyway. (though, it's up to the caller whether it itself would crash or not)

ryanofsky (commented):
utACK 9b348ff9eb75e04249dfa873a34e0d1a31993b9c. Seems right, though maybe it would be simpler with an RAII helper class:

```
struct CFReleaser
{
    CFReleaser(CFTypeRef cf) m_cf(cf) {}
    ~CFReleaser() { if (m_cf) CFRelease(m_cf); }
    CFTypeRef m_cf;
};
```

danra:
@ryanofsky I agree, just didn't want to do two things at once - introduce such a class as well as fix problematic code.
Once this is merged I'm for doing a separate pull requests with a commit that introduces such a class, followed by commits which make use of it.

jonasschnelli:
utACK 9b348ff9eb75e04249dfa873a34e0d1a31993b9c.
I think we should also get rid of the OSX Growl support.

laanwj:
I really like the idea of using a RAII class here. Much easier to check whether all possible exit paths are covered, and exception-safety comes for free.

jonasschnelli:
I agree an RAII approach would be much better.
But until someone comes up with an RAII for that pure OSX stuff, I think this is an valid improvement.

Tested ACK 9b348ff9eb75e04249dfa873a34e0d1a31993b9c.
Could not find any leaks with macOS instruments (leak detector) during notifications.

PR #11160, opened 2017-08-26
Improve versionbits_computeblockversion test code consistency

danra:
In this test, `nTime` is used for all the calls to `Mine()`, each time being set to the correct time beforehand, except for in the last few calls to `Mine()` where `nStartTime` is used directly, even though `nTime` is still set to `nStartTime` beforehand. `nTime` just remains unused for these last few calls to `Mine()`.

Changed the last few calls to `Mine()` to use `nTime` instead, improving consistency. This also fixes an unused value static analyzer warning about `nTime` being set to a value which is never used.

### DIFF
--- a/src/test/versionbits_tests.cpp
+++ b/src/test/versionbits_tests.cpp
@@ -314,20 +314,20 @@ BOOST_AUTO_TEST_CASE(versionbits_computeblockversion)
 
     // Mine one period worth of blocks, and check that the bit will be on for the
     // next period.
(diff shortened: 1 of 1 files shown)


promag:
PR title longer than the patch :trollface:

danra:
looks like a spurious failure, please rerun

sipa:
utACK 81d4f3c50c505afe2f8352f8614c6490e3b65699, but can you use a shorter commit title (the first line of the commit message is treated specially, and shouldn't be too long, as it goes into release notes etc).

danra:
@promag @sipa Shortened title :)

MarcoFalke:
utACK 946638d

PR #11161, opened 2017-08-26
Remove redundant explicitly defined copy ctors

danra:
CFeeRate and CTxMemPoolEntry have explicitly defined copy ctors which has the same functionality as the implicit default copy ctors which would have been generated otherwise.

Besides being redundant, it violates the rule of three (see https://en.wikipedia.org/wiki/Rule_of_three_(C%2B%2B_programming) ).
(Of course, the rule of three doesn't -really- cause a resource management issue here, but the reason for that is exactly that there is no need for an explicit copy ctor in the first place since no resources are being managed).

### DIFF
--- a/src/policy/feerate.h
+++ b/src/policy/feerate.h
@@ -26,7 +26,6 @@ public:
     explicit CFeeRate(const CAmount& _nSatoshisPerK): nSatoshisPerK(_nSatoshisPerK) { }
     /** Constructor for a fee rate in satoshis per kB. The size in bytes must not exceed (2^63 - 1)*/
     CFeeRate(const CAmount& nFeePaid, size_t nBytes);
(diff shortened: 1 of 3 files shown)


promag:
Remove periods fro PR title and commit message.

Otherwise trivial ACK 8a14a86.

danra:
@promag Done

promag:
Please rebase (remove merge commit).

danra:
@promag Done

sipa:
utACK f92b20a85e37b2ce91cc48fa9d7a1f29e353ff13

MarcoFalke:
Please squash your commits according to https://github.com/bitcoin/bitcoin/blob/master/CONTRIBUTING.md#squashing-commits.

No need for two commits on a 8-line-changed patch doing the same thing.

danra:
@MarcoFalke Done

laanwj:
utACK b426e24

PR #11163, opened 2017-08-26
Remove redundant explicitly defined CTxMemPoolEntry copy ctor

danra:
CTxMemPoolEntry has an explicitly defined copy ctor which has the same functionality as the implicit default copy ctor which would have been generated otherwise.

Besides being redundant, it violates the rule of three (see https://en.wikipedia.org/wiki/Rule_of_three_(C%2B%2B_programming) ).
The rule of three doesn't -really- cause a resource management issue here, but the reason for that is exactly that there is no need for an explicit copy ctor in the first place since no resources are being managed.


promag:
This is the same as #11161, maybe add this commit there and make the PR generic?

danra:
Will merge into #11161.

PR #7302, opened 2016-01-05
C++11 build/runtime fixes

theuni:
In preparation for switching to the c++11 requirement, I've done depends builds on all supported OSs and fixed up any snags along the way.

Note that this does not enable c++11 anywhere, it simply enables c++11 builds to pass.

Once merged, we can begin moving forward with enabling.

PR #11164, opened 2017-08-26
Fix boost headers included as user instead of system headers

danra:
In most of the project, boost headers are included as system headers.
Fix the few inconsistent places where they aren't.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -827,14 +827,14 @@ TEMP_CPPFLAGS="$CPPFLAGS"
 CPPFLAGS="$CPPFLAGS $BOOST_CPPFLAGS"
 AC_MSG_CHECKING([for mismatched boost c++11 scoped enums])
 AC_LINK_IFELSE([AC_LANG_PROGRAM([[
-  #include "boost/config.hpp"
-  #include "boost/version.hpp"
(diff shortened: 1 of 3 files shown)


laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/11164/commits/5ac072caa242d7ecf724e0c3a23f4a7c477a3a1e

fanquake:
utACK 5ac072c

configure.ac
```diff
@@ -827,14 +827,14 @@ TEMP_CPPFLAGS="$CPPFLAGS"
 CPPFLAGS="$CPPFLAGS $BOOST_CPPFLAGS"
 AC_MSG_CHECKING([for mismatched boost c++11 scoped enums])
 AC_LINK_IFELSE([AC_LANG_PROGRAM([[
-  #include "boost/config.hpp"
-  #include "boost/version.hpp"
+  #include <boost/config.hpp>
```
laanwj: @theuni as this is part of a check, I wonder if there was a specific resaon for using `"` instead of `<`?

fanquake: For reference this code came from #7302

Issue #10924, opened 2017-07-25
[RPC][Wallet][Segwit] Bug: Transaction sent to imported P2WSH does not appear in listtransaction.


NicolasDorier:
It seems that there is a bug when a node sends coins to an imported P2WSH.
Such transaction does not appear in `listtransaction` unlike P2SH.

Note that if node A sends to node B where B imported the P2WSH, things works as expected, and both A and B would see the transaction in `listtransaction`.

The stangest thing is that `gettransaction` seems to know about the transaction. (but not the details)

Repro:
https://github.com/NicolasDorier/bitcoin/commits/importaddresssegwit
If you switch `use_p2wsh` to false, it passes, else, it does not. 
Failing on `assert(self.nodes[0].listtransactions("*", 1, 0, True)[0]["txid"]  == txid)`

ping @sipa this is the bug I bumped on TumbleBit that I explained to you.

sipa:
A short update (thanks @NicolasDorier): there is no CTxDestination for native witness outputs, which means any `importaddress` is unable to mark the importer scriptPubKey as anything but change. This leads to inconsistent `gettransaction`/`listtransactions` output.

There are 2 solutions here:
* make `importaddress` fail to import addresses with no `CTxDestination` (or warn at least)
* Add BIP173 support

sipa:
Actually, (1) is already warned about in the `importaddress` RPC help text.

NicolasDorier:
my attempt to fix the issue, https://github.com/bitcoin/bitcoin/pull/10947 it would need to have bech32 support though.

ryanofsky:
Should #10947 be tagged for 0.15, or should this be untagged?

NicolasDorier:
Should not be tagged for 0.15. I think the best is to wait next version.

MarcoFalke:
Cleared milestone 0.15.

NicolasDorier:
Closing as this was solved by #11167

PR #11089, opened 2017-08-18
Enable various p2sh-p2wpkh functionality

luke-jr:
A subset of #9017

PR #11167, opened 2017-08-26
Full BIP173 (Bech32) support

sipa:
Builds on top of #11117.

This adds support for:
* Creating BIP173 addresses for testing (through `addwitnessaddress`, though by default it still produces P2SH versions)
* Sending to BIP173 addresses (including non-v0 ones)
* Analysing BIP173 addresses (through `validateaddress`)

It includes a reformatted version of the [C++ Bech32 reference code](https://github.com/sipa/bech32/tree/master/ref/c%2B%2B) and an independent implementation of the address encoding/decoding logic (integrated with CTxDestination). All BIP173 test vectors are included.

Not included (and intended for other PRs):
* Full wallet support for SegWit (which would include automatically adding witness scripts to the wallet during automatic keypool topup, SegWit change outputs, ...) [see #11403]
* Splitting base58.cpp and tests/base58_tests.cpp up into base58-specific code, and "address encoding"-code [see #11372]
* Error locating in UI for BIP173 addresses.

### DIFF
--- a/src/base58.cpp
+++ b/src/base58.cpp
@@ -5,15 +5,16 @@
 #include "base58.h"
 
 #include "hash.h"
+#include "script/script.h"
 #include "uint256.h"
 
-#include <assert.h>
-#include <stdint.h>
-#include <string.h>
-#include <vector>
-#include <string>
 #include <boost/variant/apply_visitor.hpp>
 #include <boost/variant/static_visitor.hpp>
+
+#include <algorithm>
+#include <assert.h>
+#include <string.h>
+
 
 /** All alphanumeric characters except for "0", "I", "O", and "l" */
 static const char* pszBase58 = "123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz";
@@ -212,86 +213,55 @@ int CBase58Data::CompareTo(const CBase58Data& b58) const
 
 namespace
 {
-/** base58-encoded Bitcoin addresses.
- * Public-key-hash-addresses have version 0 (or 111 testnet).
- * The data vector contains RIPEMD160(SHA256(pubkey)), where pubkey is the serialized public key.
- * Script-hash-addresses have version 5 (or 196 testnet).
- * The data vector contains RIPEMD160(SHA256(cscript)), where cscript is the serialized redemption script.
- */
-class CBitcoinAddress : public CBase58Data {
-public:
-    bool Set(const CKeyID &id);
-    bool Set(const CScriptID &id);
-    bool Set(const CTxDestination &dest);
-    bool IsValid() const;
-    bool IsValid(const CChainParams &params) const;
-
-    CBitcoinAddress() {}
-    CBitcoinAddress(const CTxDestination &dest) { Set(dest); }
-    CBitcoinAddress(const std::string& strAddress) { SetString(strAddress); }
-    CBitcoinAddress(const char* pszAddress) { SetString(pszAddress); }
-
-    CTxDestination Get() const;
-};
-
-class CBitcoinAddressVisitor : public boost::static_visitor<bool>
+class DestinationEncoder : public boost::static_visitor<std::string>
 {
 private:
-    CBitcoinAddress* addr;
+    const CChainParams& m_params;
 
 public:
-    explicit CBitcoinAddressVisitor(CBitcoinAddress* addrIn) : addr(addrIn) {}
+    DestinationEncoder(const CChainParams& params) : m_params(params) {}
 
-    bool operator()(const CKeyID& id) const { return addr->Set(id); }
-    bool operator()(const CScriptID& id) const { return addr->Set(id); }
-    bool operator()(const CNoDestination& no) const { return false; }
+    std::string operator()(const CKeyID& id) const
+    {
+        std::vector<unsigned char> data = m_params.Base58Prefix(CChainParams::PUBKEY_ADDRESS);
+        data.insert(data.end(), id.begin(), id.end());
+        return EncodeBase58Check(data);
+    }
+
+    std::string operator()(const CScriptID& id) const
+    {
+        std::vector<unsigned char> data = m_params.Base58Prefix(CChainParams::SCRIPT_ADDRESS);
+        data.insert(data.end(), id.begin(), id.end());
+        return EncodeBase58Check(data);
+    }
+
+    std::string operator()(const CNoDestination& no) const { return ""; }
 };
 
+CTxDestination DecodeDestination(const std::string& str, const CChainParams& params)
+{
+    std::vector<unsigned char> data;
+    uint160 hash;
+    if (DecodeBase58Check(str, data)) {
+        // base58-encoded Bitcoin addresses.
+        // Public-key-hash-addresses have version 0 (or 111 testnet).
+        // The data vector contains RIPEMD160(SHA256(pubkey)), where pubkey is the serialized public key.
+        const std::vector<unsigned char>& pubkey_prefix = params.Base58Prefix(CChainParams::PUBKEY_ADDRESS);
+        if (data.size() == hash.size() + pubkey_prefix.size() && std::equal(pubkey_prefix.begin(), pubkey_prefix.end(), data.begin())) {
+            std::copy(data.begin() + pubkey_prefix.size(), data.end(), hash.begin());
+            return CKeyID(hash);
+        }
+        // Script-hash-addresses have version 5 (or 196 testnet).
+        // The data vector contains RIPEMD160(SHA256(cscript)), where cscript is the serialized redemption script.
+        const std::vector<unsigned char>& script_prefix = params.Base58Prefix(CChainParams::SCRIPT_ADDRESS);
+        if (data.size() == hash.size() + script_prefix.size() && std::equal(script_prefix.begin(), script_prefix.end(), data.begin())) {
+            std::copy(data.begin() + script_prefix.size(), data.end(), hash.begin());
+            return CScriptID(hash);
+        }
+    }
+    return CNoDestination();
+}
 } // namespace
-
-bool CBitcoinAddress::Set(const CKeyID& id)
-{
-    SetData(Params().Base58Prefix(CChainParams::PUBKEY_ADDRESS), &id, 20);
-    return true;
-}
-
-bool CBitcoinAddress::Set(const CScriptID& id)
-{
-    SetData(Params().Base58Prefix(CChainParams::SCRIPT_ADDRESS), &id, 20);
-    return true;
-}
-
-bool CBitcoinAddress::Set(const CTxDestination& dest)
-{
-    return boost::apply_visitor(CBitcoinAddressVisitor(this), dest);
-}
-
-bool CBitcoinAddress::IsValid() const
-{
-    return IsValid(Params());
-}
-
-bool CBitcoinAddress::IsValid(const CChainParams& params) const
-{
-    bool fCorrectSize = vchData.size() == 20;
-    bool fKnownVersion = vchVersion == params.Base58Prefix(CChainParams::PUBKEY_ADDRESS) ||
-                         vchVersion == params.Base58Prefix(CChainParams::SCRIPT_ADDRESS);
-    return fCorrectSize && fKnownVersion;
-}
-
-CTxDestination CBitcoinAddress::Get() const
-{
-    if (!IsValid())
-        return CNoDestination();
-    uint160 id;
-    memcpy(&id, vchData.data(), 20);
-    if (vchVersion == Params().Base58Prefix(CChainParams::PUBKEY_ADDRESS))
-        return CKeyID(id);
-    else if (vchVersion == Params().Base58Prefix(CChainParams::SCRIPT_ADDRESS))
-        return CScriptID(id);
-    else
-        return CNoDestination();
-}
 
 void CBitcoinSecret::SetKey(const CKey& vchSecret)
 {
@@ -328,22 +298,20 @@ bool CBitcoinSecret::SetString(const std::string& strSecret)
 
 std::string EncodeDestination(const CTxDestination& dest)
 {
-    CBitcoinAddress addr(dest);
-    if (!addr.IsValid()) return "";
-    return addr.ToString();
+    return boost::apply_visitor(DestinationEncoder(Params()), dest);
 }
 
 CTxDestination DecodeDestination(const std::string& str)
 {
-    return CBitcoinAddress(str).Get();
+    return DecodeDestination(str, Params());
 }
 
 bool IsValidDestinationString(const std::string& str, const CChainParams& params)
 {
-    return CBitcoinAddress(str).IsValid(params);
+    return IsValidDestination(DecodeDestination(str, params));
 }
 
 bool IsValidDestinationString(const std::string& str)
 {
-    return CBitcoinAddress(str).IsValid();
+    return IsValidDestinationString(str, Params());
 }
--- a/src/base58.cpp
+++ b/src/base58.cpp
@@ -4,9 +4,11 @@
 
 #include "base58.h"
 
+#include "bech32.h"
 #include "hash.h"
 #include "script/script.h"
 #include "uint256.h"
+#include "utilstrencodings.h"
 
 #include <boost/variant/apply_visitor.hpp>
 #include <boost/variant/static_visitor.hpp>
@@ -235,7 +237,31 @@ public:
         return EncodeBase58Check(data);
     }
 
-    std::string operator()(const CNoDestination& no) const { return ""; }
+    std::string operator()(const WitnessV0KeyHash& id) const
+    {
+        std::vector<unsigned char> data = {0};
+        ConvertBits<8, 5, true>(data, id.begin(), id.end());
+        return bech32::Encode(m_params.Bech32HRP(), data);
+    }
+
+    std::string operator()(const WitnessV0ScriptHash& id) const
+    {
+        std::vector<unsigned char> data = {0};
+        ConvertBits<8, 5, true>(data, id.begin(), id.end());
+        return bech32::Encode(m_params.Bech32HRP(), data);
+    }
+
+    std::string operator()(const WitnessUnknown& id) const
+    {
+        if (id.version < 1 || id.version > 16 || id.length < 2 || id.length > 40) {
+            return {};
+        }
+        std::vector<unsigned char> data = {(unsigned char)id.version};
+        ConvertBits<8, 5, true>(data, id.program, id.program + id.length);
+        return bech32::Encode(m_params.Bech32HRP(), data);
+    }
+
+    std::string operator()(const CNoDestination& no) const { return {}; }
 };
 
 CTxDestination DecodeDestination(const std::string& str, const CChainParams& params)
@@ -257,6 +283,40 @@ CTxDestination DecodeDestination(const std::string& str, const CChainParams& par
         if (data.size() == hash.size() + script_prefix.size() && std::equal(script_prefix.begin(), script_prefix.end(), data.begin())) {
             std::copy(data.begin() + script_prefix.size(), data.end(), hash.begin());
             return CScriptID(hash);
+        }
+    }
+    data.clear();
+    auto bech = bech32::Decode(str);
+    if (bech.second.size() > 0 && bech.first == params.Bech32HRP()) {
+        // Bech32 decoding
+        int version = bech.second[0]; // The first 5 bit symbol is the witness version (0-16)
+        // The rest of the symbols are converted witness program bytes.
+        if (ConvertBits<5, 8, false>(data, bech.second.begin() + 1, bech.second.end())) {
+            if (version == 0) {
+                {
+                    WitnessV0KeyHash keyid;
+                    if (data.size() == keyid.size()) {
+                        std::copy(data.begin(), data.end(), keyid.begin());
+                        return keyid;
+                    }
+                }
+                {
+                    WitnessV0ScriptHash scriptid;
+                    if (data.size() == scriptid.size()) {
+                        std::copy(data.begin(), data.end(), scriptid.begin());
+                        return scriptid;
+                    }
+                }
+                return CNoDestination();
+            }
+            if (version > 16 || data.size() < 2 || data.size() > 40) {
+                return CNoDestination();
+            }
+            WitnessUnknown unk;
+            unk.version = version;
+            std::copy(data.begin(), data.end(), unk.program);
+            unk.length = data.size();
+            return unk;
         }
     }
     return CNoDestination();
--- a/src/bech32.cpp
+++ b/src/bech32.cpp
@@ -1,0 +1,191 @@
+// Copyright (c) 2017 Pieter Wuille
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include "bech32.h"
+
+namespace
+{
+
+typedef std::vector<uint8_t> data;
+
+/** The Bech32 character set for encoding. */
+const char* CHARSET = "qpzry9x8gf2tvdw0s3jn54khce6mua7l";
+
+/** The Bech32 character set for decoding. */
+const int8_t CHARSET_REV[128] = {
+    -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    15, -1, 10, 17, 21, 20, 26, 30,  7,  5, -1, -1, -1, -1, -1, -1,
+    -1, 29, -1, 24, 13, 25,  9,  8, 23, -1, 18, 22, 31, 27, 19, -1,
+     1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1,
+    -1, 29, -1, 24, 13, 25,  9,  8, 23, -1, 18, 22, 31, 27, 19, -1,
+     1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1
+};
+
+/** Concatenate two byte arrays. */
+data Cat(data x, const data& y)
+{
+    x.insert(x.end(), y.begin(), y.end());
+    return x;
+}
+
+/** This function will compute what 6 5-bit values to XOR into the last 6 input values, in order to
+ *  make the checksum 0. These 6 values are packed together in a single 30-bit integer. The higher
+ *  bits correspond to earlier values. */
+uint32_t PolyMod(const data& v)
+{
+    // The input is interpreted as a list of coefficients of a polynomial over F = GF(32), with an
+    // implicit 1 in front. If the input is [v0,v1,v2,v3,v4], that polynomial is v(x) =
+    // 1*x^5 + v0*x^4 + v1*x^3 + v2*x^2 + v3*x + v4. The implicit 1 guarantees that
+    // [v0,v1,v2,...] has a distinct checksum from [0,v0,v1,v2,...].
+
+    // The output is a 30-bit integer whose 5-bit groups are the coefficients of the remainder of
+    // v(x) mod g(x), where g(x) is the Bech32 generator,
+    // x^6 + {29}x^5 + {22}x^4 + {20}x^3 + {21}x^2 + {29}x + {18}. g(x) is chosen in such a way
+    // that the resulting code is a BCH code, guaranteeing detection of up to 3 errors within a
+    // window of 1023 characters. Among the various possible BCH codes, one was selected to in
+    // fact guarantee detection of up to 4 errors within a window of 89 characters.
+
+    // Note that the coefficients are elements of GF(32), here represented as decimal numbers
+    // between {}. In this finite field, addition is just XOR of the corresponding numbers. For
+    // example, {27} + {13} = {27 ^ 13} = {22}. Multiplication is more complicated, and requires
+    // treating the bits of values themselves as coefficients of a polynomial over a smaller field,
+    // GF(2), and multiplying those polynomials mod a^5 + a^3 + 1. For example, {5} * {26} =
+    // (a^2 + 1) * (a^4 + a^3 + a) = (a^4 + a^3 + a) * a^2 + (a^4 + a^3 + a) = a^6 + a^5 + a^4 + a
+    // = a^3 + 1 (mod a^5 + a^3 + 1) = {9}.
+
+    // During the course of the loop below, `c` contains the bitpacked coefficients of the
+    // polynomial constructed from just the values of v that were processed so far, mod g(x). In
+    // the above example, `c` initially corresponds to 1 mod (x), and after processing 2 inputs of
+    // v, it corresponds to x^2 + v0*x + v1 mod g(x). As 1 mod g(x) = 1, that is the starting value
+    // for `c`.
+    uint32_t c = 1;
+    for (auto v_i : v) {
+        // We want to update `c` to correspond to a polynomial with one extra term. If the initial
+        // value of `c` consists of the coefficients of c(x) = f(x) mod g(x), we modify it to
+        // correspond to c'(x) = (f(x) * x + v_i) mod g(x), where v_i is the next input to
+        // process. Simplifying:
+        // c'(x) = (f(x) * x + v_i) mod g(x)
+        //         ((f(x) mod g(x)) * x + v_i) mod g(x)
+        //         (c(x) * x + v_i) mod g(x)
+        // If c(x) = c0*x^5 + c1*x^4 + c2*x^3 + c3*x^2 + c4*x + c5, we want to compute
+        // c'(x) = (c0*x^5 + c1*x^4 + c2*x^3 + c3*x^2 + c4*x + c5) * x + v_i mod g(x)
+        //       = c0*x^6 + c1*x^5 + c2*x^4 + c3*x^3 + c4*x^2 + c5*x + v_i mod g(x)
+        //       = c0*(x^6 mod g(x)) + c1*x^5 + c2*x^4 + c3*x^3 + c4*x^2 + c5*x + v_i
+        // If we call (x^6 mod g(x)) = k(x), this can be written as
+        // c'(x) = (c1*x^5 + c2*x^4 + c3*x^3 + c4*x^2 + c5*x + v_i) + c0*k(x)
+
+        // First, determine the value of c0:
+        uint8_t c0 = c >> 25;
+
+        // Then compute c1*x^5 + c2*x^4 + c3*x^3 + c4*x^2 + c5*x + v_i:
+        c = ((c & 0x1ffffff) << 5) ^ v_i;
+
+        // Finally, for each set bit n in c0, conditionally add {2^n}k(x):
+        if (c0 & 1)  c ^= 0x3b6a57b2; //     k(x) = {29}x^5 + {22}x^4 + {20}x^3 + {21}x^2 + {29}x + {18}
+        if (c0 & 2)  c ^= 0x26508e6d; //  {2}k(x) = {19}x^5 +  {5}x^4 +     x^3 +  {3}x^2 + {19}x + {13}
+        if (c0 & 4)  c ^= 0x1ea119fa; //  {4}k(x) = {15}x^5 + {10}x^4 +  {2}x^3 +  {6}x^2 + {15}x + {26}
+        if (c0 & 8)  c ^= 0x3d4233dd; //  {8}k(x) = {30}x^5 + {20}x^4 +  {4}x^3 + {12}x^2 + {30}x + {29}
+        if (c0 & 16) c ^= 0x2a1462b3; // {16}k(x) = {21}x^5 +     x^4 +  {8}x^3 + {24}x^2 + {21}x + {19}
+    }
+    return c;
+}
+
+/** Convert to lower case. */
+inline unsigned char LowerCase(unsigned char c)
+{
+    return (c >= 'A' && c <= 'Z') ? (c - 'A') + 'a' : c;
+}
+
+/** Expand a HRP for use in checksum computation. */
+data ExpandHRP(const std::string& hrp)
+{
+    data ret;
+    ret.reserve(hrp.size() + 90);
+    ret.resize(hrp.size() * 2 + 1);
+    for (size_t i = 0; i < hrp.size(); ++i) {
+        unsigned char c = hrp[i];
+        ret[i] = c >> 5;
+        ret[i + hrp.size() + 1] = c & 0x1f;
+    }
+    ret[hrp.size()] = 0;
+    return ret;
+}
+
+/** Verify a checksum. */
+bool VerifyChecksum(const std::string& hrp, const data& values)
+{
+    // PolyMod computes what value to xor into the final values to make the checksum 0. However,
+    // if we required that the checksum was 0, it would be the case that appending a 0 to a valid
+    // list of values would result in a new valid list. For that reason, Bech32 requires the
+    // resulting checksum to be 1 instead.
+    return PolyMod(Cat(ExpandHRP(hrp), values)) == 1;
+}
+
+/** Create a checksum. */
+data CreateChecksum(const std::string& hrp, const data& values)
+{
+    data enc = Cat(ExpandHRP(hrp), values);
+    enc.resize(enc.size() + 6); // Append 6 zeroes
+    uint32_t mod = PolyMod(enc) ^ 1; // Determine what to XOR into those 6 zeroes.
+    data ret(6);
+    for (size_t i = 0; i < 6; ++i) {
+        // Convert the 5-bit groups in mod to checksum values.
+        ret[i] = (mod >> (5 * (5 - i))) & 31;
+    }
+    return ret;
+}
+
+} // namespace
+
+namespace bech32
+{
+
+/** Encode a Bech32 string. */
+std::string Encode(const std::string& hrp, const data& values) {
+    data checksum = CreateChecksum(hrp, values);
+    data combined = Cat(values, checksum);
+    std::string ret = hrp + '1';
+    ret.reserve(ret.size() + combined.size());
+    for (auto c : combined) {
+        ret += CHARSET[c];
+    }
+    return ret;
+}
+
+/** Decode a Bech32 string. */
+std::pair<std::string, data> Decode(const std::string& str) {
+    bool lower = false, upper = false;
+    for (size_t i = 0; i < str.size(); ++i) {
+        unsigned char c = str[i];
+        if (c < 33 || c > 126) return {};
+        if (c >= 'a' && c <= 'z') lower = true;
+        if (c >= 'A' && c <= 'Z') upper = true;
+    }
+    if (lower && upper) return {};
+    size_t pos = str.rfind('1');
+    if (str.size() > 90 || pos == str.npos || pos == 0 || pos + 7 > str.size()) {
+        return {};
+    }
+    data values(str.size() - 1 - pos);
+    for (size_t i = 0; i < str.size() - 1 - pos; ++i) {
+        unsigned char c = str[i + pos + 1];
+        int8_t rev = (c < 33 || c > 126) ? -1 : CHARSET_REV[c];
+        if (rev == -1) {
+            return {};
+        }
+        values[i] = rev;
+    }
+    std::string hrp;
+    for (size_t i = 0; i < pos; ++i) {
+        hrp += LowerCase(str[i]);
+    }
+    if (!VerifyChecksum(hrp, values)) {
+        return {};
+    }
+    return {hrp, data(values.begin(), values.end() - 6)};
+}
+
+} // namespace bech32
--- a/src/bech32.h
+++ b/src/bech32.h
@@ -1,0 +1,25 @@
+// Copyright (c) 2017 Pieter Wuille
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+// Bech32 is a string encoding format used in newer address types.
+// The output consists of a human-readable part (alphanumeric), a
+// separator character (1), and a base32 data section, the last
+// 6 characters of which are a checksum.
+//
+// For more information, see BIP 173.
+
+#include <stdint.h>
+#include <string>
+#include <vector>
+
+namespace bech32
+{
+
+/** Encode a Bech32 string. Returns the empty string in case of failure. */
+std::string Encode(const std::string& hrp, const std::vector<uint8_t>& values);
+
+/** Decode a Bech32 string. Returns (hrp, data). Empty hrp means failure. */
+std::pair<std::string, std::vector<uint8_t>> Decode(const std::string& str);
+
+} // namespace bech32
--- a/src/chainparams.cpp
+++ b/src/chainparams.cpp
@@ -137,6 +137,8 @@ public:
         base58Prefixes[EXT_PUBLIC_KEY] = {0x04, 0x88, 0xB2, 0x1E};
         base58Prefixes[EXT_SECRET_KEY] = {0x04, 0x88, 0xAD, 0xE4};
 
+        bech32_hrp = "bc";
+
         vFixedSeeds = std::vector<SeedSpec6>(pnSeed6_main, pnSeed6_main + ARRAYLEN(pnSeed6_main));
 
         fDefaultConsistencyChecks = false;
@@ -236,6 +238,8 @@ public:
         base58Prefixes[EXT_PUBLIC_KEY] = {0x04, 0x35, 0x87, 0xCF};
         base58Prefixes[EXT_SECRET_KEY] = {0x04, 0x35, 0x83, 0x94};
 
+        bech32_hrp = "tb";
+
         vFixedSeeds = std::vector<SeedSpec6>(pnSeed6_test, pnSeed6_test + ARRAYLEN(pnSeed6_test));
 
         fDefaultConsistencyChecks = false;
@@ -330,6 +334,8 @@ public:
         base58Prefixes[SECRET_KEY] =     std::vector<unsigned char>(1,239);
         base58Prefixes[EXT_PUBLIC_KEY] = {0x04, 0x35, 0x87, 0xCF};
         base58Prefixes[EXT_SECRET_KEY] = {0x04, 0x35, 0x83, 0x94};
+
+        bech32_hrp = "bcrt";
     }
 };
 
--- a/src/qt/bitcoinaddressvalidator.cpp
+++ b/src/qt/bitcoinaddressvalidator.cpp
@@ -67,7 +67,7 @@ QValidator::State BitcoinAddressEntryValidator::validate(QString &input, int &po
         if (((ch >= '0' && ch<='9') ||
             (ch >= 'a' && ch<='z') ||
             (ch >= 'A' && ch<='Z')) &&
-            ch != 'l' && ch != 'I' && ch != '0' && ch != 'O')
+            ch != 'I' && ch != 'O') // Characters invalid in both Base58 and Bech32
         {
             // Alphanumeric and not a 'forbidden' character
         }
--- a/src/script/ismine.cpp
+++ b/src/script/ismine.cpp
@@ -61,6 +61,7 @@ isminetype IsMine(const CKeyStore &keystore, const CScript& scriptPubKey, bool& 
     {
     case TX_NONSTANDARD:
     case TX_NULL_DATA:
+    case TX_WITNESS_UNKNOWN:
         break;
     case TX_PUBKEY:
         keyID = CPubKey(vSolutions[0]).GetID();
--- a/src/script/standard.cpp
+++ b/src/script/standard.cpp
@@ -30,6 +30,7 @@ const char* GetTxnOutputType(txnouttype t)
     case TX_NULL_DATA: return "nulldata";
     case TX_WITNESS_V0_KEYHASH: return "witness_v0_keyhash";
     case TX_WITNESS_V0_SCRIPTHASH: return "witness_v0_scripthash";
+    case TX_WITNESS_UNKNOWN: return "witness_unknown";
     }
     return nullptr;
 }
@@ -73,6 +74,12 @@ bool Solver(const CScript& scriptPubKey, txnouttype& typeRet, std::vector<std::v
         if (witnessversion == 0 && witnessprogram.size() == 32) {
             typeRet = TX_WITNESS_V0_SCRIPTHASH;
             vSolutionsRet.push_back(witnessprogram);
+            return true;
+        }
+        if (witnessversion != 0) {
+            typeRet = TX_WITNESS_UNKNOWN;
+            vSolutionsRet.push_back(std::vector<unsigned char>{(unsigned char)witnessversion});
+            vSolutionsRet.push_back(std::move(witnessprogram));
             return true;
         }
         return false;
@@ -198,6 +205,23 @@ bool ExtractDestination(const CScript& scriptPubKey, CTxDestination& addressRet)
     {
         addressRet = CScriptID(uint160(vSolutions[0]));
         return true;
+    } else if (whichType == TX_WITNESS_V0_KEYHASH) {
+        WitnessV0KeyHash hash;
+        std::copy(vSolutions[0].begin(), vSolutions[0].end(), hash.begin());
+        addressRet = hash;
+        return true;
+    } else if (whichType == TX_WITNESS_V0_SCRIPTHASH) {
+        WitnessV0ScriptHash hash;
+        std::copy(vSolutions[0].begin(), vSolutions[0].end(), hash.begin());
+        addressRet = hash;
+        return true;
+    } else if (whichType == TX_WITNESS_UNKNOWN) {
+        WitnessUnknown unk;
+        unk.version = vSolutions[0][0];
+        std::copy(vSolutions[1].begin(), vSolutions[1].end(), unk.program);
+        unk.length = vSolutions[1].size();
+        addressRet = unk;
+        return true;
     }
     // Multisig txns have more than one address...
     return false;
@@ -266,6 +290,27 @@ public:
     bool operator()(const CScriptID &scriptID) const {
         script->clear();
         *script << OP_HASH160 << ToByteVector(scriptID) << OP_EQUAL;
+        return true;
+    }
+
+    bool operator()(const WitnessV0KeyHash& id) const
+    {
+        script->clear();
+        *script << OP_0 << ToByteVector(id);
+        return true;
+    }
+
+    bool operator()(const WitnessV0ScriptHash& id) const
+    {
+        script->clear();
+        *script << OP_0 << ToByteVector(id);
+        return true;
+    }
+
+    bool operator()(const WitnessUnknown& id) const
+    {
+        script->clear();
+        *script << CScript::EncodeOP_N(id.version) << std::vector<unsigned char>(id.program, id.program + id.length);
         return true;
     }
 };
--- a/src/script/standard.h
+++ b/src/script/standard.h
@@ -64,6 +64,7 @@ enum txnouttype
     TX_NULL_DATA, //!< unspendable OP_RETURN script that carries data
     TX_WITNESS_V0_SCRIPTHASH,
     TX_WITNESS_V0_KEYHASH,
+    TX_WITNESS_UNKNOWN, //!< Only for Witness versions not already defined above
 };
 
 class CNoDestination {
@@ -72,14 +73,42 @@ public:
     friend bool operator<(const CNoDestination &a, const CNoDestination &b) { return true; }
 };
 
+struct WitnessV0ScriptHash : public uint256 {};
+struct WitnessV0KeyHash : public uint160 {};
+
+//! CTxDestination subtype to encode any future Witness version
+struct WitnessUnknown
+{
+    unsigned int version;
+    unsigned int length;
+    unsigned char program[40];
+
+    friend bool operator==(const WitnessUnknown& w1, const WitnessUnknown& w2) {
+        if (w1.version != w2.version) return false;
+        if (w1.length != w2.length) return false;
+        return std::equal(w1.program, w1.program + w1.length, w2.program);
+    }
+
+    friend bool operator<(const WitnessUnknown& w1, const WitnessUnknown& w2) {
+        if (w1.version < w2.version) return true;
+        if (w1.version > w2.version) return false;
+        if (w1.length < w2.length) return true;
+        if (w1.length > w2.length) return false;
+        return std::lexicographical_compare(w1.program, w1.program + w1.length, w2.program, w2.program + w2.length);
+    }
+};
+
 /**
  * A txout script template with a specific destination. It is either:
  *  * CNoDestination: no destination set
- *  * CKeyID: TX_PUBKEYHASH destination
- *  * CScriptID: TX_SCRIPTHASH destination
+ *  * CKeyID: TX_PUBKEYHASH destination (P2PKH)
+ *  * CScriptID: TX_SCRIPTHASH destination (P2SH)
+ *  * WitnessV0ScriptHash: TX_WITNESS_V0_SCRIPTHASH destination (P2WSH)
+ *  * WitnessV0KeyHash: TX_WITNESS_V0_KEYHASH destination (P2WPKH)
+ *  * WitnessUnknown: TX_WITNESS_UNKNOWN destination (P2W???)
  *  A CTxDestination is the internal data type encoded in a bitcoin address
  */
-typedef boost::variant<CNoDestination, CKeyID, CScriptID> CTxDestination;
+typedef boost::variant<CNoDestination, CKeyID, CScriptID, WitnessV0ScriptHash, WitnessV0KeyHash, WitnessUnknown> CTxDestination;
 
 /** Check whether a CTxDestination is a CNoDestination. */
 bool IsValidDestination(const CTxDestination& dest);
@@ -104,7 +133,7 @@ bool Solver(const CScript& scriptPubKey, txnouttype& typeRet, std::vector<std::v
  * Parse a standard scriptPubKey for the destination address. Assigns result to
  * the addressRet parameter and returns true if successful. For multisig
  * scripts, instead use ExtractDestinations. Currently only works for P2PK,
- * P2PKH, and P2SH scripts.
+ * P2PKH, P2SH, P2WPKH, and P2WSH scripts.
  */
 bool ExtractDestination(const CScript& scriptPubKey, CTxDestination& addressRet);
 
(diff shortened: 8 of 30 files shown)


luke-jr:
Would prefer to have simply sending-to (maybe validating/analyzing too?) as a separate PR, before wallet upgrades.

I'm not sure when it would make sense to convert between P2SH and BIP173...

sipa:
@luke-jr I agree, but I consider `addwitnessaddress` an RPC to aid with testing, not full support.

gmaxwell:
addwitnessaddress is very much not actual support, it's a test shim.

luke-jr:
But it modifies the wallet, no? Seems useful to review independently from the rest. Especially since it has the additional considerations of what happens if you try to use it and then downgrade to an older version...

sipa:
@luke-jr It does. But consider that we've since 0.13.1 had support to receiving and spending native witness outputs in the wallet (without that, testing the consensus logic for it would have been much harder), just no way to encode such outputs as strings. This merely extends that ability to test to include the string representation.

sipa:
Added support in Python framework, and some integrated some functional tests into the `segwit.py` test.

src/bech32.cpp
```diff
@@ -0,0 +1,157 @@
...
+    std::string ret = hrp + '1';
+    ret.reserve(ret.size() + combined.size());
+    for (size_t i = 0; i < combined.size(); ++i) {
+        ret += CHARSET[combined[i]];
+    }
+    return ret;
+}
+
+/** Decode a Bech32 string. */
+std::pair<std::string, data> Decode(const std::string& str) {
```
promag: IMO `bool Decode(const std::string& str, const std::string& hrp, data& d)` feels better, and this way below it can early return.

src/base58.cpp
```diff
@@ -235,13 +237,64 @@ class DestinationEncoder : public boost::static_visitor<std::string>
...
+                    return id;
+                }
+                if (data.size() == 32) {
+                    WitnessV0ScriptHash id;
+                    memcpy(id.begin(), data.data(), 32);
+                    return id;
+                }
+                return CNoDestination();
+            }
+            if (version > 16 || data.size() < 2 || data.size() > 40) return CNoDestination();
```
promag: Return in new line?

src/base58.cpp
```diff
@@ -3,15 +3,18 @@
...
 #include "hash.h"
 #include "uint256.h"
+#include "utilstrencodings.h"
 
 #include <assert.h>
 #include <stdint.h>
 #include <string.h>
 #include <vector>
 #include <string>
+#include <algorithm>
```
promag: Sort.

src/bech32.h
```diff
@@ -0,0 +1,35 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+#include <stdint.h>
+#include <vector>
```
promag: Sort.

src/base58.cpp
```diff
@@ -235,13 +237,64 @@ class DestinationEncoder : public boost::static_visitor<std::string>
...
+    }
+
     std::string operator()(const CNoDestination& no) const { return ""; }
 };
 
 CTxDestination DecodeDestination(const std::string& str, const CChainParams& params)
 {
     std::vector<unsigned char> data;
     uint160 hash;
+    auto bech = bech32::Decode(str);
```
promag: Must come first? If not which is the cheapest?

src/base58.cpp
```diff
@@ -212,79 +215,102 @@ int CBase58Data::CompareTo(const CBase58Data& b58) const
...
-    uint160 id;
-    memcpy(&id, vchData.data(), 20);
-    if (vchVersion == Params().Base58Prefix(CChainParams::PUBKEY_ADDRESS))
-        return CKeyID(id);
-    else if (vchVersion == Params().Base58Prefix(CChainParams::SCRIPT_ADDRESS))
-        return CScriptID(id);
-    else
+    std::vector<unsigned char> data;
+    uint160 hash;
+    auto bech = bech32::Decode(str);
```
promag: Must come first? If not which is the cheapest?

promag (commented):
Fast code review. Move 4bc71a6  to #11117??

src/bech32.cpp
sipa: It's not often used in the Bitcoin Core codebase, but using pairs for multiple returned values is very typical in C++ (see the return type of [std::map::insert](http://en.cppreference.com/w/cpp/container/map/insert) for example). I'd rather stick with that.

src/base58.cpp
sipa: Done.

src/base58.cpp
sipa: bech32 is far cheaper (no basis conversion, no SHA256).

src/base58.cpp
sipa: done

src/bech32.h
sipa: Done.

src/base58.cpp
sipa: I haven't benchmarked, but Bech32 should be far cheaper (no SHA256, no basis conversion)

src/chainparams.cpp
```diff
@@ -330,6 +334,8 @@ class CRegTestParams : public CChainParams {
         base58Prefixes[SECRET_KEY] =     std::vector<unsigned char>(1,239);
         base58Prefixes[EXT_PUBLIC_KEY] = {0x04, 0x35, 0x87, 0xCF};
         base58Prefixes[EXT_SECRET_KEY] = {0x04, 0x35, 0x83, 0x94};
+
+        bech32_hrp = "rt";
```
sipa: I arbitrarily chose a bech32 prefix for regtest. Feel free to bikeshed (it doesn't even need to be just 2 characters).

src/base58.cpp
promag: Right, in terms of functionality the order doesn't matter. But at the moment most addresses  (don't know numbers) are base58 so for now move `bech32::Decode()` after `DecodeBase58Check()`?

It would be cool to move this out of `base58.cpp`, follow up maybe?

src/bech32.cpp
```diff
@@ -0,0 +1,140 @@
...
+    bool lower = false, upper = false;
+    bool ok = true;
+    for (size_t i = 0; ok && i < str.size(); ++i) {
+        unsigned char c = str[i];
+        if (c < 33 || c > 126) ok = false;
```
promag: Early return?

src/bech32.cpp
```diff
@@ -0,0 +1,140 @@
...
+        if (c >= 'a' && c <= 'z') lower = true;
+        if (c >= 'A' && c <= 'Z') upper = true;
+    }
+    if (lower && upper) ok = false;
```
promag: Early return?

src/bech32.cpp
```diff
@@ -0,0 +1,140 @@
...
+    size_t pos = str.rfind('1');
+    if (ok && str.size() <= 90 && pos != str.npos && pos >= 1 && pos + 7 <= str.size()) {
+        data values;
+        values.resize(str.size() - 1 - pos);
+        for (size_t i = 0; i < str.size() - 1 - pos; ++i) {
+            unsigned char c = str[i + pos + 1];
+            if (CHARSET_REV[c] == -1) ok = false;
```
promag: Early return?

src/base58.cpp
sipa: > Right, in terms of functionality the order doesn't matter. But at the moment most addresses (don't know numbers) are base58 so for now move bech32::Decode() after DecodeBase58Check()?

I was using a fail-fast approach, making the thing that most quickly fails first. You're right that as long as there are hardly any bech32 addresses, putting Base58 would be overall faster. But none of this is performance critical anyway...

src/bech32.cpp
sipa: Done.

src/bech32.cpp
sipa: Done.

src/bech32.cpp
sipa: Done.

src/bech32.h
```diff
@@ -0,0 +1,18 @@
+// Copyright (c) 2017 Pieter Wuille
+// Distributed under the MIT software license, see the accompanying
+// file COPYING or http://www.opensource.org/licenses/mit-license.php.
+
+#include <stdint.h>
+#include <string>
+#include <vector>
+
+namespace bech32
```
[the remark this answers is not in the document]
sipa: Done.

src/test/base58_tests.cpp
```diff
@@ -218,14 +247,39 @@ BOOST_AUTO_TEST_CASE(base58_keys_valid_gen)
...
             }
             std::string address = EncodeDestination(dest);
-            BOOST_CHECK_MESSAGE(address == exp_base58string, "mismatch: " + strTest);
+            BOOST_CHECK_MESSAGE(address == exp_base58string, strprintf("mismatch: %s: \"%s\" != \"%s\"", strTest, address, exp_base58string));
+
+            CTxDestination dest2 = DecodeDestination(address);
+            BOOST_CHECK_MESSAGE(dest == dest2, strprintf("mismatch in encoding: %s", strTest));
+
+            for (char& c : address) {
+                if (c >= 'a' && c <= 'z') {
```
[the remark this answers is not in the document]
sipa: Unfortunately, toupper is locale-dependent, so it can't be used for consistent behaviour.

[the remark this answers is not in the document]
jimpo: `toupper(c, std::locale::classic)`?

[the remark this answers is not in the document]
sipa: Seems overkill.

instagibbs:
Include a test with v1+ address?

sipa:
@instagibbs Done

instagibbs:
super surprising to me that regtest has a different human readable part: `bcrt1`.

Legacy addresses shared between testnet and regtest. Perhaps note this somewhere in the code and/or in this PR description.

laanwj:
> super surprising to me that regtest has a different human readable part: bcrt1.

That was surprising to me too, but I found it a pleasant surprise. It's not like prefixes are a very rare resource, this decreases ambiguity, and fits in with regtest getting its own RPC port in #10825.

instagibbs:
Definitely not arguing against it!

src/wallet/rpcwallet.cpp
```diff
@@ -1241,13 +1254,31 @@ UniValue addwitnessaddress(const JSONRPCRequest& request)
...
+
     Witnessifier w(pwallet);
     bool ret = boost::apply_visitor(w, dest);
     if (!ret) {
         throw JSONRPCError(RPC_WALLET_ERROR, "Public key or redeemscript not known to wallet, or the key is uncompressed");
     }
 
-    pwallet->SetAddressBook(w.result, "", "receive");
+    CScript redeemscript = GetScriptForDestination(w.result);
+    if (!w.already_witness) pwallet->AddCScript(redeemscript);
```
instagibbs: move this inside of the else statement below

src/wallet/rpcwallet.cpp
```diff
@@ -1241,13 +1254,31 @@ UniValue addwitnessaddress(const JSONRPCRequest& request)
...
+    }
+
     Witnessifier w(pwallet);
     bool ret = boost::apply_visitor(w, dest);
     if (!ret) {
         throw JSONRPCError(RPC_WALLET_ERROR, "Public key or redeemscript not known to wallet, or the key is uncompressed");
     }
 
-    pwallet->SetAddressBook(w.result, "", "receive");
+    CScript redeemscript = GetScriptForDestination(w.result);
```
instagibbs: this is more aptly described as a witness program script, since `w.result`  will always be a witness program.

sipa: Fixed.

src/wallet/rpcwallet.cpp
sipa: Done.

src/base58.cpp
```diff
@@ -212,86 +213,51 @@ int CBase58Data::CompareTo(const CBase58Data& b58) const
...
-        return CScriptID(id);
-    else
-        return CNoDestination();
+    std::vector<unsigned char> data;
+    uint160 hash;
+    if (DecodeBase58Check(str, data)) {
+        // Base58 decoding
+        const std::vector<unsigned char>& pubkey_prefix = params.Base58Prefix(CChainParams::PUBKEY_ADDRESS);
+        if (data.size() == 20 + pubkey_prefix.size() && std::equal(pubkey_prefix.begin(), pubkey_prefix.end(), data.begin())) {
+            memcpy(hash.begin(), &data[1], 20);
```
theuni: Why is this not ```&data[pubkey_prefix.size()]```  (or ```data.data() +  pubkey_prefix.size()```) ?

theuni:
utACK other than those two nits, and excluding the bech32 implementation itself. It looks sane, but I'd like to audit more thoroughly before ACKing.

src/base58.cpp
sipa: Nice catch, fixed!

meshcollider (commented):
utACK https://github.com/bitcoin/bitcoin/pull/11167/commits/efceb6ec12321d698203b26be1a4cdeed49514b7 ... https://github.com/bitcoin/bitcoin/pull/11167/commits/8a42338bb968c213e3837e5b5c008355ce05cc23

I'm not really familiar with the technicalities of how Bech32 encoding and decoding work, but the code is consistent with what's in BIP 173. Didn't review the tests very thoroughly but they look ok

src/wallet/rpcwallet.cpp
```diff
@@ -1213,17 +1225,18 @@ UniValue addwitnessaddress(const JSONRPCRequest& request)
...
             "\nAdd a witness address for a script (with pubkey or redeemscript known).\n"
             "It returns the witness script.\n"
 
             "\nArguments:\n"
             "1. \"address\"       (string, required) An address known to the wallet\n"
+            "2. p2sh            (bool, optional) Embed inside P2SH (default: true)\n"
 
             "\nResult:\n"
-            "\"witnessaddress\",  (string) The value of the new address (P2SH of witness script).\n"
+            "\"witnessaddress\",  (string) The value of the new address (P2SH or BIP173).\n"
```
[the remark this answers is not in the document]
sipa: Bech32 is the generic encoding format (and should be contrasted with Base58Check), BIP173 specifies a specific way of encoding SegWit addresses using Bech32 (like P2SH/BIP13 specifies a specific way of encoding BIP16 addresses using Base58Check).

src/bech32.h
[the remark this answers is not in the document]
sipa: Done again - I somehow lost this change.

src/script/standard.cpp
```diff
@@ -198,6 +203,23 @@ bool ExtractDestination(const CScript& scriptPubKey, CTxDestination& addressRet)
     {
         addressRet = CScriptID(uint160(vSolutions[0]));
         return true;
+    } else if (whichType == TX_WITNESS_V0_KEYHASH) {
+        WitnessV0KeyHash hash;
+        memcpy(hash.begin(), &vSolutions[0][0], 20);
```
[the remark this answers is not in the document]
sipa: Fixed.

src/wallet/rpcwallet.cpp
```diff
@@ -1213,17 +1225,18 @@ UniValue addwitnessaddress(const JSONRPCRequest& request)
...
+    if (request.fHelp || request.params.size() < 1 || request.params.size() > 2)
     {
-        std::string msg = "addwitnessaddress \"address\"\n"
+        std::string msg = "addwitnessaddress \"address\" ( p2sh )\n"
             "\nAdd a witness address for a script (with pubkey or redeemscript known).\n"
             "It returns the witness script.\n"
 
             "\nArguments:\n"
             "1. \"address\"       (string, required) An address known to the wallet\n"
+            "2. p2sh            (bool, optional) Embed inside P2SH (default: true)\n"
```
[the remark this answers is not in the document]
sipa: Fixed.

src/bech32.cpp
```diff
@@ -0,0 +1,138 @@
...
+        if (CHARSET_REV[c] == -1) {
+            return std::make_pair(std::string(), data());
+        }
+        values[i] = CHARSET_REV[c];
+    }
+    std::string hrp;
+    for (size_t i = 0; i < pos; ++i) {
+        hrp += LowerCase(str[i]);
+    }
+    if (!VerifyChecksum(hrp, values)) return std::make_pair(std::string(), data());
```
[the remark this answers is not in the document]
sipa: Fixed.

src/bech32.cpp
```diff
@@ -0,0 +1,138 @@
...
+std::string Encode(const std::string& hrp, const data& values) {
+    data checksum = CreateChecksum(hrp, values);
+    data combined = Cat(values, checksum);
+    std::string ret = hrp + '1';
+    ret.reserve(ret.size() + combined.size());
+    for (size_t i = 0; i < combined.size(); ++i) {
+        ret += CHARSET[combined[i]];
+    }
+    return ret;
+}
```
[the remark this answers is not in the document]
sipa: Fixed.

sipa:
I think I've addressed all comments. I've also added a few extra comments in some places.

> quick nit reminder, TX_WITNESS_UNKNOWN isn't explicitly handled in CombineSignatures/SignStep

Fixed.

sipa:
Rebased on top of #11297 to hopefully address the Travis issue.

NicolasDorier:
should fix https://github.com/bitcoin/bitcoin/issues/10924 I made tests on https://github.com/NicolasDorier/bitcoin/commits/importaddresssegwit can you include them @sipa  ?

sipa:
@NicolasDorier Cherry-picked with some refactoring and merging into segwit.py.

NicolasDorier:
seems good for me.

Will try later this PR against NTumbleBit. I am using P2SH-P2WSH right now because core was not supporting otherwise. I should be able to move to pure P2WSH thanks to this PR.

ecdsa:
FYI, this BIP is supported by Electrum, and will be part of the next release

src/base58.cpp
```diff
@@ -212,86 +216,103 @@ int CBase58Data::CompareTo(const CBase58Data& b58) const
...
+    }
 
-bool CBitcoinAddress::IsValid() const
-{
-    return IsValid(Params());
-}
+    std::string operator()(const WitnessUnknown& id) const
+    {
+        if (id.version < 1 || id.version > 16 || id.length < 2 || id.length > 40) {
+            return "";
```
promag: Here and in other places where an empty string is returned prefer:
```cpp
return {};
```
See https://stackoverflow.com/a/26588207.

src/base58.cpp
```diff
@@ -212,86 +216,103 @@ int CBase58Data::CompareTo(const CBase58Data& b58) const
...
+            return CKeyID(hash);
+        }
+        const std::vector<unsigned char>& script_prefix = params.Base58Prefix(CChainParams::SCRIPT_ADDRESS);
+        if (data.size() == 20 + script_prefix.size() && std::equal(script_prefix.begin(), script_prefix.end(), data.begin())) {
+            memcpy(hash.begin(), &data[script_prefix.size()], 20);
+            return CScriptID(hash);
+        }
+    }
+    auto bech = bech32::Decode(str);
+    if (bech.first == params.Bech32HRP() && bech.second.size() > 0) {
```
promag: Swap conditions?

src/bech32.cpp
```diff
@@ -0,0 +1,140 @@
...
+    uint32_t mod = PolyMod(enc) ^ 1;
+    data ret;
+    ret.resize(6);
+    for (size_t i = 0; i < 6; ++i) {
+        ret[i] = (mod >> (5 * (5 - i))) & 31;
+    }
+    return ret;
+}
+
+}
```
promag:
```cpp
} // namespace
```

MeshCollider: Ohh heh I must have misclicked and put my comment on the wrong line, I thought I pointed this one out already but it seems I just commented on the end of a function

src/base58.cpp
sipa: Fixed.

src/base58.cpp
sipa: Done.

src/bech32.cpp
sipa: Done.

test/functional/segwit.py
```diff
@@ -570,6 +584,30 @@ def run_test(self):
...
+            else:
+                scriptPubKey = "a9142f8c469c2f0084c48e11f998ffbe7efa7549f26d87"
+                transaction = "01000000000100e1f5050000000017a9142f8c469c2f0084c48e11f998ffbe7efa7549f26d8700000000"
+
+            self.nodes[1].importaddress(scriptPubKey, "", False)
+            rawtxfund = self.nodes[1].fundrawtransaction(transaction)['hex']
+            rawtxfund = self.nodes[1].signrawtransaction(rawtxfund)["hex"]
+            txid = self.nodes[1].decoderawtransaction(rawtxfund)["txid"]
+
+            self.nodes[1].sendrawtransaction(rawtxfund)
```
promag:
```py
txid = self.nodes[1].sendrawtransaction(rawtxfund)
```
And remove above?

instagibbs:
tACK https://github.com/bitcoin/bitcoin/pull/11167/commits/0571b9271b27ce2767e9c7034c23facb495fab65

Did a couple commits on top to get hdkey info returned with native segwit keyhashes. Something like this can be done as a quick followup, perhaps a modified version of #11089: https://github.com/instagibbs/bitcoin/commits/fullseg

src/bech32.cpp
```diff
@@ -0,0 +1,140 @@
+// Copyright (c) 2017 Pieter Wuille
```
achow101: Is this the right copyright header (here and in `bech32.h`)?

src/bech32.h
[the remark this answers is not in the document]
achow101: I don't see the comment... I think you lost it again.

src/script/ismine.cpp
```diff
@@ -142,6 +142,9 @@ isminetype IsMine(const CKeyStore &keystore, const CScript& scriptPubKey, bool&
             return ISMINE_SPENDABLE;
         break;
     }
+
+    case TX_WITNESS_UNKNOWN:
```
achow101: I think this should be above with the `TX_NULL_DATA` and `TX_NONSTANDARD` cases.

achow101 (commented):
Concept ACK. I just did a brief pass through of things that jumped out to me. Will do more review and testing later.

sdaftuar (commented):
I've only code-reviewed the Bech32 implementation (bcd3f74b649da1482c242371ddebeaf78d8b7813 ) so far, which looks good.  Left a few comment suggestions that would have been helpful for my review; feel free to take or leave, but I figured I would post them at least to help others who may be reviewing as well.

(Will continue to review the rest of the PR.)

src/bech32.cpp
sipa: I think so.

src/script/ismine.cpp
sipa: Nice catch, done.

src/bech32.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+/** Concatenate two byte arrays. */
+data Cat(data x, const data& y)
+{
+    x.insert(x.end(), y.begin(), y.end());
+    return x;
+}
+
+/** Find the polynomial with value coefficients mod the generator as 30-bit. */
+uint32_t PolyMod(const data& values)
```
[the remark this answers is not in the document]
sipa: I've added a long explanation.

src/bech32.cpp
```diff
@@ -0,0 +1,140 @@
...
+{
+    uint32_t chk = 1;
+    for (size_t i = 0; i < values.size(); ++i) {
+        uint8_t top = chk >> 25;
+        chk = (chk & 0x1ffffff) << 5 ^ values[i] ^
```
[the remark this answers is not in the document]
sipa: Done.

src/bech32.cpp
```diff
@@ -0,0 +1,140 @@
...
+            (-((top >> 0) & 1) & 0x3b6a57b2UL) ^
+            (-((top >> 1) & 1) & 0x26508e6dUL) ^
+            (-((top >> 2) & 1) & 0x1ea119faUL) ^
+            (-((top >> 3) & 1) & 0x3d4233ddUL) ^
+            (-((top >> 4) & 1) & 0x2a1462b3UL);
```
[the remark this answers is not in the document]
sipa: I've replaced it with much more readable conditions. I've verified that at least GCC 6.3 compiles it to efficient code (without any jumps).

src/bech32.cpp
```diff
@@ -0,0 +1,140 @@
...
+{
+    return PolyMod(Cat(ExpandHRP(hrp), values)) == 1;
+}
+
+/** Create a checksum. */
+data CreateChecksum(const std::string& hrp, const data& values)
+{
+    data enc = Cat(ExpandHRP(hrp), values);
+    enc.resize(enc.size() + 6);
+    uint32_t mod = PolyMod(enc) ^ 1;
```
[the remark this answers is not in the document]
sipa: Done.

src/bech32.h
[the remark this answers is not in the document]
sipa: Done.

src/bech32.cpp
achow101: Ok, just checking since most files in this repo have `Copyright (c) The Bitcoin Core developers`

src/test/base58_tests.cpp
```diff
@@ -218,14 +247,39 @@ BOOST_AUTO_TEST_CASE(base58_keys_valid_gen)
...
+                dest = id;
+            } else if (exp_addrType == "p2wsh") {
+                WitnessV0ScriptHash id;
+                memcpy(id.begin(), exp_payload.data(), 32);
+                dest = id;
+            } else if (exp_addrType.size() == 5 && exp_addrType.substr(0, 4) == "p2w?") {
+                WitnessUnknown unk;
+                memcpy(unk.program, exp_payload.data(), exp_payload.size());
+                unk.length = exp_payload.size();
+                unk.version = exp_addrType[4] - 'a';
```
gmaxwell: gross.  Can these tests be changed to work with whole script pubkeys rather than tweaking the addrType?

sipa: > gross

I agree. 

> Can these tests be changed to work with whole script pubkeys rather than tweaking the addrType?

Done. I've added an extra commit to first refactor the tests into testing scriptPubKeys.

src/test/data/base58_keys_valid.json
```diff
@@ -422,5 +422,59 @@
...
+    ],
+    [
+        "tb1qqqqqp399et2xygdj5xreqhjjvcmzhxw4aywxecjdzew6hylgvsesrxh6hy",
+        "0020000000c4a5cad46221b2a187905e5266362b99d5e91c6ce24d165dab93e86433",
+        {
+            "isPrivkey": false,
+            "isTestnet": true,
+            "tryCaseFlip": true
+        }
     ]
```
gmaxwell: no bcrt tests?

gmaxwell (approved):
utACK.

src/test/data/base58_keys_valid.json
sipa: Fixed.

sipa:
Addressed @gmaxwell's comment about `bcrt` tests, by adding an extra commit that introduces regtest examples to base58_tests, and then extending those in the BIP173 tests commit.

jonasschnelli:
Maybe cherry-pick  https://github.com/jonasschnelli/bitcoin/commit/f86c0585dccd29dca63cef3e6a0d434f49abac49 to enable sending to BIP173 addresses through the GUI

danra:
@sipa (More of a BIP 173 question) - what's the reasoning for xoring 1 to the checksum both when creating and validating it?

More specifically: `polymod` in `bech32_create_checksum()` is xored with 1 and `bech32_verify_checksum()` checks for `bench32_polymod(...) == 1`. Instead, one could avoid the xor, and check  `bench32_polymod(...) == 0`.

src/base58.cpp
```diff
@@ -212,86 +213,51 @@ int CBase58Data::CompareTo(const CBase58Data& b58) const
...
-        return CScriptID(id);
-    else
-        return CNoDestination();
+    std::vector<unsigned char> data;
+    uint160 hash;
+    if (DecodeBase58Check(str, data)) {
+        // Base58Check decoding
+        const std::vector<unsigned char>& pubkey_prefix = params.Base58Prefix(CChainParams::PUBKEY_ADDRESS);
+        if (data.size() == 20 + pubkey_prefix.size() && std::equal(pubkey_prefix.begin(), pubkey_prefix.end(), data.begin())) {
+            memcpy(hash.begin(), &data[pubkey_prefix.size()], 20);
```
danra: - Use `std::copy[_n]` instead of `memcpy`
- Suggest DRYing constant lengths 20, 32. Either defining them here or using existing definition from another header.

danra (changes requested):

src/base58.cpp
```diff
@@ -5,16 +5,17 @@
...
+#include "script/script.h"
 #include "uint256.h"
 
 #include <assert.h>
-#include <stdint.h>
 #include <string.h>
-#include <vector>
-#include <string>
+
+#include <algorithm>
```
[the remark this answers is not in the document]
sipa: We don't have strict rules about header include order, and I'm not convinced they would be a good trade-off (more work for reviewers, only marginally more readable code).

src/base58.cpp
```diff
@@ -212,86 +213,51 @@ int CBase58Data::CompareTo(const CBase58Data& b58) const
 
 namespace
 {
-/** base58-encoded Bitcoin addresses.
- * Public-key-hash-addresses have version 0 (or 111 testnet).
- * The data vector contains RIPEMD160(SHA256(pubkey)), where pubkey is the serialized public key.
- * Script-hash-addresses have version 5 (or 196 testnet).
- * The data vector contains RIPEMD160(SHA256(cscript)), where cscript is the serialized redemption script.
- */
```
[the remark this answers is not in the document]
sipa: I've merged it into the implementation.

src/base58.cpp
sipa: Both done, in multiple places.

src/base58.cpp
```diff
@@ -328,22 +294,20 @@ bool CBitcoinSecret::SetString(const std::string& strSecret)
...
 {
-    return CBitcoinAddress(str).IsValid(params);
+    return IsValidDestination(DecodeDestination(str, params));
 }
 
 bool IsValidDestinationString(const std::string& str)
 {
-    return CBitcoinAddress(str).IsValid();
+    return IsValidDestination(DecodeDestination(str, Params()));
 }
```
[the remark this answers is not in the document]
sipa: Done.

sipa:
@jonasschnelli Cool, cherry-picked your commit.

@danra The xor 1 to verify the checksum is actually explained in this PR:

```c++
    // PolyMod computes what value to xor into the final values to make the checksum 0. However,
    // if we required that the checksum was 0, it would be the case that appending a 0 to a valid
    // list of values would result in a new valid list. For that reason, Bech32 requires the
    // resulting checksum to be 1 instead.
```

I'll update the BIP to explain it there as well.

src/base58.cpp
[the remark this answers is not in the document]
danra: IMHO Having the headers not arranged is just a more overhead to the reader. When it's a few headers, it's just a bit of overhead. When it's plenty, the overhead grows.

In any case, this specific comment is just about grouping standard header together, sorting them alphabetically, a guideline which I got the impression is mostly observed in the code.

src/base58.cpp
```diff
@@ -5,16 +5,17 @@
...
 
 #include <assert.h>
-#include <stdint.h>
 #include <string.h>
-#include <vector>
-#include <string>
+
+#include <algorithm>
 #include <boost/variant/apply_visitor.hpp>
 #include <boost/variant/static_visitor.hpp>
```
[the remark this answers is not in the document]
danra: The reasoning for this is the same as why user includes should come before system includes - avoid hidden dependencies.

While boost is canonical and well-tested, it's probably still better practice to have non-standard system headers be included before standard system headers, if only for consistency.

src/base58.cpp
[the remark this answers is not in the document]
danra: I thought I searched for the contents and didn't find it, I guess I missed it. Thanks!

src/base58.cpp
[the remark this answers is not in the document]
sipa: They're already sorted alphabetically and grouped together (but with C headers separate for C++ headers), which seems perfectly reasonable to me.

src/base58.cpp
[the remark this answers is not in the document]
sipa: Interesting, I hadn't heard about this reason. I'll take it into account in the future.

src/base58.cpp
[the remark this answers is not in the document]
sipa: I only did it after you commented.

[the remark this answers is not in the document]
danra: Ah, I misunderstood. Thanks!

sipa:
Rebased on top of #11116, and fixed a bug that it discovered (`Solver` would treat non-length 20/32 v0 witness outputs as WitnessUnknown rather than failing). Thanks, @jimpo!

gmaxwell (approved):
ACK

src/qt/bitcoinaddressvalidator.cpp
```diff
@@ -67,7 +67,7 @@ QValidator::State BitcoinAddressEntryValidator::validate(QString &input, int &po
         if (((ch >= '0' && ch<='9') ||
             (ch >= 'a' && ch<='z') ||
             (ch >= 'A' && ch<='Z')) &&
-            ch != 'l' && ch != 'I' && ch != '0' && ch != 'O')
+            ch != 'I' && ch != 'O')
```
instagibbs: nit: I think this deserves an in-line comment.

src/base58.cpp
```diff
@@ -259,6 +285,34 @@ CTxDestination DecodeDestination(const std::string& str, const CChainParams& par
             return CScriptID(hash);
         }
     }
+    auto bech = bech32::Decode(str);
+    if (bech.second.size() > 0 && bech.first == params.Bech32HRP()) {
+        // Bech32 decoding
+        int version = bech.second[0];
+        if (ConvertBits<5, 8, false>(data, bech.second.begin() + 1, bech.second.end())) {
```
instagibbs: Not obvious to me here why it's `bech.second.begin() + 1`.

src/script/standard.cpp
```diff
@@ -175,6 +182,7 @@ bool Solver(const CScript& scriptPubKey, txnouttype& typeRet, std::vector<std::v
 
 bool ExtractDestination(const CScript& scriptPubKey, CTxDestination& addressRet)
 {
+    valtype witness_program;
```
instagibbs: unused?

instagibbs (commented):
re-ACK, with some small nits

src/qt/bitcoinaddressvalidator.cpp
sipa: Fixed.

src/base58.cpp
sipa: Added a comment.

src/script/standard.cpp
sipa: Removed.

src/base58.cpp
```diff
@@ -212,86 +215,108 @@ int CBase58Data::CompareTo(const CBase58Data& b58) const
...
 
-    bool operator()(const CKeyID& id) const { return addr->Set(id); }
-    bool operator()(const CScriptID& id) const { return addr->Set(id); }
-    bool operator()(const CNoDestination& no) const { return false; }
-};
-
-} // namespace
+    std::string operator()(const CKeyID& id) const
+    {
+        std::vector<unsigned char> data = m_params.Base58Prefix(CChainParams::PUBKEY_ADDRESS);
```
promag: Call `data.reserve(prefix.size() + id.size())` before filling data to avoid reallocation?

Otherwise use copy constructor? Same below.

src/bech32.cpp
```diff
@@ -0,0 +1,191 @@
...
+    // if we required that the checksum was 0, it would be the case that appending a 0 to a valid
+    // list of values would result in a new valid list. For that reason, Bech32 requires the
+    // resulting checksum to be 1 instead.
+    return PolyMod(Cat(ExpandHRP(hrp), values)) == 1;
+}
+
+/** Create a checksum. */
+data CreateChecksum(const std::string& hrp, const data& values)
+{
+    data enc = Cat(ExpandHRP(hrp), values);
```
promag: Avoid copy?

src/bech32.cpp
```diff
@@ -0,0 +1,191 @@
...
+    enc.resize(enc.size() + 6); // Append 6 zeroes
```
promag: I believe this is the 3rd possible reallocation on the same vector:
 - 1st L106: `ret.resize()`
 - 2nd L30: `x.insert(...)`
 - 3rd L130: `enc.resize(enc.size() + 6)`

Suggestion, add 2nd argument `capacity = 0` to `ExpandHRP`?

src/bech32.cpp
sipa: No copy occurs here. The temporary returned by `ExpandHRP` directly becomes the first argument to `Cat` (note that it takes a by-value argument, and not a reference, for this reason). `Cat` updates it to append `values`, and then either uses copy elision or moving to assign it to `enc`.

src/base58.cpp
sipa: The copy constructor is used here when initializing data.

src/test/base58_tests.cpp
```diff
@@ -127,22 +84,25 @@ BOOST_AUTO_TEST_CASE(base58_keys_valid_parse)
...
         std::vector<unsigned char> exp_payload = ParseHex(test[1].get_str());
         const UniValue &metadata = test[2].get_obj();
         bool isPrivkey = find_value(metadata, "isPrivkey").get_bool();
-        bool isTestnet = find_value(metadata, "isTestnet").get_bool();
-        if (isTestnet)
+        bool isTestnet = find_value(metadata, "chain").get_str() == "testnet";
+        bool regtest = find_value(metadata, "chain").get_str() == "regtest";
+        bool try_case_flip = find_value(metadata, "tryCaseFlip").isNull() ? false : find_value(metadata, "tryCaseFlip").get_bool();
+        if (regtest) {
+            SelectParams(CBaseChainParams::REGTEST);
```
laanwj: I wonder why we don't use the SelectParams by string here.

src/test/base58_tests.cpp
```diff
@@ -188,50 +160,37 @@ BOOST_AUTO_TEST_CASE(base58_keys_valid_gen)
         std::vector<unsigned char> exp_payload = ParseHex(test[1].get_str());
         const UniValue &metadata = test[2].get_obj();
         bool isPrivkey = find_value(metadata, "isPrivkey").get_bool();
-        bool isTestnet = find_value(metadata, "isTestnet").get_bool();
-        if (isTestnet)
+        bool isTestnet = find_value(metadata, "chain").get_str() == "testnet";
+        bool regtest = find_value(metadata, "chain").get_str() == "regtest";
```
laanwj: Same here, `SelectParams(find_value(metadata, "chain").get_str())` would do?

src/bech32.cpp
```diff
@@ -0,0 +1,191 @@
...
+namespace
+{
+
+typedef std::vector<uint8_t> data;
+
+/** The Bech32 character set for encoding. */
+const char* CHARSET = "qpzry9x8gf2tvdw0s3jn54khce6mua7l";
+
+/** The Bech32 character set for decoding. */
+const int8_t CHARSET_REV[128] = {
```
laanwj: As this is indexed using a `uint8_t` I'd feel slightly more at ease if we extend this array to size 256, or add an explicit check against 128 in the loop where it is referenced. I know the check at the beginning of `Decode()` makes it impossible to hit this, but it seems something that is easily forgotten with a future code change.

laanwj:
utACK +/- nits

src/base58.cpp
danra: @sipa Should be rebased onto 1e46ebdf8618e585568ffc1b093c79cc9be07b57

src/bech32.cpp
```diff
@@ -0,0 +1,191 @@
...
+    // (a^2 + 1) * (a^4 + a^3 + a) = (a^4 + a^3 + a) * a^2 + (a^4 + a^3 + a) = a^6 + a^5 + a^4 + a
+    // = a^3 + 1 (mod a^5 + a^3 + 1) = {9}.
+
+    // During the course of the loop below, `c` contains the bitpacked coefficients of the
+    // polynomial constructed from just the values of v that were processed so far, mod g(x). In
+    // the above example, `c` initially corresponds to 1 mod (x), and after processing 2 inputs of
+    // v, it corresponds to x^2 + v0*x + v1 mod g(x). As 1 mod g(x) = 1, that is the starting value
+    // for `c`.
+    uint32_t c = 1;
+    for (size_t i = 0; i < v.size(); ++i) {
```
danra: Use a range-for loop, `i` isn't used except for accessing element (and in comments)

src/bech32.cpp
```diff
@@ -0,0 +1,191 @@
...
+        if (c0 & 8)  c ^= 0x3d4233dd; //  {8}k(x) = {30}x^5 + {20}x^4 +  {4}x^3 + {12}x^2 + {30}x + {29}
+        if (c0 & 16) c ^= 0x2a1462b3; // {16}k(x) = {21}x^5 +     x^4 +  {8}x^3 + {24}x^2 + {21}x + {19}
+    }
+    return c;
+}
+
+/** Convert to lower case. */
+inline unsigned char LowerCase(unsigned char c)
+{
+    return (c >= 'A' && c <= 'Z') ? (c - 'A') + 'a' : c;
```
danra: Here and later, `'A' <= c && c <= 'Z'` is better style (`c` is 'between' the values. Easier to read, harder to get wrong when used consistently)

src/base58.cpp
```diff
@@ -212,86 +215,108 @@ int CBase58Data::CompareTo(const CBase58Data& b58) const
...
+            return CKeyID(hash);
+        }
+        // Script-hash-addresses have version 5 (or 196 testnet).
+        // The data vector contains RIPEMD160(SHA256(cscript)), where cscript is the serialized redemption script.
+        const std::vector<unsigned char>& script_prefix = params.Base58Prefix(CChainParams::SCRIPT_ADDRESS);
+        if (data.size() == hash.size() + script_prefix.size() && std::equal(script_prefix.begin(), script_prefix.end(), data.begin())) {
+            std::copy(data.begin() + script_prefix.size(), data.end(), hash.begin());
+            return CScriptID(hash);
+        }
+    }
```
theuni: data.clear()

src/bech32.cpp
```diff
@@ -0,0 +1,191 @@
...
+    return ret;
+}
+
+/** Verify a checksum. */
+bool VerifyChecksum(const std::string& hrp, const data& values)
+{
+    // PolyMod computes what value to xor into the final values to make the checksum 0. However,
+    // if we required that the checksum was 0, it would be the case that appending a 0 to a valid
+    // list of values would result in a new valid list. For that reason, Bech32 requires the
+    // resulting checksum to be 1 instead.
```
danra: I might be wrong, but isn't the problem with prepending a 0, rather than appending one?

src/bech32.cpp
```diff
@@ -0,0 +1,191 @@
...
+}
+
+/** Create a checksum. */
+data CreateChecksum(const std::string& hrp, const data& values)
+{
+    data enc = Cat(ExpandHRP(hrp), values);
+    enc.resize(enc.size() + 6); // Append 6 zeroes
+    uint32_t mod = PolyMod(enc) ^ 1; // Determine what to XOR into those 6 zeroes.
+    data ret;
+    ret.resize(6);
```
danra: Just `data ret(6);`

src/bech32.cpp
```diff
@@ -0,0 +1,191 @@
...
+namespace bech32
+{
+
+/** Encode a Bech32 string. */
+std::string Encode(const std::string& hrp, const data& values) {
+    data checksum = CreateChecksum(hrp, values);
+    data combined = Cat(values, checksum);
+    std::string ret = hrp + '1';
+    ret.reserve(ret.size() + combined.size());
+    for (size_t i = 0; i < combined.size(); ++i) {
```
danra: Use range-for loop

src/bech32.cpp
```diff
@@ -0,0 +1,191 @@
...
+        if (c >= 'a' && c <= 'z') lower = true;
+        if (c >= 'A' && c <= 'Z') upper = true;
+    }
+    if (lower && upper) return std::make_pair(std::string(), data());
+    size_t pos = str.rfind('1');
+    if (str.size() > 90 || pos == str.npos || pos == 0 || pos + 7 > str.size()) {
+        return std::make_pair(std::string(), data());
+    }
+    data values;
+    values.resize(str.size() - 1 - pos);
```
danra: Just `data values(str.size() - 1 - pos);`

src/bech32.cpp
```diff
@@ -0,0 +1,191 @@
...
+    }
+    return ret;
+}
+
+/** Decode a Bech32 string. */
+std::pair<std::string, data> Decode(const std::string& str) {
+    bool lower = false, upper = false;
+    for (size_t i = 0; i < str.size(); ++i) {
+        unsigned char c = str[i];
+        if (c < 33 || c > 126) return std::make_pair(std::string(), data());
```
danra: Here and later, just `return {}`

src/bech32.cpp
```diff
@@ -0,0 +1,191 @@
...
+        return std::make_pair(std::string(), data());
+    }
+    data values;
+    values.resize(str.size() - 1 - pos);
+    for (size_t i = 0; i < str.size() - 1 - pos; ++i) {
+        unsigned char c = str[i + pos + 1];
+        if (CHARSET_REV[c] == -1) {
+            return std::make_pair(std::string(), data());
+        }
+        values[i] = CHARSET_REV[c];
```
danra: Access `CHARSET_REV` just once, store `const int8_t& val = CHARSET_REV[c]`

src/bech32.cpp
```diff
@@ -0,0 +1,191 @@
...
+    }
+    std::string hrp;
+    for (size_t i = 0; i < pos; ++i) {
+        hrp += LowerCase(str[i]);
+    }
+    if (!VerifyChecksum(hrp, values)) {
+        return std::make_pair(std::string(), data());
+    }
+    return std::make_pair(hrp, data(values.begin(), values.end() - 6));
```
danra: Here too `std::make_pair` can be removed in favor of `{}`

danra (changes requested):

src/test/bech32_tests.cpp
```diff
@@ -0,0 +1,61 @@
...
+    if (s1.size() != s2.size()) return false;
+    for (size_t i = 0; i < s1.size(); ++i) {
+        char c1 = s1[i];
+        if (c1 >= 'A' && c1 <= 'Z') c1 -= ('A' - 'a');
+        char c2 = s2[i];
+        if (c2 >= 'A' && c2 <= 'Z') c2 -= ('A' - 'a');
+        if (c1 != c2) return false;
+    }
+    return true;
+}
```
danra: Suggest using `std::tolower()` and standard algorithms to simplify the implementation

src/test/base58_tests.cpp
```diff
@@ -127,22 +84,25 @@ BOOST_AUTO_TEST_CASE(base58_keys_valid_parse)
...
             continue;
         }
         std::string exp_base58string = test[0].get_str();
         std::vector<unsigned char> exp_payload = ParseHex(test[1].get_str());
         const UniValue &metadata = test[2].get_obj();
         bool isPrivkey = find_value(metadata, "isPrivkey").get_bool();
-        bool isTestnet = find_value(metadata, "isTestnet").get_bool();
-        if (isTestnet)
+        bool isTestnet = find_value(metadata, "chain").get_str() == "testnet";
+        bool regtest = find_value(metadata, "chain").get_str() == "regtest";
```
danra: `isRegtest` for consistency

src/base58.cpp
```diff
@@ -212,86 +215,108 @@ int CBase58Data::CompareTo(const CBase58Data& b58) const
...
-bool CBitcoinAddress::Set(const CTxDestination& dest)
-{
-    return boost::apply_visitor(CBitcoinAddressVisitor(this), dest);
-}
+    std::string operator()(const WitnessV0ScriptHash& id) const
+    {
+        std::vector<unsigned char> data = {0};
+        ConvertBits<8, 5, true>(data, id.begin(), id.end());
+        return bech32::Encode(m_params.Bech32HRP(), data);
+    }
```
danra: `WitnessV0KeyHash` and `WitnessV0ScriptHash` overloads have identical implementation, refactor it to a private template (templated on `id` type) function and call it from these two?

src/base58.cpp
```diff
@@ -212,86 +215,108 @@ int CBase58Data::CompareTo(const CBase58Data& b58) const
...
+                WitnessV0KeyHash keyid;
+                if (data.size() == keyid.size()) {
+                    std::copy(data.begin(), data.end(), keyid.begin());
+                    return keyid;
+                }
+                WitnessV0ScriptHash scriptid;
+                if (data.size() == scriptid.size()) {
+                    std::copy(data.begin(), data.end(), scriptid.begin());
+                    return scriptid;
+                }
```
danra: Braces around each of `keyid` and `scriptid` segments would reduce their visibility to just the scope where they're needed.

src/base58.cpp
```diff
@@ -212,86 +215,108 @@ int CBase58Data::CompareTo(const CBase58Data& b58) const
...
+                return CNoDestination();
+            }
+            if (version > 16 || data.size() < 2 || data.size() > 40) {
+                return CNoDestination();
+            }
+            WitnessUnknown unk;
+            unk.version = version;
+            memcpy(unk.program, data.data(), data.size());
```
danra: `std::copy`

src/base58.cpp
```diff
@@ -212,86 +215,108 @@ int CBase58Data::CompareTo(const CBase58Data& b58) const
...
+            std::copy(data.begin() + script_prefix.size(), data.end(), hash.begin());
+            return CScriptID(hash);
+        }
+    }
+    auto bech = bech32::Decode(str);
+    if (bech.second.size() > 0 && bech.first == params.Bech32HRP()) {
+        // Bech32 decoding
+        int version = bech.second[0]; // The first 5 bit symbol is the witness version (0-16)
+        // The rest of the symbols are converted witness program bytes.
+        if (ConvertBits<5, 8, false>(data, bech.second.begin() + 1, bech.second.end())) {
```
danra: Early return if the condition fails would reduce indentation

danra: Looks like there's no enforcing of the following rule from the BIP regarding decoding:

> Re-arrange those bits into groups of 8 bits. **Any incomplete group at the end MUST be 4 bits or less, MUST be all zeroes**, and is discarded.

src/script/standard.cpp
```diff
@@ -198,6 +205,23 @@ bool ExtractDestination(const CScript& scriptPubKey, CTxDestination& addressRet)
     {
         addressRet = CScriptID(uint160(vSolutions[0]));
         return true;
+    } else if (whichType == TX_WITNESS_V0_KEYHASH) {
+        WitnessV0KeyHash hash;
+        memcpy(hash.begin(), vSolutions[0].data(), 20);
```
danra: `std::copy`

src/script/standard.cpp
```diff
@@ -198,6 +205,23 @@ bool ExtractDestination(const CScript& scriptPubKey, CTxDestination& addressRet)
...
+        memcpy(hash.begin(), vSolutions[0].data(), 32);
+        addressRet = hash;
+        return true;
+    } else if (whichType == TX_WITNESS_UNKNOWN) {
+        WitnessUnknown unk;
+        unk.version = vSolutions[0][0];
+        memcpy(unk.program, vSolutions[1].data(), vSolutions[1].size());
+        unk.length = vSolutions[1].size();
+        addressRet = unk;
+        return true;
```
danra: The header comment for `ExtractDestination` is out of date now that P2WPKH and P2WSH are supported, it says

> Currently only works for P2PK, P2PKH, and P2SH scripts.

src/script/standard.h
```diff
@@ -72,14 +73,44 @@ class CNoDestination {
...
+struct WitnessUnknown
+{
+    unsigned int version;
+    unsigned int length;
+    unsigned char program[40];
+
+    friend bool operator==(const WitnessUnknown& w1, const WitnessUnknown& w2) {
+        if (w1.version != w2.version) return false;
+        if (w1.length != w2.length) return false;
+        if (memcmp(w1.program, w2.program, w1.length)) return false;
```
danra: `std::equal`

src/script/standard.h
```diff
@@ -72,14 +73,44 @@ class CNoDestination {
...
+        return true;
+    }
+
+    friend bool operator<(const WitnessUnknown& w1, const WitnessUnknown& w2) {
+        if (w1.version < w2.version) return true;
+        if (w1.version > w2.version) return false;
+        if (w1.length < w2.length) return true;
+        if (w1.length > w2.length) return false;
+        if (memcmp(w1.program, w2.program, w1.length) < 0) return true;
```
danra: `return std::lexicographical_compare(...);`

src/bech32.cpp
sipa: Fixed. I've added a `reserve` call in `ExpandHRP` (which reserves 90 + hrp.size(), the maximum valid expanded size the checksum is computed over).

src/test/base58_tests.cpp
sipa: Good idea, fixed.

src/test/base58_tests.cpp
sipa: Fixed.

src/bech32.cpp
sipa: Fixed.

musabi:
Resolved?

src/bech32.cpp
```diff
@@ -0,0 +1,191 @@
...
+
+#include "bech32.h"
+
+namespace
+{
+
+typedef std::vector<uint8_t> data;
+
+/** The Bech32 character set for encoding. */
+const char* CHARSET = "qpzry9x8gf2tvdw0s3jn54khce6mua7l";
```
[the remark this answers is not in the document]
sipa: I think the implementation should focus on explaining why it is a correct implementation of the BIP. It doesn't need to include things the BIP already explains well - that's just gratuitously splitting the knowledge over multiple places.

Yes, the code for that optimization is in my repository, but it's very ad-hoc and not really intended for publishing.

src/bech32.cpp
```diff
@@ -0,0 +1,191 @@
...
+const int8_t CHARSET_REV[128] = {
+    -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+    15, -1, 10, 17, 21, 20, 26, 30,  7,  5, -1, -1, -1, -1, -1, -1,
+    -1, 29, -1, 24, 13, 25,  9,  8, 23, -1, 18, 22, 31, 27, 19, -1,
+     1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1,
+    -1, 29, -1, 24, 13, 25,  9,  8, 23, -1, 18, 22, 31, 27, 19, -1,
+     1,  0,  3, 16, 11, 28, 12, 14,  6,  4,  2, -1, -1, -1, -1, -1
+};
```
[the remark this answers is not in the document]
sipa: Sounds good, but seems out of scope for this PR.

src/bech32.cpp
```diff
@@ -0,0 +1,191 @@
...
+    // x^6 + {29}x^5 + {22}x^4 + {20}x^3 + {21}x^2 + {29}x + {18}. g(x) is chosen in such a way
+    // that the resulting code is a BCH code, guaranteeing detection of up to 3 errors within a
+    // window of 1023 characters. Among the various possible BCH codes, one was selected to in
+    // fact guarantee detection of up to 4 errors within a window of 89 characters.
+
+    // Note that the coefficients are elements of GF(32), here represented as decimal numbers
+    // between {}. In this finite field, addition is just XOR of the corresponding numbers. For
+    // example, {27} + {13} = {27 ^ 13} = {22}. Multiplication is more complicated, and requires
+    // treating the bits of values themselves as coefficients of a polynomial over a smaller field,
+    // GF(2), and multiplying those polynomials mod a^5 + a^3 + 1. For example, {5} * {26} =
```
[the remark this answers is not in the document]
sipa: Yes and no.

There is only one field of size 32 _up to isomorphism_. However, the multiplication of GF(32) elements _when represented as bits_ absolutely depends on the choice of the polynomial - and this is the one used in Bech32. It just so happens that you could do a mapping of the input values to the respective elements in an isomorphic representation (which uses another polynomial), do the multiplication, and map things back.

I think the current explanation is sufficiently accurate.

src/bech32.cpp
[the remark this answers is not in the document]
danra: Make sense regarding the information location. Where in your repo can one find the optimization code you used?

src/bech32.cpp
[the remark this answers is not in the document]
danra: > However, the multiplication of GF(32) elements when represented as bits absolutely depends on the choice of the polynomial - and this is the one used in Bech32
Yes, that's what I meant - emphasizing the fact that this polynomial was chosen, not that it is *the* one required for multiplication. I think this point could be better made. Though, I agree that even without this nit, the documentation is great :)

src/bech32.cpp
sipa: I prefer variables up front.

src/bech32.cpp
sipa: Thanks for taking the time to think through it, but no, this is about appending.

If the requirement was that the checksum was zero, it would mean that c(x) (the codeword polynomial) mod g(x) = 0, or in other words, c(x) is a multiple of g(x) (the generator). If c(x) is a multiple of g(x), then x*c(x) is a multiple of g(x) as well. The corresponding codeword for c(x)*x is the characters of the original with a 0 appended.

There is also a measure taken to avoid being able to prepend zeroes, namely the fact that for the codeword [a,b,c,d] we use `x^4 + a*x^3 + b*x^2 + c*x + d`, rather than `a*x^3 + b*x^2 + c*x + d` (see the comment above, about initializing the `c` variable to 1).

src/bech32.cpp
sipa: Fixed.

src/test/bech32_tests.cpp
sipa: Unfortunately, `tolower` is locale-dependent. I don't feel that the complexity of avoiding that is worth it for such a trivial algorithm.

src/test/base58_tests.cpp
sipa: Already removed in response to @laanwj's comment above.

src/base58.cpp
sipa: I think being explicit here is clearer. I also don't expect that this correspondence will be maintained in future witness versions.

danra:
there’s a version of tolower where you can specify the locale yourself, so there is no concern about the wrong one being set in the system

src/base58.cpp
sipa: It would, but I believe it would reduce readability: there are a number of alternatives being tried, and each has its branch to process it. If it were unindented, it would need to be reindented anyway if more cases were added.

src/base58.cpp
sipa: Fixed.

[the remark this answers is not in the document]
sipa: I prefer to do that later.

src/bech32.cpp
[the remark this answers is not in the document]
sipa: https://github.com/sipa/ezbase32/blob/master/dist32.cpp

src/bech32.cpp
sipa: This wasn't done originally, as the reference code is C++03, but by now it's diverged enough that I'll just bite the bullet and make it C++11 inside Core.

src/bech32.cpp
sipa: Done.

src/bech32.cpp
sipa: Done.

src/bech32.cpp
sipa: Done.

src/bech32.cpp
sipa: Done.

src/bech32.cpp
sipa: Cool, done.

src/test/bech32_tests.cpp
sipa: Locales scare me. As a reviewer, I would be far less confident that that code is correct due to it as opposed to an absolutely trivial algorithm that obvious does what it should.

src/base58.cpp
sipa: Done.

src/base58.cpp
sipa: Done.

src/script/standard.cpp
sipa: Done.

src/script/standard.cpp
sipa: Fixed.

src/script/standard.h
sipa: Done.

src/script/standard.h
sipa: Done.

sipa:
I believe I've addressed or at least responded to all comments by @promag, @laanwj, @theuni, and @danra.

I've also updated the Bech32 tests with some new additions to the BIP.

src/test/bech32_tests.cpp
theuni: Agree. Seeing "locale" always means a non-trivial amount of googling is in store.

src/bech32.cpp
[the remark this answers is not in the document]
sipa: Well I think it is _the_ one required for multiplication.

The specific representation of GF(32) we use, is defined as bits representing polynomials over GF(2) modulo exactly the polynomial `a^5 + a^3 + 1` and no other.

Yes, other representations using other polynomials exist, and given that all size 32 fields are isomorphic, it's guaranteed that a mapping exists between them which is consistent with addition and multiplication. But we're not using any of those :)

test/functional/segwit.py
```diff
@@ -570,6 +584,29 @@ def run_test(self):
...
+            rawtxfund = self.nodes[1].fundrawtransaction(transaction)['hex']
+            rawtxfund = self.nodes[1].signrawtransaction(rawtxfund)["hex"]
+            txid = self.nodes[1].sendrawtransaction(rawtxfund)
+
+            assert_equal(self.nodes[1].gettransaction(txid, True)["txid"], txid)
+            assert_equal(self.nodes[1].listtransactions("*", 1, 0, True)[0]["txid"], txid)
+
+            # Assert it is properly saved
+            self.stop_node(1)
+            self.start_node(1)
```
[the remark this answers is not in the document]
sipa:
```python
  File "./segwit.py", line 606, in run_test
    self.restart_node(1)
AttributeError: 'SegWitTest' object has no attribute 'restart_node'
```

src/base58.cpp
danra: I agree with the principle. But, the only line I suggest dedenting is
`if (ConvertBits<5, 8, false>(data, bech.second.begin() + 1, bech.second.end()))`
which is *after* a case has been decided (Bech32 decoding), and if this condition fails, the decoding fails, so early return makes sense. It's unlikely any additional cases would be added in case of failure here.

src/test/bech32_tests.cpp
danra: Ok, makes sense. In that case I adding a utility function for 'english' [is|to][lower|upper], there's too much repetition of the same (simple) functionality.

laanwj: Yes, that really doesn't help. Please don't introduce dependence on locale. We prefer deterministic, well-defined, nation-independent string processing, even if that means writing some extra code at least we **know** what it's doing then. Also this is specific to bechs32 (part of the standard) so it's better to keep it self-contained.

laanwj:
The commentary seems to have reached the level of nano-nits, which is likely an indication this is ready for merge.

src/bech32.cpp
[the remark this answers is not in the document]
danra: > Or do you mean that I should point out that Bech32 could have used any of the other irreducible polynomials?

Yes. I just suggest a more precise phrasing. e.g., instead of
> Multiplication is more complicated, and requires [...] multiplying those polynomials mod a^5 + a^3 + 1

write 

> Multiplication is more complicated, and requires [...] multiplying those polynomials mod an irreducible polynomial, chosen as a^5 + a^3 + 1 in Bech32."

src/test/bech32_tests.cpp
danra: @laanwj Just to repeat an earlier comment, `<locale>` has overloads of these functions where you can specify the locale, so there are no hidden dependencies on the default system locale. I do agree it might require more googling, and is perhaps a bit less obvious.

src/base58.cpp
danra: @theuni Good catch!
@sipa Suggest splitting `DecodeDestination` to two functions, one for decoding Base58 and one for Bech32. This would have trivially eliminated the bug above. Another example of something which would be fixed is `hash` being used only for Base58 decoding but being visible for the entirety of the Bech32 decoding as well.

TheBlueMatt (commented):
postumous Matt was here and didn't finish review.

src/base58.cpp
theuni: +1 to this in some future follow-up.

src/bech32.cpp
```diff
@@ -0,0 +1,191 @@
...
+}
+
+/** Expand a HRP for use in checksum computation. */
+data ExpandHRP(const std::string& hrp)
+{
+    data ret;
+    ret.reserve(hrp.size() + 90);
+    ret.resize(hrp.size() * 2 + 1);
+    for (size_t i = 0; i < hrp.size(); ++i) {
+        unsigned char c = hrp[i];
```
theuni: Unlikely to matter, but shouldn't this be ```LowerCase(hrp[i])``` ?

MarcoFalke:
Removing from backport

PR #11169, opened 2017-08-27
[GUI] Make tabs toolbar no longer have a context menu

### LINKED ISSUE #11168: [GUI] Tabs Toolbar menu option makes tabs disappear for the session
In the GUI, if you right click on the tabs at the top, a menu appears with a checkbox item that says `Tabs Toolbar`. Clicking that makes the tabs at the top disappear. However there is no way for the tabs to be brought back except by restarting the software.

IMO it really isn't useful to hide the tabs and I'm not sure if there is an easy way to make a menu which brings them back.

achow101:
Adds a contextMenuPolicy of Qt::PreventContextMenu to prevent the tabs toolbar from showing a context menu that allows it to be hidden.

Fixes #11168

### DIFF
--- a/src/qt/bitcoingui.cpp
+++ b/src/qt/bitcoingui.cpp
@@ -454,6 +454,7 @@ void BitcoinGUI::createToolBars()
     if(walletFrame)
     {
         QToolBar *toolbar = addToolBar(tr("Tabs toolbar"));
+        toolbar->setContextMenuPolicy(Qt::PreventContextMenu);
         toolbar->setMovable(false);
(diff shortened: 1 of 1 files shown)


MeshCollider:
Trivial utACK, makes a lot of sense. Travis failure is with p2p-segwit, not related

sipa:
Concept ACK

MarcoFalke:
Concept ACK

laanwj:
Tested ACK e254830
I'm surprised this we discover this hidden "feature" only now.

hebasto:
The ultimate fix of #11168 is https://github.com/bitcoin-core/gui/pull/219 :)

PR #11170, opened 2017-08-27
Add label support to "addwitnessaddress"

rawodb:


gmaxwell:
This RPC should be going away, not enhanced. It's a testing shim.

MeshCollider:
Agree with Greg, hopefully will be properly replaced and removed soon so it's a NACK from me

rawodb:
I see no other option to add witness addresses at this point.

The prior createwitnessaddress that didn't have a solvable check was removed.

PR #11172, opened 2017-08-28
Fixed a grammar issue.

ChandlerScheurkogel:


MeshCollider:
A little bit picky imo, it's not even wrong in the first place. Probably should be 'the' anyway

fanquake:
trivial ACK 1c6ee8a

promag:
Agree with @MeshCollider, the "fix" isn't worth it.

Either way, remove period from PR title and commit message and add reference to `linearize documentation`?

laanwj:
Yeah, I agree this isn't worth it, also the old sentence seems fine to me.

Leviathn:
Old sentence was perfectly fine.

PR #11173, opened 2017-08-28
RPC: Fix currency unit string in the help text

AkioNak:
1. The RPC help text should use the constant `CURRENCY_UNIT` defined in `policy/feerate.cpp` instead of the literal `'BTC'`. 
  In the following 2 RPC commands, `'BTC'` is written directly in the help text.
  This commit changes them to use that constant.
     1) `estimatesmartfee`
     2) `estimaterawfee`

2. Some RPC command use `'satoshis'` as the unit. 
  It should be written as `'satoshis'` instead of `'Satoshis'` in the RPC help text.
  So, this commit fixes this typo in `getblocktemplate`.


MeshCollider:
utACK https://github.com/bitcoin/bitcoin/pull/11173/commits/54cd27b096d8542cf60c76f122de3606a83ba17b

instagibbs:
utACK https://github.com/bitcoin/bitcoin/pull/11173/commits/54cd27b096d8542cf60c76f122de3606a83ba17b

MarcoFalke:
utACK 54cd27b

jnewbery:
For both `feerate` help strings, I think the following text is more precise:

`"(numeric, optional) estimated fee rate (in " + CURRENCY_UNIT + "-per-kilobyte)\n"`

AkioNak:
@jnewbery Thank you for your comment.
In other RPCs, there are 2 patterns for the notation of fee rate unit like this case:
  1. `... + CURRENCY_UNIT + "/kB" ...` (In most cases)
  2. `... + CURRENCY_UNIT + " per KB" ...`
        (getmempoolinfo - mempoolminfee, fundrawtransaction - feeRate)

So I think case 1. is better than using `-per-kilobyte` for these help strings of this PR.
And case 2, I think it is better to modify `per KB` to `/kB` because KB implies KiB or Kibibyte (2^10 byte) and they just display the **unit**.

AkioNak:
@jnewbery I fixed the `feerate` help strings.

MeshCollider:
`/kB` is better than `-per-kilobyte` imo but I prefer @jnewberys string overall with that small adjustment.

Also please squash commits into 1

jnewbery:
New help strings look good to me. Please squash commits and I'll re-review.

AkioNak:
@jnewbery @MeshCollider I squashed the commits.

MeshCollider:
re-utACK https://github.com/bitcoin/bitcoin/pull/11173/commits/47ba2c312a440baff5ebc911e3bb014643f4e9c3

jnewbery:
Tested ACK 47ba2c312a440baff5ebc911e3bb014643f4e9c3

MarcoFalke:
re-utACK 47ba2c3

PR #11176, opened 2017-08-28
build: Rename --enable-experimental-asm to --enable-asm and enable by default

laanwj:
Now that 0.15 is branched off, enable assembler SHA256 optimizations by default, but still allow disabling them, for example if something goes wrong with auto-detection on a platform.

Also add mention of the use of asm in the configure summary.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1297,6 +1297,7 @@ echo "  with zmq      = $use_zmq"
 echo "  with test     = $use_tests"
 echo "  with bench    = $use_bench"
 echo "  with upnp     = $use_upnp"
+echo "  use asm       = $use_asm"
 echo "  debug enabled = $enable_debug"
 echo "  werror        = $enable_werror"
 echo 
--- a/configure.ac
+++ b/configure.ac
@@ -177,14 +177,14 @@ AC_ARG_ENABLE([glibc-back-compat],
   [use_glibc_compat=$enableval],
   [use_glibc_compat=no])
 
-AC_ARG_ENABLE([experimental-asm],
-  [AS_HELP_STRING([--enable-experimental-asm],
-  [Enable experimental assembly routines (default is no)])],
-  [experimental_asm=$enableval],
-  [experimental_asm=no])
+AC_ARG_ENABLE([asm],
+  [AS_HELP_STRING([--enable-asm],
+  [Enable assembly routines (default is yes)])],
+  [use_asm=$enableval],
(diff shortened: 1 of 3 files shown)


promag:
Should the flag mention SHA256 like `--enable-asm-sha256`?

Either way LGTM.

laanwj:
> Should the flag mention SHA256 like --enable-asm-sha256?

No, I think we should have a general asm flag, build-side, as this does. This is similar to OpenSSL's flag to use assembly or not. Something else creates too many combinations to test, and I don't see why it would be useful.

promag:
Ok, just had to ask.

utACK 538cc0c.

kallewoof:
If this is meant for 0.15, the [release notes](https://github.com/bitcoin/bitcoin/blob/0.15/doc/release-notes.md) need to be updated. They currently read

> SHA256 hashing has been optimized for architectures supporting SSE 4 (See PR 10182). SHA256 is around 50% faster on supported hardware, which results in around 5% faster IBD and block validation. In version 0.15, SHA256 hardware optimization is disabled in release builds by default, but can be enabled by using --enable-experimental-asm when building.

laanwj:
No, this is not meant for 0.15, certainly not for 0.15.0

> Now that 0.15 is branched off, enable assembler SHA256 optimizations by default

kallewoof:
Sorry about that. I should've read the PR message more closely.

sipa:
@laanwj There are plenty more versions of SHA256 to add (for starters, a RORX version, a SHA-NI version, but later perhaps a 4-way parallel SSE4 or 8-way parallel AVX2 version). Do we all want to include them under "--enable-asm", or will they first go into a reinstated "--enable-experimental-asm" before being promoted?

laanwj:
> Do we all want to include them under "--enable-asm" or will they first go into a reinstated "--enable-experimental-asm" before being promoted?

Both are valid options, we should probably decide on a case by case basis - depending on how much we trust the code, how long it can be in master until the release.

luke-jr (approved):
utACK

fanquake:
utACK 538cc0c

droark:
Posthumous review comment: Is USE_ASM meant to be a catch-all for any other assembly optimizations added in the future, or is it solely for the SHA256 optimizations? I ask because the language in the changes is a little ambiguous (IMO). I'd like to fix that. I just don't want to step on toes first.

Thanks.

PR #11177, opened 2017-08-28
Support for SegWit Addresses in RPC  calls and change addresses

rawodb:
This is the approach I've taken to enabling segwit addresses for rpc calls on some bitcoin nodes I run.

* **-defaultwitnessaddress** startup parameters has been added to return segwitaddresses by defaults
* **Witnessifier** visitor has been moved to the top of the file for use in getaccountaddress/getnewaddress

* **getaccountaddress** has been added the optional witness flag which will return Witnessified addresses
* **getnewaddress** has been added the optional witness flag which will return Witnessified addresses
* **addwitnessaddress** has been added an optional label parameter
* Added check for witness used destination scripts in GetAccountPubkey 

* Change addresses default to P2SH wrapped segwit addresses when **-defaultwitnessaddress** is used

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -453,6 +453,7 @@ std::string HelpMessage(HelpMessageMode mode)
         strUsage += HelpMessageOpt("-limitdescendantcount=<n>", strprintf("Do not accept transactions if any ancestor would have <n> or more in-mempool descendants (default: %u)", DEFAULT_DESCENDANT_LIMIT));
         strUsage += HelpMessageOpt("-limitdescendantsize=<n>", strprintf("Do not accept transactions if any ancestor would have more than <n> kilobytes of in-mempool descendants (default: %u).", DEFAULT_DESCENDANT_SIZE_LIMIT));
         strUsage += HelpMessageOpt("-vbparams=deployment:start:end", "Use given start/end times for specified version bits deployment (regtest-only)");
+        strUsage += HelpMessageOpt("-defaultwitnessaddress", "Return witness addresses by default in RPC calls (default: false)");
     }
     strUsage += HelpMessageOpt("-debug=<category>", strprintf(_("Output debugging information (default: %u, supplying <category> is optional)"), 0) + ". " +
         _("If <category> is not supplied or if <category> = 1, output all debugging information.") + " " + _("<category> can be:") + " " + ListLogCategories() + ".");
--- a/src/rpc/client.cpp
+++ b/src/rpc/client.cpp
@@ -120,6 +120,7 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "estimaterawfee", 1, "threshold" },
     { "prioritisetransaction", 1, "dummy" },
     { "prioritisetransaction", 2, "fee_delta" },
+    { "getnewaddress", 1, "witness" },
     { "setban", 2, "bantime" },
     { "setban", 3, "absolute" },
     { "setnetworkactive", 0, "state" },
--- a/src/rpc/client.cpp
+++ b/src/rpc/client.cpp
@@ -120,6 +120,7 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "estimaterawfee", 1, "threshold" },
     { "prioritisetransaction", 1, "dummy" },
     { "prioritisetransaction", 2, "fee_delta" },
+    { "getaccountaddress", 1, "witness" },
     { "getnewaddress", 1, "witness" },
     { "setban", 2, "bantime" },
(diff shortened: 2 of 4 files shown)


src/init.cpp
```diff
@@ -453,6 +453,7 @@ std::string HelpMessage(HelpMessageMode mode)
         strUsage += HelpMessageOpt("-limitdescendantcount=<n>", strprintf("Do not accept transactions if any ancestor would have <n> or more in-mempool descendants (default: %u)", DEFAULT_DESCENDANT_LIMIT));
         strUsage += HelpMessageOpt("-limitdescendantsize=<n>", strprintf("Do not accept transactions if any ancestor would have more than <n> kilobytes of in-mempool descendants (default: %u).", DEFAULT_DESCENDANT_SIZE_LIMIT));
         strUsage += HelpMessageOpt("-vbparams=deployment:start:end", "Use given start/end times for specified version bits deployment (regtest-only)");
+        strUsage += HelpMessageOpt("-defaultwitnessaddress", "Return witness addresses by default in rpc calls");
```
promag: Upper case RPC. Add default value.

src/rpc/client.cpp
```diff
@@ -120,6 +120,8 @@ static const CRPCConvertParam vRPCConvertParams[] =
     { "estimaterawfee", 1, "threshold" },
     { "prioritisetransaction", 1, "dummy" },
     { "prioritisetransaction", 2, "fee_delta" },
+    { "getaccountaddress", 1, "witness" },
```
promag: This call is deprecated. Don't change it.

src/wallet/rpcwallet.cpp
```diff
@@ -221,10 +224,25 @@ UniValue getnewaddress(const JSONRPCRequest& request)
         throw JSONRPCError(RPC_WALLET_KEYPOOL_RAN_OUT, "Error: Keypool ran out, please call keypoolrefill first");
     }
     CKeyID keyID = newKey.GetID();
+    CBitcoinAddress addr = CBitcoinAddress(keyID);
 
     pwallet->SetAddressBook(keyID, strAccount, "receive");
+    bool witnessify = gArgs.GetBoolArg("-defaultwitnessaddress", false);
+    if (request.params.size() >= 2) {
```
promag: Remove this test, `isNull()` below is enough because `request.params[]` returns null `UniValue` for out of bounds elements.

src/wallet/rpcwallet.cpp
```diff
@@ -221,10 +224,25 @@ UniValue getnewaddress(const JSONRPCRequest& request)
...
+        if (!request.params[1].isNull()) {
+            witnessify = request.params[1].get_bool();
+        }
+    }
 
-    return CBitcoinAddress(keyID).ToString();
+    if (witnessify) {
+		Witnessifier w(pwallet);
```
promag: Wrong indentation.

src/wallet/rpcwallet.cpp
```diff
@@ -1265,7 +1266,11 @@ UniValue addwitnessaddress(const JSONRPCRequest& request)
         throw JSONRPCError(RPC_WALLET_ERROR, "Public key or redeemscript not known to wallet, or the key is uncompressed");
     }
 
-    pwallet->SetAddressBook(w.result, "", "receive");
+    std::string strLabel = "";
+    if (request.params.size() >= 2) {
```
promag: !isNull() instead. See developer notes.

src/wallet/rpcwallet.cpp
```diff
@@ -1265,7 +1266,11 @@ UniValue addwitnessaddress(const JSONRPCRequest& request)
         throw JSONRPCError(RPC_WALLET_ERROR, "Public key or redeemscript not known to wallet, or the key is uncompressed");
     }
 
-    pwallet->SetAddressBook(w.result, "", "receive");
+    std::string strLabel = "";
```
promag: Just `std::string label;`

promag:
Please don't mention users in commit messages. You also have to fixup some commits, for instance,  you should add `-defaultwitnessaddress` help message in the first commit you use the argument.

sipa:
I don't think we should be making witness addresses configurable on a per-address basis. Because of automatic keypool topic during rescan in HD mode, we need to automatically witnessify newly created keys anyway, as soon as there is a possibility of a witness address.

I'd prefer a wallet-wide setting that turns all addresses (including change) into witnesses.

rawodb:
Rebased off master, added support for P2SH wrapped change addresses when using -defaultwitnessaddress

PR #11178, opened 2017-08-28
Add iswitness parameter to decode- and fundrawtransaction RPCs

meshcollider:
Suggested in https://github.com/bitcoin/bitcoin/pull/10481#issuecomment-325244946, this adds the option to explicitly choose whether a serialized transaction should be decoded as a witness or non-witness transaction rather than relying on the heuristic checks in #10481. The parameter defaults to relying on #10481 if not included, but it overrides that if included.


MeshCollider:
@achow101 @roconnor-blockstream you might want to take a look at this :)

src/core_read.cpp
```diff
@@ -108,7 +108,7 @@ bool CheckTxScriptsSanity(const CMutableTransaction& tx)
     return true;
 }
 
-bool DecodeHexTx(CMutableTransaction& tx, const std::string& strHexTx, bool fTryNoWitness)
+bool DecodeHexTx(CMutableTransaction& tx, const std::string& strHexTx, bool fTryNoWitness, bool fTryWitness)
```
achow101: `snake_case`

achow101 (commented):
utACK 55e2b9c8a13b5f80d03936f4275cd9e3828e4b3f

MeshCollider:
Changed style as requested, thanks @achow101

luke-jr:
Please don't change the style of *existing* code, only the new code... :/

MeshCollider:
@luke-jr the style refactor only touches a tiny amount of code inside that single function, in a similar way to adding braces to existing if-statements when modifying nearby code. I'd argue that this small change to existing code style is worth it for the consistency, especially between try_witness and try_no_witness

promag (commented):
Missing tests.

src/core_read.cpp
```diff
@@ -108,38 +108,44 @@ bool CheckTxScriptsSanity(const CMutableTransaction& tx)
...
+        catch (const std::exception&) {
             return false;
         }
     }
-    catch (const std::exception&) {
-        return false;
-    }
-
+    
     return true;
```
sipa: If you move this `return true;` inside the `if (try_witness)` block, the `if(!try_witness)` inside the `catch` above becomes unnecessary. I think that'd be a cleaner flow.

src/rpc/rawtransaction.cpp
```diff
@@ -418,13 +418,15 @@ UniValue createrawtransaction(const JSONRPCRequest& request)
...
-    if (request.fHelp || request.params.size() != 1)
+    if (request.fHelp || request.params.size() < 1 || request.params.size() > 2)
         throw std::runtime_error(
-            "decoderawtransaction \"hexstring\"\n"
+            "decoderawtransaction \"hexstring\" ( iswitness )\n"
             "\nReturn a JSON object representing the serialized, hex-encoded transaction.\n"
 
             "\nArguments:\n"
             "1. \"hexstring\"      (string, required) The transaction hex string\n"
+            "2. iswitness          (boolean, optional) Whether the transaction hex is a serialized witness transaction \n"
```
sipa: Nit: trailing space

MeshCollider: Fixed

src/core_read.cpp
MeshCollider: Good idea yep, done

src/core_read.cpp
```diff
@@ -129,18 +129,22 @@ bool DecodeHexTx(CMutableTransaction& tx, const std::string& strHexTx, bool fTry
...
-        if (!ssData.empty()) {
+    if (try_witness) {
+        CDataStream ssData(txData, SER_NETWORK, PROTOCOL_VERSION);
+        try {
+            ssData >> tx;
+            if (!ssData.empty()) {
+                return false;
+            }
+        }
+        catch (const std::exception&) {
```
sipa: Nit: `catch` on the same line as `}`.

src/core_read.cpp
```diff
@@ -129,18 +129,22 @@ bool DecodeHexTx(CMutableTransaction& tx, const std::string& strHexTx, bool fTry
...
-    CDataStream ssData(txData, SER_NETWORK, PROTOCOL_VERSION);
-    try {
-        ssData >> tx;
-        if (!ssData.empty()) {
+    if (try_witness) {
+        CDataStream ssData(txData, SER_NETWORK, PROTOCOL_VERSION);
+        try {
+            ssData >> tx;
+            if (!ssData.empty()) {
+                return false;
```
sipa: Even shorter: invert this test and make it return true.

MeshCollider: Nice :+1:

src/core_read.cpp
```diff
@@ -108,39 +108,39 @@ bool CheckTxScriptsSanity(const CMutableTransaction& tx)
...
-        if (!ssData.empty()) {
+    if (try_witness) {
+        CDataStream ssData(txData, SER_NETWORK, PROTOCOL_VERSION);
+        try {
+            ssData >> tx;
+            if (ssData.empty()) {
+                return true;
+            }
+        } catch (const std::exception&) {
             return false;
```
sipa: And now this line is unnecessary as well :)

MeshCollider: Heh, changed. Next you'll want me to move the try-catch outside the if statements so there only needs to be one of them ;) Thanks for reviewing!

sipa: Nope, that wouldn't actually work. If both flags were true, and the first attempt to deserialize throws, you still want to try the second.

src/core_read.cpp
```diff
@@ -108,39 +108,39 @@ bool CheckTxScriptsSanity(const CMutableTransaction& tx)
...
-    std::vector<unsigned char> txData(ParseHex(strHexTx));
+    std::vector<unsigned char> txData(ParseHex(hex_tx));
 
-    if (fTryNoWitness) {
+    if (try_no_witness) {
         CDataStream ssData(txData, SER_NETWORK, PROTOCOL_VERSION | SERIALIZE_TRANSACTION_NO_WITNESS);
         try {
             ssData >> tx;
-            if (ssData.eof() && CheckTxScriptsSanity(tx)) {
+            if (ssData.eof() && (CheckTxScriptsSanity(tx) || !try_witness)) {
```
sipa: One final tiny nit: swap the arguments, as checking a boolean is much faster than CheckTxScriptsSanity.

sipa:
utACK 3d16658226cd08911200dbd90db762dc3409e443

MeshCollider:
Rebased

ryanofsky (commented):
utACK 6f39ac04375a5f6ef803da59ba0b606123d63142

Probably could merge given 3 acks.

laanwj:
utACK 6f39ac04375a5f6ef803da59ba0b606123d63142

PR #11184, opened 2017-08-28
Use IsMine to validate custom change address

### LINKED ISSUE #11137: Coin Control doesn't recognize my segwit address as a valid change address
### Describe the issue

1. Open wallet on testnet.
1. Create an address.
2. run addwitnessaddress with new address.
3. Send money with coin control enabled:
    * Check 'Custom Change Address'
    * Specify segwit address

### Expected behaviour

Should work just fine.

### Actual behaviour

1. Get a warning saying I don't control this address, and presumably won't actually get the change.

2. I do get the change.

### Screenshots.
If the issue is related to the GUI, screenshots can be added to this issue via drag & drop.

0.14.2

screen shot 2017-08-25 at 12 15 35 pm

dooglus:
... so it recogizes SegWit change addresses.

Fixes #11137.

### DIFF
--- a/src/qt/sendcoinsdialog.cpp
+++ b/src/qt/sendcoinsdialog.cpp
@@ -789,10 +789,8 @@ void SendCoinsDialog::coinControlChangeEdited(const QString& text)
         }
         else // Valid address
         {
-            CKeyID keyid;
-            addr.GetKeyID(keyid);
-            if (!model->havePrivKey(keyid)) // Unknown change address
-            {
+            const CTxDestination dest = addr.Get();
+            if (!model->IsSpendable(dest)) {
                 ui->labelCoinControlChangeLabel->setText(tr("Warning: Unknown change address"));
 
                 // confirmation dialog
@@ -800,7 +798,7 @@ void SendCoinsDialog::coinControlChangeEdited(const QString& text)
                     QMessageBox::Yes | QMessageBox::Cancel, QMessageBox::Cancel);
 
                 if(btnRetVal == QMessageBox::Yes)
-                    CoinControlDialog::coinControl->destChange = addr.Get();
+                    CoinControlDialog::coinControl->destChange = dest;
                 else
                 {
                     ui->lineEditCoinControlChange->setText("");
@@ -819,7 +817,7 @@ void SendCoinsDialog::coinControlChangeEdited(const QString& text)
                 else
                     ui->labelCoinControlChangeLabel->setText(tr("(no label)"));
 
-                CoinControlDialog::coinControl->destChange = addr.Get();
+                CoinControlDialog::coinControl->destChange = dest;
             }
         }
     }
--- a/src/qt/walletmodel.cpp
+++ b/src/qt/walletmodel.cpp
@@ -561,9 +561,9 @@ bool WalletModel::getPubKey(const CKeyID &address, CPubKey& vchPubKeyOut) const
     return wallet->GetPubKey(address, vchPubKeyOut);
 }
 
-bool WalletModel::havePrivKey(const CKeyID &address) const
+bool WalletModel::IsSpendable(const CTxDestination& dest) const
 {
-    return wallet->HaveKey(address);
+    return IsMine(*wallet, dest) & ISMINE_SPENDABLE;
 }
 
 bool WalletModel::getPrivKey(const CKeyID &address, CKey& vchPrivKeyOut) const
--- a/src/qt/walletmodel.h
+++ b/src/qt/walletmodel.h
@@ -190,7 +190,7 @@ public:
     UnlockContext requestUnlock();
 
     bool getPubKey(const CKeyID &address, CPubKey& vchPubKeyOut) const;
-    bool havePrivKey(const CKeyID &address) const;
+    bool IsSpendable(const CTxDestination& dest) const;
     bool getPrivKey(const CKeyID &address, CKey& vchPrivKeyOut) const;
     void getOutputs(const std::vector<COutPoint>& vOutpoints, std::vector<COutput>& vOutputs);
     bool isSpent(const COutPoint& outpoint) const;


src/qt/sendcoinsdialog.cpp
```diff
@@ -789,9 +789,7 @@ void SendCoinsDialog::coinControlChangeEdited(const QString& text)
         }
         else // Valid address
         {
-            CKeyID keyid;
-            addr.GetKeyID(keyid);
-            if (!model->havePrivKey(keyid)) // Unknown change address
+            if (model->isMine(addr.Get()) != ISMINE_SPENDABLE) // Unknown change address
```
promag: IMO remove the comment, the code is more expressive.

src/qt/walletmodel.cpp
```diff
@@ -561,9 +561,9 @@ bool WalletModel::getPubKey(const CKeyID &address, CPubKey& vchPubKeyOut) const
     return wallet->GetPubKey(address, vchPubKeyOut);
 }
 
-bool WalletModel::havePrivKey(const CKeyID &address) const
+isminetype WalletModel::isMine(const CTxDestination &dest) const
```
promag: `const CTxDestination& dest`.

src/qt/walletmodel.h
```diff
@@ -7,6 +7,7 @@
 
 #include "paymentrequestplus.h"
 #include "walletmodeltransaction.h"
+#include "script/ismine.h"
```
promag: Sort include.

src/qt/walletmodel.h
```diff
@@ -190,7 +191,7 @@ class WalletModel : public QObject
     UnlockContext requestUnlock();
 
     bool getPubKey(const CKeyID &address, CPubKey& vchPubKeyOut) const;
-    bool havePrivKey(const CKeyID &address) const;
+    isminetype isMine(const CTxDestination &dest) const;
```
promag: `const CTxDestination& dest`.

src/qt/sendcoinsdialog.cpp
promag: Move `{` to this line.

src/qt/walletmodel.cpp
MarcoFalke: function names are PascalCase.

src/qt/walletmodel.h
promag: Another option, just for the record, that doesn't require to include `script/ismine.h`, is `bool isSpendable(const CKeyID& address)`.

src/qt/walletmodel.cpp
dooglus: Where can I find guidelines about where to put the whitespace? I went with the existing layout. See how the line I replaced had `(const CKeyID &address)`?

src/qt/sendcoinsdialog.cpp
MarcoFalke: if you feel like it, you might add a `const CTxDestination dest = addr.Get();` before that line and then use `if (!bool(model->IsMine(dest) & ISMINE_SPENDABLE))`.

Though, it shouldn't make a difference.

maflcko (commented):
utACK 89daeb84915d7fd21230771ed266dde1f7d7afda

src/qt/walletmodel.cpp
MarcoFalke: This is mentioned in the developer notes. Though, you could always just run clang format (also mentioned in the developer notes)

src/qt/walletmodel.h
dooglus: I don't see `bool isSpendable(const CKeyID& address)` anywhere. And the point is that we want to be able to check the spendability of general addresses, not only `CKeyID`s.

promag:
BTW, squash and reword, commit message is too long.

dooglus:
Addressed @MarcoFalke's nit re. masking out ISMINE_SPENDABLE.

Squashed.

Reworded commit message.

src/qt/walletmodel.cpp
dooglus: Check `src/qt/walletmodel.h` - they're all lowerThenUpper

promag:
Thanks @dooglus but could you remove period. Suggestion `Use IsMine to validate custom change address`.

src/qt/walletmodel.cpp
promag: Yes but it's outdated.

src/qt/sendcoinsdialog.cpp
```diff
@@ -789,10 +789,8 @@ void SendCoinsDialog::coinControlChangeEdited(const QString& text)
         }
         else // Valid address
         {
-            CKeyID keyid;
-            addr.GetKeyID(keyid);
-            if (!model->havePrivKey(keyid)) // Unknown change address
-            {
+            const CTxDestination dest = addr.Get();
+            if (!bool(model->isMine(dest) & ISMINE_SPENDABLE)) {
```
sipa: The `bool(..)` is unnecessary.

dooglus:
Removed period from commit message

Removed `bool`

Squashed again

dooglus:
Replaced commit message with @promag's suggested one

src/qt/walletmodel.cpp
dooglus: So what's the plan? Shouldn't we go through the code fixing all the outdated style at once rather than fixing one at a time?

dooglus: I mean, I can change just the one I touched if that's preferable.

dooglus:
It appears that calling the function `IsMine` instead of `isMine` caused an error:

      CXX      qt/qt_libbitcoinqt_a-walletmodeltransaction.o
    qt/walletmodel.cpp:566:12: error: too many arguments to function call, expected 1, have 2; did you mean '::IsMine'?
        return IsMine(*wallet, dest);
               ^~~~~~
               ::IsMine
    ./script/ismine.h:39:12: note: '::IsMine' declared here
    isminetype IsMine(const CKeyStore& keystore, const CTxDestination& dest, SigVersion = SIGVERSION_BASE);
               ^

I'll call it 'IsSpendable()` instead.

MarcoFalke:
utACK 848ef973d7ecb054e0d6d6952fd69a0816f3b7cd

promag:
@dooglus i believe you only had to call like `::IsMine()`. Either way please update PR title too, and elaborate more about the motivation in the description. Ty.

MeshCollider:
@promag title is still correct I believe, because it still uses IsMine() to check

promag:
@MeshCollider it's too long 🙄.

dooglus:
I finally understood what you meant about IsSpendable(), so did it that way. IsSpendable() calls IsMine(), so the commit log is correct. I've used the same string for the PR title.

sipa:
Should change to a watch-only address be permitted?

dooglus:
Change to any address is permitted. The warning is for if you do not have control of the address.

promag:
> should change to a watch-only address be permitted?

@sipa follow up discussion? For now `ISMINE_SPENDABLE` is clear and does prevents possible problems that `ISMINE_WATCH_ONLY` could have.

src/qt/sendcoinsdialog.cpp
```diff
@@ -789,10 +789,8 @@ void SendCoinsDialog::coinControlChangeEdited(const QString& text)
         }
         else // Valid address
         {
-            CKeyID keyid;
-            addr.GetKeyID(keyid);
-            if (!model->havePrivKey(keyid)) // Unknown change address
-            {
+            const CTxDestination dest = addr.Get();
+            if (!model->IsSpendable(dest)) {
```
jonasschnelli: Wouldn't `if (!model->IsSpendable(addr.Get())) {` be simpler?

jonasschnelli:
utACK 848ef973d7ecb054e0d6d6952fd69a0816f3b7cd

src/qt/sendcoinsdialog.cpp
dooglus: Yes it would, and that's how I had it until @MarcoFalke [made this comment](https://github.com/bitcoin/bitcoin/pull/11184#discussion_r135662398).

Is there a style guide that tells me how such things should be? Lots of people seem to have ideas about what's right, but I'd like to see something in writing, as it were.

MarcoFalke: `dest` can be used in a few places below, but as I mentioned above this is just personal preference for one style over another. Just use what you feel makes more sense, no need to discuss.

dooglus: Good point. I use it in the few places below now too.

MeshCollider:
utACK https://github.com/bitcoin/bitcoin/pull/11184/commits/c41224dfd51c896341bbf2fa23e160bf5ffe27c3

promag:
utACK c41224d.

practicalswift:
Concept ACK! Solves the issue discussed in my PR #11184 which I've now closed.

laanwj:
Ouch, did no one notice that this was filed against 0.15? Had to revert it.
We need a new PR for master.

PR #11190, opened 2017-08-29
[tests] Remove unused imports (script.OP_0 and script.sha256)

practicalswift:
Remove unused imports (`script.OP_0` and `script.sha256`).

### DIFF
--- a/test/functional/test_framework/address.py
+++ b/test/functional/test_framework/address.py
@@ -4,7 +4,7 @@
 # file COPYING or http://www.opensource.org/licenses/mit-license.php.
 """Encode and decode BASE58, P2PKH and P2SH addresses."""
 
-from .script import hash256, hash160, sha256, CScript, OP_0
+from .script import CScript, hash160, hash256
 from .util import bytes_to_hex_str, hex_str_to_bytes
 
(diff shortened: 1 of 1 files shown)


MeshCollider:
The typo fix is wrong unfortunately, should be 'a' (because 'one' effectively starts with a 'w')

fanquake:
Agree with @MeshCollider that typo fix isn't correct.

practicalswift:
@MeshCollider @fanquake Thanks for reviewing! Typo fix removed :-)

practicalswift:
The build error seems unrelated.

MeshCollider:
Yeah Concept ACK for removing the unused import

jnewbery:
There are many instances in the test/functional/test_framework directory of unused imports, wildcard imports, names being imported indirectly through wildcard imports, imports on the same line, imports not in PEP8 ordering, imports not being sorted, etc.

I'd rather have a single PR that fixes all of those nits, rather than tackle it file-by-file.

That said, utACK c7e0abe1904bac179771985c5223a323a17f97f5 if people disagree with my preferred approach.

MarcoFalke:
Agree with @jnewbery if a fix is reasonable to do for the whole project in a single commit, that should be preferred. However, if that is not feasible for whatever reason or not worth it, nothing should be touched. (Unless you need to change the line anyway.)

PR #11191, opened 2017-08-29
RPC: Improve help text and behavior of RPC-logging.

AkioNak:
1. It is allowed `libevent` logging to be updated during runtime,
  but still described that restriction in the help text.
  So we delete these text.
2. Add a descrption about the evaluation order of `<include>` and
  `<exclude>` to clarify how debug loggig categories to be set.
3. Add a description about the available logging category `"all"`
  which is not explained.
4. Add `"optional"` to the help text of `<include>` and `<exclude>`.
5. Add missing new lines before `"Argument:"`.
6. `"0"`,`"1"` are allowed in both array of `<include>` and `<exclude>`.
  `"0"` is **ignored** and `"1"` is treated **same as** `"all"`.
  It is confusing, so forbid them.
7. It always returns all logging categories with status.
  Fix the help text to match this behavior.

### DIFF
--- a/src/init.cpp
+++ b/src/init.cpp
@@ -925,7 +925,8 @@ bool AppInitParameterInteraction()
         // Special-case: if -debug=0/-nodebug is set, turn off debugging messages
         const std::vector<std::string> categories = gArgs.GetArgs("-debug");
 
-        if (find(categories.begin(), categories.end(), std::string("0")) == categories.end()) {
+        if (std::none_of(categories.begin(), categories.end(),
+            [](std::string cat){return cat == "0" || cat == "none";})) {
             for (const auto& cat : categories) {
                 uint32_t flag = 0;
                 if (!GetLogCategory(&flag, &cat)) {
--- a/src/rpc/misc.cpp
+++ b/src/rpc/misc.cpp
@@ -577,6 +577,9 @@ uint32_t getCategoryMask(UniValue cats) {
         if (!GetLogCategory(&flag, &cat)) {
             throw JSONRPCError(RPC_INVALID_PARAMETER, "unknown logging category " + cat);
         }
+        if (flag == BCLog::NONE) {
+            return 0;
+        }
         mask |= flag;
     }
     return mask;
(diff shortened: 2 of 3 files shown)


src/rpc/misc.cpp
```diff
@@ -574,7 +574,7 @@ uint32_t getCategoryMask(UniValue cats) {
     for (unsigned int i = 0; i < cats.size(); ++i) {
         uint32_t flag = 0;
         std::string cat = cats[i].get_str();
-        if (!GetLogCategory(&flag, &cat)) {
+        if (!GetLogCategory(&flag, &cat) || (cat == "0" || cat == "1")) {
```
MarcoFalke: According to the help text, this is a valid "category":

```
  -debug=<category>
       Output debugging information (default: 0, ...

**Comment by MarcoFalke** [comment]
Concept ACK

File: src/rpc/misc.cpp (hunk 1 above)
Comment (AkioNak) [reply src/rpc/misc.cpp]: @MarcoFalke Thank you for your review.
Oh, indeed.
I think `-debug=1`(or0) is very clear for initial parameter, but I'm afraid that it is a little bit hard to understand that `"0"` is ignored in this RPC command.
So, I will add `{BCLog::NONE, "none"}` to `LogCategories[]` and add to the help text that `"0"` and `"1"` are the shorthand of `"none"` and `"all"`.

Comment (MarcoFalke) [reply src/rpc/misc.cpp]: I wasn't even aware of this rpc. You might want to wait for feedback by someone who used this.

Comment (AkioNak) [reply src/rpc/misc.cpp]: Then, I will wait for feedback for a while, thanks.

**Comment by AkioNak** [comment]
@MarcoFalke I added the followings:
1. > So, I will add `{BCLog::NONE, "none"}` to `LogCategories[]` and add to the help text that `"0"` and `"1"` are the shorthand of `"none"` and `"all"`.
2. made `"0"` and `"none"` to be  treated as same in `-debug` for`bitcoind`.
3. make `"0"` and `"none"` means ignore other specified logging category of `"include"` or `"exclude"` for this RPC like when they are given to `-debug`.

File: src/rpc/misc.cpp [hunk 2]
```diff
@@ -577,6 +577,8 @@ uint32_t getCategoryMask(UniValue cats) {
         if (!GetLogCategory(&flag, &cat)) {
             throw JSONRPCError(RPC_INVALID_PARAMETER, "unknown logging category " + cat);
         }
+        if (flag == BCLog::NONE)
+            return 0;
```
kallewoof: Style: Either use same line (`if (..) return 0;`) or use braces (
```
if (..) {
    return 0;
}
```

src/rpc/misc.cpp
```diff
@@ -586,16 +588,31 @@ UniValue logging(const JSONRPCRequest& request)
 {
     if (request.fHelp || request.params.size() > 2) {
         throw std::runtime_error(
-            "logging [include,...] <exclude>\n"
+            "logging (<include> <exclude>)\n"
```
kallewoof: The standard is to have put space after `(` and before `)` for optional args, i.e. `( <include> <exclude> )`

src/rpc/misc.cpp
```diff
@@ -586,16 +588,31 @@ UniValue logging(const JSONRPCRequest& request)
...
             "Gets and sets the logging configuration.\n"
-            "When called without an argument, returns the list of categories that are currently being debug logged.\n"
-            "When called with arguments, adds or removes categories from debug logging.\n"
+            "When called without an argument, returns the list of categories with status that are currently being debug logged or not.\n"
+            "When called with arguments, adds or removes categories from debug logging and return the list above.\n"
```
kallewoof: `and returns the list` (`s`)

src/rpc/misc.cpp
```diff
@@ -586,16 +588,31 @@ UniValue logging(const JSONRPCRequest& request)
...
+			"The argument are evaluated for \"include\" first. That is, \"exclude\" takes precedence.\n"
```
kallewoof: A bit hard to understand. Maybe: `The arguments are evaluated in order \"include\", \"exclude\". If an item is both included and excluded, it will thus end up being excluded.`

src/rpc/misc.cpp
```diff
@@ -586,16 +588,31 @@ UniValue logging(const JSONRPCRequest& request)
...
             "The valid logging categories are: " + ListLogCategories() + "\n"
-            "libevent logging is configured on startup and cannot be modified by this RPC during runtime."
-            "Arguments:\n"
-            "1. \"include\" (array of strings) add debug logging for these categories.\n"
-            "2. \"exclude\" (array of strings) remove debug logging for these categories.\n"
-            "\nResult: <categories>  (string): a list of the logging categories that are active.\n"
+            "In addition, the following are available as a category name with special meanings:\n"
```
kallewoof: `as category names`

kallewoof (changes requested):

AkioNak:
@kallewoof Thank you for your review.
I fixed that which you pointed out.

kallewoof (approved):
utACK

fanquake:
Please squash 1830d5c

src/init.cpp
```diff
@@ -925,7 +925,8 @@ bool AppInitParameterInteraction()
         // Special-case: if -debug=0/-nodebug is set, turn off debugging messages
         const std::vector<std::string> categories = gArgs.GetArgs("-debug");
 
-        if (find(categories.begin(), categories.end(), std::string("0")) == categories.end()) {
+        if (std::none_of(categories.begin(), categories.end(),
+            [](std::string cat){return cat == std::string("0") || cat == std::string("none");})) {
```
promag:
```cpp
[](const auto& cat){return cat == "0" || cat == "none";}
```

promag (commented):
utACK.

src/init.cpp
AkioNak: @promag thank you for your review.
1.  auto specifier can not use in lambda epression when -std=c++11.
2. std::string has == operater with c-string.

so I will change to following:
``` [](std::string cat){return cat == "0" || cat == "none";})) { ```

Comment (promag) [reply src/init.cpp]: BTW, this is not the 1st time I would prefer to read:
```cpp
std::set<std::string> categories = gArgs.GetUniqueArgs("-debug");

if (!categories.count("0") && !categories.count("none")) {
```
Or simply:
```cpp
if (!gArgs.IsArgSet("-debug", "0") && !gArgs.IsArgSet("-debug", "none")) {
```

AkioNak:
@promag fixed simplify the right side of the equations.
@fanquake squashed.

laanwj:
Lightly tested ACK c60c49b6794279325725a5c3c1a8d3dc6764b966

PR #11194, opened 2017-08-29
fees: optimize decay

theuni:
As discussed briefly last week with @morcos and @sdaftuar.
Rather than jumping around, visit elements in-order. This produces a noticeable speedup.

I added the assertions because it's not entirely clear from the existing code if the bucket size is allowed to differ from the individual sizes. I can't imagine that that could be the case, so I'll remove them if preferred.

Additionally, I assume we could skip UpdateMovingAverages() altogether if no entries have ever been added, but it's not obvious to me how to do that without breaking something.

### DIFF
--- a/src/policy/fees.cpp
+++ b/src/policy/fees.cpp
@@ -231,13 +231,25 @@ void TxConfirmStats::Record(int blocksToConfirm, double val)
 
 void TxConfirmStats::UpdateMovingAverages()
 {
-    for (unsigned int j = 0; j < buckets.size(); j++) {
-        for (unsigned int i = 0; i < confAvg.size(); i++)
-            confAvg[i][j] = confAvg[i][j] * decay;
-        for (unsigned int i = 0; i < failAvg.size(); i++)
-            failAvg[i][j] = failAvg[i][j] * decay;
-        avg[j] = avg[j] * decay;
-        txCtAvg[j] = txCtAvg[j] * decay;
+    for (auto& elem : confAvg) {
+        assert(elem.size() == buckets.size());
+        for (auto& bucket : elem) {
+            bucket *= decay;
+        }
+    }
+    for (auto& elem : failAvg) {
+        assert(elem.size() == buckets.size());
+        for (auto& bucket : elem) {
+            bucket *= decay;
+        }
+    }
+    assert(avg.size() == buckets.size());
+    for (auto& bucket : avg) {
+        bucket *= decay;
+    }
+    assert(txCtAvg.size() == buckets.size());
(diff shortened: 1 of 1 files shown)


gmaxwell:
> Additionally, I assume we could skip UpdateMovingAverages() altogether if no entries have ever been added, but it's not obvious to me how to do that without breaking something.

An optimization like this would be important for IBD.

theuni:
@gmaxwell Indeed. When profiling IBD for the first few hundred thousand blocks, some of my results showed this function alone accounting for ~5-10% of all cpu time spent on the message handling thread.

This change reduces those percentages to roughly half iirc, but obviously avoiding it entirely when possible would be best.

promag:
> Additionally, I assume we could skip UpdateMovingAverages() altogether if no entries have ever been added, but it's not obvious to me how to do that without breaking something.

How about not `Record()` it if current tip is too old? In other words, just `Record()` and `UpdateMovingAverages()` when approaching live tip?

gmaxwell:
FWIW, I benchmarked a sync completely bypassing CBlockPolicyEstimator::processBlock and it was effectively the same speed (0.5% faster, but that could well be noise).

promag:
You mean speed in IBD, not CPU usage right?

gmaxwell:
Yes, though on this host (and in this benchmark) IBD is more or less cpu limited.

src/policy/fees.cpp
```diff
@@ -231,13 +231,25 @@ void TxConfirmStats::Record(int blocksToConfirm, double val)
...
+            bucket *= decay;
+        }
+    }
+    assert(avg.size() == buckets.size());
+    for (auto& bucket : avg) {
+        bucket *= decay;
+    }
+    assert(txCtAvg.size() == buckets.size());
+    for (auto& bucket : txCtAvg) {
+        bucket *= decay;
```
Dllieu: It might be more clear to write

```c++
auto applyDecay = [&decay, &buckets](std::vector<double>& transactions)
{
    assert(transactions.size() == buckets.size());
    for (double& bucket : transactions)
    {
        bucket *= decay;
    }
};

for (auto& elem : confAvg)
{
    applyDecay(elem);
}

for (auto& elem : failAvg)
{
    applyDecay(elem);
}

applyDecay(avg);
applyDecay(txCtAvg);
```

TheBlueMatt (commented):
Please lets not add more asserts in non-consensus/net code. Without the asserts happy with this or @Dllieu's suggestion.

laanwj:
Last activity on this was in 2017, going to close this and add "up for grabs". Let me know if you want me to reopen it.

PR #11200, opened 2017-08-31
Allow for aborting rescans in the GUI

achow101:
A cancel button is added to the `showProgress` dialog that is used only for rescans. When clicked, `AbortRescan` is called directly to cancel the rescan.

Rescans triggered from the debug console will now be cancelable by clicking the cancel button.

Rescans triggered by a command (e.g. `importmulti`) will now give an error indicating that the rescan was aborted by the user (either by the `abortrescan` command or by clicking cancel).

### DIFF
--- a/src/interfaces/wallet.h
+++ b/src/interfaces/wallet.h
@@ -65,6 +65,9 @@ public:
     virtual bool changeWalletPassphrase(const SecureString& old_wallet_passphrase,
         const SecureString& new_wallet_passphrase) = 0;
 
+    //! Abort a rescan.
+    virtual void abortRescan() = 0;
+
     //! Back up wallet.
     virtual bool backupWallet(const std::string& filename) = 0;
 
--- a/src/wallet/rpcdump.cpp
+++ b/src/wallet/rpcdump.cpp
@@ -534,9 +534,11 @@ UniValue importwallet(const JSONRPCRequest& request)
         int64_t nFilesize = std::max((int64_t)1, (int64_t)file.tellg());
         file.seekg(0, file.beg);
 
-        pwallet->ShowProgress(_("Importing..."), 0); // show progress dialog in GUI
+        // Use uiInterface.ShowProgress instead of pwallet.ShowProgress because pwallet.ShowProgress has a cancel button tied to AbortRescan which
+        // we don't want for this progress bar shoing the import progress. uiInterface.ShowProgress does not have a cancel button.
+        uiInterface.ShowProgress(_("Importing..."), 0, false); // show progress dialog in GUI
         while (file.good()) {
-            pwallet->ShowProgress("", std::max(1, std::min(99, (int)(((double)file.tellg() / (double)nFilesize) * 100))));
+            uiInterface.ShowProgress("", std::max(1, std::min(99, (int)(((double)file.tellg() / (double)nFilesize) * 100))), false);
             std::string line;
             std::getline(file, line);
             if (line.empty() || line[0] == '#')
@@ -599,7 +601,7 @@ UniValue importwallet(const JSONRPCRequest& request)
             }
         }
         file.close();
-        pwallet->ShowProgress("", 100); // hide progress dialog in GUI
+        uiInterface.ShowProgress("", 100, false); // hide progress dialog in GUI
         pwallet->UpdateTimeFirstKey(nTimeBegin);
     }
     pwallet->RescanFromTime(nTimeBegin, reserver, false /* update */);
--- a/src/wallet/rpcdump.cpp
+++ b/src/wallet/rpcdump.cpp
@@ -172,7 +172,13 @@ UniValue importprivkey(const JSONRPCRequest& request)
         }
     }
     if (fRescan) {
-        pwallet->RescanFromTime(TIMESTAMP_MIN, reserver, true /* update */);
+        int64_t scanned_time = pwallet->RescanFromTime(TIMESTAMP_MIN, reserver, true /* update */);
+        if (pwallet->IsAbortingRescan()) {
+            throw JSONRPCError(RPC_MISC_ERROR, "Rescan aborted by user.");
+        }
+        if (scanned_time > TIMESTAMP_MIN) {
+            throw JSONRPCError(RPC_WALLET_ERROR, "Rescan was unable to fully rescan the blockchain. Some transactions may be missing.");
+        }
     }
 
     return NullUniValue;
@@ -310,7 +316,13 @@ UniValue importaddress(const JSONRPCRequest& request)
     }
     if (fRescan)
     {
-        pwallet->RescanFromTime(TIMESTAMP_MIN, reserver, true /* update */);
+        int64_t scanned_time = pwallet->RescanFromTime(TIMESTAMP_MIN, reserver, true /* update */);
+        if (pwallet->IsAbortingRescan()) {
+            throw JSONRPCError(RPC_MISC_ERROR, "Rescan aborted by user.");
+        }
+        if (scanned_time > TIMESTAMP_MIN) {
+            throw JSONRPCError(RPC_WALLET_ERROR, "Rescan was unable to fully rescan the blockchain. Some transactions may be missing.");
+        }
         pwallet->ReacceptWalletTransactions();
     }
 
@@ -479,7 +491,13 @@ UniValue importpubkey(const JSONRPCRequest& request)
     }
     if (fRescan)
     {
-        pwallet->RescanFromTime(TIMESTAMP_MIN, reserver, true /* update */);
+        int64_t scanned_time = pwallet->RescanFromTime(TIMESTAMP_MIN, reserver, true /* update */);
+        if (pwallet->IsAbortingRescan()) {
+            throw JSONRPCError(RPC_MISC_ERROR, "Rescan aborted by user.");
+        }
+        if (scanned_time > TIMESTAMP_MIN) {
+            throw JSONRPCError(RPC_WALLET_ERROR, "Rescan was unable to fully rescan the blockchain. Some transactions may be missing.");
+        }
         pwallet->ReacceptWalletTransactions();
     }
 
@@ -604,7 +622,14 @@ UniValue importwallet(const JSONRPCRequest& request)
         uiInterface.ShowProgress("", 100, false); // hide progress dialog in GUI
         pwallet->UpdateTimeFirstKey(nTimeBegin);
     }
-    pwallet->RescanFromTime(nTimeBegin, reserver, false /* update */);
+    uiInterface.ShowProgress("", 100, false); // hide progress dialog in GUI
+    int64_t scanned_time = pwallet->RescanFromTime(nTimeBegin, reserver, false /* update */);
+    if (pwallet->IsAbortingRescan()) {
+        throw JSONRPCError(RPC_MISC_ERROR, "Rescan aborted by user.");
+    }
+    if (scanned_time > nTimeBegin) {
+        throw JSONRPCError(RPC_WALLET_ERROR, "Rescan was unable to fully rescan the blockchain. Some transactions may be missing.");
+    }
     pwallet->MarkDirty();
 
     if (!fGood)
@@ -1214,6 +1239,9 @@ UniValue importmulti(const JSONRPCRequest& mainRequest)
         int64_t scannedTime = pwallet->RescanFromTime(nLowestTimestamp, reserver, true /* update */);
         pwallet->ReacceptWalletTransactions();
 
+        if (pwallet->IsAbortingRescan()) {
+            throw JSONRPCError(RPC_MISC_ERROR, "Rescan aborted by user.");
+        }
         if (scannedTime > nLowestTimestamp) {
             std::vector<UniValue> results = response.getValues();
             response.clear();
--- a/src/interfaces/wallet.cpp
+++ b/src/interfaces/wallet.cpp
@@ -132,6 +132,7 @@ public:
     {
         return m_wallet.ChangeWalletPassphrase(old_wallet_passphrase, new_wallet_passphrase);
     }
+    void abortRescan() override { m_wallet.AbortRescan(); }
     bool backupWallet(const std::string& filename) override { return m_wallet.BackupWallet(filename); }
     std::string getWalletName() override { return m_wallet.GetName(); }
     bool getKeyFromPool(bool internal, CPubKey& pub_key) override
--- a/src/qt/walletview.cpp
+++ b/src/qt/walletview.cpp
@@ -315,9 +315,9 @@ void WalletView::showProgress(const QString &title, int nProgress)
         progressDialog = new QProgressDialog(title, "", 0, 100);
         progressDialog->setWindowModality(Qt::ApplicationModal);
         progressDialog->setMinimumDuration(0);
-        progressDialog->setCancelButton(0);
         progressDialog->setAutoClose(false);
         progressDialog->setValue(0);
+        progressDialog->setCancelButtonText(tr("Cancel"));
     }
     else if (nProgress == 100)
     {
@@ -327,8 +327,13 @@ void WalletView::showProgress(const QString &title, int nProgress)
             progressDialog->deleteLater();
         }
     }
-    else if (progressDialog)
-        progressDialog->setValue(nProgress);
+    else if (progressDialog) {
+        if (progressDialog->wasCanceled()) {
+            getWalletModel()->wallet().abortRescan();
+        } else {
+            progressDialog->setValue(nProgress);
+        }
+    }
 }
 
 void WalletView::requestedSyncWarningInfo()


src/qt/bitcoingui.cpp
```diff
@@ -1119,16 +1120,21 @@ void BitcoinGUI::detectShutdown()
...
     {
         progressDialog = new QProgressDialog(title, "", 0, 100);
         progressDialog->setWindowModality(Qt::ApplicationModal);
         progressDialog->setMinimumDuration(0);
-        progressDialog->setCancelButton(0);
         progressDialog->setAutoClose(false);
         progressDialog->setValue(0);
+
+        if (cancel) {
+            progressDialog->setCancelButtonText("Cancel");
```
MeshCollider: Does this "Cancel" need a translation?

meshcollider (commented):
utACK modulo Travis issues

src/qt/bitcoingui.cpp
```diff
@@ -1138,8 +1144,14 @@ void BitcoinGUI::showProgress(const QString &title, int nProgress)
             progressDialog->deleteLater();
         }
     }
-    else if (progressDialog)
-        progressDialog->setValue(nProgress);
+    else if (progressDialog) {
+        if (progressDialog->wasCanceled()) {
+            cancel();
+        }
```
promag: `} else {`

src/qt/bitcoingui.cpp
```diff
@@ -1119,16 +1120,21 @@ void BitcoinGUI::detectShutdown()
     }
 }
 
-void BitcoinGUI::showProgress(const QString &title, int nProgress)
+void BitcoinGUI::showProgress(const QString &title, int nProgress, const std::function<void(void)> &cancel)
```
promag: `&` are next to the type. Same for remaining changes.

promag (commented):
Should it be possible to abort rescans caused by the RPC calls?

I'm not very fond of callbacks. Have you thought in alternatives?

achow101:
I don't know what is causing the travis failures.

> Should it be possible to abort rescans caused by the RPC calls?

Yes. There is an `abortrescan` RPC call that lets you abort a rescan. But it is not accessible from the GUI.

src/txdb.cpp
```diff
@@ -386,7 +386,7 @@ bool CCoinsViewDB::Upgrade() {
             if (count++ % 256 == 0) {
                 uint32_t high = 0x100 * *key.second.begin() + *(key.second.begin() + 1);
                 int percentageDone = (int)(high * 100.0 / 65536.0 + 0.5);
-                uiInterface.ShowProgress(_("Upgrading UTXO database") + "\n"+ _("(press q to shutdown and continue later)") + "\n", percentageDone);
+                uiInterface.ShowProgress(_("Upgrading UTXO database") + "\n"+ _("(press q to shutdown and continue later)") + "\n", percentageDone, std::function<void(void)>());
```
ryanofsky: In commit "Add cancel function reference to ShowProgress"

No need to write `std::function<void(void)>()` here and 12 other places. You can just pass {} or nullptr.

src/qt/bitcoingui.cpp
```diff
@@ -1139,8 +1144,14 @@ void BitcoinGUI::showProgress(const QString &title, int nProgress, const std::fu
             progressDialog->deleteLater();
         }
     }
-    else if (progressDialog)
-        progressDialog->setValue(nProgress);
+    else if (progressDialog) {
+        if (progressDialog->wasCanceled()) {
+            cancel();
```
ryanofsky: In commit "Add cancel button to rescan progress dialog"

Maybe assert(cancel)

ryanofsky (commented):
utACK e6e8b0c259594b09e405fc9123d1f8473e42b0e4 with RPC bugfixes

src/wallet/rpcdump.cpp
```diff
@@ -535,7 +556,14 @@ UniValue importwallet(const JSONRPCRequest& request)
     file.close();
     pwallet->ShowProgress("", 100, std::function<void(void)>()); // hide progress dialog in GUI
     pwallet->UpdateTimeFirstKey(nTimeBegin);
-    pwallet->RescanFromTime(nTimeBegin, false /* update */);
+    int64_t scanned_time = pwallet->RescanFromTime(nTimeBegin, false /* update */);
+    if (scanned_time > TIMESTAMP_MIN) {
```
ryanofsky: In commit "Give an error when rescan is aborted by the user"

s/TIMESTAMP_MIN/nTimeBegin/ here to fix test failures

src/txdb.cpp
achow101: Oh, I didn't realize that that would work. Done.

src/qt/bitcoingui.cpp
achow101: Done

src/qt/bitcoingui.cpp
achow101: Done

src/wallet/rpcdump.cpp
achow101: Done

src/qt/bitcoingui.cpp
achow101: Yes, it does. Done

ryanofsky (commented):
utACK c8720f0a320eff0cc83a459dc87f28087d518825. Only changes were implementing various suggestions.

> I'm not very fond of callbacks. Have you thought in alternatives?

Reason callback seems to be used is to avoid the need for qt code to call CWallet's AbortRescan method or set the fAbortRescan flag directly. The alternatives would be for Qt to call a different function or set a different flag, but don't seem like they would be much better or worse.

src/wallet/wallet.cpp
```diff
@@ -1568,13 +1568,13 @@ CBlockIndex* CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, bool f
         fAbortRescan = false;
         fScanningWallet = true;
 
-        ShowProgress(_("Rescanning..."), 0); // show rescan progress in GUI as dialog or on splashscreen, if -rescan on startup
+        ShowProgress(_("Rescanning..."), 0, boost::bind(&CWallet::AbortRescan, this)); // show rescan progress in GUI as dialog or on splashscreen, if -rescan on startup
```
jonasschnelli: nit: use `std::bind` (and below)

src/ui_interface.h
```diff
@@ -95,7 +95,7 @@ class CClientUIInterface
     boost::signals2::signal<void (CWallet* wallet)> LoadWallet;
 
     /** Show progress e.g. for verifychain */
-    boost::signals2::signal<void (const std::string &title, int nProgress)> ShowProgress;
+    boost::signals2::signal<void (const std::string &title, int nProgress, const std::function<void(void)>& cancel)> ShowProgress;
```
jonasschnelli: use `nullptr` as default parameter for `cancel` to avoid adding `nullptr` to all non-callback `ShowProgress` calls?

jonasschnelli (commented):
Concept ACK

src/ui_interface.h
achow101: I couldn't set a default parameter in the signal, so I'm not sure it is possible to do that. Or I'm just doing something wrong.

src/wallet/wallet.cpp
achow101: Done

jonasschnelli:
Tested a bit and there is a concurrency issue (maybe not related to this code but since you finally can test multiple rescans during the same runtime this PR may reveal the issue):

Sometimes the progress window does not show up (happened to me only if I canceled a rescan before) and the GUI is fully in locked state (due to the rescan).
In that case, the rescan is happening and the GUI is locked (without showing a progress screen) via `cs_main` in `transactiontablemodel.cpp`

jonasschnelli:
Needs rebase

src/qt/bitcoingui.cpp
```diff
@@ -487,7 +487,7 @@ void BitcoinGUI::setClientModel(ClientModel *_clientModel)
         connect(_clientModel, SIGNAL(message(QString,QString,unsigned int)), this, SLOT(message(QString,QString,unsigned int)));
 
         // Show progress dialog
-        connect(_clientModel, SIGNAL(showProgress(QString,int)), this, SLOT(showProgress(QString,int)));
+        connect(_clientModel, SIGNAL(showProgress(QString,int, bool, std::function<void(void)>)), this, SLOT(showProgress(QString,int, std::function<void(void)>)));
```
ryanofsky: Could just write `void()` instead of `void(void)` everywhere

src/qt/splashscreen.cpp
```diff
@@ -181,7 +181,7 @@ static void ShowProgress(SplashScreen *splash, const std::string &title, int nPr
 #ifdef ENABLE_WALLET
 void SplashScreen::ConnectWallet(CWallet* wallet)
 {
-    wallet->ShowProgress.connect(boost::bind(ShowProgress, this, _1, _2, false));
+    wallet->ShowProgress.connect(boost::bind(ShowProgress, this, _1, _2, false, _4));
```
ryanofsky: You added both `resume_possible` and `cancel` arguments to the ShowProgress signal in wallet.h so this should be `_3, _4`, instead of `false, _4`.

(Another alternative if you want to ensure wallet resume_possible value is always false, is to not add resume_possible in wallet.h, and then change this to `false, _3`. But adding the argument and then silently ignoring like this does now isn't good.)

ryanofsky (commented):
utACK 4657e40f1516d506095f969dc726da6f97d57a99. Only change since last review is rebase and some  whitespace diffs.

promag:
utACK 4657e40. Have you thought adding tests for the new RPC errors? (might be impossible?)

jonasschnelli:
Reminder: this PR currently causes a deadlock (at least on OSX):
IMO it would require https://github.com/bitcoin/bitcoin/pull/11281 to work flawless.

ryanofsky:
@jonasschnelli, I don't see how this could be causing any deadlock that wasn't occurring before. The cancel callback isn't acquiring any locks, just updating an atomic\<bool>. Are you sure there is a problem in the current version of this PR?

achow101:
@ryanofsky I think it is more that a deadlock currently exists in the rescanning code which this PR makes more obvious and easier to hit.

src/qt/bitcoingui.cpp
achow101: Done

src/qt/splashscreen.cpp
achow101: Done

achow101: Actually changing this seems to be causing a segfault, so I will leave it as is.

ryanofsky:
> @ryanofsky I think it is more that a deadlock currently exists in the rescanning code which this PR makes more obvious and easier to hit.

I know that's the claim, but I don't understand what the deadlock is. What does the rescan code have to do with transactiontablemodel and why would shorter aborted scans cause more UI problems than longer full scans?

ryanofsky (commented):
utACK 449dd52b417d2a49b4de4440b391f6ebccb093b4. Same as before, just gets rid of spurious whitespace changes and void(void)

promag:
Please see an alternative concept in #11826.

achow101:
Rebased

achow101:
Review/Merge beg

promag:
@achow101 needs rebase.

achow101:
Rebased

src/qt/bitcoingui.h
```diff
@@ -232,8 +234,8 @@ private Q_SLOTS:
     void detectShutdown();
 
     /** Show progress dialog e.g. for verifychain */
-    void showProgress(const QString &title, int nProgress);
-    
+    void showProgress(const QString &title, int nProgress, bool resume_possible, const std::function<void()>& cancel);
```
luke-jr: Would be nice to have nullptr as a default for cancel.

src/wallet/rpcdump.cpp
```diff
@@ -608,10 +628,18 @@ UniValue importwallet(const JSONRPCRequest& request)
...
         file.close();
-        pwallet->ShowProgress("", 100); // hide progress dialog in GUI
+        pwallet->ShowProgress("", 100, false, nullptr); // hide progress dialog in GUI
         pwallet->UpdateTimeFirstKey(nTimeBegin);
     }
-    pwallet->RescanFromTime(nTimeBegin, reserver, false /* update */);
+    pwallet->ShowProgress("", 100, false, nullptr); // hide progress dialog in GUI
+    int64_t scanned_time = pwallet->RescanFromTime(nTimeBegin, reserver, false /* update */);
+    if (scanned_time > nTimeBegin) {
+        if (pwallet->IsAbortingRescan()) {
```
luke-jr: This is ugly, since RescanFromTime changes without a signature change. Better to have it throw an exception to be caught from callers, when cancelled.

achow101: I don't quite understand what you mean.

src/qt/bitcoingui.h
achow101: This was suggested before and I was not able to get it to work.

ryanofsky:
utACK 7e53d957709c0bac88cabfc0f9370b33888504db. No changes since last review in Qt code. Just conflict resolution in other parts of the code.

achow101:
Review beg for reviews from people other than @ryanofsky

src/wallet/rpcdump.cpp
sipa: @luke-jr In what way does `RescanFromTime` change?

MarcoFalke:
utACK, but calling `rescanblockchain` in the GUI console no longer gives me the tiny "Rescanning ..." pop up, which prevents me from testing. Any idea?

Pop up on master:

screenshot from 2018-03-18 09-10-58

achow101:
That's strange. I'm seeing that now too, but I don't know why.

achow101:
Ok, I fixed that problem and also had to rebase.

src/qt/splashscreen.cpp
MarcoFalke: The following diff appears to be working for me. Could you help me find the steps to reproduce the segfault? Otherwise, I think this is a sensible suggestion.

```diff
--- a/src/qt/splashscreen.cpp
+++ b/src/qt/splashscreen.cpp
@@ -181,7 +181,7 @@ static void ShowProgress(SplashScreen *splash, const std::string &title, int nPr
 #ifdef ENABLE_WALLET
 void SplashScreen::ConnectWallet(CWallet* wallet)
 {
-    wallet->ShowProgress.connect(boost::bind(ShowProgress, this, _1, _2, false, _4));
+    wallet->ShowProgress.connect(boost::bind(ShowProgress, this, _1, _2, _3, _4));
     connectedWallets.push_back(wallet);
 }
 #endif
@@ -203,7 +203,7 @@ void SplashScreen::unsubscribeFromCoreSignals()
     uiInterface.ShowProgress.disconnect(boost::bind(ShowProgress, this, _1, _2, _3, _4));
 #ifdef ENABLE_WALLET
     for (CWallet* const & pwallet : connectedWallets) {
-        pwallet->ShowProgress.disconnect(boost::bind(ShowProgress, this, _1, _2, false, _4));
+        pwallet->ShowProgress.disconnect(boost::bind(ShowProgress, this, _1, _2, _3, _4));
     }
 #endif
 }

File: src/wallet/rpcdump.cpp (hunk 12 above)
Comment (MarcoFalke) [reply src/wallet/rpcdump.cpp]: `RescanFromTime` uses `ScanForWalletTransactions` internally. And the interface of `ScanForWalletTransactions` is changed in this pull request.

File: src/wallet/wallet.cpp [hunk 13]
```diff
@@ -1773,6 +1773,7 @@ CBlockIndex* CWallet::ScanForWalletTransactions(CBlockIndex* pindexStart, CBlock
         }
         if (pindex && fAbortRescan) {
             LogPrintf("Rescan aborted at block %d. Progress=%f\n", pindex->nHeight, gvp);
+            ret = pindex;
```
MarcoFalke: Please revert this change. This breaks other rpcs that use `ScanForWalletTransactions` directly

src/wallet/rpcdump.cpp
```diff
@@ -479,7 +493,14 @@ UniValue importpubkey(const JSONRPCRequest& request)
     }
     if (fRescan)
     {
-        pwallet->RescanFromTime(TIMESTAMP_MIN, reserver, true /* update */);
+        int64_t scanned_time = pwallet->RescanFromTime(TIMESTAMP_MIN, reserver, true /* update */);
+        if (scanned_time > TIMESTAMP_MIN) {
+            if (pwallet->IsAbortingRescan()) {
```
MarcoFalke: Why is this check within the scope of the `(scanned_time > TIMESTAMP_MIN)` check? Wouldn't the following make more sense:

```cpp
if (pwallet->IsAbortingRescan()) throw JSONRPCError(RPC_MISC_ERROR, "Rescan aborted by user.");
if (scanned_time > TIMESTAMP_MIN) throw JSONRPCError(RPC_WALLET_ERROR, "Rescan was unable to fully rescan the blockchain. Some transactions may be missing.");

**Comment by MarcoFalke** [comment]
utACK c18cfb8fe0e5de02471011bef0e37f0a6d5920e6, but needs feedback addressed.

File: src/wallet/wallet.cpp (hunk 13 above)
Comment (achow101) [reply src/wallet/wallet.cpp]: Done

File: src/wallet/rpcdump.cpp (hunk 14 above)
Comment (achow101) [reply src/wallet/rpcdump.cpp]: Yes, done.

File: src/wallet/rpcdump.cpp (hunk 12 above)
Comment (achow101) [reply src/wallet/rpcdump.cpp]: That change was reverted.

File: src/qt/splashscreen.cpp (hunk 10 above)
Comment (achow101) [reply src/qt/splashscreen.cpp]: Hmm, that diff works for me. I don't remember what I did that caused a segfault. Added this change.

**Comment by MarcoFalke** [comment]
re-utACK 1de7f88f4cad0f59a6684e179177707c8efd7b3a (only changes were to address the feedback)

**Review by ryanofsky — COMMENTED** [review]
utACK 1de7f88f4cad0f59a6684e179177707c8efd7b3a. Changes since last review were rebase, checking for abort condition even if RescanFromTime succeeds, and forwarding `resume_possible` argument to `showProgress`.

**Comment by MarcoFalke** [comment]
@jonasschnelli You tested this a bit last year. Mind to re-test?

**Comment by TheBlueMatt** [comment]
This effectively reverts #10770. Please, no.

**Comment by achow101** [comment]
@TheBlueMatt I don't see any other way to implement such cancellations in the GUI. Perhaps you could suggest something?

**Comment by TheBlueMatt** [comment]
We cancel just fine during init already (using StartShutdown()). If we want to *also* support cancelling via a rescan progress dialog, the rescan progress dialog should know how to make that happen. I still really don't think we need a generic interface here - we only use the dialog for one thing (and I'm not aware of any other cancelable proposed things to show there) - and making it common between init and rescan still makes no sense to me - they are very, very different things.

**Comment by ryanofsky** [comment]
@TheBlueMatt , your objection seems the same as the "I'm not very fond of callbacks" criticism which came up earlier: https://github.com/bitcoin/bitcoin/pull/11200#pullrequestreview-59957173. I didn't think this was a big deal (or really a very substantive criticism), but it would be pretty easy to address by replacing the callback with an wallet `AbortRescans()` function that Qt code could call directly. Would this make everybody happy?

**Comment by TheBlueMatt** [comment]
Its a bit more broad, I'm a bit more than "not fond of" passing callbacks from random internal places all the way to GUI, plus we started with that and deliberately got rid of them. Adding them *back* would need more justification than "well, this is a way that it can be done".

**Comment by ryanofsky** [comment]
I think support cancelling rescans is a worthwhile feature, and there are lots of ways it can be implemented. Which way to choose is mostly a question of how much much existing code you want to rewrite.

Having qt code just call an `AbortRescan` function when the cancel button is pressed is probably the simplest possible approach. Having the qt code call a generic `cancel` callback is a little more generic, and maybe a little cleaner, but does involve callbacks. Changing the progress dialog to be aware of the differences between init and rescan would be another way to approach this, and probably a good idea, but just a larger change.

**Comment by achow101** [comment]
I changed the implementation to completely remove the callback. The `showProgress` dialog that is used for rescans just calls `AbortRescan` directly when cancel is clicked. This should work since the dialog is used only for rescanning.

**Comment by achow101** [comment]
Rebased

**Comment by MarcoFalke** [comment]
Needs trivial rebase to accommodate a file rename.

**Comment by achow101** [comment]
Rebased

File: src/wallet/rpcdump.cpp [hunk 15]
```diff
@@ -172,7 +172,13 @@ UniValue importprivkey(const JSONRPCRequest& request)
         }
     }
     if (fRescan) {
-        pwallet->RescanFromTime(TIMESTAMP_MIN, reserver, true /* update */);
+        int64_t scanned_time = pwallet->RescanFromTime(TIMESTAMP_MIN, reserver, true /* update */);
```
promag: This code is repeated 4 times, maybe it's worth moving it to a function?

src/interfaces/wallet.h
```diff
@@ -247,6 +247,9 @@ class Wallet
     //! Register handler for watchonly changed messages.
     using WatchOnlyChangedFn = std::function<void(bool have_watch_only)>;
     virtual std::unique_ptr<Handler> handleWatchOnlyChanged(WatchOnlyChangedFn fn) = 0;
+
+    //! Abort a rescan
```
promag: Nit, period.

src/qt/walletmodel.cpp
```diff
@@ -558,3 +558,8 @@ bool WalletModel::isMultiwallet()
 {
     return m_node.getWallets().size() > 1;
 }
+
+void WalletModel::AbortRescan() const
+{
+    wallet().AbortRescan();
```
promag: Nit, `m_wallet->AbortRescan()`?

src/qt/walletmodel.h
```diff
@@ -204,6 +204,8 @@ class WalletModel : public QObject
     QString getWalletName() const;
 
     bool isMultiwallet();
+
+    void AbortRescan() const;
```
promag: Nit, `void abortRescan() const`.

promag (commented):
Some comments, will test later.

promag:
> Changing the progress dialog to be aware of the differences between init and rescan would be another way to approach this, and probably a good idea, but just a larger change.

I agree with @ryanofsky here. I'll try to push #11826 forward.

src/interfaces/wallet.h
ryanofsky: Would suggest renaming this method `abortRescan` and moving it above `backupWallet` (both here and in `wallet.cpp`). These interfaces are organized with regular methods on top and callback methods below. Also roughly with whole-wallet operations above operations on individual addresses and transactions.

jonasschnelli:
Could finally test this.
I think it is an important feature because the way how users currently abort rescan is by force quitting the application (which is really not what we want).

Tested ACK bc2003cdf7102e03710dbc19e9c397b92e124fb4.
I'd like to see @promag's nits addresses (lowercase methodname in QT, interface call via class member).

src/qt/walletmodel.cpp
ryanofsky: In commit "Add cancel button to rescan progress dialog"

Would suggest dropping this method and just calling model.wallet().abortRescan() directly. I don't think there's a benefit to wrapping individual `Wallet` calls inside a `WalletModel` methods unless they're doing extra work like synchronization or caching.

src/wallet/rpcdump.cpp
ryanofsky: In commit "Give an error when rescan is aborted by the user"

> This code is repeated 4 times, maybe it's worth moving it to a function?

Could be nice, though to be honest a lot of other code is duplicated between the import* RPCs too. So effort might be better spent deduping larger chunks or deprecating older RPCs in favor of importmulti.

ryanofsky (commented):
utACK bc2003cdf7102e03710dbc19e9c397b92e124fb4. This PR has changed quite a lot since it was opened but is very simple at this point and provides useful functionality.

src/wallet/rpcdump.cpp
achow101: I think that should be part of a separate refactor for deduplicating all of the import* RPCs.

src/interfaces/wallet.h
achow101: Done (moved and added period).

src/qt/walletmodel.cpp
achow101: Removed and called directly from walletview

src/qt/walletmodel.h
achow101: Removed

src/wallet/rpcdump.cpp
```diff
@@ -534,9 +534,9 @@ UniValue importwallet(const JSONRPCRequest& request)
         int64_t nFilesize = std::max((int64_t)1, (int64_t)file.tellg());
         file.seekg(0, file.beg);
 
-        pwallet->ShowProgress(_("Importing..."), 0); // show progress dialog in GUI
+        uiInterface.ShowProgress(_("Importing..."), 0, false); // show progress dialog in GUI
```
ryanofsky: In commit "Add cancel button to rescan progress dialog"

These changes from wallet to global `ShowProgress` calls seem unrelated. Are they intentional? It might make sense to change these if removing the wallet `ShowProgress` signal, but since that is still used in `CWallet::ScanForWalletTransactions`, this change just makes these calls inconsistent.

If this change is necessary it would be good to have a comment saying why CWallet::ShowProgress is not used here.

ryanofsky (commented):
utACK 66ee39f227b28a32c6c1cb6cd68c0eee0cc8637f

src/wallet/rpcdump.cpp
achow101: Because `CWallet::ShowProgress` has a cancel button now which is directly tied to `AbortRescan` (because it isn't used for anything except showing rescan progress), I needed a different progress bar which does not have a cancel button. This change makes that `ShowProgress` use `BitcoinGUI::ShowProgress` (which is tied toe `uiInterface`)which does not have a cancel button (and is also not used anywhere else AFAICT). This progress bar indicates the import progress, not rescan progress which is why a different `ShowProgress` is needed

ryanofsky (commented):
utACK deaffd812e3cdaeb7ec978c2d86f4bf688add96f, thanks for the comment!

achow101:
Squashed the squashme. Is this ready for merging?

MarcoFalke:
utACK ae1d2b0308d7fe9df7fc18699c69e9587e6154bd. Going to test this later...

src/wallet/rpcdump.cpp
promag: Sure.

PR #7552, opened 2016-02-17
rpc: Add `verifyrawtransactions` call

laanwj:
Add a RPC call to verify a set of raw transactions without propagating them. It has an extensible API to customize verification options.

Implements #4162.

```
verifyrawtransactions ["hexstring",...] ( options )

Verifies one or more raw transactions (serialized, hex-encoded). If transactions depend on each other, they must be provided in order.

Arguments:
1. ["hexstring",...] (array of strings, required) The hex string of the raw transactions)
2. options   (json object, optional)
     {
       "include_mempool"          (boolean, optional, default=true) Whether to include the mem pool
       "check_final"              (boolean, optional, default=true) Check that the transactions will be final by next block
       "check_standard"           (boolean, optional, default=true) Perform transaction standard checks
     }

Result:
null if the verification was successful, otherwise an error object:
{
  "index":n,                (numeric) Index in transactions array of failed transaction
  "hash":"hex",             (string) Transaction hash of failed transaction
  "code": n,                (numeric) Reject code
  "reason": "text"          (string) Reject reason
  "debug_message": "text"   (string) Reject debug message
}
```

TODO:
- ~~RPC test~~
-  ~~Call  `AreInputsStandard()` to check inputs for standardness~~

PR #11201, opened 2017-08-31
[WIP] [RPC] Add verifyrawtransaction RPC

justicz:
Currently, there is not an RPC that allows for the validation of a raw transaction without actually submitting that transaction to the network. This PR implements `verifyrawtransaction`, which accepts a hex encoded raw transaction and determines if it could be included in the next block. It does this by creating a "fake" block with the proposed transaction, and then checking that the block is valid with `TestBlockValidity`.

This is my first attempt at contributing to bitcoin, so if I've already managed to break any unspoken (or spoken) rules, please let me know 😅

### DIFF
--- a/src/rpc/rawtransaction.cpp
+++ b/src/rpc/rawtransaction.cpp
@@ -948,11 +948,88 @@ UniValue verifyrawtransactions(const JSONRPCRequest& request)
         transactions.insert(std::pair<uint256, CTransactionRef>(ref->GetHash(), ref));
     }
 
+    // Find the ancestors of all of the transactions, including ones in mempool
+    std::map<uint256, CTransactionRef> ancestors;
+    for (auto const& hash_tx : transactions) {
+        CTransactionRef tx = hash_tx.second;
+        for (const CTxIn& txin : tx->vin) {
+            // Check the mempool
+            const uint256& hash = txin.prevout.hash;
+            CTxMemPool::txiter it = mempool.mapTx.find(hash);
+
+            // Ancestor not in mempool
+            if (it == mempool.mapTx.end()) {
+                continue;
+            }
+
+            // Add the ancestor to the map
+            ancestors.insert(std::pair<uint256, CTransactionRef>(hash, MakeTransactionRef(it->GetTx())));
+
+            // CalculateMemPoolAncestors only works on mempool entries, which is why we are iterating over vin
+            CTxMemPool::setEntries tx_ancestors;
+            uint64_t no_limit = std::numeric_limits<uint64_t>::max();
+            std::string dummy;
+            mempool.CalculateMemPoolAncestors(*it, tx_ancestors, no_limit, no_limit, no_limit, no_limit, dummy, false);
+
+            // Insert ancestors' ancestors into the map
+            for (CTxMemPool::txiter ancestor_it : tx_ancestors) {
+                CTransactionRef atf = MakeTransactionRef(ancestor_it->GetTx());
+                ancestors.insert(std::pair<uint256, CTransactionRef>(atf->GetHash(), atf));
+            }
+        }
+    }
+
+    // Merge the ancestors into the transactions
+    transactions.insert(ancestors.begin(), ancestors.end());
+
+    // Now, come up with a valid topological sort over the transactions
+    std::vector<CTransactionRef> ordered_tx;
+    std::vector<CTransactionRef> block_tx;
+
+    // Iterate over the transactions, removing all of those with no
+    // dependencies in the list each time. Technically, this is worst
+    // case n^2 since the transactions could be dependent in reverse
+    // order. In practice, I think this is better than e.g. implementing
+    // an asymptotically efficient topological sort since input sizes will
+    // likely be small and this is way easier to implement.
+    while (transactions.size() > 0) {
+        auto hash_tx = transactions.begin();
+        bool found_independent = false;
+        while (hash_tx != transactions.end()) {
+            CTransactionRef tx = hash_tx->second;
+            bool is_independent = true;
+            for (const CTxIn& txin : tx->vin) {
+                txit = transactions.find(txin.prevout.hash);
+                if (txit != transactions.end()) {
+                    is_independent = false;
+                    break;
+                }
+            }
+            if (is_independent) {
+                found_independent = true;
+                // Don't insert mempool ancestors
+                if (ancestors.find(hash_tx->first) == ancestors.end()) {
+                    ordered_tx.push_back(tx);
+                }
+                block_tx.push_back(tx);
+                hash_tx = transactions.erase(hash_tx);
+            } else {
+                hash_tx++;
+            }
+        }
+        if (found_independent) {
+            continue;
+        }
+        // We got all the way through without finding an independent tx
+        result.push_back(Pair("valid",  false));
+        result.push_back(Pair("reason", "circular transaction dependency"));
+        return result;
+    }
+
     // Should we check against our local policy, or just consensus rules?
     if (use_local_policy) {
         // For each input transaction
-        for (auto const& hash_tx : transactions) {
-            CTransactionRef tx(hash_tx.second);
+        for (auto const& tx : ordered_tx) {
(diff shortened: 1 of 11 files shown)


gmaxwell:
Well there is already a recommend way to do this: use a block proposal.

The general limitation with a 'verifyrawtransaction' approach is that it doesn't let you test a graph of unconfirmed transactions, but a block proposal does.

Looks like internally your rpc is doing more or less what a block proposal does, but handling more of the things around it. So perhaps it would be useful to do.

It could obviously be extended to accept an array of transactions to do the right thing with respect to multiple... but another open question is what about parents in the mempool. Right now your code will fail a transaction that spends the output of a mempool transaction.  Sometimes that would be desirable but I think often not.

luke-jr:
Seems awfully inefficient to call CNB just to throw away the chosen transaction-set. Also, the block fees will be wrong. If you don't want to build a block from scratch, maybe abstract that out of CNB first (separate commit, same PR)?

(To revise your PR, just use `git push --force` over the same branch.)

justicz:
Got it, thanks for the feedback. I'll address these issues now

sdaftuar:
I agree with the general usefulness of a `verifyrawtransaction` RPC (especially one that takes a list of unconfirmed transactions, and ideally one that allows for checking based on inputs that may be in the mempool), but one issue with this block proposal approach is that you won't be testing against local policy, just the consensus rules.  So be aware that a True result here won't necessarily mean that the node would accept the transactions and relay them.

achow101:
Concept ACK, but as others have said, there are other things to consider when verifying a transaction that are not done here. Those need to be documented.

jtimon:
Seems related to https://github.com/bitcoin/bitcoin/pull/7552

justicz:
Thanks much for that link @jtimon, somehow I missed that.

It seems like everyone agrees that there is the potential for some useful functionality here, there's just some uncertainty in what exactly the API for this should look like. After reading https://github.com/bitcoin/bitcoin/pull/7552, it seems like this uncertainty is largely what has kept `verifyrawtransaction` functionality from being included in the past.

What if `verifyrawtransaction` became `verifyrawtransactions` as in https://github.com/bitcoin/bitcoin/pull/7552, and took an argument as follows:

```
{
    "transactions": [ "hex_tx0", "hex_tx1", ... ],
    "use_local_policy": true // default true
}
```

I think this achieves all of the behavior we want; users can set "use_local_policy" to false if they only want to check consensus rules, and the default behavior handles ancestors in the mempool correctly as brought up by @gmaxwell. I think I can do this with a minor refactor of `AcceptToMemoryPoolWorker`?

Does this sound good?

justicz:
Alright, I went ahead and did my best to implement what I described in my previous post.

dcousens:
Can't `signrawtransaction` be used to verify a transaction?

src/rpc/rawtransaction.cpp
```diff
@@ -887,6 +889,231 @@ UniValue signrawtransaction(const JSONRPCRequest& request)
...
+            // Call AcceptToMemoryPool with the dryrun flag set
+            if (!AcceptToMemoryPool(mempool, state, std::move(tx), true, &missing_inputs, nullptr, false, true, 0)) {
+                if (state.IsInvalid()) {
+                    result.push_back(Pair("valid",  false));
+                    result.push_back(Pair("reason", strprintf("%i: %s", state.GetRejectCode(), state.GetRejectReason())));
+                    return result;
+                } else if (missing_inputs) {
+                    // We want to ignore the "missing inputs" case here, since the input
+                    // may actually be in the mempool. If the input is truly missing, we
+                    // will find out later after the ancestor and consensus check.
```
sdaftuar: If ATMP fails for a transaction due to missing inputs, then almost none of the policy checks will be evaluated for the transaction, because that's one of the first things we (must) do -- for instance you can't check the feerate of the transaction without having the inputs available.

sdaftuar: Actually on further review, it looks like you've added these "dry run" transactions to the mempool, so now I think if ATMP returns "missing inputs", that should reflect an actual failure?  Is there a case where that doesn't hold?

sdaftuar:
@justicz Thanks for working on this.  I think there's still work to be done here, but this seems like a promising direction.

Rather than add the `dryRunMap` to the mempool, though, I think it'd be cleaner if we instead just added a caching layer in front of the mempool, which holds these dry run transactions.  ATMPWorker already takes a mempool as an argument, so I think it'd be straightforward to abstract the mempool interface away (and I guess clean up the mempool interface so that callers aren't grabbing at internals), and then pass in a cache that holds dry run transactions in it, as well as transactions that get evicted from the mempool (eg due to RBF).  Then I think we could ensure that the mempool itself is a const object in this context that is clearly unaffected, with all changes happening only in the cache.

If we can achieve that, that would make the review and maintenance burden of this feature pretty minimal, which I think would be awesome.

src/txmempool.cpp
```diff
@@ -357,13 +357,19 @@ void CTxMemPool::AddTransactionsUpdated(unsigned int n)
...
 
-bool CTxMemPool::addUnchecked(const uint256& hash, const CTxMemPoolEntry &entry, setEntries &setAncestors, bool validFeeEstimate)
+bool CTxMemPool::addUnchecked(const uint256& hash, const CTxMemPoolEntry &entry, setEntries &setAncestors, bool validFeeEstimate, bool dryRun)
 {
     NotifyEntryAdded(entry.GetSharedTx());
     // Add to memory pool without checking anything.
     // Used by AcceptToMemoryPool(), which DOES do
     // all the appropriate checks.
     LOCK(cs);
+
```
sdaftuar: The `NotifyEntryAdded` callback a few lines up shouldn't trigger for dry run transactions.

src/validation.cpp
```diff
@@ -839,48 +839,59 @@ static bool AcceptToMemoryPoolWorker(const CChainParams& chainparams, CTxMemPool
...
-        {
-            LogPrint(BCLog::MEMPOOL, "replacing tx %s with %s for %s BTC additional fees, %d delta bytes\n",
-                    it->GetTx().GetHash().ToString(),
-                    hash.ToString(),
-                    FormatMoney(nModifiedFees - nConflictingFees),
-                    (int)nSize - (int)nConflictingSize);
-            if (plTxnReplaced)
-                plTxnReplaced->push_back(it->GetSharedTx());
+        if (!fDryRun) {
+            // Remove conflicting transactions from the mempool
```
sdaftuar: Since you only are removing conflicts in non-dryrun situations, doesn't that mean that someone calling verifyrawtransactions with a pair of tx's, one of which would replace existing transactions, and one of which depends on an existing transaction, would potentially succeed?  Yet when actually trying to send, the second would necessarily fail.

justicz:
Thanks for the comments @sdaftuar! I think I agree with all of the points you made and will work on separating things out of the mempool more cleanly.

src/rpc/rawtransaction.cpp
justicz: Yeah -- this code is leftover from before I figured out the issue you mentioned in your first comment. I was working on getting rid of this when you left your review 😝

src/validation.cpp
justicz: I think you're right. I'll fix this

src/txmempool.cpp
justicz: 👍

justicz:
Hey @sdaftuar, it's not done yet (still need to implement eviction/properly count descendants) but I've been working on implementing the feature in the way you described over here: https://github.com/justicz/bitcoin/pull/3 . I broke the relevant parts of the CTxMemPool interface out into a CTxMemPoolBase class and then implemented a CTxMockMemPool that I pass into ATMP.

Is this more what you had in mind? I'm modifying some of the const CTxMemPool methods to accept a cache so that I can avoid duplicating a lot of the logic.

justicz:
Closing this PR, will make a new one in the future if I can come up with an implementation I am happy with

wtogami:
Please ping when you do come up with a better implementation. I have strong interest in this!

PR #11205, opened 2017-08-31
Make fixed CAmounts and related sanity function constexpr

danra:
This would allow the fixed amounts and sanity function to be used in constexpr contexts.

### DIFF
--- a/src/amount.h
+++ b/src/amount.h
@@ -11,8 +11,8 @@
 /** Amount in satoshis (Can be negative) */
 typedef int64_t CAmount;
 
-static const CAmount COIN = 100000000;
-static const CAmount CENT = 1000000;
+constexpr CAmount COIN = 100000000;
+constexpr CAmount CENT = 1000000;
 
 /** No amount larger than this (in satoshi) is valid.
  *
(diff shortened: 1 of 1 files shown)


MarcoFalke:
If we'd really wanted to do that, a project wide diff would be appropriate.

danra:
@MarcoFalke I don't mind doing this project-wide where appropriate, but I'd appreciate pointers to the files where it would be relevant. I believe it would make sense to constexpr-ize conceptual things which it would make sense to test or express more complex concepts at compile time.

src/amount.h
```diff
@@ -23,7 +23,7 @@ static const CAmount CENT = 1000000;
  * critical; in unusual circumstances like a(nother) overflow bug that allowed
  * for the creation of coins out of thin air modification could lead to a fork.
  * */
-static const CAmount MAX_MONEY = 21000000 * COIN;
-inline bool MoneyRange(const CAmount& nValue) { return (nValue >= 0 && nValue <= MAX_MONEY); }
+constexpr CAmount MAX_MONEY = 21000000 * COIN;
+constexpr bool MoneyRange(const CAmount& nValue) { return (nValue >= 0 && nValue <= MAX_MONEY); }
```
promag: Rename `nValue` with `amount` and remove `()`?

promag (commented):
IMO either mention `amount.h` or do what @MarcoFalke said.

laanwj:
No problem with this, though there seems to be no specific reason for doing this until we need those values in a constexpr context.

MarcoFalke:
Agree with @laanwj. This is obviously an improvement, but I fail to see the rationale to do this refactoring only in this specific file. This will just cause a bunch of "Make constexpr" commits over the next couple of months.

Either do a scripted diff for the whole project if people agree to do that or leave it until there is an actual use case.

laanwj:
Closing this, as there is no agreement to do this.

PR #11206, opened 2017-08-31
Update optionsdialog.ui

polyetilen:
Typo.


promag:
Since it's you first contribution:
 - the tittle and commit could say something like `Fix mnemonic of hide tray icon checkbox in options dialog`;
 - being such a simple patch, the PR description can be empty.

fanquake:
This is a duplicate of/included in #10911.

Issue #11171, opened 2017-08-27
RC2 Exits After Initialization


xyzzy099:
I installed RC2 on my Windows7 box, and it exits shortly after initialization consistently.  The debug log contents look OK, with the only errors contained in the snippet below:
---

2017-08-27 14:00:58 GUI: PaymentServer::LoadRootCAs: Loaded  56  root certifica
es
2017-08-27 14:00:58 tor: Error connecting to Tor control socket
2017-08-27 14:00:58 tor: Not connected to Tor control port 127.0.0.1:9051, tryi
g to reconnect
2017-08-27 14:00:58 GUI: TransactionTablePriv::refreshWallet
2017-08-27 14:00:59 trying connection 79.195.178.192:8333 lastseen=366.6hrs
2017-08-27 14:00:59 Added connection peer=0
2017-08-27 14:01:00 tor: Error connecting to Tor control socket
2017-08-27 14:01:00 tor: Not connected to Tor control port 127.0.0.1:9051, tryi
g to reconnect
2017-08-27 14:01:03 tor: Error connecting to Tor control socket
2017-08-27 14:01:03 tor: Not connected to Tor control port 127.0.0.1:9051, tryi
g to reconnect
2017-08-27 14:01:04 connection to 79.195.178.192:8333 timeout
2017-08-27 14:01:04 trying connection 81.163.206.173:8333 lastseen=175.4hrs
2017-08-27 14:01:04 connection from 213.5.31.38:54224 accepted
2017-08-27 14:01:04 Added connection peer=1
2017-08-27 14:01:04 connection from 52.192.180.114:40814 accepted
2017-08-27 14:01:04 connection from 54.191.174.46:38180 dropped (banned)
2017-08-27 14:01:04 received: version (106 bytes) peer=1
2017-08-27 14:01:04 sending version (102 bytes) peer=1
2017-08-27 14:01:04 send version message: version 70015, blocks=482210, us=[::]
0, peer=1
2017-08-27 14:01:04 sending verack (0 bytes) peer=1
2017-08-27 14:01:04 receive version message: /bitcoin-seeder:0.01/: version 600
0, blocks=350000, us=73.119.168.246:8333, peer=1
2017-08-27 14:01:04 added time data, samples 2, offset -209 (-3 minutes)
2017-08-27 14:01:04 sending alert (168 bytes) peer=1
2017-08-27 14:01:05 received: verack (0 bytes) peer=1
[...]

PR #11208, opened 2017-08-31
Fixing offscreen GUI issue

### LINKED ISSUE #11207: [Windows] GUI window can still appear off screen.
The GUI does not appear on windows if the last position of the window is off screen, as would be with multiple monitors and then subsequently disconnecting one monitor. I believe this is related to #7869 and #11171. It seems that e9ff818b69c2f8ce4a151d1a81a3e22a4319c93d did not fix this problem.

This is easy to replicate; open regedit.exe (windows registry editor), find the registry keys in `HKEY_CURRENT_USER\Software\Bitcoin\Bitcoin-Qt`. Edit the `nWindowPos` entry to be a point that is off screen. The next start of Bitcoin Core will result in the window being off screen and there is no way to get it back without doing `-resetguisettings`

meshcollider:
Fixes https://github.com/bitcoin/bitcoin/issues/11207

This still obviously needs to be confirmed as the cause of the issues in question, this is the current assumption

Edit: c.f. https://github.com/bitcoin/bitcoin/pull/11335

### DIFF
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -857,10 +857,9 @@ void restoreWindowGeometry(const QString& strSetting, const QSize& defaultSize, 
 
     parent->resize(size);
     parent->move(pos);
-
-    if ((!pos.x() && !pos.y()) || (QApplication::desktop()->screenNumber(parent) == -1))
-    {
-        QRect screen = QApplication::desktop()->screenGeometry();
+    
+    QRect screen = QApplication::desktop()->screenGeometry();
+    if ((!pos.x() && !pos.y()) || pos.x() > screen.width() || pos.y() > screen.height() || (QApplication::desktop()->screenNumber(parent) == -1)) {
         QPoint defaultPos((screen.width() - defaultSize.width()) / 2,
                           (screen.height() - defaultSize.height()) / 2);
         parent->resize(defaultSize);
--- a/src/qt/guiutil.cpp
+++ b/src/qt/guiutil.cpp
@@ -857,9 +857,10 @@ void restoreWindowGeometry(const QString& strSetting, const QSize& defaultSize, 
 
     parent->resize(size);
     parent->move(pos);
-    
-    QRect screen = QApplication::desktop()->screenGeometry();
-    if ((!pos.x() && !pos.y()) || pos.x() > screen.width() || pos.y() > screen.height() || (QApplication::desktop()->screenNumber(parent) == -1)) {
+
+    int screen_number = QApplication::desktop()->screenNumber(parent);
+    QRect screen = QApplication::desktop()->availableGeometry(screen_number);
+    if ((!pos.x() && !pos.y()) || screen_number == -1 || !screen.contains(parent->frameGeometry())) {
         QPoint defaultPos((screen.width() - defaultSize.width()) / 2,
                           (screen.height() - defaultSize.height()) / 2);
         parent->resize(defaultSize);


MarcoFalke:
Tagged for backport in case there will be another rc

jtimon:
utACK 7765030

src/qt/guiutil.cpp
```diff
@@ -857,10 +857,9 @@ void restoreWindowGeometry(const QString& strSetting, const QSize& defaultSize,
 
     parent->resize(size);
     parent->move(pos);
-
-    if ((!pos.x() && !pos.y()) || (QApplication::desktop()->screenNumber(parent) == -1))
-    {
-        QRect screen = QApplication::desktop()->screenGeometry();
+    
+    QRect screen = QApplication::desktop()->screenGeometry();
```
promag: Use `availableGeometry()`.

promag (commented):
Concept ACK.

achow101:
ACK 776503029d5f9e6976d65aa6d8ee1958f8afef04

Tested on Windows 10, confirms that this fixes the issue I identified in #11207. However #11171 sounds like it is a different problem.

src/qt/guiutil.cpp
MeshCollider: Yep sounds good. I'll fix it to availableGeometry(screen_number) too now that I think about it. Will push it as a separate commit for review before squashing.

src/qt/guiutil.cpp
```diff
@@ -857,10 +857,9 @@ void restoreWindowGeometry(const QString& strSetting, const QSize& defaultSize,
...
+    if ((!pos.x() && !pos.y()) || pos.x() > screen.width() || pos.y() > screen.height() || (QApplication::desktop()->screenNumber(parent) == -1)) {
```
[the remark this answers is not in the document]
MeshCollider: parent->geometry() does not include the window frame, so perhaps frameGeometry()? But yeah I guess a check for full containment makes sense rather than just the point pos()

AllanDoensen:
I believe this was fixed here : #10156

src/qt/guiutil.cpp
```diff
@@ -857,10 +857,10 @@ void restoreWindowGeometry(const QString& strSetting, const QSize& defaultSize,
 
     parent->resize(size);
     parent->move(pos);
-
-    if ((!pos.x() && !pos.y()) || (QApplication::desktop()->screenNumber(parent) == -1))
-    {
-        QRect screen = QApplication::desktop()->screenGeometry();
+
```
promag: Remove whitespace.

src/qt/guiutil.cpp
```diff
@@ -857,10 +857,10 @@ void restoreWindowGeometry(const QString& strSetting, const QSize& defaultSize,
...
     parent->resize(size);
     parent->move(pos);
-
-    if ((!pos.x() && !pos.y()) || (QApplication::desktop()->screenNumber(parent) == -1))
-    {
-        QRect screen = QApplication::desktop()->screenGeometry();
+    
+    int screen_number = QApplication::desktop()->screenNumber(parent);
+    QRect screen = QApplication::desktop()->availableGeometry(screen_number);
+    if ((!pos.x() && !pos.y()) || screen_number == -1 || !screen.contains(parent->frameGeometry())) {
```
promag: If I understand correctly, `!pos.x() && !pos.y()` should be replaced with `!settings.contains(strSettings + "Pos")`.

src/qt/guiutil.cpp
```diff
@@ -857,10 +857,10 @@ void restoreWindowGeometry(const QString& strSetting, const QSize& defaultSize,
 
     parent->resize(size);
     parent->move(pos);
-
-    if ((!pos.x() && !pos.y()) || (QApplication::desktop()->screenNumber(parent) == -1))
-    {
-        QRect screen = QApplication::desktop()->screenGeometry();
+    
+    int screen_number = QApplication::desktop()->screenNumber(parent);
+    QRect screen = QApplication::desktop()->availableGeometry(screen_number);
```
promag: What if the result of `availableGeometry()` when `screen_number = 1`?

MeshCollider: Do you mean -1? If the parameter is -1, it returns the default screen

src/qt/guiutil.cpp
MeshCollider: Safer to check after converting it to a point, just to make sure conversion worked properly?

src/qt/guiutil.cpp
promag: Fixed above comment.

src/qt/guiutil.cpp
promag: But what if the window position is `QPoint(0, 0)`?

luke-jr (approved):
utACK

src/qt/guiutil.cpp
```diff
@@ -858,9 +858,9 @@ void restoreWindowGeometry(const QString& strSetting, const QSize& defaultSize,
     parent->resize(size);
     parent->move(pos);
 
-    if ((!pos.x() && !pos.y()) || (QApplication::desktop()->screenNumber(parent) == -1))
-    {
-        QRect screen = QApplication::desktop()->screenGeometry();
+    int screen_number = QApplication::desktop()->screenNumber(parent);
+    QRect screen = QApplication::desktop()->availableGeometry(screen_number);
+    if ((!pos.x() && !pos.y()) || screen_number == -1 || !screen.contains(parent->frameGeometry())) {
```
jonasschnelli: Does this allow a window that is halfway moved out of the screen (I think we should allow this)?

MeshCollider: No I don't think it does, but I thought about that and the expected behaviour of most software is that it launches within the screen even if you moved it partially out last time?

sipa:
utACK

laanwj:
> I believe this was fixed here : #10156

I was about to remark this - there have been a few PRs before in the last few years 'fixing' this issue, but it keeps returning (or wasn't fixed properly after all). So normally we would make sure of this by unit tests but I don't think that's an option here due to the GUI-ness. So I suggest creating a manual testplan before/after, so we can verify that this bug is squashed.

jonasschnelli:
Just tested on macOS and windows 8.1 (VM):
Current master OSX: if I move the window outside of the boundary, it will relocate the window to the maximum boundaries without overlapping the visible area.
With this PR OX: overlapping the boundaries will always lead to reopen it in the screen center.
(I think this is acceptable, but slightly different to the current state).

Could not reproduce the issue on Windows 8.1 (did run on a very large screen, closes Bitcoin-Qt while having the main window on the right bottom, reduce screen by times 4, re-started Qt and had the window visible)

laanwj:
So it's unclear whether this is actually needed. Removing the milestone.

laanwj:
Closing in favor of #11335

PR #11210, opened 2017-09-01
Stop test_bitcoin-qt touching ~/.bitcoin

### LINKED ISSUE #11192: test: test_bitcoin-qt creates an empty `~/.bitcoin`
Ideally the tests wouldn't touch the default data directory. Both the core unit tests and the functional tests do this correctly. However,  `src/qt/test/test_bitcoin-qt` seems to create an empty `~/.bitcoin` directory if it doesn't exist yet.
Instead, it should set up a temporary data directory like the other tests.

meshcollider:
Fixes https://github.com/bitcoin/bitcoin/issues/11192

The directory remains unused, but this stops the tests touching ~/.bitcoin at all (namely creating it if it doesn't exist)

### DIFF
--- a/src/qt/test/test_main.cpp
+++ b/src/qt/test/test_main.cpp
@@ -53,6 +53,10 @@ int main(int argc, char *argv[])
     SetupNetworking();
     SelectParams(CBaseChainParams::MAIN);
     noui_connect();
+    ClearDatadirCache();
+    fs::path pathTemp = fs::temp_directory_path() / strprintf("test_bitcoin-qt_%lu_%i", (unsigned long)GetTime(), (int)GetRand(100000));
+    fs::create_directories(pathTemp);
+    gArgs.ForceSetArg("-datadir", pathTemp.string());
 
     bool fInvalid = false;
 
@@ -96,6 +100,8 @@ int main(int argc, char *argv[])
         fInvalid = true;
     }
 #endif
+
+    fs::remove_all(pathTemp);
 
     return fInvalid;
 }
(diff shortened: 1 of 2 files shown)


MeshCollider:
Travis failure is in an unrelated test, please rerun :)

jonasschnelli:
utACK dea086f498097d19a2c9acbfc753c9c2d68dbb03

cdecker:
utACK dea086f498097d19a2c9acbfc753c9c2d68dbb03

src/qt/test/test_main.cpp
```diff
@@ -53,6 +53,10 @@ int main(int argc, char *argv[])
     SetupNetworking();
     SelectParams(CBaseChainParams::MAIN);
     noui_connect();
+    ClearDatadirCache();
```
sipa: Any reason to only do this for Qt?

laanwj: We already do it on `src/test/test_bitcoin.cpp`

laanwj:
utACK, thanks!

laanwj:
@MarcoFalke mentioned needing a test for this on travis (that the tests don't touch `~/.bitcoin`) but this can be done in a separate PR

MarcoFalke:
Making sure that `~/.bitcoin` istn't touched implies that something creates it first. Not sure if we want that in our `.travis.yml`.

luke-jr:
@MarcoFalke Eh, we can just add a check that `~/.bitcoin` still doesn't exist at the end?

MarcoFalke:
Iirc we did that and it failed occasionally. This is on my list of things
to take a look at ...

PR #11212, opened 2017-09-01
Change the type of BlockStatus to unsigned int 

jjz:
1. The type used when using BlockStatus is unsigned int
2. The 16bits range is sufficient to use
3. unsigned int is whatever unsigned integer the compiler likes best

### DIFF
--- a/src/chain.h
+++ b/src/chain.h
@@ -122,7 +122,7 @@ struct CDiskBlockPos
 
 };
 
-enum BlockStatus: uint32_t {
+enum BlockStatus: unsigned int {
     //! Unused.
     BLOCK_VALID_UNKNOWN      =    0,
 


MarcoFalke:
Is there any observable effect of this?

jjz:
There is no obvious impact on performance , but the code is clean . the same data , why use different data types?

gmaxwell:
Please do not switch to types of implementation size without good reason, they complicate review and analysis. Various thoughtful secure coding practice guides prohibit them entirely, and although we don't we probably don't want more of them without good reason.

The type your description is calling for is uint_fast16_t but it's very likely that we'll add many more flags in the future, and that type is the uint32_t on all our supported systems in any case.

jjz:
So should we change this code 
` unsigned int nStatus;`
to 
`uint32_t  nStatus`? Line 203 in the chain.h @gmaxwell

PR #11213, opened 2017-09-01
[trivial] [rpc] Fix getrawtransaction help for per-txout chainstate db

jnewbery:
Builds on #10275 since that is also touching this help string and I don't want to cause an unnecessary rebase.

Only commit for this PR is:

```
[trivial] fix getrawtransaction help

After per-txout chainstate db, getrawtransaction no longer works for
transactions with unspent outputs (unless -txindex is being used)
```


jnewbery:
This is wrong

PR #11219, opened 2017-09-02
[docs] explain how to recompile a modified unit test

Sjors:
It was not obvious to me to run `make` inside the test directory, especially because `make src/test` and `make src/test/test_bitcoin` result in `make: Nothing to be done for ...`.

### DIFF
--- a/src/test/README.md
+++ b/src/test/README.md
@@ -5,7 +5,10 @@ and tests weren't explicitly disabled.
 
 After configuring, they can be run with `make check`.
 
-To run the bitcoind tests manually, launch `src/test/test_bitcoin`.
+To run the bitcoind tests manually, launch `src/test/test_bitcoin`. To recompile
+after a test file was modified, run `make` and then run the test again. If you
+modify a non-test file, use `make -C src/test` to recompile only what's needed
+to run the bitcoind tests.
 
 To add more bitcoind tests, add `BOOST_AUTO_TEST_CASE` functions to the existing
 .cpp files in the `test/` directory or add new .cpp files that


achow101:
Use `make -C src/test` instead of changing to the test directory and running `make`.

Sjors:
@achow101 thanks, I'll update the instruction.

laanwj:
Worded like this it is not clear to me what is the advantage of this, compared to re-running make in the top-level directory.
After all, make detects what files are changed and only recompiles those.
Or is this when you change a non-test compilation unit and only want to see how it affects the unit test?

Sjors:
The latter scenario was the problem I ran into. I push a hopefully more clear instruction.

laanwj:
Looks good to me now. Tested ACK 478d4fb79c878c560ed9179b24b48e884a3b0c13

src/test/README.md
```diff
@@ -5,7 +5,10 @@ and tests weren't explicitly disabled.
 
 After configuring, they can be run with `make check`.
 
-To run the bitcoind tests manually, launch `src/test/test_bitcoin`.
+To run the bitcoind tests manually, launch `src/test/test_bitcoin`. To recompile
+after a test file was modified, run `make` and then run the test again. If you
+modify a non-test file, use `make -C src/test` to recompile only what's needed
+to run the bitcoind tests.
```
glowang, line 11: I'm a bit confused by this description...excuse me for asking some beginner questions here...😂 
I couldn't find a lot of info on how to use make and `-C`, but I did find one here. According to this [post](https://www.gnu.org/software/make/manual/make.html#Phony-Targets) about phony targets, I derived that the dir after -C flag is the target subdir to build:

> Phony targets are also useful in conjunction with recursive invocations of make (see Recursive Use of make). In this situation the makefile will often contain a variable which lists a number of sub-directories to be built. A simplistic way to handle this is to define one rule with a recipe that loops over the sub-directories, like this:
> 
> SUBDIRS = foo bar baz
> 
> subdirs:
>         for dir in $(SUBDIRS); do \
>           $(MAKE) -C $$dir; \
>         done
> 

If so, `make -C src/test` is only building the test directory, right? Why is it that if one `modify a non-test file`, one needs to rebuild the `src/test`? They should not have to rebuild `src/test` if they didn't modify any test file?

MarcoFalke: the test depend on bitcoin core, so if a non-test file (a bitcoin core file) like init.cpp is modified, it will need to be recompiled as well.

init.cpp is then put into libbitcoin_server, which is then linked into test_bitcoin

glowang: Sounds like either the a test or a non-test file gets changed, bitcoin core (the src dir?) needs to be recompiled. Then when is `make -C src/test` relevant? Because this command only compiles what is in `src/test` right? The documentation says `If you
modify a non-test file, use `make -C src/test` to recompile only what's needed
to run the bitcoind tests`, and it seems to imply that when a non-test file gets changed, we only need to compile `src/test`?

Sjors: Tests depend on individual source files, not all of `bitcoind`. So if you're working on `src/descriptor.cpp` it's faster to do `make -C src/test && src/test/test_bitcoin  --run_test=descriptor_tests`

MarcoFalke: > Tests depend on individual source files, not all of bitcoind

I don't think this is true. Tests depend on all the libraries that bitcoind depends on. The only thing they don't depend on is bitcoind.cpp

Issue #8670, opened 2016-09-06
Replacing Boost Test Framework


JeremyRubin:
The unit tests are pretty slow (especially under wine) which can cause timeouts in travis.

Furthermore, it's another boost dependency we can probably get rid of.

This issue is to discuss what features people would like in a new test framework.

To start a list of features that people might like:
- printing the expression, value, and line number on failure
- checkpointing in case of segfaults
- regex filter tests
- progress meter/output currently running test
- test benchmarking info/ability to regression test performance

Please chime in with anything else that you think should go into it.

Christewart:
It would be nice to have some sort of support for property based testing, as per #8469. Conveniently the framework I integrate in that pull request supports integration with Boost.

JeremyRubin:
@Christewart we're trying to get rid of boost.

MarcoFalke:
There is a bunch of `BOOST_CHECK()`s in the code( where `BOOST_CHECK_EQUAL()`s should be used instead) causing at least one assignment error (#8673). So would it make sense to have `BOOST_CHECK()` fail on non-boolean values, if it is possible at all?

JeremyRubin:
@MarcoFalke I have no clue how to do that; short of a compile time regex (http://cpptruths.blogspot.com/2011/07/compile-time-regex-matcher-using.html).

JeremyRubin:
I also like that GDB can attach itself on a failure. I don't like that it seems to launch a gui gdb needing root to attach on x sessions though.

Christewart:
@JeremyRubin I think it would be worthwhile to building in a property based testing framework into our custom framework if we are hell bent on having building a framework from scratch. I'm not sure if I support completely eliminating the Boost dependency, since the overhead of maintaining the new framework might be more trouble than its worth. 

Would it just be more wise to run our own fork of Boost and fix the performance issues you mentioned above? Could those features you mentioned above be layered on top of Boost?

sipa:
Feature request: a command line option to make the application `abort()` on the first error, allowing the creation of a core dump and inspect afterwards in gdb.

laanwj:
> The unit tests are pretty slow (especially under wine) which can cause timeouts in travis.

Also on FreeBSD: `test_bitcoin takes` 236.803s, whereas on Linux only 40s on older hardware.

It's starting to look like boost::test is optimized specifically for Linux.

laanwj:
[...]

PR #6898, opened 2015-10-28
Rewrite CreateNewBlock

morcos:
WARNING: This hasn't been used for mainnet mining yet as far as I know.

~~NOTE: This includes a commit which changes the default block priority size to 0.  This code has been optimized for not including transactions in a priority space.  If block priority space > 0 then the code does not offer too much of a performance improvement over the existing code.~~ (EDIT: now its fast regardless)

The mempool is explicitly assumed to be responsible for maintaining consistency of transactions with respect to not spending non-existent outputs, not double spending, script validity and coinbase maturity.  Only finality of transactions is checked before assembling a block.

A final call to TestBlockValidity is still performed.  If an invalid block has been constructed, then an error is thrown ~~logged and a NULL pointer is returned.~~ (EDIT: reverted to original behavior with more informative error)

This code produces the same blocks as the code it replaces with the following exceptions:
- The fee rate sort tie breaker is hash instead of priority
- The priority sort has a secondary tie breaker of hash
- ~~The priority block is not limited to transactons above AllowFree()~~ (EDIT: fixed)
- ~~The fee rate sorted part of the block is not limited to transactions above minRelayTxFee~~ (EDIT: fixed)
- (EDIT: The blocks are the same if you keep scanning to the end to try to fill up the last remaining space, but this code stops after trying 50 times to fill the last 1000 bytes)

Comparing this to the original code over 2000 calls to CreateNewBlock over the last 2 days.
(blockprioritysize=0, maxmempool=300, dbcache=1000, maxsigcachesize=1000000)
Time in ms

|  | Time to assemble block | Additional time to perform final TestBlockValidity |
| --- | --- | --- |
| master | 2602 | 240 |
| this pull | 14 | 438 |

The extra slowness in TestBlockValidity is because the cache used to be warmed up in the assembly code.

PR #11220, opened 2017-09-02
Check specific validation error in miner tests

Sjors:
## Problem

`BOOST_CHECK_THROW` merely checks that some `std::runtime_error` is
thrown, but not which one.

Here's an example of how this can cause a test to pass when a developer
introduces a consensus bug. The test for the sigops limit assumes
that `CreateNewBlock` fails with `bad-blk-sigops`. However it can
also fail with bad-txns-vout-negative, if a naive developer lowers
`BLOCKSUBSIDY` to `1*COIN`.

## Solution

`BOOST_CHECK_EXCEPTION` allows an additional predicate function. This
commit uses this for all exceptions that are checked for in
`miner_tets.cpp`:
* `bad-blk-sigops`
* `bad-cb-multiple`
* `bad-txns-inputs-missingorspent`
* `block-validation-failed`

If the function throws a different error, the test will fail. Although the message produced by Boost is a bit [confusing](http://boost.2283326.n4.nabble.com/Test-BOOST-CHECK-EXCEPTION-error-message-still-vague-tt4683257.html#a4683554), it does show which error was actually thrown. Here's what the above `1*COIN` bug would result in:

schermafbeelding 2017-09-02 om 23 42 29
 
## Other considerations

A more elegant solution in my opinion would be to subclass `std::runtime_error` for each `INVALID_TRANSACTION` type, but this would involve touching consensus code.

I put the predicates in `test_bitcoin.h` because I assume they can be reused in other test files. However [serialize_tests.cpp](https://github.com/bitcoin/bitcoin/blob/v0.15.0rc3/src/test/serialize_tests.cpp#L245) also uses `BOOST_CHECK_EXCEPTION` and it defines the predicate in the test file itself.

Instead of four `IsRejectInvalidReasonX(std::runtime_error const& e)` functions, I'd prefer something reusable like `bool IsRejectInvalidReason(String reason)(std::runtime_error const& e)`, which would be used like `BOOST_CHECK_EXCEPTION(functionThatThrows(), std::runtime_error, IsRejectInvalidReason("bad-blk-sigops")`. I couldn't figure out how to do that in C++.

### DIFF
--- a/src/test/miner_tests.cpp
+++ b/src/test/miner_tests.cpp
@@ -26,6 +26,17 @@
 
 BOOST_FIXTURE_TEST_SUITE(miner_tests, TestingSetup)
 
+// BOOST_CHECK_EXCEPTION predicates to check the specific validation error
+class HasReason {
+public:
+    HasReason(const std::string& reason) : m_reason(reason) {}
+    bool operator() (const std::runtime_error& e) const {
+        return std::string(e.what()).find(m_reason) != std::string::npos;
+    };
+private:
+    const std::string m_reason;
+};
+
 static CFeeRate blockMinFeeRate = CFeeRate(DEFAULT_BLOCK_MIN_TX_FEE);
 
 static BlockAssembler AssemblerForTest(const CChainParams& params) {
@@ -264,7 +275,8 @@ BOOST_AUTO_TEST_CASE(CreateNewBlock_validity)
         mempool.addUnchecked(hash, entry.Fee(LOWFEE).Time(GetTime()).SpendsCoinbase(spendsCoinbase).FromTx(tx));
         tx.vin[0].prevout.hash = hash;
     }
-    BOOST_CHECK_THROW(AssemblerForTest(chainparams).CreateNewBlock(scriptPubKey), std::runtime_error);
+
+    BOOST_CHECK_EXCEPTION(AssemblerForTest(chainparams).CreateNewBlock(scriptPubKey), std::runtime_error, HasReason("bad-blk-sigops"));
     mempool.clear();
 
     tx.vin[0].prevout.hash = txFirst[0]->GetHash();
@@ -304,7 +316,7 @@ BOOST_AUTO_TEST_CASE(CreateNewBlock_validity)
     // orphan in mempool, template creation fails
     hash = tx.GetHash();
     mempool.addUnchecked(hash, entry.Fee(LOWFEE).Time(GetTime()).FromTx(tx));
-    BOOST_CHECK_THROW(AssemblerForTest(chainparams).CreateNewBlock(scriptPubKey), std::runtime_error);
+    BOOST_CHECK_EXCEPTION(AssemblerForTest(chainparams).CreateNewBlock(scriptPubKey), std::runtime_error, HasReason("bad-txns-inputs-missingorspent"));
     mempool.clear();
 
     // child with higher feerate than parent
@@ -332,7 +344,8 @@ BOOST_AUTO_TEST_CASE(CreateNewBlock_validity)
     hash = tx.GetHash();
     // give it a fee so it'll get mined
     mempool.addUnchecked(hash, entry.Fee(LOWFEE).Time(GetTime()).SpendsCoinbase(false).FromTx(tx));
-    BOOST_CHECK_THROW(AssemblerForTest(chainparams).CreateNewBlock(scriptPubKey), std::runtime_error);
+    // Should throw bad-cb-multiple
+    BOOST_CHECK_EXCEPTION(AssemblerForTest(chainparams).CreateNewBlock(scriptPubKey), std::runtime_error, HasReason("bad-cb-multiple"));
     mempool.clear();
 
     // double spend txn pair in mempool, template creation fails
@@ -345,7 +358,7 @@ BOOST_AUTO_TEST_CASE(CreateNewBlock_validity)
     tx.vout[0].scriptPubKey = CScript() << OP_2;
     hash = tx.GetHash();
     mempool.addUnchecked(hash, entry.Fee(HIGHFEE).Time(GetTime()).SpendsCoinbase(true).FromTx(tx));
-    BOOST_CHECK_THROW(AssemblerForTest(chainparams).CreateNewBlock(scriptPubKey), std::runtime_error);
+    BOOST_CHECK_EXCEPTION(AssemblerForTest(chainparams).CreateNewBlock(scriptPubKey), std::runtime_error, HasReason("bad-txns-inputs-missingorspent"));
     mempool.clear();
 
     // subsidy changing
@@ -389,7 +402,8 @@ BOOST_AUTO_TEST_CASE(CreateNewBlock_validity)
     tx.vout[0].nValue -= LOWFEE;
     hash = tx.GetHash();
     mempool.addUnchecked(hash, entry.Fee(LOWFEE).Time(GetTime()).SpendsCoinbase(false).FromTx(tx));
-    BOOST_CHECK_THROW(AssemblerForTest(chainparams).CreateNewBlock(scriptPubKey), std::runtime_error);
+    // Should throw block-validation-failed
+    BOOST_CHECK_EXCEPTION(AssemblerForTest(chainparams).CreateNewBlock(scriptPubKey), std::runtime_error, HasReason("block-validation-failed"));
     mempool.clear();
 
     // Delete the dummy blocks again.


luke-jr:
Most of the rejection reasons are not standardised, and we only care that it fails. It might make the most sense to check that it's a `consensus_rule_violation` exception or such, but I'm not sure checking the specific string is useful.

Your example is invalid, because for the `bad-blk-sigops` test, we would not care what the outcome is if the subsidy is broken. Instead, the subsidy should be tested separately, not implicitly.

sdaftuar:
Concept ACK.  I think in general checking for the specific error we expect makes it less likely that we inadvertently introduce a bug in the tests; certainly this kind of thing has cropped up in the functional test suite, where bugs introduced into a test harness cause the test to not actually be testing anything, or to be testing the wrong thing.

IMO the downside (in general) to additional specificity is that it adds overhead when making changes to the consensus code rejection reasons, because now all the tests that were hardcoded for a particular reason need updating.  This patch seems pretty small though, and I don't think that should be much of a concern here.  In fact I think we largely already do similar reject-reason-pattern-matching in the functional tests as well.

Sjors:
I'll see what I can do about Travis failing on everything but OSX. It's throwing `multiple definition of IsRejectInvalidReason` in a bunch of places.

I agree the example isn't great. I did actually run into it while trying an extremely high value for MAX_BLOCK_SIGOPS_COST (250x) in an [experiment](https://github.com/Sjors/bitcoin/pull/1). In attempt to make the test pass, I increased the transaction generation loops from 1001 to 250001, which obviously(?) took forever to run. I then tried fewer loops to get a sense of performance. To my surprise the test passed for some seemingly random intermediate value. Once I found the actual error message I realized it was an error in the test; it just ran of coins somewhere in the loop.

Sjors:
I pushed a few improvements, apparently to the satisfaction of Travis:

1. moved changes from `test_bitcoin.h` to `testutil.h`, which seems more appropriate
2. instead of one function for each validation error type, there's now a `class CheckRejectInvalid` whose constructor takes a string
3. `#define` for validation error strings so they're all in one place. That should make later refactoring easier and also prevent typos.

So now each test looks like this:

```cpp
BOOST_CHECK_EXCEPTION(
    functionThatThrows(),
    std::runtime_error,
    CheckRejectInvalid(REJECT_INVALID_BAD_CB_MULTIPLE)
);
```

Unfortunately as a result of this change, when a test throws a different exception, it no longer shows what that exception was. However, this is easy for a developer to debug: just put `BOOST_CHECK(functionThatThrows())` above the failing `BOOST_CHECK_EXCEPTION ` and run the test again.

Sjors:
@sipa and @morcos worked on `miner_tests.cpp` earlier this year.

@morcos since you refactored `CreateNewBlock` in late 2015 (#6898), could you check that I picked the right specific validation errors? Just to make sure the tests weren't already misbehaving.

src/test/testutil.h
```diff
@@ -9,7 +9,27 @@
...
 fs::path GetTempPath();
 
+// Validation errors used in tests (prevent typos and facilitate refactoring):
+#define REJECT_INVALID_BAD_BLK_SIGOPS "bad-blk-sigops"
+#define REJECT_INVALID_BAD_CB_MULTIPLE "bad-cb-multiple"
+#define REJECT_INVALID_BAD_TXNS_INPUTS_MISSINGORSPENT "bad-txns-inputs-missingorspent"
+#define REJECT_INVALID_BLOCK_VALIDATION_FAILED "block-validation-failed"
+
+// BOOST_CHECK_THROW predicates to check the specific validation error
+class CheckRejectInvalid {
```
MarcoFalke: Maybe move all of this to the cpp?

Sjors:
@MarcoFalke  since #11234 just snatched `testutil.h` out from under me, I moved everything into `miner_test.cpp`.

There are a lot of other tests that use `BOOST_CHECK_THROW` without checking the precise error, so it might make sense to revert #11234 at some point.

MarcoFalke:

Sjors:
Done, as well as rebased to latest master.

Sjors:
Done, as well as rebased to latest master.

Sjors:
I pushed again to fix commit message and code comment (it was saying "BOOST_CHECK_THROW" instead of "BOOST_CHECK_EXCEPTION" in two places).

Sjors:
The `p2p-segwit.py` test failed in one environment. Random time-out? All machines passed on my previous force push.
schermafbeelding 2017-10-03 om 11 16 49

MeshCollider:
@Sjors probably just an unrelated failure, Travis does that sometimes. I've restarted that test for you.

src/test/miner_tests.cpp
```diff
@@ -26,6 +27,25 @@
 
 BOOST_FIXTURE_TEST_SUITE(miner_tests, TestingSetup)
 
+// Validation errors used in tests (prevent typos and facilitate refactoring):
+#define REJECT_INVALID_BAD_BLK_SIGOPS "bad-blk-sigops"
```
promag: IMO inline these strings.

src/test/miner_tests.cpp
```diff
@@ -17,6 +17,7 @@
 #include "uint256.h"
 #include "util.h"
 #include "utilstrencodings.h"
+#include <boost/algorithm/string.hpp>
```
promag: Move down near other boost includes (try to keep it sorted too).

src/test/miner_tests.cpp
```diff
@@ -26,6 +27,25 @@
...
+// Validation errors used in tests (prevent typos and facilitate refactoring):
+#define REJECT_INVALID_BAD_BLK_SIGOPS "bad-blk-sigops"
+#define REJECT_INVALID_BAD_CB_MULTIPLE "bad-cb-multiple"
+#define REJECT_INVALID_BAD_TXNS_INPUTS_MISSINGORSPENT "bad-txns-inputs-missingorspent"
+#define REJECT_INVALID_BLOCK_VALIDATION_FAILED "block-validation-failed"
+
+// BOOST_CHECK_EXCEPTION predicates to check the specific validation error
+class CheckRejectInvalid {
+public:
+    CheckRejectInvalid(std::string const& arg) {
```
promag:
```cpp
CheckRejectInvalid(const std::string& reason) : m_reason(reason) {}
```

src/test/miner_tests.cpp
```diff
@@ -26,6 +27,25 @@
...
+        reason = arg;
+    };
+    bool operator() (std::runtime_error const& e) const {
+        return boost::contains(e.what(), reason);
```
promag: Use `std::string::find`?

src/test/miner_tests.cpp
```diff
@@ -26,6 +27,25 @@
...
+#define REJECT_INVALID_BAD_TXNS_INPUTS_MISSINGORSPENT "bad-txns-inputs-missingorspent"
+#define REJECT_INVALID_BLOCK_VALIDATION_FAILED "block-validation-failed"
+
+// BOOST_CHECK_EXCEPTION predicates to check the specific validation error
+class CheckRejectInvalid {
+public:
+    CheckRejectInvalid(std::string const& arg) {
+        reason = arg;
+    };
+    bool operator() (std::runtime_error const& e) const {
```
promag: Nit, `const std::runtime_error& e`.

src/test/miner_tests.cpp
```diff
@@ -264,7 +284,9 @@ BOOST_AUTO_TEST_CASE(CreateNewBlock_validity)
         mempool.addUnchecked(hash, entry.Fee(LOWFEE).Time(GetTime()).SpendsCoinbase(spendsCoinbase).FromTx(tx));
         tx.vin[0].prevout.hash = hash;
     }
-    BOOST_CHECK_THROW(AssemblerForTest(chainparams).CreateNewBlock(scriptPubKey), std::runtime_error);
+
+    // This should throw bad-blk-sigops
+    BOOST_CHECK_EXCEPTION(AssemblerForTest(chainparams).CreateNewBlock(scriptPubKey), std::runtime_error, CheckRejectInvalid(REJECT_INVALID_BAD_BLK_SIGOPS));
```
promag: IMO this sounds better:
```cpp
..., std::runtime_error, HasReason("bad-blk-sigops"));
```

promag (commented):
Agree with better testing the exception.

src/test/miner_tests.cpp
Sjors: That should indeed help with #8670.

src/test/miner_tests.cpp
Sjors: I can rename `CheckRejectInvalid` to `HasReason`; seems more readable indeed.

src/test/miner_tests.cpp
```diff
@@ -26,6 +27,25 @@
...
+class CheckRejectInvalid {
+public:
+    CheckRejectInvalid(std::string const& arg) {
+        reason = arg;
+    };
+    bool operator() (std::runtime_error const& e) const {
+        return boost::contains(e.what(), reason);
+    };
+private:
+    std::string reason;
```
MarcoFalke: µnit: Can be `const`

src/test/miner_tests.cpp
Sjors: I used `#define` to make it easier to find all occurrences in the codebase if these rejection reasons ever need to be changed. It also helps against typos, although typos would be immediately obvious when you run the test.

Happy to change it to inline if nobody objects.

src/test/miner_tests.cpp
Sjors: Thanks. If it wasn't obvious, I'm not terribly familiar with C++. :-)

src/test/miner_tests.cpp
promag: The test will fail if there is a typo. Having inline is good because a rename forces you to update the relevant tests and double check everything related.

src/test/miner_tests.cpp
Sjors: I got rid of this dependency.

src/test/miner_tests.cpp
Sjors: Using inline now.

src/test/miner_tests.cpp
Sjors: Done, and renamed the private variable from `reason` to `m_reason`

src/test/miner_tests.cpp
Sjors: Done

src/test/miner_tests.cpp
Sjors: Done

src/test/miner_tests.cpp
Sjors: Done (Github needs a check-mark emoticon)

Sjors:
I believe I addressed all code style issued raised. Rebased to latest master. I also signed the commit this time.

Sjors:
@JeremyRubin it would be great to get your feedback on this PR, given that you're working on these validation errors. I was able to cherry-pick my commit on top of your #11523 without a problem.

JeremyRubin:
Concept Ack!

I would think one improvement might be if the strings were not to be literals (as Luke points out, they aren't standardized). I think it's ok for now though.

Sjors:
@JeremyRubin the downside of squashing commits is you can't see history :-) I actually used `#define REJECT_INVALID_BAD_CB_MULTIPLE "invalid_bad_cb_multiple"` in an earlier version, but switched back to literal strings  based on @promag's [feedback](https://github.com/bitcoin/bitcoin/pull/11220#discussion_r142387624).

sipa:
Concept ACK

MarcoFalke:
utACK 59d0ede514f7754ad4df11ffe11f27e2ea63bf3e

Sjors:
Rebased to account for #11389.

PR #11221, opened 2017-09-03
Refactor: simpler read

gnuser:


src/streams.h
```diff
@@ -345,18 +345,16 @@ class CDataStream
 
         // Read from the beginning of the buffer
         unsigned int nReadPosNext = nReadPos + nSize;
-        if (nReadPosNext >= vch.size())
+        if (nReadPosNext > vch.size()) {
+            throw std::ios_base::failure("CDataStream::read(): end of data");
+        }
+        memcpy(pch, &vch[nReadPos], nSize);
```
sipa: This is not correct. If nReadPos = 0, and nSize = 0, this will call `operator[]` on `vch`, which is illegal when it is empty.

MeshCollider: Note that that's a bug with the original code though, not with the refactor

dcousens (approved):

src/streams.h
laanwj: probably better to use `vch.data() + nNreadPos`?

promag: @sipa above there is an early return if `nSize == 0` so that doesn't happen because of the `throw`.

promag:
utACK 9db9d62.

Nit, commit message could mention `CDataStream::read` instead `read` only.

src/streams.h
```diff
@@ -345,18 +345,16 @@ class CDataStream
 
         // Read from the beginning of the buffer
         unsigned int nReadPosNext = nReadPos + nSize;
-        if (nReadPosNext >= vch.size())
+        if (nReadPosNext > vch.size()) {
+            throw std::ios_base::failure("CDataStream::read(): end of data");
+        }
+        memcpy(pch, &vch[nReadPos], nSize);        
+        if (nReadPosNext == vch.size())
```
promag: Nit, `{` on this line.

src/streams.h
sipa: Agree, there is no bug here. It would still be clearer using `vch.data() + nReadPos` though.

PR #11101, opened 2017-08-21
bitcoin_qt.m4: Minor fixes, improve code orthogonality

lemzwerg:
Please use

  git diff --word-diff-regex=. --word-diff=color

to view the differences.

The commit contains the following minor fixes.

  . Only include '<QtCore/qconfig.h> for Qt5's QT_VERSION tests.
    For Qt4, also include '<QtCore/qglobal.h> for the QT_VERSION tests.
    This makes the tests work even if '-fPIC' or '-fPIE' is not used.
  . Change quotes from incorrect double quotes to brackets in
    'BITCOIN_QT_CONFIGURE'.
  . Add some missing braces around variables in autoconf messages.
  . Add a missing dollar sign in '_BITCOIN_QT_IS_STATIC'.

Additionally, the patch does the following.

  . Add double qoutes to string tests where arguments could (theoretically)
    contain spaces.
  . Remove double quotes where not necessary.
  . Consequently quote all arguments of M4 functions.
  . Mark empty M4 arguments with '[]'.
  . Other minor cosmetic fixes.

PR #11222, opened 2017-09-03
bitcoin_qt.m4: Minor fixes and clean-ups.

lemzwerg:
This is the same as https://github.com/bitcoin/bitcoin/pull/11101 but moved to a separate branch.

### DIFF
--- a/build-aux/m4/bitcoin_qt.m4
+++ b/build-aux/m4/bitcoin_qt.m4
@@ -254,10 +254,10 @@ AC_DEFUN([BITCOIN_QT_CONFIGURE],[
       bitcoin_enable_qt_dbus=yes
     fi
     if test x$use_dbus = xyes && test x$have_qt_dbus = xno; then
-      AC_MSG_ERROR("libQtDBus not found. Install libQtDBus or remove --with-qtdbus.")
+      AC_MSG_ERROR([libQtDBus not found. Install libQtDBus or remove --with-qtdbus.])
     fi
     if test x$LUPDATE = x; then
-      AC_MSG_WARN("lupdate is required to update qt translations")
+      AC_MSG_WARN([lupdate is required to update qt translations])
     fi
   ],[
     bitcoin_enable_qt=no
--- a/build-aux/m4/bitcoin_qt.m4
+++ b/build-aux/m4/bitcoin_qt.m4
@@ -388,7 +388,7 @@ AC_DEFUN([_BITCOIN_QT_FIND_STATIC_PLUGINS],[
            [bitcoin_cv_need_platformsupport=no])
          ])
          if test x$bitcoin_cv_need_platformsupport = xyes; then
-           BITCOIN_QT_CHECK(AC_CHECK_LIB([${QT_LIB_PREFIX}PlatformSupport],[main],,BITCOIN_QT_FAIL(lib$QT_LIB_PREFIXPlatformSupport not found)))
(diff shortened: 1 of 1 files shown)


fanquake:
Please update the PR title to something more explanatory, as they are now included in merge commits.

lemzwerg:
Better  now?

practicalswift:
utACK 40c8a65c5653d6cfd7859700fe4e7d6468ce2231

laanwj:
@theuni Can you take a look at the build system changes here please?

build-aux/m4/bitcoin_qt.m4
```diff
@@ -321,7 +321,7 @@ AC_DEFUN([_BITCOIN_QT_IS_STATIC],[
     [bitcoin_cv_static_qt=yes],
     [bitcoin_cv_static_qt=no])
   ])
-  if test xbitcoin_cv_static_qt = xyes; then
+  if test x$bitcoin_cv_static_qt = xyes; then
```
theuni: This was clearly broken before, but it didn't matter because we define QT_STATICPLUGIN elsewhere. Let's just delete this.

theuni:
I wasn't comfortable reviewing e90d91c6f31e4b8fe74029fecfacf56190c217aa accurately, so I scripted it. Could you please update the commit message to make it a scripted diff? See here for an explanation: https://github.com/bitcoin/bitcoin/blob/master/doc/developer-notes.md#scripted-diffs

The script I used, which matches your commit, is: 
```shell
sed -iE -e 's/"x\(yes\|no\|auto\)\?"/x\1/g' -e 's/test \(x\$[^ ]*\)/test "\1"/g' build-aux/m4/bitcoin_qt.m4
```

theuni:
Did you use some tool or script for the m4 macros? I think it's fine, but I'm nervous about some issue turning up months from now because of an errant bracket.

lemzwerg:
No time for creating a script, sorry.  I suggest that you do a comparison with
```
git diff --word-diff-regex=. --word-diff=color e90d91c^ e90d91c
```
for review; this shows the changes of this commit in the most compact form.

Just a final note regarding bitcoin's M4 stuff for configuring Qt: I now believe that this is the wrong way.  Without the help of `qmake` it will always remain a guesswork what compiler flags should be really used, and what code should be used for static linking.  Have a look at

http://repo.or.cz/ttfautohint.git/blob_plain/HEAD:/m4/autotroll.m4
http://repo.or.cz/ttfautohint.git/blob/HEAD:/configure.ac#l52
http://repo.or.cz/ttfautohint.git/blob/HEAD:/frontend/static-plugins.cpp.in

for my current solution.  For example, if I configure ttfautohint for cross-building with [mxe](http://mxe.cc), `static-plugins.cpp` automatically contains the following code.

```c
// This file is autogenerated by qmake. It imports static plugin classes for
// static plugins specified using QTPLUGIN and QT_PLUGIN_CLASS.<plugin> variables.
#include <QtPlugin>
Q_IMPORT_PLUGIN(QWindowsIntegrationPlugin)
```

lemzwerg:
FYI: autotroll has been updated with my changes and is now available from https://github.com/tsuna/autotroll.

theuni:
@lemzwerg I'd really rather not change all this around arbitrarily. And I'd certainly not require qmake. Especially considering that they'll be replacing it with qbs in the future.

Everything but the last commit looks good to me (after addressing https://github.com/bitcoin/bitcoin/pull/11222#discussion_r142265041). I assume the last is probably ok too, but I'd rather not risk introducing some obscure new error if this doesn't fix a reported bug.

laanwj:
Needs update for @cfields' comments.

fanquake:
Closing in favour of #11711.

PR #11226, opened 2017-09-03
[WIP] Add Clang thread safety analysis annotations: GUARDED_BY(lock) / EXCLUSIVE_LOCKS_REQUIRED(lock)

practicalswift:
Add additional Clang thread safety analysis annotations as discussed with @sipa (https://github.com/bitcoin/bitcoin/pull/10866#issuecomment-316238630) and others.

This is a follow-up to #10866 ~~which this PR is rebased upon (awaiting merge of #10866). The first three commits (8a1dc09, e022990 and cbae151) can be reviewed in #10866.~~

This PR contains three types of changes:
1. Add missing locks (based on the `GUARDED_BY(...)` analysis below). Please review thoroughly.
2. Add `GUARDED_BY(...)` annotations to allow for Clang's thread safety analysis at compile-time. Does not change run-time behaviour.
3. Add `EXCLUSIVE_LOCKS_REQUIRED(...)` annotations to allow for Clang's thread safety analysis at compile-time. Does not change run-time behaviour.

Background reading: The ["C/C++ Thread Safety Analysis"](https://research.google.com/pubs/archive/42958.pdf) paper (Hutchins, Ballman & Sutherland, 2014) paper describing Clang thread safety analysis and how it is used for the Google C++ codebase:

> They essentially provide a static type system for threads, and can detect potential race conditions and deadlocks. This paper describes Clang Thread Safety Analysis, a tool which uses annotations to declare and enforce thread safety policies in C and C++ programs.
> […]
> It has been deployed on a large scale at Google; all C++ code at Google is now compiled with thread safety analysis enabled by default.
> […]
> The `GUARDED_BY` attribute declares that a thread must lock `mu` before it can read or write to `balance`, thus ensuring that the increment and decrement operations are atomic. Similarly, `REQUIRES` declares that the calling thread must lock mu before calling `withdrawImpl`. Because the caller is assumed to have locked `mu,` it is safe to modify `balance` within the body of the method.

### DIFF
--- a/src/addrman.cpp
+++ b/src/addrman.cpp
@@ -65,7 +65,7 @@ double CAddrInfo::GetChance(int64_t nNow) const
     return fChance;
 }
 
-CAddrInfo* CAddrMan::Find(const CNetAddr& addr, int* pnId)
+CAddrInfo* CAddrMan::Find(const CNetAddr& addr, int* pnId) EXCLUSIVE_LOCKS_REQUIRED(cs_addrMan)
 {
     std::map<CNetAddr, int>::iterator it = mapAddr.find(addr);
     if (it == mapAddr.end())
@@ -312,7 +312,7 @@ bool CAddrMan::Add_(const CAddress& addr, const CNetAddr& source, int64_t nTimeP
     return fNew;
 }
 
-void CAddrMan::Attempt_(const CService& addr, bool fCountFailure, int64_t nTime)
+void CAddrMan::Attempt_(const CService& addr, bool fCountFailure, int64_t nTime) EXCLUSIVE_LOCKS_REQUIRED(cs_addrMan)
 {
     CAddrInfo* pinfo = Find(addr);
 
@@ -334,7 +334,7 @@ void CAddrMan::Attempt_(const CService& addr, bool fCountFailure, int64_t nTime)
     }
 }
 
-CAddrInfo CAddrMan::Select_(bool newOnly)
+CAddrInfo CAddrMan::Select_(bool newOnly) EXCLUSIVE_LOCKS_REQUIRED(cs_addrMan)
 {
     if (size() == 0)
         return CAddrInfo();
@@ -480,7 +480,7 @@ void CAddrMan::GetAddr_(std::vector<CAddress>& vAddr) EXCLUSIVE_LOCKS_REQUIRED(c
     }
 }
 
-void CAddrMan::Connected_(const CService& addr, int64_t nTime)
+void CAddrMan::Connected_(const CService& addr, int64_t nTime) EXCLUSIVE_LOCKS_REQUIRED(cs_addrMan)
 {
     CAddrInfo* pinfo = Find(addr);
 
@@ -500,7 +500,7 @@ void CAddrMan::Connected_(const CService& addr, int64_t nTime)
         info.nTime = nTime;
 }
 
-void CAddrMan::SetServices_(const CService& addr, ServiceFlags nServices)
+void CAddrMan::SetServices_(const CService& addr, ServiceFlags nServices) EXCLUSIVE_LOCKS_REQUIRED(cs_addrMan)
 {
     CAddrInfo* pinfo = Find(addr);
 
--- a/src/addrman.cpp
+++ b/src/addrman.cpp
@@ -78,7 +78,7 @@ CAddrInfo* CAddrMan::Find(const CNetAddr& addr, int* pnId)
     return nullptr;
 }
 
-CAddrInfo* CAddrMan::Create(const CAddress& addr, const CNetAddr& addrSource, int* pnId)
+CAddrInfo* CAddrMan::Create(const CAddress& addr, const CNetAddr& addrSource, int* pnId) EXCLUSIVE_LOCKS_REQUIRED(cs)
 {
     int nId = nIdCount++;
     mapInfo[nId] = CAddrInfo(addr, addrSource);
@@ -90,7 +90,7 @@ CAddrInfo* CAddrMan::Create(const CAddress& addr, const CNetAddr& addrSource, in
     return &mapInfo[nId];
 }
 
-void CAddrMan::SwapRandom(unsigned int nRndPos1, unsigned int nRndPos2)
+void CAddrMan::SwapRandom(unsigned int nRndPos1, unsigned int nRndPos2) EXCLUSIVE_LOCKS_REQUIRED(cs)
 {
     if (nRndPos1 == nRndPos2)
         return;
@@ -110,7 +110,7 @@ void CAddrMan::SwapRandom(unsigned int nRndPos1, unsigned int nRndPos2)
     vRandom[nRndPos2] = nId1;
 }
 
-void CAddrMan::Delete(int nId)
+void CAddrMan::Delete(int nId) EXCLUSIVE_LOCKS_REQUIRED(cs)
 {
     assert(mapInfo.count(nId) != 0);
     CAddrInfo& info = mapInfo[nId];
@@ -124,7 +124,7 @@ void CAddrMan::Delete(int nId)
     nNew--;
 }
 
-void CAddrMan::ClearNew(int nUBucket, int nUBucketPos)
+void CAddrMan::ClearNew(int nUBucket, int nUBucketPos) EXCLUSIVE_LOCKS_REQUIRED(cs)
 {
     // if there is an entry in the specified bucket, delete it.
     if (vvNew[nUBucket][nUBucketPos] != -1) {
@@ -139,7 +139,7 @@ void CAddrMan::ClearNew(int nUBucket, int nUBucketPos)
     }
 }
 
-void CAddrMan::MakeTried(CAddrInfo& info, int nId)
+void CAddrMan::MakeTried(CAddrInfo& info, int nId) EXCLUSIVE_LOCKS_REQUIRED(cs)
 {
     // remove the entry from all new buckets
     for (int bucket = 0; bucket < ADDRMAN_NEW_BUCKET_COUNT; bucket++) {
@@ -187,7 +187,7 @@ void CAddrMan::MakeTried(CAddrInfo& info, int nId)
     info.fInTried = true;
 }
 
-void CAddrMan::Good_(const CService& addr, int64_t nTime)
+void CAddrMan::Good_(const CService& addr, int64_t nTime) EXCLUSIVE_LOCKS_REQUIRED(cs)
 {
     int nId;
 
@@ -239,7 +239,7 @@ void CAddrMan::Good_(const CService& addr, int64_t nTime)
     MakeTried(info, nId);
 }
 
-bool CAddrMan::Add_(const CAddress& addr, const CNetAddr& source, int64_t nTimePenalty)
+bool CAddrMan::Add_(const CAddress& addr, const CNetAddr& source, int64_t nTimePenalty) EXCLUSIVE_LOCKS_REQUIRED(cs)
 {
     if (!addr.IsRoutable())
         return false;
@@ -459,7 +459,7 @@ int CAddrMan::Check_()
 }
 #endif
 
-void CAddrMan::GetAddr_(std::vector<CAddress>& vAddr)
+void CAddrMan::GetAddr_(std::vector<CAddress>& vAddr) EXCLUSIVE_LOCKS_REQUIRED(cs)
 {
     unsigned int nNodes = ADDRMAN_GETADDR_MAX_PCT * vRandom.size() / 100;
     if (nNodes > ADDRMAN_GETADDR_MAX)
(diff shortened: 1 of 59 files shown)


sipa:
Awesome, I'm very happy to see someone work on this again.

sipa:
You must have introduced a lock inversion, based on the Travis output.

promag:
Squash commits `Add GUARDED_BY(cs_KeyStore)`?

practicalswift:
@sipa Thanks for letting me know! I've now removed the `LOCK(walletInstance->cs_KeyStore)` in `wallet.cpp` that caused the lock inversion. Do you which variables `cs_KeyStore` is meant to guard?

practicalswift:
@promag Fixed! :-) Regarding the repeated `GUARDED_BY(cs)` commit messages - these actually refer to different `cs`:es. To make things less confusing I've now disambiguated them by adding the class names in the commit message.

Furthermore I've added an commit which renames renames `CAddrMan.cs` to `CAddrMan.cs_addrMan`, and `CTxMemPool.cs` to `CTxMemPool.cs_txMemPool`.

TheBlueMatt:
Concept ACK. This will generate some turbulence in terms of needing to rebase a ton of existing PRs, so maybe it makes sense to break this into several PRs which only touch given modules? In any case, excited to see this stuff happen.

laanwj:
Concept ACK. Though I'm not sure what a good time to merge a huge, spread-out (though low-risk) change like this is. Maybe just before the 0.16 split?

practicalswift:
@laanwj Good point! When is the 0.16 split expected to happen?

The smaller PR #10866 lays the groundwork for all the thread-safety-analysis work. Having that PR merged a bit would help a lot :-)

practicalswift:
@TheBlueMatt Sure, I can split this PR into multiple parts to ease reviewing! Just let me know which parts/modules to split it on :-) To avoid duplicating the commits in #10866 we should probably wait until #10866 is merged before splitting this PR up. Sounds reasonable?

practicalswift:
@sipa @TheBlueMatt @laanwj Thanks for reviewing and thanks for the encouragement! Do you happen to know which core developers that would be appropriate reviewers for this work (on a deeper `GUARDED_BY(…)` level for each module)?

The types of errors I might need some help from reviewers to avoid:
* False positives: Too restrictive lock requirements. Requiring a lock where a lock is not needed. More concretely: unnecessary `GUARDED_BY(…)` annotations. (The `EXCLUSIVE_LOCKS_REQUIRED(…)` annotations follow more or less mechanically for the `GUARDED_BY(…)`.)
* False negatives: Missing lock requirements. Not requiring a lock where a lock is needed. More concretely: missing `GUARDED_BY(…)` annotations.

And perhaps most importantly: help reviewing that the *added* locks in this PR looks reasonable. In constrast to the annotations they alter run-time behaviour.

There are a couple of `AssertLockHeld(…)`:s that I've been unable to map to variables being guarded. I might need some help to track those down to add the corresponding `GUARDED_BY(…)`:s (and hence `EXCLUSIVE_LOCKS_REQUIRED(…)`). I'm a bit hesitant to just add `EXCLUSIVE_LOCKS_REQUIRED(…)` to match the `AssertLockHeld(…)` (which would be trivial) since that methodology hides the fact that the underlying `GUARDED_BY(…)` hasn't been identified.

practicalswift:
The code base contains 37 locks and in the process of documenting them I compiled this list of the initial creators of each lock. These lock introducers would likely be the best possible reviewers for the `GUARDED_BY(…)`:s in this PR for "their" respective locks :-)

Locks introduced by currently active core devs:
* @TheBlueMatt (10 locks): `cs_filter, cs_args, cs_vSend, cs_most_recent_block, cs_callbacks_pending (previously m_cs_callbacks_pending), cs_vAddedNodes, cs_SubVer, cs_addrName, cs_addrLocal, cs_sendProcessing`
* @sipa (8 locks): `cs_KeyStore, cs_addrMan (previously cs), cs_mapLocalHost, cs_LastBlockFile, cs_vOneShots, cs_pathCached (previously csPathCached), cs_nTimeOffset, cs_nBlockSequenceId`
* @theuni (3 locks): `cs_hSocket, cs_vRecv, cs_vProcessMsg`
* @morcos (2 locks): `cs_feeEstimator, cs_feeFilter`
* @gmaxwell (1 lock): `cs_warnings`

Other lock introducers:
* Locks included in the first commit in the repo (Satoshi locks!): `cs_main, cs_vNodes, cs_db, cs_inventory, cs_Shutdown`
* gavinandresen: `cs_wallet, cs_txMemPool (previously cs), cs_setBanned`
* sje397: `cs_totalBytesSent, cs_totalBytesRecv`
* Clark Gaebel: `cs_test (previously cs)`
* domob1812: `cs_rpcWarmup`
* Philip Kaufmann: `cs_proxyInfos`

practicalswift:
I'm now down to only four remaining `AssertLockHeld(…)`:s for which I'm to identify which specific variable the lock in question is being a guard for.

These are the four cases I need help with:

* [[validation.cpp] GetBlockScriptFlags(...)](https://github.com/bitcoin/bitcoin/blob/ea729d55b4dbd17a53ced474a8457d4759cfb5a5/src/validation.cpp#L1588-L1589) – `AssertLockHeld(cs_main);`
* [[validation.cpp] AcceptBlockHeader(...)](https://github.com/bitcoin/bitcoin/blob/ea729d55b4dbd17a53ced474a8457d4759cfb5a5/src/validation.cpp#L3030-L3032) – `AssertLockHeld(cs_main);`
* [[wallet/wallet.cpp] CWallet::RemoveWatchOnly(...)](https://github.com/bitcoin/bitcoin/blob/ea729d55b4dbd17a53ced474a8457d4759cfb5a5/src/wallet/wallet.cpp#L335-L337) – `AssertLockHeld(cs_wallet);`
* [[wallet/wallet.cpp] CWallet::RescanFromTime(...)](https://github.com/bitcoin/bitcoin/blob/ea729d55b4dbd17a53ced474a8457d4759cfb5a5/src/wallet/wallet.cpp#L1530-L1533) – `AssertLockHeld(cs_wallet);`

For each of the above cases I see two alternatives:
1. A `GUARDED_BY(…)` for some variable accessed directly or indirectly is missing.
2. The lock is not technically necessary.

practicalswift:
I've now documented the five places where I'm opting out of thread-safety analysis for various reasons (using the `NO_THREAD_SAFETY_ANALYSIS` annotation): b72ad1f43d96b6f63e7a4046f31af085da4e86db

Please let me know if there is a way to remove any of these opt-outs.

TheBlueMatt:
@practicalswift 

* GetBlockScriptFlags() needs cs_main for versionbitscache (maybe others, but probably just that?)
* AcceptBlockHeader is mostly for mapBlockIndex it looks like
* RemoveWatchOnly looks like it may need to be an atomic operation, but I haven't dug into it too much. It may very well be the case that you'd miss a NotifyWatchonlyChanged fire if you have an AddWatchOnly and a RemoveWatchOnly call at the same time (though I don't know if there are any negative effects from doing so).
* RescanFromTime indeed does not appear to require cs_wallet (obviously ScanForWalletTransactions does, but it takes it itself). It obviously needs cs_main, though.

practicalswift:
@TheBlueMatt Thanks! I'll look into the suggested fixes.

practicalswift:
@jonasschnelli @TheBlueMatt In `CDB::PeriodicFlush(CWalletDBWrapper& dbw)` ([code](https://github.com/bitcoin/bitcoin/blob/39ae41389a8694ec1ab9059117ed2cdd2003d0d3/src/wallet/db.cpp#L604-L612)) we are currently trying to lock `bitdb.cs_db` – shouldn't that lock be on `env->cs_db` instead (`CDBEnv *env = dbw.env`)? Or am I reading it wrong? :-)

practicalswift:
@laanwj, after some digging I found your `bitdb` reference cleanup commit be9e1a968debbb7ede8ed50e9288a62ff15d1e1e, perhaps you know the answer to the question about the `bitdb.cs_db` lock above? :-)

practicalswift:
Rebased!

ryanofsky:
This change is great, but I might suggest breaking off and grouping the ADD_LOCK commits into a few separate PRs. That way the runtime changes can be more easily and carefully vetted, and this PR can exclusively focus on adding compiler annotations and also be more easily reviewed.

practicalswift:
@ryanofsky Very good point! I'll do that! Thanks.

practicalswift:
@ryanofsky One slight problem is that this PR won't compile cleanly unless the locks are added. But you thinking was that the individual "add locks"-PR should be merged before this PR gets merged? That works for me :-)

ryanofsky:
Yeah, I was thinking there would be a handful of other prs which this would depend on since the LOCKs are in different parts of the code, and different people might want to review them. Those PRs would all have their own branches, and the branch for this PR could just `git merge` in those branches until they get merged to master.

practicalswift:
@ryanofsky Excellent! That'll be our plan!

TheBlueMatt:
Alright, with the depends, it seems like time to start breaking this up, no? There is no way this is going to get merged in one go, though its been great to see this keep getting rebased since it has caught a few errors already :).

ryanofsky:
Pretty sure this is already happening. There have been a bunch of new prs adding locking. As they are merged, the "ADD LOCK" commits here can go away and this will shrink, only adding annotations without changing behaviour.

TheBlueMatt:
@ryanofsky I was referring to adding the annotations - as far as I've seen the only @practicalswift PRs in the locking space have been adding missing locks detected by the annotations, not actually adding the annotations.

ryanofsky:
Since annotations don't change behaviour, it seems to me that having one PR that *just* adds annotations is fine and probably easier to review than lots of different PRs adding annotations.

Adding new locks is much more dicey, and changes that add new locks should be smaller and get more careful review. This is happening now with #11634, #11623, #11596, etc, and I'm assuming there will be more.

@practicalswift, you may want to add WIP / Work in progress to the description here to indicate that this is currently a work in progress, and that this PR will eventually only add annotations and not change locking or other runtime behavior, since right now it is doing a little of everything.

TheBlueMatt:
I tend to disagree - if we add annotations and get them wrong we're
likely to fool ourselves in the future on locking changes - making sure
that a function which doesn't access variables which need locks but
which needs results from two other function calls to be consistent has
the appropriate annotation may end up pretty important. Even better, we
should take this as an opportunity to fix obviously bad locking and
reduce the coverage of locks when there are easy places to do so!

practicalswift:
@ryanofsky @TheBlueMatt What about doing both: I start by submitting the `LOCK(…)`:s as separate PR:s and when these are merged I continue by submitting the `GUARDED_BY(…)`/`EXCLUSIVE_LOCKS_REQUIRED(…)` in separate PR:s. All while keeping this PR updated with both types of changes. Sounds good?

Regarding the latter, should I submit the annotation PR:s on a per-lock basis (i.e. one PR with all `cs_wallet` annotations, etc.)?

MarcoFalke:
Needs rebase.
For the purpose of easing review and speeding up merge: What about submitting a subset of this that only renames the comments `// protected by ${cs}` to `GUARDED_BY(${cs})`. Assuming that the existing comments are reviewed such a subset should be refactoring-only and easier to merge.

practicalswift:
@MarcoFalke Do you mean a PR with all `GUARDED_BY(…)`:s in this PR added instead as comments `// GUARDED_BY(…)`? Sure, confirm and I'll fix that.

Regarding rebasing – I suggest we merge (or close in the case of NACK) the locking related [PR:s I currently have open](https://github.com/bitcoin/bitcoin/pulls/practicalswift) and I'll rebase this PR on top of `master` after that. Makes sense?

MarcoFalke:
Something like e7f4866d499c4f1a5c17fa140be090da7c940c86, but at this point I am not sure if it is going to help review a lot.

practicalswift:
Closing. Will split this into subsets and submit as separate PR:s.

practicalswift:
To facilitate reviewing the most important parts of this thread-safety PR has been submitted as individual PRs:

* #12665 – Add compile time checking for all run time locking assertions
* #11795 – Avoid locking cs_vNodes twice when calling FindNode(...)
* #11762 – Avoid locking mutexes that are already held by the same thread
* #11694 – Add missing cs_main lock in getblocktemplate(...)
* #11689 – Fix missing locking in CTxMemPool::check(…) and CTxMemPool::setSanityCheck(…)
* #11652 – Add missing locks to init.cpp (in AppInitMain + ThreadImport) and validation.cpp
* #11634 – Add missing cs_wallet/cs_KeyStore locks to wallet
* #11617 – Call FlushStateToDisk(...) regardless of fCheckForPruning
* #11596 – Add missing cs_main locks when accessing chainActive

Reviews of these PR:s would be really appreciated! :-)

I'll tackle the remaining parts once the above PR:s have been merged or closed. Complete thread safety annotation coverage is not far away now! :-)

PR #11227, opened 2017-09-03
WIP: switch to libevent for node socket handling

theuni:
Not yet ready for review. This can be considered a staging area. Chunks of this will be PR'd until only the actual libevent switch-over is ready, at which point this PR should be ready for review.

These changes remove our old select() loop for socket handling in favor of libevent, which uses epoll/kqueue/etc. as a back-end. In addition to being faster and more efficient, this allows us to drop some annoying restrictions, namely that select can only handle 1024 sockets in practice on most systems.

Note that this does _not_ yet make the switch to libevent for outgoing connections, that work is happening in parallel, and will be easier to merge after this.

Also, for any reviewers, several of these commits would individually introduce some regressions or slow-downs, but they've been split up in order to clarify why some of the changes are being made.

Depends on:
- [x] #10663 
- [x] #11363

Todo:
- Add a ton of documentation
- RAII the libevent structures
- Add some tests where possible

### DIFF
--- a/src/net.cpp
+++ b/src/net.cpp
@@ -511,7 +511,7 @@ void CConnman::DumpBanlist()
 
 void CNode::CloseSocketDisconnect()
 {
-    fDisconnect = true;
+    Disconnect();
     LOCK(cs_hSocket);
     if (hSocket != INVALID_SOCKET)
     {
@@ -591,7 +591,7 @@ void CConnman::Ban(const CSubNet& subNet, const BanReason &banReason, int64_t ba
         LOCK(cs_vNodes);
         for (CNode* pnode : vNodes) {
             if (subNet.Match((CNetAddr)pnode->addr))
-                pnode->fDisconnect = true;
+                pnode->Disconnect();
         }
     }
     if(banReason == BanReasonManuallyAdded)
@@ -1093,7 +1093,7 @@ bool CConnman::AttemptToEvictConnection()
     LOCK(cs_vNodes);
     for(std::vector<CNode*>::const_iterator it(vNodes.begin()); it != vNodes.end(); ++it) {
         if ((*it)->GetId() == evicted) {
-            (*it)->fDisconnect = true;
+            (*it)->Disconnect();
             return true;
         }
     }
@@ -1444,27 +1444,27 @@ void CConnman::ThreadSocketHandler()
                 if (pnode->nLastRecv == 0 || pnode->nLastSend == 0)
                 {
                     LogPrint(BCLog::NET, "socket no message in first 60 seconds, %d %d from %d\n", pnode->nLastRecv != 0, pnode->nLastSend != 0, pnode->GetId());
-                    pnode->fDisconnect = true;
+                    pnode->Disconnect();
                 }
                 else if (nTime - pnode->nLastSend > TIMEOUT_INTERVAL)
                 {
                     LogPrintf("socket sending timeout: %is\n", nTime - pnode->nLastSend);
-                    pnode->fDisconnect = true;
+                    pnode->Disconnect();
                 }
                 else if (nTime - pnode->nLastRecv > (pnode->nVersion > BIP0031_VERSION ? TIMEOUT_INTERVAL : 90*60))
                 {
                     LogPrintf("socket receive timeout: %is\n", nTime - pnode->nLastRecv);
-                    pnode->fDisconnect = true;
+                    pnode->Disconnect();
                 }
                 else if (pnode->nPingNonceSent && pnode->nPingUsecStart + TIMEOUT_INTERVAL * 1000000 < GetTimeMicros())
                 {
                     LogPrintf("ping timeout: %fs\n", 0.000001 * (GetTimeMicros() - pnode->nPingUsecStart));
-                    pnode->fDisconnect = true;
+                    pnode->Disconnect();
                 }
                 else if (!pnode->fSuccessfullyConnected)
                 {
                     LogPrintf("version handshake timeout from %d\n", pnode->GetId());
-                    pnode->fDisconnect = true;
+                    pnode->Disconnect();
                 }
             }
         }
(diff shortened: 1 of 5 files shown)


theuni:
I believe that the proxy_test failure is a false-positive.

Looks like the issue is that disconnection happens so quickly now after an error that the testnode fails.

src/net.cpp
```diff
@@ -2397,6 +2370,9 @@ bool CConnman::Start(CScheduler& scheduler, const Options& connOptions)
         LOCK(m_cs_interrupt_event);
         m_interrupt_event = event_new(m_event_base, -1, 0, [](evutil_socket_t, short, void* data) {static_cast<CConnman*>(data)->InterruptEvents();}, (this));
     }
+    m_event_timeout = timeval{60, 0};
+    const timeval* common_timeout = event_base_init_common_timeout(m_event_base, &m_event_timeout);
+    memcpy(&m_event_timeout, common_timeout, sizeof(m_event_timeout));
```
sipa: The documentation on `event_base_init_common_timeout` says "To do this, call this function with the common duration. It will return a pointer to a different, opaque timeout value. (Don't depend on its actual contents!)".

I'm not sure whether that means you need to use the exact pointer returned, or just a copy of its value (as this code does).

src/net.cpp
```diff
@@ -456,7 +471,14 @@ void CConnman::DumpBanlist()
 
 void CNode::CloseSocketDisconnect()
 {
-    Disconnect();
+    if (m_write_event) {
```
sipa: I would prefer it if you could have some RAII wrappers around the libevent structures before adding all this.

src/net.cpp
```diff
@@ -454,7 +454,7 @@ void CConnman::DumpBanlist()
 
 void CNode::CloseSocketDisconnect()
 {
-    fDisconnect = true;
+    Disconnect();
```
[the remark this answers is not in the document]
theuni: Probably so, will give it a shot.

src/net.cpp
```diff
@@ -914,12 +914,7 @@ size_t CConnman::SocketSendData(CNode *pnode) const
     while (it != pnode->vSendMsg.end()) {
         const auto &data = *it;
         assert(data.size() > pnode->nSendOffset);
-        int nBytes = 0;
-        {
-            if (pnode->hSocket == INVALID_SOCKET)
```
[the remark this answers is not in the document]
theuni: I believe all of these are gone by the end of the change set, the removal may make more sense in an earlier commit, though.

src/net.cpp
theuni: Agreed, see PR description :)

src/net.cpp
```diff
@@ -1205,11 +1302,14 @@ void CConnman::AcceptConnection(const SOCKET& listen_socket, bool whitelisted) {
         return;
     }
 
-    if (!IsSelectableSocket(hSocket))
-    {
-        LogPrintf("connection from %s dropped: non-selectable socket\n", addr.ToString());
-        CloseSocket(hSocket);
-        return;
+    const char* method = event_base_get_method(m_event_base);
+    if (!method || (strncmp(method, "select", 6) == 0)) {
```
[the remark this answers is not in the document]
theuni: It's only needed if select ends up being used, which should be very unlikely on most platforms. I'll add a comment here, and print something to the log so that it's clear which back-end is being used.

src/net.cpp
theuni: Good eye! I wasn't sure how to store this properly either. This usage is taken from the official documentation. I'll add a comment about its sanity and a link to the docs.

laanwj:
Concept ACK, woohoo!

src/net.cpp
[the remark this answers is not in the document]
laanwj: We could just bail out if select is being used. But not sure what platforms would be affected.

src/net.cpp
```diff
@@ -841,36 +872,126 @@ const uint256& CNetMessage::GetMessageHash() const
...
+            pnode->vProcessMsg.splice(pnode->vProcessMsg.end(), pnode->vRecvMsg, pnode->vRecvMsg.begin(), it);
+            pnode->nProcessQueueSize += nSizeAdded;
+            if (pnode->nProcessQueueSize > nReceiveFloodSize) {
+                LogPrint(BCLog::NET, "Disabling receive due to oversized process queue of size: %lu. peer=%i\n", pnode->nProcessQueueSize, pnode->GetId());
+                pnode->fPauseRecv = true;
+                event_del(pnode->m_read_event);
+            }
 
+        }
+        WakeMessageHandler();
```
jimpo: Should this be moved into the above `if` block? (It used to be inside the `if (notify)` statement)

src/net.cpp
```diff
@@ -1150,7 +1150,7 @@ void CConnman::AcceptConnection(const ListenSocket& hListenSocket) {
         }
     }
 
-    bool whitelisted = hListenSocket.whitelisted || IsWhitelistedRange(addr);
+    whitelisted = whitelisted || IsWhitelistedRange(addr);
```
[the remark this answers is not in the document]
sipa: > ||=

I don't think such an operator exists in C++.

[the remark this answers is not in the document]
jimpo: Oops, you're right. Should have checked.

src/net.cpp
theuni: Yes, thank you.

src/net.cpp
theuni: I can say for certain that the operator doesn't exist because I tried it first :)

laanwj: Hehe I also remember it doesn't exist - I looked it up a long time ago and IIRC the rationale is that's because `|=` is the same when using only booleans. In any case this is ok as it is.

promag:
Needs rebase.

fanquake:
Removing "Up for grabs", given we are now in the process of trying to remove libevent.

PR #11230, opened 2017-09-04
[tests] fixup dbcrash interaction with add_nodes()

jnewbery:
Another conflict with #11121. Apologies - this is entirely my fault. I didn't run the extended test suite after rebasing on master.

@MarcoFalke @sdaftuar

### DIFF
--- a/test/functional/dbcrash.py
+++ b/test/functional/dbcrash.py
@@ -64,7 +64,7 @@ class ChainstateWriteCrashTest(BitcoinTestFramework):
 
     def setup_network(self):
         # Need a bit of extra time for the nodes to start up for this test
-        self.add_nodes(self.num_nodes, timewait=90)
+        self.add_nodes(self.num_nodes, extra_args=self.extra_args, timewait=90)
         self.start_nodes()
         # Leave them unconnected, we'll use submitblock directly in this test
 


test/functional/dbcrash.py
```diff
@@ -64,7 +64,7 @@ def set_test_params(self):
 
     def setup_network(self):
         # Need a bit of extra time for the nodes to start up for this test
-        self.add_nodes(self.num_nodes, timewait=90)
+        self.add_nodes(self.num_nodes, extra_args=self.extra_args, timewait=90)
```
MarcoFalke: We might as well get rid of passing member variables to methods.

maflcko (commented):
utACK 8fdb6f9126fe1e13b60de11fd34c204247679ef4

sdaftuar:
ACK, this fixes `dbcrash.py` now

test/functional/dbcrash.py
jnewbery: Agree, but I haven't looked into how big a change that would be (I'd need to look at all calls to `add_nodes()` to make sure they're not calling this method with no `extra_args` method and expecting the node to be started with any `extra_args` equal to `[]`)

PR #11231, opened 2017-09-04
Improve netaddress implementation

danra:
Making most of the implementation simpler and safer.

### DIFF
--- a/src/netaddress.cpp
+++ b/src/netaddress.cpp
@@ -110,7 +110,7 @@ bool CNetAddr::IsIPv4() const
 
 bool CNetAddr::IsIPv6() const
 {
-    return (!IsIPv4() && !IsTor() && !IsInternal());
+    return !IsIPv4() && !IsTor() && !IsInternal();
 }
 
 bool CNetAddr::IsRFC1918() const
@@ -118,7 +118,7 @@ bool CNetAddr::IsRFC1918() const
     return IsIPv4() && (
         GetByte(3) == 10 ||
         (GetByte(3) == 192 && GetByte(2) == 168) ||
-        (GetByte(3) == 172 && (GetByte(2) >= 16 && GetByte(2) <= 31)));
+        (GetByte(3) == 172 && GetByte(2) >= 16 && GetByte(2) <= 31));
 }
 
 bool CNetAddr::IsRFC2544() const
@@ -128,7 +128,7 @@ bool CNetAddr::IsRFC2544() const
 
 bool CNetAddr::IsRFC3927() const
 {
-    return IsIPv4() && (GetByte(3) == 169 && GetByte(2) == 254);
+    return IsIPv4() && GetByte(3) == 169 && GetByte(2) == 254;
 }
 
 bool CNetAddr::IsRFC6598() const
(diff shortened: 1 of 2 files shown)


gmaxwell:
please don't add another boost dependency for a simple equality comparison.

danra:
@gmaxwell There is already a dependency on this boost library in other places in the code. It doesn't have a standard equivalent, and it makes the code clearer here.

gmaxwell:
@danra we are working on eliminating boost, and adding the additional header here visibly slows down the compile.  If it were for something more complex than a match there might be an argument.

danra:
@gmaxwell If you don't mind, I'd be happy to discuss this a bit, since I think the approach of just "eliminating boost" is wrong.

First of all, I don't that goal listed under the github "Projects" - I only see a goal of transitioning from boost to C++11. If removing boost in general from the project is a goal, even for cases where there is no standard substitute, I suppose it should be listed under "Projects" as well.

Second, I think a reasonable compilation time increase is worth improving code clarity. Consider the "endswith" functionality here:
- checking it inline instead of using a function requires a couple of if conditions, making it less readable and more error prone, which also means harder to review and less secure. One reading the code has only so much of an attention span, and it's bad to force details which are unnecessary in the context - implementing "endswith" inline is certainly out of the scope of the code where it is used.
- Writing our own function is for the bitcoin repo in some utility header is fine, but it means a) using less tested code than the one boost uses and b) just more work which could be spent on more useful things.

danra:
@gmaxwell Do you have a specific suggestion for an elegant one-liner? I don't see one which doesn't involve also checking the size, or doing some min() with it

sipa:
I don't see anything wrong with 

```c++
template<typename I>
bool beginswith(const T& a, const T& b) { return a.size() >= b.size() && std::equal(b.begin(), b.end(), a.begin()); }
```

MeshCollider:
@danra It's written on there: `The long term future goal is to remove Boost as a dependency.`

jimpo:
Should this be using `.size()` everywhere on the `std::array`s instead of `sizeof`?

danra:
@MeshCollider You're right. I think it's a bad goal, though (unlike boost->C++11 migration under which it's listed). For instance, is Bitcoin Core really interested in rewriting boost::multi_index?

danra:
@jimpo I thought `sizeof` would be the same (and compile-time evaluable) since std::array is an aggregate time, but it seems [that is not guaranteed](https://stackoverflow.com/questions/19103244/is-the-size-of-stdarray-defined-by-standard). Will fix, thanks!

danra:
@sipa It's not wrong, but this sort of utility functions tend to pile up, and I think it's better to use well-tested (e.g. boost provided) utility functions where it's available, rather than maintain them yourself, even if it costs at some compilation time.

sipa:
@danra That's a fair point, though I personally think that a boost dependency is worse than the cost of maintaining such a simple function.

danra:
@gmaxwell @sipa Reverted boost dependency, due to both your review and the fact that using the boost function changed the semantics slightly - the ".onion" address (with no preceding characters) was previously rejected immediately by the condition, which is probably safer than the situation after my modification which resulted in passing an empty string to `DecodeBase32` (that function *should* handle it correctly, but it's probably better to code defensively)

@jimpo Fixed sizeof->size() (hopefully that would also fix onion address-related tests failing on some platforms :)

src/netaddress.cpp
```diff
@@ -291,30 +303,31 @@ std::string CNetAddr::ToString() const
...
 bool operator==(const CNetAddr& a, const CNetAddr& b)
 {
-    return (memcmp(a.ip, b.ip, 16) == 0);
+    return a.ip == b.ip;
 }
 
 bool operator!=(const CNetAddr& a, const CNetAddr& b)
 {
-    return (memcmp(a.ip, b.ip, 16) != 0);
+    return !(a==b);
```
promag: Nit, add spaces `!(a == b)`.

laanwj:
Closing due to inactivity, let me know if you start work on this again and want to re-open this.

PR #11234, opened 2017-09-05
Remove redundant testutil.cpp|h files

meshcollider:
The only function in testutil.cpp, `GetTempPath()` simply called `fs::temp_directory_path()` directly. This just tidies things up by removing that redundant function and the file containing it

I can understand wanting a general util file for tests to use, but if there's nothing in it, we might as well remove it, it can always be added back later when it's put to use.


dcousens (approved):

practicalswift:
utACK d1138e36208dac87a27f415f39c70377636e93f1. Nice cleanup!

laanwj:
> I can understand wanting a general util file for tests to use

That's why I kept it around, but yeah, if someone wants a test utility cpp/h they can re-add it.

MarcoFalke:
utACK d1138e36208dac87a27f415f39c70377636e93f1. The file collapsed after removal of boost.

PR #11235, opened 2017-09-05
Trivial: Fix a grammar mistake and a typo in the french translations

aeyoll:

PR #11236, opened 2017-09-05
Add note on translations to CONTRIBUTING.md

meshcollider:
The readme mentions it but there is no note in contributing.md, so this just adds a note there as suggested by https://github.com/bitcoin/bitcoin/pull/11235#issuecomment-327110644


laanwj:
Yes, makes sense, utACK

PR #11237, opened 2017-09-05
qt: Fixing division by zero in time remaining

### LINKED ISSUE #10291: crazy 'estimated time left until synced' estimate
Running "Bitcoin version v0.14.0.0-b0c9b33" when it first started up I saw this:

screenshot - 17-04-27 - 03 17 02 pm

1. The estimated time is way too high. It will only take a few hours to sync the remaining blocks.

2. The estimation is shown as a huge number of seconds instead of converting it to "48 years". Later, the estimation changed to "3 hours", so the code is capable of doing such conversions, but in this case it didn't.

meshcollider:
Fixes https://github.com/bitcoin/bitcoin/issues/10291, https://github.com/bitcoin/bitcoin/issues/11265

progressDelta may be 0 (or even negative according to 11265), this checks for that and prints unknown if it is, because we cannot calculate an estimate for the time remaining (would be infinite or negative).

### DIFF
--- a/src/qt/modaloverlay.cpp
+++ b/src/qt/modaloverlay.cpp
@@ -99,15 +99,18 @@ void ModalOverlay::tipUpdate(int count, const QDateTime& blockDate, double nVeri
                 progressDelta = progressStart-sample.second;
                 timeDelta = blockProcessTime[0].first - sample.first;
                 progressPerHour = progressDelta/(double)timeDelta*1000*3600;
-                remainingMSecs = remainingProgress / progressDelta * timeDelta;
+                remainingMSecs = (progressDelta > 0) ? remainingProgress / progressDelta * timeDelta : -1;
                 break;
             }
         }
         // show progress increase per hour
         ui->progressIncreasePerH->setText(QString::number(progressPerHour*100, 'f', 2)+"%");
 
-        // show expected remaining time
-        ui->expectedTimeLeft->setText(GUIUtil::formatNiceTimeOffset(remainingMSecs/1000.0));
+        if(remainingMSecs >= 0) {	
+            ui->expectedTimeLeft->setText(GUIUtil::formatNiceTimeOffset(remainingMSecs / 1000.0));
+        } else {
+            ui->expectedTimeLeft->setText(QObject::tr("unknown"));
+        }
 
         static const int MAX_SAMPLES = 5000;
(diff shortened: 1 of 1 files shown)


jonasschnelli:
utACK a676f81148cfcf57ea27effb3b69aef78b497994

MarcoFalke:
utACK a676f81

src/qt/modaloverlay.cpp
```diff
@@ -82,36 +82,38 @@ void ModalOverlay::tipUpdate(int count, const QDateTime& blockDate, double nVeri
...
             QPair<qint64, double> sample = blockProcessTime[i];
 
             // take first sample after 500 seconds or last available one
             if (sample.first < (currentDate.toMSecsSinceEpoch() - 500 * 1000) || i == blockProcessTime.size() - 1) {
-                progressDelta = progressStart-sample.second;
+                progressDelta = blockProcessTime[0].second - sample.second;
                 timeDelta = blockProcessTime[0].first - sample.first;
-                progressPerHour = progressDelta/(double)timeDelta*1000*3600;
-                remainingMSecs = remainingProgress / progressDelta * timeDelta;
+                progressPerHour = progressDelta / (double) timeDelta * 1000 * 3600;
```
laanwj: Please don't combine refactoring and a bugfix in one commit. This makes it extremely hard to review what is changed. If you want to do this in one PR, that's okay, but please do it as two separate commits.

MeshCollider:
Split into 2 commits, thanks @laanwj

src/qt/modaloverlay.cpp
```diff
@@ -99,15 +99,18 @@ void ModalOverlay::tipUpdate(int count, const QDateTime& blockDate, double nVeri
...
                 progressPerHour = progressDelta/(double)timeDelta*1000*3600;
-                remainingMSecs = remainingProgress / progressDelta * timeDelta;
+                remainingMSecs = (progressDelta > 0) ? remainingProgress / progressDelta * timeDelta : -1;
                 break;
             }
         }
         // show progress increase per hour
         ui->progressIncreasePerH->setText(QString::number(progressPerHour*100, 'f', 2)+"%");
 
-        // show expected remaining time
```
laanwj: Did you remove this comment intentionally?

MeshCollider: Ah must have just been a bad commit split, sorry about that

PR #11238, opened 2017-09-05
Add assertions before potential null deferences

meshcollider:
Picked up by the static analyzer [Facebook Infer](http://fbinfer.com/) which I was playing around with for another research project. Just adding some asserts before dereferencing potentially null pointers.

### DIFF
--- a/src/miner.cpp
+++ b/src/miner.cpp
@@ -145,6 +145,7 @@ std::unique_ptr<CBlockTemplate> BlockAssembler::CreateNewBlock(const CScript& sc
 
     LOCK2(cs_main, mempool.cs);
     CBlockIndex* pindexPrev = chainActive.Tip();
+    assert(pindexPrev != nullptr);
     nHeight = pindexPrev->nHeight + 1;
 
     pblock->nVersion = ComputeBlockVersion(pindexPrev, chainparams.GetConsensus());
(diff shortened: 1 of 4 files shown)


JeremyRubin:
Seems reasonable, utAck.

src/addrdb.cpp
```diff
@@ -49,7 +49,10 @@ bool SerializeFileDB(const std::string& prefix, const fs::path& path, const Data
         return error("%s: Failed to open file %s", __func__, pathTmp.string());
 
     // Serialize
-    if (!SerializeDB(fileout, data)) return false;
+    if (!SerializeDB(fileout, data)) {
+        fileout.fclose();
```
laanwj: This is an autofile - doesn't it close automatically on release?

promag: Maybe remove explicit close and move `fileout` to an inner scope?

practicalswift:
utACK modulo the autofile fix which is redundant.

Generally I think it is a good very idea to add `assert(…)`:s where it help static analyzers understand the code better. It makes implicit assumptions explicit which helps static analyzers and humans alike to understand and reason about the code.

My experience is that if a code path makes a static analyzer confused it is very likely to be confusing for humans too. This holds even in a lot of cases where it could be argued that the static analyzer should be "smarter" and in some way misunderstands the code. Confusion leads to bugs, so better avoid confusion by liberal use of `assert`:s and other mechanisms to make tacit assumptions explicit.

MeshCollider:
Removed redundant file close as requested, thanks for the reviews :)

practicalswift:
utACK c0019924406e1ce8368465c768de11019ad5eeed

promag:
utACK c001992.

laanwj:
Can you please change the title to `Add nullptr assertions` or such? The current title is incorrect, as adding assertions does not *fix* anything.

laanwj:
utACK c00199244

PR #11239, opened 2017-09-05
Gauge double

jlopp:
Default gauge function truncates any precision; this is needed for the chainstats txrate metric.


jlopp:
bah, I have to fix this default PR branch problem.

PR #11245, opened 2017-09-05
[0.15] Mention offscreen issue in release notes

meshcollider:
@gmaxwell suggested adding the off-screen GUI issue as a known bug to the 0.15.0 release notes including the fix

### DIFF
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -50,7 +50,9 @@ Known Issues
 There have been a number of reports of the Bitcoin-Qt GUI on Windows being spawned off-screen,
 causing it not to appear (See [Issue 11207](https://github.com/bitcoin/bitcoin/issues/11207)).
 Users can fix this problem by re-running bitcoin-qt with the `-resetguisettings` parameter, or
-using Windows keyboard shortcuts to move the GUI back on screen. This will be fixed in 0.15.1.
+using Windows keyboard shortcuts to move the GUI back on screen. Select the window from the task
+bar, type Alt+Space on the keyboard, and select Move from the menu. You can then use the arrow keys
+to move the window back onto the screen. This will be fixed in 0.15.1.
 
 Compatibility
 ==============
--- a/doc/release-notes.md
+++ b/doc/release-notes.md
@@ -44,6 +44,14 @@ option to rebuild the chainstate data structures in the old format.
 If your node has pruning enabled, this will entail re-downloading and
 processing the entire blockchain.
 
+Known Issues
+------------
+
+There have been a number of reports of the Bitcoin-Qt GUI on Windows being spawned off-screen,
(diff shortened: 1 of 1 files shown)


laanwj:
Thanks, ACK

jonasschnelli:
Related to #11207 #11208

doc/release-notes.md
```diff
@@ -44,6 +44,14 @@ option to rebuild the chainstate data structures in the old format.
 If your node has pruning enabled, this will entail re-downloading and
 processing the entire blockchain.
 
+Known Issues
+------------
+
+There have been a number of reports of the Bitcoin-Qt GUI on Windows being spawned off-screen,
+causing it not to appear (See [Issue 11207](https://github.com/bitcoin/bitcoin/issues/11207)).
+Users can fix this problem by re-running bitcoin-qt with the `-resetguisettings` parameter, or
+using Windows keyboard shortcuts to move the GUI back on screen. This will be fixed in 0.15.1.
```
MarcoFalke: Can you explain what "Windows keyboard shortcuts" mean?

MarcoFalke:
Concept ACK

doc/release-notes.md
MeshCollider: I didn't want to make it a long tutorial on fixing the issue, the keyboard shortcut was given as a comment on the issue I linked, I can add it in though if you like?

jonasschnelli:
This issue is gone somewhere between 0.14.1 and 0.15.0rc2 (could be a Qt upstream fix).
Not sure if we should add this... the issue may appear for self-compiled Bitcoin-Qt <Qt5.7.

luke-jr:
I thought the issue was a new regression in 0.15 rcs?

`-resetguisettings` clears *all* the GUI settings - not an acceptable workaround IMO.

MarcoFalke:
> I thought the issue was a new regression in 0.15 rcs?

The issue comes back every major release, in a slightly different form each time probably.

gmaxwell:
I've been seeing a LOT of people reporting it on 0.15rc including RC2 and RC3 but I have had no reason to believe it was newly introduced.  @jonasschnelli Why do you believe that it is gone?

jonasschnelli:
@gmaxwell: I just tested 0.14.1 and 0.15.0rc2 on Windows 7, 8.1 and 10. And 0.14.1 has the issue (when moving the window out of the screen boundaries with RedEdit) while 0.15.0rc2 does not have the issue. My assumption then is, that either it's gone by a Qt upstream fix or by a change within our codebase.

laanwj:
Looks like even the original reporter cannot reproduce it anymore: https://github.com/bitcoin/bitcoin/issues/11207#issuecomment-327312446

MeshCollider:
If this is no longer an issue, feel free to close this PR

MarcoFalke:
Should probably be closed if this is no longer an issue

laanwj:
Closing for now.

PR #11256, opened 2017-09-06
RPC: add weight to mempool entry output

esotericnonsense:
Tested against master using the REST api (/rest/mempool/contents), simple addition of a field.

Personal use case is for fee analysis software.

### DIFF
--- a/test/functional/segwit.py
+++ b/test/functional/segwit.py
@@ -240,45 +240,45 @@ class SegWitTest(BitcoinTestFramework):
         #                      tx2 (segwit input, paying to a non-segwit output) ->
         #                      tx3 (non-segwit input, paying to a non-segwit output).
         # tx1 is allowed to appear in the block, but no others.
-        txid1 = send_to_witness(1, self.nodes[0], find_unspent(self.nodes[0], 50), self.pubkey[0], False, Decimal("49.996"))
-        hex_tx = self.nodes[0].gettransaction(txid)['hex']
-        tx = FromHex(CTransaction(), hex_tx)
+        tx_hex = self.nodes[0].gettransaction(txid)['hex']
+        tx = FromHex(CTransaction(), tx_hex)
         assert(tx.wit.is_null()) # This should not be a segwit input
-        assert(txid1 in self.nodes[0].getrawmempool())
 
-        hex_tx = self.nodes[0].gettransaction(txid1)['hex']
-        tx = FromHex(CTransaction(), hex_tx)
+        txid1 = send_to_witness(1, self.nodes[0], find_unspent(self.nodes[0], 50), self.pubkey[0], False, Decimal("49.996"))
+        assert(txid1 in self.nodes[0].getrawmempool())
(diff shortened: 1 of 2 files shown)


promag:
No test affected 😞 care to improve by asserting the new field in the relevant RPC's 😉?

laanwj:
Concept ACK

MarcoFalke:
I guess you need to update the documentation as well.

esotericnonsense:
Tests are failing.
txid2 and txid3 work as expected.
txid1 fails on both my test and https://github.com/bitcoin/bitcoin/pull/11203 (add wtxid to mempool entry output).

ajtowns:
txid1 is failing because at that point "tx" is actually referring to the input to txid1, not the transaction for txid1. It works fine for me if I add

`         tx = FromHex(CTransaction(), self.nodes[0].gettransaction(txid1)['hex'])
`

prior to the assert_equal lines (and uncomment them obviously).

esotericnonsense:
Doh. You're absolutely right. Fixed.

The final commit 'Refactor segwit 3-tx-chain' changes all references to 'tx' to 'tx/tx1/tx2/tx3' in order to clarify that. It has a large diff and can be dropped if necessary (only affects code style).

morcos:
Concept ACK, but it turns out GetTxSize is not actually what we claim it is.  See the calculation of `GetVirtualTransactionSize` which potentially includes number of sig ops in the calculation.

I thinke the right path forward is: 
- a small BIP documenting this usage of virtual transaction size (which is used almost everywhere in our code
- documentation update pointing to this new BIP instead of 141 for defining virtual size
- this PR makes even more sense given this lack of easy conversion, but I think the newly added tests should at least be documented to note that they only work in the case that sig ops don't factor in, and perhaps we should add a test with a tx that has more sigops to show what happens in that case.

luke-jr:
@morcos AFAIK that's exclusively used for node policy, and as such isn't a topic for standardisation...?

test/functional/segwit.py
```diff
@@ -240,28 +240,45 @@ def run_test(self):
...
         assert(txid1 in self.nodes[0].getrawmempool())
+        tx1_hex = self.nodes[0].gettransaction(txid1)['hex']
+        tx1 = FromHex(CTransaction(), tx1_hex)
+
+        # Check that wtxid is properly reported in mempool entry (txid1)
+        assert_equal(int(self.nodes[0].getmempoolentry(txid1)["wtxid"], 16), tx1.calc_sha256(True))
+
+        # Check that weight and sizei (actually vsize) are properly reported in mempool entry (txid1)
+        assert_equal(self.nodes[0].getmempoolentry(txid1)["size"], (self.nodes[0].getmempoolentry(txid1)["weight"] + 3) // 4)
+        assert_equal(self.nodes[0].getmempoolentry(txid1)["weight"], len(tx1.serialize())*3 + len(tx1.serialize_with_witness()))
```
luke-jr: serialize needs to be made serialize_without_witness

MeshCollider:
Closing this, please see #14649

PR #11259, opened 2017-09-06
Remove duplicate destination decoding

promag:
Follow up of #11117, this patch removes an extra unnecessary destination decoding that was identified while reviewing #11117. It is also the only case where `IsValidDestinationString` is called before `DecodeDestination`.

For reference see [comment](https://github.com/bitcoin/bitcoin/pull/11117#discussion_r137145517).

### DIFF
--- a/src/base58.cpp
+++ b/src/base58.cpp
@@ -232,8 +232,6 @@ public:
     CBitcoinAddress(const char* pszAddress) { SetString(pszAddress); }
 
     CTxDestination Get() const;
-    bool GetKeyID(CKeyID &keyID) const;
-    bool IsScript() const;
 };
 
 class CBitcoinAddressVisitor : public boost::static_visitor<bool>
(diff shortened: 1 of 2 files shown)


laanwj:
utACK https://github.com/bitcoin/bitcoin/pull/11259/commits/6bc6e241176b4a3acd1be2c12b7ffa7f1f65bd56

sipa:
utACK 6bc6e241176b4a3acd1be2c12b7ffa7f1f65bd56

MeshCollider:
utACK https://github.com/bitcoin/bitcoin/commit/6bc6e241176b4a3acd1be2c12b7ffa7f1f65bd56

promag:
As per @laanwj suggestion on IRC, added a commit to remove unused methods that caused
```
base58.cpp:308:6: warning: ‘bool {anonymous}::CBitcoinAddress::IsScript() const’ defined but not used [-Wunused-function]
base58.cpp:298:6: warning: ‘bool {anonymous}::CBitcoinAddress::GetKeyID(CKeyID&) const’ defined but not used [-Wunused-function]
```
Also rebased.

sipa:
utACK 8d0041e607a0e5bff52152f3d8af504f3703baa7

laanwj:
Thanks. re-utACK 86e6dd4b6, builds without warnings (except in leveldb, which we can't do anything about here) now.

Issue #11004, opened 2017-08-07
New Travis job for CHECK_DOCS steps


eklitzke:
This is a request to change how the Travis job is configured for Bitcoin. I'd submit a PR to do this myself, but the change can only be made by someone who has Travis configuration permissions for the bitcoin repo.

## Background

Currently there are a number of static-analysis style checks that are run in Travis, guarded by the `CHECK_DOC` environment variable. This variable is only set for the first test runner (the arm-linux-gnueabihf host), so the checks are only run once. All of these checks run very fast, and do not require building Bitcoin or running the regular C++/Python tests.

## Request

Configure a new Travis job that runs against PRs, and only runs the `CHECK_DOC` tests. At the same time, the `CHECK_DOC` stuff in the existing job would be removed.

## Motivation
Creating a separate Travis job for these tests would be an improvement because:
 * It would be more immediately obvious that a `CHECK_DOC` failure is related to one of the check scripts; currently it's confusing to new contributors (e.g. me) to see an ARM failure, when the failure is not actually related to any new architecture-specific code.
 * The job would either pass quickly or fail quickly, giving faster feedback to developers who submit PRs.
 * The main `.travis.yml` file in the repo could be cleaned up, by removing the `CHECK_DOC` guards in the `before_script` and `script` sections.

MarcoFalke:
There is some overhead in booting and setting up the vm, so I thought it would be fine to misuse the first entry in the build matrix.

eklitzke:
I will defer to you of course, but I've found this to be helpful in other projects. As an example, suppose it takes one minute on average to launch a new Travis worker, and the CHECK_DOC tests can run in a few seconds (or at least, less than a minute). Then if someone submits a PR there's a good chance that they'll see a red X on their pull request while the tab is still open, rather than an hour later when they go to check how the full test suite did. Feel free to close though if you think the current system makes more sense.

MarcoFalke:
For some reason the red X is not shown as soon as one of the entries in the build matrix failed. It only shows when all entries are finished with evaluation. So we'd need to address that first, if possible. Otherwise the new entry in the build matrix won't help.

jnewbery:
Concept ACK. I like the idea of running the static tools separately, both to make it more obvious where problems are and potentially for faster feedback.

> For some reason the red X is not shown as soon as one of the entries in the build matrix failed. It only shows when all entries are finished with evaluation.

[...]

PR #11260, opened 2017-09-06
travis: Assert default datadir isn't created, Run scripted diff only once

maflcko:
It is sufficient to check the scripted diffs on one arch, i.e. `CHECK_DOC`==1.

Also, the default datadir should not be created by just running the tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -43,9 +43,9 @@ install:
     - if [ -n "$PACKAGES" ]; then travis_retry sudo apt-get update; fi
     - if [ -n "$PACKAGES" ]; then travis_retry sudo apt-get install --no-install-recommends --no-upgrade -qq $PACKAGES; fi
 before_script:
-    - if [ "$TRAVIS_EVENT_TYPE" = "pull_request" ]; then contrib/devtools/commit-script-check.sh $TRAVIS_COMMIT_RANGE; fi
-    - unset CC; unset CXX
+    - if [ "$CHECK_DOC" = 1 -a "$TRAVIS_EVENT_TYPE" = "pull_request" ]; then contrib/devtools/commit-script-check.sh $TRAVIS_COMMIT_RANGE; fi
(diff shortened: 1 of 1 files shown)


MarcoFalke:
With "SHA" I meant "scritped diff", whatever.

practicalswift:
Concept ACK regarding the assertion

Is it due to performance reasons we want to run scripted diff only for the `arm-linux-gnueabihf` build (`CHECK_DOC=1`)? One benefit of doing the scripted diff for all builds is to make it apparent from the build status overview that all builds fails. If only `arm-linux-gnueabihf` is red when a scripted diff fails it might look like a platform specific issue and hence require further investigation to see that it is a general scripted diff failure.

This argument could be raised for all checks that for performance reasons are specific to a given Travis build, but scripted diff should be fairly quick in the general case?

MarcoFalke:
No need to run an arch independent check on each arch. If you feel like factoring out the CHECK_DOC into a different build stage (c.f. https://github.com/bitcoin/bitcoin/issues/11004#issuecomment-321404956); Go for it!

practicalswift:
@MarcoFalke Makes sense! Factoring out `CHECK_DOC` into a different build stage sounds like a good plan.

Concept ACK :-)

MarcoFalke:
So without 5368614aa106b96e02ffb9e68eb86ebe197bc07f, it actually fails: https://travis-ci.org/MarcoFalke/bitcoin/jobs/272689632

laanwj:
Concept ACK, will leave review of the travis changes to @theuni

MeshCollider:
Concept ACK, the Travis changes look right but I don't know enough about it to say for sure :)

theuni (approved):
Looks good to me. utACK fa40b0eb514e77928a13a007df170e5ede18ee3e

laanwj:
Thanks, will merge after the general issue with travis is resolved.

MarcoFalke:
This is orthogonal to the travis issue and should not affect it.

PR #11261, opened 2017-09-06
scripted-diff: Use <cxxx> instead of deprecated <xxx.h> when including C compatibility headers

practicalswift:
Use `<cxxx>` instead of deprecated `<xxx.h>` when including C compatibility headers.

Includes changed:

```
<assert.h>
<errno.h>
<limits.h>
<math.h>
<signal.h>
<stdarg.h>
<stdint.h>
<stdio.h>
<stdlib.h>
<string.h>
```

### DIFF
--- a/doc/developer-notes.md
+++ b/doc/developer-notes.md
@@ -441,6 +441,8 @@ Source code organization
     failures when those indirect dependencies change. Furthermore, it obscures what the real code
     dependencies are.
 
+- Use `#include <cxxx>` instead of the deprecated form `#include <xxx.h>` when including C compatibility headers (such as `<cstdint>`, `<cstdlib>`, `<cstring>`, etc.).
+
 - Don't import anything into the global namespace (`using namespace ...`). Use
   fully specified types such as `std::string`.
 
--- a/src/addrman.h
+++ b/src/addrman.h
@@ -15,7 +15,7 @@
 
 #include <map>
 #include <set>
-#include <stdint.h>
+#include <cstdint>
 #include <vector>
 
 /**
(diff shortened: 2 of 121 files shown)


laanwj:
Another huge diff. Are there any immediate drawbacks from this, except consistency?
Also if we decide to do this it needs mention in the developer notes.

practicalswift:
@laanwj In addition to the consistency argument you raised, it could be argued that we should try to avoid deprecated stuff in general :-)  The use of the `.h` form for C compatibility headers is deprecated in C++11. I'll add a note about it in the developer notes.

To make reviewing easy I made this change as a scripted-diff.

practicalswift:
Added a note about the preference of `#include <cxx>` in the developer notes.

Another argument for normalization is to allow for easy detection of duplicate includes and quick grepping for all usages of say `<csignal>` (instead of having to search for both `<signal.h>` and `<csignal>`).

dcousens (approved):

danra:
NACK
I would love to see this change. However, it would also require using all included functionality through the std namespace, since the non-legacy headers aren't guaranteed to contain it in the global namespace.
If you can get the bigger change accepted (prefixing included things with `std::` / adding the appropriate `using` declarations where needed) I would gladly support it.

practicalswift:
@danra I'm afraid such prefixing would produce a gigantic diff. Just changing `{,u}int64_t` to `std::{u,}int64_t` would touch ~2000 lines :-)

danra:
@practicalswift You can use non-namespace `using` declarations. But regardless, I doubt this kind of big change would be approved.
But given that, doing your proposed change would make the code less compliant with the C++ standard, not more so.
In fact, to maximize portability and compliance with the standard, there are probably many instances of `#include <cxxx>` in the project which should be replaced with `#include <xxx.h>`, which is guaranteed to have its functionality visible in both the `std` and the global namespaces.

practicalswift:
Closing this PR due to failure to reach ACK consensus :-)

PR #11274, opened 2017-09-07
[tests] Cleanup wildcard imports in functional tests

mess110:
* Unless it makes sense (p2p-compactblocks, p2p-fullblocktest, p2p-segwit, replace-by-fee)
* Also remove double new lines
* Remove whitespaces


laanwj:
Thanks for your contribution.

However this was discussed before, last time in #9876, and back then it was decided not to do this.

MarcoFalke:
Thanks. Though, this looks like a duplicate of  #10366

mess110:
Always search before doing work, always search before doing work, always sea...

Issue #11257, opened 2017-09-06
[RPC] Bug: RPC Batching does not work with more than 1 multi wallet 


NicolasDorier:
***UPDATE:***

Turns out the issue was caused not by testnet, but by having 2 wallets @ryanofsky found out !

----------

Step to reproduce:

This work on regtest just fine:
```
bitcoind -wallet=test -regtest -rpcpassword=abc -rpcuser=abc
```

Try to make a batch RPC request to /wallet/test:

```
[{"jsonrpc":"1.0","id":1,"method":"getbalance","params":["*"]}]
```
You get the expected result.

However; same things with testnet:
```
bitcoind -wallet=test -testnet -rpcpassword=abc -rpcuser=abc
```

Result:

"Wallet file not specified (must request wallet RPC through /wallet/<filename> uri-path)."

This problem is limited to batched RPC request on testnet as just making the exact same request without the '[' and ']' works fine. It works also fine on regtest, which is crazy. I will look into later.

sipa:
Very strange! Does it work if you use the wallet URL?

ryanofsky:
Seems to be a bug in JSONRPCExecOne. It is creating a request object without assigning a url.

sipa:
Is this a 0.15.0 issue?

ryanofsky:
> Is this a 0.15.0 issue?

Actually, I'm not sure the bug I see in JSONRPCExecOne is the same one Nicolas is reporting. The bug in JSONRPCExecOne (which is only called for batch requests and not single requests) is that is that request uri field is not filled, so if more than one wallet file is loaded, "Wallet file not specified" errors will trigger when a wallet RPC is called. This is a bug in 0.15, but not a regression because previous releases did not support loading multiple wallets.

Nicolas's issue is more confusing because despite multiwallet being mentioned in issue title, steps to reproduce only show bitcoind being started with a single -wallet argument. Also, strangely the bug happens with testnet but not regtest.

NicolasDorier:
Sorry I was in holidays @ryanofsky . This bug came into a "production" testnet. I confirm that my bitcoind has two wallets.

NicolasDorier:
@sipa yes, 0.15.0 issue. Hopefully ryanosky's fix can make it in time. Testing it now.

NicolasDorier:
Tested, fixed by https://github.com/bitcoin/bitcoin/pull/11277

sipa:
Please don't close issues until the fix is actually merged.

PR #11277, opened 2017-09-07
Fix uninitialized URI in batch RPC requests

ryanofsky:
This fixes "Wallet file not specified" errors when making batch wallet RPC calls with more than one wallet loaded. This issue was reported by @NicolasDorier in https://github.com/bitcoin/bitcoin/issues/11257

Request URI is not used for anything except multiwallet request dispatching, so this change has no other effect.

### DIFF
--- a/src/httprpc.cpp
+++ b/src/httprpc.cpp
@@ -192,7 +192,7 @@ static bool HTTPReq_JSONRPC(HTTPRequest* req, const std::string &)
 
         // array of requests
         } else if (valRequest.isArray())
-            strReply = JSONRPCExecBatch(valRequest.get_array());
+            strReply = JSONRPCExecBatch(jreq, valRequest.get_array());
         else
             throw JSONRPCError(RPC_PARSE_ERROR, "Top-level object parse error");
 
--- a/test/functional/test_framework/authproxy.py
+++ b/test/functional/test_framework/authproxy.py
@@ -138,17 +138,20 @@ class AuthServiceProxy(object):
             self.__conn.request(method, path, postdata, headers)
             return self._get_response()
 
-    def __call__(self, *args, **argsn):
+    def get_request(self, *args, **argsn):
         AuthServiceProxy.__id_count += 1
 
         log.debug("-%s-> %s %s"%(AuthServiceProxy.__id_count, self._service_name,
                                  json.dumps(args, default=EncodeDecimal, ensure_ascii=self.ensure_ascii)))
         if args and argsn:
             raise ValueError('Cannot handle both named and positional arguments')
-        postdata = json.dumps({'version': '1.1',
-                               'method': self._service_name,
-                               'params': args or argsn,
-                               'id': AuthServiceProxy.__id_count}, default=EncodeDecimal, ensure_ascii=self.ensure_ascii)
+        return {'version': '1.1',
+                'method': self._service_name,
+                'params': args or argsn,
+                'id': AuthServiceProxy.__id_count}
+
+    def __call__(self, *args, **argsn):
+        postdata = json.dumps(self.get_request(*args, **argsn), default=EncodeDecimal, ensure_ascii=self.ensure_ascii)
         response = self._request('POST', self.__url.path, postdata.encode('utf-8'))
         if response['error'] is not None:
             raise JSONRPCException(response['error'])
@@ -158,7 +161,7 @@ class AuthServiceProxy(object):
         else:
             return response['result']
 
-    def _batch(self, rpc_call_list):
+    def batch(self, rpc_call_list):
         postdata = json.dumps(list(rpc_call_list), default=EncodeDecimal, ensure_ascii=self.ensure_ascii)
         log.debug("--> "+postdata)
         return self._request('POST', self.__url.path, postdata.encode('utf-8'))
--- a/src/rpc/server.cpp
+++ b/src/rpc/server.cpp
@@ -389,11 +389,10 @@ bool IsDeprecatedRPCEnabled(const std::string& method)
     return find(enabled_methods.begin(), enabled_methods.end(), method) != enabled_methods.end();
 }
 
-static UniValue JSONRPCExecOne(const UniValue& req)
+static UniValue JSONRPCExecOne(JSONRPCRequest jreq, const UniValue& req)
 {
     UniValue rpc_result(UniValue::VOBJ);
 
-    JSONRPCRequest jreq;
     try {
         jreq.parse(req);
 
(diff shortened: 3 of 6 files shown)


test/functional/test_framework/authproxy.py
```diff
@@ -138,17 +138,20 @@ def _request(self, method, path, postdata):
             self.__conn.request(method, path, postdata, headers)
             return self._get_response()
 
-    def __call__(self, *args, **argsn):
+    def _call(self, *args, **argsn):
```
MarcoFalke: Rename this to _prepare_post_dict or something?

NicolasDorier:
I really want that for 0.15, will test soon.

NicolasDorier:
tACK c0c652ccc24f4c3dbd200f14822f10c373a5198d @laanwj I hope this can make it to 0.15.
I enjoy multi-wallet, but I really can't use it in any project with batch support, as it would require me to rewrite things that already work.

jnewbery:
I don't like the fact that you're creating a 'private' `_call` method in the `AuthServiceProxy` class, and then calling it (and the existing private method `_batch`) from a test script.

One of the advantages of the `TestNode` class is that on the whole the test scripts don't interface directly with the `AuthServiceProxy` class. That was deliberate and would allow us to:
- remove the slightly cludgy `AuthServiceProxyWrapper` coverage class easily
- modify the auth service proxy or even swap it out for a new implementation without having to change the scripts.

Is there a way you can add the batch functionality to the `TestNode` class?

src/httprpc.cpp
```diff
@@ -192,7 +192,7 @@ static bool HTTPReq_JSONRPC(HTTPRequest* req, const std::string &)
 
         // array of requests
         } else if (valRequest.isArray())
-            strReply = JSONRPCExecBatch(valRequest.get_array());
```
promag: Kind of unrelated, but above in L186 we could call `JSONRPCExecOne` I believe.

JeremyRubin:
needs rebase! utack c0c652c

ryanofsky (commented):
> I don't like the fact that you're creating a 'private' _call method in the AuthServiceProxy class, and then calling it (and the existing private method _batch) from a test script.
>
> One of the advantages of the TestNode class is that on the whole the test scripts don't interface directly with the AuthServiceProxy class. That was deliberate and would allow us to:
> remove the slightly cludgy AuthServiceProxyWrapper coverage class easily
modify the auth service proxy or even swap it out for a new implementation without having to change the scripts.
>
> Is there a way you can add the batch functionality to the TestNode class?

All I'm really doing here is adding a simple 3-line test to multiwallet.py using the existing _batch method:

```python
batch = w1._batch([w1.getblockchaininfo._get_request(), w1.getwalletinfo._get_request()])
assert_equal(batch[0]["result"]["chain"], "regtest")
assert_equal(batch[1]["result"]["walletname"], "w1")
```

This only required two minor improvements to AuthServiceProxy and AuthServiceProxyWrapper, so I think the approach is good for this PR, which is supposed to be an RPC bugfix and not a test framework restructuring.

But to discuss possible future test framework improvements, TestNode seems does not seem like a great place to move RPC sending code. The division where authproxy handles json & http stuff and test_node handles test stuff seems like a good one, and I think don't think it'd be desirable to scramble it all up without a clear motivation. I also don't think that batch rpc requests should follow a different code path from single rpc requests.

As far as I know, nothing in this PR makes it more difficult to remove AuthServiceProxyWrapper in the future if that is a goal. The actual change I'm making to AuthServiceProxyWrapper is a bugfix and cleanup that makes the class less invasive.

If calling methods that begin with underscores is a problem (and I think it's a pretty superficial problem if it is a problem at all), we could simply drop the underscores, or make get_request and batch be standalone functions instead of member functions, or be operators like % and <<.

ryanofsky:
Added 2 commits 236a4378a388395c3b9a85e4a908bee6a6c58b4f -> 0611fae0b9891869cd9aaa9ddd1311dd63baa999 ([pr/mb.4](https://github.com/ryanofsky/bitcoin/commits/pr/mb.4) -> [pr/mb.5](https://github.com/ryanofsky/bitcoin/commits/pr/mb.5), [compare](https://github.com/ryanofsky/bitcoin/compare/pr/mb.4...pr/mb.5))

I went ahead and added % and << operators for making batch rpc calls in a new commit. I realized batch rpc calls weren't being included in coverage logs, and operator syntax seemed like a less invasive way for AuthServiceProxyWrapper to be able to log the calls.

jnewbery:
> If calling methods that begin with underscores is a problem (and I think it's a pretty superficial problem if it is a problem at all), we could simply drop the underscores, or make get_request and batch be standalone functions instead of member functions, or be operators like % and <<.

The new operator overrides are cute, but perhaps unnecessary.

My objection to using 'private' methods in the test script is simply one of convention. Obviously leading underscores don't mean anything to the python interpreter, but to us humans it means 'this method/member is private and won't be accessed directly outside of this class/module'. There's no need to break that assumption needlessly.

> But to discuss possible future test framework improvements, TestNode seems does not seem like a great place to move RPC sending code.

I disagree, but like you say this is a PR for fixing a bitcoind bug and not test framework restructuring. This PR does seem to be an improvement to the AuthServiceProxyWrapper class.

Concept ACK with a weak preference to remove the syntactic sugar and rename the methods to make them 'public'.

ryanofsky:
Updated 3ca5398f2efb3a09b9be65b6ee4aa7acff2c3cd9 -> 4526d21e52aa94f12121fcf01047c04f82c4990a ([pr/mb.5](https://github.com/ryanofsky/bitcoin/commits/pr/mb.5) -> [pr/mb.6](https://github.com/ryanofsky/bitcoin/commits/pr/mb.6)) to get rid of operator overloading.

JeremyRubin:
Chiming in to say that _call is fine -- it's just to indicate that it's an intended for internal use function in python, 'truly private' methods are double underscore.

jnewbery:
>  'truly private' methods are double underscore.

There's no such thing as truly private methods in Python: https://docs.python.org/3/tutorial/classes.html#private-variables

JeremyRubin:
I mean, it's semantics, but technically neither does C++ because you can get around it (in some cases; like without private static globals?) as follows:

```c++
struct A {
private:
void foo() {}
}
struct B {
void foo() {}
}

auto x = new A{};
((B*) x)->foo();
```

In any case, the scare quotes were around 'truly private': it's not truly private, but it does make you call a mangled name. Then again, not many languages stop you from manipulating an object in your memory space if you want to manipulate it.

laanwj:
@JeremyRubin That only works for virtual methods, AFAIK, and it works because the slot is at the same offset in the vtable. Non-virtual methods have the class name encoded into the symbol name

That said, let's try to respect private methods even though it's not perfectly enforced, it's meant as a convention to communicate to other people working on the code that the method shouldn't be used externally...

sipa:
@JeremyRubin Even if it works, I'm pretty sure that it's UB.

NicolasDorier:
Can we add milestone 0.15.1?

luke-jr (approved):
tACK

laanwj:
utACK 4526d21

ruimarinho:
Tested and working as expected. Mixing node with non-node requests now works fine. @laanwj any chance of backporting this to 0.15.1?

MarcoFalke:
@ruimarinho Thanks for the ping. Somehow the backport label was missing and I forgot to include it in #11447

MarcoFalke:
Removing backport tag: #11647

